PR #647, opened 2019-05-08
ICU-20567 Use 7Zip to package Windows binaries on CI build bots

daniel-ju:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20567
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #650, opened 2019-05-10
ICU-20549 Publish Windows build artifacts from CI builds (x64 Release)

jefgen:
This runs the `distrelease.ps1` script to package up the Windows binaries on the Azure CI build machines (for x64 Release) into a zip file, and then publishes the file as a build "artifact" for the pipline.

This will allow us to use the generated binaries from the CI build machines for the BRS tasks.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20549
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
       env:
         BUILD: ICU4J
     - script: |
-        cat `find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;`
+        cd icu4j && cat `find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;`
       condition: failed() # only run if the build fails.
       displayName: 'List failures (if any)'
 #-------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jefgen:
@daniel-ju  Please take a look. 
I think we should try to merge this PR first, then we can see if PR #647 works on the build bots.

.ci-builds/.azure-pipelines.yml
```diff
@@ -58,6 +58,18 @@ jobs:
       inputs:
         filename: icu4c/source/allinone/icucheck.bat
         arguments: 'x64 Release'
+    - task: PowerShell@2
+      displayName: 'PowerShell: Distrelease script'
+      inputs:
+        targetType: filePath
+        filePath: 'icu4c/packaging/distrelease.ps1'
+        arguments: '-arch x64'
+        workingDirectory: icu4c
```
daniel-ju: Just checking, does this mean the script will be run from the `icu4c` directory?

jefgen: Right. It effectively does a `cd` to this directory before running the script, just like the instructions say to do. :)

PR #651, opened 2019-05-11
ICU-12760 support UDISPCTX_TYPE_SUBSTITUTE_HANDLING

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-12760
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/locdspnm.cpp
+++ b/icu4c/source/common/locdspnm.cpp
@@ -291,6 +291,7 @@ class LocaleDisplayNamesImpl : public LocaleDisplayNames {
     UnicodeString formatCloseParen;
     UnicodeString formatReplaceCloseParen;
     UDisplayContext nameLength;
+    UDisplayContext substitute;
 
     // Constants for capitalization context usage types.
     enum CapContextUsage {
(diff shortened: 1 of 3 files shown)


icu4c/source/common/locdspnm.cpp
```diff
@@ -560,6 +572,7 @@ LocaleDisplayNamesImpl::adjustForUsageAndContext(CapContextUsage usage,
 UnicodeString&
 LocaleDisplayNamesImpl::localeDisplayName(const Locale& loc,
                                           UnicodeString& result) const {
+  printf("X\n");
```
fabalbon: Please remove.

icu4c/source/test/intltest/locnmtst.cpp
```diff
@@ -422,4 +423,59 @@ void LocaleDisplayNamesTest::TestRootEtc() {
...
+  test_assert_equal("wxyz", temp);
+  ldn->variantDisplayName("abc", temp);
+  test_assert_equal("abc", temp);
+  ldn->keyDisplayName("efg", temp);
+  test_assert_equal("efg", temp);
+  ldn->keyValueDisplayName("ca", "ijk", temp);
+  test_assert_equal("ijk", temp);
+}
+
+void LocaleDisplayNamesTest::VerifyNoSubstitute(LocaleDisplayNames* ldn) {
```
sffc: Please add a test for some key that is present in root but not in the language bundle.

FrankYFTang: I believe no such thing exist
see https://github.com/unicode-org/icu/blob/master/icu4c/source/data/lang/root.txt
and
https://github.com/unicode-org/icu/blob/master/icu4c/source/data/region/root.txt

sffc: OK.

PR #652, opened 2019-05-15
ICU-20602 add LocaleBuilder::copyErrorTo

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20602
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/localebuilder.cpp
+++ b/icu4c/source/common/localebuilder.cpp
@@ -433,4 +433,13 @@ Locale LocaleBuilder::build(UErrorCode& errorCode)
     return product;
 }
 
+UBool LocaleBuilder::copyErrorTo(UErrorCode &outErrorCode) {
+    if (U_FAILURE(outErrorCode)) {
+        // Do not overwrite the older error code
+        return TRUE;
+    }
+    outErrorCode = status_;
+    return U_FAILURE(outErrorCode);
+}
+
 U_NAMESPACE_END
--- a/icu4c/source/common/unicode/localebuilder.h
+++ b/icu4c/source/common/unicode/localebuilder.h
@@ -276,6 +276,17 @@ public:
      */
     Locale build(UErrorCode& status);
 
+    /**
+     * Sets the UErrorCode if an error occurred while recording sets.
+     * Preserves older error codes in the outErrorCode.
+     * @param outErrorCode Set to an error code that occurred while setting subtags.
+     *                  Unchanged if there is no such error or if outErrorCode
+     *                  already contained an error.
+     * @return TRUE if U_FAILURE(outErrorCode)
+     * @draft ICU 65
+     */
+    UBool copyErrorTo(UErrorCode &outErrorCode);
+
 private:
     UErrorCode status_;
     char language_[9];
--- a/icu4c/source/test/intltest/localebuildertest.cpp
+++ b/icu4c/source/test/intltest/localebuildertest.cpp
@@ -55,8 +55,19 @@ void LocaleBuilderTest::runIndexedTest( int32_t index, UBool exec, const char* &
 
 void LocaleBuilderTest::Verify(LocaleBuilder& bld, const char* expected, const char* msg) {
     UErrorCode status = U_ZERO_ERROR;
+    UErrorCode copyStatus = U_ZERO_ERROR;
+    UErrorCode errorStatus = U_ILLEGAL_ARGUMENT_ERROR;
+    if (bld.copyErrorTo(copyStatus)) {
+        errln(msg, u_errorName(copyStatus));
+    }
+    if (!bld.copyErrorTo(errorStatus) || errorStatus != U_ILLEGAL_ARGUMENT_ERROR) {
+        errln("Should always get the previous error and return FALSE");
+    }
     Locale loc = bld.build(status);
     if (U_FAILURE(status)) {
+        errln(msg, u_errorName(status));
+    }
+    if (status != copyStatus) {
         errln(msg, u_errorName(status));
     }
     std::string tag = loc.toLanguageTag<std::string>(status);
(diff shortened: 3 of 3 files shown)


icu4c/source/common/localebuilder.cpp
```diff
@@ -433,4 +433,9 @@ Locale LocaleBuilder::build(UErrorCode& errorCode)
     return product;
 }
 
+UBool LocaleBuilder::copyErrorTo(UErrorCode &outErrorCode) {
```
sffc: This does not

> Preserves older error codes in the outErrorCode.

Please add

```cpp

        if (U_FAILURE(outErrorCode)) {
            // Do not overwrite the older error code
            return TRUE;
        }
```

icu4c/source/common/localebuilder.cpp
```diff
@@ -433,4 +433,9 @@ Locale LocaleBuilder::build(UErrorCode& errorCode)
...
+    outErrorCode = status_;
+    return U_FAILURE(outErrorCode) ? TRUE : FALSE;
```
sffc: U_FAILURE already returns a UBool.  No need to pass it through another ternary.

icu4c/source/test/intltest/localebuildertest.cpp
```diff
@@ -190,6 +197,9 @@ void LocaleBuilderTest::TestLocaleBuilder() {
         status = U_ZERO_ERROR;
         bld.clear();
         while (true) {
+            status = U_ZERO_ERROR;
+            UErrorCode copyStatus = U_ZERO_ERROR;
+            bld.copyErrorTo(copyStatus);
```
sffc: You copy the error before you start mutating the LocaleBuilder.  Wouldn't you want to put this after you call all the various builder methods?

I would also like to see a concrete test where the error code is not just U_ZERO_ERROR.  Give it bad input on purpose, and check that the error code is as expected.  Maybe you already have that, but it's not obvious from looking at the test.

icu4c/source/test/intltest/localebuildertest.cpp
```diff
@@ -55,10 +55,17 @@ void LocaleBuilderTest::runIndexedTest( int32_t index, UBool exec, const char* &
 
 void LocaleBuilderTest::Verify(LocaleBuilder& bld, const char* expected, const char* msg) {
     UErrorCode status = U_ZERO_ERROR;
```
sffc: FYI for the future: in tests, you can and should use IcuTestErrorCode from testlog.h.

icu4c/source/common/localebuilder.cpp
FrankYFTang: DONE

icu4c/source/common/localebuilder.cpp
FrankYFTang: DONE

FrankYFTang:
PTAL

FrankYFTang:
> I don't see:
> 
> > I would also like to see a concrete test where the error code is not just U_ZERO_ERROR. Give it bad input on purpose, and check that the error code is as expected. Maybe you already have that, but it's not obvious from looking at the test.

Inside the LocaleBuilderTest::Verify function

sffc:
Ok.  Trying to read your verify function.  So you have data with "X", and then you have

```cpp
            else if (strcmp("X", method) == 0) {
                if (U_SUCCESS(status)) {
                    errln("FAIL: No error return - test case: %s", actions.c_str());
                }
```

And then you also have

```cpp
            if (status != copyStatus) {
                errln("copyErrorTo not matching");
            }
```

However, between those two lines, you have

```cpp
            if (U_FAILURE(status)) {
                if (strcmp("X", testCase[i]) == 0) {
                    // This failure is expected
                    break;
```

Can you verify that `copyStatus == status` also in the failure case?

FrankYFTang:
> Ok. Trying to read your verify function. So you have data with "X", and then you have
> 
> ```c++
>             else if (strcmp("X", method) == 0) {
>                 if (U_SUCCESS(status)) {
>                     errln("FAIL: No error return - test case: %s", actions.c_str());
>                 }
> ```
> 
> And then you also have
> 
> ```c++
>             if (status != copyStatus) {
>                 errln("copyErrorTo not matching");
>             }
> ```
> 
> However, between those two lines, you have
> 
> ```c++
>             if (U_FAILURE(status)) {
>                 if (strcmp("X", testCase[i]) == 0) {
>                     // This failure is expected
>                     break;
> ```
> 
> Can you verify that `copyStatus == status` also in the failure case?

DONE. PTAL

FrankYFTang:
> It looks like you used the new squash commit button. There was a bug in that button which I fixed yesterday at 9pm PDT, but it looks like you clicked the button before I fixed it.
> 
> Please force-push from your local branch (reset this PR to what it was before you squashed), then you can try the squash flow again.

PTAL

icu4c/source/common/unicode/localebuilder.h
```diff
@@ -276,6 +276,17 @@ class U_COMMON_API LocaleBuilder : public UObject {
      */
     Locale build(UErrorCode& status);
 
+    /**
+     * Sets the UErrorCode if an error occurred while recording sets.
+     * Preserves older error codes in the outErrorCode.
+     * @param outErrorCode Set to an error code if it does not contain one already
+     *                  and an error occurred while recording sets.
```
sffc: Suggested documentation of `@param outErrorCode`:

```
Set to an error code that occurred while setting subtags.
Unchanged if there is no such error or if outErrorCode
already contained an error.
```

FrankYFTang: DONE PTAL

PR #653, opened 2019-05-16
ICU-20608 Add platform defines for Emscripten.

tweenk:
[ICU-20608]

Adds basic platform defines for Emscripten, a C++ toolchain for the Web that can target asm.js and WebAssembly. Note that this does not make icu4c fully build with Emscripten yet.

[ICU-20608]: https://unicode-org.atlassian.net/browse/ICU-20608

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -135,6 +135,14 @@
 /** Fuchsia is a POSIX-ish platform. @internal */
 #define U_PF_FUCHSIA 4100
 /* Maximum value for Linux-based platform is 4499 */
+/**
+ * Emscripten is a C++ transpiler for the Web that can target asm.js or
+ * WebAssembly. It provides some POSIX-compatible wrappers and stubs and
+ * some Linux-like functionality, but is not fully compatible with
+ * either.
+ * @internal
+ */
+#define U_PF_EMSCRIPTEN 5010
 /** z/OS is the successor to OS/390 which was the successor to MVS. @internal */
 #define U_PF_OS390 9000
 /** "IBM i" is the current name of what used to be i5/OS and earlier OS/400. @internal */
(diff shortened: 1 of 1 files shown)


icu4c/source/common/unicode/platform.h
```diff
@@ -139,6 +139,14 @@
...
 /** "IBM i" is the current name of what used to be i5/OS and earlier OS/400. @internal */
 #define U_PF_OS400 9400
+/**
+ * Emscripten is a C++ transpiler for the Web that can target asm.js or
+ * WebAssembly. It provides some POSIX-compatible wrappers and stubs and
+ * some Linux-like functionality, but is not fully compatible with
+ * either.
+ * @internal
+ */
+#define U_PF_EMSCRIPTEN 10000
```
markusicu: I would like a smaller (4-digit) numeric value here. Discussed a bit in the ICU team meeting today; @srl295 will have an opinion too.
- Since this is a little like POSIX and a little like Linux, we could use 3800 (at the end of the Unix-y/BSD-y platforms).
- Or 4500 (after the Linux-y ones).
- Or something in the 5000..8999 range to put more distance between other platforms and this somewhat "weird" one.

The actual numbers don't matter much, except that it's convenient to have similar platforms grouped together for range checks like `elif U_PF_LINUX <= U_PLATFORM && U_PLATFORM <= 4499`.

Steven?

srl295: @markusicu How about in the 5000s: this is a VM, and though it is POSIX 'inspired', a lot of things are probably POSIX 'inspired'.

markusicu: I am fine with "in the 5000s". How about 5010, reserving 5000 for a possible future range-start constant?

tweenk: Changed to 5010.

tweenk:
Updated pull request with the new platform value - sorry for the delay.

PR #655, opened 2019-05-20
ICU-20588 mutex cleanup

aheninger:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20588
- [x] Issue accepted
- [ ] Tests included


aheninger:
Incomplete, not ready for review. Want to see build bot results.

icu4c/source/common/umutex.cpp
```diff
@@ -35,34 +36,98 @@ U_NAMESPACE_BEGIN
...
+
+
+std::mutex *UMutex::getMutex() {
+    std::mutex *retPtr = fMutex.load(std::memory_order_acquire);
+    if (retPtr == nullptr) {
+        std::call_once(*pInitFlag, umtx_init);
+        std::lock_guard<std::mutex> guard(*initMutex);
+        if (fMutex.load() == nullptr) {
+            fMutex = new(fStorage) std::mutex();
+            retPtr = fMutex;
```
axelandrejs: Thank you for fixing this! I like this solution.

icu4c/source/common/umutex.cpp
```diff
@@ -35,34 +36,98 @@ U_NAMESPACE_BEGIN
 #error U_USER_MUTEX_CPP not supported
 #endif
 
+// Check that UMutex is trivially constructable & destructable, which ensures that
+// static instances are not running static constructors or destructors.
+#if 0    // Omit because some build bot compilers are old, and lack support for the test.
```
jefgen: I wonder if maybe this could be skipped for older compilers?
`
#if (defined(__GNUG__) && __GNUC__ < 5) || (defined(__clang__) && __clang_major__ < 5)
// skip
#else
// test
#endif
`

icu4c/source/common/umutex.h
```diff
@@ -186,30 +187,45 @@ template<class T> void umtx_initOnce(UInitOnce &uio, void (U_CALLCONV *fp)(T, UE
...
 
-    std::mutex   fMutex = {};    // Note: struct - pubic members - because most access is from
-    //                           //       plain C style functions (umtx_lock(), etc.)
+    static void cleanup();
+
+private:
+    alignas(std::mutex) char fStorage[sizeof(std::mutex)];
+    std::atomic<std::mutex *> fMutex;
+
+    /** All initialized UMutexes are kept in a linked list, so that they can be found,
```
jefgen: Thank you Andy! This is a neat solution, so that they will also be destroyed in the order they were created too.

icu4c/source/common/cmemory.h
```diff
@@ -101,12 +101,18 @@ typedef union {
...
+ *    myMutex->~mutex();
+ * DO NOT use delete.
+ * DO NOT use with class UMutex, which has specific support for static instances.
+ *
+ * STATIC_NEW is intended for use when
  *   - We want a static (or global) object.
- *   - We don't want it to ever be destructed, to avoid order-of-destruction
- *     or use-after-destruction problems.
- *   - We want to avoid an ordinary heap allocated object, to avoid triggering
- *     memory leak reports, from valgrind, for example.
```
jefgen: Thanks Andy!
If this is merged, then there are a few other places in the ICU code that could likely use this new macro as well. :)

aheninger:
Jeff and Axel, thanks for the earlier comments. I think it's now complete.

PR #656, opened 2019-05-21
ICU-20616 Allow bidi marks around the sign in exponent parsing.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20616
- [ ] Issue accepted
- [x] Tests included


sffc:
That makes sense.  There may be other places in number parsing outside of ScientificMatcher that could benefit from such a change, too.  I filed a new ticket to investigate:

https://unicode-org.atlassian.net/browse/ICU-20619

PR #659, opened 2019-05-25
ICU-20625 Fixing uninitialized memory in DTPG detected by Valgrind.

sffc:
Commit that introduced this (~2 yrs ago): https://github.com/unicode-org/icu/commit/e2e48c9dce85e1a460c32df5e6b8bd84c6fcf0fb

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20625
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -2547,7 +2547,8 @@ UChar SkeletonFields::getFirstChar() const {
 }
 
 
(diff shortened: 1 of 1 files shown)


sffc:
The field `addedDefaultDayPeriod` was not being set in the constructor.  It appears that there are some code paths that read from that field before it is initialized elsewhere.

PR #660, opened 2019-05-25
ICU-20629 Fixing DTPG hour-format locale and region fallback logic.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20629
- [ ] Issue accepted
- [ ] Tests included


roubert:
It looks as if your proposed changes would cause `IntlTestDateTimePatternGeneratorAPI.testC` to fail. See ICU-20431 for the most recent change concerning this test case. I believe that the test case is correct and identical in ICU4J, in which case your proposed change doesn't work as expected. But I know that I don't understand the ICU resource loading logic well enough, so it'd probably be better if you could find a code reviewer who does understand this better than I do.

sffc:
I ran into even more bugs that turned out to be related to this method:

https://unicode-org.atlassian.net/browse/ICU-20640

I pushed some more commits to fix the issues in that bug as well as the test failure.  I will probably commit against the original ticket (ICU-20629) and mark ICU-20640 as Fixed By Other Ticket.

sffc:
Reassigning to @pedberg-icu.

icu4c/source/data/misc/likelySubtags.txt
```diff
@@ -65,6 +65,7 @@ likelySubtags:table(nofallback){
     arn{"arn_Latn_CL"}
     aro{"aro_Latn_BO"}
     arq{"arq_Arab_DZ"}
+    ars{"ars_Arab_SA"}
```
pedberg-icu: The likelySubtags data is auto-generated in CLDR based on the <territoryInfo> data in supplementalData. Please file a CLDR issue to ensure that "ars" data gets added for "SA", otherwise this change will get wiped out with the next CLDR update.

sffc: Already done: https://unicode-org.atlassian.net/browse/CLDR-13071

sffc:
Pushed one more commit to fix a Java test failure and C++ dependencies.  Includes a fix for ICU-9982.

https://unicode-org.atlassian.net/browse/ICU-9982

icu4c/source/data/misc/likelySubtags.txt
pedberg-icu: Thanks

icu4c/source/common/util.h
```diff
@@ -185,6 +185,15 @@ class U_COMMON_API ICU_Utility /* not : public UObject because all methods are s
      */
     static int32_t parseInteger(const UnicodeString& rule, int32_t& pos, int32_t limit);
 
+    /**
+     * Parse an integer at pos using only ASCII digits.
+     * Base 10 only.
+     * @param pos INPUT-OUTPUT parameter.  On input, the first
+     * character to parse.  On output, the character after the last
+     * parsed character.
+     */
```
pedberg-icu: offset (or position, or something), not character.
	     * @param pos INPUT-OUTPUT parameter.  On input, the offset
	     * of the first character to parse.  On output, the offset after the
	     * last parsed character.

sffc: Done.  I copied and pasted it from the other function.  I changed it here and in the old function.

sffc:
See also https://github.com/unicode-org/cldr/pull/39

sffc:
Tests are passing.  Squash coming up.

sffc:
@pedberg-icu I addressed your feedback and squashed.  PTAL

sffc:
Tests are passing.  Just need a stamp.

icu4c/source/common/util.h
pedberg-icu: Well, OK. There might not be a character after the last parsed character, so it is really just the index after the last parsed character. But this is okay for now.

PR #642, opened 2019-05-04
ICU-20593 Buildtool: adding some options and improving exec mode.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20593
- [x] Updated PR title and link in previous line to include Issue number
- [ ] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #661, opened 2019-05-25
ICU-20627 Adding alias locales to res_index (ignore xml deprecates).

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20627
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/python/icutools/databuilder/__main__.py
+++ b/icu4c/source/python/icutools/databuilder/__main__.py
@@ -84,12 +84,6 @@ flag_parser.add_argument(
     action = "store_true"
 )
 flag_parser.add_argument(
-    "--ignore_xml_deprecates",
-    help = "Whether to ignore XML deprecates files for building res_index.",
-    default = False,
-    action = "store_true"
-)
-flag_parser.add_argument(
     "--seqmode",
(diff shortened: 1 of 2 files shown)


icu4c/source/python/icutools/databuilder/__main__.py
```diff
@@ -83,12 +83,6 @@
     default = False,
     action = "store_true"
 )
-flag_parser.add_argument(
-    "--ignore_xml_deprecates",
```
markusicu: Should we keep obsolote options for one or more releases as no-ops, with the help text giving the status?
It does not seem necessary to break users who specify the option.
Or better to make them fail fast so they know this is no longer an option?

sffc: I just added the option in #642.  It isn't in ICU 64.2.

sffc: I added the option in #642 to give users (Google) a way to disable the feature.  But now that I'm proposing to make this the default behavior, I'm deleting the option I added.  If we wanted to give a way for users to get the old behavior, we could make an option named `--use-xml-deprecates`.

sffc:
I will look at the test failures on Tuesday.

icu4c/source/python/icutools/databuilder/__main__.py
markusicu: ok no concern then. tnx.

PR #674, opened 2019-06-05
ICU-20654 Add hour/minute for "this hour/minute"

FrankYFTang:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20654
- [x] Issue accepted
- [x] Tests included


FrankYFTang:
> Please fix the broken tests.

DONE.

sffc:
LGTM, please squash.

FrankYFTang:
> LGTM, pending API approval. Thanks for fixing both C and J

API got approved by TC, could you approve this PR now? Thanks

FrankYFTang:
> Sorry. I just noticed that you didn't add a test in C++ that uses the new constants. You did in Java. Can you add a test in C++ that directly uses `UDAT_ABSOLUTE_HOUR` and `UDAT_ABSOLUTE_MINUTE`, maybe inside of a big parameterized test like kEnglishNoQuantity?
> 
> Also, I notice that you added UDAT_ABSOLUTE_HOUR and _MINUTE but not _SECOND. Is that intentional?

ok, i will add test in C. There are no need to have _SECOND because there are no "this second", but only "now" and that is already implemented.

PR #680, opened 2019-06-10
ICU-20659 Fix DTD link in XML data files

gvictor:
- http://www.unicode.org/repos/cldr/trunk/common/dtd/ldml.dtd returns
  HTTP 302 error and redirects to an html page, not a dtd content
  apparently.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20659
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/data/xml/brkitr/de.xml
+++ b/icu4c/source/data/xml/brkitr/de.xml
@@ -6,9 +6,9 @@
 
  See http://bugs.icu-project.org/trac/ticket/10750#comment:5 - this file is needed to emit break iterator rules properly.
 -->
-<!DOCTYPE ldml SYSTEM "http://www.unicode.org/repos/cldr/trunk/common/dtd/ldml.dtd"
+<!DOCTYPE ldml SYSTEM "../../dtd/cldr-35.1/common/dtd/ldml.dtd"
 [
-   <!ENTITY % icu SYSTEM "http://www.unicode.org/repos/cldr/trunk/common/dtd/ldmlICU.dtd">
+   <!ENTITY % icu SYSTEM "../../dtd/cldr-35.1/common/dtd/ldmlICU.dtd">
    %icu;
 ]
 >
--- a/icu4c/source/data/dtd/cldr-35.1/common/dtd/ldml.dtd
+++ b/icu4c/source/data/dtd/cldr-35.1/common/dtd/ldml.dtd
@@ -1,0 +1,3146 @@
+<!--
+Copyright © 2003-2019 Unicode, Inc. and others. All rights reserved. Distributed under the Terms of Use in http://www.unicode.org/copyright.html.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of the Unicode data files and any associated documentation (the "Data Files") or Unicode software and any associated documentation (the "Software") to deal in the Data Files or Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, and/or sell copies of the Data Files or Software, and to permit persons to whom the Data Files or Software are furnished to do so, provided that (a) the above copyright notice(s) and this permission notice appear with all copies of the Data Files or Software, (b) both the above copyright notice(s) and this permission notice appear in associated documentation, and (c) there is clear notice in each modified Data File or in the Software as well as in the documentation associated with the Data File(s) or Software that the data or software has been modified.
+
+THE DATA FILES AND SOFTWARE ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT OF THIRD PARTY RIGHTS. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR HOLDERS INCLUDED IN THIS NOTICE BE LIABLE FOR ANY CLAIM, OR ANY SPECIAL INDIRECT OR CONSEQUENTIAL DAMAGES, OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THE DATA FILES OR SOFTWARE.
+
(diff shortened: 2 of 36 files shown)


icu4c/source/data/xml/brkitr/de.xml
```diff
@@ -6,9 +6,9 @@
 
  See http://bugs.icu-project.org/trac/ticket/10750#comment:5 - this file is needed to emit break iterator rules properly.
 -->
-<!DOCTYPE ldml SYSTEM "http://www.unicode.org/repos/cldr/trunk/common/dtd/ldml.dtd"
+<!DOCTYPE ldml SYSTEM "https://raw.githubusercontent.com/unicode-org/cldr/master/common/dtd/ldml.dtd"
```
srl295: Probably should be a specific version X instead of master. Or perhaps, copy version X INTO ICU's source base, and use that. I'm not sure if using this URL is the best practice.

icu4c/source/data/xml/brkitr/de.xml
```diff
@@ -6,9 +6,9 @@
...
 [
-   <!ENTITY % icu SYSTEM "http://www.unicode.org/repos/cldr/trunk/common/dtd/ldmlICU.dtd">
+   <!ENTITY % icu SYSTEM "https://raw.githubusercontent.com/unicode-org/cldr/master/common/dtd/ldmlICU.dtd">
```
srl295: no…  move ldmlICU.dtd out of CLDR (new ticket required) and move it into the ICU source base. Then make this a relative link.

icu4c/source/data/xml/brkitr/de.xml
srl295: I kind of think we should basically "vendor" the CLDR dtd here, only for this use. And change this to a relative link:   `../dtd/cldr-35/ldml.dtd`. And then NOT update that dtd, ever, unless these override files actually depend on something that changes.

sffc: The status quo is that it points to CLDR SVN Trunk.  So, pointing to CLDR GitHub Master is the equivalent.

If there's a way to avoid the need for an HTTP(S) request when running Ldml2IcuConverter, that would be a desirable outcome.  I'm not sure about how we'd keep the ICU copy of ldml.dtd in sync with the CLDR copy, though.  That file does tend to change from version to version I believe.

srl295: It's the status quo, but not a good idea.  CLDR used to have HTTP references long ago also and removed them.

>  I'm not sure about how we'd keep the ICU copy of ldml.dtd in sync with the CLDR copy, though. That file does tend to change from version to version I believe.

So here's the thing. Yes, the dtd changes but that may not mean _the version which ICU points to needs to change_.  From a cursory look, most files haven't had a substantive change in many years, and unlikely a change which is affected by the DTD. We can always upgrade the vendored DTD if/when the need arises.

sffc: It seems like the most "correct" thing to do would be to pull the dtd from the local CLDR clone that is being used when running Ldml2IcuConverter.  Is that something we can do?  Like, could we reference an environment variable like `$CLDR_DIR` in the path?

srl295: Why is that more correct?

sffc: Well actually.  I guess, when running Ldml2IcuConverter, the DTD that applies to the files *in the CLDR project* should be the one from the CLDR project.  But, these files live in ICU.  So, if the DTD were to ever change in a breaking way, the ICU files would need to be updated.

The DTD version from which these files were written should be versioned, and saving the DTD file into ICU is a way to version it.

icu4c/source/data/xml/brkitr/de.xml
gvictor: But seems CLDR project has references to ldmlICU.dtd, and I am not sure if moving it out of CLDR does not break CLDR tools.
https://github.com/unicode-org/cldr/search?q=ldmlICU&type=Code
Can we file a separate ticket for this migration?

icu4c/source/data/xml/brkitr/de.xml
gvictor: Referencing an environment variable like $CLDR_DIR in the path would be ideal for our case, since we, on Android, checkouts both ICU and CLDR directory. But I have no strong opinion here, and "vendor" the CLDR dtd would work for us too.

However, as @sffc said, pointing to CLDR GitHub Master is the equivalent. And the XML file is currently broken because of the CLDR github migration. Can we fix the regression first before any further improvement?

icu4c/source/data/xml/brkitr/de.xml
srl295: CLDR needs to change, and ldml2icu is moving out of CLDR.

icu4c/source/data/xml/brkitr/de.xml
srl295: CLDR's tooling only checks that the path ends in `common/dtd/ldmlICU.dtd`.  so I propose that right now, we *copy* the file in question to `icu/icu4c/source/data/dtd/cldr-35.1/common/dtd/ldmlICU.dtd` and then change the paths to `../../../dtd/cldr-35.1/common/dtd/ldmlICU.dtd`

then removing the file from CLDR can be done as a separate step.

icu4c/source/data/xml/brkitr/de.xml
srl295: CLDR's direct usage is:
- skip this DTD in certain processing
- a test case

that's it.

icu4c/source/data/xml/brkitr/de.xml
srl295: ^ and ldml.dtd similarly

gvictor: Ok. I will update this PR to copy the CLDR version 35.1 of the DTD files to icu/icu4c/source/data/dtd/cldr-35.1/common/dtd/

gvictor: It's now done.

icu4c/source/data/xml/brkitr/de.xml
gvictor: Sorry for late reply.

I have cloned the file into ICU project now, and filed https://unicode-org.atlassian.net/browse/CLDR-13130.

Please take a look.

gvictor:
Hi, Steven and Shane,

I have made the requested change in the force push and resolved the conversation. I still see the error "Merging can be performed automatically once the requested changes are addressed.", and am not sure how github detects "requested changes". Could you check if the change makes sense to you?

Victor

sffc:
@srl295 Can you take a look at Victor's changes?

gvictor:
Could you help merge this? I am not authorised to merge this.

PR #681, opened 2019-06-10
ICU-20622 Fixing several MeasureFormat problems

mihnita:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20622
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/measfmt.cpp
+++ b/icu4c/source/i18n/measfmt.cpp
@@ -55,28 +55,23 @@ UOBJECT_DEFINE_RTTI_IMPLEMENTATION(MeasureFormat)
 class NumericDateFormatters : public UMemory {
 public:
     // Formats like H:mm
-    SimpleDateFormat hourMinute;
+    UnicodeString hourMinute;
 
     // formats like M:ss
-    SimpleDateFormat minuteSecond;
+    UnicodeString minuteSecond;
 
     // formats like H:mm:ss
-    SimpleDateFormat hourMinuteSecond;
+    UnicodeString hourMinuteSecond;
 
     // Constructor that takes the actual patterns for hour-minute,
     // minute-second, and hour-minute-second respectively.
     NumericDateFormatters(
             const UnicodeString &hm,
             const UnicodeString &ms,
-            const UnicodeString &hms,
-            UErrorCode &status) : 
-            hourMinute(hm, status),
-            minuteSecond(ms, status), 
-            hourMinuteSecond(hms, status) {
-        const TimeZone *gmt = TimeZone::getGMT();
-        hourMinute.setTimeZone(*gmt);
-        minuteSecond.setTimeZone(*gmt);
-        hourMinuteSecond.setTimeZone(*gmt);
+            const UnicodeString &hms) :
+            hourMinute(hm),
+            minuteSecond(ms),
+            hourMinuteSecond(hms) {
     }
 private:
     NumericDateFormatters(const NumericDateFormatters &other);
@@ -233,8 +228,7 @@ static NumericDateFormatters *loadNumericDateFormatters(
     NumericDateFormatters *result = new NumericDateFormatters(
         loadNumericDateFormatterPattern(resource, "hm", status),
         loadNumericDateFormatterPattern(resource, "ms", status),
-        loadNumericDateFormatterPattern(resource, "hms", status),
-        status);
+        loadNumericDateFormatterPattern(resource, "hms", status));
     if (U_FAILURE(status)) {
         delete result;
         return NULL;
@@ -706,55 +700,6 @@ UnicodeString &MeasureFormat::formatMeasure(
     return appendTo;
 }
 
-// Formats hours-minutes-seconds as 5:37:23 or similar.
-UnicodeString &MeasureFormat::formatNumeric(
-        const Formattable *hms,  // always length 3
-        int32_t bitMap,   // 1=hourset, 2=minuteset, 4=secondset
-        UnicodeString &appendTo,
-        UErrorCode &status) const {
-    if (U_FAILURE(status)) {
-        return appendTo;
-    }
-    UDate millis = 
-        (UDate) (((uprv_trunc(hms[0].getDouble(status)) * 60.0
-             + uprv_trunc(hms[1].getDouble(status))) * 60.0
-                  + uprv_trunc(hms[2].getDouble(status))) * 1000.0);
-    switch (bitMap) {
-    case 5: // hs
-    case 7: // hms
-        return formatNumeric(
-                millis,
-                cache->getNumericDateFormatters()->hourMinuteSecond,
-                UDAT_SECOND_FIELD,
-                hms[2],
-                appendTo,
-                status);
-        break;
-    case 6: // ms
-        return formatNumeric(
-                millis,
-                cache->getNumericDateFormatters()->minuteSecond,
-                UDAT_SECOND_FIELD,
-                hms[2],
-                appendTo,
-                status);
-        break;
-    case 3: // hm
-        return formatNumeric(
-                millis,
-                cache->getNumericDateFormatters()->hourMinute,
-                UDAT_MINUTE_FIELD,
-                hms[1],
-                appendTo,
-                status);
-        break;
-    default:
-        status = U_INTERNAL_PROGRAM_ERROR;
-        return appendTo;
-        break;
-    }
-}
-
 static void appendRange(
         const UnicodeString &src,
         int32_t start,
@@ -770,71 +715,112 @@ static void appendRange(
     dest.append(src, end, src.length() - end);
 }
 
-// Formats time like 5:37:23
+
+// Formats numeric time duration as 5:00:47 or 3:54.
 UnicodeString &MeasureFormat::formatNumeric(
-        UDate date, // Time since epoch 1:30:00 would be 5400000
-        const DateFormat &dateFmt, // h:mm, m:ss, or h:mm:ss
-        UDateFormatField smallestField, // seconds in 5:37:23.5
-        const Formattable &smallestAmount, // 23.5 for 5:37:23.5
+        const Formattable *hms,  // always length 3
+        int32_t bitMap,   // 1=hour set, 2=minute set, 4=second set
         UnicodeString &appendTo,
         UErrorCode &status) const {
     if (U_FAILURE(status)) {
         return appendTo;
     }
-    // Format the smallest amount with this object's NumberFormat
-    UnicodeString smallestAmountFormatted;
 
-    // We keep track of the integer part of smallest amount so that
-    // we can replace it later so that we get '0:00:09.3' instead of
-    // '0:00:9.3'
-    FieldPosition intFieldPosition(UNUM_INTEGER_FIELD);
-    (*numberFormat)->format(
-            smallestAmount, smallestAmountFormatted, intFieldPosition, status);
-    if (
-            intFieldPosition.getBeginIndex() == 0 &&
-            intFieldPosition.getEndIndex() == 0) {
+    UnicodeString pattern;
+
+    double hours = hms[0].getDouble(status);
+    double minutes = hms[1].getDouble(status);
+    double seconds = hms[2].getDouble(status);
+    if (U_FAILURE(status)) {
+        return appendTo;
+    }
+
+    // All possible combinations: "h", "m", "s", "hm", "hs", "ms", "hms"
+    if (bitMap == 5 || bitMap == 7) { // "hms" & "hs" (we add minutes if "hs")
+        pattern = cache->getNumericDateFormatters()->hourMinuteSecond;
+        hours = uprv_trunc(hours);
+        minutes = uprv_trunc(minutes);
+    } else if (bitMap == 3) { // "hm"
+        pattern = cache->getNumericDateFormatters()->hourMinute;
+        hours = uprv_trunc(hours);
+    } else if (bitMap == 6) { // "ms"
+        pattern = cache->getNumericDateFormatters()->minuteSecond;
+        minutes = uprv_trunc(minutes);
+    } else { // h m s, handled outside formatNumeric. No value is also an error.
         status = U_INTERNAL_PROGRAM_ERROR;
         return appendTo;
     }
 
-    // Format time. draft becomes something like '5:30:45'
-    // #13606: DateFormat is not thread-safe, but MeasureFormat advertises itself as thread-safe.
-    FieldPosition smallestFieldPosition(smallestField);
-    UnicodeString draft;
-    static UMutex dateFmtMutex;
-    umtx_lock(&dateFmtMutex);
-    dateFmt.format(date, draft, smallestFieldPosition, status);
-    umtx_unlock(&dateFmtMutex);
-
-    // If we find field for smallest amount replace it with the formatted
-    // smallest amount from above taking care to replace the integer part
-    // with what is in original time. For example, If smallest amount
-    // is 9.35s and the formatted time is 0:00:09 then 9.35 becomes 09.35
-    // and replacing yields 0:00:09.35
-    if (smallestFieldPosition.getBeginIndex() != 0 ||
-            smallestFieldPosition.getEndIndex() != 0) {
-        appendRange(draft, 0, smallestFieldPosition.getBeginIndex(), appendTo);
-        appendRange(
-                smallestAmountFormatted,
-                0,
-                intFieldPosition.getBeginIndex(),
-                appendTo);
-        appendRange(
-                draft,
-                smallestFieldPosition.getBeginIndex(),
-                smallestFieldPosition.getEndIndex(),
-                appendTo);
-        appendRange(
-                smallestAmountFormatted,
-                intFieldPosition.getEndIndex(),
-                appendTo);
-        appendRange(
-                draft,
-                smallestFieldPosition.getEndIndex(),
-                appendTo);
-    } else {
-        appendTo.append(draft);
+    const DecimalFormat *numberFormatter = dynamic_cast<const DecimalFormat*>(numberFormat->get());
+    if (!numberFormatter) {
+        status = U_INTERNAL_PROGRAM_ERROR;
+        return appendTo;
     }
+    number::LocalizedNumberFormatter numberFormatter2;
+    if (auto* lnf = numberFormatter->toNumberFormatter(status)) {
+        numberFormatter2 = lnf->integerWidth(number::IntegerWidth::zeroFillTo(2));
+    } else {
+        return appendTo;
+    }
+
+    FormattedStringBuilder fsb;
+
+    UBool protect = FALSE;
+    const int32_t patternLength = pattern.length();
+    for (int32_t i = 0; i < patternLength; i++) {
+        char16_t c = pattern[i];
+
+        // Also set the proper field in this switch
+        // We don't use DateFormat.Field because this is not a date / time, is a duration.
+        double value = 0;
+        switch (c) {
+            case u'H': value = hours; break;
+            case u'm': value = minutes; break;
+            case u's': value = seconds; break;
+        }
+
+        // For undefined field we use UNUM_FIELD_COUNT, for historical reasons.
+        // See cleanup bug: https://unicode-org.atlassian.net/browse/ICU-20665
+        // But we give it a clear name, to keep "the ugly part" in one place.
+        constexpr UNumberFormatFields undefinedField = UNUM_FIELD_COUNT;
+
+        // There is not enough info to add Field(s) for the unit because all we have are plain
+        // text patterns. For example in "21:51" there is no text for something like "hour",
+        // while in something like "21h51" there is ("h"). But we can't really tell...
+        switch (c) {
+            case u'H':
+            case u'm':
+            case u's':
+                if (protect) {
+                    fsb.appendCodePoint(c, undefinedField, status);
+                } else {
+                    UnicodeString tmp;
+                    if ((i + 1 < patternLength) && pattern[i + 1] == c) { // doubled
+                        tmp = numberFormatter2.formatDouble(value, status).toString(status);
+                        i++;
+                    } else {
+                        numberFormatter->format(value, tmp, status);
+                    }
+                    // TODO: Use proper Field
+                    fsb.append(tmp, undefinedField, status);
+                }
+                break;
+            case u'\'':
+                // '' is escaped apostrophe
+                if ((i + 1 < patternLength) && pattern[i + 1] == c) {
+                    fsb.appendCodePoint(c, undefinedField, status);
+                    i++;
+                } else {
+                    protect = !protect;
+                }
+                break;
+            default:
+                fsb.appendCodePoint(c, undefinedField, status);
+        }
+    }
+
+    appendTo.append(fsb.toTempUnicodeString());
+
     return appendTo;
 }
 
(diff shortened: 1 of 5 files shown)


mihnita:
This is for Shane.
He already reviewed a version without `FormattedStringBuilder`

The plan is to review and cleanup this one, then do the C/C++ implementation using it as a model.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
-        result.nextFieldPosition(intFieldPosition);
-        smallestAmountFormatted = result.toString();
-        // Give up if there is no integer field.
-        if (intFieldPosition.getBeginIndex() == 0 && intFieldPosition.getEndIndex() == 0) {
-            throw new IllegalStateException();
-        }
+        // We can create it on demand, but all of the patterns (right now) have mm and ss.
+        // So unless it is hours only we will need a 0-padded 2 digits formatter.
+        LocalizedNumberFormatter numberFormatter2 = numberFormatter.integerWidth(IntegerWidth.zeroFillTo(2));
+        FormattedStringBuilder fsb = new FormattedStringBuilder(pattern.length());
```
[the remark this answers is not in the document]
mihnita: Done (going with the default)

Although in Java the StringBuilder resizes the internal buffer as needed, no matter if it was allocated with default or an explicit value.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
-                // Add the rest of the smallest field
-                appendTo.append(smallestAmountFormatted,
-                        intFieldPosition.getEndIndex(),
-                        smallestAmountFormatted.length());
-                appendTo.append(draft, smallestFieldPosition.getEndIndex(), draft.length());
-            } else {
-                // As fallback, just use the formatted duration.
-                appendTo.append(draft);
-            }
+            appendable.append(fsb.toString());
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -823,12 +823,10 @@ private void formatMeasuresSlowTrack(
...
 
     // type is one of "hm", "ms" or "hms"
-    private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String type) {
+    private static String loadNumericDurationFormat(ICUResourceBundle r, String type) {
         r = r.getWithFallback(String.format("durationUnits/%s", type));
         // We replace 'h' with 'H' because 'h' does not make sense in the context of durations.
-        DateFormat result = new SimpleDateFormat(r.getString().replace("h", "H"));
-        result.setTimeZone(TimeZone.GMT_ZONE);
-        return result;
+        return r.getString().replace("h", "H");
```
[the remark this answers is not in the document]
mihnita: To be fair, I think this one is a bit cleaner...

This is in one place, but there are 2 switches looking for 'H', so at some point one might update one and miss the other one.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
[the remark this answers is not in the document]
sffc: LocalizedNumberFormatter also resizes the buffer if needed; it's just best to avoid it if necessary.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
+                            fsb.append(numberFormatter2.format(value), null); // TODO: Use proper Field
+                            i++;
+                        } else {
+                            fsb.append(numberFormatter.format(value), null); // TODO: Use proper Field
+                        }
+                    }
+                    break;
+                case '\'':
+                    // '' is escaped apostrophe
+                    if ((i + 1 < pattern.length()) && pattern.charAt(i + 1) == c) {
```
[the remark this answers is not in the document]
mihnita: Keeping as is, consistent style with the doubled Hms, where we look and skip next

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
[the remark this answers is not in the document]
mihnita: Keeping as is, if you don't feel strongly about it.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
-            draft = formatter.format(duration, new StringBuffer(), smallestFieldPosition).toString();
+            // There is not enough info to add Field(s) for the unit because all we have are plain
+            // text patterns. For example in "21:51" there is no text for something like "hour",
+            // while in something like "21h51" there is ("h"). But we can't really tell...
+            switch (c) {
+                case 'H':
+                case 'm':
+                case 's':
+                    if (protect) {
+                        fsb.append(c.toString(), null);
```
sffc: Please use `fsb.appendCodePoint(c, null)`

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
-            throw new IllegalStateException();
-        }
+        // We can create it on demand, but all of the patterns (right now) have mm and ss.
+        // So unless it is hours only we will need a 0-padded 2 digits formatter.
+        LocalizedNumberFormatter numberFormatter2 = numberFormatter.integerWidth(IntegerWidth.zeroFillTo(2));
+        FormattedStringBuilder fsb = new FormattedStringBuilder();
+
+        boolean protect = false;
+        for (int i = 0; i < pattern.length(); i++) {
+            Character c = pattern.charAt(i);
```
sffc: Can you use type `char` not `Character` to avoid the boxing?

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
+                            i++;
+                        } else {
+                            fsb.append(numberFormatter.format(value), null); // TODO: Use proper Field
+                        }
+                    }
+                    break;
+                case '\'':
+                    // '' is escaped apostrophe
+                    if ((i + 1 < pattern.length()) && pattern.charAt(i + 1) == c) {
+                        fsb.append(c.toString(), null);
```
sffc: Please use `fsb.appendCodePoint(c, null)`

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
```diff
@@ -861,116 +859,77 @@ private static DateFormat loadNumericDurationFormat(ICUResourceBundle r, String
...
+                        i++;
+                    } else {
+                        protect = !protect;
+                    }
+                    break;
+                default:
+                    fsb.append(c.toString(), null);
```
sffc: Please use `fsb.appendCodePoint(c, null)`

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+                        numberFormatter->format(value, tmp, status);
+                    }
+                    // TODO: Use proper Field
+                    fsb.append(tmp, (FormattedStringBuilder::Field) NULL, status);
+                }
+                break;
+            case '\'':
+                // '' is escaped apostrophe
+                if ((i + 1 < patternLength) && pattern[i + 1] == c) {
+                    fsb.append(UnicodeString(c), (FormattedStringBuilder::Field) NULL, status);
```
sffc: ditto

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+                    i++;
+                } else {
+                    protect = !protect;
+                }
+                break;
+            default:
+                fsb.append(UnicodeString(c), (FormattedStringBuilder::Field) NULL, status);
```
sffc: ditto

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+        }
     }
+
+    appendTo.append(fsb.toUnicodeString());
```
sffc: You can do `toTempUnicodeString()` here.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
-            smallestAmount, smallestAmountFormatted, intFieldPosition, status);
-    if (
-            intFieldPosition.getBeginIndex() == 0 &&
-            intFieldPosition.getEndIndex() == 0) {
+
+    UnicodeString pattern;
+
+    double hours = hms[0].getDouble(status);
+    double minutes = hms[1].getDouble(status);
+    double seconds = hms[2].getDouble(status);
```
sffc: You should check for `U_FAILURE(status)` here.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
-                smallestAmountFormatted,
-                intFieldPosition.getEndIndex(),
-                appendTo);
-        appendRange(
-                draft,
-                smallestFieldPosition.getEndIndex(),
-                appendTo);
-    } else {
-        appendTo.append(draft);
+    const NumberFormat *numberFormatter = numberFormat->get();
```
sffc: Please look at `formatMeasure` for how to get the cached LocalizedNumberFormatter.  It is inefficient to use the public-facing DecimalFormat APIs in ICU implementation code.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+
+        // There is not enough info to add Field(s) for the unit because all we have are plain
+        // text patterns. For example in "21:51" there is no text for something like "hour",
+        // while in something like "21h51" there is ("h"). But we can't really tell...
+        switch (c) {
+            case 'H':
+            case 'm':
+            case 's':
+                if (protect) {
+                    fsb.append(UnicodeString(c), (FormattedStringBuilder::Field) NULL, status);
```
sffc: For undefined field, please use

`UNUM_FIELD_COUNT`

I know, yuck yuck.  Why does UNUM_FIELD_COUNT mean null field when this is not a number-specific format?  For historical reasons.  I filed a bug to clean this up:

https://unicode-org.atlassian.net/browse/ICU-20665

Also, please use `appendCodePoint` when you are appending a single character.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+    const NumberFormat *numberFormatter = numberFormat->get();
+    auto* numberFormatter2 = (NumberFormat*) numberFormatter->clone();
+    numberFormatter2->setMinimumIntegerDigits(2);
+
+    FormattedStringBuilder fsb;
+
+    UBool protect = FALSE;
+    const int32_t patternLength = pattern.length();
+    for (int32_t i = 0; i < patternLength; i++) {
+        char c = pattern[i];
```
sffc: `pattern` is a UnicodeString, and the `UnicodeString::operator[]` returns a `char16_t`, not a `char`.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+
+        // Also set the proper field in this switch
+        // We don't use DateFormat.Field because this is not a date / time, is a duration.
+        double value = 0;
+        switch (c) {
+            case 'H': value = hours; break;
```
sffc: `c` is a UTF-16 code unit, so you should use `u'H'`, not `'H'`, which is platform-dependent.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+            case 'm': value = minutes; break;
+            case 's': value = seconds; break;
+        }
+
+        // There is not enough info to add Field(s) for the unit because all we have are plain
+        // text patterns. For example in "21:51" there is no text for something like "hour",
+        // while in something like "21h51" there is ("h"). But we can't really tell...
+        switch (c) {
+            case 'H':
```
sffc: Same here; please use `u'H'` (also for the other switch conditions)

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
mihnita: This would damage strings with surrogates.
Java string is UTF-16, so a char (or Character) is not a code point, is a code unit.

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
mihnita: I used Character so that I can do `c.toString()`
But I can do `char` and `String.valueOf(c)`.
Not sure if the performance makes a difference, but well, does not hurt :-)

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
mihnita: Done.

icu4c/source/i18n/measfmt.cpp
mihnita: Can't use appendCodePoint because I iterate over code units (the `[]` operator returns `char16_t`)

Options:
```
int patternLen = pattern.countChar32();
for (int i = 0; i < patternLen; i++) {
    UChar32 c = pattern.char32At(i);
    ...
    fsb.appendCodePoint(c, (FormattedStringBuilder::Field) NULL, status);
}
```
or
```
int patternLen = pattern.length();
for (int i = 0; i < patternLen; i++) {
    char16_t c = pattern[i];
    fsb.append(UnicodeString(c), (FormattedStringBuilder::Field) NULL, status);
}
```

`UnicodeString(c)` might be expensive.
But `char32At` is also expensive, because it has to iterate the whole string, for each get.
Maybe convert the whole string to a `UChar32` array.

For the first option (`char32At`) the documentation says _"Return the code point that contains the code unit at offset `<tt>`offset`</tt>`."_
I find that a bit confusing. Is that the offset in code points, or code units?
Trying it seems to indicate that it is code point offset (as the name suggests). The doc makes things more confusing instead of more clear...

icu4c/source/i18n/measfmt.cpp
mihnita: ditto :-)

icu4c/source/i18n/measfmt.cpp
mihnita: Done

icu4c/source/i18n/measfmt.cpp
mihnita: Done

icu4c/source/i18n/measfmt.cpp
mihnita: I'm not sure I understand.

I use the existing formatter, I don't create a new one.

**Me**
```
const SharedNumberFormat *numberFormat;
...
UnicodeString &MeasureFormat::formatNumeric( ... ) {
    ...
    const NumberFormat *numberFormatter = numberFormat->get();
    ...
}
```

**formatMeasure does this:**
```
UnicodeString &MeasureFormat::formatMeasure( ... const NumberFormat &nf ... ) {
    auto* df = dynamic_cast<const DecimalFormat*>(&nf);
    // check df == nullptr) { ... }
    number::FormattedNumber result;
    if (auto* lnf = df->toNumberFormatter(status)) {
        result = lnf->unit(amtUnit)
            .unitWidth(getUnitWidth(fWidth))
            .formatDouble(amtNumber.getDouble(status), status);
    }
}
```

So it takes a `NumberFormat` as parameter, then casts it to `DecimalFormat`, then gets a `toNumberFormatter` from it.

How does it get invoked (so that we know where the parameter comes from)?

A. in **MeasureFormat::format** 
```    formatMeasure(*amount, **numberFormat, appendTo, pos, status);```
B. in **MeasureFormat::formatMeasures**
```     formatMeasure(measures[0], **numberFormat, appendTo, pos, status);```
C. in **MeasureFormat::formatMeasures**
```     formatMeasure(measures[i], *nf, results[i], pos, status);
D. in **MeasureFormat::formatMeasuresSlowTrack**
```     formatMeasure(measures[i], *nf, results[i], fpos, status);
     formatMeasure(measures[i], *nf, results[i], dontCare, status);```

In A and B the `**numberFormat` is the same one I am using. No caching.
In C there is the cached integer formatter, sometimes
```
        const NumberFormat *nf = cache->getIntegerFormat();
        if (i == measureCount - 1) {
            nf = numberFormat->get();
        }
```
In D is the cached integer formatter: `const NumberFormat *nf = cache->getIntegerFormat();`

So C / D get use some caching, but that is the `getIntegerFormat`, which is the wrong formatter.
I don't want an integer formatter, because I needs to do stuff like 12:34.5678 (with decimals)

File: icu4c/source/i18n/measfmt.cpp (hunk 14 above)
Comment (mihnita) [reply icu4c/source/i18n/measfmt.cpp]: Done UNUM_FIELD_COUNT.

Can't use appendCodePoint, since the character is char16_t, so it might be a utf-16 surrogate, not a code point.

File: icu4c/source/i18n/measfmt.cpp (hunk 15 above)
Comment (mihnita) [reply icu4c/source/i18n/measfmt.cpp]: Done.

File: icu4c/source/i18n/measfmt.cpp (hunk 17 above)
Comment (mihnita) [reply icu4c/source/i18n/measfmt.cpp]: Done

File: icu4c/source/i18n/measfmt.cpp (hunk 16 above)
Comment (mihnita) [reply icu4c/source/i18n/measfmt.cpp]: Done.

File: icu4c/source/i18n/measfmt.cpp (hunk 9 above)
Comment (sffc) [reply icu4c/source/i18n/measfmt.cpp]: > Is that the offset in code points, or code units?

The offset is always in code units.

File: icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java (hunk 5 above)
Comment (sffc) [reply icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java]: > This would damage strings with surrogates.

The implementation checks if it needs to allocate 1 or 2 code units.  In C++, it checks with U16_LENGTH, and in Java with Character.charCount.

C++ U16_LENGTH is defined as:

    #define U16_LENGTH(c) ((uint32_t)(c)<=0xffff ? 1 : 2)

Java Character.charCount is defined as:

> If the specified character is equal to or greater than 0x10000, then the method returns 2. Otherwise, the method returns 1.

Therefore I think the method works even if you are dealing with a surrogate code unit.

I filed a ticket to clarify: https://unicode-org.atlassian.net/browse/ICU-20666

FormattedStringBuilder is an internal API that isn't perfect but we continually improve on it.

> The Java StringBuilder has an append(char), but FormattedStringBuilder does not.

It was built with the idea that you operate in code points.

Comment (sffc) [reply icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java]: Another way to put it is, a code unit (including a surrogate code unit) can be treated as if it were a code point.

Comment (sffc) [reply icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java]: Java style is to loop over code units.  ICU style is to loop over code points (but to index into strings using code unit offsets).  That's why the ICU internal API has an appendCodePoint, but the Java API has an append(char).

If you want to re-write your loop ICU-style, you could do:

```java
for (int cp, i = 0; i < pattern.length(); i += Character.charCount(cp)) {
  cp = Character.codePointAt(pattern, i);
  // ...
}
```

The alternative is to use `appendCodePoint(c, null)` despite the misnomer.  Converting to a string is not efficient and isn't a good option.

icu4c/source/i18n/measfmt.cpp
```diff
@@ -770,71 +715,96 @@ static void appendRange(
...
+                } else {
+                    UnicodeString tmp;
+                    if ((i + 1 < patternLength) && pattern[i + 1] == c) { // doubled
+                        numberFormatter2->format(value, tmp, status);
+                        i++;
+                    } else {
+                        numberFormatter->format(value, tmp, status);
+                    }
+                    // TODO: Use proper Field
+                    fsb.append(tmp, (FormattedStringBuilder::Field) NULL, status);
```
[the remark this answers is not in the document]
mihnita: Done

icu4c/source/i18n/measfmt.cpp
mihnita: Done.

icu4c/source/i18n/measfmt.cpp
mihnita: Done

icu4c/source/i18n/measfmt.cpp
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/text/MeasureFormat.java
mihnita: Done

icu4c/source/i18n/measfmt.cpp
sffc: The field `numberFormat` is the right one to use.  I'm just saying that it is more efficient to call `.toNumberFormatter()` and then use `.setIntegerWidth()` than it is to call `.clone()` and `.setMinimumIntegerDigits()`.

sffc: Something like this:

```cpp
const auto* df = dynamic_cast<const DecimalFormat*>(numberFormat->get());
// handle nullptr
LocalizedNumberFormatter lnf2;
if (auto* lnf = df->toNumberFormatter(status)) {
    lnf2 = lnf->integerWidth(IntegerWidth::zeroFillTo(2));
} else {
    // return
}
```

It's more efficient because you don't generally have to allocate heap memory, and also note that `.setMinimumIntegerDigits()` under the hood does a bit of extra work that is avoided here.

sffc:
Please squash.

icu4c/source/i18n/measfmt.cpp
mihnita: Done. Thanks a lot.

PR #686, opened 2019-06-12
ICU-20664 Add CI builds that exercise the Data Filtering.

jefgen:
This adds two extra CI builds to exercise the Data Filtering on Linux with Clang and on Windows with MSVC.

Notes:
I'm only using the simple example file from the [ICU Data Build Tool](https://github.com/unicode-org/icu/blob/master/docs/userguide/icu_data/buildtool.md) document for now, and adding it to the `.ci-builds` sub-folder.
We could move it somewhere else if you wanted, or use more advanced data filtering -- this is mainly just to have CI build coverage for now.
Also, these builds don't run any tests, as the reduced data bundle would cause test failures.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20664
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -37,6 +37,19 @@ jobs:
         CC: clang
         CXX: clang++
 #-------------------------------------------------------------------------
+- job: ICU4C_Clang_Ubuntu_DataFilter_1604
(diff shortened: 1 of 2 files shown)


jefgen:
Ah, I realized that I didn't change the display name on the Clang build. 
It still said "`Build and Test`".
I force-pushed a small update to change it to "`Build with Data Filter`"

PR #687, opened 2019-06-13
ICU-20404 Simplify word break rules.

aheninger:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20404
- [ ] Issue accepted
- [ ] Tests included


aheninger:
This PR recasts the way that word break rules deal with format and extend characters, removing them from being explicitly included at the end of most rules.

This makes it easier to reason about how rules can chain into others - chaining can occur only at text positions where a rule has matched, and the change gets trailing extend character matches out of most rules.

The change also reduces the size of the word break binary data by about 10%. The data size for root word and word_POSIX together drops form 89,884 to 80,984 bytes.

PR #688, opened 2019-06-16
ICU-20669 French spellout rules: Add unit tests, with several failing

adrogon:
This PR only adds more tests to the French spellout rules, a few of them are failing on purpose, to show where the library is currently failing at returning them.

I probably cannot file a proper PR for the production code part, that would make the tests pass, hence the draft status of this one.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20669
- [ ] Issue accepted
- [x] Tests included


sffc:
Hi, thanks for the contribution.  Please use `logKnownIssue()` when you add a test that is intentionally failing.  Search the test suite for examples.

srl295:
the bug was moved to [CLDR-13116]

sffc:
I'm going to close this PR since this appears to be a CLDR bug, not an ICU bug.

PR #695, opened 2019-06-23
ICU-20048 remove uses of UAlignedMemory

aheninger:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20048
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/ucnv.cpp
+++ b/icu4c/source/common/ucnv.cpp
@@ -25,6 +25,8 @@
 
 #if !UCONFIG_NO_CONVERSION
 
+#include <memory>
+
 #include "unicode/ustring.h"
 #include "unicode/ucnv.h"
 #include "unicode/ucnv_err.h"
@@ -158,7 +160,6 @@ ucnv_safeClone(const UConverter* cnv, void *stackBuffer, int32_t *pBufferSize, U
     UConverter *localConverter, *allocatedConverter;
     int32_t stackBufferSize;
     int32_t bufferSizeNeeded;
-    char *stackBufferChars = (char *)stackBuffer;
     UErrorCode cbErr;
     UConverterToUnicodeArgs toUArgs = {
         sizeof(UConverterToUnicodeArgs),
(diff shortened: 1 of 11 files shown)


icu4c/source/common/ucnv.cpp
```diff
@@ -225,22 +226,18 @@ ucnv_safeClone(const UConverter* cnv, void *stackBuffer, int32_t *pBufferSize, U
...
+    /* Adjust (if necessary) the stackBuffer pointer to be aligned correctly for a UConverter.
      */
-    if (U_ALIGNMENT_OFFSET(stackBuffer) != 0) {
-        int32_t offsetUp = (int32_t)U_ALIGNMENT_OFFSET_UP(stackBufferChars);
-        if(stackBufferSize > offsetUp) {
-            stackBufferSize -= offsetUp;
-            stackBufferChars += offsetUp;
+    size_t altStackBufferSize = stackBufferSize;  // Incompatible types for passing by reference.
+    if (stackBuffer) {
+        if (std::align(alignof(UConverter), bufferSizeNeeded, stackBuffer, altStackBufferSize)) {
```
markusicu: You are only testing the return value for nullptr. If I read the docs right, you need to capture the returned pointer, and if it's not nullptr then assign it to stackBuffer, since it may have been adjusted up by a few bytes. Just like the old code adjusted up the stackBufferChars.
https://en.cppreference.com/w/cpp/memory/align

aheninger: I think it's OK. The pointer parameter to std::align() is passed by reference, and is updated in place, in addition to being returned.

cppreference.com is a little vague, it doesn't say up front that the input pointer is modified, but does say further down 
> The function modifies the pointer only if it would be possible to fit ...

 http://www.cplusplus.com/reference/memory/align/ is more explicit, saying

> The function updates ptr to point to such [the aligned] address, and decreases space by the number of bytes used for alignment, so that these arguments can be used in repeated calls to this function.

markusicu: Thanks. cppreference.com is really misleading on this one.

srl295:
added a comment.. breaks on gcc 4.8/4.9

PR #697, opened 2019-06-30
ICU-20650 Memory Alignment in quickSort() function.

aheninger:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20650
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/uarrsort.cpp
+++ b/icu4c/source/common/uarrsort.cpp
@@ -34,6 +34,10 @@ enum {
     STACK_ITEM_SIZE=200
 };
 
+static constexpr int32_t sizeInMaxAlignTs(int32_t sizeInBytes) {
+    return (sizeInBytes + sizeof(max_align_t) - 1) / sizeof(max_align_t);
+}
+
 /* UComparator convenience implementations ---------------------------------- */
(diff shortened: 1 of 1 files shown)


icu4c/source/common/uarrsort.cpp
```diff
@@ -238,27 +232,19 @@ subQuickSort(char *array, int32_t start, int32_t limit, int32_t itemSize,
...
     /* allocate two intermediate item variables (x and w) */
-    if(itemSize<=STACK_ITEM_SIZE) {
-        p=xw;
-    } else {
-        p=uprv_malloc(2*itemSize);
-        if(p==NULL) {
-            *pErrorCode=U_MEMORY_ALLOCATION_ERROR;
-            return;
-        }
+    icu::MaybeStackArray<max_align_t, sizeInMaxAlignTs(STACK_ITEM_SIZE)> x;
```
markusicu: I would prefer still making only one xw of twice an item size so that we heap-allocate at most once, and for w use p + sizeInMaxAlignTs(itemSize) where p = xw.getAlias().

PR #699, opened 2019-07-05
ICU-20693 Draft pull request for early review of some of the new mapp…

hagbard:
…er class replacements

##### Checklist

- [ ] Issue accepted
- [ ] Tests included


hagbard:
Code is still very much work in progress, and comments could do with a lot of improvement.
Tests don't even exist yet.

There's one very much unfinished visitor class in here, but apart from that it's all working nicely (produces almost identical output, with only known differences for the completed mapper classes).

Known differences:
* New code will not produce empty collation data file (this might be required)
* New code fixes a bug which was suppressing the "tender='false'" attribute being emitted for a lot of currencies.

srl295:
what's the relation between this and #721 ?

hagbard:
This is obsolete now - thanks for reminding me it was still around.

PR #701, opened 2019-07-09
ICU-20593 Data slicing: allow more characters in res filter rules.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20593
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/python/icutools/databuilder/filtration_schema.json
+++ b/icu4c/source/python/icutools/databuilder/filtration_schema.json
@@ -38,7 +38,7 @@
                         "type": "array",
                         "items": {
                             "type": "string",
-                            "pattern": "^[+-]/((\\w+|\\*)(/\\w+|/\\*)*)?$"
+                            "pattern": "^[+-]/[\\S]+$"
                         }
                     }
                 },


sffc:
The problem is that keys such as `%%Parent` need to be whitelisted when additive mode is used.  The alternative here would be to always unconditionally include keys starting with `%%`, or at least `%%Parent`.

icu4c/source/python/icutools/databuilder/filtration_schema.json
```diff
@@ -38,7 +38,7 @@
                         "type": "array",
                         "items": {
                             "type": "string",
-                            "pattern": "^[+-]/((\\w+|\\*)(/\\w+|/\\*)*)?$"
+                            "pattern": "^[+-]/(([\\w%]+|\\*)(/[\\w%]+|/\\*)*)?$"
```
markusicu: This is fine for % but I think genrb allows anything that is an "invariant character", possibly minus spaces and quotation marks (I would have to look at the parser code).

How about just `"^[+-]/\\s$"` and let the genrb parser deal with key validity?

sffc:
The commit message is already up to date.  I fixed the PR title to match the commit message.

PR #705, opened 2019-07-12
ICU-20688 SHIFT-JIS Conversion Issues - Patch

DGN001:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20688
- [x] Issue accepted
- [ ] Tests included


DGN001:
I was able to reproduce it on the source code which I download from below link :-
https://github.com/unicode-org/icu/releases/tag/release-64-2

sffc:
What's the status of this PR, awaiting author or reviewer?

srl295:
@sffc neither. it's for a bug analysis. I'm back, and will continue this. please leave it open.

grhoten:
I left a more detailed reply in the associated Jira issue. These authoritative IBM tables should not be modified. IBM did this on purpose. One of the other flavors of Shift-JIS can be used instead, like https://github.com/unicode-org/icu-data/blob/main/charset/data/ucm/windows-932-2000.ucm.

markusicu:
Right. The ibm-*.ucm files were generated from IBM conversion tables and reflect actual IBM conversion behavior. We provide other mapping tables that can be used. We might consider moving to non-IBM conversion tables in our default data, but not as a patch of the IBM data files.

PR #706, opened 2019-07-17
ICU-20712 Add ARM64 support for ICU4C UWP, add to CI builds.

jefgen:
This change adds support for building the ICU4C UWP projects for ARM64.
(Note: The "UWP" projects for ICU4C only build the "common" and "i18n" projects.)

Additionally, this change also adds a CI build for ARM32 and ARM64, so that we can have build coverage for these as well.

Co-Authored-By: Daniel Ju <daju@microsoft.com>

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20712
- [x] Issue accepted
- [ ] Tests included


jefgen:
The CI build failure on MSYS2 GCC x86_64 is unrelated to this change.
In fact it looks like the build is failing on the master branch, has been failing since last week.

I've filed a separate ticket [ICU-20718] for investigating what/who might have caused the failure.

jefgen:
Note: I force pushed to pick up the MSYS build fixes, but doing a force push dismisses any reviews/approvals. I'll need somebody to sign-off again in order to complete the PR.
Thanks!

PR #708, opened 2019-07-19
ICU-20493 Warn on global/static c’tors

srl295:
[ICU-20493]

- add -Wglobal-constructors to clang (only for libs)
- Note: ignore 'runstatedir' flap in configure

[ICU-20493]: https://unicode-org.atlassian.net/browse/ICU-20493

### DIFF
--- a/icu4c/source/configure
+++ b/icu4c/source/configure
@@ -755,7 +755,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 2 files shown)


aheninger:
Waiting on some resolution for the failing Azure CI builds...

jefgen:
The failing MSYS2 builds are due to an update to MSYS that was released, which causes the `shell cmd` command to no longer work.

I have a fix in PR #711 which resolves the issue. This was recently merged to master and the CI builds are now back to a working state.

If this PR could be rebased on the latest changes from master, then I suspect that the failing CI builds should hopefully be fixed. :)

srl295:
Ok. Will do

markusicu:
In the PR text, "Note: ignore 'runstatedir' flap in configure" did you mean s/flap/flag/ ?

srl295:
Flap. An unreleased autoconf version keeps adding stuff, when I use the
latest release it removes it. So it flaps back and forth.

markusicu:
I know "flip-flop" but "flap" is new to me :-)

PR #709, opened 2019-07-19
ICU-20718 Disable the MSYS2 CI builds for now.

jefgen:
It seems that an update to the Azure DevOps images that run the CI builds has caused the MSYS2 build to start failing on the master and maint-64 branches.
I'm not sure what the underlying root cause it, but I'm temporarily disabling this flavor of the CI build to unblock other PRs.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20718
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #711, opened 2019-07-22
ICU-20718 Fix failing MSYS2 build due to current path confusion.

jefgen:
The Azure DevOps VM images that are used in the CI builds recently updated the version of MSYS that is installed, which is causing the MSYS builds to fail. It seems that this new version has issues with using the `shell cmd` method in the makefiles to obtain the current path.
However, we can avoid this entirely by using the `cygpath` utility, which also avoids the overhead of launching a new `cmd` instance just to get the path.
Unfortunately, it seems that older versions of MSYS don't have this utility program included by default, though versions of MSYS 2 and up do have it included by default.
We can continue using the old approach for the older versions and use the `cygpath` approach for new versions.

This PR supplants/replaces PR #709.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20718
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/config/mh-mingw
+++ b/icu4c/source/config/mh-mingw
@@ -103,13 +103,29 @@ LIBICU = $(STATIC_PREFIX_WHEN_USED)$(ICUPREFIX)
 #SH## with usually. Many applications will want to add ${ICULIBS_I18N} as well. 
 #SH#ICULIBS="${ICULIBS_BASE} ${ICULIBS_I18N} ${ICULIBS_COMMON} ${ICULIBS_DATA} "
 
-# The #M# is used to delete lines for icu-config
-# Current full path directory.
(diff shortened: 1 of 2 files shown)


jefgen:
Note: I force-pushed to update the commit message, but the contents are the same.

sffc:
Thanks for keeping support for both old and new versions of MSYS!  I assume you tested it on both?

jefgen:
Thanks. Yes, I did try building on MSYS1, MSYS2 and MSYS3.

Though the version of MSYS1 that I was able to get working doesn't actually "build" as the version of the compiler is so old that it doesn't have support for C++11's `std::mutex` and `std::atomic`.
However, I did confirm that the paths that the makefile computes are the same -- so if somebody does have a working version of MSYS1, then it *should* build.

I was able to build and test on MSYS2 and MSYS3 without any issues though.

PR #716, opened 2019-07-26
ICU-13764 Add separate CI build with warnings as errors (clang)

jefgen:
This adds a separate CI build that enables `-Werror` (treat warnings as errors) for the Clang compliler.
This change also fixes a number of warnings so that the ICU4C build is clean with both `Wall`, `Wextra`, and `Wextra-semi` now.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-13764
- [x] Issue accepted
- [ ] Tests included


.ci-builds/.azure-pipelines.yml
```diff
@@ -37,6 +37,19 @@ jobs:
...
         CXX: clang++
 #-------------------------------------------------------------------------
+- job: ICU4C_Clang_Ubuntu_1604_WarningsAsErrors
+  displayName: 'C: Linux Clang WarningsAsErrors (Ubuntu 16.04)'
+  timeoutInMinutes: 30
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+    - script: |
+        export CPPFLAGS="-Wall -Werror" && cd icu4c/source && ./runConfigureICU Linux && make -j2 check
```
sffc: Probably you only need to build, not run the tests

jefgen: Good point! I'll update to remove the check.

jefgen:
> Any reason you made this a new build, instead of augmenting the existing builds?

Good question. Personally, I would probably prefer to just have this on the "regular" build, but I think the feedback when the ticket was discussed was that having a separate build would allow for ignoring noisy warnings if you are only interested in seeing if the tests pass with your change.

Since the clang builds are really rather fast (compared to MSVC), I think adding a separate build doesn't add to much overhead.
Though we could just enable this on the main build later if that was okay with the TC.

.ci-builds/.azure-pipelines.yml
```diff
@@ -44,8 +44,8 @@ jobs:
     vmImage: 'Ubuntu 16.04'
   steps:
     - script: |
-        export CPPFLAGS="-Wall -Werror" && cd icu4c/source && ./runConfigureICU Linux && make -j2 check
-      displayName: 'Build and Test (WarningsAsErrors)'
+        export CPPFLAGS="-Wall -Werror -Wextra" && cd icu4c/source && ./runConfigureICU Linux && make -j2
```
sffc: Please use `make -j2 tests` so that the tests are also built.

sffc: Alternative: if the tests have too many warnings and you don't want to clean them up, add a comment saying something to that extent and keep it at just building the library code.

sffc:
Does `-Wall` not imply `-Wextra-semi` ?  I guess not?

.ci-builds/.azure-pipelines.yml
jefgen: Thanks Shane, that's a good point. I'll see about fix the warnings/errors in the tests.

jefgen:
> Does `-Wall` not imply `-Wextra-semi` ? I guess not?

Apparently not, at least with Clang version 6.0 which is what I have locally and what is on the Ubuntu VMs that the builds use.
(I went ahead and fixed all the warnings from Wextra-semi in this PR anyways though.)

sffc:
Apparently there's also `-Weverything`:

https://embeddedartistry.com/blog/2017/3/7/clang-weverything

jefgen:
Interesting... thanks Shane!

I think cleaning up *all* the warnings might take some time though...
```
$ grep ': warning: ' with-Weverything-enabled.log | wc -l
351472
```
(Though, to be fair there are probably a large number of duplicate messages output for one issue in a header file, etc.)

jefgen:
Thanks for the review!
Since this has 4 commits, I'll squash and then force-push -- which will require one more sign off.

PR #717, opened 2019-07-26
ICU-20675 ignore .vscode configuration files.

younies:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20675
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ __pycache__/
 
 # Specific file name ignores
 /.git
+.vscode
 icu4c/bin
 icu4c/bin32uwp
 icu4c/bin64


sffc:
Please fill in the checklist in the OP, including the blank where you put the issue number.

.gitignore
```diff
@@ -16,6 +16,7 @@ icu4c/lib32uwp
 icu4c/lib64
 icu4c/lib64uwp
 icu4c/libARMuwp
+icu4c/source/.vscode/*
```
sffc: You don't need the `/*` in .gitignore syntax.  If you give a directory path, the whole directory is ignored.

sffc: Actually, a better solution is to just put `.vscode/` at the top level.  That ignores all `.vscode` directories anywhere in the directory tree.

younies: Done :)

Thanks for the note.

younies:
> Thanks, though I think this should likely be for the top level folder path and not inside the ICU4c subfolder.

Shane suggested this too, and I make it in the top.

sffc:
Alternatively, my setup actually is that I have `.vscode/` as a global ignore in my personal git config settings.  So it's not strictly necessary to add it here.  But it does still reduce a step if you don't have your own personal gitignore file, so no harm done.

PR #718, opened 2019-07-26
ICU-20048 Remove use of std::align() for compat with g++ 4.9.

aheninger:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20048
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/ucnv.cpp
+++ b/icu4c/source/common/ucnv.cpp
@@ -225,13 +225,16 @@ ucnv_safeClone(const UConverter* cnv, void *stackBuffer, int32_t *pBufferSize, U
         }
     }
 
-
     /* Adjust (if necessary) the stackBuffer pointer to be aligned correctly for a UConverter.
+     * TODO(Jira ICU-20736) Redo this using std::align() once g++4.9 compatibility is no longer needed.
      */
-    size_t altStackBufferSize = stackBufferSize;  // Incompatible types for passing by reference.
     if (stackBuffer) {
-        if (std::align(alignof(UConverter), bufferSizeNeeded, stackBuffer, altStackBufferSize)) {
-            stackBufferSize = static_cast<int32_t>(altStackBufferSize);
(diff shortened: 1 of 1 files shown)


icu4c/source/common/ucnv.cpp
```diff
@@ -225,13 +225,16 @@ ucnv_safeClone(const UConverter* cnv, void *stackBuffer, int32_t *pBufferSize, U
...
     /* Adjust (if necessary) the stackBuffer pointer to be aligned correctly for a UConverter.
+     * TODO(Jira ICU-20736) Redo this using std::align() once g++4.9 compatibility is no longer needed.
      */
-    size_t altStackBufferSize = stackBufferSize;  // Incompatible types for passing by reference.
     if (stackBuffer) {
-        if (std::align(alignof(UConverter), bufferSizeNeeded, stackBuffer, altStackBufferSize)) {
-            stackBufferSize = static_cast<int32_t>(altStackBufferSize);
+        uintptr_t p = reinterpret_cast<uintptr_t>(stackBuffer);
+        uintptr_t aligned_p = (p + alignof(UConverter) - 1) & ~(alignof(UConverter) - 1);
+        ptrdiff_t sizeAdjustment = aligned_p - p;
```
markusicu: It's not a sizeAdjustment, it's a pointerAdjustment or offset or alignIncrement... right?

aheninger: Well, it's both. It is the amount by which the stack buffer size is about to be adjusted, and the amount by which the pointer was adjusted. I was thinking the former when I named it.

I'm perfectly happy to change the name if you think it's clearer.

markusicu: Hm, you are right, but it's confusing looking at the line where it's computed. "offset" would be clearer.

aheninger: Done

PR #719, opened 2019-07-31
ICU-20747 Change dependency for common_uwp from onecoreuap.lib to onecore.lib

rohanp-msft:
onecore.lib is on all devices, while onecoreuap.lib is only on devices that support UWP. Moving down to OneCore.lib means being able to target more devices.
See https://docs.microsoft.com/en-us/windows/win32/apiindex/windows-umbrella-libraries for additional info.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20747
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/common_uwp.vcxproj
+++ b/icu4c/source/common/common_uwp.vcxproj
@@ -114,7 +114,7 @@
       </DataExecutionPrevention>
       <TurnOffAssemblyGeneration>true</TurnOffAssemblyGeneration>
       <IgnoreSpecificDefaultLibraries>vccorlib.lib;msvcrt.lib</IgnoreSpecificDefaultLibraries>
-      <AdditionalDependencies>onecoreuap.lib</AdditionalDependencies>
+      <AdditionalDependencies>onecore.lib</AdditionalDependencies>
     </Link>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)'=='Release'">


jefgen:
Thanks @rohanp-msft for the PR! :)

Would you mind changing the title of your commit message to include the JIRA ticket number?

It should look something like this:
"ICU-20747 Change dependency for common_uwp from onecoreuap.lib to onecore.lib"

(Or perhaps the slightly shorter: "ICU-20747 Modify common_uwp to use onecore.lib instead of onecoreuap.lib". But I'll leave the actual message up to you.)

You can use the 'git amend' feature to change the message on the commit, and then force-push the changes to your branch.
For example: 
`git commit --amend -m "ICU-20747 New message title"`

Thanks!

rohanp-msft:
> Thanks @rohanp-msft for the PR! :)
> Would you mind changing the title of your commit message to include the JIRA ticket number?
> It should look something like this:
> "ICU-20747 Change dependency for common_uwp from onecoreuap.lib to onecore.lib"
> (Or perhaps the slightly shorter: "ICU-20747 Modify common_uwp to use onecore.lib instead of onecoreuap.lib". But I'll leave the actual message up to you.)
> You can use the 'git amend' feature to change the message on the commit, and then force-push the changes to your branch.
> For example:
> git commit --amend -m "ICU-20747 New message title"
> Thanks!

Done!

jefgen:
Note for posterity:

The TravisCI build shows as failing. However, looking at the builds it seems that there is some sort of infra issue with Travis as the failure is only happening on the Java build and the error message is that ant isn't installed.

PR #721, opened 2019-08-01
ICU-20693 New LDML to ICU tooling.

hagbard:
This is the first ready to review version of the new ICU LDML conversion tools.
Tests are coming, but since I'm using JUnit and Truth locally we still need a discussion about where that should go.

Config is very nonexistent yet, but to run:
* set the following environment variables
-DCLDR_DIR=<locale cldr release root dir>
-DICU_DIR=<locale icu release root dir>
-DICU_SRC=<path to the root of the source for the tools (ending with tools/cldr/src)>
-DCLDR_DTD_CACHE=<temporary directory for DTD cache>

* Run the LdmlConverter class.

After a short pause, you should see each locale ID (with time info) printed out.
If you don't change the IcuConverterConfig class, this will take about 10 minutes or so.

##### Checklist

- [x] Issue accepted
- [ ] Tests included


hagbard:
I realize now I didn't exclude one of the tests, so this probably won't build right now (lacking dependencies).

tools/cldr/src/com/ibm/icu/dev/tool/cldr/IcuConverterConfig.java
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * The converter config intended to generate the standard ICU data files.
+ */
+// TODO: Bind this to Ant or some other more user friendly config file format.
+// TODO: Uncomment "ars" once that's in CLDR.
+final class IcuConverterConfig implements LdmlConverter.ConverterConfig {
+    // The set of locale IDs used by ICU. Some of these are aliases however, so don't expect
+    // to find XML files for all of them.
+    private static final ImmutableSet<String> ICU_LOCALE_IDS = ImmutableSet.of(
```
hagbard: Copied from the ICU Ant config file as of release 35.1 (same for everything else here).

hagbard:
Right, I updated this to the latest version, which now produces 99.99% identical output to the current tools (with all remaining diffs being understood). I still need to wire it all up to something like Ant and provide a "patching" mechanism for manually edited files in the ICU directories affected by the tool, but at this point the code is ready for review.

I'll start to work exclusively on unit testing and Ant integration now, and consider this code complete and ready for submission.

tools/cldr/src/com/ibm/icu/dev/tool/cldr/mapper/TransformsMapper.java
```diff
@@ -0,0 +1,174 @@
...
+     * @param ruleFileOutputDir the directory into which transliteration rule files will be written.
+     * @return the IcuData instance to be written to a file.
+     */
+    public static IcuData process(CldrDataSupplier src, Path ruleFileOutputDir) {
+        RuleVisitor visitor = new RuleVisitor(p -> {
+            Path file = ruleFileOutputDir.resolve(p);
+            try {
+                return new PrintWriter(Files.newBufferedWriter(file, CREATE, TRUNCATE_EXISTING));
+            } catch (IOException e) {
+                throw new RuntimeException("error opening file: " + file, e);
```
hagbard: I'll change to throw UncheckedIOException

tools/cldr/src/com/ibm/icu/dev/tool/cldr/IcuTextWriter.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /** Write a file in ICU data format with the specified header. */
+    static void writeToFile(IcuData icuData, Path outDir, List<String> header) {
+        try {
+            Files.createDirectories(outDir);
+            try (Writer w = Files.newBufferedWriter(outDir.resolve(icuData.getName() + ".txt"));
+                PrintWriter out = new PrintWriter(w)) {
+                new IcuTextWriter(icuData).writeTo(out, header);
+            }
+        } catch (IOException e) {
```
hagbard: I'll change to throw UncheckedIOException

hagbard:
> In today’s ICU meeting we had a long discussion of the code location. Please use `/tools/cldr/cldr-to-icu/src/org/unicode/icu/tool/cldrtoicu/...`

Okay, I'll do that tomorrow. Might be worth making a new PR once that's done since it'll get really messy on this one I think.

Where should tests go?  I'd assume /tools/cldr/cldr-to-icu/**test**/org/unicode/icu/tool/cldrtoicu/...

Can I also assume that this will be a new project with the ability to adopt new dependencies on Java libraries (e.g. especially JUnit and Truth for testing) ?

hagbard:
As discussed by email, it'll be:
/tools/cldr/cldr-to-icu/src/{main|test}/org/unicode/icu/tool/cldrtoicu/...
Since that's better for splitting test vs non-test code and matches a more standard Maven open-source pattern of directories.

hagbard:
Okay, I've moved all the packages, as requested by Markus, and I've now got everything working with a standard "pom.xml" file in Maven. This is pretty much ready for review.

Note the now (to make it possible to run code without lots of arcane setup) the CLDR API java file is included in this pull request as part of the local maven repo. It's not ideal, but it means you can develop and work with the API without needing to directly depend on the code in the CLDR project as well (which is really difficult in Maven). It means that the ICU project code is now standalone from the CLDR project and doesn't need to rely on Ant, even indirectly (though obviously the CLDR data is still needed).

.gitignore
```diff
@@ -1009,3 +1009,6 @@ tools/unicode/c/genuca/x64
 tools/unicode/c/genuca/x86
 tools/unicode/c/icudefs.txt
 tools/unicodetools/com/ibm/rbm/lib
+
+# icu conversion tools (cldr-to-icu)
+tools/cldr/cldr-to-icu/target
```
srl295: Please move this into `tools/cldr/cldr-to-icu/.gitignore` where it needs to be only `/target`

tools/cldr/cldr-to-icu/README.txt
```diff
@@ -0,0 +1,71 @@
...
+Note that these instructions do not currently support configuration of the converter for things
+such as limiting the set of files produced. That is supported in code and could be easily added
+to the binary, or encapsulated via an Ant task, but currently it is not directly supported.
+See the IcuConverterConfig class for the API by which this can be supported.
+
+
+Notes on JDK versions:
+----------------------
+
+It is possible that some JDK versions will not, by default, support the set of features necessary
```
srl295: From maven we should be able to specify the exact JDK version used, so is this statement still needed?

tools/cldr/cldr-to-icu/README.txt
```diff
@@ -0,0 +1,71 @@
...
+
+To run the converter tool, you need to know the locations of several key directories before
+hand. These are:
+
+
+Important directories:
+----------------------
+
+<CLDR_DIR>   = The root directory of the CLDR release.
+<ICU_DIR>    = The root directory of the ICU release (probably a parent directory of where
```
srl295: ICU_DIR shouldn't need to be specified, or should at least default to the project-relative directory

tools/cldr/cldr-to-icu/README.txt
```diff
@@ -0,0 +1,71 @@
...
+               this README is located).
+<DTD_CACHE>  = The temporary cache directory in which DTD files are downloaded (this is the
```
srl295: Caching should not be needed at all, see [CLDR-13165] (I am deleting the caching entity resolver from CLDR.)

tools/cldr/cldr-to-icu/README.txt
```diff
@@ -0,0 +1,71 @@
...
+               same directory as would be used when running tools from the CLDR project).
+<OUT_DIR>    = The output directory into which ICU data files should be written.
```
srl295: OUT_DIR also should be defaulted to the appropriate location relative to ICU_DIR

tools/cldr/cldr-to-icu/lib/com/ibm/icu/icu-utilities/0.1-SNAPSHOT/icu-utilities-0.1-SNAPSHOT.pom
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.ibm.icu</groupId>
+  <artifactId>icu-utilities</artifactId>
+  <version>0.1-SNAPSHOT</version>
+  <description>POM was created from install:install-file</description>
```
srl295: neither this (nor the .jar ) should be here in git. As a stop-gap it should be generated and published as part of icu4j's release.   @yumaoka  ?

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
```
srl295: no, let's set up a real maven file on the CLDR side to publish cldr.jar

tools/cldr/cldr-to-icu/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+                    <target>8</target>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+
+    <repositories>
+        <repository>
+            <id>local-maven-repo</id>
```
srl295: like to avoid this, see other comments.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
```diff
@@ -0,0 +1,379 @@
...
+
+    // Currently hard-coded "hacks" which could be encoded via the builder if wanted.
+
+    @Override public Map<String, String> getForcedAliases(IcuLocaleDir dir) {
+        switch (dir) {
+        case COLL:
+            return ImmutableMap.<String, String>builder()
+                // It is not at all clear why this is being done (we expect "sr_Latn_ME" normally).
+                // TODO: Find out and document this properly.
+                .put("sr_ME", "sr_Cyrl_ME")
```
srl295: is there a bug for this?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    // The primary set of locale IDs to be generated. Other, directory specific, sets should be
+    // subsets of this. Some of these ID are aliases, so XML files may not exist for all of them.
+    //
+    // This was further modified (in order to better match the set of generated ICU files) by:
+    // * Removing "es_003" (which just seems to be ignored in current code)
+    // * Adding:  "en_NH", "sr_XK", "yue_CN", "yue_HK" (deprecated locale IDs in the manual config)
+    // * Adding: "no_NO_NY" (a not even structurally valid ID that exists for very legacy reasons)
+    private static final ImmutableSet<String> ICU_LOCALE_IDS = ImmutableSet.of(
```
srl295: no... please remove this list.  This should be generated from CLDR data + configuration. I strongly object to a static list in this file!

hagbard: Oh I'm very against this, but this list is not trivially computable from the CLDR data (or not by any way I know). When I had this just make all locales, you get about 20 or 30 more locales.

So, if you want this to go away, tell me how to get exactly this list from CLDR data (since, at least for migration of tools) there's a strong requirement to produce identical results.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
hagbard: I'm going to go through all by TODOs and generate about 20 tickets when this is all done (since during review some TODOs will be resolved, and I don't want to span a lot of tickets until after review).

tools/cldr/cldr-to-icu/pom.xml
hagbard: So would I, but right now it's the only way to make running this vaguely hermetic. Yes, we should redo the CLDR build stuff, but that's several weeks of work.

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
hagbard: Well sure, that's ideal, but should it hold up submitting this code?

tools/cldr/cldr-to-icu/lib/com/ibm/icu/icu-utilities/0.1-SNAPSHOT/icu-utilities-0.1-SNAPSHOT.pom
hagbard: FYI there's now an "api-jar" target in the Ant file which produces "cldr-api.jar" which is a smaller JAR than using "cldr.jar".

tools/cldr/cldr-to-icu/README.txt
hagbard: This one I disagree on. The default behaviour of overwriting some, but maybe not all, existing files in the release directory seems very dangerous.

tools/cldr/cldr-to-icu/README.txt
hagbard: "Should not" eventually ... or "should not" now ?

tools/cldr/cldr-to-icu/README.txt
hagbard: I think I can make Maven use the project directory for this, but I'm a Maven noob, so any advice welcome.

tools/cldr/cldr-to-icu/README.txt
hagbard: This is a "coded" bit of advice that's really for Googlers. Any Google setup uses an internal JDK by default which has had the SAX parser's behaviour limited for security reasons. The errors are weird and non-obvious and have taken me, and others, days to track down. The fix is just to use the non-production JDK that's also available (but not the default).

I guess the error message is pretty reasonable now on its own, so maybe this isn't required in the actual readme any more...

.gitignore
hagbard: Ahh thanks! Will do.

tools/cldr/cldr-to-icu/README.txt
srl295: - Already has no effect as of current CLDR / ICU master, as no DTDs are loaded via http/s
> `find ~/src/cldr ~/src/icu -name '*.xml' | fgrep .dtd | fgrep http   # no results`
- CachingEntityResolver is being removed in https://github.com/unicode-org/cldr/pull/108 (including from the API)

tools/cldr/cldr-to-icu/README.txt
srl295: I'll have to see what the overall workflow looks like then.

tools/cldr/cldr-to-icu/lib/com/ibm/icu/icu-utilities/0.1-SNAPSHOT/icu-utilities-0.1-SNAPSHOT.pom
srl295: If you want to go that route (getting things in and iterate), then I would prefer a set of instructions which includes running the mvn install yourself - but still, not including the jars or the mvn repo in the icu repository. As I understand it, for this route the 'cldr-to-icu' tool is not going to be the primary generator for ICU 65's data.

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
srl295: let's submit the code with tools/cldr-to-icu/lib being `.gitignore`'d then

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
srl295: Do you mean `icu-config.xml`?  If that's what this is replacing, please put this in a properties file (or equivalent), and make it a command line parameter, not baked into the source.

hagbard: Yes, that's what's being replaced. I can make a config/properties file, but so far nobody has suggested it's important for it to be easily configurable. Is there a use case other than "generate all the ICU data files" that would need a different list?

Review the rest (or as much as you want) and I'll come back to this tomorrow (I have to be in the hospital for at least some of tomorrow, so I might not get to it until later).

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
hagbard: That's an interesting idea. I'll have to fix the readme to explain why nothing works "out the box", but it is easy enough to install the 2 necessary JAR files. If it only needed to support Unix, I'd just write a bash script to set it all up.

tools/cldr/cldr-to-icu/lib/com/ibm/icu/icu-utilities/0.1-SNAPSHOT/icu-utilities-0.1-SNAPSHOT.pom
hagbard: Yeah, I was working on the assumption it'd be used until 2 days ago :)

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
srl295: it should just be mvn command lines, I'd think. so even if you weren't on unix you could just run the command lines. thanks, appreciate it..

tools/cldr/cldr-to-icu/README.txt
hagbard: Well, it's not my call, since I'm not doing the releases but a default behaviour of "overwrite a non-specific set of files in an existing directory" seems weird to me.

tools/cldr/cldr-to-icu/README.txt
hagbard: Yay, I'll see about getting all the code and docs related to this stripped out of the API code before I push another version of this PR. Thanks!

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
srl295: ok, take care. i will review the rest.

> Is there a use case other than "generate all the ICU data files" that would need a different list?

yes. "I only want these files" or "I want to include these files from seed".

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
hagbard: Well, you (may) need to build the cldr-api.jar, so that needs documenting too.

tools/cldr/cldr-to-icu/README.txt
hagbard: I may have misunderstood you. I thought you were saying that DTD_CACHE is no longer needed. It doesn't work for me when I omit it.

I suspect it is still needed when dealing with older releases at the moment (and since I am heavily testing on CLDR-35.1 / ICU-64.2 since that's the only release data I can test on) I need to leave it there for now. We can remove it again once the dust has settled and people are happy that we will never need to run the new tools on an old release of the data.

srl295: fair enough, but I expect to merge the PR into CLDR 36.

tools/cldr/cldr-to-icu/README.txt
srl295: that's actually been the usual workflow: `cd ~/src/icu/icu4c/source/data && ant clean && ant all`. Delete all generated data and rebuild. Verify, and check in.

tools/cldr/cldr-to-icu/lib/org/unicode/cldr/cldr-api/0.1-SNAPSHOT/cldr-api-0.1-SNAPSHOT.pom
srl295: as of https://github.com/unicode-org/cldr/commit/7a782455a103c45a5ae5b7656b14691fbd9322e1 the API is included in `cldr.jar`

srl295: > there's now an "api-jar" target in the Ant file

OK, i see you are referring to https://github.com/unicode-org/cldr/pull/107 — I'm not sure this is really needed right now, and as I predicted it is much more complex with an ant-based built. Just use cldr.jar?

hagbard: As mentioned there, the CLDR is a huge dependency mess and probably includes duplicates of classes I'm using via the ICU library dependencies in Maven. As it stands, the Ant file is broken and needs some kind of a fix though.

tools/cldr/cldr-to-icu/pom.xml
srl295: https://github.com/unicode-org/icu/pull/765 is a draft towards some of that several weeks.

hagbard: So that seems to only be redoing things to push the JAR to Maven after building with Ant.
For making it build with Maven (which is what you'd want to get rid of some of the really messy dependency issues) you need a fairly non-trivial refactoring of lots of code in and around the "util" package. Right now "util" has circular deps with lots of other classes in other packages, which need to be broken if we don't want a monolithic JAR for the API that contains 10M of unused code.

srl295: > So that seems to only be redoing things to push the JAR to Maven after building with Ant.

For the ICU `utilities.jar` side — yes.

srl295: > push the JAR to Maven after building with Ant

… which is how ICU's maven deployments work

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
hagbard: I've updated the build stuff and removed the API jar, but I haven't done the config stuff yet (and I'd rather do that once everything else is in since I'd like a longer discussion about use cases and config). For example if I could determine the list via examining the CLDR data, all the better.

tools/cldr/cldr-to-icu/pom.xml
hagbard: Right, but that won't fix the cyclic deps and non-hermetic/non-incremental building that Ant has (Ant is only reliable if you previously did a "clean" before a build since it relies on time stamps rather than content hashing to determine "freshness"). I'd be interested in being given the chance to move CLDR to build with Maven.

srl295: > Right, but that won't fix the cyclic deps and non-hermetic/non-incremental building that Ant has

it's meant as an incremental step

>  I'd be interested in being given the chance to move CLDR to build with Maven.

Right. Perhaps we can collaborate.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuTextWriter.java
```diff
@@ -0,0 +1,311 @@
...
+    private final IcuData icuData;
+    private int depth = 0;
+    private boolean valueWasInline = false;
+
+    IcuTextWriter(IcuData icuData) {
+        this.icuData = checkNotNull(icuData);
+    }
+
+    private void writeTo(PrintWriter out, List<String> header) throws IOException {
+        out.write('\uFEFF');
```
srl295: related to https://unicode-org.atlassian.net/browse/ICU-10197 - may want to add the utf-8 coding to the header. (Can be a follow up).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuTextWriter.java
```diff
@@ -0,0 +1,311 @@
...
+    private static void printValue(PrintWriter out, String value, boolean quote) {
+        if (quote) {
+            out.append('"').append(value).append('"');
+        } else {
+            out.append(value);
+        }
+    }
+
+    // Can a string be broken here? If not, backup until we can.
+    private static int goodBreak(String quoted, int end) {
```
srl295: should this actually use the line break iterator (that's what it's for!!) to backup to a base char? i.e. to not split `é` into `e`+\n+`´`? A previous ICU's LB iterator should be fine.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PathMatcher.java
```diff
@@ -0,0 +1,259 @@
...
+/**
+ * An immutable matcher for {@link CldrPath} instances. A path matcher specification looks like
+ * {@code "foo/*[@x="z"]/bar[@y=*]"}, where element names and attribute values can be wildcards.
+ *
+ * <p>Note that the path fragment represented by the specification does not include either leading
+ * or trailing {@code '/'}. This is because matching can occur at any point in a {@code CdlrPath}.
+ * The choice of where to match in the path is governed by the match method used (e.g.
+ * {@link PathMatcher#matchesSuffixOf(CldrPath)}.
+ */
+public abstract class PathMatcher {
```
srl295: seems like it shoul dmove into CLDR API… eventually.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java
hagbard: Given the discussion started in https://docs.google.com/document/d/1QmpLD82RKBS0CDbWkIQaWFd80Ne_XiEtXBK88n9a5Yg/edit#
which needs people to look at it and come to a concensus, I'm strongly in favour of not letting this specific detail block the initial submission. There's a bunch more thinking and work around the config stuff to be done, but it's not important if it's done now or in a much small targeted PR in a week or so.

srl295: OK- i can see that.

hagbard:
I made a small fix to the locale regex config (I had tried to be clever and it backfired). Other than that, I'll intended to submit this on Tuesday (giving anyone else who wants to review a chance to shout before then).

Steven - please note that I cannot submit myself, and I need a "proper" LGTM when you're happy.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuTextWriter.java
hagbard: Ta, I'll add a TODO now and I (or anyone else) can followup later.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuTextWriter.java
hagbard: Ahh, thanks. I'm actually not super familiar with this, so I'll see about making a change, but you'll have to check it carefully to make sure I'm doing the right thing (I'll add unit tests since this is going to be rarely hit IRL).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PathMatcher.java
hagbard: I'm still not happy with the API, it has some strange wrinkles, so it needs a bit of a rethink before then, but in general yes, it or something like it should be generally available.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuTextWriter.java
hagbard: Mind if I do this in a followup CL (and add a TODO now) since it seems more than just a few lines changed? Or, we could just not line wrap at all since it doesn't seem structurally necessary (though this should be done later since it changes output) ??

macchiati: Yes, looks more complicated. It should use a linebreak iterator, but it should also have a specialized CharSequence that knows about the escapes that could be used.

Alternatively, if there are any escapes, then unescape the whole string, find the linebreak, then escape (that's likely to be much simpler).

hagbard: Ahh thanks. Seems like since it's producing identical output now, this is probably best handled via a followup ticket/pr. I'll be sure to promote the TODO that I just added to a ticket when I do the sweep.

hagbard:
Unless anyone has any more comments, I'd like to get this is today.

PR #722, opened 2019-08-02
ICU-20737 Removing Python dependency on distutils.

sffc:
Deletes tstfiles.mk and merges the list into BUILDRULES.py

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20737
- [x] Issue accepted
- [x] Tests included


sffc:
Ping @srl295

PR #724, opened 2019-08-03
ICU-20618 regex nested look around, fix restore of original input bounds when leaving blocks

aheninger:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20618
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/regexcmp.cpp
+++ b/icu4c/source/i18n/regexcmp.cpp
@@ -561,7 +561,7 @@ UBool RegexCompile::doParseActions(int32_t action)
         //               sequence; don't change without making updates there too.
         //
         // Compiles to
-        //    1    START_LA     dataLoc     Saves SP, Input Pos
+        //    1    LA_START     dataLoc     Saves SP, Input Pos, Active input region.
         //    2.   STATE_SAVE   4            on failure of lookahead, goto 4
         //    3    JMP          6           continue ...
         //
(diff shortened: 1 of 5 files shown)


aheninger:
The problem can occur when a regex pattern contains nested look-arounds, a look-behind expression that itself contained a look-ahead, for example.

Matching operations can be bounded to operate on a limited range of the input text, by the region() function, or by nature of iterating through a string with find(), for example.

Look ahead or behind patterns further adjust this region; look behind limits the match to not extend beyond the position at the time the look behind was entered, for example, while "transparent bounds" can loosen the limits. The upshot is that the active region is modified when a look around expression is entered, and restored when it is exited.

The bug was that the restore did not allow for nested look-arounds, but just assumed that exiting was returning to the outer, non-look-around matching context. Without nesting, all was well.

The fix is to explicitly save the input region when entering a look-around block, and to restore from the saved values when exiting the block. Which requires changing the behavior of the match engine opcodes involved with look-around entry and exit. (Regex patterns are compiled into a p-code like opcodes which are subsequently interpreted during matching. The handling of look-around is one of the trickier bits of this.)

The exact design of the opcodes and sequences for look-around patterns could be simplified, but for now I just want to fix the specific bug while minimizing other changes.

younies:
I have two questions regarding to this PR:
  1 - shall we fill a jira ticket for simplifying the the opcodes and sequences for look-around patterns to do it in the future ?
  2 - I do not see any test cases added.

aheninger:
Younies asks

> shall we fill a jira ticket for simplifying the the opcodes and sequences for look-around patterns to do it in the future ?

On balance, I don't think so. The current implementation works. There wouldn't be a performance difference. So, significant effort for a relatively minor cleanup. Probably best to put any extra development resources into something more pressing. And not have yet another low priority ticket that sits around for years.

> I do not see any test cases added.

All of the additions to `source/test/testdata/regextst.txt` are test cases for this change. No new test code was needed, just new test cases for the data driven test. The test file format is described at the top of the file.

PR #728, opened 2019-08-05
ICU-20733 Fix build error with /Zc:twoPhase and vs2017

deepak1556:
Fixes https://unicode-org.atlassian.net/browse/ICU-20733

Don't export LocalizedNumberRangeFormatter class to workaround
an issue in MSSTL <xxatomic> that requires the sizeof(T) of
exported class.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20733
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/numberrangeformatter.h
+++ b/icu4c/source/i18n/unicode/numberrangeformatter.h
@@ -185,28 +185,6 @@ class DecimalQuantity;
 class UFormattedNumberRangeData;
 class NumberRangeFormatterImpl;
 
-} // namespace impl
-
-/**
- * \cond
- * Export an explicit template instantiation. See datefmt.h
- * (When building DLLs for Windows this is required.)
- */
-#if U_PLATFORM == U_PF_WINDOWS && !defined(U_IN_DOXYGEN)
-} // namespace icu::number
-U_NAMESPACE_END
-
-template struct U_I18N_API std::atomic< U_NAMESPACE_QUALIFIER number::impl::NumberRangeFormatterImpl*>;
-
-U_NAMESPACE_BEGIN
-namespace number {  // icu::number
-#endif
-/** \endcond */
-
-// Other helper classes would go here, but there are none.
-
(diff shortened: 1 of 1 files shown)


jefgen:
@deepak1556 Thanks for this PR!

The Azure CI Java build looks to have hit a network error, so I've requested a rebuild for it.
(Though they should be fine since this change isn't touching anything related to ICU4J).

However, the OSX builds are reporting the following linker error:

```
Undefined symbols for architecture x86_64:

  "icu_64::number::NumberRangeFormatterSettings<icu_64::number::LocalizedNumberRangeFormatter>::clone() const", referenced from:

      NumberRangeFormatterTest::toObject() in numbertest_range.o

  "icu_64::number::NumberRangeFormatterSettings<icu_64::number::LocalizedNumberRangeFormatter>::numberFormatterBoth(icu_64::number::UnlocalizedNumberFormatter&&)", referenced from:

      NumberRangeFormatterTest::testCopyMove() in numbertest_range.o

  "icu_64::number::NumberRangeFormatterSettings<icu_64::number::LocalizedNumberRangeFormatter>::clone()", referenced from:

      NumberRangeFormatterTest::toObject() in numbertest_range.o
```

(Aside: This wasn't noticed before when testing your patch as the Azure CI builds which run on my forked repo don't build anything OSX.)

The odd thing is that all of the public methods of the LocalizedNumberRangeFormatter are exposed with U_I18N_API, but perhaps the compiler on OSX needs the entire class exported?

It seems like a hack, but maybe the entire class could be exported only on OSX, and just export the public methods everywhere else? (I'm not familiar with development on OSX, so maybe I'm missing something...)

jefgen:
Hello,
From discussion in the ICU-TC call on 2019-11-06, it was agreed to close PRs older than 3 months that have been waiting on author. Note the the JIRA ticket itself will remain open, this is just closing the PR on GitHub.

(For this particular issue, if I get some time in the future I might revisit this ticket though.)

PR #729, opened 2019-08-05
WIP change BC, add dayPeriod for gregorian and fields

ziryan:
##### Checklist

- [ ] Issue accepted
- [ ] Tests included


jefgen:
Thank you for opening this PR!

This should likely be submitted to the CLDR project instead though, as that is where the data for ICU comes from:
https://github.com/unicode-org/cldr/

PR #726, opened 2019-08-03
ICU-20751 Fix xtensa build

bkuhls:
Solves build error
In file included from double-conversion.h:42:0,
                 from number_decimalquantity.cpp:19:
double-conversion-utils.h:119:2: error: #error Target architecture was not detected as supported by Double-Conversion.
 #error Target architecture was not detected as supported by Double-Conversion.

detected by buildroot autobuilders:
http://autobuild.buildroot.net/results/e6c/e6c6f826d8280f2bef923b3327d2a9a13ce2f7e1/

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-_____
- [ ] Updated PR title and link in previous line to include Issue number
- [ ] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #731, opened 2019-08-06
ICU-20756 add xtensa architecture support to double-conversion-utils.h

jcmvbkbc:
Building of icu4c for the Xtensa architecture currently fails with the
following error message:

  double-conversion-utils.h:120:2: error: #error Target architecture was
  not detected as supported by Double-Conversion.

Xtensa uses 64-bit doubles and passes the test case in the
double-conversion-utils.h. Add it to the list of supported
architectures.

Signed-off-by: Max Filippov <jcmvbkbc@gmail.com>

##### Checklist

- [+] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20756
- [+] Updated PR title and link in previous line to include Issue number
- [-] Issue accepted
- [-] Tests included
- [-] Documentation is changed or added


jcmvbkbc:
Duplicates PR #726

PR #732, opened 2019-08-07
ICU-20758 Add a StringPiece constructor for any string view type

roubert:
https://unicode-org.atlassian.net/browse/ICU-20758

Using the C++ SFINAE (substitution failure is not an error) technique,
it's possible to provide an icu::StringPiece constructor that accepts
any string view type that might be available in the current compilation
unit, without adding any additional dependencies or configuration flags
to ICU.

### DIFF
--- a/icu4c/source/common/unicode/stringpiece.h
+++ b/icu4c/source/common/unicode/stringpiece.h
@@ -31,6 +31,9 @@
 
 #if U_SHOW_CPLUSPLUS_API
 
+#include <cstddef>
+#include <type_traits>
+
 #include "unicode/uobject.h"
 #include "unicode/std_string.h"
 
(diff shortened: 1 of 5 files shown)


icu4c/source/common/unicode/stringpiece.h
```diff
@@ -77,6 +80,29 @@ class U_COMMON_API StringPiece : public UMemory {
    */
   StringPiece(const std::string& str)
     : ptr_(str.data()), length_(static_cast<int32_t>(str.size())) { }
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Constructs from some other implementation of a string piece class, from any
+   * C++ record type that has these two methods:
+   *
+   *   struct OtherStringPieceClass {
```
markusicu: Is this indentation sufficient for doxygen to turn it into a `<pre>` block?

icu4c/source/configure.ac
```diff
@@ -903,6 +903,14 @@ else
 fi
 AC_SUBST(U_HAVE_STRTOD_L)
 
+# Check if C++17 std::string_view is available.
```
markusicu: Do we need this?

Why not extend the conditions for U_CPLUSPLUS_VERSION in unicode/platform.h to detect C++17 (`__cplusplus >= 201703L`), and then make the test conditional on `#if U_CPLUSPLUS_VERSION >= 17`?

roubert: In some environments (eg. current Apple Xcode), std::string_view is available even when a C++11 build is requested, so I thought it better to just run the std::string_view unit test whenever it's actually available instead of trying to test for conditions where it should be available.

icu4c/source/common/unicode/stringpiece.h
roubert: No, apparently it was not. I ran Doxygen and it seems as if with the configuration used for ICU4C there isn't any indentation sufficient to automatically turn this into a code block. I've updated to use a `\code` annotation now instead.

icu4c/source/configure.ac
markusicu: It adds a lot of "configure" machinery... but ok.

icu4c/source/common/unicode/stringpiece.h
markusicu: Shane has successfully used a relative indent of 4 spaces to make it work. But `\code` is good, too.

roubert:
> lgtm please squash

done

PR #734, opened 2019-08-08
ICU-20773 Adds an explanation to the pre-flighting.

filmil:
A non-obvious feature of the pre-flighting is that it will report a
buffer overflow error code during pre-flighting, while still reporting
the discovered target buffer size correctly.

So in order to pre-flight correctly, one must ignore that particular error
code.  While the documentation is technically correct in that the code will
report overflow when overflow would have happened had we requested a
conversion, it maybe stands to reason to emphasize that this particular
use case need special handling.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20773
- [ ] Issue accepted
- [ ] Tests included


markusicu:
Discuss in ticket.

markusicu:
For reasons stated in the ticket, I propose closing this PR.
I think we need to figure out a more "global" way to tell new ICU users about general basics.

PR #735, opened 2019-08-08
ICU-20517 Removing dead code in rbnf.cpp

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20517
- [x] Issue accepted
- [ ] Tests included


sffc:
I was a little bit too aggressive with removing the dead code.  I added some back again.

PR #736, opened 2019-08-08
ICU-20766 Fix failing TravisCI build for ICU4J due to missing ANT.

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20766
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,11 @@ matrix:
 
     - name: "j"
       language: java
+      addons:
+        apt:
+          packages:
+            - ant
+            - ant-optional
       env:      BUILD=ICU4J
       before_script:
         - cd icu4j


.travis.yml
```diff
@@ -25,6 +25,11 @@ matrix:
 
     - name: "j"
       language: java
+      addons:
+        apt:
+          packages:
+            - ant
+            - ant-optional
```
sffc: What is ant-optional and do we need it?  (minor issue)

jefgen: Turns out that we do need it. :)
I first tried with just `ant` and while that did unblock the ICU4J builds, they eventually failed later on because the Ant JUnit plugins/extensions weren't installed.
They are part of the Ant Optional package, so it seems that we do need it.
(Not sure if that might be worth adding a comment to the YML file for or not though...?)

PR #744, opened 2019-08-09
ICU-20759 Add ucal_getDefaultTimeZone C API for host OS timezone detection

jefgen:
In ICU4C there is a C++ API for getting the host OS timezone (not ICU's default time zone) which doesn't cache the result: `TimeZone::detectHostTimeZone`.

This is useful for applications that want to determine the host OS timezone, and to detect if it has changed.

However, there is no corresponding C API for this functionality.

This adds a new public C API `ucal_getHostTimeZone` which exposes this functionality to C API users.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20759
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/ucal.cpp
+++ b/icu4c/source/i18n/ucal.cpp
@@ -34,7 +34,7 @@ U_NAMESPACE_USE
 static TimeZone*
 _createTimeZone(const UChar* zoneID, int32_t len, UErrorCode* ec) {
     TimeZone* zone = NULL;
-    if (ec!=NULL && U_SUCCESS(*ec)) {
+    if (ec != NULL && U_SUCCESS(*ec)) {
         // Note that if zoneID is invalid, we get back GMT. This odd
         // behavior is by design and goes back to the JDK. The only
         // failure we will see is a memory allocation failure.
@@ -69,7 +69,7 @@ ucal_openCountryTimeZones(const char* country, UErrorCode* ec) {
 U_CAPI int32_t U_EXPORT2
 ucal_getDefaultTimeZone(UChar* result, int32_t resultCapacity, UErrorCode* ec) {
     int32_t len = 0;
-    if (ec!=NULL && U_SUCCESS(*ec)) {
+    if (ec != NULL && U_SUCCESS(*ec)) {
         TimeZone* zone = TimeZone::createDefault();
         if (zone == NULL) {
             *ec = U_MEMORY_ALLOCATION_ERROR;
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/ucal.cpp
```diff
@@ -91,6 +91,23 @@ ucal_setDefaultTimeZone(const UChar* zoneID, UErrorCode* ec) {
     }
 }
 
+U_DRAFT int32_t U_EXPORT2
+ucal_getHostTimeZone(UChar* result, int32_t resultCapacity, UErrorCode* ec) {
+    int32_t len = 0;
+    if (ec!=NULL && U_SUCCESS(*ec)) {
```
axelandrejs: Space before and after !=

jefgen: Thanks, fixed!

icu4c/source/i18n/ucal.cpp
```diff
@@ -91,6 +91,23 @@ ucal_setDefaultTimeZone(const UChar* zoneID, UErrorCode* ec) {
     }
 }
 
+U_DRAFT int32_t U_EXPORT2
+ucal_getHostTimeZone(UChar* result, int32_t resultCapacity, UErrorCode* ec) {
+    int32_t len = 0;
+    if (ec != NULL && U_SUCCESS(*ec)) {
+        TimeZone *zone = TimeZone::detectHostTimeZone();
```
jefgen: Note for Yoshito: I didn't use a `LocalPointer` here as the rest of the file currently doesn't use it at all.

yumaoka: OK.

icu4c/source/i18n/ucal.cpp
```diff
@@ -91,6 +91,23 @@ ucal_setDefaultTimeZone(const UChar* zoneID, UErrorCode* ec) {
...
+        if (zone == NULL) {
+            *ec = U_MEMORY_ALLOCATION_ERROR;
```
yumaoka: It's interesting - but I agree TimeZone::detectHostTimeZone() would return null only when unknown zone object cannot be created because of out of memory.

jefgen: Right, `TimeZone::getUnknown()` should give back the unknown zone if things fail.
I think maybe I might add a brief line to the API docs that this might also return NULL with U_MEMORY_ALLOCATION_ERROR in very bad cases where we can't even clone the unknown zone. What do you think?

PR #745, opened 2019-08-09
ICU-20712 Add support for Windows UWP ARM/ARM64 data DLLs.

jefgen:
This adds support for building Windows UWP ARM and ARM64 data DLLs.

This also changes the `pkgdata` options to decouple ARM/ARM64 from UWP, and adds a generic architecture option (`-a`) for passing the value of the `/MACHINE:` arch on the command line. (Ex: `-a ARM64`).

Additionally this change also adds another option (`-b`) to set `/DYNAMICBASE` on the data DLL, which is required for the ARM64 builds.

(Both of these changes also work towards future work to enable ARM64 desktop builds.)

Note: I think it would be better in general to completely remove the fixed base addresses from ICU altogether. I’ve filed another ticket/bug for this here: https://unicode-org.atlassian.net/browse/ICU-20768

This is also related to the following ticket, but the ticket checker only accepts one ticket on a PR:
[ICU-20670 Add support to generate ICU data DLL for Windows arm64](https://unicode-org.atlassian.net/browse/ICU-20670)

This change also *removes* a number of ifdefs/forks for `U_PLATFORM_HAS_WINUWP_API` and changes them to forks for `ICU_DATA_DIR_WINDOWS`, in order to both unblock the usage of the data DLL in the UWP scenario, as well as to further reduce the divergence of the UWP projects.

(cc/FYI: @marswe).

Note: There are currently multiple commits in this PR; once this is reviewed I will squash them into a single commit and then force-push.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20712
- [x] Issue accepted
- [ ] Tests included


daniel-ju:
Do we have to update the `#if U_PLATFORM_HAS_WINUWP_API == 0` conditional in `umapfile.cpp` as well to fix the build errors?

jefgen:
> Do we have to update the `#if U_PLATFORM_HAS_WINUWP_API == 0` conditional in `umapfile.cpp` as well to fix the build errors?

Thanks Daniel.
Hmm, no I think this is related to the changing the import libs around to fix the WACK issues.
These are supposed to be present in `onecore.lib`. I'll need to investigate some more...

jefgen:
Thanks for the review!
I'm going to squash all of the commits into one commit next, and force-push, which will require one more review.

PR #748, opened 2019-08-12
ICU-13533 Cleanup testing with no data under the address sanitizer

aheninger:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-13533
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/nucnvtst.c
+++ b/icu4c/source/test/cintltst/nucnvtst.c
@@ -3095,17 +3095,18 @@ TestHZ() {
     UChar *uTarget;
     char *cTarget;
     const char *cTargetLimit;
-    char *cBuf;
-    UChar *uBuf,*test;
+    char *cBuf = NULL;
+    UChar *uBuf = NULL;
+    UChar *test;
     int32_t uBufSize = 120;
     UErrorCode errorCode=U_ZERO_ERROR;
-    UConverter *cnv;
+    UConverter *cnv = NULL;
     int32_t* offsets = (int32_t*) malloc(uBufSize * sizeof(int32_t) * 5);
     int32_t* myOff= offsets;
     cnv=ucnv_open("HZ", &errorCode);
     if(U_FAILURE(errorCode)) {
         log_data_err("Unable to open HZ converter: %s\n", u_errorName(errorCode));
-        return;
+        goto cleanup;
     }
 
     uBuf =  (UChar*)malloc(uBufSize * sizeof(UChar)*5);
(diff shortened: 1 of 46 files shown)


aheninger:
This PR fixes some, but not all, of the problems identified by the address sanitizer when running ICU tests with no data.

This is something that I had started last year, then set aside. This PR mostly consists of picking up the old svn branch, rebasing it up to the current master, and then making sure everything still works.

No data testing under Asan is interesting because it exercises error handling paths in the code that aren't hit with normal testing. It finds numerous memory leaks and a few uses of bad pointers. Most of the problems are in test code, but a few are in the library code itself.

This is a large set of changes. For reviewing, I suggest looking carefully at the changes to the library code itself (few and small), and skimming and spot checking the changes to tests.

The changes to tests are mostly to fix memory leaks, usually to make sure that temporary objects are being correctly deleted when a function returns early due to other errors. Usually one of the following, whatever was easiest in the context:

- Replace heap allocated objects with local (stack) objects. This is the easiest change when the objects are created with constructors. Not usable with factory methods.

- Use LocalPointer<type>. Or, sometimes, std::unique_ptr.

- Add an explicit delete. Quickest if the function had almost correct error handling already, and switching it over to LocalPointer would be tedious.

I was going for quick and limited changes, more than refactoring everything that could be improved.

icu4c/source/test/cintltst/nucnvtst.c
```diff
@@ -3314,17 +3317,18 @@ TestISO_2022_JP() {
     UChar *uTarget;
     char *cTarget;
     const char *cTargetLimit;
-    char *cBuf;
-    UChar *uBuf,*test;
+    char *cBuf = NULL;
+    UChar *uBuf = NULL;
+    UChar *test;
```
gnrunge: Does this variable serve a purpose?

icu4c/source/test/intltest/tzfmttst.cpp
```diff
@@ -728,7 +728,8 @@ void TimeZoneFormatTest::RunAdoptDefaultThreadSafeTests(int32_t threadNumber) {
     if (threadNumber % 2 == 0) {
         for (int32_t i = 0; i < kAdoptDefaultIteration; i++) {
             std::unique_ptr<icu::StringEnumeration> timezones(
-            icu::TimeZone::createEnumeration());
+                    icu::TimeZone::createEnumeration());
+            if (!assertTrue(WHERE, (bool)timezones, false, true)) {return;}  // Fails with missing data.
```
gnrunge: Comment above may help the eye.

PR #751, opened 2019-08-13
ICU-20524 Adding variables for relative file replacement directory.

sffc:
Also fixes some bugs with file replacement.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20524
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/docs/userguide/icu_data/buildtool.md
+++ b/docs/userguide/icu_data/buildtool.md
@@ -702,8 +702,21 @@ configuration file:
         {
           src: "Zawgyi_my.txt"
           dest: "translit/Zawgyi_my.txt"
-        }
+        },
+        "misc/dayPeriods.txt"
       ]
     }
+
+`directory` should either be an absolute path, or a path starting with one of
+the following, and it should not contain a trailing slash:
+
+- "$SRC" for the *icu4c/source/data* directory in the source tree
+- "$FILTERS" for the directory containing filters.json
+- "$CWD" for your current working directory
+
+When the entry in the `replacements` array is an object, the `src` and `dest`
+fields indicate, for each file in the source directory (`src`), what file in
+the ICU hierarchy it should replace (`dest`). When the entry is a string, the
+same relative path is used for both `src` and `dest`.
 
 Whole-file substitution happens before all other filters are applied.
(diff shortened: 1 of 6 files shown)


sffc:
Note: I don't think this needs an API proposal.  I just changed the code and updated the docs.  I haven't emailed icu-design for any of the other data filter tool changes either.

Also, there is some confusion about what features are available in ICU 64 versus master.  There was at least one support email from a user thinking they could use a documented feature but the feature wasn't available yet in ICU 64.  I'm not sure the best way to solve this.  Once we get the userguide set up better, we can make a versioned userguide.

docs/userguide/icu_data/buildtool.md
```diff
@@ -702,8 +702,21 @@ configuration file:
...
+`directory` should either be an absolute path, or a path starting with one of
+the following:
+
+- "$SRC" for the *icu4c/source/data* directory in the source tree
+- "$FILTERS" for the directory containing filters.json
+- "$CWD" for your current working directory
+
+When the entry in the `replacements` array is an object, the `src` and `dest`
+fields indicate, for each file in the source directory (`src`), what file in
+the ICU heirarchy it should replace (`dest`). When the entry is a string, the
```
markusicu: typo --> hierarchy

icu4c/source/python/icutools/databuilder/filtration_schema.json
```diff
@@ -49,17 +49,25 @@
         "fileReplacements": {
             "type": "object",
             "properties": {
-                "directory": { "type": "string" },
+                "directory": {
+                    "type": "string",
+                    "pattern": "^(/|\\$SRC|\\$FILTERS|\\$CWD)?.*$"
```
markusicu: should the variable patterns each include a trailing `/` ?

sffc: There is not expected to be a trailing slash.  Maybe it would be better to actually forbid one.

sffc:
Amended to fix the regular expression as discussed.

sffc:
Amending to attempt to fix the new CI failure.

PR #758, opened 2019-08-15
ICU-20776 Checks for nullness in UMutex::lock()

filmil:
Adds `U_ASSERT` check before using `m->lock()` to make allocation issues
more apparent at least in debug builds.

There is probably quite a few places like this, but let's try fixing
broken windows.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20776
- [ ] Tests included
- [ ] Issue accepted

### DIFF
--- a/icu4c/source/common/umutex.cpp
+++ b/icu4c/source/common/umutex.cpp
@@ -98,6 +98,7 @@ std::mutex *UMutex::getMutex() {
             gListHead = this;
         }
     }
+    U_ASSERT(retPtr != nullptr);
     return retPtr;
 }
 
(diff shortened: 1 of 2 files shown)


aheninger:
Note that a memory allocation failure -- new() returning nullptr -- is not possible here, because a placement new into existing storage is being used, not a heap allocation.

The change still is reasonable, I think, to document the expectations and to catch any unanticipated issues.

filmil:
Thanks for explaining.  That said, please note that the JIRA issue will need to be accepter (it's "New" at the moment) for this to be merged.

PR #759, opened 2019-08-15
ICU-20601 Wrap public ICU API compound macros in do { } while.

roubert:
https://unicode-org.atlassian.net/browse/ICU-20601

This is the normal way of making compound macros behave like statements
and allowing macro invocations that look like function calls to work
just like if they were function calls.

Before ICU 65, function like multi statement ICU macros were just
defined as series of statements wrapped in { } blocks and the caller
could choose to either treat them as if they were actual functions and
end the invocation with a trailing ; creating an empty statement after
the block or else omit this trailing ; using the knowledge that the
macro would expand to { }.

To continue supporting code that depends on that old behaviour, on being
able to omit the trailing ; when invoking one of these macros, the new
surrounding do { } while is done through preprocessor defines that can
easily be overridden at compile time, when needed.

### DIFF
--- a/icu4c/source/common/unicode/umachine.h
+++ b/icu4c/source/common/unicode/umachine.h
@@ -140,6 +140,41 @@
 #define U_FINAL final
 #endif
 
+// Before ICU 65, function-like, multi-statement ICU macros were just defined as
+// series of statements wrapped in { } blocks and the caller could choose to
+// either treat them as if they were actual functions and end the invocation
+// with a trailing ; creating an empty statement after the block or else omit
+// this trailing ; using the knowledge that the macro would expand to { }.
+//
+// But doing so doesn't work well with macros that look like functions and
+// compiler warnings about empty statements (ICU-20601) and ICU 65 therefore
+// switches to the standard solution of wrapping such macros in do { } while.
+//
+// This will however break existing code that depends on being able to invoke
+// these macros without a trailing ; so to be able to remain compatible with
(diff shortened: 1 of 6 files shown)


icu4c/source/common/unicode/umachine.h
```diff
@@ -140,6 +140,40 @@
...
+//
+// CPPFLAGS='-DUPRV_MACRO_BEGIN="" -DUPRV_MACRO_END=""' runConfigureICU ...
+
+/**
+ * \def UPRV_MACRO_BEGIN
+ * Defined as the "do" keyword by default.
+ * @internal
+ */
+#ifndef UPRV_MACRO_BEGIN
+#define UPRV_MACRO_BEGIN do
```
markusicu: I find these names a bit too generic/ambiguous.

Suggestion: UPRV_BLOCK_MACRO_BEGIN & UPRV_BLOCK_MACRO_END

icu4c/source/common/unicode/umachine.h
```diff
@@ -140,6 +140,40 @@
 #define U_FINAL final
 #endif
 
+// Before ICU 65, function like multi statement ICU macros were just defined as
```
markusicu: Before ICU 65, function-like, multi-statement ICU macros were just defined as

icu4c/source/common/unicode/umachine.h
roubert: Done.

PR #760, opened 2019-08-15
ICU-20601 Wrap ICU implementation compound macros in do { } while.

roubert:
This does the same for the ICU implementation code as was done for the public ICU API in PR #759.

This is not necessary to do in order to enable users of ICU to use -Wextra-semi-stmt when building their own code (that has now already been resolved), but it will make it possible to use this warning when building ICU itself which is useful both to weed out mistakes in the ICU code itself and to prevent future changes that cause this warning to creep in into the public ICU API. It'll also make macro definitions consistent in ICU between API and implementation code.

### DIFF
--- a/icu4c/source/common/ubidiimp.h
+++ b/icu4c/source/common/ubidiimp.h
@@ -387,41 +387,49 @@ typedef union {
 } BidiMemoryForAllocation;
 
 /* Macros for initial checks at function entry */
-#define RETURN_IF_NULL_OR_FAILING_ERRCODE(pErrcode, retvalue)   \
-        if((pErrcode)==NULL || U_FAILURE(*pErrcode)) return retvalue
-#define RETURN_IF_NOT_VALID_PARA(bidi, errcode, retvalue)   \
-        if(!IS_VALID_PARA(bidi)) {  \
-            errcode=U_INVALID_STATE_ERROR;  \
-            return retvalue;                \
-        }
-#define RETURN_IF_NOT_VALID_PARA_OR_LINE(bidi, errcode, retvalue)   \
-        if(!IS_VALID_PARA_OR_LINE(bidi)) {  \
-            errcode=U_INVALID_STATE_ERROR;  \
-            return retvalue;                \
-        }
-#define RETURN_IF_BAD_RANGE(arg, start, limit, errcode, retvalue)   \
-        if((arg)<(start) || (arg)>=(limit)) {       \
-            (errcode)=U_ILLEGAL_ARGUMENT_ERROR;     \
-            return retvalue;                        \
-        }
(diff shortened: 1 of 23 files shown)


icu4c/source/common/ubidiimp.h
```diff
@@ -387,41 +387,49 @@ typedef union {
...
-        if(!IS_VALID_PARA_OR_LINE(bidi)) {  \
-            errcode=U_INVALID_STATE_ERROR;  \
-            return;                \
-        }
-#define RETURN_VOID_IF_BAD_RANGE(arg, start, limit, errcode)   \
-        if((arg)<(start) || (arg)>=(limit)) {       \
-            (errcode)=U_ILLEGAL_ARGUMENT_ERROR;     \
-            return;                        \
-        }
+#define RETURN_IF_NULL_OR_FAILING_ERRCODE(pErrcode, retvalue) UPRV_BLOCK_MACRO_BEGIN { \
```
markusicu: Option/question: This didn't use a `{ block }` before. Did you change it so that it would not accidentally consume a following "else"?
(It's better now, but wondering.)

icu4c/source/i18n/decNumberLocal.h
```diff
@@ -259,7 +259,7 @@
   /* 2,000,000,000 (as is needed for negative exponents of            */
   /* subnormals).  The unsigned integer pow is used as a temporary    */
   /* variable. */
-  #define TODIGIT(u, cut, c, pow) {       \
+  #define TODIGIT(u, cut, c, pow) UPRV_BLOCK_MACRO_BEGIN { \
```
markusicu: Optional: This one might interfere with syncing with the upstream decnumber library. I know we already add copyright headers; do you have a sense of how much else we have patched?

I wonder if we should, in this case,
- use literal "do" and "while (0)"
- submit an upstream ticket or patch

icu4c/source/common/ubidiimp.h
roubert: Yes, I updated all these to not accidentally consume trailing else-statements.

icu4c/source/i18n/decNumberLocal.h
roubert: I don't know. The fact that the file has a perfectly normal ICU path name made me just assume that it is to be treated like any other ICU source file. If it ought to be treated differently, I would have expected it to be located in some kind of separate location in order to clearly delineate that it should be treated differently from the rest of the ICU source code.

markusicu: We keep a pristine copy of the upstream code in the "vendor" folder, and then copy into the regular ICU library location with modifications as necessary. Looking at https://github.com/unicode-org/icu/tree/master/vendor/double-conversion/upstream I don't see this file, so this change is fine.

PR #764, opened 2019-08-17
ICU-20768 Remove fixed DLL base addresses when building Windows DLLs.

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20768
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/cintltst.vcxproj
+++ b/icu4c/source/test/cintltst/cintltst.vcxproj
@@ -75,8 +75,6 @@
       <AdditionalLibraryDirectories>..\..\..\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <ProgramDatabaseFile>.\x86\Debug/cintltst.pdb</ProgramDatabaseFile>
       <SubSystem>Console</SubSystem>
-      <RandomizedBaseAddress>false</RandomizedBaseAddress>
-      <FixedBaseAddress>false</FixedBaseAddress>
       <DataExecutionPrevention>
       </DataExecutionPrevention>
     </Link>
@@ -102,7 +100,6 @@
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <ProgramDatabaseFile>.\x86\Release/cintltst.pdb</ProgramDatabaseFile>
       <SubSystem>Console</SubSystem>
-      <RandomizedBaseAddress>false</RandomizedBaseAddress>
       <DataExecutionPrevention>
       </DataExecutionPrevention>
     </Link>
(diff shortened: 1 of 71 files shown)


daniel-ju:
>       <RandomizedBaseAddress>false</RandomizedBaseAddress>

Is this supposed to be removed as well (in addition to debug config below)?

---
Refers to: icu4c/source/layoutex/layoutex.vcxproj:80 in 9f80b7b. [](commit_id = 9f80b7be88562b809f586537bff042798523d502, deletion_comment = True)

daniel-ju:
I did a quick search of <RandomizedBaseAddress> and it seems like there are more occurrences under the `test` directory, do we want to remove those too?

jefgen:
Thanks Daniel!

I believe all of the layoutex stuff is deprecated, so I'm not sure how much it matters.
That said, we should probably remove it completely from everything.
I'll push another commit for the layoutex and tests.

markusicu:
layoutex is not deprecated. it is usable on top of the HarfBuzz-LayoutEngine wrapper.
http://userguide.icu-project.org/layoutengine

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -169,7 +168,6 @@ static UOption options[]={
     /*21*/    UOPTION_DEF("zos-pds-build", 'z', UOPT_NO_ARG),
     /*22*/    UOPTION_DEF("windows-uwp-build", 'u', UOPT_NO_ARG),
     /*23*/    UOPTION_DEF("windows-DLL-arch", 'a', UOPT_REQUIRES_ARG),
-    /*24*/    UOPTION_DEF("windows-dynamicbase", 'b', UOPT_NO_ARG),
```
markusicu: I suggest you retain this option but make it a no-op and adjust its help text.

icu4c/source/test/cintltst/cintltst.vcxproj
```diff
@@ -75,7 +75,6 @@
       <AdditionalLibraryDirectories>..\..\..\lib;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
       <ProgramDatabaseFile>.\x86\Debug/cintltst.pdb</ProgramDatabaseFile>
       <SubSystem>Console</SubSystem>
-      <RandomizedBaseAddress>false</RandomizedBaseAddress>
       <FixedBaseAddress>false</FixedBaseAddress>
```
markusicu: I see this in a couple of files. Is it unnecessary/redundant now? I suggest doing a grep for "BaseAddress" in _all_ files.

jefgen: Right, this is redundant/unnecessary now. (This sets `/FIXED:NO` when linking, which is also the default, so its unneeded in the `vcxproj` file).
I'll grep for "BaseAddress" in all files and remove any remaining left-overs.
Thanks Markus!

jefgen:
Thanks everyone for the reviews!

PR #767, opened 2019-08-22
ICU-20783 C++ covariant return types: clone(), freeze() & friends

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20783

### DIFF
--- a/icu4c/source/i18n/rbt.cpp
+++ b/icu4c/source/i18n/rbt.cpp
@@ -191,8 +191,8 @@ RuleBasedTransliterator::~RuleBasedTransliterator() {
     }
 }
 
-Transliterator* // Covariant return NOT ALLOWED (for portability)
-RuleBasedTransliterator::clone(void) const {
+RuleBasedTransliterator*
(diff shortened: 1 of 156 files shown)


icu4c/source/i18n/rbt.cpp
```diff
@@ -191,8 +191,8 @@ RuleBasedTransliterator::~RuleBasedTransliterator() {
     }
 }
 
-Transliterator* // Covariant return NOT ALLOWED (for portability)
-RuleBasedTransliterator::clone(void) const {
+RuleBasedTransliterator* // Covariant return NOT ALLOWED (for portability)
```
aheninger: Remove comment.

markusicu: done; found & removed a few more too

aheninger:
This regular expression
    \(.*\).*clone\(

finds a fair number of clones with casts, where the cast could potentially be removed.
I don't think it's necessary to change these to proceed with this PR.

PR #768, opened 2019-08-23
ICU-20782 Enable Windows desktop ARM and ARM64 support

jefgen:
This pull-request enables "classic" desktop builds of ICU4C for both ARM (32-bit) and ARM64 (64-bit) on Windows.

All but the two samples in the "`allinone`" project now have ARM and ARM64 project configurations, and build for Windows Desktop ARM/ARM64.

Note: In order to build the ARM/ARM64 data DLL, you need to first build x64 Release, as the ARM/ARM build uses the x64 bits in order to be able to cross-compile for ARM/ARM64. This allows for completely building ARM/ARM64 binaries using only x64 hardware.

The ARM/ARM64 builds require using a newer version of the Windows SDK than 8.1, so they have a separate `WindowsTargetPlatformVersion` which uses Windows 10 SDK version 10.0.16299.0 (aka RS3), which is the first version of the Windows SDK to support building ARM64 desktop applications.

In order to exercise building in the Azure DevOps CI pipeline, I had to extend the timeout of the CI builds as they now build x64, ARM, and ARM64 in that order.
(Unfortunately we can't run the tests for ARM/ARM64 in the CI builds, as that would require having actual ARM/ARM64 hardware to run on in the Azure DevOps CI pipelines, which doesn't exist.)

In addition to the above this PR also greatly cleans-up the ICU4C `.vcxproj` files, in order to remove redundant parts, fix inconsistencies, and make them more readable.
This introduces two new variables in the shared `*.props` files: `IcuBinOutputDir` and `IcuLibOutputDir` in order to further reduce the amount of duplicated lines in the individual `.vcxproj` files themselves.

(Note: All of the changes in the PR were edited manually by hand, as there is sadly no VS tool to assist with cleaning-up `.vcxproj` files.)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20782
- [x] Issue accepted
- [ ] Tests included

ALLOW_MANY_COMMITS=true


jefgen:
I didn't do the samples because I ran out of time. There are more samples than just these two. I was going to file another ticket for the samples. 😊


Thanks,
Jeff
________________________________

PR #772, opened 2019-08-27
ICU-20693 Basic Eclipse settings

hagbard:
This is by no means complete, but represents the basic settings for classpaths and import ordering.
Importantly, when IntelliJ is set up and told to use/import Eclipse settings, it produces 100% identical import statements (consistency is the most important thing).

##### Checklist

- [x] Issue accepted
- [ ] Tests included


hagbard:
Just to be clear this is what you get when you do a vanilla Eclipse import of the Maven project and then set the import order to match the other CLDR Eclipse project. I've not made any choices about anything else.

PR #774, opened 2019-08-27
ICU-20693 Unit tests for most conversion code and mappers.

hagbard:
Unit tests for most of the conversion classes and mappers. Still some to do and better regressions tests need looking at too. Added a few comments of behaviours that are tested for, but seem "odd".

Note that there was a bit of refactoring of non-test classes which is mainly because when I looked at coverage I realized some method were being called from exactly one place and could just be inlined with something just as straightforward or removed entirely. This reduces the surface area a bit and makes testing easier (since it's more orthogonal now).

Also added some more documentation where it was previously missed. The tool still produces identical data against the old code for the CLDR 35.1 data.

##### Checklist

- [x] Issue accepted
- [x] Tests included


tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/mapper/Bcp47MapperTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+        ImmutableList<IcuData> icuData = Bcp47Mapper.process(cldrData);
+        IcuData tzData = icuData.get(1);
+
+        // Only the type-map paths/values are split into the timezone data.
+        assertThat(tzData).getPaths().hasSize(4);
+        assertThat(tzData).hasValuesFor("/typeMap/tz/\"foo:bar\"", "escaped");
+        assertThat(tzData).hasValuesFor("/typeMap/tz/foo_bar", "unescaped");
+
+        // TODO: Check this is correct! Having foo/bar as the value rather than foo:bar seems wrong.
```
hagbard: @macchiati @markusicu - Please have a look at this since I think it's doing exactly what the old converter did (but perhaps this case was never hit or should not be hit). Either way it feels wrong that the path now contains "foo:bar" above, but the alias "name" is "foo/bar". Am happy to fix, or comment one way or the other.

hagbard:
Added missing copyright headers.

hagbard:
When someone gets around to reviewing this, [please ping me and I'll stop updating it with more tests.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuData.java
```diff
@@ -98,48 +93,11 @@ public void replace(RbPath rbPath, RbValue rbValue) {
         add(rbPath, rbValue);
     }
 
-    public void setVersion(String versionString) {
```
hagbard: While trying to test IcuData it became clear that these methods were in the wrong place, so I added the abstract mapper base class for them.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuData.java
```diff
@@ -155,16 +113,6 @@ private RbValue replacePathValues(String value) {
         return Collections.unmodifiableSet(paths);
     }
 
-    /** Returns whether the given path is present in this instance. */
```
hagbard: Some of these methods were used exactly once, so I inlined them to reduce the surface area.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/IcuDataDumper.java
```diff
@@ -260,14 +260,14 @@ void processLine(String line) {
...
             if (pathStack.isEmpty()) {
-                return RbPath.empty();
+                return RbPath.of();
             }
             List<String> segments = new ArrayList<>();
             Iterables.addAll(segments, pathStack);
             if (segments.get(0).matches("<[0-9]{4}>")) {
                 segments.remove(0);
             }
-            return segments.isEmpty() ? RbPath.empty() : RbPath.of(Lists.reverse(segments));
```
hagbard: Not even worth the special case treatment.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/AbstractPathValueMapper.java
```diff
@@ -0,0 +1,100 @@
...
+import org.unicode.icu.tool.cldrtoicu.RbPath;
+import org.unicode.icu.tool.cldrtoicu.RbValue;
+
+import com.google.common.collect.ListMultimap;
+
+/**
+ * An abstract parent class for any mappers based on {@code PathValueTransformer}. This ensures
+ * that transformation results are correctly processed when being added to IcuData instances.
+ */
+public abstract class AbstractPathValueMapper {
```
hagbard: This is where the code that was in IcuData ended up. It seems better to not have IcuData need to know about the semantics of the PathValueTransformer Result instances - that's very much a "mapper" concept.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/AbstractPathValueMapper.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Adds transformation results on the specified multi-map to this data instance. Results are
+     * handled differently according to whether they are grouped, or represent an alias value. If
+     * the value of an ungrouped result is itself a resource bundle path (including possibly having
+     * an array index) then the referenced value is assumed to be an existing path whose value is
+     * then substituted.
+     */
+    // TODO: Fix this to NOT implicitly rely of ordering of referenced values.
```
hagbard: I'm looking at a fix for this - currently it could fail if the replacement path exists later than the path referencing it.
If fails in an obviously enough way though and doesn't fail on any real-world data at the moment (but it definitely needs to be fixed).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/LocaleMapper.java
```diff
@@ -39,7 +40,7 @@
  * <p>This is currently driven by the {@code ldml2icu_locale.txt} configuration file via a
  * {@code RegexTransformer}, but could use any {@link PathValueTransformer} implementation.
  */
-public final class LocaleMapper {
+public final class LocaleMapper extends AbstractPathValueMapper {
```
hagbard: This class and SupplementalMapper were rejigged to use the new base class. This looks like a lot, but it's really only moving big blocks of code around, and making previously static methods instance methods.
Unfortunately the diff doesn't do a great job of spotting this (there's only about 10 lines actually different and everything else is code moving around a bit)

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/TransformsMapper.java
```diff
@@ -96,27 +106,6 @@ public static IcuData process(CldrDataSupplier src, Path ruleFileOutputDir) {
         RuleVisitor(Function<Path, PrintWriter> outFn) {
             this.outFn = checkNotNull(outFn);
             icuData.setFileComment("File: root.txt");
-
-            // I have _no_ idea what any of this is about, I'm just trying to mimic the original
```
hagbard: Moved into a separate method since it's so "special case-y".

hagbard:
Since this went stale due to changes in CLDR, I've had to roll in some more updates.

hagbard:
Still waiting for someone to merge this (there are 4 or 5 more PRs waiting now).

sffc:
This PR should have targeted the maint branch.  PR to fix:

https://github.com/unicode-org/icu/pull/842

PR #788, opened 2019-08-29
ICU-13745 fix UBS in gregocal

srl295:
[ICU-13745] fix undefined behavior in GregorianCalendar::setGregorianChange
- Julian days outside of INT32_MIN..INT32_MAX are normalized
- Add a test case

[ICU-13745]: https://unicode-org.atlassian.net/browse/ICU-13745

### DIFF
--- a/icu4c/source/i18n/gregocal.cpp
+++ b/icu4c/source/i18n/gregocal.cpp
@@ -324,26 +324,26 @@ GregorianCalendar::setGregorianChange(UDate date, UErrorCode& status)
     if (U_FAILURE(status)) 
         return;
 
-    fGregorianCutover = date;
-
     // Precompute two internal variables which we use to do the actual
     // cutover computations.  These are the normalized cutover, which is the
     // midnight at or before the cutover, and the cutover year.  The
     // normalized cutover is in pure date milliseconds; it contains no time
     // of day or timezone component, and it used to compare against other
     // pure date values.
-    int32_t cutoverDay = (int32_t)ClockMath::floorDivide(fGregorianCutover, (double)kOneDay);
-    fNormalizedGregorianCutover = cutoverDay * kOneDay;
+    double cutoverDay = ClockMath::floorDivide(date, (double)kOneDay);
 
-    // Handle the rare case of numeric overflow.  If the user specifies a
-    // change of UDate(Long.MIN_VALUE), in order to get a pure Gregorian
-    // calendar, then the epoch day is -106751991168, which when multiplied
-    // by ONE_DAY gives 9223372036794351616 -- the negative value is too
-    // large for 64 bits, and overflows into a positive value.  We correct
-    // this by using the next day, which for all intents is semantically
-    // equivalent.
-    if (cutoverDay < 0 && fNormalizedGregorianCutover > 0) {
-        fNormalizedGregorianCutover = (cutoverDay + 1) * kOneDay;
+    // Handle the rare case of numeric overflow where the user specifies a time
+    // outside of INT32_MIN .. INT32_MAX number of days.
+    
+    if (cutoverDay <= INT32_MIN) {
+        cutoverDay = INT32_MIN;
+        fGregorianCutover = fNormalizedGregorianCutover = cutoverDay * kOneDay;
+    } else if (cutoverDay >= INT32_MAX) {
+        cutoverDay = INT32_MAX;
+        fGregorianCutover = fNormalizedGregorianCutover = cutoverDay * kOneDay;
+    } else {
+        fNormalizedGregorianCutover = cutoverDay * kOneDay;
+        fGregorianCutover = date;
     }
 
     // Normalize the year so BC values are represented as 0 and negative
@@ -360,7 +360,7 @@ GregorianCalendar::setGregorianChange(UDate date, UErrorCode& status)
     fGregorianCutoverYear = cal->get(UCAL_YEAR, status);
     if (cal->get(UCAL_ERA, status) == BC) 
         fGregorianCutoverYear = 1 - fGregorianCutoverYear;
-    fCutoverJulianDay = cutoverDay;
+    fCutoverJulianDay = (int32_t)cutoverDay;
     delete cal;
 }
 
--- a/icu4c/source/test/intltest/calregts.cpp
+++ b/icu4c/source/test/intltest/calregts.cpp
@@ -96,6 +96,7 @@ CalendarRegressionTest::runIndexedTest( int32_t index, UBool exec, const char* &
         CASE(52,TestPersianCalOverflow);
         CASE(53,TestIslamicCalOverflow);
         CASE(54,TestWeekOfYear13548);
+        CASE(55,Test13745);
     default: name = ""; break;
     }
 }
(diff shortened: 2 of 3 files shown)


srl295:
does not have a specific test case yet

icu4c/source/i18n/gregocal.cpp
```diff
@@ -332,7 +332,7 @@ GregorianCalendar::setGregorianChange(UDate date, UErrorCode& status)
     // normalized cutover is in pure date milliseconds; it contains no time
     // of day or timezone component, and it used to compare against other
     // pure date values.
-    int32_t cutoverDay = (int32_t)ClockMath::floorDivide(fGregorianCutover, (double)kOneDay);
+    int64_t cutoverDay = ClockMath::floorDivide(fGregorianCutover, (double)kOneDay);
```
markusicu: I am not sure if your change fixes the issue completely.

This is what Fredrik found last year:
> The float-cast-overflow UBSan [undefined-behavior sanitizer] check ...
>
> ... setGregorianChange() is written to expect numeric overflow for certain inputs, in which case UBSan will erroneously fail like this:
>
> .../icu/source/i18n/gregocal.cpp:335:35: runtime error: -2.08066e+300 is outside the range of representable values of type 'int'

By casting the double result to int64_t rather than int32_t, you widen the range of values that fit, but doubles can still be much larger. The code (and the comment below) seems to assume that the UDate can be no lower than "Long.MIN_VALUE" which would be true in Java where we do use long, but not in C++ where we use UDate=double, unless the cutover setter limits the range of values.

Also, the code following the floorDivide() still has a hack of detecting an overflow by checking if dividing a negative value yields a positive quotient, but it seems like there is a 50/50 chance that an overflow would give us a negative value (assuming it just stuffs the double bits into the int64_t).

Wouldn't it be better to put the quotient (cutoverDay) into a double, do a range check in "double" space to detect what we need (I am not quite sure what the proper replacement for the "Handle the rare case..." hack should be), and prevent an overflow later when converting to the int32_t fCutoverJulianDay?

markusicu: Re-reading the code... This is the setGregorianChange() public API, so there is nothing else to limit the range of values.

And right after the floorDivide we re-multiply to get the fNormalizedGregorianCutover to be the start of the day corresponding to the caller's given point in time.

So what this apparently tried to do is to deal with where date-as-a-double divided by millis-per-day does not fit into a long, and compensate.

How about we change the cutoverDay into a double, and if it's below INT32_MIN or above INT32_MAX we adjust both the fGregorianCutover and the fNormalizedGregorianCutover to limit * kOneDay -- and remove the rest of the overflow detection stuff?

Maybe
```
double cutoverDay = ClockMath::floorDivide(fGregorianCutover, (double)kOneDay);
if (cutoverDay < INT32_MIN) {
    fGregorianCutover = fNormalizedGregorianCutover = (double)INT32_MIN * kOneDay;
} else if (cutoverDay > INT32_MAX) {
    fGregorianCutover = fNormalizedGregorianCutover = (double)INT32_MAX * kOneDay;
} else {
    fNormalizedGregorianCutover = cutoverDay * kOneDay;
}
```

srl295: there's also a floorDivide that returns an int64_t

markusicu: Yes, but signature matching is on the parameters, not on where the return value ends up.

Given that the inputs are both doubles, calling the double version looks proper, and we should deal with the double result.

srl295: OK. Working on a fix. Also seems we can just do `floorDivide(date, …)`

srl295:
@markusicu ptal

icu4c/source/i18n/gregocal.cpp
```diff
@@ -324,26 +324,28 @@ GregorianCalendar::setGregorianChange(UDate date, UErrorCode& status)
     if (U_FAILURE(status)) 
         return;
 
-    fGregorianCutover = date;
+    fGregorianCutover = (int64_t)date;
```
markusicu: Why? It gets pinned below if necessary.
I assume that the field is a UDate=double.

icu4c/source/i18n/gregocal.cpp
```diff
@@ -324,26 +324,28 @@ GregorianCalendar::setGregorianChange(UDate date, UErrorCode& status)
...
+    
+    if (cutoverDay <= INT32_MIN) {
+        cutoverDay = INT32_MIN;
+        fGregorianCutover = fNormalizedGregorianCutover = cutoverDay * kOneDay;
+    } else if (cutoverDay >= INT32_MAX) {
+        cutoverDay = INT32_MAX;
+        fGregorianCutover = fNormalizedGregorianCutover = cutoverDay * kOneDay;
+    } else {
+        fNormalizedGregorianCutover = cutoverDay * kOneDay;
+        fGregorianCutover = (int64_t)date;
```
markusicu: ditto why cast

icu4c/source/i18n/gregocal.cpp
srl295: fixed. (actually this line was a bad merge, line 348 is identical in the normal case)

icu4c/source/i18n/gregocal.cpp
srl295: fixed

srl295:
^ ok sorry, NOW it's fixed

markusicu:
code lgtm

not done yet:

please fix the commit message typo (missing "u")

please adjust the PR description

srl295:
corrected

markusicu:
There are test failures. From Travis Linux/gcc:
```
../../../icu4c/source/test/intltest/calregts.cpp: In member function ‘void CalendarRegressionTest::Test13745()’:
../../../icu4c/source/test/intltest/calregts.cpp:1561:94: error: call of overloaded ‘assertEquals(const char [21], long int, UDate)’ is ambiguous
     assertEquals("getGregorianChange()", INT32_MAX * MILLIS_IN_DAY, cal->getGregorianChange());
                                                                                              ^
```
I suggest you make the MILLIS_IN_DAY constant of type double. That should make the second argument a double and resolve the ambiguity.

srl295:
Ok.

Obtener Outlook para Android<https://aka.ms/ghei36>

________________________________

markusicu:
Travis is happy now, but some of the Windows build bots are still failing. Maybe an off-by-one / precision problem around limits of the double mantissa??

One of the failures is
```
         Test13745 {
         FAIL: getGregorianChange(); got -1.855425871872e+17; expected 1.855425871008e+17
         FAIL: getGregorianChange(); got 1.855425871008e+17; expected -1.855425871872e+17
      
         } ERRORS (2) in Test13745
```

srl295:
@markusicu from the build logs:
> ##[warning]icu4c\source\test\intltest\calregts.cpp(1556,0): Warning C4307: '+': integral constant overflow
> ##[warning]icu4c\source\test\intltest\calregts.cpp(1564,0): Warning C4307: '-': integral constant overflow

markusicu:
> @markusicu from the build logs:
> 
> > ##[warning]icu4c\source\test\intltest\calregts.cpp(1556,0): Warning C4307: '+': integral constant overflow
> > ##[warning]icu4c\source\test\intltest\calregts.cpp(1564,0): Warning C4307: '-': integral constant overflow

Aha, those would be from the expressions `INT32_MAX+1L` and `INT32_MIN-1L`. If you first cast the max/min constants to UDate before adding/subtracting, then it should work, right?

srl295:
@markusicu ladies and gentlemen… do we have a winner?

srl295:
Thanks all.

PR #789, opened 2019-08-29
ICU-20577 Restore C MeasureFormat ability to adopt non-decimal number…

pedberg-icu:
… formatter

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20577
- [x] Issue accepted
- [x] Tests included

This restores the ability of C++ MeasureFormat to use a non-decimal number format (as in ICU 63). I filed a separate enhancement request to add this capability to ICU4J. 

We can pretty much just re-use the ICU 63 code except for getting the unit patterns; they used to be cached in the MeasureFormat object, but now we need to get them from LongNameHandler.

### DIFF
--- a/icu4c/source/i18n/number_longnames.cpp
+++ b/icu4c/source/i18n/number_longnames.cpp
@@ -265,6 +265,26 @@ UnicodeString LongNameHandler::getUnitDisplayName(
     return simpleFormats[DNAM_INDEX];
 }
 
+UnicodeString LongNameHandler::getUnitPattern(
+        const Locale& loc,
+        const MeasureUnit& unit,
+        UNumberUnitWidth width,
+        StandardPlural::Form pluralForm,
+        UErrorCode& status) {
+    if (U_FAILURE(status)) {
+        return ICU_Utility::makeBogusString();
+    }
+    UnicodeString simpleFormats[ARRAY_LENGTH];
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/number_longnames.cpp
```diff
@@ -265,6 +265,20 @@ UnicodeString LongNameHandler::getUnitDisplayName(
...
+        const MeasureUnit& unit,
+        UNumberUnitWidth width,
+        StandardPlural::Form pluralForm,
+        UErrorCode& status) {
+    if (U_FAILURE(status)) {
+        return ICU_Utility::makeBogusString();
+    }
+    UnicodeString simpleFormats[ARRAY_LENGTH];
+    getMeasureData(loc, unit, width, simpleFormats, status);
+    return simpleFormats[pluralForm];
```
markusicu: This function may get called twice, and each time you fetch strings for all forms an return one of them. How about instead taking two UnicodeString's by reference and assigning to them the pluralForm and the OTHER form (e.g., pattern / otherPattern)? The call site would then not need to call this again for OTHER. (Return a boolean = U_SUCCESS)

pedberg-icu:
> lgtm but what do you think of my optimization idea?

I think that is a great suggestion, thanks, I will implement it now.

sffc:
LGTM

PR #790, opened 2019-08-29
ICU-20799 delete calendarToAdopt upon error

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20799
- [X] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/test/intltest/dtfmttst.cpp
+++ b/icu4c/source/test/intltest/dtfmttst.cpp
@@ -127,6 +127,7 @@ void DateFormatTest::runIndexedTest( int32_t index, UBool exec, const char* &nam
     TESTCASE_AUTO(TestMinuteSecondFieldsInOddPlaces);
     TESTCASE_AUTO(TestDayPeriodParsing);
     TESTCASE_AUTO(TestParseRegression13744);
+    TESTCASE_AUTO(TestAdoptCalendarLeak);
 
     TESTCASE_AUTO_END;
 }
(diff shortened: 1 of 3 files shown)


icu4c/source/test/intltest/dtfmttst.cpp
```diff
@@ -5540,6 +5541,17 @@ void DateFormatTest::TestParseRegression13744() {
     assertEquals("Error index", inDate.length(), pos.getErrorIndex());
 }
 
+void DateFormatTest::TestAdoptCalendarLeak() {
```
jefgen: Thank for adding a test Frank.
I wonder if it might be worth adding some comments like:
This test relies on the locale fullName exceeding ULOC_FULLNAME_CAPACITY in order for setKeywordValue to fail.

ULOC_FULLNAME_CAPACITY

jefgen:
Looks like the test case is crashing on some of the build bots (Clang/Linux Ubuntu 16.04 and MSVC 64-bit release).

icu4c/source/test/intltest/dtfmttst.cpp
```diff
@@ -5540,6 +5541,17 @@ void DateFormatTest::TestParseRegression13744() {
...
+    UErrorCode status;
```
jefgen: This needs to be initialized in order to prevent the test from randomly crashing:
`UErrorCode status = U_ZERO_ERROR;`

PR #793, opened 2019-08-30
ICU-20795 Add comment to ICU4C Readme.html about U_SHOW_CPLUSPLUS_API.

jefgen:
Probably worth adding a note to the README.html file that all of the public C++ headers are now guarded with the macro "`U_SHOW_CPLUSPLUS_API`".

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20795
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/readme.html
+++ b/icu4c/readme.html
@@ -744,7 +744,8 @@
         across library and compiler versions is very hard to achieve.
         Most ICU C++ APIs are in header files that contain a comment with
         <code>\brief C++ API</code>.
-        Consider not installing these header files.</li>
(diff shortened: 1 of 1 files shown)


daniel-ju:
Thanks for adding this! Would it be easier for you to merge this change first, and I can rebase in #792  after?

jefgen:
I think it would be better if the version number PR is merged first, and then this change can be rebased later if there is any conflict. (Though I'm hoping that since the same lines aren't modified there shouldn't be any conflicts, but I guess we'll see...)

PR #804, opened 2019-08-31
ICU-20797 fix UBS compilation error and UBS in test code

srl295:
[ICU-20797]

Two issues here:

- fix 2 build issue in i18n when compiling with clang++ -fsanitize=undefined
the following two symbols were not exported (and they should be):
  typeinfo for icu::CollationCacheEntry
  typeinfo for icu::numparse::impl::CodePointMatcher

- remove undefined behavior warning in NumberFormatTestTuple.. minor, but very annoying
when repeated many times during every test run. Tends to mask real errors.

> numberformattesttuple.cpp:319:5: runtime error: member access within null pointer of type 'NumberFormatTestTuple'

- [ ] Issue accepted (not yet)

[ICU-20797]: https://unicode-org.atlassian.net/browse/ICU-20797

### DIFF
--- a/icu4c/source/i18n/collationtailoring.h
+++ b/icu4c/source/i18n/collationtailoring.h
@@ -92,7 +92,7 @@ private:
     CollationTailoring(const CollationTailoring &other);
 };
 
-struct CollationCacheEntry : public SharedObject {
+struct U_I18N_API CollationCacheEntry : public SharedObject {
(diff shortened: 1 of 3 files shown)


srl295:
i guess this could go under accepted [ICU-13655] too.

srl295:
No rush on this one. Only test code.

markusicu:
ping

PR #806, opened 2019-08-31
ICU-20693 Fix weird bug in Maven pom file and add comments

hagbard:
I have no idea why, but my Maven setup had used version 65.1 of the ICU4J dependency in the pom file. Perhaps it was configured locally (e.g. my setup or in Google) for some reason but it stopped working abruptly for me tonight (Maven would not build and could not resolve the dependency). I set it back to 64.2 and added some comments, and everything appears to work again. Very odd.

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/cldr-to-icu/pom.xml
+++ b/tools/cldr/cldr-to-icu/pom.xml
@@ -79,6 +79,7 @@
     </repositories>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


markusicu:
65 does not exist yet. If Maven is trying to pull ICU 65 from a public repository, it has to fail.

PR #204, opened 2018-10-10
ICU-20162: use maven for C api change tool

srl295:
reopen from #163 

- delete a bunch of .jars from the repository
- for now, i'm keeping the exact versions we used to use
- upgrading xalan is as easy as modifying the pom.xml
- ignore Makefile.local in copyright scan

##### Checklist

- [x] Issue filed at [ICU-20162]
- [x] Update PR title to include Issue number
- [x] Issue accepted
- [x] Documentation is changed or added

[ICU-20162]: https://unicode-org.atlassian.net/browse/ICU-20162

PR #807, opened 2019-09-03
ICU-20162 use maven to build icu4c api change tool

srl295:
[ICU-20162]

- handle covariant return types by “simplifying” some functions’ return type to `void*`
- removes three .jar files from git-lfs
- make calls ant calls maven
- have travis verify buildability
- support APIChangeReport.md (markdown)
- 'signature simplification' is no longer new, do not mark it as new.

[ICU-20162]: https://unicode-org.atlassian.net/browse/ICU-20162


srl295:
i added a 'release tools' instance, this could also make sure that @hagbard ’s tool builds also.

srl295:
for #808

roubert:
Sorry, I won't be able to review this, I hardly know anything at all about Maven.

.travis.yml
```diff
@@ -134,3 +134,11 @@ matrix:
...
       script:
           - perl tools/scripts/cpysearch/cpyscan.pl
+
+    - name: "j: icu4c release tools"
+      language: "java"
+      script:
+        - mvn -f tools/release/java/pom.xml package dependency:analyze
+      cache:
+        directories:
+        - $HOME/.m2
```
hagbard: This is only for the server side continuous build right? Otherwise assuming $HOME/.m2 for the Maven package cache might be an issue.

tools/release/java/build.xml
```diff
@@ -62,29 +58,56 @@
...
-                <pathelement path="."/>
-            </classpath>
+        </java>
+    </target>
+
+    <target name="apireport_md" depends="tools">
+        <java jar="${apireport.jar}" fork="yes" failonerror="true">
+
+<!--
+            <arg value = "-DELETEME-newver"/>
```
hagbard: could you use properties here?

Please add some user level comments describing when/why someone would set these.

tools/release/java/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+      <version>2.7.2</version>
+      <scope>runtime</scope>
+    </dependency>
+
+
+
+  </dependencies>
+
+  <build>
+    <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
```
hagbard: Perhaps say why? This kind of comment can become confusing if people continue to follow it after the reason for it has gone. Is there a technical reason? Is it a policy thing? At what point could Maven defaults be trusted (if ever) ?

tools/release/java/pom.xml
```diff
@@ -0,0 +1,132 @@
...
+        <plugin>
+          <artifactId>maven-project-info-reports-plugin</artifactId>
+          <version>3.0.0</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
```
hagbard: Do you need to "shade" it if it's only being run as a main tool?
I thought shading was for re-writing library packages/classes so they don't clash with things when your library is used by someone else. That shouldn't matter for top level binaries (and either way, please comment why this is being done).

tools/release/java/build.xml
```diff
@@ -14,35 +14,31 @@
...
 
-    <target name="tools"  depends="init" description="compile release tools">
-        <javac includeantruntime="false" includes="**/*.java"
-            srcdir="${src.dir}"
-            destdir="${build.dir}"
-            source="1.7"
-            debug="on" deprecation="off"
-            encoding="utf8"/>
+    <target name="tools"  depends="init" description="compile release tools" unless="apireport.jar.exists">
+      <exec command="mvn package"/> <!-- complains, but it's cross platform -->
```
hagbard: I use:

        <exec executable="mvn" searchpath="true">
            <arg value="compile"/>
        </exec>

Which might be neater (the binary is clearly delimited from the arguments).

tools/release/java/build.xml
```diff
@@ -62,29 +58,56 @@
...
+            <arg value = "--resultfile"/>
+            <arg value = "${basedir}/APIChangeReport.md"/>
+
         </java>
     </target>
     <target name="apireport_xml" depends="tools">
-        <java classname="com.ibm.icu.dev.tools.docs.StableAPI" fork="yes" failonerror="true">
+        <java jar="${apireport.jar}" fork="yes" failonerror="true">
 
 <!--
```
hagbard: I'm definitely not a fan of these uncommented "magic" comments. It feels like they'll just rot over time.
Could we have a flag that's always set but takes a value (e.g. --newver=${want.new.version}") and then the property is overrideable normally on the command line instead of having to edit the build file.

.travis.yml
srl295: @hagbard yes only affects travis speedup

tools/release/java/build.xml
srl295: this is preexisting. it overwrites the version number rather than have it calculated (if the TC wants to change the title to a specific version number) I can document though.

tools/release/java/build.xml
srl295: > . It feels like they'll just rot over time.

they already have rotted. It's probably better just to remove them entirely from ant and document that they can be set manually

tools/release/java/pom.xml
srl295: this is from the archetype generated pom.xml

tools/release/java/pom.xml
srl295: yes, but it includes the dependencies inline in the .jar

tools/release/java/build.xml
srl295: this is a great suggestion.

tools/release/java/build.xml
srl295: I'm just dropping the comments.

tools/release/java/build.xml
srl295: dropping them

tools/release/java/pom.xml
srl295: I'm just going to remove the comments. They are standard practice for maven, so I don't think there's a need to comment.

markusicu:
I don't know my way around this stuff, but I trust you, and it's tools.
Please first update the PR description to match the final state.
Also, is the Simplification anchor up to date with the output in the other PR?

tools/release/java/pom.xml
srl295: commented why it's being done

srl295:
> Please first update the PR description to match the final state.

done

> Also, is the Simplification anchor up to date with the output in the other PR?

yes

tools/release/java/src/main/java/com/ibm/icu/dev/tools/docs/StableAPI.java
```diff
@@ -218,7 +218,8 @@ private InputStream loadStream(String name, String argName, File argFile) {
     	InputStream stream = null;
     	if(argFile != null) {
     		try {
-	    		stream = new FileInputStream(argFile);
+                stream = new FileInputStream(argFile);
```
markusicu: funky indentation here & around... but i will approve anyway in the interest of the rc

tools/release/java/src/main/resources/com/ibm/icu/dev/tools/docs/genReport_md.xslt
```diff
@@ -0,0 +1,237 @@
...
+  <xsl:param name="rightVer" />
+  <xsl:param name="dateTime" />
+  <xsl:param name="rightMilestone" />
+  <xsl:param name="leftMilestone" />
+  <xsl:param name="notFound" />
+  <xsl:param name="ourYear" />
+  <xsl:output method="text" indent="no" omit-xml-declaration="yes" />
+  <xsl:template match="/">
+<![CDATA[<!--]]>
+ Copyright © <xsl:value-of select="$ourYear" />: Unicode, Inc. and others.
```
markusicu: Should we fix this to 2019 and later? -- like we generally use the first year we added a file.

tools/release/java/src/main/resources/com/ibm/icu/dev/tools/docs/genReport_md.xslt
```diff
@@ -0,0 +1,237 @@
...
+<xsl:for-each select="/list/simplifications/simplification">
+- **`<xsl:value-of select="base" />`**<xsl:for-each select="change">
+  - `<xsl:value-of select="text()" />`</xsl:for-each>
+</xsl:for-each>
+
+## Colophon
+
+Contents generated by StableAPI tool on <xsl:value-of select="$dateTime" />
+
+Copyright (C) 2016-<xsl:value-of select="$ourYear" />, Unicode, Inc. and others.
```
markusicu: Should this be just 2019 and later like in a new source file?

tools/release/java/src/main/resources/com/ibm/icu/dev/tools/docs/genReport.xslt
```diff
@@ -30,7 +30,7 @@ doctype-system="http://www.w3.org/TR/html4/loose.dtd"
   <xsl:template match="/">
     <html>
 	<xsl:comment>
-     Copyright (C) 2016 and later: Unicode, Inc. and others.
+     Copyright © <xsl:value-of select="$ourYear" />: Unicode, Inc. and others.
```
markusicu: I think we should keep it 2016 and later for the HTML version.

tools/release/java/src/main/resources/com/ibm/icu/dev/tools/docs/genreport_xml.xslt
```diff
@@ -29,7 +29,7 @@
 
   <xsl:template match="/">
     <xsl:comment>
-      Copyright (C)  <xsl:value-of select="$ourYear" />: Unicode, Inc. and others.
+      Copyright © <xsl:value-of select="$ourYear" />: Unicode, Inc. and others.
```
markusicu: 2016 and later?

tools/release/java/src/main/java/com/ibm/icu/dev/tools/docs/StableAPI.java
srl295: Will re-indent in a new PR

srl295:
formerly #204

tools/release/java/src/main/resources/com/ibm/icu/dev/tools/docs/genreport_xml.xslt
srl295: To fix in #933

tools/release/java/src/main/java/com/ibm/icu/dev/tools/docs/StableAPI.java
srl295: reindented in #933

PR #815, opened 2019-09-05
ICU-20835 Update instructions for commit checker.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20795
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/commit-checker/README.md
+++ b/tools/commit-checker/README.md
@@ -24,7 +24,9 @@ Optional: save your Jira credentials in a `.env` file in this directory:
     JIRA_USERNAME=hello
     JIRA_PASSWORD=world
 
-This is required if you want to process sensitive tickets.
+This is required if you want to process sensitive tickets.  Note: JIRA_PASSWORD needs to be an API Token generated according to the following instructions:
+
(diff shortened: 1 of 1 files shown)


tools/commit-checker/README.md
```diff
@@ -39,4 +41,23 @@ Run the tool and save the result into REPORT.md; set fixVersion to the upcoming
         --rev-range "release-63-1..upstream/maint/maint-64"
         > REPORT.md
```
jefgen: I wonder if it might be worth adding a line or two saying something like:
Note: If the maintenance branch has not yet been created then use the master branch in the revision.
Ex: "release-64-2..upstream/master"
(Also this assumes that the unicode-org repo is named upstream).

sffc: Done

sffc:
I changed the ticket number to use the same ticket as #852 instead of re-using the ICu 65 ticket.  This will avoid putting a ticket on the ICU 65 ticket on the master branch.

sffc:
Ping on this and #852

PR #824, opened 2019-09-10
ICU-20453 Add more warning types to CI build.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20453
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -44,7 +44,7 @@ jobs:
     vmImage: 'Ubuntu 16.04'
   steps:
     - script: |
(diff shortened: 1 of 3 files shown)


sffc:
I think there are still some warnings that might not be covered by the CI build.  At least, they don't seem to be included in -Wall.  Can you verify?

These are both warning categories that caused problems with the V8 integration last cycle.

sffc:
According to this list, the two warnings I'm adding are not included in either Wall or Westra:

https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

sffc:
Also: I propose this PR for the maintenance branch.

jefgen:
It doesn't look like Clang's setting for `-Wall` enables `-Wundef` or `-Wnon-virtual-dtor` either.

markusicu:
ICU meeting says ok to merge into maint-65 branch.

PR #827, opened 2019-09-11
ICU-20796 BRS 65 update C API guard macros (manual version)

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20796
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/unicode/localebuilder.h
+++ b/icu4c/source/common/unicode/localebuilder.h
@@ -279,6 +279,7 @@ public:
      */
     Locale build(UErrorCode& status);
 
+#ifndef U_HIDE_DRAFT_API
     /**
      * Sets the UErrorCode if an error occurred while recording sets.
      * Preserves older error codes in the outErrorCode.
@@ -289,6 +290,7 @@ public:
      * @draft ICU 65
      */
     UBool copyErrorTo(UErrorCode &outErrorCode) const;
+#endif  /* U_HIDE_DRAFT_API */
 
 private:
     friend class LocaleMatcher::Result;
--- a/icu4c/source/common/unicode/uniset.h
+++ b/icu4c/source/common/unicode/uniset.h
@@ -1655,11 +1655,10 @@ private:
                      const UnicodeSet* inclusions,
                      UErrorCode &status);
 
-#ifndef U_HIDE_DRAFT_API   // Skipped: ucpmap.h is draft only.
+    // UCPMap is now stable ICU 63
     void applyIntPropertyValue(const UCPMap *map,
                                UCPMapValueFilter *filter, const void *context,
                                UErrorCode &errorCode);
-#endif  /* U_HIDE_DRAFT_API */
 
     /**
      * Set the new pattern to cache.
--- a/icu4c/source/i18n/unicode/calendar.h
+++ b/icu4c/source/i18n/unicode/calendar.h
@@ -2283,7 +2283,6 @@ private:
      * should only be called if this calendar is not lenient.
      * @see #isLenient
      * @see #validateField(int, int&)
-     * @internal
      */
     void validateFields(UErrorCode &status);
 
@@ -2293,7 +2292,6 @@ private:
      * <code>U_ILLEGAL_ARGUMENT_ERROR</code> will be set.  Subclasses may
      * use this method in their implementation of {@link
      * #validateField(int, int&)}.
-     * @internal
      */
     void validateField(UCalendarDateFields field, int32_t min, int32_t max, UErrorCode& status);
 
--- a/icu4c/source/i18n/unicode/dcfmtsym.h
+++ b/icu4c/source/i18n/unicode/dcfmtsym.h
@@ -457,13 +457,11 @@ private:
      * to non-resource bundle strings,
      * then regular UnicodeString copies must be used instead of fastCopyFrom().
      *
-     * @internal
      */
     UnicodeString fSymbols[kFormatSymbolCount];
 
     /**
      * Non-symbol variable for getConstSymbol(). Always empty.
-     * @internal
      */
     UnicodeString fNoSymbol;
 
--- a/icu4c/source/i18n/unicode/formattedvalue.h
+++ b/icu4c/source/i18n/unicode/formattedvalue.h
@@ -26,7 +26,9 @@ U_NAMESPACE_BEGIN
  */
 
 
-#ifndef U_HIDE_DRAFT_API
+// The following cannot have #ifndef U_HIDE_DRAFT_API because
+// class FormattedValue depends on it, and FormattedValue cannot be
+// hidden becauseclass FormattedNumber (stable ICU 60) depends on it.
 /**
  * Represents a span of a string containing a given field.
  *
@@ -55,6 +57,7 @@ class U_I18N_API ConstrainedFieldPosition : public UMemory {
     /** @draft ICU 64 */
     ~ConstrainedFieldPosition();
 
+#ifndef U_HIDE_DRAFT_API
     /**
      * Resets this ConstrainedFieldPosition to its initial state, as if it were newly created:
      *
@@ -224,18 +227,23 @@ class U_I18N_API ConstrainedFieldPosition : public UMemory {
         int32_t field,
         int32_t start,
         int32_t limit);
+#endif  /* U_HIDE_DRAFT_API */
 
   private:
     int64_t fContext = 0LL;
     int32_t fField = 0;
     int32_t fStart = 0;
     int32_t fLimit = 0;
+#ifndef U_HIDE_DRAFT_API
     int32_t fCategory = UFIELD_CATEGORY_UNDEFINED;
+#else   /* U_HIDE_DRAFT_API */
+    int32_t fCategory = 0;
+#endif  /* U_HIDE_DRAFT_API */
     int8_t fConstraint = 0;
 };
 
-#endif // U_HIDE_DRAFT_API
-
+// The following cannot have #ifndef U_HIDE_DRAFT_API because
+// class FormattedNumber (stable ICU 60) depends on it.
 /**
  * An abstract formatted value: a string with associated field attributes.
  * Many formatters format to classes implementing FormattedValue.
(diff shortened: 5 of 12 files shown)


markusicu:
On Travis, the header test is failing due to problems with ConstrainedFieldPosition. Please fix.

icu4c/source/common/unicode/localebuilder.h
```diff
@@ -279,6 +279,7 @@ class U_COMMON_API LocaleBuilder : public UObject {
      */
     Locale build(UErrorCode& status);
 
+#ifndef U_HIDE_DRAFT_API
```
markusicu: FYI Not strictly necessary. I understand this helps with drafts from different versions. My tool should find problems when we promote draft 64 but not 65 and forget this here.

icu4c/source/common/unicode/uniset.h
```diff
@@ -1655,11 +1655,10 @@ class U_COMMON_API UnicodeSet U_FINAL : public UnicodeFilter {
                      const UnicodeSet* inclusions,
                      UErrorCode &status);
 
-#ifndef U_HIDE_DRAFT_API   // Skipped: ucpmap.h is draft only.
+    // UCPMap is now stable ICU 63
```
markusicu: FYI -- Nice catch. My tool won't find these because there is no tag here.

icu4c/source/i18n/unicode/calendar.h
```diff
@@ -2283,7 +2283,7 @@ class U_I18N_API Calendar : public UObject {
      * should only be called if this calendar is not lenient.
      * @see #isLenient
      * @see #validateField(int, int&)
-     * @internal
+     * @internal (private)
```
markusicu: May be better to just remove the tag?

icu4c/source/i18n/unicode/calendar.h
```diff
@@ -2293,7 +2293,7 @@ class U_I18N_API Calendar : public UObject {
      * <code>U_ILLEGAL_ARGUMENT_ERROR</code> will be set.  Subclasses may
      * use this method in their implementation of {@link
      * #validateField(int, int&)}.
-     * @internal
+     * @internal (private)
```
markusicu: ditto

icu4c/source/i18n/unicode/dcfmtsym.h
```diff
@@ -457,13 +457,13 @@ class U_I18N_API DecimalFormatSymbols : public UObject {
      * to non-resource bundle strings,
      * then regular UnicodeString copies must be used instead of fastCopyFrom().
      *
-     * @internal
+     * @internal (private)
```
markusicu: ditto

icu4c/source/i18n/unicode/dcfmtsym.h
```diff
@@ -457,13 +457,13 @@ class U_I18N_API DecimalFormatSymbols : public UObject {
...
      */
     UnicodeString fSymbols[kFormatSymbolCount];
 
     /**
      * Non-symbol variable for getConstSymbol(). Always empty.
-     * @internal
+     * @internal (private)
```
markusicu: ditto

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -674,7 +674,7 @@ class U_I18N_API Precision : public UMemory {
     } fType;
 
     union PrecisionUnion {
-        /** @internal */
+        /** @internal (private) */
```
markusicu: May be better to just remove the tag?

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -686,7 +686,7 @@ class U_I18N_API Precision : public UMemory {
             /** @internal */
             impl::digits_t fMaxSig;
         } fracSig;
-        /** @internal */
+        /** @internal (private) */
```
markusicu: ditto

icu4c/source/i18n/unicode/numberrangeformatter.h
```diff
@@ -855,7 +859,7 @@ class U_I18N_API FormattedNumberRange : public UMemory, public FormattedValue {
 
     /**
      * Internal constructor from data type. Adopts the data pointer.
-     * @internal
+     * @internal (private)
```
markusicu: May be better to just remove the tag?

icu4c/source/i18n/unicode/plurfmt.h
```diff
@@ -524,7 +524,7 @@ class U_I18N_API PluralFormat : public Format {
 
 private:
      /**
-      * @internal
+      * @internal (private)
```
markusicu: May be better to just remove the tag?

icu4c/source/i18n/unicode/plurfmt.h
```diff
@@ -536,13 +536,13 @@ class U_I18N_API PluralFormat : public Format {
          * @param number The number to be plural-formatted.
          * @param ec Error code.
          * @return The selected PluralFormat keyword.
-         * @internal
+         * @internal (private)
```
markusicu: ditto -- and more in this file

icu4c/source/i18n/unicode/formattedvalue.h
```diff
@@ -288,7 +290,6 @@ class U_I18N_API FormattedValue /* not : public UObject because this is an inter
      */
     virtual Appendable& appendTo(Appendable& appendable, UErrorCode& status) const = 0;
 
-#ifndef U_HIDE_DRAFT_API
```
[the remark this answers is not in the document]
pedberg-icu: > This was an intentional part of Steven's promotion PR. toString, toTempString, and appendTo were all on FormattedNumber from before, and only nextPosition was added new. So, we left nextPosition hidden both in the base class (here) and in the child class. Also note that un-hiding this may require un-hiding ConstrainedFieldPosition.

We cannot have #ifndef macros around virtual functions. That would mean clients would get different vtable layouts depending on which conditionals they build with, and lead to binary incompatibility. This is specifically called out in the [instructions for guard conditionals](https://sites.google.com/site/icusite/processes/release/tasks/docs#TOC-Verify-that-draft-is-surrounded-by-ifndef-U_HIDE_DRAFT_API-etc) and in the [C coding guidelines](http://userguide.icu-project.org/dev/codingguidelines#TOC-C-C-Hiding-Un--stable-APIs).

I guess I will un-hide ConstrainedFieldPosition.

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -2410,14 +2410,14 @@ class U_I18N_API LocalizedNumberFormatter
 class U_I18N_API FormattedNumber : public UMemory, public FormattedValue {
   public:
 
-#ifndef U_HIDE_DRAFT_API
+    // Default constructor cannot have #ifndef U_HIDE_DRAFT_API
```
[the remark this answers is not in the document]
pedberg-icu: Because C++ will auto-generate one if it is conditionalized out. This is specifically called out in the [C coding guidelines](http://userguide.icu-project.org/dev/codingguidelines#TOC-C-C-Hiding-Un--stable-APIs).

icu4c/source/i18n/unicode/calendar.h
pedberg-icu: Done

icu4c/source/i18n/unicode/dcfmtsym.h
pedberg-icu: Done

icu4c/source/i18n/unicode/numberformatter.h
pedberg-icu: Done

icu4c/source/i18n/unicode/numberformatter.h
pedberg-icu: Done

icu4c/source/i18n/unicode/numberrangeformatter.h
pedberg-icu: Done

icu4c/source/i18n/unicode/plurfmt.h
pedberg-icu: Done

icu4c/source/i18n/unicode/formattedvalue.h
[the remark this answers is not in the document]
sffc: I wonder if that recommendation is still necessary. We had a similar issue when we put `std::atomic` as a class member in a public header. Despite the fact that the layout of `std::atomic` is implementation-dependent, we decided it was OK because we only guarantee binary compatibility in C APIs (C++ users build and ship ICU in tandem with the rest of their application).

But, it's not really a problem if you change it to expose nextPosition. As a way to avoid exposing ConstrainedFieldPosition, you could forward declare it since it's only used by reference in this function.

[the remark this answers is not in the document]
pedberg-icu: It IS a problem because the library builds without anything hidden, but clients might build with say U_HIDE_DRAFT_API defined. It is not a matter of different binary versions of the library, but different builds of code that uses the SAME library version seeing different vtable layouts.

It is not necessary to really go out of our way to use the conditionals to hide everything that is DRAFT, say. The important thing is the API status tag. Then we conditionalize as much as we can easily do without causing any problems.

icu4c/source/i18n/unicode/numberrangeformatter.h
```diff
@@ -725,7 +727,9 @@ class U_I18N_API FormattedNumberRange : public UMemory, public FormattedValue {
     // Copydoc: this method is new in ICU 64
     /** @copydoc FormattedValue::nextPosition() */
     UBool nextPosition(ConstrainedFieldPosition& cfpos, UErrorCode& status) const U_OVERRIDE;
+#endif  /* U_HIDE_DRAFT_API */
```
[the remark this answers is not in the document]
pedberg-icu: Yep, that is why we have the Travis conditionals test. Anyway, I did unhide the concrete override here and in a few other places. Also, though I had removed the guard conditionals around the ConstrainedFieldPosition class as a whole (since it is used by unguarded items), I added them to its methods. And oddly, I had to restore some @internal status tags to the private fields in the Precision class to avoid doxygen complaints (not sure why it cares about private items). Anyway, all tests are passing now.

PR #828, opened 2019-09-12
ICU-20814 new tool for checking API tags vs. ifdef guards; add FORCE …

markusicu:
…guards, and move some guards, to make this work

https://unicode-org.atlassian.net/browse/ICU-20814

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,11 @@ matrix:
         - PATH="${PREFIX}/bin:$PATH" make -C test/hdrtst check
 
     # clang release build with some options to enforce useful constraints.
+    # Includes dependency checker on an in-source, optimized build.
+    # Includes checking @draft etc. API tags vs. ifndef guards like
+    # U_HIDE_DRAFT_API and U_FORCE_HIDE_DRAFT_API.
+    # (FORCE guards make this tool pass but won't compile to working code.
+    # See the testtagsguards.sh script for details.)
     - name: "c: linux clang"
       dist:     xenial
       language: cpp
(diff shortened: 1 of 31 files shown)


markusicu:
I tried to make it work for @internal as well but ran into trouble after hiding low-level macros like for the platform, charset and endianness. I punted on this for now. We may need a different tag for some of those.

markusicu:
rebased, no longer need to add normal guards
PTAL

sffc:
How about if you add this new test to the .travis.yml file?  You'll want to do that eventually and it makes sense to do it in the same PR.

markusicu:
> Changes LGTM, but please add this to .travis.yml

I have just made an attempt at doing so, see the new .travis.yml changes. Let’s see if it works.

sffc:
It looks like the travis CI didn't get the change to .travis.yml.  It is testing commit ecbf19abdc8dd6f376dd72772a97ab7d71a5836c which doesn't have that change in it.

I tried merging the branch manually in my local workspace, and it succeeds cleanly, including the change to .travis.yml.  I have no other explanation than that .travis.yml made a mistake and still tested the old commit, maybe with a race condition of some kind.

I tried pushing the "restart build" button on the CI, but that just restarts the build based on the same hash; in our case, we need the hash to be recomputed.

Try force-pushing again (run `git commit --amend` to update the timestamp of the commit) and see if the CI works this time.

markusicu:
> Try force-pushing again (run `git commit --amend` to update the timestamp of the commit) and see if the CI works this time.

done

markusicu:
started from the wrong location; trying again

markusicu:
Finally works with `diff -u --minimal`. PTAL

PR #832, opened 2019-09-18
ICU-20819 Correct placement of parentheses in preprocessor macro.

roubert:
The purpose of the FIELD_NAME_STR() macro is to create a string literal
by using the # preprocessing operator and then skip the first 5 chars of
this string by using the +5 pointer arithmetic. This is all intentional,
but if the parentheses are misplaced the compiler might think that this
is a mistake, a failed string concatenation (-Wstring-plus-int).


roubert:
Ping?

PR #836, opened 2019-09-20
ICU-20478 Sort variant in (for|to)LanguageTag of icu::Locale and ULocale

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20478
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -1110,6 +1110,19 @@ _appendRegionToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool stri
     }
 }
 
+static void _sortVariants(VariantListEntry* first) {
+    for (VariantListEntry* var1 = first; var1 != NULL; var1 = var1->next) {
+        for (VariantListEntry* var2 = var1->next; var2 != NULL; var2 = var2->next) {
+            // Swap var1->variant and var2->variant.
+            if (uprv_compareInvCharsAsAscii(var1->variant, var2->variant) > 0) {
+                const char* temp = var1->variant;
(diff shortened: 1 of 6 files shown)


icu4c/source/common/uloc_tag.cpp
```diff
@@ -1110,6 +1110,19 @@ _appendRegionToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool stri
     }
 }
 
+static void _sortVariants(VariantListEntry* first) {
+    for (VariantListEntry* var1 = first; var1 != NULL; var1 = var1->next) {
+        for (VariantListEntry* var2 = var1->next; var2 != NULL; var2 = var2->next) {
+            // Swap var1->variant and var2->variant.
+            if (strcmp(var1->variant, var2->variant) > 0) {
```
markusicu: In the ICU4C library code, we want to use uprv_str...() functions not raw str...().

In this case, we would also get inconsistent sort orders for EBCDIC vs. ASCII, which would make your test cases fail on an EBCDIC machine, so actually the right one to use is uprv_compareInvCharsAsAscii(s1, s2) from internal uinvchar.h. (EBCDIC sorts digits after letters.)

(This is not a concern in Java which uses Unicode strings.)

FrankYFTang: PTAL adddressed!

markusicu:
FYI remember that if you following the "Details" link on the jira-ticket checker you can squash from the UI there

FrankYFTang:
please approve after the force-push. Thanks

PR #839, opened 2019-09-24
ICU-20831 Eliminate a data race in SimpleDateFormat::format.

eggrobin:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20831
- [ ] Issue accepted
- [ ] Tests included

##### A note regarding testing.
The issue at hand is a fairly rare data race; the only way I found to reliably reproduce it in a unit test was to add a 20 millisecond sleep in [`rbbi_cache.cpp`](https://github.com/unicode-org/icu/blob/c78d9fa13762b7ae1ee9aec623d0b818d02a9956/icu4c/source/common/rbbi_cache.cpp#L434-L435), and to then simultaneously call `format` with different values from many threads on a `SimpleDateFormat` created as follows.
```C++
  icu::ErrorCode error;
  // NOTE: The choice of pattern and locale here is important: the resulting
  // date must start with a lowercase letter in its default capitalization, so
  // that we actually go through the capitalization code.
  icu::Locale french("fr_FR");
  icu::SimpleDateFormat simple_date_format(icu::UnicodeString::fromUTF8("cccc"),
                                           french,
                                           error);
  simple_date_format.setContext(UDISPCTX_CAPITALIZATION_FOR_STANDALONE, error);
```
The calls to `format` will then produce such results as `{ "DimanChe", "DimancChe", "DimancHe", "Dimanche", "Jeudi", "Lundi", "Mardi", "MercrEdi", "MercreDi", "Mercredi", "SamedI", "Samedi", "VendrEdi", "VendreDi", "VendreEdi", "Vendredi" }`, with the oddly-capitalized (and sometimes oddly-spelled) values exhibiting the data race.

Unfortunately, since the test is only effective when adding sleep statements in ICU internals, I cannot really include it in this pull request.

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -1981,9 +1981,11 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
                 break;
         }
         if (titlecase) {
+            BreakIterator* const mutableCapitalizationBrkIter = fCapitalizationBrkIter->clone();
             UnicodeString firstField(appendTo, beginOffset);
-            firstField.toTitle(fCapitalizationBrkIter, fLocale, U_TITLECASE_NO_LOWERCASE | U_TITLECASE_NO_BREAK_ADJUSTMENT);
+            firstField.toTitle(mutableCapitalizationBrkIter, fLocale, U_TITLECASE_NO_LOWERCASE | U_TITLECASE_NO_BREAK_ADJUSTMENT);
             appendTo.replaceBetween(beginOffset, appendTo.length(), firstField);
+            delete mutableCapitalizationBrkIter;
         }
     }
 #endif
--- a/icu4c/source/i18n/unicode/smpdtfmt.h
+++ b/icu4c/source/i18n/unicode/smpdtfmt.h
@@ -1645,7 +1645,7 @@ private:
 
     UBool fHaveDefaultCentury;
 
-    BreakIterator* fCapitalizationBrkIter;
+    const BreakIterator* fCapitalizationBrkIter;
 };
 
 inline UDate


sffc:
Can you check for this issue and fix it in Java if necessary?

pedberg-icu:
> Please check and fix in ICU4J if necessary.



> Please check and fix in ICU4J if necessary.

Wll do ICU4J in a separate PR

sffc:
> Wll do ICU4J in a separate PR

Same ticket, for ICU 65, or a new ticket for future?  If this is in a separate near-term PR, I would prefer that they go into the same release.

pedberg-icu:
> > Wll do ICU4J in a separate PR
> 
> Same ticket, for ICU 65, or a new ticket for future? If this is in a separate near-term PR, I would prefer that they go into the same release.

Same *release* or same *PR*? The plan was to have two different PRs for ICU 65 (one for C, one for J, since the latter would be done by me instead of the person created this PR)

aheninger:
An alternative fix would be to mutex the existing use of fCapitalizationBrkIter, rather than cloning the break iterator.

Which is better is not obvious, it depends on the usage. For single threaded cases, or multi-threaded but without heavy contention, mutexing is going to be faster than cloning.

sffc:
> Same release or same PR? The plan was to have two different PRs for ICU 65 (one for C, one for J, since the latter would be done by me instead of the person created this PR)

OK.  Two PRs for ICU 65 SGTM.  So we will use the same ticket for the second PR.

> An alternative fix would be to mutex the existing use of fCapitalizationBrkIter, rather than cloning the break iterator.
> 
> Which is better is not obvious, it depends on the usage. For single threaded cases, or multi-threaded but without heavy contention, mutexing is going to be faster than cloning.

eggrobin  discussed this option on the ticket and said:

> exclusively locking on all calls to format may cause contention issues

PR #840, opened 2019-09-25
ICU-20832 use uint32_t instead of uint16_t to avoid overflows for very long strings

adamsitnik:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20832
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/usearch.cpp
+++ b/icu4c/source/i18n/usearch.cpp
@@ -317,7 +317,7 @@ inline uint16_t initializePatternCETable(UStringSearch *strsrch,
         uprv_free(pattern->ces);
     }
 
-    uint16_t  offset      = 0;
+    uint32_t  offset      = 0;
     uint16_t  result      = 0;
     int32_t   ce;
 
@@ -388,7 +388,7 @@ inline uint16_t initializePatternPCETable(UStringSearch *strsrch,
         uprv_free(pattern->pces);
     }
 
-    uint16_t  offset = 0;
+    uint32_t  offset = 0;
     uint16_t  result = 0;
     int64_t   pce;
 
--- a/icu4c/source/test/cintltst/usrchtst.c
+++ b/icu4c/source/test/cintltst/usrchtst.c
@@ -2894,6 +2894,43 @@ exit:
    return;
 }
 
+static void TestUInt16Overflow(void) {
+    const int32_t uint16_overflow = UINT16_MAX + 1;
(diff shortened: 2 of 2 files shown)


icu4c/source/i18n/usearch.cpp
```diff
@@ -317,7 +317,7 @@ inline uint16_t initializePatternCETable(UStringSearch *strsrch,
         uprv_free(pattern->ces);
     }
 
-    uint16_t  offset      = 0;
+    uint32_t  offset      = 0;
```
pedberg-icu: Hmm. There is inconsistency in the places this offset value is used; some are int32_t (like pattern->cesLength, pattern->pcesLength) and some are uint32_t (like the corresponding parameter in addTouint64_tArray). We may need to be more consistent about usage.

icu4c/source/i18n/usearch.cpp
```diff
@@ -388,7 +388,7 @@ inline uint16_t initializePatternPCETable(UStringSearch *strsrch,
         uprv_free(pattern->pces);
     }
 
-    uint16_t  offset = 0;
+    uint32_t  offset = 0;
```
pedberg-icu: Same comment as above about int32_t vs uint32_t.

icu4c/source/i18n/usearch.cpp
jefgen: Thanks Peter. I could file a follow up ticket for 66 or 67, and this ticket could be just for the maint-65 release.

jefgen:
Thank you for reporting the issue (filing a ticket/issue) and for creating the PR!

In the ICU-TC call this ticket/PR was discussed and the general censuses was that this would be okay for the `maint/maint-65` branch, pending review on the changes from Peter. (Thank you Peter).

I'll change the target branch on this PR to be into the maint branch in moment.

icu4c/source/i18n/usearch.cpp
jefgen: I filed a follow-up ticket: https://unicode-org.atlassian.net/browse/ICU-20833

Aside: Since `addTouint64_tArray` uses `uint32_t` I wonder if we'd want to use that everywhere in here?

pedberg-icu: Yeah, probably uint32_t everywhere is best. Thanks for the followup! Will approve this one for 65.

PR #843, opened 2019-09-25
ICU-20693 Adding support for inferred ids.

hagbard:
Support for inferred IDs in the Ant XML file.

This dramatically reduces maintenance burden of the Ant build file and infers the regions, default scripts and variants from the base language / language+script. Basically it avoids needing to micro-manage the list of locales anywhere near as much.

Lots of checks to ensure that locale lists stay canonical too (good for long term code health).

##### Checklist

- [x] Issue accepted
- [x] Tests included


hagbard:
Dear reviewers, please only look at the change in https://github.com/unicode-org/icu/pull/843/commits/7b4cda8dcf268bbfe8947a18d8885b94425210be
Since that's the commit for this PR (the other 2 commits are on the previous PR which isn't in yet).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/ConvertIcuDataTask.java
```diff
@@ -73,59 +89,118 @@ public void setEmitReport(boolean emit) {
...
         @SuppressWarnings("unused")
         public void setTarget(String target) {
-            this.target = checkNotNull(target);
+            this.target = whitespace().trimFrom(target);
+        }
+
+        @Override
+        public void init() throws BuildException {
+            checkBuild(!source.isEmpty(), "Alias source must not be empty");
+            checkBuild(!target.isEmpty(), "Alias source must not be empty");
```
srl295: typo, `source` -> `target`

srl295: subminor, no reason to block this over it.

PR #844, opened 2019-09-25
ICU-20693 No arbitrary ordering in path/value visitation for new API

hagbard:
This is an optional, but very "nice to have" PR which removes all non-test use of the "ARBITRARY" order. As soon as everything is in and we're happy to change ICU files a bit, I will remove all ability to control path ordering from the API and go with DTD order everything (this will change supplemental data order a lot). This isolates ICU from any XML file ordering, which seems like a good thing. This PR is just me getting rid of the really egregious case ("arbitrary ordering") first since it really doesn't have any need to exist any more.

##### Checklist

- [x] Issue accepted
- [x] Tests included


hagbard:
Dear reviewers, please only look at the change in https://github.com/unicode-org/icu/pull/844/commits/eea9c3f2530c0ff36bc7a2b3aaddfec08c62efe0
Since that's the commit for this PR (the other 2 commits are on the previous PR which isn't in yet).

PR #845, opened 2019-09-25
ICU-20796 Updating double-conversion from upstream master.

sffc:
Upstream double-conversion added, removed, and renamed some files, so it took some time to touch them up for ICU.  I think I did it all correctly, but you might want to do a sanity check to confirm.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20796
- [x] Issue accepted
- [ ] Tests included


sffc:
Also PTAL at the updated README.md and pull-from-upstream.sh.  I rewrote the tool.

sffc:
Amended to fix test failures.

vendor/double-conversion/UPDATING.md
```diff
@@ -11,18 +17,36 @@ Run `pull-from-upstream.sh` as below:
 
 	./pull-from-upstream.sh <tag/branch>
 
-You will be prompted to download the tarball. If confirmed, the script will overwrite the contents of the upstream directory.
+The script runs in 4 steps.
```
jefgen: Thank you for updating the docs and rewriting the tool!

icu4c/source/i18n/double-conversion-utils.h
```diff
@@ -54,16 +56,23 @@
 #define DOUBLE_CONVERSION_NO_RETURN __attribute__((noreturn))
 #endif
 #endif
-#ifndef UNREACHABLE
+#ifndef DOUBLE_CONVERSION_UNREACHABLE
 #ifdef _MSC_VER
 void DOUBLE_CONVERSION_NO_RETURN abort_noreturn();
 inline void abort_noreturn() { abort(); }
-#define UNREACHABLE()   (abort_noreturn())
+#define DOUBLE_CONVERSION_UNREACHABLE()   (abort_noreturn())
```
jefgen: Not necessary right now, but I wonder if we might want to consider patching this to change it to use the ICU macro `UPRV_UNREACHABLE`, incase somebody customized that macro in ICU to do something other than call `abort()`.

It looks like the two macros are mostly the same, with the only difference being that the double-conversion macro is annotated as no-return for MSVC.
I think that this was done in the double-conversion code to support older versions of Visual Studio (ex: 2010), but I think ICU requires VS2015 due to the C++11 support.

sffc: Good thought.  I liked changing the ASSERT macro because it eliminated the inclusion of an extra header file.  I don't think changing this one would do the same.  Let's just leave it the way it is; less likely to break things, as you said about the return value semantics.

PR #853, opened 2019-09-26
ICU-20709 Tweak SignDisplay EXCEPT_ZERO semantics for ECMAScript.

sffc:
As discussed in the ticket, we have had more in-depth discussion about what would be the best semantics for SignDisplay.EXCEPT_ZERO.  I would like to make these semantics the default behavior in ICU.  The changes involve numbers very close to zero.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20709
- [x] Issue accepted
- [x] Tests included


sffc:
Closing in favor of #858.

PR #855, opened 2019-09-27
ICU-20303 Break Iterator, improve handling of look-ahead rules.

aheninger:
- Merge the look-ahead results slots used when multiple rules share a common accepting state.
- Sequentially number the look-ahead result slot. Will eventually allow replacing the runtime map with an array.
- Inhibit chaining out of look-ahead rules. This could never actually happen; when a hard break
  rule matches, the engine is stopped immediately, but the state table was being constructed
  as if it could  happen. Reduces table size for line break rules.
- Remove incorrect handling of fAccepting and fLookAhead fields of a state table row
  when removing duplicate states. Look-ahead slot number was being mis-interpreted as a state number.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20303
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/rbbi.cpp
+++ b/icu4c/source/common/rbbi.cpp
@@ -883,9 +883,15 @@ int32_t RuleBasedBreakIterator::handleNext() {
                 return lookaheadResult;
             }
         }
+
+        // If we are at the position of the '/' in a look-ahead (hard break) rule;
+        // record the current position, to be returned later, if the full rule matches.
+        // TODO: Move this check before the previous check of fAccepting.
+        //       This would enable hard-break rules with no following context.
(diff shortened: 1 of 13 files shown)


aheninger:
The C++ code is ready for review.
Needs to be ported to Java before merging.

Java and C++ must be in synch, or there will be ICU4J test failures in the data round trip test - the ICU4J rule builder will not recreate the same binary data.

aheninger:
Rebased up to master, which was recently refreshed with the latest from ICU 65.

This wanted to happen before Craig creates a branch from here and begins the Java port, to avoid merge conflicts from regenerating the ICU4J data jar.

PR #858, opened 2019-09-28
ICU-20709 Change EXCEPT_ZERO to check sign after rounding.

sffc:
**Tips for reviewing this PR:** I was very careful to split this PR into 3 disjoint commits which all pass tests on their own.  I suggest looking at each commit individually.

**Background:** As discussed in the ticket, we have had more in-depth discussion about what would be the best semantics for SignDisplay.EXCEPT_ZERO. I would like to make these semantics the default behavior in ICU. The changes involve numbers very close to zero.

This PR does not belong on the maintenance branch.  It is going on master for ICU 67.

@jefgen: assigning to you, but feel free to send it over to Markus or someone else.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20709
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/number_compact.cpp
+++ b/icu4c/source/i18n/number_compact.cpp
@@ -215,19 +215,25 @@ void CompactData::CompactDataSink::put(const char *key, ResourceValue &value, UB
 /// END OF CompactData.java; BEGIN CompactNotation.java ///
 ///////////////////////////////////////////////////////////
 
-CompactHandler::CompactHandler(CompactStyle compactStyle, const Locale &locale, const char *nsName,
-                               CompactType compactType, const PluralRules *rules,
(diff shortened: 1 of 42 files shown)


sffc:
Ping

jefgen:
@sffc Unfortunately, I don't think I'll have time to fully review this change. Is there perhaps somebody else that might be able to review this? Thanks.

sffc:
@FrankYFTang See tips for review in the first post.

sffc:
I don't want to squash for reasons described in the OP. However I'll add one more commit to change the 4 to a constant

sffc:
OK, fixed in a follow-up commit.  If there is a strong desire, I can rewrite the first commit on the branch to include the latest change.  My preference is to just put all four commits in order on master.

PR #869, opened 2019-10-01
ICU-20693 Pseudo-locale "alt path" filtering support.

hagbard:
https://unicode-org.atlassian.net/browse/ICU-20963

Adding support for two types of path "filtering" which are used to modify CLDR data during the ICU data build process.

alt-path support gives the equivalent functionality to the "data modifiers" file as used by FilterFactory
See: https://android.googlesource.com/platform/external/cldr/+/refs/heads/master/tools/java/org/unicode/cldr/tool/dataModifiers.txt
(Modified from https://github.com/unicode-org/cldr/blob/master/tools/java/org/unicode/cldr/tool/dataModifiers.txt)

pseudo-locale support replaces the need for a separate CLDRFilePseudolocalizer tool, allowing pseudo locales to just be treated as any other locale (controlled via a build flag).

This should complete the functionality needed for Google and Android special case behaviour during builds.


tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -283,6 +294,19 @@
...
+                 The target path will only be replaced if both it, and the source path, exist in
+                 the CLDR data (new paths will not be added if only the source path exists).
+
+                 Since the paths must represent the same semantic type of data, they must be in the
+                 same "namespace" (same element names) and must not contain value attributes. Thus
+                 they can only differ by distinguishing attributes (either added or modified).
+
+                 This feature is typically used to select alternate translations (e.g. short forms)
+                 for certain paths. -->
+            <!-- <altPath target="//path/to/value[@attr='foo']" source="//path/to/value[@attr='bar']"/> -->
```
hagbard: No actual need for this by default, but the comment + example should be enough for people to work with.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleData.java
```diff
@@ -0,0 +1,126 @@
...
+ * not possible to "move" values between arbitrary paths. Target and source paths must be in the
+ * same "namespace" (i.e. share the same element names) but attributes can differ.
+ *
+ * <p>Note also that the mapping is not recursive, so mapping {@code A -> B} and {@code B -> C}
+ * will NOT cause {@code A} to be mapped to {@code C}.
+ *
+ * <p>Typically this class is expected to be used for selecting alternate values of locale data
+ * based on the {@code "alt"} path attribute (e.g. selecting the short form of a region name).
+ */
+public final class AlternateLocaleData {
```
hagbard: This is a neat abstraction to avoid any knowledge of alt-paths in the core logic. It returns a supplier which modifies CLDR data on the fly based on a given map. This means that all mapping classes just see a "normal" supplier and don't have to care.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleData.java
```diff
@@ -0,0 +1,126 @@
...
+                    if (altValue != null) {
+                        return altValue.replacePath(value.getPath());
+                    }
+                }
+                return value;
+            }
+        }
+    }
+
+    private static boolean hasSameNamespace(CldrPath x, CldrPath y) {
```
hagbard: Enforces the requirement that you cannot swap different paths (which would leave CldrValues potentially broken since value attributes would then not be associated with an element on their path).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/FilteredData.java
```diff
@@ -0,0 +1,67 @@
...
+import org.unicode.cldr.api.CldrData;
+import org.unicode.cldr.api.CldrPath;
+import org.unicode.cldr.api.CldrValue;
+
+/**
+ * A class which allows data from some underlying {@link CldrData} source to be filtered or
+ * removed (but not added).
+ */
+// TODO: Once DTD ordering is the only allowed order, this can be extended to allow adding paths.
+abstract class FilteredData implements CldrData {
```
hagbard: Base class representing filtered CLDR data (used by both alt-path and pseudo locales).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -137,76 +138,48 @@ private static PathMatcher supplementalMatcher(String... spec) {
...
-            LdmlConverter::processKeyTypeData),
-
-        // Batching by type.
-        DTD_LDML(LDML, c -> c.processAll(LDML)),
-        DTD_SUPPLEMENTAL(SUPPLEMENTAL, c -> c.processAll(SUPPLEMENTAL)),
-        DTD_BCP47(BCP47, c -> c.processAll(BCP47));
-
-        public static final ImmutableSet<OutputType> ALL =
-            ImmutableSet.of(DTD_BCP47, DTD_SUPPLEMENTAL, DTD_LDML);
+        LOCALES(LDML),
```
hagbard: Simplify the enum to stop it doing dispatching (that's now done below).
This is needed because the filtering work wants to operate on a single data instance (previously we would run over the same data multiple times, depending on the output types). Now we just map the data from a CLDR supplier once and then split it afterwards.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -137,76 +138,48 @@ private static PathMatcher supplementalMatcher(String... spec) {
...
 
         CldrDataType getCldrType() {
             return type;
         }
     }
 
+    // Map to convert the rather arbitrarily defined "output types" to the directories into which
+    // the data is written. This is only for "LDML" types since other mappers don't need to split
+    // data into multiple directories.
+    private static final ImmutableListMultimap<OutputType, IcuLocaleDir> TYPE_TO_DIR =
```
hagbard: This is so that we can decide what processing to do based on the desired "output type". Important since otherwise simple stuff like "regenerate only collation data" would take 15 minutes rather than 2 minutes.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -344,7 +273,20 @@ private void processAndSplitLocaleFiles(
             if (!availableIds.contains(id)) {
                 continue;
             }
-            IcuData icuData = icuFn.apply(id);
+
+            IcuData icuData = new IcuData(id, true);
```
hagbard: Now we write a single IcuData instance and load specials data only once.
All mappers are executed and then we split results according to the resource bundle.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -445,42 +396,77 @@ private void processAndSplitLocaleFiles(
...
-    }
-
-    private void processKeyTypeData() {
-        Bcp47Mapper.process(src).forEach(d -> write(d, "misc"));
-    }
-
-    private void processTransforms() {
-        Path transformDir = createDirectory(config.getOutputDir().resolve("translit"));
-        write(TransformsMapper.process(src, transformDir, fileHeader), transformDir);
+    private void processSupplemental() {
```
hagbard: This is just the unwrapping of the callbacks from the enum (above).

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/ConvertIcuDataTask.java
```diff
@@ -50,6 +59,11 @@
     // Per directory overrides (fully specified locale IDs).
     private final SetMultimap<IcuLocaleDir, String> perDirectoryIds = HashMultimap.create();
     private final IcuConverterConfig.Builder config = IcuConverterConfig.builder();
+    // Don't try and resolve actual paths until inside the execute method.
```
hagbard: Just support for more configuration in the Ant build.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/AbstractPathValueMapper.java
```diff
@@ -55,12 +55,11 @@
...
-     * Returns a new {@code IcuData} instance produced by post-processing a set of results
-     * generated by calling sub-class method {@link #addResults()}. This is the only method which
-     * need be directly invoked by the sub-class implementation (other methods are optionally used
-     * from within the {@link #addResults()} callback).
+     * Post-processes results generated by calling the subclass method {@link #addResults()}. This
+     * is the only method which need be directly invoked by the sub-class implementation (other
+     * methods are optionally used from within the {@link #addResults()} callback).
      */
-    final IcuData generateIcuData(String icuName, boolean hasFallback) {
+    final void addIcuData(IcuData icuData) {
```
hagbard: These mapper changes (here and below) are just to switch to taking and modifying the IcuData rather than creating and returning one. This is all part of the change to process the CLDR data just once to build a single data instance.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/LocaleMapper.java
```diff
@@ -47,50 +41,30 @@
...
 
-    private final String localeId;
-    private final CldrDataSupplier src;
+    private final CldrData unresolved;
     private final Optional<CldrData> icuSpecialData;
 
     private LocaleMapper(
-        String localeId,
-        CldrDataSupplier src,
+        CldrData unresolved,
```
hagbard: Since we only want to generate data once, we pass in the data (which will cache and is reusable) rather than passing in the supplier (which generates new and independent each time).

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
```diff
@@ -0,0 +1,152 @@
...
+    public static void assertValuesUnordered(CldrData data, CldrValue... values) {
+        Set<CldrValue> captured = new HashSet<>();
+        data.accept(ARBITRARY, captured::add);
+        assertThat(captured).containsExactlyElementsIn(values);
+    }
+
+    private static CldrValue ldml(String path, String value) {
+        return CldrValue.parseValue("//ldml/" + path, value);
+    }
+}
```
hagbard: Damn, we can either leave these missing EOF newlines (2 or 3 that I can see, only in tests) or do a quick "in github edit" to add them. I thought IntelliJ was configured to never do this, but obviously not.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PseudoLocales.java
```diff
@@ -0,0 +1,384 @@
...
+        @Override public CldrData getDataForType(CldrDataType type) {
+            return src.getDataForType(type);
+        }
+    }
+
+    private interface PseudoText {
+        void addFragment(String text, boolean isLocalizable);
+    }
+
+    private static final class PseudoLocaleData extends FilteredData {
```
markusicu: Shouldn't the implementation of the pseudo-locale transformations remain in CLDR code, ideally with some public CLDR API? Or as an option on the CLDR data API??

ICU need not be the only user of pseudo-localized CLDR data.
For example, it should be possible to add pseudolocales to the CLDR JSON data.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/ConvertIcuDataTask.java
```diff
@@ -50,6 +59,11 @@
...
+    private final Map<String, String> altPathMap = new HashMap<>();
+    // TODO: Maybe move into CLDR API; e.g. withPseduoLocales()
```
markusicu: typo Pseduo

markusicu:
FYI I only glanced over a lot of the framework changes. They look plausible, but it would be nice if someone more familiar with the flow and logic of this converter (Mark and/or Steven) could also review.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PseudoLocales.java
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        @Override public CldrDataSupplier withDraftStatusAtLeast(CldrDraftStatus draftStatus) {
+            return new PseudoSupplier(src.withDraftStatusAtLeast(draftStatus));
+        }
+
+        @Override public CldrData getDataForLocale(String localeId, CldrResolution resolution) {
+            if (PseudoType.getLocaleIds().contains(localeId)) {
+                // Use resolved data to ensure we get all the values (e.g. values in "en_001").
+                CldrData enData = src.getDataForLocale("en", RESOLVED);
```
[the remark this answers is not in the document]
hagbard: Hmm, you know I thought I had done that. Thanks for catching. I might be missing a "push".

markusicu:
FYI In hindsight, it feels like this could have benefited from being three PRs: Refactoring / alt paths / pseudolocales. But I don't want to make you jump through hoops now.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PseudoLocales.java
hagbard: I'd be happy to move it once we agree on the API. I did it like this since it wasn't initially obvious how we would handle it in the common API (I actually initially didn't think the pseudo locales were legal CLDR language codes, but rather Google/ICU specific hacks).

markusicu: Ok, in that case please leave a TODO with CLDR & ICU follow-up ticket numbers.

FYI XA & XB officially listed in http://www.unicode.org/reports/tr35/tr35.html#unicode_region_subtag_validity

hagbard:
> FYI In hindsight, it feels like this could have benefited from being three PRs: Refactoring / alt paths / pseudolocales. But I don't want to make you jump through hoops now.

Yeah, at the time it was all a bit rushed to try and get in for the cut-off.

tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -39,6 +39,15 @@
              CldrDraftStatus for more details. -->
         <property name="minDraftStatus" value="contributed"/>
 
+        <!-- A regular expression to match the locale IDs to be generated (useful for
+             debugging specific regions). This is applied after locale ID specifications
+             have been expanded into full locale IDs, so the value "en" will NOT match
+             "en_GB" or "en_001" etc. -->
+        <property name="localeIdFilter" value=""/>
+
+        <!-- Whether to synthetically generate "pseudo locale" data ("ar_XA" and "en_XB"). -->
```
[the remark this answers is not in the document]
hagbard: Yeah, I got that wrong so many times ... thanks for the mnemonic.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/PseudoLocales.java
hagbard: Done. CLDR-13381 (not sure it also needs an ICU ticket though).

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
hagbard: Done in latest push

PR #885, opened 2019-10-11
ICU-20862 Fix setKeywordValue U_BUFFER_OVERFLOW_ERROR bug.

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20862
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -1340,7 +1340,31 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
 void
 Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErrorCode &status)
 {
-    uloc_setKeywordValue(keywordName, keywordValue, fullName, ULOC_FULLNAME_CAPACITY, &status);
+    if (U_FAILURE(status)) {
+        return;
+    }
+    int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
+    int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
+                                             bufferLength, &status) + 1;
+    /* Handle the case the current buffer is not enough to hold the new id */
+    if (status == U_BUFFER_OVERFLOW_ERROR) {
+        U_ASSERT(newLength > bufferLength);
+        char* newFullName = (char *)uprv_malloc(newLength);
+        if (newFullName == nullptr) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+            return;
+        }
+        uprv_strcpy(newFullName, fullName);
+        if (fullName != fullNameBuffer) {
+            // if full Name is already on the heap, need to free it.
+            uprv_free(fullName);
+        }
+        fullName = newFullName;
+        status = U_ZERO_ERROR;
+        uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status);
+    } else {
+        U_ASSERT(newLength <= bufferLength);
+    }
     if (U_SUCCESS(status) && baseName == fullName) {
         // May have added the first keyword, meaning that the fullName is no longer also the baseName.
         initBaseName(status);
--- a/icu4c/source/test/intltest/loctest.cpp
+++ b/icu4c/source/test/intltest/loctest.cpp
@@ -271,6 +271,8 @@ void LocaleTest::runIndexedTest( int32_t index, UBool exec, const char* &name, c
     TESTCASE_AUTO(TestPointerConvertingIterator);
     TESTCASE_AUTO(TestTagConvertingIterator);
     TESTCASE_AUTO(TestCapturingTagConvertingIterator);
+    TESTCASE_AUTO(TestSetUnicodeKeywordValueInLongLocale);
+    TESTCASE_AUTO(TestSetUnicodeKeywordValueNullInLongLocale);
     TESTCASE_AUTO_END;
 }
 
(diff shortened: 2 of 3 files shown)


roubert:
> Notice: the branch changed across the force-push!

Please tell me when you think that this PR is ready for review (ie. when you won't change it with any force-push again).

FrankYFTang:
> > Notice: the branch changed across the force-push!
> 
> Please tell me when you think that this PR is ready for review (ie. when you won't change it with any force-push again).

sorry, get into some error of int size.

FrankYFTang:
Please review @roubert

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,25 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
 void
 Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErrorCode &status)
 {
-    uloc_setKeywordValue(keywordName, keywordValue, fullName, ULOC_FULLNAME_CAPACITY, &status);
+    int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
+    int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
+                                         bufferLength, &status) + 1;
+    /* Handle the case the current buffer is not enough to hold the new id */
+    if (status == U_BUFFER_OVERFLOW_ERROR || newLength > bufferLength) {
```
roubert: This should be `&&` instead of `||`, shouldn't it?

Unless something inside of `uloc_setKeywordValue()` is broken, it should not be possible for one of these two conditions to be true while the other is not.

If you want to be defensive against such a possible bug, you could add an assertion here to verify that these two conditions are always either both true or both false:

```
    if (status == U_BUFFER_OVERFLOW_ERROR) {
        U_ASSERT(newLength > bufferLength);
        // ...
    } else {
        U_ASSERT(newLength <= bufferLength);
    }
```

But the way the code is written now, with `||` instead of `&&`, it looks as if a value of `status` that is something other than `U_BUFFER_OVERFLOW_ERROR` could possibly be overwritten with `U_ZERO_ERROR` below (which would be wrong do to).

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,25 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
...
+            return;
+        }
+        uprv_strcpy(newFullName, fullName);
+        if (fullName != fullNameBuffer) {
+            // if full Name is already on the heap, need to free it.
+            uprv_free(fullName);
+        }
+        fullName = newFullName;
+        status = U_ZERO_ERROR;
+        newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status) + 1;
```
roubert: There is no reason to assign a new value to `newLength` here, that variable is never accessed again after this statement.

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,25 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
 void
 Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErrorCode &status)
 {
-    uloc_setKeywordValue(keywordName, keywordValue, fullName, ULOC_FULLNAME_CAPACITY, &status);
+    int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
+    int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
+                                         bufferLength, &status) + 1;
```
roubert: The indentation here is odd.

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,25 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
...
+    /* Handle the case the current buffer is not enough to hold the new id */
+    if (status == U_BUFFER_OVERFLOW_ERROR || newLength > bufferLength) {
+        char* newFullName = (char *)uprv_malloc(sizeof(char)*(newLength));
```
roubert: I don't think that `sizeof(char)` adds any value here, it's by definition 1 and I'm quite certain that everyone who ever reads this code knows that.

icu4c/source/test/intltest/loctest.cpp
```diff
@@ -3952,3 +3954,71 @@ void LocaleTest::TestCapturingTagConvertingIterator() {
...
     assertFalse("3.hasNext()", iter.hasNext());
 }
+
+void LocaleTest::TestSetUnicodeKeywordValueInLongLocale() {
+    IcuTestErrorCode status(*this, "TestSetUnicodeKeywordValueInLongLocale");
+    const char* value = "efghijkl";
+    icu::Locale l("de");
+    char keyword[3];
+    CharString expected;
+    expected.append("de-u", status);
```
roubert: It'd be simpler to do this directly in the constructor:

```
    CharString expected("de-u", status);
```

icu4c/source/test/intltest/loctest.cpp
```diff
@@ -3952,3 +3954,71 @@ void LocaleTest::TestCapturingTagConvertingIterator() {
...
+    }
+}
+
+void LocaleTest::TestSetUnicodeKeywordValueNullInLongLocale() {
+    IcuTestErrorCode status(*this, "TestSetUnicodeKeywordValueNullInLongLocale");
+    const char *exts[] = {"cf", "cu", "em", "kk", "kr", "ks", "kv", "lb", "lw",
+      "ms", "nu", "rg", "sd", "ss", "tz"};
+    for (int32_t i = 0; i < UPRV_LENGTHOF(exts); i++) {
+        CharString tag;
+        tag.append("de-u", status);
```
roubert: It'd be simpler to do this directly in the constructor:

```
        CharString tag("de-u", status);
```

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,25 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
 void
 Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErrorCode &status)
 {
-    uloc_setKeywordValue(keywordName, keywordValue, fullName, ULOC_FULLNAME_CAPACITY, &status);
+    int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
+    int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
```
aheninger: Need to bail out early if the incoming status shows a failure. uloc_setKeywordValue(..., status) will return a meaningless value. And the incoming failure could be U_BUFFER_OVERFLOW_ERROR, which will further confuse the following logic.

icu4c/source/common/locid.cpp
FrankYFTang: see existing code in Locale::init

icu4c/source/common/locid.cpp
FrankYFTang: again, see existing code in  Locale::init

FrankYFTang:
all comments addressed. PTAL

icu4c/source/common/locid.cpp
```diff
@@ -1357,7 +1361,9 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
         }
         fullName = newFullName;
         status = U_ZERO_ERROR;
-        newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status) + 1;
+        uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status) + 1;
+    } else {
+        U_ASSERT(status != U_BUFFER_OVERFLOW_ERROR);
```
roubert: This lines makes no sense, after `if (status == U_BUFFER_OVERFLOW_ERROR)` it couldn't ever happen that this assertion isn't true, unless you have a compiler so broken that it generates bad code for if-statements.

icu4c/source/common/locid.cpp
```diff
@@ -1357,7 +1361,9 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
         }
         fullName = newFullName;
         status = U_ZERO_ERROR;
-        newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status) + 1;
+        uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status) + 1;
```
roubert: The final `+ 1` here makes no sense at all, this value is never used for anything.

FrankYFTang:
Sorry, PTAL

FrankYFTang:
Please approve again after the SQ @roubert

icu4c/source/common/locid.cpp
```diff
@@ -1340,7 +1340,31 @@ Locale::getUnicodeKeywordValue(StringPiece keywordName,
...
 Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErrorCode &status)
 {
-    uloc_setKeywordValue(keywordName, keywordValue, fullName, ULOC_FULLNAME_CAPACITY, &status);
+    if (U_FAILURE(status)) {
+        return;
+    }
+    int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
+    int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
+                                             bufferLength, &status) + 1;
+    /* Handle the case the current buffer is not enough to hold the new id */
```
sffc: Would be better to use a localStatus here

PR #888, opened 2019-10-16
ICU-20872 remove extra ; after function {}

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20872
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/numberformatter.h
+++ b/icu4c/source/i18n/unicode/numberformatter.h
@@ -2600,7 +2600,7 @@ StringClass FormattedNumber::toDecimalNumber(UErrorCode& status) const {
     StringByteSink<StringClass> sink(&result);
     toDecimalNumber(sink, status);
     return result;
-};
+}
 #endif // U_HIDE_DRAFT_API
 
 /**


jefgen:
> @jefgen How did this get past our build bots that were specifically designed to prevent this from happening?

I think this might be because the build bots are still on Clang version 6.0 ([link](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md)), and it looks like it wasn't until Clang [version 7.0](http://releases.llvm.org/7.0.0/tools/clang/docs/ReleaseNotes.html) that `Wextra-semi` also enabled `Wc++98-compat-extra-semi`.

So I think we missed this when we added `Wextra-semi` to the build bots, so adding `Wc++98-compat-extra-semi` seems like the right thing since the bots are still on v6.0

sffc:
It looks like the built bot doesn't like the new flag:

```
checking whether the C compiler works... no
configure: error: in `/home/vsts/work/1/s/icu4c/source':
```

Is this flag *only* supported in Clang 7+?  Maybe Clang 6 doesn't know about that flag and exits when it sees it.

FrankYFTang:
> It looks like the built bot doesn't like the new flag:
> 
> ```
> checking whether the C compiler works... no
> configure: error: in `/home/vsts/work/1/s/icu4c/source':
> ```
> 
> Is this flag _only_ supported in Clang 7+? Maybe Clang 6 doesn't know about that flag and exits when it sees it.

I remove the flag change. Could we just land this only with the .h change?

PR #890, opened 2019-10-20
ICU-20863 Regex, lazy creation and reduced size of map from capture g…

aheninger:
…roup names to numbers.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20863
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/repattrn.cpp
+++ b/icu4c/source/i18n/repattrn.cpp
@@ -138,18 +138,20 @@ RegexPattern &RegexPattern::operator = (const RegexPattern &other) {
     }
 
     // Copy the named capture group hash map.
-    int32_t hashPos = UHASH_FIRST;
-    while (const UHashElement *hashEl = uhash_nextElement(other.fNamedCaptureMap, &hashPos)) {
-        if (U_FAILURE(fDeferredStatus)) {
-            break;
-        }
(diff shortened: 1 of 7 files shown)


icu4c/source/i18n/repattrn.cpp
```diff
@@ -191,27 +193,38 @@ void RegexPattern::init() {
...
 
     // fNamedCaptureMap owns its key strings, type (UnicodeString *)
     uhash_setKeyDeleter(fNamedCaptureMap, uprv_deleteUObject);
+    return true;
 }
 
-
 //--------------------------------------------------------------------------
 //
 //   zap            Delete everything owned by this RegexPattern.
```
gvictor: [Intend to comment line 262 in github UI]
nit: a style question. Although uhash_close() does check the pointer null before deletion. Should null check be added  before calling uhash_close()?

For example: RBBIRuleScanner::~RBBIRuleScanner() does check null before calling uhash_close().

aheninger: OK, Changed it for consistency with other related code. If I were doing it all from scratch, I think I would omit all the extra NULL checks.

markusicu: FYI I wouldn't add such NULL checks, just like we don't check for nullptr before `delete ptr;`.

PR #892, opened 2019-10-21
ICU-20875 Include <cstddef> for max_align_t

jmroot:
The definition of `max_align_t` is not guaranteed to be available unless
the appropriate header is included. Since use of `<stddef.h>` from C++ is
deprecated, that's `<cstddef>`, and `max_align_t` is thus defined under the
std namespace rather than in the global namespace.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20875
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/uarrsort.cpp
+++ b/icu4c/source/common/uarrsort.cpp
@@ -18,6 +18,8 @@
 *   Internal function for sorting arrays.
 */
 
+#include <cstddef>
+
 #include "unicode/utypes.h"
 #include "cmemory.h"
 #include "uarrsort.h"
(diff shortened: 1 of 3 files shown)


jefgen:
Thanks for the PR!

Would you be able to edit or [amend ](https://www.atlassian.com/git/tutorials/rewriting-history#git-commit--amend) your commit message to include the ticket number in the commit title?

It should look like this: `ICU-20875 Include <cstddef> for max_align_t`

This will make the jira-ticket checker happy. :) 
Thanks!

PR #893, opened 2019-10-24
ICU-20863 Regex Named Capture map, add a missing nullptr check.

aheninger:
https://unicode-org.atlassian.net/browse/ICU-20863

Follow-up on PR #890


markusicu:
once merged, please cherry-pick to maint/maint-66 where #890 already landed (because we cut the maintenance branch from master after you merged that)

PR #894, opened 2019-10-25
ICU-12863 Add list style APIs to C and C++

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-12863
- [ ] Issue accepted
- [x] Tests included

@jswalden

### DIFF
--- a/icu4c/source/i18n/listformatter.cpp
+++ b/icu4c/source/i18n/listformatter.cpp
@@ -80,7 +80,6 @@ UPRV_FORMATTED_VALUE_SUBCLASS_AUTO_IMPL(FormattedList)
 
 
 static Hashtable* listPatternHash = nullptr;
-static const char STANDARD_STYLE[] = "standard";
 
 U_CDECL_BEGIN
 static UBool U_CALLCONV uprv_listformatter_cleanup() {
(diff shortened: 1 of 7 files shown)


icu4c/source/i18n/listformatter.cpp
```diff
@@ -179,6 +178,42 @@ const ListFormatInternal* ListFormatter::getListFormatInternal(
     return result;
 }
 
+static const char* typeWidthToStyleString(UListFormatType type, UListFormatWidth width) {
+    switch (type) {
+        case ULISTFMT_TYPE_CONJUNCTION:
+            switch (width) {
+                case ULISTFMT_WIDTH_SHORT:
+                    return "standard";
+                default:
```
FrankYFTang: Please add  case for ULISTFMT_WIDTH_SHORT: nor ULISTFMT_WIDTH_NARROW: here

icu4c/source/i18n/listformatter.cpp
```diff
@@ -179,6 +178,42 @@ const ListFormatInternal* ListFormatter::getListFormatInternal(
...
+                    return nullptr;
+            }
+            break;
+
+        case ULISTFMT_TYPE_DISJUNCTION:
+            switch (width) {
```
FrankYFTang: Please add  case for ULISTFMT_WIDTH_SHORT: nor ULISTFMT_WIDTH_NARROW: here

icu4c/source/i18n/listformatter.cpp
sffc: CONJUNCTION and DISJUNCTION only support SHORT width.  Why do I need to add cases for the others?  It will just cause a U_ILLEGAL_ARGUMENT_ERROR.

FrankYFTang: Disagree. Read https://unicode.org/reports/tr35/tr35-general.html#ListPatterns
Notice "type attribute value" include all 9 combinations
Also see https://github.com/unicode-org/icu/blob/master/icu4c/source/data/locales/en.txt#L2007

icu4c/source/test/intltest/listformattertest.cpp
```diff
@@ -611,6 +611,55 @@ void ListFormatterTest::TestBadStylesFail() {
...
 
+void ListFormatterTest::TestCreateStyled() {
+    IcuTestErrorCode status(*this, "TestCreateStyled");
+    // Locale en has interesting data
+    struct TestCase {
+        const char* locale;
+        UListFormatterType type;
+        UListFormatterWidth width;
+        const char16_t* expected;
+    } cases[] = {
```
FrankYFTang: Please add more tests to cover
1. where there are only 1 item in the list
2. where there are only 2 items in the list

sffc: Done.

sffc:
Ping (I squashed and need another stamp)

PR #895, opened 2019-10-29
ICU-20863 Regex Named Capture map, add a missing nullptr check.

aheninger:
https://unicode-org.atlassian.net/browse/ICU-20863


aheninger:
This is a cherry-pick of e94657e6 from master into maint-66.
The PR on master is https://github.com/unicode-org/icu/pull/893

PR #896, opened 2019-10-29
ICU-20705 Add udtitvfmt_formatCalendarToResult

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20705
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/udateintervalformat.cpp
+++ b/icu4c/source/i18n/udateintervalformat.cpp
@@ -128,8 +128,27 @@ udtitvfmt_formatToResult(
     }
     auto* resultImpl = UFormattedDateIntervalApiHelper::validate(result, *status);
     DateInterval interval = DateInterval(fromDate,toDate);
-    resultImpl->fImpl = reinterpret_cast<const DateIntervalFormat*>(formatter)
-        ->formatToValue(interval, *status);
+    if (resultImpl != nullptr) {
+        resultImpl->fImpl = reinterpret_cast<const DateIntervalFormat*>(formatter)
+            ->formatToValue(interval, *status);
+    }
+}
+
+U_DRAFT void U_EXPORT2
+udtitvfmt_formatCalendarToResult(
+                const UDateIntervalFormat* formatter,
+                UCalendar*      fromCalendar,
+                UCalendar*      toCalendar,
+                UFormattedDateInterval* result,
+                UErrorCode*     status) {
+    if (U_FAILURE(*status)) {
+        return;
+    }
+    auto* resultImpl = UFormattedDateIntervalApiHelper::validate(result, *status);
+    if (resultImpl != nullptr) {
+        resultImpl->fImpl = reinterpret_cast<const DateIntervalFormat*>(formatter)
+            ->formatToValue(*(Calendar *)fromCalendar, *(Calendar *)toCalendar, *status);
+    }
 }
 
 
--- a/icu4c/source/test/cintltst/cdateintervalformattest.c
+++ b/icu4c/source/test/cintltst/cdateintervalformattest.c
@@ -21,6 +21,7 @@
 static void TestDateIntervalFormat(void);
 static void TestFPos_SkelWithSeconds(void);
 static void TestFormatToResult(void);
+static void TestFormatCalendarToResult(void);
 
 void addDateIntervalFormatTest(TestNode** root);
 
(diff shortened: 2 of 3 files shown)


icu4c/source/i18n/udateintervalformat.cpp
```diff
@@ -132,5 +132,20 @@ udtitvfmt_formatToResult(
...
+                const UDateIntervalFormat* formatter,
+                UFormattedDateInterval* result,
+                UCalendar*      fromCalendar,
+                UCalendar*      toCalendar,
+                UErrorCode*     status) {
+    if (U_FAILURE(*status)) {
+        return;
+    }
+    auto* resultImpl = UFormattedDateIntervalApiHelper::validate(result, *status);
+    resultImpl->fImpl = reinterpret_cast<const DateIntervalFormat*>(formatter)
```
jefgen: Just a suggestion/comment:
It looks like the `validate` method can return `nullptr` and set status to `U_INVALID_FORMAT_ERROR` in some cases. Given this I wonder if maybe the code should check the status before de-referencing the `resultImpl` to prevent a crash or seg-fault.
(I noticed that the method above doesn't do this, but perhaps that's another bug/ticket).

FrankYFTang: DONE. I check the nullptr of resultImpl

icu4c/source/i18n/udateintervalformat.cpp
```diff
@@ -128,8 +128,27 @@ udtitvfmt_formatToResult(
...
+                UCalendar*      toCalendar,
+                UFormattedDateInterval* result,
+                UErrorCode*     status) {
+    if (U_FAILURE(*status)) {
+        return;
+    }
+    auto* resultImpl = UFormattedDateIntervalApiHelper::validate(result, *status);
+    if (resultImpl != nullptr) {
+        resultImpl->fImpl = reinterpret_cast<const DateIntervalFormat*>(formatter)
+            ->formatToValue(*(Calendar *)fromCalendar, *(Calendar *)toCalendar, *status);
```
sffc: Although unpleasant, this is consistent with the pattern used for UCalendar in ucal.cpp (C-style cast, no type check or null check)

icu4c/source/test/cintltst/cdateintervalformattest.c
```diff
@@ -380,4 +382,75 @@ static void TestFormatToResult() {
...
+            {UFIELD_CATEGORY_DATE, UDAT_MINUTE_FIELD, 45, 47}};
+        checkMixedFormattedValue(
+            message,
+            udtitvfmt_resultAsValue(fdi, &ec),
+            expectedString,
+            expectedFieldPositions,
+            UPRV_LENGTHOF(expectedFieldPositions));
+    }
+    {
+        const char* message = "Field position test 1";
```
sffc: Give the tests different messages

icu4c/source/test/cintltst/cdateintervalformattest.c
```diff
@@ -380,4 +382,75 @@ static void TestFormatToResult() {
     udtitvfmt_closeResult(fdi);
 }
 
+static void TestFormatCalendarToResult() {
```
sffc: Can you write a test where the calendar in the locale is different from the calendar in the UCalendar objects, which is the main point of adding this new API?

icu4c/source/i18n/udateintervalformat.cpp
FrankYFTang: > Although unpleasant, this is consistent with the pattern used for UCalendar in ucal.cpp (C-style cast, no type check or null check)

Notice this is different from ucal.cpp. In ucal.cpp, it is the direct use of the passing parameter after casting. But in here, the pointer is not directly passing in from the coder, but return from the validate function. There are no guarantee that while the input parameter is not null the validate will always return not null.

sffc: The use of `result` is guarded, but `fromCalendar` and `toCalendar` are not guarded.  But again, that's consistent with ucal.cpp, so no action required, but if you wanted to be extra safe you could do a null or branding check.

icu4c/source/test/cintltst/cdateintervalformattest.c
```diff
@@ -422,7 +426,7 @@ static void TestFormatCalendarToResult() {
             UPRV_LENGTHOF(expectedFieldPositions));
     }
     {
-        const char* message = "Field position test 1";
+        const char* message = "Field position test 2";
```
sffc: Make this one 3 and the next one 4

FrankYFTang: why?

sffc: Nevermind, I saw another change but that was in the previous function.

sffc:
Proposal was approved.  LGTM please squash

PR #899, opened 2019-10-30
ICU-20744 Add VS Code configuration files for c++

younies:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20744
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/docs/ide4c/vscode/README.md
+++ b/docs/ide4c/vscode/README.md
@@ -1,0 +1,10 @@
+<!--- © 2020 and later: Unicode, Inc. and others. ---> 
+<!--- License & terms of use: http://www.unicode.org/copyright.html#License --->
+
+# Configuring VS Code for ICU4C
+
+  - create `.vscode` folder in icu4c/source
+  - Copy `tasks.json` and `launch.json` files into `.vscode` folder
+
+NOTE
+ Before you build `icu4c` from vs code. run [`./runConfigureICU` command first](http://userguide.icu-project.org/icufaq).
--- a/docs/ide4c/vscode/launch.json
+++ b/docs/ide4c/vscode/launch.json
@@ -1,0 +1,100 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html#License
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "intltest Launch",
+            "type": "cppdbg",
+            "request": "launch",
+            "targetArchitecture": "x64",
+            "program": "${workspaceFolder}/test/intltest/intltest",
+            "environment": [
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add which test cases you want to run, leaving it empty will run all the test cases.
+                "rbbi/RBBITest" // Just for example.
+            ],
+            "stopAtEntry": false, // If you want to add a break point in the beginning of the program (i.e. test cases).
+            "cwd": "${workspaceFolder}",
+            "externalConsole": false, //set to true to see output in an external terminal instead
+            "MIMode": "gdb", // Use "lldb" for mac osx.
+            "logging": {
+                "trace": true,
+                "traceResponse": true,
+                "engineLogging": true,
+                "programOutput": true
+            },
+            "preLaunchTask": "buildTest"
+        },
+        {
+            "name": "cintltst Launch",
+            "type": "cppdbg",
+            "request": "launch",
+            "targetArchitecture": "x64",
+            "program": "${workspaceFolder}/test/cintltst/cintltst",
+            "environment": [
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add which test cases you want to run, leaving it empty will run all the test cases.
+                "/tsutil/cstrcase" // Just for example.
+            ],
+            "stopAtEntry": false, // If you want to add a break point in the beginning of the program (i.e. test cases).
+            "cwd": "${workspaceFolder}",
+            "externalConsole": false, //set to true to see output in an external terminal instead
+            "MIMode": "gdb", // Use "lldb" for mac osx.
+            "logging": {
+                "trace": true,
+                "traceResponse": true,
+                "engineLogging": true,
+                "programOutput": true
+            },
+            "preLaunchTask": "buildTest"
+        },
+        {
+            "name": "iotest Launch",
+            "type": "cppdbg",
+            "request": "launch",
+            "targetArchitecture": "x64",
+            "program": "${workspaceFolder}/test/iotest/iotest",
+            "environment": [
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add which test cases you want to run, leaving it empty will run all the test cases.
+                "/datadriv" // Just for example.
+            ],
+            "stopAtEntry": false, // If you want to add a break point in the beginning of the program (i.e. test cases).
+            "cwd": "${workspaceFolder}",
+            "externalConsole": false, //set to true to see output in an external terminal instead
+            "MIMode": "gdb", // Use "lldb" for mac osx.
+            "logging": {
+                "trace": true,
+                "traceResponse": true,
+                "engineLogging": true,
+                "programOutput": true
+            },
+            "preLaunchTask": "buildTest"
+        }
+    ]
+}
--- a/docs/ide4c/vscode/tasks.json
+++ b/docs/ide4c/vscode/tasks.json
@@ -1,0 +1,38 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html#License
+
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "Makefile", // You can choose your own label.
+            "type": "shell",
+            "command": "make",
+            // Make this the default build command.
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
+            "presentation": {
+                "reveal": "always"
+            },
+            // You can pass all the building arguments here.
+            "args": [
+                "all",
+                "--debug", // Enable debug mode.
+                "-j", // Use more than one core in parallel
+                "-l90" // Use 90 percent of the CPU
+            ]
+        },
+        {
+            "label": "buildTest", // for building test cases before running it.
+            "command": "make", // Could be any other shell command
+            "args": [
+                "tests",
+                "-j", // Use more than one core in parallel
+                "-l90" // Use 90 percent of the CPU
+            ],
+            "type": "shell"
+        }
+    ]
+}


docs/ide4c/vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+            // You can pass all the building arguments here.
+            "args": [
+                "all",
+                "--debug", // Enable debug mode.
+                "-j8" // Use 8 cores.
+            ]
+        },
+        {
+            "label": "buildTest", // for building test cases before running it.
+            "command": "make", // Could be any other shell command
```
sffc: `make tests` does this.  Better than cd into the directory.

docs/ide4c/vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+            "args": [
+                "-j16",
+                "-C",
+                "${workspaceFolder}/test/intltest"
+            ],
+            "type": "shell"            
+        },
+        {
+            "label": "runTestCases", // For excuting the test cases.
```
sffc: Suggestion: rename this to "intltest", and then add two more targets, "cintltst" and "iotest" for the other test suites.

docs/ide4c/vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+            "command": "${workspaceFolder}/test/intltest/intltest", // Could be any other shell command
+            "args": [
```
sffc: Please give an example for how you would specify a single test.  For example,

```js
"args": [
    // Specify a specific test case here; for example:
    // "format/NumberTest"
]
```

docs/ide4c/vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+                "all",
+                "--debug", // Enable debug mode.
+                "-j8" // Use 8 cores.
+            ]
+        },
+        {
+            "label": "buildTest", // for building test cases before running it.
+            "command": "make", // Could be any other shell command
+            "args": [
+                "-j16",
```
sffc: `-j16` is a bit much, no?  I think Andy found that `-j6` gave the best performance and anything more than `-j6` had rapidly diminishing returns.

docs/ide4c/vscode/tasks.json
```diff
@@ -0,0 +1,41 @@
...
+                "-C",
+                "${workspaceFolder}/test/intltest"
+            ],
+            "type": "shell"            
+        },
+        {
+            "label": "runTestCases", // For excuting the test cases.
+            "command": "${workspaceFolder}/test/intltest/intltest", // Could be any other shell command
```
sffc: Is there a way to make this command depend on a previous one like `buildTest`?

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,36 @@
...
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add wich test cases you want to run, leaving it empty will run all the test cases.
```
jefgen: very minor nit: typo (wich -> which)

jefgen:
Thanks for doing this @younies! :)

sffc:
Please also add the copyright headers link the Travis build is asking you to do.  For Markdown, you can use HTML-style comments.

docs/ide4c/vscode/tasks.json
younies: You are right, however, tests should be added as args

docs/ide4c/vscode/tasks.json
younies: Okay, I will make it -j6. However, we could not proof it is the best. For example, using Core i9 process may differ. Therefore, I will leave a comment states that is just the number of cores.

docs/ide4c/vscode/tasks.json
younies: Yes, in the launch, we request buildTests first

docs/ide4c/vscode/tasks.json
younies: This is in the launch task.

younies:
Sorry, I performed the squash early, I was testing something.

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+             "rbbi/RBBITest" // Just for example.
+            ],
+            "stopAtEntry": true, // If you want to add a break point in the begining of the program (i.e. test cases).
```
sffc: Why default to true instead of false?  It is kind-of confusing to step from the main function down into your test case.  I usually leave a breakpoint in my test case and let the code run until it gets there.

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add wich test cases you want to run, leaving it empty will run all the test cases.
+             "rbbi/RBBITest" // Just for example.
```
sffc: Inconsistent indentation

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+            "logging": {
+                "trace": true,
+                "traceResponse": true,
+                "engineLogging": true,
+                "programOutput": true
+            },
+            "preLaunchTask": "buildTest"
+        },
+        {
+            "name": "cintltest Launch",
```
sffc: it's `cintltst`, no "e"

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+            "MIMode": "gdb", // Use "lldb" for mac osx.
+            "logging": {
+                "trace": true,
+                "traceResponse": true,
+                "engineLogging": true,
+                "programOutput": true
+            },
+            "preLaunchTask": "buildTest"
+        }
+    ]
```
sffc: Do you want to add `iotest` as well?  (example argument: `/datadriv`)

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add wich test cases you want to run, leaving it empty will run all the test cases.
```
sffc: I find it useful to have an example default test arg, because I forget whether to start with a "/" or not (you need the leading "/" for cintltst but not intltest).  Here is one:

`"/tsutil/cstrcase"`

docs/ide4c/vscode/README.md
```diff
@@ -0,0 +1,7 @@
+# Configuring VS Code for ICU4C
+
+  - create `.vscode` folder in icu4c/source
```
markusicu: I prefer to work from one or even two folder levels higher up. For example, with config+make I always use out-of-source builds because I don't want to contaminate my source tree, and I often configure two or more variants (debug vs. release vs. hardcode-UTF-8...) in parallel into separate output folders.

Does this kind of thing make sense for VS Code?
Does it support out-of-source builds?

docs/ide4c/vscode/README.md
```diff
@@ -0,0 +1,7 @@
...
+  - Copy `tasks.json` and `launch.json` files into `.vscode` folder
+
+NOTE
+ before you build `icu4c` from vs code. run `./runConfigureICU` command first.
```
markusicu: Please specify whether this has to be an in-source build, or else how to tell VS Code about the relationship between the source and output folders.

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,66 @@
...
+    "configurations": [
+        {
+            "name": "intltest Launch", 
+            "type": "cppdbg",
+            "request": "launch",
+            "targetArchitecture": "x64",
+            "program": "${workspaceFolder}/test/intltest/intltest",
+            "environment": [
+                {
+                    "name": "LD_LIBRARY_PATH", // Use "DYLD_LIBRARY_PATH" for mac osx.
```
markusicu: The red highlighting here of // comments indicates that they are not supported in .json, right? Does VS Code handle them, or does one have to strip them while copying into the .vscode folder?

sffc: Vscode ignores // comments.

docs/ide4c/vscode/launch.json
younies: because the debugging can start even there is no break-points have been set. However, I agree with you. I will make the default  value false.

docs/ide4c/vscode/launch.json
younies: done!

docs/ide4c/vscode/launch.json
younies: Yes, this is the name in icu4c/source/test

docs/ide4c/vscode/launch.json
younies: Done!

docs/ide4c/vscode/launch.json
younies: Yes, vscode ignores them. and I think those comments are pretty useful when in this place.

docs/ide4c/vscode/launch.json
sffc: Except that you need the leading `/` in `/tsutil/cstrcase`

docs/ide4c/vscode/launch.json
sffc: Please change "cintltest launch" to "cintltst launch"

docs/ide4c/vscode/launch.json
sffc: Please add `iotest` unless there's a reason you don't want to include it

docs/ide4c/vscode/README.md
sffc: We should aim at supporting one "suggested" workflow with these pre-built IDE files; people can customize them if they would like to do so.

I personally prefer my workspace root to be icu4c/source, largely because it reduces 2 levels of indentation in the file selector tree in the left panel of vscode.

docs/ide4c/vscode/launch.json
younies: Done !

docs/ide4c/vscode/README.md
younies: Done ... I added a link to the page that talks about build/test ICU. There is no place in master/docs that explain the building process yet.

docs/ide4c/vscode/launch.json
sffc: Ping

docs/ide4c/vscode/launch.json
sffc: Ping

docs/ide4c/vscode/launch.json
sffc: Ping

docs/ide4c/vscode/launch.json
```diff
@@ -0,0 +1,67 @@
...
+                    "value": "${workspaceRoot}/lib:${workspaceRoot}/stubdata:${workspaceRoot}/tools/ctestfw"
+                },
+                {
+                    "name": "ICU_DATA",
+                    "value": "${workspaceRoot}/data/out"
+                }
+            ],
+            "args": [ // Here you add wich test cases you want to run, leaving it empty will run all the test cases.
+            	"tsutil/cstrcase" // Just for example.
+	    ],
```
sffc: Please fix indentation in this file (you have mixed tabs and spaces)

docs/ide4c/vscode/launch.json
younies: done

docs/ide4c/vscode/launch.json
younies: done!

sffc:
Approved.  @markusicu, if you have any more comments, can you share them soon so that we can get this 3-month-old PR checked in?  Thanks!

jefgen:
This is unrelated to your change... but the red syntax highlighting on the JSON file was bothering me in the diff view...
So after some searching, it turns out that GitHub actually does have a way to support JSON with comments in terms of the syntax highlighting.

It looks like you can add the following to the `.gitattributes` file:
`*.json linguist-language=jsonc`

This will cause the syntax highlighter to use the "jsonc" format, which supports comments.

I wonder if we'd want to maybe add this to the PR -- or perhaps I should file a separate ticket for it?

sffc:
> I wonder if we'd want to maybe add this to the PR -- or perhaps I should file a separate ticket for it?

New ticket sounds good.

jefgen:
Thanks!
I filed a new ticket, and created a new PR here:  https://github.com/unicode-org/icu/pull/969

PR #902, opened 2019-10-31
ICU-20884 initialized buffer uloc_getKeywordValue

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20884
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -766,6 +766,9 @@ uloc_getKeywordValue(const char* localeID,
                      char* buffer, int32_t bufferCapacity,
                      UErrorCode* status)
 {
+    if (buffer != nullptr) {
+        buffer[0] = '\0';
+    }
     const char* startSearchHere = NULL;
     const char* nextSeparator = NULL;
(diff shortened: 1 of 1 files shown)


sffc:
LGTM

Wait for reply from ICU-TC with regard to the 66 maintenance branch before merging.

FrankYFTang:
ping

sffc:
The right place to ping is on the thread on icu-core.

FrankYFTang:
I am very confused how the PR get all other files in here. Need some help to remove them.

jefgen:
> how the PR get all other files in here.

I think this is likely because the commit was originally made using a branch off of `master`, but then the target for the PR was changed to `maint/maint-66` afterwards.

One way to fix this would be to rebase the changes in your branch `ICU-20884` onto the `maint/maint-66` branch, and then force push.

sffc:
Yeah, I think at first the diff was the same on master and maint/maint-66, but the longer we wait the more likely they are to diverge.  The easiest thing would be to rebase your commit against the head of maint/maint-66.

PR #903, opened 2019-10-31
ICU-20418 Add concise number skeleton support.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20418
- [ ] Issue accepted
- [ ] Tests included

DISABLE_JIRA_ISSUE_MATCH=true

### DIFF
--- a/icu4c/source/common/stringpiece.cpp
+++ b/icu4c/source/common/stringpiece.cpp
@@ -51,6 +51,47 @@ void StringPiece::set(const char* str) {
     length_ = 0;
 }
 
+int32_t StringPiece::find(StringPiece needle, int32_t offset) {
+  if (length() == 0 && needle.length() == 0) {
+    return 0;
+  }
+  // TODO: Improve to be better than O(N^2)?
+  for (int32_t i = offset; i < length(); i++) {
+    int32_t j = 0;
+    for (; j < needle.length(); i++, j++) {
+      if (data()[i] != needle.data()[j]) {
+        i -= j;
+        goto outer_end;
+      }
+    }
+    return i - j;
+    outer_end: void();
+  }
+  return -1;
+}
+
+int32_t StringPiece::compare(StringPiece other) {
+  int32_t i = 0;
+  for (; i < length(); i++) {
+    if (i == other.length()) {
+      // this is longer
+      return 1;
+    }
+    char a = data()[i];
+    char b = other.data()[i];
+    if (a < b) {
+      return -1;
+    } else if (a > b) {
+      return 1;
+    }
+  }
+  if (i < other.length()) {
+    // other is longer
+    return -1;
+  }
+  return 0;
+}
+
 U_EXPORT UBool U_EXPORT2
 operator==(const StringPiece& x, const StringPiece& y) {
   int32_t len = x.size();
--- a/icu4c/source/i18n/measunit.cpp
+++ b/icu4c/source/i18n/measunit.cpp
@@ -1990,10 +1990,10 @@ MeasureUnit MeasureUnit::getTeaspoon() {
 // End generated code
 
 static int32_t binarySearch(
-        const char * const * array, int32_t start, int32_t end, const char * key) {
+        const char * const * array, int32_t start, int32_t end, StringPiece key) {
     while (start < end) {
         int32_t mid = (start + end) / 2;
-        int32_t cmp = uprv_strcmp(array[mid], key);
+        int32_t cmp = StringPiece(array[mid]).compare(key);
         if (cmp < 0) {
             start = mid + 1;
             continue;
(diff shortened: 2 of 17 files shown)


sffc:
@echeran Assigned to you to review.  I made three orthogonal commits to make it easier to review.  You can do one commit at a time.

sffc:
This is just the ICU4C implementation.  I will follow soon with the ICU4J.

icu4c/source/i18n/number_skeletons.cpp
```diff
@@ -564,6 +583,14 @@ skeleton::parseStem(const StringSegment& segment, const UCharsTrie& stemTrie, Se
         CHECK_NULL(seen, precision, status);
             blueprint_helpers::parseDigitsStem(segment, macros, status);
             return STATE_NULL;
+        case u'E':
+        CHECK_NULL(seen, notation, status);
+            blueprint_helpers::parseScientificStem(segment, macros, status);
+            return STATE_NULL;
+        case u'0':
+        CHECK_NULL(seen, notation, status);
```
sffc: This should check `integerWidth`, not `notation`

icu4c/source/test/intltest/numbertest.h
```diff
@@ -114,16 +114,38 @@ class NumberFormatterApiTest : public IntlTestWithFieldPosition {
...
-    void assertFormatDescending(const char16_t* message, const char16_t* skeleton,
-                                const UnlocalizedNumberFormatter& f, Locale locale, ...);
+    /**
+     * skeleton is the full length skeleton, which must round-trip.
+     *
+     * conciseSkeleton should be the shortest available skeleton.
+     * The concise skeleton can be read but not printed.
+     */
+    void assertFormatDescending(
+      const char16_t* message,
```
echeran: What is the reason/benefit of using char16_t* here instead of StringPiece?  Or are they not designed to be compatible?

icu4c/source/i18n/measunit.cpp
```diff
@@ -2138,6 +2138,53 @@ int32_t MeasureUnit::internalGetIndexForTypeAndSubtype(const char *type, const c
...
+
+    // If not found, try breaking apart numerator and denominator
+    int32_t perIdx = coreUnitIdentifier.find("-per-", 0);
+    if (perIdx == -1) {
+        // String does not contain "-per-"
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return false;
+    }
+    StringPiece numeratorStr(coreUnitIdentifier, 0, perIdx);
+    StringPiece denominatorStr(coreUnitIdentifier, perIdx + 5);
```
echeran: Is the length of the denominator name always guaranteed to be 5 characters long?  Is it worth declaring a constant for this?

icu4c/source/common/stringpiece.cpp
```diff
@@ -51,6 +51,47 @@ void StringPiece::set(const char* str) {
     length_ = 0;
 }
 
+int32_t StringPiece::find(StringPiece needle, int32_t offset) {
+  if (length() == 0 && needle.length() == 0) {
+    return 0;
+  }
+  // TODO: Improve to be better than O(N^2)?
```
echeran: I looked up in a reference book to see what it said.  For sufficiently short strings, the O(mn) algorithm is hard to beat in practice.  For lengths < 5, it's near impossible.  Only for sufficiently long strings does the Knuth-Morris-Pratt algorithm of intelligently advancing i make a difference (https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm)

icu4c/source/i18n/number_skeletons.cpp
```diff
@@ -564,6 +583,14 @@ skeleton::parseStem(const StringSegment& segment, const UCharsTrie& stemTrie, Se
         CHECK_NULL(seen, precision, status);
             blueprint_helpers::parseDigitsStem(segment, macros, status);
             return STATE_NULL;
+        case u'E':
```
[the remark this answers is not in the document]
sffc: I don't exactly remember why CHECK_NULL is not indented.  I think it was my IDE's fault.  This weird indentation happens all over this file and I don't really want to change it and increase git blame size.

icu4c/source/test/intltest/numbertest.h
sffc: StringPiece is an 8-bit char string.  UnicodeString is a 16-bit char16_t string.  I could make the argument a UnicodeString, but the style of implicitly converting a raw string to a UnicodeString is somewhat frowned upon.  Generally I like to pass raw strings until I actually need to do work on them.

icu4c/source/i18n/number_skeletons.cpp
sffc: Done

icu4c/source/common/stringpiece.cpp
sffc: Acknowledged

sffc:
Thanks.  Ported to Java.  PTAL

sffc:
I would like to check this in as the units work depends on this PR.  Can you review the Java port please?

PR #906, opened 2019-11-04
ICU-20693 Fix erroneous addition of cldrVersion

hagbard:
ICU-20693 The cldrVersion value is added as a bit of a hack in the supplementalData.txt file (because that data doesn't come from CLDR XML directly). However it should only go into the "main" misc/supplementalData.txt, not the curr/supplementalData.txt file.
This fix just sets the flag to add it to false for the curr/ version.

This can be seen via the diffs for the new tools:

    diff -r -I '^[ /]\*' /icu-65-1/icu4c/source/data/curr/supplementalData.txt /tmp/icu65/curr/supplementalData.txt
    ...
    5207a5244
    >     cldrVersion{"36"}

##### Checklist

- [x] Issue accepted


srl295:
lgtm but pending CI

PR #910, opened 2019-11-06
ICU-20693 Reworking Ant structure to better explain and reflect 'tailorings'

hagbard:
This restructures the build-icu-data.xml Ant build file a bit and adds more support to manage the additional mechanisms needed to support the special case rewriting of locale parents in certain ICU directories.

Once the build system is changed to use the dependency graph information produced by the conversion tool (rather than the files produced by the old system), this PR should provide the last step in the chain to create 100% compatible ICU output.

##### Checklist

- [x] Issue accepted


hagbard:
Only look at the changes in:
https://github.com/unicode-org/icu/pull/910/commits/452e5258826f21d122634ddf8a4b2e15413ad75f
since anything before that are the previous PRs I've already sent out.

sffc:
LGTM2

PR #911, opened 2019-11-06
ICU-20693 Adding support for generating ICU data files directly into release directories

hagbard:
Adding a new Ant task to handle tidying up of the release directories before writing the new ICU data files. Note that this approach is NOT what I've documented in the workflow proposal document, but it should be good enough to begin testing the data for the next release. We can iterate on this later (possibly after the 66 or even 67 releases).

The new task is paranoid and checked files it might delete for an exactly matching heading. Later I intend to modify the header to include a line unique to files generated by LdmlConverter, which should prevent any risk of accidental unwanted file deletion.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20693
- [X] Issue accepted


hagbard:
Only look at diffs in:
https://github.com/unicode-org/icu/pull/911/commits/6b8607c5076967cee451b57ed00ebebe3c20a2d6
Everything else is other PRs that this work is based on.

PR #912, opened 2019-11-06
ICU-20693 Fix alt values to support per-locale mappings and remove source paths.

hagbard:
This should now match the desired behaviour used by Google to remap alt-values for paths.
You can now specify a list of locale IDs for an <altPath> element and any source paths are removed from the filtered data.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20693
- [X] Issue accepted


hagbard:
Only look at diffs in:
https://github.com/unicode-org/icu/pull/912/commits/c2915b7b022c305cdc24cc5b9ebbd129deac8c8d
The other stuff is a different (pending) PR.

PR #913, opened 2019-11-07
ICU-20693 Fixing how forced parent IDs work

hagbard:
The current code doesn't quite handle forced parent IDs correctly, and allows something like:

    "zh_Hant"  -- forced parent --> "zh"

to write the %%Parent value of "zh". However when the parent is the "truncated" locale ID, it should not be written. This PR fixes that (tested locally with a full conversion which definitely removed the issue).

I also fixed a trivial issue to prevent the supplementalData file in "curr/" from having a cldrVersion field.

Latest diffs from this PR can be found at:
https://docs.google.com/spreadsheets/d/10UqiRFEXVgFygSchF7ZhWcz5pC0_gqAE5sgPGIB36iM/edit
see the new diffs at the bottom of each directory tab.

##### Checklist

- [x] Issue accepted

### DIFF
--- a/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
+++ b/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
@@ -326,12 +326,10 @@ public final class LdmlConverter {
                 // If we add an explicit parent locale, it forces the data to be written. This is
                 // where we check for forced overrides of the parent relationship (which is a per
                 // directory thing).
-                parent
-                    .map(p -> config.getForcedParents(dir).getOrDefault(id, p))
-                    .ifPresent(p -> {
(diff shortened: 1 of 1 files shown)


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -442,7 +459,7 @@ private void processSupplemental() {
                 break;
 
             case CURRENCY_DATA:
-                processSupplemental("supplementalData", CURRENCY_DATA_PATHS, "curr", true);
+                processSupplemental("supplementalData", CURRENCY_DATA_PATHS, "curr", false);
```
hagbard: this is the fix to stop the cldrVersion being written (now matches the current data better).

hagbard:
If you could merge this yourself if you're happy with it, that'd be great.

PR #914, opened 2019-11-07
ICU-20739 Force seconds if the skeleton has fractional seconds

mihnita:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20739
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
@@ -2649,6 +2649,25 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
                 if (subField > 0) subField += value.length();
                 type[field] = subField;
             }
+
+            // #20739, we have a skeleton with milliseconde, but no seconds
+            if (!original.isFieldEmpty(FRACTIONAL_SECOND) && original.isFieldEmpty(SECOND)) {
+                // Force the use of seconds
+                for (int i = 0; i < types.length; ++i) {
+                    int[] row = types[i];
+                    if (row[1] == SECOND) {
+                        // first entry for SECOND
+                        original.populate(SECOND, (char)row[0], row[3]);
+                        baseOriginal.populate(SECOND, (char)row[0], row[3]);
+                        // We add value.length, same as above, when type is first initialized.
+                        // The value we want to "fake" here is "s", and 1 means "s".length()
+                        int subField = row[2];
+                        type[SECOND] = (subField > 0) ? subField + 1 : subField;
+                        break;
+                    }
+                }
+            }
+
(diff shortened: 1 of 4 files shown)


icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -2649,6 +2649,25 @@ DateTimeMatcher set(String pattern, FormatParser fp, boolean allowDuplicateField
...
+                // Force the use of seconds
+                for (int i = 0; i < types.length; ++i) {
+                    int[] row = types[i];
+                    if (row[1] == SECOND) {
+                        // first entry for SECOND
+                        original.populate(SECOND, (char)row[0], (char)row[3]);
+                        baseOriginal.populate(SECOND, (char)row[0], (char)row[3]);
+                        // We add value.length, same as above, when type is first initialized.
+                        // The value we want to "fake" here is "s", and 1 means "s".length()
+                        int subField = row[2];
```
sffc: According to the `types` table at the bottom of the file, `row[2]` is `NUMERIC`, which is defined as `0x100`, so I think you can just assign `type[SECOND] = row[2]` directly, like they're doing in the next loop with `DAYPERIOD`.  Also makes it more consistent with C++.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -2649,6 +2649,25 @@ DateTimeMatcher set(String pattern, FormatParser fp, boolean allowDuplicateField
...
             }
+
+            // #20739, we have a skeleton with milliseconde, but no seconds
+            if (!original.isFieldEmpty(FRACTIONAL_SECOND) && original.isFieldEmpty(SECOND)) {
+                // Force the use of seconds
+                for (int i = 0; i < types.length; ++i) {
+                    int[] row = types[i];
+                    if (row[1] == SECOND) {
+                        // first entry for SECOND
+                        original.populate(SECOND, (char)row[0], (char)row[3]);
```
sffc: I wouldn't nitpick about this if I weren't already leaving another comment, but you only need to cast `row[0]`, not `row[3]`.

mihnita: I'll fix it if you think I should fix the first one.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
mihnita: Made C++ consistent with Java (and itself)
Without adding that length the "mS" Java the unit tests fails.

sffc:
LGTM please squash

PR #915, opened 2019-11-08
ICU-20893 Unicode 13 beta

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20893


markusicu:
C++ is usable. Test failures in intltest spoof & rbbi.
Java TBD.

markusicu:
Ready to go, unless the test bots have failures.

Except: We still need to sync the root line break rules into the various tailorings. That will go into a follow-up PR.

PR #917, opened 2019-11-12
ICU-20844 reduce restriction on minInt=minFrac=0

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20844
- [x] Issue accepted
- [x] Tests included

- In NumberPropertyMapper::oldToNew, remove restriction against minInt=minFrac=0 (only if maxInt != 0)
- Ensure DecimalFormat::doFastFormatInt32 always produces at least 1 digit

This does not break any existing tests but does fix the problem demonstrated by the newly added test for this.

This PR is just for C. If the approach seems OK, I will create a separate PR for Java.

### DIFF
--- a/icu4c/source/i18n/number_mapper.cpp
+++ b/icu4c/source/i18n/number_mapper.cpp
@@ -125,10 +125,8 @@ MacroProps NumberPropertyMapper::oldToNew(const DecimalFormatProperties& propert
     }
     // Validate min/max int/frac.
     // For backwards compatibility, minimum overrides maximum if the two conflict.
-    // The following logic ensures that there is always a minimum of at least one digit.
     if (minInt == 0 && maxFrac != 0) {
-        // Force a digit after the decimal point.
-        minFrac = minFrac <= 0 ? 1 : minFrac;
+        minFrac = (minFrac < 0 || (minFrac == 0 && maxInt == 0)) ? 1 : minFrac;
         maxFrac = maxFrac < 0 ? -1 : maxFrac < minFrac ? minFrac : maxFrac;
         minInt = 0;
         maxInt = maxInt < 0 ? -1 : maxInt > kMaxIntFracSig ? -1 : maxInt;
--- a/icu4c/source/i18n/decimfmt.cpp
+++ b/icu4c/source/i18n/decimfmt.cpp
@@ -1834,7 +1834,8 @@ void DecimalFormat::doFastFormatInt32(int32_t input, bool isNegative, UnicodeStr
     char16_t localBuffer[localCapacity];
     char16_t* ptr = localBuffer + localCapacity;
     int8_t group = 0;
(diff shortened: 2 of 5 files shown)


pedberg-icu:
OK, I added a test that pattern contains "#.#" after setting minInt=minFrac=0.

I also added tests for currency formatting; then to make them (slow path) work for the 0 case without breaking anything else, I needed to add a field minIntFracZero in MacroProps and MicroProps that remembers the request for minInt=minFrac=0, and in that case forces NumberFormatterImpl::writeNumber to output a digit for 0 if it would otherwise not output anything. There may be a better way to do this.

sffc:
> I needed to add a field minIntFracZero in MacroProps and MicroProps that remembers the request for minInt=minFrac=0, and in that case forces NumberFormatterImpl::writeNumber to output a digit for 0 if it would otherwise not output anything.

I think this is probably why I put the restriction there in the first place.  However, I agree with your use case that this is something we should support, especially since we apparently used to support it, so we might need to take a small hit on code elegance.

pedberg-icu:
I will go ahead and squash this.

sffc:
I actually do have review feedback, I just haven't posted it yet.

I have some changes I'm testing locally.  I based it on the commit before you squashed.  Is it OK if I re-force-push the original commits to your branch?

pedberg-icu:
> Is it OK if I re-force-push the original commits to your branch?

Sure, sorry for the premature squash.

sffc:
OK, I re-pushed your first two commits and added a new one, https://github.com/unicode-org/icu/pull/917/commits/3bbed2156d7c662f1d2c3fe558958049f409279d

I proposed changing the test expectation in NumberFormatterApiTest.  I don't think this will be a controversial change since it only affects the ICU 60 API.  If we do that, the code can be simplified because we don't need the extra field in the macros/micros anymore.

pedberg-icu:
> I proposed changing the test expectation in NumberFormatterApiTest...

That seems like a good change to me, that test expectation in NumberFormatterApiTest seemed a little odd. I will re-squash the commits.

icu4c/source/i18n/number_mapper.cpp
```diff
@@ -127,8 +127,7 @@ MacroProps NumberPropertyMapper::oldToNew(const DecimalFormatProperties& propert
     // For backwards compatibility, minimum overrides maximum if the two conflict.
     // The following logic ensures that there is always a minimum of at least one digit.
```
sffc: Please delete this line because this comment is no longer true.

pedberg-icu: OK, removed

pedberg-icu:
> lgtm in ICU4C except for one small thing

OK, I removed the now-incorrect comment

> Do you plan to make this change in ICU4J, too?

Yes, under a different PR

PR #921, opened 2019-11-16
ICU-20890 ICU-20861 Refactoring locale_dependencies.py into JSON files

sffc:
This also fixes ICU-20861

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20890
- [x] Issue accepted
- [ ] Tests included


sffc:
Commentary:

Although it's unlikely that we will truly "bazelfy" this script in the near future, I kept the same style of keeping all I/O operations outside of library code.  The only I/O operations are coded in `__main__.py` and the test file, making them easy to swap out.

I went ahead and deleted the old locale_dependencies.py and copied it to all the JSON directories verbatim, except that I did update the zh_HK alias in rbnf.  If someone generates data with the *old* tool, they will have to manually perform that copy again.  If it is necessary, I can document that in cldr-icu-readme.txt.  I do not want to do anything more than that to help users of the old tool because the expectation is that the old tool is going away.

icu4c/source/python/icutools/databuilder/__main__.py
```diff
@@ -237,6 +238,25 @@ def add_copy_input_requests(requests, config, common_vars):
...
+class IO(object):
+    """I/O operations required when computing the build actions"""
+
+    def __init__(self, src_dir):
+        self.src_dir = src_dir
+
+    def glob(self, pattern):
+        result_paths = pyglob.glob(os.path.join(self.src_dir, pattern))
+        # For the purposes of icutools.databuilder, force Unix-style directory separators.
+        return [v.replace("\\", "/")[len(self.src_dir)+1:] for v in sorted(result_paths)]
```
hagbard: Is there no "proper" way to handle Windows style directory paths in Python?
I see you appear to be trying to limit the '\' replacement to path segments up until the src_dir length, or something.
Is this still safe in the face of a Unix file with a '\' in?
Perhaps a comment explaining the approach would be good (or is it simpler to convert src_dir first since no patterns should have '\\' in)?

icu4c/source/python/icutools/databuilder/filtration.py
```diff
@@ -56,6 +56,13 @@ def _file_to_file_stem(cls, file):
         limit = file.filename.rfind(".")
         return file.filename[start+1:limit]
 
+    @classmethod
+    def _file_to_subdir(cls, file):
```
hagbard: This is where my Python lets me down, but it seems odd to have a class method requiring an instance that doesn't use it (I guess it's copied from _file_to_file_stem. Also weird that there's no idiomatic util for this sort of thing.

icu4c/source/python/icutools/databuilder/filtration.py
```diff
@@ -151,83 +158,86 @@ def match(self, file):
...
+    def _get_parent_locale(self, locale, tree):
+        """Gets the parent locale in the given tree, according to dependency data."""
+        dependency_data = self.dependency_data_by_tree[tree]
+        if locale in dependency_data["parents"]:
+            return dependency_data["parents"][locale]
+        if locale in dependency_data["aliases"]:
+            return dependency_data["aliases"][locale]
         if LANGUAGE_ONLY_REGEX.match(locale):
             return "root"
         i = locale.rfind("_")
```
hagbard: Isn't it true at this point that it's a hard failure to not have 'root' here? Everything else should have been dealt with.
Then the comment on this can say "only returns None if the input was "root".

icu4c/source/python/icutools/databuilder/test/filtration_test.py
```diff
@@ -1,9 +1,13 @@
...
+import io as pyio
+import json
+import os
 import unittest
 
 from .. import InFile
+from ..comment_stripper import CommentStripper
 from ..filtration import Filter
 
 EXAMPLE_FILE_STEMS = [
```
hagbard: No variants tested here ... worth adding?

icu4c/source/python/icutools/databuilder/test/filtration_test.py
```diff
@@ -60,12 +65,33 @@
     "zh"
 ]
 
+
+class TestIO(object):
+    def __init__(self):
+        pass
+
+    def read_locale_deps(self, tree):
```
hagbard: Is this a golden file test running with real data? If so, please raise a ticket to not make it do that.

icu4c/source/python/icutools/databuilder/filtration.py
hagbard: Do you mean in this PR or another. Feel free to do it in another since otherwise it messes up my approval (and you might want to merge today).

sffc: I already made the `@staticmethod` change.

sffc:
Before I merge I have to:

1. Update the LOCALE_DEPS.json files
2. Squash to one commit

hagbard:
Ok, feel free to ask Markus or someone there to re-approve after any changes.
I don't think you want to be updating the json files to my output though, not until we switch to using this for the release. Keeping current behaviour unchanged (and self consistent) would seem to be the right thing at the moment.

sffc:
> I don't think you want to be updating the json files to my output though, not until we switch to using this for the release. Keeping current behaviour unchanged (and self consistent) would seem to be the right thing at the moment.

I would normally agree, but the problem is that the current behavior is wrong, for reasons described in [ICU-20861](https://unicode-org.atlassian.net/browse/ICU-20861), and your tool fixes it.  So I would rather check in the JSON files generated from your tool instead of copying the locale_dependencies.py file generated from the old tool.

sffc:
Done, the new LOCALE_DEPS.json files are in the PR.  PTAL

sffc:
I fixed a build error, merged from maint, and squashed.  The version numbers have been updated to CLDR 36.1.  PTAL

sffc:
Thanks.  I cleared it by Markus and he's also onboard with pulling these JSON files from your tool "early".

PR #926, opened 2019-11-26
ICU-20907 Disable optimization for MSVC on ARM64 on versions below 16.4

jefgen:
The Visual Studio/MSVC optimizer has trouble with these functions when building for Windows ARM64 on versions below 16.4.
(Thanks to @marswe for help with this issue!)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20907
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/normalizer2impl.cpp
+++ b/icu4c/source/common/normalizer2impl.cpp
@@ -2088,6 +2088,13 @@ uint8_t Normalizer2Impl::getPreviousTrailCC(const uint8_t *start, const uint8_t 
 // minDecompNoCP etc. and smallFCD[] are intended to help with any loss of performance,
 // at least for ASCII & CJK.
 
+// Ticket 20907 - The optimizer in MSVC/Visual Studio versions below 16.4 has trouble with this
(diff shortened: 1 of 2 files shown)


sffc:
Please target this against maint/maint-66 if you want this to go into ICU 66.

icu4c/source/common/normalizer2impl.cpp
```diff
@@ -2088,6 +2088,10 @@ uint8_t Normalizer2Impl::getPreviousTrailCC(const uint8_t *start, const uint8_t
 // minDecompNoCP etc. and smallFCD[] are intended to help with any loss of performance,
 // at least for ASCII & CJK.
 
+// Ticket 20907 - Disable optimization on Windows when building for ARM64 with Visual Studio versions below 16.4.
```
sffc: In this comment, could you add an approximate timeframe during which this workaround is required, such that if someone comes around and refactors/rewrites this function, they know when they can delete it?

jefgen: Thanks! That's a good idea. I'll expand the comment and force push.

PR #927, opened 2019-11-28
ICU-20857 Update Readme for ICU 66 Preview.

jefgen:
#### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20857
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/readme.html
+++ b/icu4j/readme.html
@@ -15,20 +15,23 @@ h3.doc { text-decoration: underline }
  link="#0000ff" vlink="#800080">
 <h1>International Components for Unicode for Java (ICU4J)</h1>
 <h2>Read Me for ICU4J 66.0.1</h2>
-(Last Update: 2019-Oct-21)
+(Last Update: 2019-Nov-27)
 <hr size="2" width="100%">
 
 <p>
 <!-- <b>Note:</b> This is major release of ICU4J. It contains bug fixes and adds implementations
 of inherited API and introduces new API or functionality. -->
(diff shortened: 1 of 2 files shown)


icu4j/readme.html
```diff
@@ -15,20 +15,23 @@
...
+(Last Update: 2019-Nov-27)
 <hr size="2" width="100%">
 
 <p>
 <!-- <b>Note:</b> This is major release of ICU4J. It contains bug fixes and adds implementations
 of inherited API and introduces new API or functionality. -->
+<b>Note:</b> This is a preview release of ICU4J 66.
+The contents of this document may not reflect the recent changes done
+for ICU 66 development. It is not recommended for production use.
 <!-- <b>Note:</b> This is a development milestone of ICU4J 66.
```
daniel-ju: nit: Should we replace the "milestone" comment with "preview" (similar to what you did in the icu4c readme)?

jefgen: Hm, I'm confused. I created a new comment that for "preview" that doesn't mention "milestone" at all.. so I'm not sure what you mean by this?

daniel-ju: Right, in this icu4j file you created a new comment, but in the icu4c file you removed the "milestone" comment and replaced it with the "preview" comment. I was just wondering if the two should be kept consistent.

jefgen: Ah, thanks! Sure, I guess it would make sense to have them both be consistent. I'll push an update in a bit.

PR #931, opened 2019-12-03
ICU-20857 BRS66 Updated ICU4J API change report

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20857
- [x] Issue accepted
- [ ] Tests included


icu4j/APIChangeReport.html
```diff
@@ -243,19 +78,13 @@ <h3>Package com.ibm.icu.util</h3>
 <li><span style='color:orange'>(draft)</span> public static final MeasureUnit PIXEL_PER_INCH</li>
```
markusicu: These should no longer show up in the "Added in ICU4J 66.0.1" section, right?

PR #934, opened 2019-12-04
ICU-20873 Add the PGP "KEYS" file to the ICU repo.

jefgen:
This is a copy of the file from: http://ssl.icu-project.org/KEYS

The ICU project's PGP KEYS file was previously hosted on a separate server that not all ICU-TC members have access too. This change copies the current KEYS file into the top-level git repo, so that we can setup a redirect on the website to point at the checked-in file, rather than hosting it separately.
(As an added bonus, this will also give us version history on the file as well).

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20873
- [x] Issue accepted
- [ ] Tests included

ALLOW_MANY_COMMITS=true


jefgen:
I was going to do that as a separate PR/commit, so we'd have a clean commit of the file as-is and then the addition separately. (But I could just add my key in the same PR as two commits I guess -- and not squash..?)

srl295:
added `ALLOW_MANY_COMMITS=true` to the description.

PR #939, opened 2019-12-10
ICU-20741 Changing SimpleDateTimeFormat::subFormat to avoid duplicated field for same position

caiolima:
There is a fallback mechanism on fields `b` and `B` to field `a` when data is not provided for a locale. This fallback is done calling `subFormat` recursively. However, when such function finishes execution, caller still executes its remaining operations, causing duplication of fields `a` and `B`  to the same position. This PR is changing the behavior for this function to return as soon as the fallback calls returns. Since we want to keep results consistent when a data change on CLDR happen, the field included  on iterator is the one requested by the pattern.

@sffc

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -996,7 +996,8 @@ SimpleDateFormat::_format(Calendar& cal, UnicodeString& appendTo,
         // Use subFormat() to format a repeated pattern character
         // when a different pattern or non-pattern character is seen
         if (ch != prevCh && count > 0) {
-            subFormat(appendTo, prevCh, count, capitalizationContext, fieldNum++, handler, *workCal, status);
+            subFormat(appendTo, prevCh, count, capitalizationContext, fieldNum++,
+                      prevCh, handler, *workCal, status);
             count = 0;
         }
         if (ch == QUOTE) {
@@ -1023,7 +1024,8 @@ SimpleDateFormat::_format(Calendar& cal, UnicodeString& appendTo,
 
     // Format the last item in the pattern, if any
     if (count > 0) {
-        subFormat(appendTo, prevCh, count, capitalizationContext, fieldNum++, handler, *workCal, status);
+        subFormat(appendTo, prevCh, count, capitalizationContext, fieldNum++,
+                  prevCh, handler, *workCal, status);
     }
 
     if (calClone != NULL) {
@@ -1402,10 +1404,11 @@ SimpleDateFormat::processOverrideString(const Locale &locale, const UnicodeStrin
 //---------------------------------------------------------------------
 void
 SimpleDateFormat::subFormat(UnicodeString &appendTo,
-                            UChar ch,
+                            char16_t ch,
                             int32_t count,
                             UDisplayContext capitalizationContext,
                             int32_t fieldNum,
+                            char16_t fieldToOutput,
                             FieldPositionHandler& handler,
                             Calendar& cal,
                             UErrorCode& status) const
@@ -1853,8 +1856,11 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
         // In either case, fall back to am/pm.
         if (toAppend == NULL || toAppend->isBogus()) {
             // Reformat with identical arguments except ch, now changed to 'a'.
-            subFormat(appendTo, 0x61, count, capitalizationContext, fieldNum,
-                      handler, cal, status);
+            // We are passing a different fieldToOutput because we want to add
+            // 'b' to field position. This makes this fallback stable when
+            // there is a data change on locales.
+            subFormat(appendTo, u'a', count, capitalizationContext, fieldNum, u'b', handler, cal, status);
+            return;
         } else {
             appendTo += *toAppend;
         }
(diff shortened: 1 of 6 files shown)


icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -1854,7 +1857,9 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
         if (toAppend == NULL || toAppend->isBogus()) {
             // Reformat with identical arguments except ch, now changed to 'a'.
             subFormat(appendTo, 0x61, count, capitalizationContext, fieldNum,
-                      handler, cal, status);
+                      fgPatternIndexToDateFormatField[patternCharIndex], handler,
```
sffc: More clear to specify what this field is

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -1990,7 +1998,7 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
     }
 #endif
 
-    handler.addAttribute(fgPatternIndexToDateFormatField[patternCharIndex], beginOffset, appendTo.length());
+    handler.addAttribute(fgPatternIndexToDateFormatField[fieldToOutput], beginOffset, appendTo.length());
```
sffc: Pass fieldToOutput directly?

icu4c/source/test/intltest/dtfmttst.cpp
```diff
@@ -5579,34 +5579,40 @@ void DateFormatTest::Test20741_ABFields() {
...
+            dateFormat.format(*calendar, result, &fpositer, status);
+
+            FieldPosition curFieldPosition;
+            FieldPosition lastFieldPosition;
+            lastFieldPosition.setBeginIndex(-1);
+            lastFieldPosition.setEndIndex(-1);
+            while(fpositer.next(curFieldPosition)) {
+                assertFalse("Field missing on pattern", pattern.indexOf(PATTERN_CHARS[curFieldPosition.getField()]) == -1);
+                if (curFieldPosition.getBeginIndex() == lastFieldPosition.getBeginIndex() && curFieldPosition.getEndIndex() == lastFieldPosition.getEndIndex()) {
+                    if (logKnownIssue("20741")) continue;
```
sffc: Remove the logKnownIssue if the bug is fixed

icu4j/main/classes/core/src/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -1568,6 +1568,20 @@ protected String subFormat(char ch, int count, int beginOffset,
         return buf.toString();
     }
 
+    /**
+     * Formats a single field; It is adding parameter ch as patternCharToOutput.
+     *
+     * @internal
+     * @deprecated This API is ICU internal only.
+     */
+    protected void subFormat(StringBuffer buf,
```
sffc: Try to avoid this overload.

icu4j/main/classes/core/src/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -1944,7 +1959,7 @@ protected void subFormat(StringBuffer buf,
             if (toAppend == null) {
                 // Time isn't exactly midnight or noon (as displayed) or localized string doesn't
                 // exist for requested period. Fall back to am/pm instead.
-                subFormat(buf, 'a', count, beginOffset, fieldNum, capitalizationContext, pos, cal);
+                subFormat(buf, 'a', count, beginOffset, fieldNum, capitalizationContext, pos, 'b', cal);
```
sffc: Add a comment pointing to the bug explaining why we have a different display field from data field.  (here and elsewhere)

icu4c/source/i18n/smpdtfmt.cpp
caiolima: Done.

icu4c/source/i18n/smpdtfmt.cpp
caiolima: Oops. You are right.

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -1408,7 +1408,7 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
                             int32_t count,
                             UDisplayContext capitalizationContext,
                             int32_t fieldNum,
-                            UDateFormatField fieldToOutput,
+                            UChar fieldToOutput,
```
sffc: Please make this and the other argument char16_t

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/DateFormatTest.java
```diff
@@ -5467,7 +5467,7 @@ public void test20741_ABFields() {
         for (String skeleton : skeletons) {
             for (int i = 0; i < locales.length; i++) {
                 ULocale locale = locales[i];
-                if (isQuick() && (i % 17 != 0)) continue;
+                // if (isQuick() && (i % 17 != 0)) continue;
```
sffc: Please add this back

PR #940, opened 2019-12-11
ICU-20906 Annotate APIs in PluralRules with why they are @internal

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20906
- [x] Issue accepted
- [ ] Tests included


sffc:
I checked all of the `@internal` APIs in PluralRules.  Most of them are used by CLDR, a few by ICU tests, and a few that I could actually make private (or package-private).

IMO, most of these classes should be moved to com.ibm.icu.impl, but that would be a more disruptive change that I'm not interested in tackling.

PR #944, opened 2019-12-13
ICU-20442 Adding support for hour-cycle on DateTimePatternGenerator

caiolima:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20707
- [x] Issue accepted
- [x] Tests included

DateTimePatternGenerator needs to consider the hour-cycle preferred by
Locale. This means that we need to to override the hour-cycle when a
locale contains "hc" keyword. This patch is adding such functionality.
In addition, "DateTimePatternGenerator::adjustFieldTypes" should adjust
hour field to properly follow tr35
spec(https://www.unicode.org/reports/tr35/tr35-dates.html#dfst-hour).

@sffc

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -654,6 +654,23 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
 
     int32_t* allowedFormats = getAllowedHourFormatsLangCountry(language, country, status);
 
+    // We need to check if there is an hour cycle on locale
+    char buffer[8];
+    int32_t count = locale.getKeywordValue("hours", buffer, sizeof(buffer), status);
+
+    fDefaultHourFormatChar = 0;
+    if (U_SUCCESS(status) && count > 0) {
+        if(uprv_strcmp(buffer, "h24") == 0) {
+            fDefaultHourFormatChar = LOW_K;
+        } else if(uprv_strcmp(buffer, "h23") == 0) {
+            fDefaultHourFormatChar = CAP_H;
+        } else if(uprv_strcmp(buffer, "h12") == 0) {
+            fDefaultHourFormatChar = LOW_H;
+        } else if(uprv_strcmp(buffer, "h11") == 0) {
+            fDefaultHourFormatChar = CAP_K;
+        }
+    }
+
     // Check if the region has an alias
     if (allowedFormats == nullptr) {
         UErrorCode localStatus = U_ZERO_ERROR;
@@ -667,13 +684,16 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
     if (allowedFormats != nullptr) {  // Lookup is successful
         // Here allowedFormats points to a list consisting of key for preferredFormat,
         // followed by one or more keys for allowedFormats, then followed by ALLOWED_HOUR_FORMAT_UNKNOWN.
-        switch (allowedFormats[0]) {
-            case ALLOWED_HOUR_FORMAT_h: fDefaultHourFormatChar = LOW_H; break;
-            case ALLOWED_HOUR_FORMAT_H: fDefaultHourFormatChar = CAP_H; break;
-            case ALLOWED_HOUR_FORMAT_K: fDefaultHourFormatChar = CAP_K; break;
-            case ALLOWED_HOUR_FORMAT_k: fDefaultHourFormatChar = LOW_K; break;
-            default: fDefaultHourFormatChar = CAP_H; break;
+        if (!fDefaultHourFormatChar) {
+            switch (allowedFormats[0]) {
+                case ALLOWED_HOUR_FORMAT_h: fDefaultHourFormatChar = LOW_H; break;
+                case ALLOWED_HOUR_FORMAT_H: fDefaultHourFormatChar = CAP_H; break;
+                case ALLOWED_HOUR_FORMAT_K: fDefaultHourFormatChar = CAP_K; break;
+                case ALLOWED_HOUR_FORMAT_k: fDefaultHourFormatChar = LOW_K; break;
+                default: fDefaultHourFormatChar = CAP_H; break;
+            }
         }
+
         for (int32_t i = 0; i < UPRV_LENGTHOF(fAllowedHourFormats); ++i) {
             fAllowedHourFormats[i] = allowedFormats[i + 1];
             if (fAllowedHourFormats[i] == ALLOWED_HOUR_FORMAT_UNKNOWN) {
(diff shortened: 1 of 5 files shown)


icu4c/source/i18n/dtptngen.cpp
```diff
@@ -654,6 +654,23 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
 
     int32_t* allowedFormats = getAllowedHourFormatsLangCountry(language, country, status);
 
+    // We need to check if there is an hour cycle on locale
+    char buffer[8];
+    int32_t count = locale.getKeywordValue("hours", buffer, sizeof(buffer), status);
```
sffc: You should check `status` after this call

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -667,21 +684,28 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
...
+                default: fDefaultHourFormatChar = CAP_H; break;
+            }
+            allowedIndex = 1;
+        } else {
+            fDefaultHourFormatChar = defaultHourFormatCharFromLocale;
         }
+
         for (int32_t i = 0; i < UPRV_LENGTHOF(fAllowedHourFormats); ++i) {
-            fAllowedHourFormats[i] = allowedFormats[i + 1];
+            fAllowedHourFormats[i] = allowedFormats[allowedIndex++];
```
sffc: Why do you need the `++` ?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -654,6 +654,23 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
 
     int32_t* allowedFormats = getAllowedHourFormatsLangCountry(language, country, status);
 
+    // We need to check if there is an hour cycle on locale
+    char buffer[8];
+    int32_t count = locale.getKeywordValue("hours", buffer, sizeof(buffer), status);
+
+    char16_t defaultHourFormatCharFromLocale = 0;
```
sffc: Nit: Why not just set to `fDefaultHourFormatChar` here?  You can make it default to 0.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -368,6 +368,24 @@ private void getAllowedHourFormats(ULocale uLocale) {
 
         String[] list = getAllowedHourFormatsLangCountry(language, country);
 
+        // We need to check if there is an hour cycle on locale
+        Character defaultCharFromLocale = null;
+        String hourCycle = uLocale.getKeywordValue("hours");
+        switch(hourCycle) {
```
sffc: This line is triggering a NPE.  You should probably just check for `null` before entering your switch.  Maybe adding a `default` would also solve the problem.

icu4c/source/i18n/dtptngen.cpp
caiolima: I renamed `allowedIndex` to `startIndex` to make this code more clear.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
caiolima: Oops. I though `switch` would handle this. I also added a case to test this path on new test.

icu4c/source/test/intltest/dtifmtts.cpp
```diff
@@ -1810,7 +1810,7 @@ void DateIntervalFormatTest::testTicket20707() {
     IcuTestErrorCode status(*this, "testTicket20707");
 
     const char16_t timeZone[] = u"UTC";
-    Locale locales[] = {"en-u-hc-h24", "en-u-hc-h23", "en-u-hc-h12", "en-u-hc-h11"};
+    Locale locales[] = {"en-u-hc-h24", "en-u-hc-h23", "en-u-hc-h12", "en-u-hc-h11", "en"};
```
sffc: Can you also make a test with an invalid value like `en-u-hc-h25`?

caiolima: Done.

sffc:
LGTM, but I would like @pedberg-icu to review also.  Don't squash until we also have his approval

sffc:
I think https://unicode-org.atlassian.net/browse/ICU-20442 is actually a better ticket number for this PR.  ICU-20707 is specifically about DateIntervalFormat.  Can you add tests regarding DateIntervalFormat to this PR?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -667,21 +684,28 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
...
+                case ALLOWED_HOUR_FORMAT_K: fDefaultHourFormatChar = CAP_K; break;
+                case ALLOWED_HOUR_FORMAT_k: fDefaultHourFormatChar = LOW_K; break;
+                default: fDefaultHourFormatChar = CAP_H; break;
+            }
+            startIndex = 1;
         }
+
         for (int32_t i = 0; i < UPRV_LENGTHOF(fAllowedHourFormats); ++i) {
-            fAllowedHourFormats[i] = allowedFormats[i + 1];
+            fAllowedHourFormats[i] = allowedFormats[i + startIndex];
```
pedberg-icu: I think this is wrong, it should remain allowedFormats[i + 1]. Regardless of the changes above, allowedFormats begins with the preferredFormat (which does NOT include b or B for compatibility reasons), and is then followed by a list of allowed values (which always include the preferred value). fAllowedHourFormats should ONLY contain the allowed-value portion of allowedFormats.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -380,11 +400,16 @@ private void getAllowedHourFormats(ULocale uLocale) {
         }
 
         if (list != null) {
-            defaultHourFormatChar = list[0].charAt(0);
-            allowedHourFormats = Arrays.copyOfRange(list, 1, list.length - 1);
+            if (defaultCharFromLocale != null) {
+                defaultHourFormatChar = defaultCharFromLocale;
+                allowedHourFormats = Arrays.copyOfRange(list, 0, list.length - 1);
```
pedberg-icu: Same comment as for C version, we should always do Arrays.copyOfRange(list, 1, list.length - 1) here

icu4c/source/i18n/dtptngen.cpp
caiolima: Thank you for the explanation. i updated the code and added a test to catch this issue as well.

sffc:
When you squash, please use ticket ICU-20442; then, in January, you can make another PR that adds tests for date interval format.

sffc:
Please wait for Peter's re-review of the changes also

sffc:
Great; @caiolima, please squash using ICU-20442 as the ticket number.

caiolima:
Ping review @markusicu

sffc:
@pedberg-icu already approved; I don't think we need to wait for @markus-icu's review.

caiolima:
Thank you very much for the review!

PR #947, opened 2019-12-15
ICU-20924 Use pkg-config to generate the path to pkgdata.inc

hughmcmaster:
On systems without `icu-config` installed, `pkgdata` will fail because it uses the legacy script to generate the path to pkgdata.inc.

Example command: `pkgdata -p bin_mkltfs -m static -v foo.txt'

`pkg-config` is always available, so I've created a patch to use it instead of `icu-config`.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20924
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/tools/pkgdata/pkgdata.cpp
+++ b/icu4c/source/tools/pkgdata/pkgdata.cpp
@@ -95,7 +95,7 @@ static int32_t pkg_archiveLibrary(const char *targetDir, const char *version, UB
 static void createFileNames(UPKGOptions *o, const char mode, const char *version_major, const char *version, const char *libName, const UBool reverseExt, UBool noVersion);
 static int32_t initializePkgDataFlags(UPKGOptions *o);
 
-static int32_t pkg_getOptionsFromICUConfig(UBool verbose, UOption *option);
+static int32_t pkg_getPkgDataPath(UBool verbose, UOption *option);
 static int runCommand(const char* command, UBool specialHandling=FALSE);
 
 #define IN_COMMON_MODE(mode) (mode == 'a' || mode == 'c')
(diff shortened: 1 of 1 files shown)


srl295:
I like the concept, but:

> `pkg-config` is always available

Not quite true. Can you change this to try `pkg-config` first and fall back to `icu-config` ? thanks.

hughmcmaster:
@srl295 Thanks for the review. I've updated the MR.

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -2203,21 +2208,18 @@ static void loadLists(UPKGOptions *o, UErrorCode *status)
...
-    if(buf[strlen(buf)-1]=='\n')
-    {
-        buf[strlen(buf)-1]=0;
-    }
-
-    if(buf[0] == 0)
-    {
-        fprintf(stderr, "%s: icu-config: invalid response from icu-config (fix PATH or use -O option)\n", progname);
+    if (!buf[0]) {
+        fprintf(stderr, "%s: invalid response from pkg-config or icu-config\n", progname);
```
srl295: (Ah, i lost my code I was writing… ) 

Basically, I was suggesting to remember which of `pkgconfigCmd` or `icuconfigCmd` was last used .And then print out  something like this:

> "pkgdata: Could not locate config file. Could not parse the results from 'icu-config --incpkgdatafile'. Check paths or use the -O option to specify a pkgdata.inc file."

hughmcmaster: Sure. I've updated the PR with both changes.

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -2158,41 +2158,46 @@ static void loadLists(UPKGOptions *o, UErrorCode *status)
...
+    return TRUE;
+}
+#endif
+
+/* Get path to pkgdata.inc. Try pkg-config first, falling back to icu-config. */
+static int32_t pkg_getPkgDataPath(UBool verbose, UOption *option) {
+#if U_HAVE_POPEN
+    static char buf[512] = "";
+    UBool pkgconfigIsValid = TRUE;
+    const char *pkgconfigCmd = "pkg-config --variable=pkglibdir icu-i18n";
```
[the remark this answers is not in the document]
hughmcmaster: Yes, `--variable` is fantastic. I'm now using `icu-uc`.

And we could, in another PR, look at adding those variables to the pkg-config file.

hughmcmaster:
Hi @srl295, any chance you can review the latest version of this MR? Thank you.

PR #949, opened 2019-12-17
ICU-20916 LocaleMatcher distinguish between equivalent locales

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20916

### DIFF
--- a/icu4c/source/common/localematcher.cpp
+++ b/icu4c/source/common/localematcher.cpp
@@ -22,7 +22,7 @@
 #include "uhash.h"
 #include "uvector.h"
 
-#define UND_LSR LSR("und", "", "")
+#define UND_LSR LSR("und", "", "", LSR::EXPLICIT_LSR)
 
 /**
  * Indicator for the lifetime of desired-locale objects passed into the LocaleMatcher.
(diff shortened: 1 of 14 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LocaleDistance.java
```diff
@@ -246,7 +284,7 @@ public int getBestIndexAndDistance(LSR desired, LSR[] supportedLSRs,
             if (favorSubtag == FavorSubtag.SCRIPT) {
                 distance >>= 2;
             }
-            if (distance >= threshold) {
+            if (distance >= roundedThreshold) {
```
macchiati: Is this a rounded threshold or "floor" threshold?

markusicu: it is rounded up when the fractions are not 0. see the beginning of this function.

if it was a floor and the shiftedThreshold had fraction bits, then we could ignore a language when it's really still a micro distance below the threshold.

markusicu:
Integration testing has been positive, C++ port is done, so I think it's ready modulo review feedback and squashing.
PTAL

markusicu:
FYI Now that "isParadigmLSR()" checks for equivalence not equality, I found C++ failing at first, and I had to change the LanguageMatcher constructor to not assume a fixed maximum number of paradigm locales. (I got lucky in that we had a test case that exercised the relevant code path and crashed.)

icu4c/source/common/localematcher.cpp
```diff
@@ -446,21 +447,32 @@ LocaleMatcher::LocaleMatcher(const Builder &builder, UErrorCode &errorCode) :
...
+        // First fill the unused slots between paradigm LSRs and other LSRs.
+        // This gap is as large as the number of locales with duplicate LSRs.
+        int32_t i = paradigmIndex;
+        int32_t j = supportedLocalesLength - 1;
+        while (i < otherIndex && otherIndex <= j) {
+            supportedLSRs[i] = supportedLSRs[j];
+            supportedIndexes[i++] = supportedIndexes[j--];
         }
-        supportedLSRsLength = otherIndex - (paradigmLimit - paradigmIndex);
+        // Swap remaining non-paradigm LSRs in place.
```
macchiati: If you always had a gap of at least one (which you could set up at the beginning), you wouldn't need the temps, which would save a (very slight) amount of time).

markusicu: True, but it would leave a gap in the middle. I would then have to either move the trailing section up to close the gap, or null the gap and modify the LocaleDistance code to skip over it (and toString() to not print it). Not worth it I think.

markusicu:
Squashed, please re-approve.

PR #952, opened 2019-12-20
ICU-20310 omit "-true" in toLanguageTag

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20310 
- [X] Issue accepted
- [X] Tests included


icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -3257,6 +3257,11 @@ public String toLanguageTag() {
         subtags = tag.getExtensions();
         for (String s : subtags) {
             buf.append(LanguageTag.SEP);
+            if (s.endsWith("-true")) {
```
roubert: Would it not be more logical to do this inside of `LanguageTag.canonicalizeExtension()` instead?

PR #954, opened 2019-12-21
ICU-20917 LocaleMatcher: prefer a more-default locale

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20917

Compares two supported locales that match equally well but not perfectly. Favors the one which has more-default subtags.

I had thought about another approach: Doing a more complicated addLikelySubtags lookup, keeping information in the LSR about how close it is to the default, and scoring on that. However, that seemed tricky, would slow down all addLikelySubtags processing -- and we only want to distinguish between two supported locales, so we need not tweak the match distance between desired and supported.

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LocaleDistance.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LocaleDistance.java
@@ -255,6 +255,8 @@ public class LocaleDistance {
         long desLangState = desLangDistance >= 0 && supportedLSRs.length > 1 ? iter.getState64() : 0;
         // Index of the supported LSR with the lowest distance.
         int bestIndex = -1;
+        // Cached lookup info from XLikelySubtags.compareLikely().
+        int bestLikelyInfo = -1;
         for (int slIndex = 0; slIndex < supportedLSRs.length; ++slIndex) {
             LSR supported = supportedLSRs[slIndex];
             boolean star = false;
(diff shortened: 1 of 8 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LocaleDistance.java
```diff
@@ -340,13 +342,29 @@ public int getBestIndexAndDistance(LSR desired, LSR[] supportedLSRs,
...
+                if (shiftedDistance < shiftedThreshold) {
+                    if (shiftedDistance == 0) {
+                        return slIndex << INDEX_SHIFT;
+                    }
+                    bestIndex = slIndex;
+                    shiftedThreshold = shiftedDistance;
+                    bestLikelyInfo = -1;
+                }
+            } else {
+                if (shiftedDistance < shiftedThreshold) {
```
macchiati: My eyes might be bleary, but it looks like this test is the same as line 345.

markusicu: Right. I duplicated it and made it do slightly different things for equivalent desired&supported vs. not equivalent. Otherwise I would have to add another variable to remember equivalence, or see if the distance is at least 1.0 and rely on the distance shift to be large enough to keep the lsr flags smaller than 1.0.

markusicu:
now complete with C++ port

markusicu:
The Azure build failure is a failure to install MSYS2. Will merge.

PR #956, opened 2019-12-21
ICU-20929 Fix unused constants in double-conversion-strtod.cpp

Nafanya:
Wrap conditionally-used constant declarations so they are also defined only when they will be used to prevent compile error with `-Wall`enabled:

```
external/icu/icu4c/source/i18n/double-conversion-strtod.cpp:55:18: error: unused variable 'kMaxExactDoubleIntegerDecimalDigits' [-Werror,-Wunused-const-variable]
static const int kMaxExactDoubleIntegerDecimalDigits = 15;

external/icu/icu4c/source/i18n/double-conversion-strtod.cpp:72:21: error: variable 'exact_powers_of_ten' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
static const double exact_powers_of_ten[] = {

external/icu/icu4c/source/i18n/double-conversion-strtod.cpp:98:18: error: unused variable 'kExactPowersOfTenSize' [-Werror,-Wunused-const-variable]
static const int kExactPowersOfTenSize = DOUBLE_CONVERSION_ARRAY_SIZE(exact_powers_of_ten);
```

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20929
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/double-conversion-strtod.cpp
+++ b/icu4c/source/i18n/double-conversion-strtod.cpp
@@ -49,10 +49,6 @@ U_NAMESPACE_BEGIN
 
 namespace double_conversion {
 
-// 2^53 = 9007199254740992.
-// Any integer with at most 15 decimal digits will hence fit into a double
-// (which has a 53bit significand) without loss of precision.
-static const int kMaxExactDoubleIntegerDecimalDigits = 15;
 // 2^64 = 18446744073709551616 > 10^19
 static const int kMaxUint64DecimalDigits = 19;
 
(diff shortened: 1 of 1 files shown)


sffc:
The question here is whether the warning is from upstream or whether it is the result of an ICU patch.  if the warning comes from upstream, we should fix it upstream, to minimize patches that ICU carries against double-conversion.

jefgen:
Thanks for creating this PR.

From taking a look at the upstream file, it looks like this might also be the same in the upstream project as well.

https://github.com/google/double-conversion/blob/master/double-conversion/strtod.cc

It looks like variable `kMaxExactDoubleIntegerDecimalDigits` is only used inside the `else` part of
`#if !defined(DOUBLE_CONVERSION_CORRECT_DOUBLE_OPERATIONS)`.

So I think this it likely not related to any ICU patches.

jefgen:
@Nafanya: Would you be able to create a PR into the https://github.com/google/double-conversion project to get this upstreamed?
If not, let me know, and I can open a PR based on your changes.
(I think I might add the macro at the top rather then move the variable definition.)

Nafanya:
Thanks @jefgen! I've created https://github.com/google/double-conversion/issues/122 and uploaded pull request https://github.com/google/double-conversion/pull/123 , adding second macro instead of moving variables as you suggested.

jefgen:
Thank you @Nafanya!

(FYI: Here is the PR in the upstream project: https://github.com/google/double-conversion/pull/123)

Question for @sffc:
It doesn't look like there is an upstream release of double-conversion with the fix yet.
That said, so you think that maybe this PR should/could be closed, and the ticket could be resolved as a dupe of a future ICU BRS task?

sffc:
Yes, that's the right procedure.  ICU will roll in the upstream fix before the release candidate in March.  It would be counterproductive to roll in this fix in advance because that will just make more work to un-patch in March, so I will close the PR.

PR #957, opened 2019-12-28
ICU-20931 Fixes uloc_acceptLanguage fallback

filmil:
uloc_acceptLanguage fallback computation will never attempt
to find a fallback if the list of available locales only contains
2-letter locales.

For example, for available locales {"fr", "es", "de"} and accept list
of {"es_MX", "nl", "ar_EG"}, no match would be found. But the call
should find "es".

This change fixes fallback computation and adds a regression test.

`TestAcceptLanguage` did not catch this bug because it always used all available locales, many
of which have IDs longer than 2 bytes.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20931
- [ ] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -2371,7 +2371,7 @@ uloc_acceptLanguage(char *result, int32_t resultAvailable,
         }
     }
 
-    for(maxLen--;maxLen>0;maxLen--) {
+    for(;maxLen>0;maxLen--) {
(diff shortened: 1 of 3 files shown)


markusicu:
Thanks, but this is no longer necessary now that I rewrote these functions with https://unicode-org.atlassian.net/browse/ICU-20700 PR #1022 "reimplement acceptLanguage() using the LocaleMatcher"

PR #958, opened 2020-01-03
ICU-20673 Allow built-in translit ID w/o data.

FrankYFTang:
Ignore errors while reading root.res by using local status.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20673
- [X] Issue accepted
- [x] Tests included

### DIFF
--- a/.ci-builds/data-filter.json
+++ b/.ci-builds/data-filter.json
@@ -8,16 +8,18 @@
     ]
   },
 // Test mixed feature filter and resource filter
+// Exlude translit data so we can run test for ICU-20673
   "featureFilters": {
     "misc": {
       "whitelist": ["supplementalData"]
-    }
+    },
+    "translit": "exclude"
   },
   "resourceFilters": [
     {
       "categories": ["misc"],
       "files": {
-	"whitelist": ["supplementalData"]
+        "whitelist": ["supplementalData"]
       },
       "rules": ["+/*"]
     }
@@ -27,8 +29,8 @@
     "directory": "$SRC",
     "replacements": [
       {
-	"src": "translit/Zawgyi_my.txt",
-	"dest": "translit/Zawgyi_my.txt"
+        "src": "brkitr/rules/line.txt",
+        "dest": "brkitr/rules/line_normal.txt"
       },
       "misc/dayPeriods.txt"
     ]
--- a/icu4c/source/i18n/translit.cpp
+++ b/icu4c/source/i18n/translit.cpp
@@ -1508,28 +1508,35 @@ UBool Transliterator::initializeRegistry(UErrorCode &status) {
      */
     //static const char translit_index[] = "translit_index";
 
+    UErrorCode lstatus = U_ZERO_ERROR;
     UResourceBundle *bundle, *transIDs, *colBund;
-    bundle = ures_open(U_ICUDATA_TRANSLIT, NULL/*open default locale*/, &status);
-    transIDs = ures_getByKey(bundle, RB_RULE_BASED_IDS, 0, &status);
+    bundle = ures_open(U_ICUDATA_TRANSLIT, NULL/*open default locale*/, &lstatus);
+    transIDs = ures_getByKey(bundle, RB_RULE_BASED_IDS, 0, &lstatus);
     const UnicodeString T_PART = UNICODE_STRING_SIMPLE("-t-");
 
     int32_t row, maxRows;
-    if (U_SUCCESS(status)) {
+    if (lstatus == U_MEMORY_ALLOCATION_ERROR) {
+        delete registry;
+        registry = nullptr;
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return FALSE;
+    }
+    if (U_SUCCESS(lstatus)) {
         maxRows = ures_getSize(transIDs);
         for (row = 0; row < maxRows; row++) {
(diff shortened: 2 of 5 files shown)


sffc:
LGTM.  If we wanted to run this test on CI without data, we could do it by modifying the build command here to only run that one test:

https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-pipelines.yml#L60

The data filter would be added here:

https://github.com/unicode-org/icu/blob/master/.ci-builds/data-filter.json

Also, please make sure there are tests that cover the two patterns that we identified as problematic:

- `::NFD; ::[:Nonspacing Mark:] Remove; ::NFC;`
- `Lower; NFKD; [:^L:]Remove; NFC;`

FrankYFTang:
Please take another look. I added unit tests. I am not sure we should  run this test on CI without data or not.

sffc:
LGTM.  I pushed a change to the Azure CI builds config file that should run the test in a no-data environment.  Adding @jefgen to review the change to the CI file.

sffc:
Looks good:

```
...
2020-01-14T09:57:24.9077471Z Spawning Python to generate data/rules.mk...
2020-01-14T09:57:25.2317187Z Note: Applying filters from ../../.ci-builds/data-filter.json.
...
(no logs that show building of translit data)
...
2020-01-14T10:02:27.9039422Z === Handling test: translit/TransliteratorTest/TestBasicTransliteratorEvenWithoutData: ===
2020-01-14T10:02:27.9039778Z    translit {
2020-01-14T10:02:27.9040183Z       TransliteratorTest {
2020-01-14T10:02:27.9040523Z          TestBasicTransliteratorEvenWithoutData {
2020-01-14T10:02:27.9066766Z          } OK:   TestBasicTransliteratorEvenWithoutData 
2020-01-14T10:02:27.9067272Z       } OK:   TransliteratorTest 
2020-01-14T10:02:27.9067620Z    } OK:   translit 
2020-01-14T10:02:27.9067966Z 
2020-01-14T10:02:27.9068613Z --------------------------------------
2020-01-14T10:02:27.9069047Z OK: All tests passed without error.
2020-01-14T10:02:27.9069851Z --------------------------------------
2020-01-14T10:02:27.9070284Z Elapsed Time: 00:00:00.004
```

.ci-builds/data-filter.json
```diff
@@ -8,16 +8,18 @@
     ]
   },
 // Test mixed feature filter and resource filter
+// Exlude translit data so we can run that test
```
jefgen: minor nit: maybe refer to the ticket number (ICU-20673) instead of just "that" so there's a breadcrumb to follow later on for somebody else?  :)

icu4c/source/i18n/translit.cpp
```diff
@@ -1508,28 +1508,29 @@ UBool Transliterator::initializeRegistry(UErrorCode &status) {
...
+    UErrorCode lstatus = U_ZERO_ERROR;
     UResourceBundle *bundle, *transIDs, *colBund;
-    bundle = ures_open(U_ICUDATA_TRANSLIT, NULL/*open default locale*/, &status);
-    transIDs = ures_getByKey(bundle, RB_RULE_BASED_IDS, 0, &status);
+    bundle = ures_open(U_ICUDATA_TRANSLIT, NULL/*open default locale*/, &lstatus);
+    transIDs = ures_getByKey(bundle, RB_RULE_BASED_IDS, 0, &lstatus);
     const UnicodeString T_PART = UNICODE_STRING_SIMPLE("-t-");
 
     int32_t row, maxRows;
-    if (U_SUCCESS(status)) {
```
jefgen: >if (U_SUCCESS(status)) { [](start = 4, length = 24)

The code before doesn't really do a great job of checking if the data loading above fails due to out-of-memory (OOM) -- but with the changes to use a local status variable now (`lstatus`) any OOM failure from `ures_open` or `ures_getByKey` would be completely hidden entirely.

Would you mind adding a check to see if `lstatus` is OOM here? 

For example, here's a similar example in another file where a localStatus is used:
[https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/numsys.cpp#L153](https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/numsys.cpp#L153)

Perhaps something like this:
```
if (lstatus == U_MEMORY_ALLOCATION_ERROR) {
    delete registry;
    registry = nullptr;
    return FALSE;
}
```

jefgen:
> I pushed a change to the Azure CI builds config file that should run the test in a no-data environment. Adding @jefgen Jeff Genovy FTE to review the change to the CI file.

The changes to the CI file LGTM.
Thanks @sffc for adding this!

icu4c/source/test/intltest/transtst.cpp
```diff
@@ -1546,6 +1563,15 @@ void TransliteratorTest::TestBasicTransliteratorEvenWithoutData() {
         if (translit.get() == nullptr || !U_SUCCESS(status)) {
             dataerrln("FAIL: createInstance %s failed", BASIC_TRANSLITERATOR_ID[i]);
         }
+        UnicodeString data(TEST_DATA);
+        UnicodeString expected(EXPECTED_RESULTS[i]);
+        translit->transliterate(data);
+        if (data != expected) {
+            std::string utf8;
```
sffc: Unused variable?

PR #961, opened 2020-01-12
ICU-11548 Improve regex static UnicodeSets handling

aheninger:
Compiled regular expression patterns make use of several shared common
UnicodeSets. This change simplifies the creation and use of these
static UnicodeSets.

- Pointer fields to the static sets are removed from the compiled patterns,
  and the static variables are accessed directly. The deleted pointers
  were a hold-over from earlier code that did not use shared statics.

- The UnicodeSet pattern literals are changed from hex constants to
  u"string literals".

- The size of fRuleSets (from regexst.h) is changed from a hard-coded 10
  to the number of UnicodeSets actually required. Doing this required
  a change to regexcst.pl to export the required size. Changing and
  rerunning this perl code resulted in massive but benign changes to
  the generated file regexcst.h, the result of perl having changed its
  order of enumeration of hashes since the file was last regenerated.

- UnicodeSets are frozen when possible. Should result in faster matching.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-11548
- [x] Issue accepted
- [ ] Tests included


aheninger:
Here are some more changes that were underway at the time I left, that would be good to get in.

There is more that could be done to improve Regular Expression's use of Unicode sets - the ticket lists some - but I don't know when that might happen.

It would be better to have this change in before beginning work on ICU-20876 "Update Grapheme Cluster boundary rules for Regex", which is scheduled for 67.1. The changes for that ticket would likely conflict with those here. Also, the changes from this PR would make an easier base to work from.

icu4c/source/i18n/regexst.cpp
```diff
@@ -35,216 +35,101 @@
...
-    0x5b, 0x3a, 0x47, 0x72, 0x61, 0x70, 0x68, 0x65, 0x6d, 0x65, 0x5f,
-//    E     x     t     e     n     d     :     ]     ]
-    0x45, 0x78, 0x74, 0x65, 0x6e, 0x64, 0x3a, 0x5d, 0x5d, 0};
-
-static const UChar gGC_ExtendPattern[] = {
-//    [     \     p     {     G     r     a     p     h     e     m     e     _
-    0x5b, 0x5c, 0x70, 0x7b, 0x47, 0x72, 0x61, 0x70, 0x68, 0x65, 0x6d, 0x65, 0x5f,
-//    E     x     t     e     n     d     }     ]
-    0x45, 0x78, 0x74, 0x65, 0x6e, 0x64, 0x7d, 0x5d, 0};
+constexpr char16_t const *gGC_ControlPattern = u"[[:Zl:][:Zp:][:Cc:][:Cf:]-[:Grapheme_Extend:]]";
```
sven-oly: These are nice cleanups to improve readability.

icu4c/source/i18n/regexcst.h
```diff
@@ -197,7 +197,7 @@ static const struct RegexTableEl gRuleParseStateTable[] = {
     , {doBadOpenParenType, 255, 206,0,  FALSE}     //  45 
     , {doOpenLookBehind, 61 /* = */, 2, 20, TRUE}     //  46      open-paren-lookbehind
     , {doOpenLookBehindNeg, 33 /* ! */, 2, 20, TRUE}     //  47 
-    , {doBeginNamedCapture, 128, 64,0,  FALSE}     //  48 
+    , {doBeginNamedCapture, 129, 64,0,  FALSE}     //  48
```
sven-oly: Not sure why the 128 & 129 changes.

icu4c/source/i18n/regexcst.h
```diff
@@ -20,117 +20,117 @@ U_NAMESPACE_BEGIN
...
-    doSetMatchMode,
-    doEnterQuoteMode,
-    doOpenCaptureParen,
-    doContinueNamedCapture,
-    doSetBackslash_d,
-    doBeginMatchMode,
-    doBackslashX,
-    doSetPosixProp,
-    doIntervalError,
+    doSetBackslash_D,
```
sven-oly: This looks like a sorted list of the parse actions.

icu4c/source/i18n/regexst.h
```diff
@@ -39,17 +40,16 @@ class RegexStaticSets : public UMemory {
...
+    UnicodeSet    fPropSets[URX_LAST_SET] {};      // The sets for common regex items, e.g. \s
+    Regex8BitSet  fPropSets8[URX_LAST_SET] {};     // Fast bitmap sets for latin-1 range for above.
 
-    UnicodeSet    fRuleSets[10];               // Sets used while parsing regexp patterns.
-    UnicodeSet    fUnescapeCharSet;            // Set of chars handled by unescape when
-                                               //   encountered with a \ in a pattern.
-    UnicodeSet    *fRuleDigitsAlias;
-    UText         *fEmptyText;                 // An empty string, to be used when a matcher
-                                               //   is created with no input.
+    UnicodeSet    fRuleSets[kRuleSet_count] {};    // Sets used while parsing regexp patterns.
```
sven-oly: This used to be an array of 10 elements, but is now only 3. Is this OK?

aheninger: Yes, the 3 is computed at build time to be the actual size required.

PR #969, opened 2020-01-29
ICU-20953 Tell GitHub to use JSONC to fix syntax highlighting for comments

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20953
- [x] Issue accepted
- [ ] Tests included

### Description:
This change informs GitHub to use the "JSONC" format for JSON files when syntax highlighting, in order to better handle comments in JSON files.
(Note: There should be no "real" impact here other than fixing the syntax highlighting.)

#### Before:
image

#### After:
image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -84,3 +84,7 @@ configure text !eol
 
 # Enable syntax highlighting on GitHub.com
 .cpyskip.txt linguist-language=Ignore-List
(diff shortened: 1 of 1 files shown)


younies:
Thanks for this catch :)

PR #971, opened 2020-01-31
ICU-20958 Prevent SEGV_MAPERR in append

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20958
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/unistr.cpp
+++ b/icu4c/source/common/unistr.cpp
@@ -1563,7 +1563,11 @@ UnicodeString::doAppend(const UChar *srcChars, int32_t srcStart, int32_t srcLeng
   }
 
   int32_t oldLength = length();
-  int32_t newLength = oldLength + srcLength;
+  int32_t newLength;
+  if (uprv_add32_overflow(oldLength, srcLength, &newLength)) {
+    setToBogus();
+    return *this;
+  }
 
   // Check for append onto ourself
   const UChar* oldArray = getArrayStart();
--- a/icu4c/source/test/intltest/ustrtest.cpp
+++ b/icu4c/source/test/intltest/ustrtest.cpp
@@ -67,6 +67,7 @@ void UnicodeStringTest::runIndexedTest( int32_t index, UBool exec, const char* &
     TESTCASE_AUTO(TestWCharPointers);
     TESTCASE_AUTO(TestNullPointers);
     TESTCASE_AUTO(TestUnicodeStringInsertAppendToSelf);
+    TESTCASE_AUTO(TestLargeAppend);
     TESTCASE_AUTO_END;
 }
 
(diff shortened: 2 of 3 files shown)


icu4c/source/common/unistr.cpp
```diff
@@ -1545,6 +1545,8 @@ UnicodeString::doAppend(const UnicodeString& src, int32_t srcStart, int32_t srcL
   return doAppend(src.getArrayStart(), srcStart, srcLength);
 }
 
+static const int32_t MAX_INT32 = 0x7FFFFFFF;
```
jefgen: Could you use `INT32_MAX`? Or is there a reason to define your own constant?

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2310,3 +2311,21 @@ void UnicodeStringTest::TestUnicodeStringInsertAppendToSelf() {
...
 }
+
+void UnicodeStringTest::TestLargeAppend() {
+    IcuTestErrorCode status(*this, "TestLargeAppend");
+    // Make a large UnicodeString
+    int32_t len = 0xAFFFFFF;
+    UnicodeString str;
+    char16_t *buf = str.getBuffer(len);
+    for (int32_t i = 0; i < len; i++) {
+      *buf++ = (char16_t)'a';
```
jefgen: nit: you could use `u'a'` here and avoid the cast.

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2310,3 +2311,21 @@ void UnicodeStringTest::TestUnicodeStringInsertAppendToSelf() {
...
     assertEquals("", u"abbcdcde", str);
 }
+
+void UnicodeStringTest::TestLargeAppend() {
+    IcuTestErrorCode status(*this, "TestLargeAppend");
+    // Make a large UnicodeString
+    int32_t len = 0xAFFFFFF;
+    UnicodeString str;
+    char16_t *buf = str.getBuffer(len);
+    for (int32_t i = 0; i < len; i++) {
```
sffc: This test looks very slow: a loop for 0xAFFFFFF iterations.  Please either restrict this test to exhaustive mode or make it faster so that it can run in under 10ms.

icu4c/source/common/unistr.cpp
```diff
@@ -1563,6 +1563,11 @@ UnicodeString::doAppend(const UChar *srcChars, int32_t srcStart, int32_t srcLeng
   }
 
   int32_t oldLength = length();
+  if (INT32_MAX - oldLength < srcLength) {
```
sffc: An alternative would be to use uprv_add32_overflow.  However, what you have written here is likely faster since we know the inputs are positive.

FYI for the future:

```cpp
int32_t oldLength = length();
int32_t newLength;
bool overflow = uprv_add32_overflow(oldLength, srcLength, &newLength);
if (overflow) {
  setToBogus();
  return *this;
}
```

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2310,3 +2311,21 @@ void UnicodeStringTest::TestUnicodeStringInsertAppendToSelf() {
...
+      *buf++ = u'a';
+    }
+    str.releaseBuffer(len);
+    UnicodeString dest;
+    // Append it 16 times
+    for (int32_t i = 0; i < 16; i++) {
```
sffc: Note: 0xAFFFFFF times 16 is 0xA4FFFFF1, which is greater than INT32_MAX, which is 0x7FFFFFFF.  Can you add that as a comment?

sffc:
There are lots of `+` operations all over this file.  Have you verified that this is the only one where overflow matters?

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2310,3 +2311,21 @@ void UnicodeStringTest::TestUnicodeStringInsertAppendToSelf() {
...
+      dest.append(str);
+    }
+    // Pass the test if it does not crash and reach here.
```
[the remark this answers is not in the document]
FrankYFTang: "On 32 bit platforms the test might fail in unexpected ways from memory allocation failures. Probably would cause the string to go bogus sooner than expected." 
how should I wrote that test then?

icu4c/source/test/intltest/ustrtest.cpp
FrankYFTang: done

icu4c/source/common/unistr.cpp
FrankYFTang: done

icu4c/source/test/intltest/ustrtest.cpp
FrankYFTang: done

FrankYFTang:
> There are lots of `+` operations all over this file. Have you verified that this is the only one where overflow matters?

no

sffc:
LGTM please squash

carnil:
Unless mistaken this issue is present before 3d77fc18b8b1 as well (which did out split a new doAppend() from the more general doReplace(), but the overflow is present before. I have not checked where exactly it might be introduced but hope this information help any downstream which ship as well older versions of icu. 

Looks indeed that for instance Ubuntu issued a USN https://usn.ubuntu.com/4305-1/ fixing versiosn back to 4.8.1.1  based ones.

PR #979, opened 2020-02-12
ICU-20972 Fix invalid conversion from const char8_t* to const char* (C++20)

stac47:
Following advise from https://stackoverflow.com/questions/57402464/is-c20-char8-t-the-same-as-our-old-char/57453713#57453713


markusicu:
The commit message needs to start with "ICU-20972 Fix invalid conv..."

stac47:
Done in #997 
Must I close this PR ?

srl295:
> Done in #997
> Must I close this PR ?

Can you instead just force push your branch to `stac47:fix_chat8_t`  ? Then you can just click Edit on this PR, and change the base from `master` to `maint/maint-66`, This is preferable so we keep the history of this PR.

markusicu:
> Can you instead just force push your branch to `stac47:fix_chat8_t` ? Then you can just click Edit on this PR, and change the base from `master` to `maint/maint-66`, This is preferable so we keep the history of this PR.

Right, that would be much better. `git commit --amend` to fix the commit message, and redirect via the GitHub UI.

PR #980, opened 2020-02-12
ICU-20973 Fix rewritten operator!= failure due non-bool return type

stac47:
[ICU-20973]

Note that the change in icu4c/source/i18n/basictz.cpp is a work around waiting the reply on the ticket raised to gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93699

### DIFF
--- a/icu4c/source/common/uchriter.cpp
+++ b/icu4c/source/common/uchriter.cpp
@@ -84,6 +84,11 @@ UCharCharacterIterator::operator==(const ForwardCharacterIterator& that) const {
         && end == realThat.end;
 }
 
+UBool
+UCharCharacterIterator::operator!=(const ForwardCharacterIterator& that) const {
+    return !operator==(that);
+}
+
 int32_t
 UCharCharacterIterator::hashCode() const {
     return ustr_hashUCharsN(text, textLength) ^ pos ^ begin ^ end;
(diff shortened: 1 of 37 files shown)


markusicu:
It's got to be a compiler bug since (a) the operator!= is explicitly defined in the base class and (b) this code has worked on dozens of compilers for over 20 years. See ticket discussion.

pepsiman:
Clang 10 agrees with GCC 10, so I don't think it's a compiler bug.

Why does the pull request fix it by adding `operator!=`, instead of changing the return types of `operator==` to `bool`?

markusicu:
> Clang 10 agrees with GCC 10, so I don't think it's a compiler bug.

So then it's a change in the standard that breaks code that has worked for some 25 years.

> Why does the pull request fix it by adding `operator!=`, instead of changing the return types of `operator==` to `bool`?

We have discussed this twice this summer. We cannot change the C API because it has to be binary compatible. We could probably change the C++ API, except where structs cross the library boundary. The details (e.g., sizeof) of type bool are up to the implementation and may differ between C and C++.

We may yet revisit this specifically for `operator!=` but Fredrik is exploring a different route (`using`).

markusicu:
We decided to go a different route, see the ticket.

ReeceSX:
\> unicode-org
\> clang, msvc, and gcc all say one thing
\> its just a compiler bug bro
\> closed without solution 

why am i not surprised

PR #986, opened 2020-02-15
ICU-20961 Return correct currency plural pattern from DecimalFormat

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20961
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/number_mapper.cpp
+++ b/icu4c/source/i18n/number_mapper.cpp
@@ -63,17 +63,8 @@ MacroProps NumberPropertyMapper::oldToNew(const DecimalFormatProperties& propert
     // AFFIXES //
     /////////////
 
-    AffixPatternProvider* affixProvider;
-    if (properties.currencyPluralInfo.fPtr.isNull()) {
-        warehouse.currencyPluralInfoAPP.setToBogus();
-        warehouse.propertiesAPP.setTo(properties, status);
-        affixProvider = &warehouse.propertiesAPP;
-    } else {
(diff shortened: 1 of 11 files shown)


sffc:
Summary: there are 3 places where an AffixPatternProvider is constructed.  Two of the places had a correct if statement that conditionally uses the currency plural provider, but the third place unconditionally used the non-currency-plural provider.  This PR makes all three sites use the same logic by moving that logic into a helper function.

The helper function is implemented differently in Java versus C++ because of object lifecycle issues.

sffc:
@jefgen - Azure CI says "Build Not Found" for the test failures above ([link](https://unicode-icu.visualstudio.com/ICU/_build/results?buildId=7030)).  I assume they got garbage collected.  But, it's only been 17 days since they were ran, and I was just now getting around to fixing the issues, but I have to wait for them to re-run again.  Is there a way to increase the retention policy to something more useful, like, say, 30 days or 90 days?

jefgen:
@sffc It looks like the default settings for Azure Pipelines projects is to only keep PR runs for 10 days -- though it does keep other runs for 30 days (ex: runs on master). That likely explains why these ones are deleted/garbage collected after 17 days.
I'll see if I can change to retention policy to 30 days for PR runs since this one is technically a 'paid' instance and not the free tier.

sffc:
The "re-run" buttons don't seem to be working.  I'll force-push (updating the timestamp on the commit) and just let all the testing restart from scratch.

sffc:
OK, looks to be working now.

PR #987, opened 2020-02-15
ICU-20956 Fix monetary symbol getters in DecimalFormat

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20956
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormatSymbols.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormatSymbols.java
@@ -888,9 +888,32 @@ public class DecimalFormatSymbols implements Cloneable, Serializable {
         if (currency == null) {
             throw new NullPointerException();
         }
+        if (currency.equals(this.currency)) {
+            return;
+        }
+        CurrencyDisplayInfo displayInfo = CurrencyData.provider.getInstance(ulocale, true);
+        setCurrencyOrNull(currency, displayInfo);
+    }
+
+    private void setCurrencyOrNull(Currency currency, CurrencyDisplayInfo displayInfo) {
         this.currency = currency;
+
+        if (currency == null) {
+            intlCurrencySymbol = "XXX";
+            currencySymbol = "\u00A4"; // 'OX' currency symbol
+            currencyPattern = null;
+            return;
+        }
+
         intlCurrencySymbol = currency.getCurrencyCode();
-        currencySymbol = currency.getSymbol(requestedLocale);
+        currencySymbol = currency.getSymbol(ulocale);
+
+        CurrencyFormatInfo formatInfo = displayInfo.getFormatInfo(currency.getCurrencyCode());
+        if (formatInfo != null) {
+            setMonetaryDecimalSeparatorString(formatInfo.monetaryDecimalSeparator);
+            setMonetaryGroupingSeparatorString(formatInfo.monetaryGroupingSeparator);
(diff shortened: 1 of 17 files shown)


icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
```diff
@@ -6613,6 +6614,25 @@ public void test20348_CurrencyPrefixOverride() {
...
+        decimalFormat.setCurrency(currency);
+
+        DecimalFormatSymbols decimalFormatSymbols = decimalFormat.getDecimalFormatSymbols();
+        assertEquals("MONETARY DECIMAL SEPARATOR", ".", decimalFormatSymbols.getMonetaryDecimalSeparatorString());
+        assertEquals("DECIMAL SEPARATOR", ",", decimalFormatSymbols.getDecimalSeparatorString());
+        assertEquals("MONETARY GROUPING SEPARATOR", " ", decimalFormatSymbols.getMonetaryGroupingSeparatorString());
+        assertEquals("GROUPING SEPARATOR", " ", decimalFormatSymbols.getGroupingSeparatorString());
+        assertEquals("CURRENCY SYMBOL", "kr", decimalFormatSymbols.getCurrencySymbol());
+
+        StringBuffer string = decimalFormat.format(new BigDecimal(12345.12), new StringBuffer(), DontCareFieldPosition.INSTANCE);
```
echeran: Change variable `string` to something else?  The call in the following line of `string.toString()` seems strange.  Other instances in the file use `buf`, `buffer`, etc. (ex: `pat`).

icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormatSymbols.java
```diff
@@ -888,9 +888,32 @@ public void setCurrency(Currency currency) {
...
+            return;
+        }
+        CurrencyDisplayInfo displayInfo = CurrencyData.provider.getInstance(ulocale, true);
+        setCurrencyOrNull(currency, displayInfo);
+    }
+
+    private void setCurrencyOrNull(Currency currency, CurrencyDisplayInfo displayInfo) {
         this.currency = currency;
+
+        if (currency == null) {
```
echeran: Should we have a test for the case when currency == null?  Or documentation of when such a case would actually occur?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
sffc: Yeah, I had copied and pasted this test from the bug thread.  Cleaned up.

icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormatSymbols.java
sffc: This code path does already get exercised 128 times in NumberFormatTest; basically whenever a DecimalFormat is instantiated with a locale without a region (like "en" instead of "en-US").  So I'd say that test coverage is already sufficient.

As far as documentation goes, the behavior should be unchanged from before.  Previously, these fallback symbols were set in the "initialize" method; now they're being set in this method.

sffc:
Hi, thanks for the LGTM.  I have squashed and need another stamp.

echeran:
Stamped. LGTM.

PR #989, opened 2020-02-16
ICU-20977 Minimal support for WebAssembly in double-conversion-utils.h

kateinoigakukun:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20977
- [ ] Issue accepted
- [ ] Tests included


sffc:
It looks like [upstream double-conversion](https://github.com/google/double-conversion/blob/master/double-conversion/utils.h) now has `__wasm32__` here:

```
    defined(__EMSCRIPTEN__) || defined(__wasm32__)
```

We roll upstream double-conversion into ICU at every feature release.  The next feature release is ICU 67, scheduled for mid April.

Closing this because we don't want to introduce unnecessary diffs in our vendor code.

PR #997, opened 2020-02-20
ICU-20972 Fix invalid conversion from const char8_t* to const char* (C++20)

stac47:
This is temporary fix proposed in #979 eligible to release 66. Version 67 should have a nicer fix to avoid all these reinterpret_cast s.


srl295:
> This is temporary fix proposed in #979 eligible to release 66. Version 67 should have a nicer fix to avoid all these reinterpret_cast s.

please close this PR and instead just update https://github.com/unicode-org/icu/pull/979#issuecomment-588625466

PR #1004, opened 2020-02-27
ICU-20987 integrate CLDR release-37-alpha1 to master (using new tooling)

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20987
- [x] Issue accepted
- [x] Tests included

This integrates the CLDR 37 data now tagged as "release-37-alpha1" (in both the cldr and cldr-staging repositories) to ICU master for ICU 67 pre-release candidate.

This is the first CLDR to ICU integration using the new tooling in the the icu directory tools/cldr/cldr-to-icu/; icu4c/source/data/cldr-icu-readme.txt has been updated to describe the new process. There are some changes related to tooling:
1. Configuring which locales to include (as well as which data to include for collation, rbnf etc.) is now handled by modifying tools/cldr/cldr-to-icu/build-icu-data.xml, instead of by modifying files such as icu4c/source/data/icu-config.xml and icu4c/source/data/build.xml (though I have tried to keep the latter up to date so the old process can still be run to produce parallel results). The format in which locales for inclusion are specified in the new build-icu-data.xml file is somewhat different; the various regional and variant locales need  not be specified for a given language and script, they will all be automatically included (unless explicitly skipped in tools/cldr/cldr-to-icu/LdmlConverter). Note that the regional locales are specified anyway for 3-letter language codes (not sure why).
2. As a result, the new tooling would include the CLDR locale ca_ES_VALENCIA, which was not previously in ICU. However, ICU code does not currently handle roundtrip of this locale ID to/from BCP47 form, so this locale is specifically excluded until we have a fix for https://unicode-org.atlassian.net/browse/ICU-20997.
3. The new tooling carries any tender="false" attribute from CLDR supplemental <currencyData><region><currency> items to the CurrencyMap in icu4c/source/data/curr/supplementalData.txt; the old tooling did not. ICU4C code does nt pay attention to this, but ICU4J code in ICUCurrencyMetaInfo does pay attention, so that methods like Currency.getKeywordValuesForLocale may return a smaller set of currencies than before.

Another significant change not related to new tooling: The CLDR locale matching data has been updated to increase the distance between "nn" and other language codes for Norwegian, so locale distance and matching tests involving "nn" produce different results.

Finally, several new locales have been added for ff_Adlm, mai, mni, pcm, sat, sd_Deva, and su.


icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -50,17 +55,22 @@
...
+# CLDR_DIR:      For most of the process, this is the path to the to root of
+#                standard CLDR sources, below which are the common and
+#                tools directories. For running LdmlConverter, this needs to be
+#                temporarily reset to the parallel root for the production data,
+#                corresponding to $CLDR_TMP_DIR/production (see description of
+#                CLDR_TMP_DIR below).
 # CLDR_CLASSES:  Path to the CLDR Tools classes directory. If not set, defaults
 #                to $CLDR_DIR/tools/java/classes
 #
 # CLDR_TMP_DIR:  Parent of temporary CLDR production data.
```
hagbard: Seems like we want a bug raised to get rid of this.

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -73,14 +83,25 @@
...
 # there are some xml files in the ICU sources that need to be updated (these xml
 # files are used in addition to the CLDR files as inputs to the CLDR data build
 # process for ICU):
 #
-#    icu4c/source/data/icu-config.xml - Update <locales> to add or remove
+# The primary file to edit for ICU 67 and later is
+#
+#    $TOOLS_ROOT/cldr/cldr-to-icu/build-icu-data.xml
+#
+# The files used in previous versions are the following; unsure whether these
```
hagbard: They are definitely not used. The new tooling uses nothing other than the new API and code/data in its own package/directory.

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -159,40 +208,53 @@ cd $ICU4C_DIR/source
...
-ant all 2>&1 | tee /tmp/cldr-newData-buildLog.txt
+ant proddata 2>&1 | tee /tmp/cldr-newData-proddataLog.txt
 
-# NOTE: if you change the CLDR data, please run "ant cleanprod" to clean out the
-# temporary production data directory (usually $CLDR_DIR/../cldr-aux/production )
+# 4b. Build the new ICU4C data files; these include .txt files and .py files.
+# These new files will replace whatever was already present in the ICU4C sources.
+# This process uses the LdmlConverter in $TOOLS_ROOT/cldr/cldr-to-icu/
+#
+# This process will take several minutes,, during most of which there will be no log
```
hagbard: extraneous ','

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -159,40 +208,53 @@ cd $ICU4C_DIR/source
...
+# output (so do not assume nothing is happening). Keep a log so you can investigate
+# anything that looks suspicious.
+#
+# This also requires temporarily redefining CLDR_DIR.
+
```
hagbard: We should definitely fix this to not need this sort of messing about.

icu4c/source/data/cldr-icu-readme.txt
pedberg-icu: Yes, filed https://unicode-org.atlassian.net/browse/ICU-20998 for this

icu4c/source/data/cldr-icu-readme.txt
pedberg-icu: OK, I will update the comment when I do the next round of CLDR-to-ICU integration for CLDR release-37-alpha2 next week.

icu4c/source/data/cldr-icu-readme.txt
pedberg-icu: Thanks, I will also fix this when I do the next round of CLDR-to-ICU integration for CLDR release-37-alpha2 next week.

icu4c/source/data/cldr-icu-readme.txt
pedberg-icu: OK, thanks, yes I could do that. I don't see a need to immediately update the doc, I will fix it the next time I do a CLDR integration (it will happen at least 2 more times before we are done with ICU 67).

markusicu:
I feel ill-equipped to review this. GitHub stops at 3000 files and so does not even show changes in library code or test code.

Next time, it would be convenient to get one commit with only version number changes in data files, and then one or more commits with all of the files that have more than just a version number change. That would make the changes reviewable.

I guess one way to review this huge commit is to download it locally and futz with diff-tool settings to ignore lines with "Version".

I trust that Steven did something like that... he approved, so Peter is not blocked.

Peter, if you still want to discuss specific changes with me and/or Mark, then please help us see what they are.

markusicu:
A mechanically simpler way to slice it into commits would be to just have a commit per folder, such as source/data/coll, source/data/locales, ...

icu4c/source/data/coll/es__TRADITIONAL.txt
```diff
@@ -1,7 +0,0 @@
-﻿// © 2016 and later: Unicode, Inc. and others.
-// License & terms of use: http://www.unicode.org/copyright.html#License
-es__TRADITIONAL{
```
jefgen: Why did this file get deleted?
Also same for `de__PHONEBOOK.txt` as well?

hagbard: This might be a mistake, maybe by the cleanup code. My tool won't generate these (it cannot generate files with invalid locale names). I think we decided to leave these manually maintained, and they are not mentioned in my anlysis:
https://docs.google.com/spreadsheets/d/10UqiRFEXVgFygSchF7ZhWcz5pC0_gqAE5sgPGIB36iM/edit#gid=474876707
So at this time I'm suspecting I'm deleting it by mistake.

There is a way to specify files not to touch in the Ant build-icu-data.xml file which might need updated.

hagbard: Hmm, it should be keeping them since they are explicitly mentioned in:
https://github.com/unicode-org/icu/blob/master/tools/cldr/cldr-to-icu/build-icu-data.xml#L350
So if it is the tool that's deleting then, it's an error.

I suspect it's actually that Peter ran an additional (and as far as these data files are concerned unnecessary) cleanup step before running the new tooling (I saw something in the updated instructions). The new tool handles these directories with the explicit assumption that some files exist and get preserved (i.e. they are now only manually maintained).

pedberg-icu: I did run ant clean in icu4c/source/data before running the new tooling. How else to handle e.g. regional locales that get removed, or that have no data in certain categories? Such files get removed in every integration. If we just preserve all existing files that are not overwritten by the new data we will end up with tons of cruft. I was assuming the tooling would generate all necessary stubs of the sort under consideration here, the old tooling did.

jefgen:
Testing manually triggering CI builds:
/azp list

hagbard:
On Fri, 28 Feb 2020 at 05:35, pedberg-icu <notifications@github.com> wrote:

> *@pedberg-icu* commented on this pull request.
> ------------------------------
>
> In icu4c/source/data/coll/es__TRADITIONAL.txt
> <https://github.com/unicode-org/icu/pull/1004#discussion_r385503715>:
>
> > @@ -1,7 +0,0 @@
>
> -﻿// © 2016 and later: Unicode, Inc. and others.
>
> -// License & terms of use: http://www.unicode.org/copyright.html#License
>
> -es__TRADITIONAL{
>
>
> I did run ant clean in icu4c/source/data before running the new tooling.
> How else to handle e.g. regional locales that get removed, or that have no
> data in certain categories? Such files get removed in every integration. If
> we just preserve all existing files that are not overwritten by the new
> data we will end up with tons of cruft. I was assuming the tooling would
> generate all necessary stubs of the sort under consideration here, the old
> tooling did.
>

Simple, we don't preserve existing files that are overwritten, and you
won't end up with lots of cruft.

1) Only files listed in the retain section are ever kept.
2) Files not listed are checked to make sure they look like generated files
(with the right header).
3) Any files not in (1) or (2) cause an error.

So for the directories which are written to by the tool (well, more
precisely, those specified in the build file to be cleaned) you should not
need any other file cleanup.

David

-- 
David Beaumont :: Îñţérñåţîöñåļîžåţîờñ Libraries :: Google
Google Switzerland GmbH., Brandschenkestrasse 110, CH-8002, Zürich -
Switzerland

pedberg-icu:
OK, when I re-do the integration on Monday (new PR, same JIRA issue) I will restore the deleted files, and update the cldr-icu-readme instructions accordingly.

PR #1006, opened 2020-03-02
ICU-20980 BRS66GA Update version numbers and README files

daniel-ju:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20980
- [x] Issue accepted
- [ ] Tests included


jefgen:
The changes/updates to the *.jar files for the version number (in this PR) will conflict with the CLDR data updates that are in PR #1007.
I think it might be better to wait for CLDR updates to be merged first, and then we can do the version update afterwards and rebuild the jar files.

daniel-ju:
> The changes/updates to the *.jar files for the version number (in this PR) will conflict with the CLDR data updates that are in PR #1007.
> I think it might be better to wait for CLDR updates to be merged first, and then we can do the version update afterwards and rebuild the jar files.

The CLDR integration is going into the master branch for ICU 67, while the version updates are going into the maint-66 branch. There shouldn't be any merge conflict issues from that, or am I mistaken?

jefgen:
> The CLDR integration is going into the master branch for ICU 67, while the version updates are going into the maint-66 branch. There shouldn't be any merge conflict issues from that, or am I mistaken?

You're absolutely correct! My mistake.
(I think I was confused by the two releases happening so close to one another, and being up too late...)

PR #1017, opened 2020-03-05
ICU-21000 Fix abort called by DateTimePatternGenerator getDefaultHourCycle

jefgen:
If you call the C++ API `getDefaultHourCycle` on an empty `DateTimePatternGenerator`
instance, or the C API `udatpg_getDefaultHourCycle`, then it hits `UPRV_UNREACHABLE` which calls `abort()`.
We should likely return an error code instead of aborting.
Add test cases for both the C++ and C APIs.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21000
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -713,19 +713,28 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
 }
 
 UDateFormatHourCycle
-DateTimePatternGenerator::getDefaultHourCycle(UErrorCode& /*status*/) const {
-  switch(fDefaultHourFormatChar) {
-    case CAP_K:
-      return UDAT_HOUR_CYCLE_11;
-    case LOW_H:
-      return UDAT_HOUR_CYCLE_12;
-    case CAP_H:
-      return UDAT_HOUR_CYCLE_23;
-    case LOW_K:
-      return UDAT_HOUR_CYCLE_24;
-    default:
-      UPRV_UNREACHABLE;
-  }
+DateTimePatternGenerator::getDefaultHourCycle(UErrorCode& status) const {
+    if (U_FAILURE(status)) {
+        return UDAT_HOUR_CYCLE_23;
+    }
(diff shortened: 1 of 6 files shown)


icu4c/source/i18n/dtptngen.cpp
```diff
@@ -713,19 +713,28 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
...
-    case CAP_H:
-      return UDAT_HOUR_CYCLE_23;
-    case LOW_K:
-      return UDAT_HOUR_CYCLE_24;
-    default:
-      UPRV_UNREACHABLE;
-  }
+DateTimePatternGenerator::getDefaultHourCycle(UErrorCode& status) const {
+    if (U_FAILURE(status)) {
+        return UDAT_HOUR_CYCLE_12;
```
pedberg-icu: I don't suppose it matters much since this is error-case handling, but usually we default to CAP_H / UDAT_HOUR_CYCLE_23 when the locale-specified time cycel is not known

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -713,19 +713,28 @@ void DateTimePatternGenerator::getAllowedHourFormats(const Locale &locale, UErro
...
+    }
+    if (fDefaultHourFormatChar == 0) {
+        // We need to return something, but the caller should ignore it
+        // anyways since the returned status is a failure.
+        status = U_UNSUPPORTED_ERROR;
+        return UDAT_HOUR_CYCLE_12;
```
pedberg-icu: Same comment here about CAP_H / UDAT_HOUR_CYCLE_23

icu4c/source/i18n/dtptngen.cpp
jefgen: Thanks Peter!
I'll updated it to use `UDAT_HOUR_CYCLE_23` instead to be consistent.

jefgen:
Thanks for taking a look @FrankYFTang and @pedberg-icu! :)

PR #1018, opened 2020-03-05
ICU-20915 LocaleMatcher no match: always getSupportedIndex()=-1; remo…

markusicu:
…ve defaultLocaleIndex field; constructor check if locales are equivalent to default, not just equal; simplify locale sorting; minor builder & test deflaking

https://unicode-org.atlassian.net/browse/ICU-20915


markusicu:
> Maybe just re-run build to see if it was a transient problem?

Apparently. I restarted the Travis build and it's all green now.

PR #1021, opened 2020-03-06
ICU-21008 Make ant checkTest run tests in exhaustive mode

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21008
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/build.xml
+++ b/icu4j/build.xml
@@ -266,7 +266,8 @@
             <delete dir="${junit.out.dir}/checkTest"/>
             <mkdir dir="${junit.out.dir}/checkTest"/>
 
-            <junit printsummary="yes">
+            <junit printsummary="yes" fork="yes">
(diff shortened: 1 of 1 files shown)


markusicu:
I would like to see some discussion on the ticket and/or in the TC before we do this.

sffc:
An email has been sent to icu-core, with one "SGTM" and no other comments.  I have accepted the ticket.

sffc:
The commit was misticketted (caught by the bot); I re-wrote the message with the correct ticket number.

sffc:
To be super clear on the effect of this PR:

- `ant check` is unaffected.
- `ant exhaustiveCheck` is unaffected.
- `ant checkTest` is unaffected when the target test is not exhaustive.
- `ant checkTest` will now run target tests even if they are exhaustive; before, it would skip them because they were exhaustive.

PR #1027, opened 2020-03-09
ICU-20626 Update Readme badges to add CI-Valgrind pipeline

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20626
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,14 @@ The ICU project is under the stewardship of [The Unicode Consortium](https://www
 
 ![ICU Logo](./tools/images/iculogo_64.png)
 
-### Build Status (master branch)
+### Build Status (`master` branch)
 
 Build | Status
 ------|-------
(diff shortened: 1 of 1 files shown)


jefgen:
>  Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

Hmm, it seems that `pr: none` in the config file also blocks comment triggers too.
https://github.com/MicrosoftDocs/vsts-docs/issues/4206#issuecomment-493451760
I'll need to create another PR to remote it, then override the config.

jefgen:
Okay, so it looks like in order to have both PR comment builds and batched builds on master, we'll actually need to have two different Pipelines setup on Azure DevOps...
Which means I'll need to force push this again to update the build definition number. (ID=29 is for the PR pipeline, and I'll need to make a new one for master...)
Sorry for all the churn...

PR #1035, opened 2020-03-10
ICU-21019 x86: Add .note.gnu.property section for Intel CET

hjl-tools:
Intel Control-flow Enforcement Technology (CET):

https://software.intel.com/en-us/articles/intel-sdm

contains shadow stack (SHSTK) and indirect branch tracking (IBT).  When
CET is enabled, ELF object files must be marked with .note.gnu.property
section.  GCC provides <cet.h> which can be included in assembly codes
to generate CET maker when compiling with -fcf-protection.

This fixes:

https://unicode-org.atlassian.net/browse/ICU-21019


srl295:
sounds reasonable, please file an issue at https://unicode-org.atlassian.net/

PR #1040, opened 2020-03-11
ICU-21013 Fix OOM not checked in vtzone.cpp

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21013
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/vtzone.cpp
+++ b/icu4c/source/i18n/vtzone.cpp
@@ -526,15 +526,15 @@ rruleParseError:
 static TimeZoneRule* createRuleByRRULE(const UnicodeString& zonename, int rawOffset, int dstSavings, UDate start,
                                        UVector* dates, int fromOffset, UErrorCode& status) {
     if (U_FAILURE(status)) {
-        return NULL;
+        return nullptr;
     }
-    if (dates == NULL || dates->size() == 0) {
+    if (dates == nullptr || dates->size() == 0) {
         status = U_ILLEGAL_ARGUMENT_ERROR;
-        return NULL;
+        return nullptr;
     }
 
     int32_t i, j;
-    DateTimeRule *adtr = NULL;
+    DateTimeRule *adtr = nullptr;
 
     // Parse the first rule
     UnicodeString rrule = *((UnicodeString*)dates->elementAt(0));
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/vtzone.cpp
```diff
@@ -1065,6 +1077,9 @@ VTimeZone::operator!=(const TimeZone& that) const {
 VTimeZone*
 VTimeZone::createVTimeZoneByID(const UnicodeString& ID) {
     VTimeZone *vtz = new VTimeZone();
+    if (vtz == nullptr) {
```
jefgen: The file is somewhat inconsistent in terms of using `NULL` or `nullptr`.
I tried to use the existing convention for each function in order to minimize churn.
However, I could just blanket change to use `nullptr` if that would be more clear.

icu4c/source/i18n/vtzone.cpp
```diff
@@ -1380,21 +1422,23 @@ VTimeZone::parse(UErrorCode& status) {
...
     }
      // Set the deleter to remove TimeZoneRule vectors to avoid memory leaks due to unowned TimeZoneRules.
     rules->setDeleter(deleteTimeZoneRule);
     
     dates = new UVector(uprv_deleteUObject, uhash_compareUnicodeString, status);
-    if (U_FAILURE(status)) {
-        goto cleanupParse;
-    }
-    if (rules == NULL || dates == NULL) {
+    if (rules == NULL) {
```
daniel-ju: `dates`

icu4c/source/i18n/vtzone.cpp
axelandrejs: I personally think nullptr is a case where moving forward is better than consistency.

icu4c/source/i18n/vtzone.cpp
jefgen: Oh good catch. Thanks!
(I'll force-push an update later with NULL->nullptr changes too.)

jefgen: Actually, for review purposes, I'll probably push more commits then squash at the end. (Though it will require more approvals though...)

icu4c/source/i18n/vtzone.cpp
jefgen: Thanks, I'll change to use `nullptr` through out the entire file while I'm at it.

jefgen:
Note: I pushed more commits so they can be reviewed separately (if wanted). I'll squash them all into one commit before merging though.

jefgen:
Note: I squashed the commits into one, so I'll need another approval for the PR.

PR #1041, opened 2020-03-11
ICU-21014 Fix OOM not checked in dtitvfmt.cpp and dtitvinf.cpp

jefgen:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21014
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/dtitvfmt.cpp
+++ b/icu4c/source/i18n/dtitvfmt.cpp
@@ -108,6 +108,10 @@ DateIntervalFormat::createInstance(const UnicodeString& skeleton,
 #endif
 
     DateIntervalInfo* dtitvinf = new DateIntervalInfo(locale, status);
+    if (dtitvinf == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return nullptr;
+    }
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/dtitvfmt.cpp
```diff
@@ -540,6 +552,9 @@ DateIntervalFormat::setDateIntervalInfo(const DateIntervalInfo& newItvPattern,
                                         UErrorCode& status) {
     delete fInfo;
     fInfo = new DateIntervalInfo(newItvPattern);
+    if (fInfo == NULL) {
```
jefgen: The file is somewhat inconsistent in terms of using `NULL` or `nullptr`.
I tried to use the existing convention for each function in order to minimize churn.
However, I could just blanket change to use `nullptr` if that would be more clear.

jefgen:
Thanks for the reviews! I squashed the two commits into one, so this will need one more approval before checking-in... sorry for the double-review.
(It's too bad that GitHub doesn't let you retain the "approved" status if the files in the PR don't change at all...)

PR #1043, opened 2020-03-12
ICU-21016 Special handling of Spanish and Hebrew list format until CLDR get the data

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21016
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/listformatter.cpp
+++ b/icu4c/source/i18n/listformatter.cpp
@@ -21,6 +21,7 @@
 #include "unicode/listformatter.h"
 #include "unicode/simpleformatter.h"
 #include "unicode/ulistformatter.h"
+#include "unicode/uscript.h"
 #include "fphdlimp.h"
 #include "mutex.h"
 #include "hash.h"
@@ -35,34 +36,203 @@
 
 U_NAMESPACE_BEGIN
 
-struct ListFormatInternal : public UMemory {
+namespace {
+
+class PatternHandler : public UObject {
+public:
+    PatternHandler(const UnicodeString& two, const UnicodeString& end, UErrorCode& errorCode) :
+        twoPattern(two, 2, 2, errorCode),
+        endPattern(end, 2, 2, errorCode) {  }
+
+    PatternHandler(const SimpleFormatter& two, const SimpleFormatter& end) :
+        twoPattern(two),
+        endPattern(end) { }
+
+    virtual ~PatternHandler();
+
+    virtual PatternHandler* clone() const { return new PatternHandler(twoPattern, endPattern); }
+
+    virtual const SimpleFormatter& getTwoPattern(const UnicodeString&) const {
+        return twoPattern;
+    }
+
+    virtual const SimpleFormatter& getEndPattern(const UnicodeString&) const {
+        return endPattern;
+    }
+
+protected:
     SimpleFormatter twoPattern;
+    SimpleFormatter endPattern;
+};
+
+PatternHandler::~PatternHandler() {
+}
+
+class ContextualHandler : public PatternHandler {
+public:
+    ContextualHandler(bool (*testFunc)(const UnicodeString& text),
+                      const UnicodeString& thenTwo,
+                      const UnicodeString& elseTwo,
+                      const UnicodeString& thenEnd,
+                      const UnicodeString& elseEnd,
+                      UErrorCode& errorCode) :
+        PatternHandler(elseTwo, elseEnd, errorCode),
+        test(testFunc),
+        thenTwoPattern(thenTwo, 2, 2, errorCode),
+        thenEndPattern(thenEnd, 2, 2, errorCode) {  }
+
+    ContextualHandler(bool (*testFunc)(const UnicodeString& text),
+                      const SimpleFormatter& thenTwo, SimpleFormatter elseTwo,
+                      const SimpleFormatter& thenEnd, SimpleFormatter elseEnd) :
+      PatternHandler(elseTwo, elseEnd),
+      test(testFunc),
+      thenTwoPattern(thenTwo),
+      thenEndPattern(thenEnd) { }
+
+    ~ContextualHandler() override;
+
+    PatternHandler* clone() const override {
+        return new ContextualHandler(
+            test, thenTwoPattern, twoPattern, thenEndPattern, endPattern);
+    }
+
+    const SimpleFormatter& getTwoPattern(
+        const UnicodeString& text) const override {
+        return (test)(text) ? thenTwoPattern : twoPattern;
+    }
+
+    const SimpleFormatter& getEndPattern(
+        const UnicodeString& text) const override {
+        return (test)(text) ? thenEndPattern : endPattern;
+    }
+
+private:
+    bool (*test)(const UnicodeString&);
+    SimpleFormatter thenTwoPattern;
+    SimpleFormatter thenEndPattern;
+};
+
+ContextualHandler::~ContextualHandler() {
+}
+
+static const char16_t *spanishY = u"{0} y {1}";
+static const char16_t *spanishE = u"{0} e {1}";
+static const char16_t *spanishO = u"{0} o {1}";
+static const char16_t *spanishU = u"{0} u {1}";
+static const char16_t *hebrewVav = u"{0} \u05D5{1}";
+static const char16_t *hebrewVavDash = u"{0} \u05D5-{1}";
+
+// Condiction to change to e.
+// Starts with "hi" or "i" but not with "hie" nor "hia"
+static bool shouldChangeToE(const UnicodeString& text) {
+    int32_t len = text.length();
+    if (len == 0) { return false; }
+    // Case insensitive match hi but not hie nor hia.
+    if ((text[0] == u'h' || text[0] == u'H') &&
+            ((len > 1) && (text[1] == u'i' || text[1] == u'I')) &&
+            ((len == 2) || !(text[2] == u'a' || text[2] == u'A' || text[2] == u'e' || text[2] == u'E'))) {
+        return true;
+    }
+    // Case insensitive for "start with i"
+    if (text[0] == u'i' || text[0] == u'I') { return true; }
+    return false;
+}
+
+// Condiction to change to u.
+// Starts with "o", "ho", and "8". Also "11" by itself.
+// re: ^((o|ho|8).*|11)$
+static bool shouldChangeToU(const UnicodeString& text) {
+    int32_t len = text.length();
+    if (len == 0) { return false; }
+    // Case insensitive match o.* and 8.*
+    if (text[0] == u'o' || text[0] == u'O' || text[0] == u'8') { return true; }
+    // Case insensitive match ho.*
+    if ((text[0] == u'h' || text[0] == u'H') &&
+            ((len > 1) && (text[1] == 'o' || text[1] == u'O'))) {
+        return true;
+    }
+    // match "^11$" and "^11 .*"
+    if ((len >= 2) && text[0] == u'1' && text[1] == u'1' && (len == 2 || text[2] == u' ')) { return true; }
+    return false;
+}
+
+// Condiction to change to VAV follow by a dash.
+// Starts with non Hebrew letter.
+static bool shouldChangeToVavDash(const UnicodeString& text) {
+    if (text.isEmpty()) { return false; }
+    UErrorCode status = U_ZERO_ERROR;
+    return uscript_getScript(text.char32At(0), &status) != USCRIPT_HEBREW;
+}
+
+PatternHandler* createPatternHandler(
+        const char* lang, const UnicodeString& two, const UnicodeString& end,
+    UErrorCode& status) {
+    if (uprv_strcmp(lang, "es") == 0) {
+        // Spanish
+        UnicodeString spanishYStr(TRUE, spanishY, -1);
+        bool twoIsY = two == spanishYStr;
+        bool endIsY = end == spanishYStr;
+        if (twoIsY || endIsY) {
+            UnicodeString replacement(TRUE, spanishE, -1);
+            return new ContextualHandler(
+                shouldChangeToE,
+                twoIsY ? replacement : two, two,
+                endIsY ? replacement : end, end, status);
+        }
+        UnicodeString spanishOStr(TRUE, spanishO, -1);
+        bool twoIsO = two == spanishOStr;
+        bool endIsO = end == spanishOStr;
+        if (twoIsO || endIsO) {
+            UnicodeString replacement(TRUE, spanishU, -1);
+            return new ContextualHandler(
+                shouldChangeToU,
+                twoIsO ? replacement : two, two,
+                endIsO ? replacement : end, end, status);
+        }
+    } else if (uprv_strcmp(lang, "he") == 0 || uprv_strcmp(lang, "iw") == 0) {
+        // Hebrew
+        UnicodeString hebrewVavStr(TRUE, hebrewVav, -1);
+        bool twoIsVav = two == hebrewVavStr;
+        bool endIsVav = end == hebrewVavStr;
+        if (twoIsVav || endIsVav) {
+            UnicodeString replacement(TRUE, hebrewVavDash, -1);
+            return new ContextualHandler(
+                shouldChangeToVavDash,
+                twoIsVav ? replacement : two, two,
+                endIsVav ? replacement : end, end, status);
+        }
+    }
+    return new PatternHandler(two, end, status);
+}
+
+}  // namespace
+
+struct ListFormatInternal : public UMemory {
     SimpleFormatter startPattern;
     SimpleFormatter middlePattern;
-    SimpleFormatter endPattern;
+    LocalPointer<PatternHandler> patternHandler;
 
 ListFormatInternal(
         const UnicodeString& two,
         const UnicodeString& start,
         const UnicodeString& middle,
         const UnicodeString& end,
+        const Locale& locale,
         UErrorCode &errorCode) :
-        twoPattern(two, 2, 2, errorCode),
         startPattern(start, 2, 2, errorCode),
         middlePattern(middle, 2, 2, errorCode),
-        endPattern(end, 2, 2, errorCode) {}
+        patternHandler(createPatternHandler(locale.getLanguage(), two, end, errorCode), errorCode) { }
 
 ListFormatInternal(const ListFormatData &data, UErrorCode &errorCode) :
-        twoPattern(data.twoPattern, errorCode),
         startPattern(data.startPattern, errorCode),
         middlePattern(data.middlePattern, errorCode),
-        endPattern(data.endPattern, errorCode) { }
+        patternHandler(createPatternHandler(
+            data.locale.getLanguage(), data.twoPattern, data.endPattern, errorCode), errorCode) { }
 
 ListFormatInternal(const ListFormatInternal &other) :
-    twoPattern(other.twoPattern),
     startPattern(other.startPattern),
     middlePattern(other.middlePattern),
-    endPattern(other.endPattern) { }
+    patternHandler(other.patternHandler->clone()) { }
 };
 
 
@@ -322,7 +492,8 @@ ListFormatInternal* ListFormatter::loadListFormatInternal(
         errorCode = U_MISSING_RESOURCE_ERROR;
         return nullptr;
     }
-    ListFormatInternal* result = new ListFormatInternal(sink.two, sink.start, sink.middle, sink.end, errorCode);
+
+    ListFormatInternal* result = new ListFormatInternal(sink.two, sink.start, sink.middle, sink.end, locale, errorCode);
     if (result == nullptr) {
         errorCode = U_MEMORY_ALLOCATION_ERROR;
         return nullptr;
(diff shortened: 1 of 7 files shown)


FrankYFTang:
Not ready for review yet. Just upload the Java implementing and tests for now. Will add C++ one soon.

FrankYFTang:
Please review, I add dependency of Regex to listformatter. Using regexp to check the condition of replacing pattern. somehow the depstest complain about "new" even after I add the regexp dependency

sffc:
If it's easy (say, less than 2 hours of work) to implement this without the regex engine, then it would be much nicer.  Otherwise, hopefully when we roll in the longer-term solution, we can remove that dependency.

Note that pulling in regex to ListFormatter also pulls it into all of its dependencies, which includes MeasureFormat (although I think NumberFormatter is safe).

FrankYFTang:
> If it's easy (say, less than 2 hours of work) to implement this without the regex engine, then it would be much nicer. Otherwise, hopefully when we roll in the longer-term solution, we can remove that dependency.
> 
> Note that pulling in regex to ListFormatter also pulls it into all of its dependencies, which includes MeasureFormat (although I think NumberFormatter is safe).

Do you mind I use regexp in Java but only use hand crafted scanner in C++?

sffc:
I'd be fine with that since regex is available in the standard library in Java, so we're not adding any new ICU dependencies

FrankYFTang:
> I'd be fine with that since regex is available in the standard library in Java, so we're not adding any new ICU dependencies

but I would need to add at least UnicodeSet, or do you have a better way that I can check a char is in Hebrew script ?

FrankYFTang:
> > I'd be fine with that since regex is available in the standard library in Java, so we're not adding any new ICU dependencies
> 
> but I would need to add at least UnicodeSet, or do you have a better way that I can check a char is in Hebrew script ?

never mind mind, I will use uscript_getScript instead

FrankYFTang:
depstest show error

"Error: in group listformatter: i18n/listformatter.o imports operator new(unsigned long) - is this a new system symbol?"

Which I cannot figure out what cause that

sffc:
> depstest show error
> 
> "Error: in group listformatter: i18n/listformatter.o imports operator new(unsigned long) - is this a new system symbol?"
> 
> Which I cannot figure out what cause that

You need to make your new class PatternHandler extend UMemory, or else `new`ing it pulls in the system allocator instead of the ICU allocator.

FrankYFTang:
> > depstest show error
> > "Error: in group listformatter: i18n/listformatter.o imports operator new(unsigned long) - is this a new system symbol?"
> > Which I cannot figure out what cause that
> 
> You need to make your new class PatternHandler extend UMemory, or else `new`ing it pulls in the system allocator instead of the ICU allocator.

ah... thanks

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+    // Condiction to change to VAV follow by a dash.
+    // Starts with non Hebrew letter.
+    private static final Pattern changeToVavDash = Pattern.compile("^[\\P{InHebrew}].*$");
+
+    // A factory function to create function based on locale
+    // Handle specal case of Spanish and Hebrew
+    private PatternHandler createPatternHandler(String two, String end) {
+        if (this.locale != null) {
+            String language = this.locale.getLanguage();
+            if (language == "es") {
```
sffc: You can't `==` strings in Java.  But, in that case, how are the tests passing?

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+// Condiction to change to e.
+// Starts with "hi" or "i" but not with "hie" nor "hia"
+// re: (?i)^(i|hi[^ae]).*$
+static bool shouldChangeToE(const UnicodeString& text) {
+    int32_t len = text.length();
+    if (len == 0) return false;
+    // Case insensitive match hi[^ae].*
+    if ((text[0] == 'h' || text[0] == 'H') &&
+        ((len > 1) && (text[1] == 'i' || text[1] == 'I')) &&
+        ((len == 2) || !(text[2] == 'a' || text[2] == 'A' || text[2] == 'e' || text[2] == 'E')))
```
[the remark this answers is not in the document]
FrankYFTang: Maybe my re is incorrect, but from the description in the bug "hi" should return true.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+        return true;
+    // Case insensitive match i[^ae].*
+    if ((text[0] == 'i' || text[0] == 'I') &&
+        ((len == 1) || !(text[1] == 'a' || text[1] == 'A' || text[1] == 'e' || text[1] == 'E')))
```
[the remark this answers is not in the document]
FrankYFTang: but "i" should return true. The bug said "Last word starts with hi/i  but not with(`hie` & `hia`)"

icu4c/source/test/intltest/listformattertest.cpp
```diff
@@ -704,4 +707,120 @@ void ListFormatterTest::TestCreateStyled() {
...
+
+        { ULISTFMT_TYPE_OR, u"desierto u oasis",         u"desierto", u"oasis",       nullptr },
+        { ULISTFMT_TYPE_OR, u"oasis, desierto u océano", u"oasis",    u"desierto",    u"océano" },
+        { ULISTFMT_TYPE_OR, u"7 u 8",                    u"7",        u"8",           nullptr },
+        { ULISTFMT_TYPE_OR, u"7 u 80",                   u"7",        u"80",          nullptr },
+        { ULISTFMT_TYPE_OR, u"7 u 800",                  u"7",        u"800",         nullptr },
+        { ULISTFMT_TYPE_OR, u"6, 7 u 8",                 u"6",        u"7",           u"8" },
+        { ULISTFMT_TYPE_OR, u"10 u 11",                  u"10",       u"11",          nullptr },
+        { ULISTFMT_TYPE_OR, u"10 o 111",                 u"10",       u"111",         nullptr },
+        { ULISTFMT_TYPE_OR, u"10 o 11.2",                u"10",       u"11.2",        nullptr },
```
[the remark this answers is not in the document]
FrankYFTang: https://github.com/tc39/proposal-intl-list-format/issues/45#issuecomment-579758548 said
Last word starts with 11, 11.000.000, 11.000, 11.000…….
which is vague 
https://translate.google.com/#view=home&op=translate&sl=en&tl=es&text=10%20or%2011
show me "10 u 11"
https://translate.google.com/#view=home&op=translate&sl=en&tl=es&text=10%20or%20111
"10 o 111"
https://translate.google.com/#view=home&op=translate&sl=en&tl=es&text=10%20or%2011.1
"10 o 11.1"

I am not sure what exactly need to be done here. The research seems incomplete and the language to specify that is still very vague now. I just pick one to implement. And I think we can spend more time to argue about which one is better but that probably won't make into 67.

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
FrankYFTang: good catch- "es" is string literal and getLanguage probably return the same from the string pool. I will change it.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+        ((len > 1) && (text[1] == 'o' || text[1] == 'O'))) return true;
+    // match "^11$" and "^11 .*"
+    if ((len >= 2) && text[0] == '1' && text[1] == '1' && (len == 2 || text[2] == ' ')) return true;
+    return false;
+}
+
+// Condiction to change to VAV follow by a dash.
+// Starts with non Hebrew letter.
+static bool shouldChangeToVavDash(const UnicodeString& text) {
+  if (text.length() == 0) return false;
```
markusicu: Optional: If you don't further use the length, you could check isEmpty().

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+  UErrorCode status = U_ZERO_ERROR;
+  return uscript_getScript(text.char32At(0), &status) != USCRIPT_HEBREW;
+}
+
+PatternHandler* createPatternHandler(
+    const char* lang, const UnicodeString& two, const UnicodeString& end,
```
markusicu: Please indent parameters by another 4 so we see the step to the body indentation.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+    UErrorCode& status) {
+    if (uprv_strcmp(lang, "es") == 0) {
+        // Spanish
+        bool twoIsY = two == UnicodeString(spanishY);
```
markusicu: Please use the Readonly-aliasing char16_t* constructor for such constants.
```
UnicodeString(TRUE, spanishY, -1);
```
Avoids copying the string contents.

https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeString.html#a28a13b2ef3a75b8a947bcd040a98e206

http://userguide.icu-project.org/strings#TOC-Maximizing-Performance-with-the-UnicodeString-Storage-Model

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+        bool endIsY = end == UnicodeString(spanishY);
```
markusicu: When you use the same string more than once please make a local variable.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
 
 U_NAMESPACE_BEGIN
 
+class PatternHandler : public UMemory {
```
markusicu: Please enclose these new classes in an anonymous namespace.

icu4c/source/test/intltest/listformattertest.cpp
```diff
@@ -704,4 +707,120 @@ void ListFormatterTest::TestCreateStyled() {
     }
 }
 
+void ListFormatterTest::TestSpanish() {
+    IcuTestErrorCode status(*this, "TestSpanish");
+    const char* locales [] = {
+      "es", "es_419" , "es_PY", "es_DO"
```
markusicu: indent by increments of 4

Please fix multiple new lines in this file/PR.

icu4c/source/test/intltest/listformattertest.cpp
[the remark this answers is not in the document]
markusicu: I am ok if we go from 80% right to 97% right for now...

icu4c/source/test/intltest/listformattertest.cpp
```diff
@@ -704,4 +707,120 @@ void ListFormatterTest::TestCreateStyled() {
...
+                    const UnicodeString inputs3[] = { cas.data1, cas.data2, cas.data3 };
+                    FormattedList result = fmt->formatStringsToValue(inputs3, UPRV_LENGTHOF(inputs3), status);
+                    assertEquals(message, cas.expected, result.toTempString(status));
+                }
+            }
+        }
+    }
+}
+
+void ListFormatterTest::TestHebrew() {
```
markusicu: This looks like the same code as the spanish test, except for the test data. Why not combine them?

icu4c/source/test/intltest/listformattertest.cpp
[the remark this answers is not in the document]
FrankYFTang: the problem is we do not know which one is 80% right and which one is 97% right

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+    private class ContextualHandler implements PatternHandler {
+        ContextualHandler(Pattern regexp, String thenTwo, String elseTwo, String thenEnd, String elseEnd) {
+            this.regexp = regexp;
+            this.thenTwoPattern = thenTwo;
+            this.elseTwoPattern = elseTwo;
+            this.thenEndPattern = thenEnd;
+            this.elseEndPattern = elseEnd;
+        }
+
+        public String getTwoPattern(String text) {
```
markusicu: `@Override`

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+            if(this.regexp.matcher(text).matches()) {
+                return this.thenTwoPattern;
+            } else {
+                return this.elseTwoPattern;
+            }
+        }
+
+        public String getEndPattern(String text) {
```
markusicu: `@Override`

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+        }
+
+        public String getTwoPattern(String text) { return this.twoPattern; }
+        public String getEndPattern(String text) { return this.endPattern; }
+
+        private final String twoPattern;
+        private final String endPattern;
+    }
+
+    // A contextual hanlder append the items by as static handler, except delegate to appender to
```
markusicu: typos

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+        ContextualHandler(Pattern regexp, String thenTwo, String elseTwo, String thenEnd, String elseEnd) {
+            this.regexp = regexp;
+            this.thenTwoPattern = thenTwo;
+            this.elseTwoPattern = elseTwo;
+            this.thenEndPattern = thenEnd;
+            this.elseEndPattern = elseEnd;
+        }
+
+        public String getTwoPattern(String text) {
+            if(this.regexp.matcher(text).matches()) {
```
markusicu: Please remove unnecessary `this.`

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+
+    // Pattern in the ICU Data which might be replaced o by u.
+    private static final String compiledO = compilePattern("{0} o {1}", new StringBuilder());
+
+    // The new pattern to replace u to o
+    private static final String compiledU = compilePattern("{0} u {1}", new StringBuilder());
+
+    // Condiction to change to e.
+    // Starts with "hi" or "i" but not with "hie" nor "hia"a
+    private static final Pattern changeToE = Pattern.compile("(?i)^((i|hi[^ae]).*|hi)$");
```
markusicu: Optional:
- You don't need the ^ and $ in the patterns because matches() wants to cover the whole string.
- I think you can simplify to `"(?i)(i.*|hi[^ae]*)"`
- even simpler: `Pattern.compile("(i.*|hi[^ae]*)", CASE_INSENSITIVE)`

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+
+    // Condiction to change to u.
+    // Starts with "o", "ho", and "8". Also "11" by itself.
+    private static final Pattern changeToU = Pattern.compile("(?i)^((o|ho|8).*|11)$");
```
markusicu: Optional: `Pattern.compile("((o|ho|8).*|11)", CASE_INSENSITIVE)`

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
+
+    // Pattern in the ICU Data which might need to add a DASH after VAV.
+    private static final String compiledVav = compilePattern("{0} \u05D5{1}", new StringBuilder());
+
+    // Pattern to add a DASH after VAV.
+    private static final String compiledVavDash = compilePattern("{0} \u05D5-{1}", new StringBuilder());
+
+    // Condiction to change to VAV follow by a dash.
```
markusicu: Please fix all "Condiction" typos.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/ListFormatterTest.java
```diff
@@ -292,4 +293,90 @@ public void TestCreateStyled() {
             assertEquals(message, expected, result);
         }
     }
+
+    @Test
+    public void TestSpanish() {
+        String[] locales = { "es", "es_419", "es_PY", "es_DO" };
```
markusicu: As in C++ -- why not combine the spanish & hebrew tests into one function with more test cases?
Less code to review & maintain...

icu4c/source/test/intltest/listformattertest.cpp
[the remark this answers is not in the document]
markusicu: What I meant is, without this fix we get Spanish list formatting right in, say, 80% of cases, and with this fix we should improve to, random number, 97%. I am ok with not yet hitting 100%.

icu4c/source/test/intltest/listformattertest.cpp
FrankYFTang: not really, somehow for the Hebrew case we only change the AND but the Spanish case we need to test both AND and OR and combine them could be very complex and tricky and expand. as you can see, we also need to test each on a list of locale.

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -526,14 +530,127 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
         builder.append(start, it.next(), 1);
         for (int idx = 2; idx < count - 1; ++idx) {
             builder.append(middle, it.next(), idx);
         }
-        return builder.append(end, it.next(), count - 1);
+        Object last = it.next();
+        return builder.append(patternHandler.getEndPattern((String)last), last, count - 1);
+    }
+
+    // A static hanlder append the items without considering the text inside the it
```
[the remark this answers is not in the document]
FrankYFTang: I change a lot on these and forgot to update the comments.

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
FrankYFTang: but ... (i.*|hi[^ae]*) will return hioa false, right?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/ListFormatterTest.java
FrankYFTang: ok

icu4c/source/test/intltest/listformattertest.cpp
FrankYFTang: ok

FrankYFTang:
PTAL

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
```diff
@@ -533,55 +533,60 @@ FormattedListBuilder formatImpl(Collection<?> items, boolean needsFields) {
...
         for (int idx = 2; idx < count - 1; ++idx) {
             builder.append(middle, it.next(), idx);
         }
         Object last = it.next();
-        return builder.append(patternHandler.getEndPattern((String)last), last, count - 1);
+        return builder.append(patternHandler.getEndPattern(String.valueOf(last)), last, count - 1);
     }
 
-    // A static hanlder append the items without considering the text inside the it
+    // A static handler just return the pattern without consider the input text.
```
markusicu: A static handler just returns the pattern without considering the input text.

icu4j/main/classes/core/src/com/ibm/icu/text/ListFormatter.java
markusicu: You are right, sorry for my bad regex. Your new version looks good.

markusicu:
PS: Make sure you look for "hidden conversations" on the $#@! GitHub UI and toggle them to be visible.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+static const char16_t *hebrewVav = u"{0} \u05D5{1}";
+static const char16_t *hebrewVavDash = u"{0} \u05D5-{1}";
+
+// Condiction to change to e.
+// Starts with "hi" or "i" but not with "hie" nor "hia"
+static bool shouldChangeToE(const UnicodeString& text) {
+    int32_t len = text.length();
+    if (len == 0) return false;
+    // Case insensitive match hi but not hie nor hia.
+    if ((text[0] == 'h' || text[0] == 'H') &&
```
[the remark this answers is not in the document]
FrankYFTang: gee.... ok

icu4c/source/i18n/listformatter.cpp
```diff
@@ -35,34 +36,193 @@
...
+static bool shouldChangeToU(const UnicodeString& text) {
+    int32_t len = text.length();
+    if (len == 0) return false;
+    // Case insensitive match o.* and 8.*
+    if (text[0] == 'o' || text[0] == 'O' || text[0] == '8') return true;
+    // Case insensitive match ho.*
+    if ((text[0] == 'h' || text[0] == 'H') &&
+        ((len > 1) && (text[1] == 'o' || text[1] == 'O'))) return true;
+    // match "^11$" and "^11 .*"
+    if ((len >= 2) && text[0] == '1' && text[1] == '1' && (len == 2 || text[2] == ' ')) return true;
```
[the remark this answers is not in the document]
FrankYFTang: that is the part we are not sure which one will be better.

FrankYFTang:
PTAL

icu4c/source/i18n/listformatter.cpp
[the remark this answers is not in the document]
markusicu: ok, i guess we'll see

markusicu:
Thanks Frank, lgtm now.
@sffc you still have unresolved questions. Are you ok with where we are?

sffc:
(lgtm please squash)

markusicu:
Thanks Shane. Frank please squash for final approval.

PR #1044, opened 2020-03-13
ICU-20984 StringPiece & ByteSink overloads for char8_t*

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20984

This is a follow-up to ICU-20972 PR #979 which fixed `char8_t *` build failures just using reinterpret_cast so that we could include it in ICU 66.

Here are some new API overloads to make working with C++20 u8"string literals" easier while keeping ICU compatible with C++11.

This includes one not originally proposed overload for `StringPiece(std::nullptr_t p)` to type-disambiguate uses of `StringPiece(nullptr)` especially when the constructor call is implicit. I decided not to add such overloads for every pointer variation here; see the PR comment thread.

Note that `StringPiece(NULL)` call sites may break, depending on the compiler. In portable code this is unavoidable, as in the adoption of char16_t (see http://site.icu-project.org/download/59). Please change such call sites from using NULL to using nullptr.

FYI @stac47 @tahonermann

I tested this with Linux clang 8 `-std=c++2a`.

### DIFF
--- a/icu4c/source/common/unicode/bytestream.h
+++ b/icu4c/source/common/unicode/bytestream.h
@@ -71,6 +71,40 @@ public:
    */
   virtual void Append(const char* bytes, int32_t n) = 0;
 
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Appends n bytes to this. Same as Append().
+   * Call AppendU8() with u8"string literals" which are const char * in C++11
+   * but const char8_t * in C++20.
+   * If the compiler does support char8_t as a distinct type,
+   * then an AppendU8() overload for that is defined and will be chosen.
+   *
+   * @param bytes the pointer to the bytes
+   * @param n the number of bytes; must be non-negative
+   * @draft ICU 67
+   */
+  inline void AppendU8(const char* bytes, int32_t n) {
+    Append(bytes, n);
+  }
+
+#if defined(__cpp_char8_t) || defined(U_IN_DOXYGEN)
+  /**
+   * Appends n bytes to this. Same as Append() but for a const char8_t * pointer.
+   * Call AppendU8() with u8"string literals" which are const char * in C++11
+   * but const char8_t * in C++20.
+   * If the compiler does support char8_t as a distinct type,
+   * then this AppendU8() overload for that is defined and will be chosen.
+   *
+   * @param bytes the pointer to the bytes
+   * @param n the number of bytes; must be non-negative
+   * @draft ICU 67
+   */
+  inline void AppendU8(const char8_t* bytes, int32_t n) {
+    Append(reinterpret_cast<const char*>(bytes), n);
+  }
+#endif
+#endif  // U_HIDE_DRAFT_API
+
   /**
    * Returns a writable buffer for appending and writes the buffer's capacity to
    * *result_capacity. Guarantees *result_capacity>=min_capacity.
--- a/icu4c/source/test/intltest/regextst.cpp
+++ b/icu4c/source/test/intltest/regextst.cpp
@@ -31,6 +31,7 @@
 
 #include "unicode/localpointer.h"
 #include "unicode/regex.h"
+#include "unicode/stringpiece.h"
 #include "unicode/uchar.h"
 #include "unicode/ucnv.h"
 #include "unicode/uniset.h"
@@ -5838,11 +5839,11 @@ void RegexTest::TestBug12884() {
     REGEX_ASSERT(status == U_REGEX_TIME_OUT);
 
     // UText, wrapping non-UTF-16 text, also takes a different execution path.
-    const char *text8 = reinterpret_cast<const char*>(u8"¿Qué es Unicode?  Unicode proporciona un número único para cada"
+    StringPiece text8(u8"¿Qué es Unicode?  Unicode proporciona un número único para cada"
                           "carácter, sin importar la plataforma, sin importar el programa,"
                           "sin importar el idioma.");
     status = U_ZERO_ERROR;
-    LocalUTextPointer ut(utext_openUTF8(NULL, text8, -1, &status));
+    LocalUTextPointer ut(utext_openUTF8(NULL, text8.data(), text8.length(), &status));
     REGEX_CHECK_STATUS;
     m.reset(ut.getAlias());
     m.find(status);
--- a/icu4c/source/test/intltest/strcase.cpp
+++ b/icu4c/source/test/intltest/strcase.cpp
@@ -1314,7 +1314,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
     Edits edits;
 
     int32_t length = CaseMap::utf8ToLower("tr", U_OMIT_UNCHANGED_TEXT,
-                                          reinterpret_cast<const char*>(u8"IstanBul"), 8, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                                          reinterpret_cast<const char*>(u8"IstanBul"), 8,
+                                          dest, UPRV_LENGTHOF(dest), &edits, errorCode);
     assertEquals(u"toLower(IstanBul)", UnicodeString(u"ıb"),
                  UnicodeString::fromUTF8(StringPiece(dest, length)));
     static const EditChange lowerExpectedChanges[] = {
@@ -1330,7 +1331,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
 
     edits.reset();
     length = CaseMap::utf8ToUpper("el", U_OMIT_UNCHANGED_TEXT,
-                                  reinterpret_cast<const char*>(u8"Πατάτα"), 6 * 2, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                                  reinterpret_cast<const char*>(u8"Πατάτα"), 6 * 2,
+                                  dest, UPRV_LENGTHOF(dest), &edits, errorCode);
     assertEquals(u"toUpper(Πατάτα)", UnicodeString(u"ΑΤΑΤΑ"),
                  UnicodeString::fromUTF8(StringPiece(dest, length)));
     static const EditChange upperExpectedChanges[] = {
@@ -1370,7 +1372,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
     // No explicit nor automatic edits.reset(). Edits should be appended.
     length = CaseMap::utf8Fold(U_OMIT_UNCHANGED_TEXT | U_EDITS_NO_RESET |
                                    U_FOLD_CASE_EXCLUDE_SPECIAL_I,
-                               reinterpret_cast<const char*>(u8"IßtanBul"), 1 + 2 + 6, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                               reinterpret_cast<const char*>(u8"IßtanBul"), 1 + 2 + 6,
+                               dest, UPRV_LENGTHOF(dest), &edits, errorCode);
     assertEquals(u"foldCase(IßtanBul)", UnicodeString(u"ıssb"),
                  UnicodeString::fromUTF8(StringPiece(dest, length)));
     static const EditChange foldExpectedChanges[] = {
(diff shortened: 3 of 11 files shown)


roubert:
I believe that the compiler is actually working as intended. You see, this:

`
131  StringPiece(T str)
`

and this:

`
159  StringPiece(T str)
`

both look like the very same type signature to the compiler (and, as far as I know, this is intentionally so). The fact that either one of them or both of them can be enabled or disabled based on conditionals doesn't change the fact that in case both of them get enabled they will have identical type signatures, `T`.

So let's take a look at what the problem is that you really want to solve here: In case someone uses a C++20 compiler and passes in a pointer of `char8_t` type, then you want to `reinterpret_cast<const char*>` this pointer.

My suggestion would be to simply solve this using the preprocessor and always typecast whenever the compiler knows `char8_t` for it does no harm to typecast to `const char*` unnecessarily if that's the input type already.

Something like this:

```
  StringPiece(T str)
      : ptr_(
#if defined(__cpp_char8_t)
          reinterpret_cast<const char*>(str.data())
#else
          str.data()
#endif
        ),
        length_(static_cast<int32_t>(str.size())) {}
```

(Or even skip the conditional altogeher and just always typecast, even for pre-C++20 compilers where that couldn't possibly ever be necessary.) That ought to be a workable solution to that problem.

I did however find another issue while looking at your code, for the methods that take pointer parameters:

```
77  StringPiece(const char* str)
85  StringPiece(const char8_t* str)
```

... etc., etc., any call with a `nullptr` argument will now become ambiguous. The most obvious solution to that would be to also add `std::nullptr_t` overloads for all of them.

roubert:
Regarding the SFINAE conditional syntax you asked about, it should really be sufficient to do like this:

```
  template <typename T,
            typename = typename std::enable_if<
                (
                    std::is_same<decltype(T().data()), const char*>::value
#if defined(__cpp_char8_t)
                 || std::is_same<decltype(T().data()), const char8_t*>::value
#endif
                ) &&
                std::is_same<decltype(T().size()), size_t>::value>::type>
  StringPiece(T str)
```

markusicu:
Thanks!

markusicu:
> I did however find another issue while looking at your code, for the methods that take pointer parameters:
> 
> ```
> 77  StringPiece(const char* str)
> 85  StringPiece(const char8_t* str)
> ```
> 
> ... etc., etc., any call with a `nullptr` argument will now become ambiguous. The most obvious solution to that would be to also add `std::nullptr_t` overloads for all of them.

Right. I added one overload for `StringPiece(std::nullptr_t p)` because that is actually being used.

However, I did not add such overloads everywhere because it seems silly to promote call sites for `StringPiece(nullptr, 99)` (danger!) or `sp.set(nullptr)` or `sp.set(nullptr, 99)`. If someone really wants to make a StringPiece empty, they should call clear() rather than use a nullptr literal.

markusicu:
PTAL

icu4c/source/common/unicode/bytestream.h
```diff
@@ -71,6 +71,40 @@ class U_COMMON_API ByteSink : public UMemory {
...
+   * but const char8_t * in C++20.
+   * If the compiler does support char8_t as a distinct type,
+   * then this AppendU8() overload for that is defined and will be chosen.
+   *
+   * @param bytes the pointer to the bytes
+   * @param n the number of bytes; must be non-negative
+   * @draft ICU 67
+   */
+  inline void AppendU8(const char8_t* bytes, int32_t n) {
+    Append(reinterpret_cast<const char*>(bytes), n);
```
jefgen: Just checking my own understanding:
The reason that this *doesn't* need to use something like `U_ALIASING_BARRIER` is because `char*` is allowed to alias anything, right?

markusicu: I hadn't even thought of that... but the C++ standards people themselves are recommending this simple cast: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1423r0.html

And yes, the justification there mentions that char pointers can alias other things.

markusicu:
Ping someone review please :-)

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -1330,7 +1331,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
 
     edits.reset();
     length = CaseMap::utf8ToUpper("el", U_OMIT_UNCHANGED_TEXT,
-                                  reinterpret_cast<const char*>(u8"Πατάτα"), 6 * 2, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                                  reinterpret_cast<const char*>(u8"Πατάτα"), 6 * 2,
```
tahonermann: I suspect this `reinterpret_cast` is no longer needed (such uses were removed in similar tests)

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -1370,7 +1372,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
     // No explicit nor automatic edits.reset(). Edits should be appended.
     length = CaseMap::utf8Fold(U_OMIT_UNCHANGED_TEXT | U_EDITS_NO_RESET |
                                    U_FOLD_CASE_EXCLUDE_SPECIAL_I,
-                               reinterpret_cast<const char*>(u8"IßtanBul"), 1 + 2 + 6, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                               reinterpret_cast<const char*>(u8"IßtanBul"), 1 + 2 + 6,
```
tahonermann: I suspect this `reinterpret_cast` is no longer needed (such uses were removed in similar tests)

icu4c/source/test/intltest/strtest.cpp
```diff
@@ -265,7 +267,7 @@ StringTest::TestStringPiece() {
         errln("StringPiece() failed");
     }
     // Construct from NULL const char * pointer.
-    StringPiece null(NULL);
+    StringPiece null((const char *)nullptr);
```
tahonermann: I think the cast here should be unnecessary.

icu4c/source/test/intltest/uts46test.cpp
```diff
@@ -160,7 +160,7 @@ void UTS46Test::TestAPI() {
     char buffer[100];
     TestCheckedArrayByteSink sink(buffer, UPRV_LENGTHOF(buffer));
     errorCode=U_ZERO_ERROR;
-    nontrans->labelToUnicodeUTF8(StringPiece(NULL, 5), sink, info, errorCode);
+    nontrans->labelToUnicodeUTF8(StringPiece((const char *)NULL, 5), sink, info, errorCode);
```
tahonermann: I think this cast should be unnecessary.

tahonermann:
I see that a solution has been implemented that addresses the ambiguous overload situation with `nullptr`.  There is an additional option you can consider that works for both `nullptr` and `NULL`.  It is a bit ugly though and won't work in all cases as it relies on overload resolution preferring non-template functions over function templates.

Given these declarations:
```
void f(const char*);
void f(const char8_t*);
```
the following calls are ambiguous:
```
  f(nullptr);
  f(NULL);
  f(0);
```
By converting the `char8_t`-based overload to a function template, overload resolution will prefer the non-template such that the above three calls will resolve to the `char`-based overload:
```
void f(const char*);
template<typename = void> void f(const char8_t*);
```
Of course, this requires that the implementation of the function template be inline in the definition and that isn't necessarily possible or desirable.  There are also a few ways to deal with that:
```
// Dispatch directly to the non-char8_t overload using reinterpret_cast:
template<typename = void> void f(const char8_t* p) {
  f(reinterpret_cast<const char*>(p));
}

// Dispatch to a tagged overload:
enum u8_tag { use_u8 };
void f(u8_tag, const char8_t*);
template<typename = void> void f(const char8_t* p) {
  f(use_u8, p);
}

// Dispatch to a differently named non-template function:
void fu8(const char8_t*);
template<typename = void> void f(const char8_t* p) {
  fu8(p);
}
```

icu4c/source/test/intltest/regextst.cpp
```diff
@@ -5838,11 +5839,11 @@ void RegexTest::TestBug12884() {
     REGEX_ASSERT(status == U_REGEX_TIME_OUT);
 
     // UText, wrapping non-UTF-16 text, also takes a different execution path.
-    const char *text8 = reinterpret_cast<const char*>(u8"¿Qué es Unicode?  Unicode proporciona un número único para cada"
+    StringPiece text8(u8"¿Qué es Unicode?  Unicode proporciona un número único para cada"
```
Quuxplusone: Nit: You could still use `StringPiece text8 = u8"...";` here, right? There's no _technical_ reason to switch to a direct-initialization with parentheses?

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -1314,7 +1314,8 @@ void StringCaseTest::TestCaseMapUTF8WithEdits() {
     Edits edits;
 
     int32_t length = CaseMap::utf8ToLower("tr", U_OMIT_UNCHANGED_TEXT,
-                                          reinterpret_cast<const char*>(u8"IstanBul"), 8, dest, UPRV_LENGTHOF(dest), &edits, errorCode);
+                                          reinterpret_cast<const char*>(u8"IstanBul"), 8,
```
Quuxplusone: @tahonermann didn't comment on this line, but I assume this `reinterpret_cast` is no longer needed for the same reason as the one on line 1334. (But on first glance I don't see how line 1317 is testing anything different from line 1457. So maybe one of the pair is _supposed_ to be testing `const char*` and the other to be testing `const char8_t*`?)

Do you _want_ test coverage for both `char` and `char8_t`? If so, maybe consider making the whole test case `template<class CharT>`, reinterpret-casting all your string literals to `const CharT*`, and then instantiating the whole thing once for `char` and once for `char8_t`. I suspect that's overkill, personally, but I dunno.

icu4c/source/test/intltest/strtest.cpp
Quuxplusone: My understanding was that the cast is needed, and that in this case the overload-resolution ambiguity is a feature, not a bug, because we do not want the user casually to express `StringPiece(nullptr)`.  `StringPiece(charptr)` where `charptr == nullptr` shouldn't crash or anything, but a literal `StringPiece(nullptr)` is not one of the intended use-cases.

However, line 555 does say `StringPiece null(nullptr);` and it Just Works... so my understanding must be wrong. Either the cast is unnecessary here, or else it's needed on line 555.

icu4c/source/test/intltest/strtest.cpp
```diff
@@ -653,6 +701,20 @@ StringTest::TestStringByteSink() {
...
+    // ICU-20984 "mitigate some C++20 char8_t breakages"
+    // For the following APIs there are overloads for both
+    // const char * and const char8_t *.
+    // A u8"string literal" has one type or the other
+    // depending on C++ version and compiler settings.
+    std::string result("abc");
+    StringByteSink<std::string> sink(&result);
+    sink.AppendU8("def", 3);
+    sink.AppendU8(u8"ghijkl", 4);
+    assertEquals("abcdefghij", "abcdef\x67\x68\x69\x6a", result.c_str());
```
Quuxplusone: `\x67` is a funny way to spell `g`; isn't this merely testing the compiler, rather than testing this library?
Alternatively, should line 714 `sink.AppendU8(u8"\u00e9")` and check that the bytes actually appended, even in C++11, are `\xc3\xa9`? (I think that would still be testing the compiler rather than this library, but at least it seems more likely to expose some weird dependency on the current locale or something.)

icu4c/source/test/intltest/tstnorm.cpp
```diff
@@ -1607,12 +1608,12 @@ BasicNormalizerTest::TestNormalizeUTF8WithEdits() {
...
 
     // Omit unchanged text.
-    expected = reinterpret_cast<const char*>(u8"aääạ\u0308ạ\u0308가각갃");
+    expected = u8"aääạ\u0308ạ\u0308가각갃";
     result.clear();
     edits.reset();
     nfkc_cf->normalizeUTF8(U_OMIT_UNCHANGED_TEXT, src, sink, &edits, errorCode);
     assertSuccess("normalizeUTF8 omit unchanged", errorCode.get());
-    assertEquals("normalizeUTF8 omit unchanged", expected.c_str(), result.c_str());
+    assertEquals("normalizeUTF8 omit unchanged", expected.data(), result.c_str());
```
Quuxplusone: You're not passing a `.data(), .size()` pair here, which I think means you're expecting both `expected` and `result` to be null-terminated C-style strings, which means `.c_str()` is semantically correct. For example, if you changed `expected` to be a `string_view` or `StringPiece`, this code would no longer work, because those types aren't logically guaranteed to be null-terminated.

If you want to use non-null-terminated string pieces here, you'll have to find a way to pass the number of chars to compare (i.e., pass both `.data` and `.size`, or pass a `string_view` or `StringPiece` that wraps up those two pieces of information into a single object).

icu4c/source/test/intltest/tstnorm.cpp
```diff
@@ -1623,12 +1624,12 @@ BasicNormalizerTest::TestNormalizeUTF8WithEdits() {
...
     UnicodeSet filter(u"[^A]", errorCode);
     FilteredNormalizer2 fn2(*nfkc_cf, filter);
-    expected = reinterpret_cast<const char*>(u8"  AäA\u0308A\u0323\u0308ạ\u0308,가각갃  ");
+    expected = u8"  AäA\u0308A\u0323\u0308ạ\u0308,가각갃  ";
     result.clear();
     edits.reset();
     fn2.normalizeUTF8(0, src, sink, &edits, errorCode);
     assertSuccess("filtered normalizeUTF8", errorCode.get());
-    assertEquals("filtered normalizeUTF8", expected.c_str(), result.c_str());
+    assertEquals("filtered normalizeUTF8", expected.data(), result.c_str());
```
Quuxplusone: Same comment as above, re `.c_str()`-correctness.

icu4c/source/test/intltest/tstnorm.cpp
```diff
@@ -1655,12 +1656,12 @@ BasicNormalizerTest::TestNormalizeUTF8WithEdits() {
...
     // Note that the result is not normalized because the inner normalizer
     // does not see text across filter spans.
-    expected = reinterpret_cast<const char*>(u8"ä\u0323\u0308ạ\u0308가각갃");
+    expected = u8"ä\u0323\u0308ạ\u0308가각갃";
     result.clear();
     edits.reset();
     fn2.normalizeUTF8(U_OMIT_UNCHANGED_TEXT, src, sink, &edits, errorCode);
     assertSuccess("filtered normalizeUTF8 omit unchanged", errorCode.get());
-    assertEquals("filtered normalizeUTF8 omit unchanged", expected.c_str(), result.c_str());
+    assertEquals("filtered normalizeUTF8 omit unchanged", expected.data(), result.c_str());
```
Quuxplusone: Same comment as above, re `.c_str()`-correctness.

icu4c/source/test/intltest/tstnorm.cpp
```diff
@@ -1777,13 +1778,13 @@ BasicNormalizerTest::TestComposeJamoTBase() {
...
-    std::string expected8(reinterpret_cast<const char*>(u8"가\u11A7가\u11A7가\u11A7"));
+    StringPiece s8(u8"\u1100\u1161\u11A7\u1100\u314F\u11A7가\u11A7");
+    StringPiece expected8(u8"가\u11A7가\u11A7가\u11A7");
     std::string result8;
-    StringByteSink<std::string> sink(&result8, static_cast<int32_t>(expected8.length()));
+    StringByteSink<std::string> sink(&result8, expected8.length());
     nfkc->normalizeUTF8(0, s8, sink, nullptr, errorCode);
     assertSuccess("normalizeUTF8(LV+11A7)", errorCode.get());
-    assertEquals("normalizeUTF8(LV+11A7)", expected8.c_str(), result8.c_str());
+    assertEquals("normalizeUTF8(LV+11A7)", expected8.data(), result8.c_str());
```
Quuxplusone: Same comment as above, re `.c_str()`-correctness.

icu4c/source/test/intltest/strcase.cpp
aheninger: I don't think there's a char8_t overload for CaseMap::utf8ToLower(), so the cast remains necessary.

icu4c/source/test/intltest/strcase.cpp
tahonermann: On further reflection, I think I was wrong here; the invoked overload is one that has a corresponding `const char*` parameter type.  The other tests where `reinterpret_cast` was removed correspond to cases that use `StringPiece`.

icu4c/source/test/intltest/strcase.cpp
tahonermann: On further reflection, I think I was wrong here; the invoked overload is one that has a corresponding `const char*` parameter type.  The other tests where `reinterpret_cast` was removed correspond to cases that use `StringPiece`.

icu4c/source/test/intltest/strcase.cpp
markusicu: Right. We have two versions of these functions, taking pointer+length vs. StringPiece. These unit tests are both for functionality and for API coverage.

icu4c/source/test/intltest/strtest.cpp
markusicu: I found that we do need a cast for NULL which is implementation-dependent, and got reminded of the same issue in the ICU 59 adoption of char16_t and our adding similar overloads for that. I must have changed this line twice, one to add a cast for NULL where that was needed, and once to replace NULL with nullptr where that was ok for test coverage. Either way, this line continues to test with a null pointer input.

I am pretty sure we want to not break call sites that use a nullptr on purpose but the parameter is actually a StringPiece, so they have been relying on the implicit conversion.

markusicu:
> By converting the `char8_t`-based overload to a function template, overload resolution will prefer the non-template such that the above three calls will resolve to the `char`-based overload:

Wow... very clever, but extremely unintuitive, and that would require a big comment for the next person to look at this. I don't think any normal programmer understands this level of voodoo.

I think I will stick with telling people that if their code breaks with NULL they should change it to nullptr. Unless there is a large amount of outrage.

icu4c/source/test/intltest/regextst.cpp
markusicu: That's right, but this here is not a unit test of StringPiece, just a test of something where the input wants to include non-ASCII characters. I don't think it matters whether I use assignment or construction.

icu4c/source/test/intltest/strcase.cpp
markusicu: As Tom found, we have two versions of the API, pointer+length vs. StringPiece. This unit test covers functionality and API.

https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1CaseMap.html

I don't intend to add char8_t and nullptr_t overloads everywhere. We might add some more, but probably not comprehensively.

icu4c/source/test/intltest/strtest.cpp
markusicu: The test has to work on a machine where string literals are not ASCII-based.

`AppendU8("def", 3)` appends system-charset bytes while `AppendU8(u8"ghijkl", 4)` appends UTF-8 bytes. I have to test the bytes that I put in. The sink does not convert.

AFAIK there is not currently a compiler for EBCDIC-based z/OS and whatever "i" is called now, but we don't want to close the door on those yet.

icu4c/source/test/intltest/tstnorm.cpp
markusicu: Right. StringPiece does not imply NUL-termination, but in this code I know what I am initializing the StringPieces with.

icu4c/source/test/intltest/uts46test.cpp
markusicu: I need the cast for NULL. I didn't add a nullptr_t overload for this constructor because I don't want to promote the use of nullptr with an explicit length, but this particular unit test case is for coverage of handling of bad input.

tahonermann:
> very clever, but extremely unintuitive

That is why I offered it as an option and not as a recommendation :)

> I think I will stick with telling people that if their code breaks with NULL they should change it to nullptr. Unless there is a large amount of outrage.

That seems quite reasonable to me.

tahonermann:
> AFAIK there is not currently a compiler for EBCDIC-based z/OS

There are at least two now:
1. IBM recently started including a build of Clang with their xlC compiler for z/OS and announced that this represents their future compiler strategy.  https://developer.ibm.com/linuxonpower/2020/02/23/ibm-c-c-and-fortran-compilers-to-adopt-llvm-open-source-infrastructure.
2. Dignus provides an LLVM based C and C++ compilers (including cross compilers) for z/OS that support C++14 and parts of C++17.  http://www.dignus.com/dcxx.

PR #1045, opened 2020-03-13
ICU-20987 integrate CLDR release-37-alpha3 to master

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20987
- [x] Issue accepted
- [x] Tests included

This integrates the CLDR 37 data tagged release-37-alpha3 to ICU master for ICU 67.
There are several things to note:
1. Several measurement units were renamed, listed below. I adjusted the MeasureUnitTest.java tools that generate create/get methods to compensate and re-generated (resulting in some internal changes in measunit.cpp); there are some questions to be resolved (such as will there be new functions corresponding to the updated unit name strings?):
  * meter-per-second-squared → meter-per-square-second
  * part-per-million → permillion
  * liter-per-100kilometers → liter-per-100-kilometer
  * inch-hg → inch-ofhg
  * millimeter-of-mercury → millimeter-ofhg
  * pound-per-square-inch → pound-force-per-square-inch
  * pound-foot → pound-force-foot
2. Fixed an error in the LocaleIdResolver.java build tool in which it was not automatically including the regional locales for 3-letter language codes; and then deleted the now-redundant locale inclusions in build-icu-data.xml
3. The territoryInfo for ms_ID was changed to favor Latn over Arab, resulting in a change to likelySubtags and thus a change to the langInfo data in ICU.
4. A region locale was added for ms_ID.


pedberg-icu:
> Thanks! What you did with the unit identifiers looks good. I would say that we keep the API names the same, and grandfather them into the tool somehow.

Sounds good with the API names, but I realized one problem with the current PR is that the type strings in measunit.cpp are no longer in alphabetical order within category as required (permillion is out of place). So I need to regenerate the mapping of those strings to index values.

pedberg-icu:
> ...  I realized one problem with the current PR is that the type strings in measunit.cpp are no longer in alphabetical order within category as required (permillion is out of place). So I need to regenerate the mapping of those strings to index values.

OK, I fixed that and adjusted the tools in MeasureUnitTest.java that generate all of the create/get methods to compensate for the new CLDR names.

hagbard:
LGTM for the tooling changes. And thank you for doing this, it looks much better now (I'm embarrassed I missed that).

pedberg-icu:
The exhaustive test failures here were already occurring in the version of master used as the starting point for this branch (see https://unicode-icu.visualstudio.com/ICU/_build?definitionId=24&_a=summary), they are not related to this PR.

I have squashed the two commits so this needs re-approval, thanks.

PR #1047, opened 2020-03-13
ICU-21009 LocaleMatcher() order[] always set all elements

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21009

FYI I introduced this problem just recently in PR #1018.


markusicu:
FYI I had run `valgrind ./intltest utility/LocaleMatcherTest` before & after the fix to verify the bug and the fix.

jefgen:
> FYI I had run `valgrind ./intltest utility/LocaleMatcherTest` before & after the fix to verify the bug and the fix.

Thanks Markus! :)

PR #1048, opened 2020-03-14
ICU-20600 build icu-data-bin files

srl295:
[ICU-20600]

make dist will now produce:
- icu4c-67.1-data-bin-b.zip
- icu4c-67.1-data-bin-l.zip

I made the builder a separate shell script, so that I can run it on backlevel ICUs.

[ICU-20600]: https://unicode-org.atlassian.net/browse/ICU-20600

PR #1049, opened 2020-03-14
ICU-20605 travis: make dist.

srl295:
[ICU-20605]

- run 'make dist' after the linux clang build
~- upgrade ubuntu xenial (16 LTS) to bionic (18 LTS) [(schedule)](https://wiki.ubuntu.com/Releases)~

HT: https://github.com/unicode-org/icu/pull/1048#discussion_r392544485

[ICU-20605]: https://unicode-org.atlassian.net/browse/ICU-20605

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,8 @@ matrix:
         - CFLAGS="-Wimplicit-fallthrough"
         - CXXFLAGS="-Wimplicit-fallthrough"
       compiler: clang
+      before_install:
+        - sudo apt-get -y install doxygen  # for make dist
       before_script:
         - cd icu4c/source
         - ./runConfigureICU Linux
@@ -87,6 +89,7 @@ matrix:
         - make -j2 check
         - ( cd test/depstest && python3 depstest.py ../../../source/ )
         - ( cd .. && source/test/hdrtst/testtagsguards.sh )
+        - make dist # not sure if this is -j safe everywhere.
 
     - name: "c: osx clang"
       language: cpp


srl295:
This should maybe be two PRs. One for make dist, and one for bumping the version.

Or should we keep trying ubuntu xenial?

.travis.yml
```diff
@@ -5,7 +5,7 @@ matrix:
     #    TODO: publish the latest built docs from master on GitHub pages.
     #          See https://gist.github.com/vidavidorra/548ffbcdae99d752da02 for ideas.
     - name:     "c: docs"
-      dist:     xenial
+      dist:     bionic
```
jefgen: oof, should likely update (or add) to azure pipelines bionic as well

.travis.yml
```diff
@@ -87,6 +87,7 @@ matrix:
         - make -j2 check
         - ( cd test/depstest && python3 depstest.py ../../../source/ )
         - ( cd .. && source/test/hdrtst/testtagsguards.sh )
+        - make dist # not sure if this is -j safe everywhere.
```
jefgen: does travis let you download artifacts?

srl295: not sure. We could push them somewhere (like a github release) though.

jefgen:
I'd vote for a separate ticket to bump the version. We should probably bump the azure pipelines one too at the same time. (or add an extra build line)

.travis.yml
jefgen: fwiw, you can publish artifacts from the pipelines.
ex: https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-pipelines.yml#L108
(This is how the windows binaries are produced.)
It should work for *nix as well, so we might be able to have a sort of rolling/evergreen make dist available.

jefgen: It looks like Travis can potentially directly deploy to a GitHub release:
https://docs.travis-ci.com/user/build-stages/deploy-github-releases/
which is definitely nicer.
AFAIK, the pipelines just give you an arbitrary file you can download. (So you'd have to copy/paste to a release).

srl295: right. But I'm not sure what release it would go to.

sffc:
Your new check is already failing

```
make[3]: *** No rule to make target 'icu-docs.zip'.  Stop.
```

PR #1052, opened 2020-03-20
ICU-20976 GCC8 fixes

srl295:
[ICU-20976]

ICU-20976 GCC8 fixes phase 1

Some initial fixes for GCC8

- setup a gcc8 buildbot with -Wextra
- rewrite ucol_sit to use icu::CharString
- workaround for gcc7+ on mac
  see https://github.com/arbor-sim/arbor/issues/562#issuecomment-409970434
- fix ucnv_2022 strcpy site

[ICU-20976]: https://unicode-org.atlassian.net/browse/ICU-20976

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,24 @@ matrix:
       after_failure:
           -  cat `find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;`
 
+    - name: "c: linux gcc 8 Wextra"
+      dist:     bionic
+      language: cpp
+      addons:
+         apt:
+           packages:
+             - g++-8
+      env: PREFIX=/tmp/icu-prefix CC=gcc-8 CXX=g++-8 CXXFLAGS="-Wextra"
+      before_script:
+        - mkdir build
+        - cd build
+        - ../icu4c/source/runConfigureICU Linux --disable-debug --enable-release --disable-layoutex --prefix="${PREFIX}" || (tail -100 config.log ; false)
+        - make -j2
+      script:
+        - make -j2 check
+        - make install
+        - PATH="${PREFIX}/bin:$PATH" make -C test/hdrtst check
+
     # gcc debug build.
     # Includes dependency checker.
     # Note - the dependency checker needs to be run on both a debug and an optimized build.
--- a/icu4c/source/i18n/ucol_sit.cpp
+++ b/icu4c/source/i18n/ucol_sit.cpp
@@ -24,6 +24,10 @@
 #include "cstring.h"
 #include "uresimp.h"
 #include "unicode/coll.h"
+#include "unicode/stringpiece.h"
+#include "charstr.h"
+
+U_NAMESPACE_USE
 
 #ifdef UCOL_TRACE_SIT
 # include <stdio.h>
@@ -82,7 +86,6 @@ static const char providerKeyword[]  = "@sp=";
 static const int32_t locElementCount = UCOL_SIT_LOCELEMENT_MAX+1;
 static const int32_t locElementCapacity = 32;
 static const int32_t loc3066Capacity = 256;
-static const int32_t locProviderCapacity = 10;
 static const int32_t internalBufferSize = 512;
 
 /* structure containing specification of a collator. Initialized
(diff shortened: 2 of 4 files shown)


icu4c/source/i18n/ucol_sit.cpp
```diff
@@ -390,37 +390,37 @@ static void
 ucol_sit_calculateWholeLocale(CollatorSpec *s) {
     // put the locale together, unless we have a done
     // locale
-    if(s->locale[0] == 0) {
+    if(s->locale.size() == 0) {
         // first the language
-        uprv_strcat(s->locale, s->locElements[UCOL_SIT_LANGUAGE]);
+        s->locale.append(s->locElements[UCOL_SIT_LANGUAGE]);
```
jefgen: `std::string` will throw on memory allocation failure, so every single one of these `append` calls might possible throw an exception if the string class needs to reallocate and OOM happens.
AFAIK, exception usage is banned inside the ICU4C library code. Some of the tests do use the STL, and (for better or worse) don't bother checking for exceptions being thrown (which is likely okay since it's test code) -- but at least within the library code, exceptions aren't allowed.

srl295: OK… what do you recommend to use instead?

jefgen: IIRC, I think Markus suggested to use the `CharString` class for things like this.

Here's an example usage in another i18n file:
[https://github.com/unicode-org/icu/blob/730207965353dad8856cd71ddaefbf36afd523e7/icu4c/source/i18n/dtptngen.cpp#L621](https://github.com/unicode-org/icu/blob/730207965353dad8856cd71ddaefbf36afd523e7/icu4c/source/i18n/dtptngen.cpp#L621)

It has/takes/uses an `UErrorCode` which can be use for checking OOM, etc.

It looks like this method (`ucol_sit_calculateWholeLocale`) is an internal helper, and all the callers do have existing status params -- so you could probably add an error status to the params for this helper. :)


---
In reply to: [395936645](https://github.com/unicode-org/icu/pull/1052#discussion_r395936645) [](ancestors = 395936645)

srl295: thx i'll do that

.travis.yml
```diff
@@ -39,6 +39,24 @@ matrix:
       after_failure:
           -  cat `find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;`
 
+    - name: "c: linux gcc 8 Wextra"
```
markusicu: PR says gcc 9, this says gcc 8. please fix one or the other.

markusicu: oh, i see you had a commit where you downgraded from 9 to 8. i will edit the pr.

markusicu:
please take a look at the travis build bot failure

markusicu:
AppVeyor build failed -- failed to install cygwin?

jefgen:
> AppVeyor build failed -- failed to install cygwin?

AppVeyor is horribly flaky... :(
I've requested a rebuild of the PR.

srl295:
Sorry for the confusion. Summary:
- GCC 8 introduced the new warnings.
- ICU on Mac may have been broken on GCC since GCC7, hence that mention.
- GCC 9 is what I originally tested with locally, BUT:
- GCC 8 is the latest I can 'easily' install on ubuntu bionic. (as `g++-8`.   `g++-9` is aliased to Clang via regex… go figure how that's a good idea. It originally seemed like what I wanted. )

srl295:
:tada: 
> c: linux gcc 8 Wextra 
Now passes (with new warnings!).  So I think this is ready to merge once the other bots are happy.

jefgen:
> > AppVeyor build failed -- failed to install cygwin?
> 
> AppVeyor is horribly flaky... :(
> I've requested a rebuild of the PR.

Never mind, it isn't AppVeyor -- rather the issue is that the Cygwin download site itself seems to be  completely down.

From the Cygwin webpage, there is this message:

> Cygwin Time Machine (is broken :-( )
Greetings! If you have come seeking the Cygwin Time Machine, I'm afraid you will have to wait a little while. The site suffered an unrecoverable disk failure on 3/20/2020, and I lost a major portion of the archive. I am in the process of restoring it, but it will take me some time to get things back to normal. Stay tuned to this page for further details. I apologize for the inconvenience.

So it seems like we might be without any Cygwin builds for some time...

PR #1056, opened 2020-03-24
ICU-20068 Shorten .gitignore by making common rules more generic

roubert:
Wildcard expressons, directory names and file names that identify
artefacts that should be ignored regardless of in which subdirectory
they are can be consolidated, thereby shortening the list considerably.

https://unicode-org.atlassian.net/browse/ICU-20068


roubert:
I've asked both of you to review for I think this would really benefit from an extra pair of eyes and (if you can spare the time) actually testing that it really does work as expected in your different development environments.

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
-icu4c/source/samples/all/*.opensdf
-icu4c/source/samples/all/*.sdf
-icu4c/source/samples/all/.vs
 icu4c/source/samples/all/all.ncb
 icu4c/source/samples/all/all.suo
-icu4c/source/samples/break/*.vcxproj.user
-icu4c/source/samples/break/Debug
 icu4c/source/samples/break/break
 icu4c/source/samples/break/break.exe
-icu4c/source/samples/break/break.out
```
sffc: I don't see which new rule matches this one

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
 icu4c/source/samples/cal/icucal
 icu4c/source/samples/cal/icucal.exe
-icu4c/source/samples/cal/release
-icu4c/source/samples/cal/x64
-icu4c/source/samples/cal/x86
-icu4c/source/samples/case/*.vcxproj.user
-icu4c/source/samples/case/Debug
 icu4c/source/samples/case/case
 icu4c/source/samples/case/case.exe
-icu4c/source/samples/case/case.out
```
sffc: Did you mean to include `*.out` as one of the wildcard ignores?  (I'll stop flagging other .out files)

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
-icu4c/source/samples/ugrep/ugrep.out
-icu4c/source/samples/ugrep/ugrep.vcproj.*.*.user
-icu4c/source/samples/ugrep/x64
-icu4c/source/samples/ugrep/x86
-icu4c/source/samples/uresb/*.res
-icu4c/source/samples/uresb/*.vcxproj.user
-icu4c/source/samples/uresb/Debug
-icu4c/source/samples/uresb/release
-icu4c/source/samples/uresb/resources.vcproj.*.*.user
-icu4c/source/samples/uresb/resources_Win32_Debug
```
sffc: How about this one?

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
+*.sdf
+*.so
+*.suo
+*.tgz
+*.user
+*.vcproj.*.*.user
+*.vcxproj.user
+cygicudata*
+derb_*
+ht_*
```
sffc: If it were up to me, I would make `derb_*` and `ht_*` directory-specific excludes

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
+tmp/
+x64/
+x86/
+
+# Files
+Makefile
+Makefile.local
+build-local.properties
+charperf
+collperf
```
sffc: `charperf` and `collperf` are specific binary names, so probably best to keep in the "specific file names" section, even though there are two of them each.

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
+*.user
+*.vcproj.*.*.user
+*.vcxproj.user
+cygicudata*
+derb_*
+ht_*
+libicu*
+libicutest*
+libsicu*
+libsicutest*
```
sffc: Some of these also look like they might be better as directory-specific excludes, like `libsicutest*`.

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
+icu4c/source/samples/ufortune/resources/*.dat
+icu4c/source/samples/ufortune/resources/*.res
+icu4c/source/samples/uresb/*.res
+icu4c/source/test-*.xml
+icu4c/source/test/perf/howExpensiveIs/*.xml
+icu4c/source/tools/gendict/*.1
+icu4c/source/tools/genrb/*.1
+icu4c/source/tools/genren/urename.*
+icu4c/source/tools/makeconv/*.1
+icu4c/source/tools/pkgdata/*.1
```
sffc: Why not make `*.1` a global wildcard?

sffc:
Also, in the PR description, please use the template. :)

.gitignore
roubert: Done.

sffc:
Could the AppVeyor error be real?

```
ln -fs libicutu67.dll.a ../../lib/libicutu.dll.a
ln: failed to create symbolic link '../../lib/libicutu.dll.a': File exists
make[2]: Leaving directory '/cygdrive/c/projects/icu/icu4c/source/tools/toolutil'
make[2]: *** [../../config/mh-cygwin64:108: ../../lib/cygicutu.dll] Error 1
make[2]: Leaving directory '/cygdrive/c/projects/icu/icu4c/source/tools/toolutil'
make[1]: Making `all' in `ctestfw'
```

jefgen:
> Could the AppVeyor error be real?

There was a long standing race condition in the Cygwin build configuration with how it removed and updated the symbolic links. The issue was more likely to occur when using a parallel build though.

I recently fixed this in PR #1066, so this is now fixed on master (and in 67 RC). 

It looks like the [gitignore ](https://github.com/roubert/icu/blob/gitignore/icu4c/source/config/mh-cygwin64) branch in this PR doesn't yet have this commit though. So likely just need to update it/merge in the master branch, and it should be fine.

.gitignore
```diff
@@ -1,1071 +1,269 @@
...
+icu4j/*.jar
+icu4j/lib/*.jar
+tools/multi/packages/*.jar
+tools/release/java/APIChangeReport*
 
-# Specific file name ignores
-/.git
-.vscode
-icu4c/bin
-icu4c/binARM
```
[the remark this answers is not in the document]
jefgen: thanks!

.gitignore
```diff
@@ -1,11 +1,104 @@
...
+tools/release/java/APIChangeReport*
 
-# Specific file name ignores
-/.git
-.vscode
-icu4c/bin
+icu4c/source/tools/icuinfo/icuplugins*.txt
+!icu4c/source/tools/icuinfo/icuplugins_windows_sample.txt
+
+# Specific file names
```
jefgen, line 101: super-nit: Some of these are really "files", rather they're folders/directories.

roubert: That's just my Unix background shining through: Every name in the file system is a file name, regardless of whether it refers to a regular file, a directory, a symbolic link, a device, etc., etc. ;-)

sffc:
Also, if you put this on master, it will target ICU 68.  I think that's probably what we want at this point.  If you think it belongs in ICU 67, you should send an email to the team.

PR #1058, opened 2020-03-25
ICU-21022 Use logKnownIssue to avoid TestDateFormatRoundTrip exhaustive fail

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21022
- [x] Issue accepted
- [x] Tests included

In exhaustive mode, TestDateFormatRoundTrip has a failure in which the successive parse/format cycles for ar_JO@calendar=islamic-civil in time-format-only mode fail to converge as rapidly as expected. For now just add a logKnownIssue that allows more cycles for convergence in this case.


pedberg-icu:
Need to verify that the ICU4C exhaustive tests pass. Note that:
- The ICU4J exhaustive tests may fail for a different reason (in IntlTestNumberFormat.TestMonster), that is a different issue
- Per Jeff e-mail, the appveyor failure is due to Cygwin download site being completely down

pedberg-icu:
OK, C exhaustive is passing with this fix.

PR #1062, opened 2020-03-25
ICU-20803 Pass ByteSink to _canonicalize()

roubert:
https://unicode-org.atlassian.net/browse/ICU-20803

This eliminates the need for the fixed size scratch buffer inside of
locale_set_default_internal() and also eliminates the need for counting
bytes, something that ByteSink will now handle correctly when needed.

None of this should have any externally visible effect (apart from
removing the arbitrary size limit imposed by the fixed size scratch
buffer), it's all about cleaning up implementation internals.

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -30,10 +30,14 @@
      l = lang, C = ctry, M = charmap, V = variant
 */
 
+#include "unicode/bytestream.h"
+#include "unicode/errorcode.h"
+#include "unicode/stringpiece.h"
 #include "unicode/utypes.h"
 #include "unicode/ustring.h"
 #include "unicode/uloc.h"
 
+#include "bytesinkutil.h"
 #include "putilimp.h"
 #include "ustr_imp.h"
 #include "ulocimp.h"
@@ -46,6 +50,7 @@
 #include "uassert.h"
 #include "charstr.h"
 
+#include <algorithm>
 #include <stdio.h> /* for sprintf */
 
 U_NAMESPACE_USE
@@ -59,7 +64,6 @@ U_CFUNC int32_t
 locale_getKeywords(const char *localeID,
             char prev,
             char *keywords, int32_t keywordCapacity,
-            char *values, int32_t valuesCapacity, int32_t *valLen,
             UBool valuesToo,
             UErrorCode *status);
 
@@ -597,11 +601,10 @@ compareKeywordStructs(const void * /*context*/, const void *left, const void *ri
     return uprv_strcmp(leftString, rightString);
 }
 
-static int32_t
+static void
 _getKeywords(const char *localeID,
              char prev,
-             char *keywords, int32_t keywordCapacity,
-             char *values, int32_t valuesCapacity, int32_t *valLen,
+             ByteSink& sink,
              UBool valuesToo,
              UErrorCode *status)
 {
@@ -613,8 +616,6 @@ _getKeywords(const char *localeID,
     const char* equalSign = NULL;
     const char* semicolon = NULL;
     int32_t i = 0, j, n;
-    int32_t keywordsLen = 0;
-    int32_t valuesLen = 0;
 
     if(prev == '@') { /* start of keyword definition */
         /* we will grab pairs, trim spaces, lowercase keywords, sort and return */
@@ -629,7 +630,7 @@ _getKeywords(const char *localeID,
             }
             if(numKeywords == maxKeywords) {
                 *status = U_INTERNAL_PROGRAM_ERROR;
-                return 0;
+                return;
             }
             equalSign = uprv_strchr(pos, '=');
             semicolon = uprv_strchr(pos, ';');
@@ -637,13 +638,13 @@ _getKeywords(const char *localeID,
             /* ';' before '=' [foo@currency;collation=pinyin] is illegal */
             if(!equalSign || (semicolon && semicolon<equalSign)) {
                 *status = U_INVALID_FORMAT_ERROR;
-                return 0;
+                return;
             }
             /* need to normalize both keyword and keyword name */
             if(equalSign - pos >= ULOC_KEYWORD_BUFFER_LEN) {
                 /* keyword name too long for internal buffer */
                 *status = U_INTERNAL_PROGRAM_ERROR;
-                return 0;
+                return;
             }
             for(i = 0, n = 0; i < equalSign - pos; ++i) {
                 if (pos[i] != ' ') {
@@ -654,7 +655,7 @@ _getKeywords(const char *localeID,
             /* zero-length keyword is an error. */
             if (n == 0) {
                 *status = U_INVALID_FORMAT_ERROR;
-                return 0;
+                return;
             }
 
             keywordList[numKeywords].keyword[n] = 0;
@@ -669,7 +670,7 @@ _getKeywords(const char *localeID,
             /* Premature end or zero-length value */
             if (!*equalSign || equalSign == semicolon) {
                 *status = U_INVALID_FORMAT_ERROR;
-                return 0;
+                return;
             }
 
             keywordList[numKeywords].valueStart = equalSign;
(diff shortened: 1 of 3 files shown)


icu4c/source/common/uloc.cpp
```diff
@@ -1199,48 +1164,48 @@ ulocimp_getLanguage(const char *localeID,
...
-            lang[i]=(char)uprv_tolower(*localeID);
-        }
-        i++;
+        char c = (char)uprv_tolower(*localeID);
+        sink.Append(&c, 1);
         localeID++;
     }
 
-    if(i==3) {
+    if(sink.size()==3) {
```
markusicu: Let's try to collect the language subtag in an internal CharString and work with that until we have the final version and append that to the sink. Then we don't need the new methods.

icu4c/source/common/uloc.cpp
```diff
@@ -1588,77 +1553,85 @@ _canonicalize(const char* localeID,
...
-    if (/*OPTION_SET(options, _ULOC_CANONICALIZE) &&*/
-        (result == NULL || resultCapacity < (int32_t)sizeof(localeBuffer))) {
-        name = localeBuffer;
-        nameCapacity = (int32_t)sizeof(localeBuffer);
-    } else {
-        name = result;
-        nameCapacity = resultCapacity;
+    LocalMemory<char> localeBuffer;
+    LocalPointer<CheckedArrayByteSink> localSink;
+    internal::MemoryByteSink* sinkp = &sink;
```
markusicu: AFAICT this is always a CheckedArrayByteSink, so please use that, and again we should not need the new methods. We may need to read from the localeBuffer instead of from sinkp->data().

If we do need different sink types, then I would start with a variable that tells us where the contents is -- I am reluctant to add API to the ByteSink classes.

icu4c/source/common/uloc.cpp
```diff
@@ -1588,77 +1553,85 @@ _canonicalize(const char* localeID,
...
+                &sentinel,
+                ULOC_FULLNAME_CAPACITY,
+                &dummy);
+
+        /* if we are doing a full canonicalization, then put results in
+           localeBuffer, if necessary; otherwise send them directly to sink. */
+        if (/*OPTION_SET(options, _ULOC_CANONICALIZE) &&*/
+            buffer == &sentinel) {
+            if (localeBuffer.allocateInsteadAndReset(
+                    ULOC_FULLNAME_CAPACITY) == nullptr) {
```
markusicu: For a small, fixed-sized array we can just use stack-allocation.

On the other hand, why not CharString + CharStringByteSink?

roubert:
The already existing code is noticeably more complex, than it otherwise would have needed to be, in order to avoid making any copies. Time and effort have clearly been spent on writing this more complicated logic. I simply assumed that this had been done because it was important and that changing this logic therefore would not be accepted.

Is it now your opinion that avoiding copies no longer is (or maybe even never was) important enough to pay the price of more complex code for?

If that's the case, the code can be simplified considerably by making copies whenever that simplifies buffer handling logic.

markusicu:
Performance is important, but I want to think carefully about evolving the ByteSink API from write-only to read-write.

If we can avoid the read-write model by building tiny strings in stack-allocated, stack-buffer CharString and copying a tiny number of bytes into a sink, then that sounds much better to me.

roubert:
OK, done.

icu4c/source/common/uloc.cpp
```diff
@@ -1199,48 +1165,52 @@ ulocimp_getLanguage(const char *localeID,
...
 U_CFUNC int32_t
+ulocimp_getLanguage(const char *localeID,
+                    char *language, int32_t languageCapacity,
+                    const char **pEnd) {
+    ErrorCode status;
+    CharString result = ulocimp_getLanguage(localeID, pEnd, status);
+    if (status.isFailure()) {
+        return 0;
+    }
+    CheckedArrayByteSink sink(language, languageCapacity);
```
markusicu: This sink does not really buy us anything. You actually have a real bug here -- language[] is not NUL-terminated.

I would
```
int32_t length = result.length();
if (length <= languageCapacity) {
    uprv_memcpy(language, result.data(), length);
    if (length < languageCapacity) { language[length] = 0; }
}
return length;
```

icu4c/source/common/uloc.cpp
```diff
@@ -1259,132 +1229,137 @@ ulocimp_getScript(const char *localeID,
...
 U_CFUNC int32_t
+ulocimp_getScript(const char *localeID,
+                  char *script, int32_t scriptCapacity,
+                  const char **pEnd) {
+    ErrorCode status;
+    CharString result = ulocimp_getScript(localeID, pEnd, status);
+    if (status.isFailure()) {
+        return 0;
+    }
+    CheckedArrayByteSink sink(script, scriptCapacity);
```
markusicu: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1259,132 +1229,137 @@ ulocimp_getScript(const char *localeID,
...
+U_CFUNC int32_t
+ulocimp_getCountry(const char *localeID,
+                   char *country, int32_t countryCapacity,
+                   const char **pEnd) {
+    ErrorCode status;
+    CharString result = ulocimp_getCountry(localeID, pEnd, status);
+    if (status.isFailure()) {
+        return 0;
+    }
+    CheckedArrayByteSink sink(country, countryCapacity);
```
markusicu: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1259,132 +1229,137 @@ ulocimp_getScript(const char *localeID,
...
     }
+}
 
-    return i;
+static int32_t
+_getVariantEx(const char *localeID,
+              char prev,
+              char *variant, int32_t variantCapacity,
+              UBool needSeparator) {
+    CheckedArrayByteSink sink(variant, variantCapacity);
```
markusicu: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1588,77 +1559,55 @@ _canonicalize(const char* localeID,
...
-        const char *d = uloc_getDefault();
-
-        len = (int32_t)uprv_strlen(d);
+    CharString tag = ulocimp_getLanguage(tmpLocaleID, &tmpLocaleID, *err);
 
-        if (name != NULL) {
-            uprv_memcpy(name, d, len);
-        }
+    if (tag.length() == I_DEFAULT_LENGTH &&
+        uprv_strncmp(origLocaleID, i_default, I_DEFAULT_LENGTH) == 0) {
```
markusicu: indent+4

icu4c/source/common/uloc.cpp
```diff
@@ -1959,7 +1918,34 @@ uloc_getBaseName(const char* localeID,
...
+    int32_t reslen = sink.NumberOfBytesAppended();
+
+    if (U_FAILURE(*err)) {
+        return reslen;
+    }
+
+    if (sink.Overflowed()) {
+        *err = U_BUFFER_OVERFLOW_ERROR;
+    } else {
+        u_terminateChars(name, nameCapacity, reslen, err);
```
markusicu: Optional: Might be useful to add a bytesinkutil function to terminateByteSink. Maybe specific to a CheckedArrayByteSink, subsuming all of the stuff like you have here lines 1928..1940 (but `sink.Append('\0')` for the actual termination). Then you could just `return ByteSinkUtil::terminateCheckedArrayByteSink(sink, *err);`

markusicu:
PS: After this PR, let's not do more enhancements for 67. We are supposed to be in code freeze except BRS tasks and critical bug fixes. If there is more to be done for the ticket, then it should go into another follow-up ticket.

icu4c/source/common/uloc.cpp
roubert: > This sink does not really buy us anything.

Done.

icu4c/source/common/uloc.cpp
roubert: Done.

icu4c/source/common/uloc.cpp
roubert: > You actually have a real bug here -- language[] is not NUL-terminated.

As far as I understand the code this is not a bug but by design, the existing code doesn't `NUL` terminate these buffers and there is no caller of these functions (they're all internal helper functions) that expects that do be done.

The new functions added by this PR, that return `CharString` instead, get automatic `NUL` termination through `CharString` and eventually all code that still calls the old functions will be updated to use the new functions instead, so the issue will eventually simply disappear and the time being, I think that keeping current behaviour intact is the best course of action.

icu4c/source/common/uloc.cpp
roubert: Something along those lines could indeed be useful. I do however think that it for the time being is a good idea to simply continue copy-pasting this recipe until all the locale memory management has been updated, for at that point in time we'll have a much clearer view of exactly what would be the most useful design for such a helper function.

roubert:
> After this PR, let's not do more enhancements for 67.

That comes naturally, for it wouldn't make much sense to make any other changes to this code before this PR has been merged and you moved this PR to 68 already two days ago.

markusicu:
I will leave it up to you whether you want to check this in for 67 and close the ticket, or leave the PR pending and potentially do more under the same ticket for 68.

roubert:
> I will leave it up to you whether you want to check this in for 67 and close the ticket, or leave the PR pending and potentially do more under the same ticket for 68.

Hm, all checks passed and the merge was conflict free, changing the ticket again and merging now is likely to be the least amount of overall work.

markusicu:
Ok, thanks!
Do you want to put remaining ideas into a follow-up ticket?

roubert:
> Do you want to put remaining ideas into a follow-up ticket?

I've already filed ICU-21035.

icu4c/source/common/uloc.cpp
roubert: > I do however think that it for the time being is a good idea to simply continue copy-pasting this recipe until all the locale memory management has been updated, for at that point in time we'll have a much clearer view of exactly what would be the most useful design for such a helper function.

PR #2863

PR #147, opened 2018-09-19
ICU-13707 Make UnicodeString safe when append or insert into itself.

sffc:
##### Checklist

- [x] Issue filed at https://unicode-org.atlassian.net :  ICU-13707
- [x] Update PR title to include Issue number
- [x] Issue accepted
- [x] Tests included
- [x] Documentation is changed or added

PR #1063, opened 2020-03-25
ICU-21032 Backport of ICU-20958 to 61.x

srl295:
[ICU-21032] Backport of [ICU-20958] aka #971 

Includes partial backports so that this can be merged:
- ICU-13634 Adding functions for integer overflow
- ICU-13634 Removing stray #include added in r41136.
- ICU-13707 Make UnicodeString safe when appended or inserted into itself. (#147)

Multiple commits to preserve history, so:

ALLOW_MANY_COMMITS=true

[ICU-21032]: https://unicode-org.atlassian.net/browse/ICU-21032
[ICU-20958]: https://unicode-org.atlassian.net/browse/ICU-20958

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,79 @@
+# AppVeyor configuration for CI Builds of ICU4C.
+
+image: Visual Studio 2017
+platform: x64
+
+# Don't clone the entire repo.
+clone_depth: 1
+
+# Cache things between builds to speed things up and save bandwidth.
+cache:
+  - c:\icu-ci-cache
+
+# Notes:
+#  - The VS2017 build is disabled (commented out) below, as we build both VS2015 and VS2017
+#    on Azure Pipelines instead, since that is much faster.
+#
+#  - For the Cygwin build, the CYG_MIRROR URL below is from the "Cygwin Time Machine" archive,
+#    and that URL maps to the 64-bit version of Cygwin 2.11.2-1 (with timestamp 1550062412).
+#    This was the last release of Cygwin 2.11 before version 3.0.0 was released.
+#    See: http://www.crouchingtigerhiddenfruitbat.org/Cygwin/timemachine.html
+#
+#    Note: The archived Cygwin repo does not keep or retain any file signature files, so we
+#    must use the "-X" or "--no-verify" option below in the setup command.
+
+environment:
+  global:
+    ICU_CI_CACHE: c:\icu-ci-cache
+    CYG_URL: https://cygwin.com/setup-x86_64.exe
+    CYG_MIRROR: http://ctm.crouchingtigerhiddenfruitbat.org/pub/cygwin/circa/64bit/2019/02/13/045332
+    CYG_PACKAGES: automake,gcc-core,gcc-g++,make,pkg-config,perl,python3
+    CYG_ROOT: c:\cygwin-root
+    CYG_CACHE: '%ICU_CI_CACHE%\cygwin64'
+    CYG_CACHED_SETUP: '%CYG_CACHE%\setup.exe'
+
+  matrix:
+    - BUILDSYSTEM: cygwin64-gcc
+      CC: gcc
+      CXX: g++
+      APPVEYOR_SAVE_CACHE_ON_ERROR: true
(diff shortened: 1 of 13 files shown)


icu4c/source/common/putil.cpp
```diff
@@ -533,6 +533,30 @@ uprv_fmin(double x, double y)
     return (x > y ? y : x);
 }
 
+#include <iostream>
```
jefgen: Why is `iostream` included here? I don't see it in the file `putil.cpp` on the master branch.

jefgen: Found where it was removed:
https://github.com/unicode-org/icu/commit/afda6f137878057910bf4a86bc45a03a73a1e35a#diff-08900170df63db033be7aa17ad18a0e1
FWIW, you probably want to cherry-pick that as well, as I don't think we want the bloated `iostream` in the library code.

srl295: THanks. Added.

jefgen: thx!

srl295:
Thanks. I'll let it bake and merge tomorrow.

srl295:
> Azure Pipelines failed to run 1 pipeline(s).

@jefgen ^ this is an old branch… perhaps it doesn't have the metadata/config files needed for this?

srl295:
Actually… it doesn't even have Travis running… !

jefgen:
> Actually… it doesn't even have Travis running… !

Yeah, looking at the tree here https://github.com/unicode-org/icu/tree/maint/maint-61 I don't see any config files for Travis, AppVeyor, or Azure, etc...

srl295:
> > Actually… it doesn't even have Travis running… !
> 
> Yeah, looking at the tree here https://github.com/unicode-org/icu/tree/maint/maint-61 I don't see any config files for Travis, AppVeyor, or Azure, etc...

But "all" checks pass am i right? 😁 

okay… I'll try just porting over the head .travis, etc.  May need to tone down some of the warnings though.

FrankYFTang:
61.x? do we have a policy of how many versions back we should maintain for the branch?

srl295:
> 61.x? do we have a policy of how many versions back we should maintain for the branch?

It's been up to whoever wants to do the work. me, in this case

.appveyor.yml
```diff
@@ -0,0 +1,79 @@
...
+#    See: http://www.crouchingtigerhiddenfruitbat.org/Cygwin/timemachine.html
+#
+#    Note: The archived Cygwin repo does not keep or retain any file signature files, so we
+#    must use the "-X" or "--no-verify" option below in the setup command.
+
+environment:
+  global:
+    ICU_CI_CACHE: c:\icu-ci-cache
+    CYG_URL: https://cygwin.com/setup-x86_64.exe
+    CYG_MIRROR: http://ctm.crouchingtigerhiddenfruitbat.org/pub/cygwin/circa/64bit/2019/02/13/045332
```
jefgen: Note: This will fail as the Cygwin time machine is down.

srl295: and maven fails also … don't think it should be a blocker

srl295:
^ oops, should have rebased before the exhaustive. @jefgen will we still get the results? Still no code change on the prior commits

jefgen:
> will we still get the results?

No, I think the force-push cancels any running builds on PR. (It's not setup to batch them like it does on the master branch.)

srl295:
C exhaustive tests passed. I think this is ready to merge.

jefgen:
- J: Linux OpenJDK (Ubuntu 16.04)
Never gets past downloading Ant dependencies due to Maven wanting HTTPS instead of HTTP (fixed in never versions.)
- C: MSVC x64 ARM32 ARM64 Release (VS 2017) + Distrelease ARM64
Builds fine, the failure is in the distrelease script
- C: MSVC 32-bit Release (VS 2017) + Distrelease
Builds fine, the failure is in the distrelease script
- C: MSYS2 GCC x86_64 Release
Build fails, but I this is expected. MSYS2 had a bunch of issues that were fixed in more recent versions.

overall, builds LGTM.

PR #1064, opened 2020-03-25
ICU-20979 BRS67RC Update version numbers, regenerate configure, update ICU4C README

daniel-ju:
##### Checklist

- [x] Issue accepted
- [ ] Tests included


icu4c/readme.html
```diff
@@ -33,7 +33,7 @@ <h1>
       <span class="only-rc">Release Candidate</span>
```
daniel-ju: Still left the RC stuff in here in case we ever want to have different READMEs in the future... otherwise we could remove it.

jefgen: Thanks!
This is probably fine for now. IIRC, I think there was some (brief) mention that maybe the README might be converted to Markdown and checked-in. (Though, I think it would mostly just point to the user-guide and download page).

---
In reply to: [398193398](https://github.com/unicode-org/icu/pull/1064#discussion_r398193398) [](ancestors = 398193398)

jefgen:
Argh, the failing AppVeyor builds are due to Cygwin 3.x changes. 

I have a PR here to fix it: #1066

jefgen:
It looks like there is a conflict with one of the jar files: `icu4j/main/shared/data/icudata.jar`
(This is from https://github.com/unicode-org/icu/commit/0d26f83f83bdd5786a0719bc3ab161eccf8770b5).

We should likely re-generate the jar file for this before checking in. (You might want to just rebase the changes on top of master first, which would also fix the AppVeyor build failure too. :) )

PR #1067, opened 2020-03-26
ICU-13259 Check whether ulocdata_getDelimiter works in various locales including fr_CA

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-13259
- [x] Issue accepted
- [x] Tests included

Original plan was to add a test to demonstrate that the reported problem no longer occurs. But in fact it does still occur, so this will need to be expanded to actually fix the problem.

### DIFF
--- a/icu4c/source/test/cintltst/cldrtest.c
+++ b/icu4c/source/test/cintltst/cldrtest.c
@@ -1333,7 +1333,7 @@ static void TestCoverage(void){
         status = U_ZERO_ERROR;
         ulocdata_getDelimiter(uld, types[i], result, 32, &status);
         if (U_FAILURE(status)){
-            log_err("ulocdata_getgetDelimiter error with type %d", types[i]);
+            log_err("ulocdata_getDelimiter error with type %d", types[i]);
         }
     }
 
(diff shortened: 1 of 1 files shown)


pedberg-icu:
Oops, need to actually call the test 🙄

jefgen:
I think this might have been fixed by https://github.com/unicode-org/icu/pull/1399

I re-based the changes on top of the most recent master (with Frank's fix), and the tests are now all passing.

Here is the branch in my fork: https://github.com/jefgen/icu/commits/ICU-13259-addTestForDelimiters
Test results: https://dev.azure.com/ms/icu/_build/results?buildId=114703&view=results

So I think if this PR could rebased on top of the most recent master, then it could be merged and the bug resolved. :)

pedberg-icu:
> I think this might have been fixed by #1399
> 
> I re-based the changes on top of the most recent master (with Frank's fix), and the tests are now all passing.
> 
> Here is the branch in my fork: https://github.com/jefgen/icu/commits/ICU-13259-addTestForDelimiters
> Test results: https://dev.azure.com/ms/icu/_build/results?buildId=114703&view=results
> 
> So I think if this PR could rebased on top of the most recent master, then it could be merged and the bug resolved. :)

Yes indeed, thanks, I rebased and the test passes, looks like #1399 fixed this. Removing the incomplete label.

jefgen (approved):
lgtm

FrankYFTang:
Could you also add similar test to ICU4J and make sure Java do the right thing. I forget to check on Java.

FrankYFTang (approved):
LGTM

FrankYFTang:
> Could you also add similar test to ICU4J and make sure Java do the right thing. I forget to check on Java.

Don't need to do in this PR. A separate PR is ok.

PR #1072, opened 2020-03-30
ICU-20979 BRS 67rc, update urename.h

pedberg-icu:
##### Checklist

- [x] Issue accepted
- [x] Tests included


icu4c/source/common/unicode/urename.h
```diff
@@ -203,6 +204,7 @@
 #define res_read U_ICU_ENTRY_POINT_RENAME(res_read)
 #define res_unload U_ICU_ENTRY_POINT_RENAME(res_unload)
 #define u_UCharsToChars U_ICU_ENTRY_POINT_RENAME(u_UCharsToChars)
+#define u_asciiToUpper U_ICU_ENTRY_POINT_RENAME(u_asciiToUpper)
```
jefgen: This one seems a bit odd... The function isn't public and seems to be an internal only API, yet the name starts with `u_` (instead of `uprv_`) and the API is also marked with the macro `U_CAPI` instead of `U_INTERNAL`.

icu4c/source/common/unicode/urename.h
```diff
@@ -1053,6 +1057,7 @@
 #define ulistfmt_format U_ICU_ENTRY_POINT_RENAME(ulistfmt_format)
 #define ulistfmt_formatStringsToResult U_ICU_ENTRY_POINT_RENAME(ulistfmt_formatStringsToResult)
 #define ulistfmt_open U_ICU_ENTRY_POINT_RENAME(ulistfmt_open)
+#define ulistfmt_openForType U_ICU_ENTRY_POINT_RENAME(ulistfmt_openForType)
```
jefgen: This one is marked as `U_CAPI`, but should probably be `U_DRAFT` as well.

icu4c/source/common/unicode/urename.h
```diff
@@ -1295,7 +1303,6 @@
 #define uprv_convertToPosix U_ICU_ENTRY_POINT_RENAME(uprv_convertToPosix)
 #define uprv_copyAscii U_ICU_ENTRY_POINT_RENAME(uprv_copyAscii)
 #define uprv_copyEbcdic U_ICU_ENTRY_POINT_RENAME(uprv_copyEbcdic)
-#define uprv_currencyLeads U_ICU_ENTRY_POINT_RENAME(uprv_currencyLeads)
```
jefgen: It looks like this function still exists. (But it doesn't look like it is used though?)

icu4c/source/common/unicode/urename.h
srl295: if not public it should be U_INTERNAL — or even statiic?

icu4c/source/common/unicode/urename.h
srl295: maybe dead stripped

icu4c/source/common/unicode/urename.h
pedberg-icu: Well, U_INTERNAL is defined to be U_CAPI so there is no implementation difference. For documentation purposes this could be updated, but that would be better as part of either the API status update or the guard conditionals update (both of which I am doing).

icu4c/source/common/unicode/urename.h
pedberg-icu: Again, U_DRAFT = U_CAPI so no impl difference, but I can fix for documentation as part of API status or guard conditional update

icu4c/source/common/unicode/urename.h
pedberg-icu: I notice a lot of variation in how the genren tool handles such internal functions, possible due to compiler variation or something. But yes, this does not seem to be used anywhere, it is internal with no U_EXPORT, etc. Can be dropped from urename.h

icu4c/source/common/unicode/urename.h
srl295: this function is called only from `i18n/currunit.cpp` and implemented in `common/ustring.cpp`, so U_INTERNAL is the right choice.

icu4c/source/common/unicode/urename.h
jefgen: Thank you Peter!

icu4c/source/common/unicode/urename.h
pedberg-icu: Fixed this in https://github.com/unicode-org/icu/pull/1081

icu4c/source/common/unicode/urename.h
pedberg-icu: Fixed this in https://github.com/unicode-org/icu/pull/1081

PR #1076, opened 2020-03-31
ICU-20979 Updating double-conversion from upstream master

sffc:
##### Checklist

- [x] Issue accepted
- [ ] Tests included

It was a clean integration this time.  Hopefully the build bots are happy.

### DIFF
--- a/icu4c/source/i18n/double-conversion-utils.h
+++ b/icu4c/source/i18n/double-conversion-utils.h
@@ -74,6 +74,12 @@ inline void abort_noreturn() { abort(); }
 #endif
 #endif
 
+#if defined(__clang__) && __has_attribute(uninitialized)
+#define DOUBLE_CONVERSION_STACK_UNINITIALIZED __attribute__((uninitialized))
+#else
+#define DOUBLE_CONVERSION_STACK_UNINITIALIZED
+#endif
+
 // Double operations detection based on target architecture.
 // Linux uses a 80bit wide floating point stack on x86. This induces double
 // rounding, which in turn leads to wrong results.
@@ -104,6 +110,7 @@ int main(int argc, char** argv) {
     defined(__ARMEL__) || defined(__avr32__) || defined(_M_ARM) || defined(_M_ARM64) || \
     defined(__hppa__) || defined(__ia64__) || \
     defined(__mips__) || \
+    defined(__nios2__) || \
     defined(__powerpc__) || defined(__ppc__) || defined(__ppc64__) || \
     defined(_POWER) || defined(_ARCH_PPC) || defined(_ARCH_PPC64) || \
     defined(__sparc__) || defined(__sparc) || defined(__s390__) || \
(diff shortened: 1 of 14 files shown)


icu4c/source/i18n/double-conversion-utils.h
```diff
@@ -74,6 +74,12 @@ inline void abort_noreturn() { abort(); }
 #endif
 #endif
 
+#if defined(__clang__) && __has_attribute(uninitialized)
+#define DOUBLE_CONVERSION_STACK_UNINITIALIZED __attribute__((uninitialized))
```
jefgen: It looks like this line is causing ~20 new warnings to be issued on the various MSVC build bots. 

I think the issue is that `__has_attribute` isn't understood by MSVC.
It looks like elsewhere in the ICU code, there is a check for ifdef first:

`#ifdef __has_attribute`

ex:
[https://github.com/unicode-org/icu/blob/20c29becd69b19c62cb6618665fd773eeb73dab1/icu4c/source/common/unicode/platform.h#L427](https://github.com/unicode-org/icu/blob/20c29becd69b19c62cb6618665fd773eeb73dab1/icu4c/source/common/unicode/platform.h#L427)

I wonder if maybe this could be changed to something like:

`#if defined(__clang__) && defined(__has_attribute) && __has_attribute(uninitialized)`

jefgen: Actually, upon second thought, it might be better to use the "custom" ICU version of the macro (`UPRV_HAS_ATTRIBUTE`), like this:
`#if defined(__clang__) && UPRV_HAS_ATTRIBUTE(uninitialized)`

I pushed a commit to branch on my fork (based off of your branch) here:
https://github.com/jefgen/icu/commit/3418c36d437039d574e16ba56c4cf53dd04daf0d

It looks like using this approach fixes the [warnings ](https://unicode-icu.visualstudio.com/ICU/_build/results?buildId=7647&view=logs&j=da564c1e-baf9-5d7d-25e9-0bb69f5e7fb5&t=ca1cae39-694b-53ec-5a87-5a46e2070573)(so down from 54 to 33 warnings), and works fine on [Travis ](
https://travis-ci.com/github/jefgen/icu/builds/157337062) as well.

sffc: I'm not particularly excited about adding more ICU patches; an upstream solution would be better.  Do you consider the new MSVC warnings a fatal issue that we need to fix now, or would it be okay from your point of view if we file an upstream issue to get resolved at a later date?  If we prepare an upstream PR that gets accepted, we could turn the crank one more time before ICU 67 GA, for example.

jefgen: No, I don't consider the new warning(s) to be fatal, nor are they bad enough to consider blocking this change from going in... (Thank you for updating the library BTW).
I agree that it would be better to pursue fixing this upstream in the double-conversion library, as that would benefit everyone. 
I'll most likely need to get internal approval first before I can contribute anything to the upstream library though, so I'm not sure if we'll have a fix before GA.

sffc: Upstream issue: https://github.com/google/double-conversion/issues/129

jefgen: Hah! Thank you kindly for filing the upstream issue! :)

PR #1077, opened 2020-03-31
ICU-21015 Fixing gcc compiler warnings

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21015
- [x] Issue accepted
- [x] Tests included

We accepted this ticket for ICU 67, and I didn't get the fix ready by the official code freeze, and then Markus said no, punt to 68.  But I would really rather see this in ICU 67.  There is one actual potential bug in library code, and there are also some changes in ICU public header files, which means that anyone using GCC and building against ICU may also encounter these errors, which is a standard we should avoid.

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -2197,6 +2197,11 @@ DateTimeMatcher::DateTimeMatcher(const DateTimeMatcher& other) {
     copyFrom(other.skeleton);
 }
 
+DateTimeMatcher& DateTimeMatcher::operator=(const DateTimeMatcher& other) {
+    copyFrom(other.skeleton);
+    return *this;
+}
+
 
 void
 DateTimeMatcher::set(const UnicodeString& pattern, FormatParser* fp) {
--- a/icu4c/source/i18n/unicode/basictz.h
+++ b/icu4c/source/i18n/unicode/basictz.h
@@ -207,6 +207,12 @@ protected:
     BasicTimeZone(const BasicTimeZone& source);
 
     /**
+     * Copy assignment.
+     * @stable ICU 3.8
+     */
+    BasicTimeZone& operator=(const BasicTimeZone&) = default;
+
+    /**
      * Gets the set of TimeZoneRule instances applicable to the specified time and after.
      * @param start     The start date used for extracting time zone rules
      * @param initial   Receives the InitialTimeZone, always not NULL
--- a/icu4c/source/i18n/unicode/numsys.h
+++ b/icu4c/source/i18n/unicode/numsys.h
@@ -74,6 +74,12 @@ public:
     NumberingSystem(const NumberingSystem& other);
 
     /**
+     * Copy assignment.
+     * @stable ICU 4.2
+     */
+    NumberingSystem& operator=(const NumberingSystem& other) = default;
+
+    /**
      * Destructor.
      * @stable ICU 4.2
      */
--- a/icu4c/source/test/cintltst/ccapitst.c
+++ b/icu4c/source/test/cintltst/ccapitst.c
@@ -3035,7 +3035,7 @@ static void TestJ1968(void) {
 
     err = U_ZERO_ERROR;
     myConvName[UCNV_MAX_CONVERTER_NAME_LENGTH-1] = ',';
-    strncpy(myConvName + UCNV_MAX_CONVERTER_NAME_LENGTH, "locale=", 7);
+    memcpy(myConvName + UCNV_MAX_CONVERTER_NAME_LENGTH, "locale=", 7);
     cnv = ucnv_open(myConvName, &err);
     if (cnv || err != U_ILLEGAL_ARGUMENT_ERROR) {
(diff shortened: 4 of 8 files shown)


icu4c/source/i18n/unicode/numsys.h
```diff
@@ -73,6 +73,12 @@ class U_I18N_API NumberingSystem : public UObject {
      */
     NumberingSystem(const NumberingSystem& other);
 
+    /**
+     * Copy assignment (defaulted).
+     * @draft ICU 67
```
sffc: What version number?

icu4c/source/i18n/unicode/basictz.h
```diff
@@ -43,6 +43,9 @@ class U_I18N_API BasicTimeZone: public TimeZone {
      */
     virtual ~BasicTimeZone();
 
+    /** Default copy assignment (inherited) */
```
sffc: What version number?

sffc:
FYI, for the public header files, the warning is (was):

```
In file included from ../../i18n/unicode/dcfmtsym.h:39,
                 from ../../i18n/unicode/numberformatter.h:16,
                 from numbertest_api.cpp:13:
../../i18n/unicode/numsys.h:74:5: note: because ‘icu::NumberingSystem’ has user-provided ‘icu::NumberingSystem::NumberingSystem(const icu::NumberingSystem&)’
   74 |     NumberingSystem(const NumberingSystem& other);
      |     ^~~~~~~~~~~~~~~
numbertest_api.cpp:63:96: warning: implicitly-declared ‘icu::NumberingSystem& icu::NumberingSystem::operator=(const icu::NumberingSystem&)’ is deprecated [-Wdeprecated-copy]
```

It turns out according to https://en.cppreference.com/w/cpp/language/copy_assignment#Implicitly-defined_copy_assignment_operator that this is deprecated according to the C++11 standard, so this warning is not just GCC being noisy; it's actually mandated by the spec.

icu4c/source/i18n/unicode/basictz.h
srl295: these were default pre-existing. So i think it should be exactly the version and stability of the copy c'tor, even if it was stable.

icu4c/source/i18n/unicode/numsys.h
srl295: same, these were default pre-existing. So i think it should be exactly the version and stability of the copy c'tor.

icu4c/source/test/cintltst/ccapitst.c
```diff
@@ -3035,7 +3035,7 @@ static void TestJ1968(void) {
 
     err = U_ZERO_ERROR;
     myConvName[UCNV_MAX_CONVERTER_NAME_LENGTH-1] = ',';
-    strncpy(myConvName + UCNV_MAX_CONVERTER_NAME_LENGTH, "locale=", 7);
+    memcpy(myConvName + UCNV_MAX_CONVERTER_NAME_LENGTH, "locale=", 7);
```
srl295, line 3038: this isn't library code, but do we want to use a #define to document the use of memcpy-for-string here?

icu4c/source/i18n/unicode/numsys.h
srl295: so stable 4.2

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -2197,6 +2197,11 @@ DateTimeMatcher::DateTimeMatcher(const DateTimeMatcher& other) {
     copyFrom(other.skeleton);
 }
 
+DateTimeMatcher& DateTimeMatcher::operator=(const DateTimeMatcher& other) {
```
markusicu, line 2200: very optional: this kind of thing could be inline in the header file

icu4c/source/i18n/unicode/basictz.h
markusicu: right, looks like stable ICU 3.8 here

i suspect that this one (and maybe some others) might have been exposed since my icu 65 changes for polymorphic clone

icu4c/source/test/cintltst/ccapitst.c
markusicu: nah, we use memcpy for strings in lots of places already

however, i think you need to copy 8 bytes in order to get myConvName NUL-terminated

srl295: yes, we do use memcpy.  The idea was to use a macro just to document the usage. But it's minor, especially in test code.

markusicu:
Shane, if you want this in 67, then please finish it today by 15:00 PT.

sffc:
^ that's a rebase to master.  Actual changes coming soon

icu4c/source/test/cintltst/ccapitst.c
sffc: The gcc warning for this one was that the string is 8 bytes long (including NUL), but strncpy is limited to copy only 7 bytes.  Changing strncpy to memcpy is equivalent and silences the warning.

sffc:
^ there's the actual changes.  PTAL

PR #1086, opened 2020-04-01
ICU-20979 BRS67RC Updating ICU4J readme

yumaoka:
For now, just updating version to 67, but not mentioning it's RC, so we don't need another update.

I also updated all dead ICU user guide links.

##### Checklist

- [x] Issue accepted
- [ ] Tests included


icu4j/readme.html
```diff
@@ -879,7 +879,7 @@ <h5> Generating Data from CLDR </h5>
 <I> Note: This procedure assumes that all 3 sources are present</I>
 <ol>
     <li>Checkout or download CLDR version 'release-36'</li>
```
daniel-ju, line 881: CLDR version 37 here.

yumaoka: Thanks - Updated this, and dropped another reference to CLDR 36

PR #1094, opened 2020-04-05
ICU-3655 No getAvailableLocales in RuleBasedNumberFormat

robertgmelo:
ICU-3655 No getAvailableLocales in RuleBasedNumberFormat

- Add static methods to return the proper list of locales
that implement RBNF rules.

### DIFF
--- a/icu4c/source/i18n/rbnf.cpp
+++ b/icu4c/source/i18n/rbnf.cpp
@@ -1116,6 +1116,30 @@ RuleBasedNumberFormat::findRuleSet(const UnicodeString& name, UErrorCode& status
     return NULL;
 }
 
+const Locale* U_EXPORT2
+RuleBasedNumberFormat::getAvailableLocales(int32_t& count)
+{
+    UErrorCode status = U_ZERO_ERROR;
+    UEnumeration* locs = ures_openAvailableLocales(U_ICUDATA_RBNF, &status);
+    count = 0;
+    if (U_SUCCESS(status)) {
+        status = U_ZERO_ERROR;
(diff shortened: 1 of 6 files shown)


markusicu:
cc @grhoten

grhoten:
What's the issue number for making this change in ICU4C, or will it be changed in this pull request too? If you're adding API, you should also have a test to call this new API.

robertgmelo:
@grhoten, 

What's the issue number for making this change in ICU4C, or will it be changed in this pull request too?

> I asked in the CR if it would be also necessary on ICU4C. I assume now it's necessary.

I'll proceed with that on ICU4C as well and also consider the tests of new API calls.

Thanks in advance.

grhoten:
Did you propose your API changes to the icu-design mailing list? Also I thought we were preferring static StringEnumeration* U_EXPORT2 getAvailableLocales(void) now because it's a string enumeration. NumberFormat has both. I don't have a strong opinion either way as long as it's available some way.

grhoten:
I looked here for your proposal: https://sourceforge.net/p/icu/mailman/icu-design/

robertgmelo:
@grhoten 

Honestly, I didn't send :( I added in the API doc only.

Yeah, when I started to look on ICU4C, I saw the 2 signatures on NumberFormat, the one of Locale and this one you've mentioned, of StringEnumeration. 

In this case, I also should this missing one, correct? Of this way, we will be compliant with mother class.

Please let me know your thoughts.

robertgmelo:
Abandoned.

PR #1096, opened 2020-04-06
ICU-21060 Fix behaviour of -per-, -and-, and dimensionless units.

hugovdm:
Without this fix:
- `"kilogram-per-meter-per-second"` incorrectly returns U_ILLEGAL_ARGUMENT_ERROR
- `"hertz-and-one-per-hour"` incorrectly becomes `"hertz-per-hour"`
- `"one-per-hour-and-hertz"` was supported, this "fix" removes that support. CLDR-13700 and CLDR-13701 may later decide what to do with negative powers and compound units within mixed units.

This fix also:

- removes support for `"one-and-one"` - considering it a not-useful use case.
- rejects `"kiloone"`, `"megaone"`, `"square-one"`, and `"square-kiloone"` for that matter - support for which was probably more side-effect than design. (They were all interpreted as just `"one"`, instead of, say, "thousands" and "millions" for `"kiloone"` and `"megaone"`.)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21060
- [x] Issue accepted
- [x] Tests included

Force-pushed to rebase onto maint/maint-67.


hugovdm:
@sffc Here's the next pull request - the second commit. It still contains the previous one as the first commit, because it looks like I don't yet have the rights to change ICU tickets to "accepted".

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -413,7 +418,8 @@ class Parser {
             }
             switch (token.getMatch()) {
                 case COMPOUND_PART_PER:
-                    if (fAfterPer) {
+                    if (fSawAnd) {
+                        // Mixed units can only contain single units.
```
hugovdm, line 422: This is according to the current proposed LDML spec.

We need to keep an eye on https://unicode-org.atlassian.net/browse/CLDR-13700 though, since that might make things quite a bit more complicated. For this, we probably need an ICU ticket as counterpart, which we can refer to in a TODO, is that correct?

hugovdm:
I filed a CLDR ticket to add the proposed negative-dimensionality identifier to the unit identifier spec:
https://unicode-org.atlassian.net/browse/CLDR-13701

As discussed offline: 
- simply not supporting mixed units was proposed-spec-compliant at the time of implementing, quick to do and seemed more future proof (in the sense that narrow implementation now, broadened in the future, is less breakage-prone than supporting things now that we then have to consider backward compatibility with as the spec changes and settles).
- the code changes are a little bit more complicated because the old code just drops the "one" and would instead need to parse it like a normal single unit, while throwing away its dimensionality (one-per-one, one-per-square-one, both mapping to 'one', though meter-per-square-meter stays exactly as is without simplification).

icu4c/source/i18n/measunit.cpp
```diff
@@ -537,9 +537,9 @@ static const char * const gSubTypes[] = {
     "solar-mass",
     "stone",
     "ton",
-    "one",
-    "percent",
-    "permille",
+    "", // FIXME(review): this should be generated by Java, from resource files.
+    "percent", // FIXME(review): should this be here? both percent and permille
```
sffc, line 541: For both of these items, file a bug to follow up in ICU 68, and make the TODO point to that bug.

icu4c/source/i18n/measunit.cpp
```diff
@@ -2020,6 +2020,20 @@ MeasureUnit::MeasureUnit(const MeasureUnit &other)
     *this = other;
 }
 
+// FIXME(review):
+// Can someone explain this please? I don't have a good grip on C++ "universal
```
sffc, line 2024: This is the move constructor.  Rather than copying the data from `other`, the data that `other` owns is moved into the new object, with no heap allocations, etc., since the old object is expected to be destroyed.  This is what `&&` means.

icu4c/source/i18n/measunit.cpp
```diff
@@ -2061,6 +2075,8 @@ MeasureUnit &MeasureUnit::operator=(MeasureUnit &&other) noexcept {
     }
     delete fImpl;
     fImpl = other.fImpl;
+    // FIXME(review): Here it is again, looks to the naive eye like "destruction
```
sffc, line 2078: Yeah, this is move assignment.

```
MeasureUnit foo(MeasureUnit::getSecond()); // move constructor
foo = MeasureUnit::getMeter(); // move assignment
```

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -94,7 +104,6 @@ const struct SIPrefixStrings {
 
 // TODO(ICU-21059): Get this list from data
 const char16_t* const gSimpleUnits[] = {
-    u"one", // note: expected to be index 0
```
sffc, line 97: Since the SimpleUnitImpl default constructor defaults to index 0, I feel like we should keep an empty string here.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -339,10 +347,14 @@ class Parser {
     }
 
 private:
+    // Tracks parser progress: the offset into fSource.
     int32_t fIndex = 0;
     StringPiece fSource;
     UCharsTrie fTrie;
 
+    // Set to true when we've seen a "-per-", after which all units are in the
```
sffc, line 355: After we've seen a "-per-" or a "per-"

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -422,24 +460,28 @@ class Parser {
...
 
                 case COMPOUND_PART_TIMES:
-                    if (fAfterPer) {
-                        result.dimensionality = -1;
+                    if (previ == 0) {
+                        // Can't start with "-".
+                        status = kUnitIdentifierSyntaxError;
+                        return;
                     }
+                    if (fAfterPer) { result.dimensionality = -1; }
```
sffc, line 468: I would put this on separate lines.  It's easier when reading code to see this when it's on its own line.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -383,37 +401,57 @@ class Parser {
...
         // 0 = no tokens seen yet (will accept power, SI prefix, or simple unit)
         // 1 = power token seen (will not accept another power token)
         // 2 = SI prefix token seen (will not accept a power or SI prefix token)
         int32_t state = 0;
         int32_t previ = fIndex;
 
-        // Maybe read a compound part
-        if (fIndex != 0) {
+        // Read a unit
+        while (hasNext()) {
```
sffc, line 428: Can you make sure that your code doesn't allow identifiers such as `square-per-meter` or `metersecond` to sneak through?

The logic before was basically, if we aren't at the beginning of the string, then a TYPE_COMPOUND_PART token is *required* here.  Not optional, but required.

I think your new logic should be, if we are at the beginning of the string, then allow an optional 'per-'; otherwise, require a '-', '-per-', or '-and-'.  Maybe that requires putting `per-` in its own section in the trie; that's probably fine.

icu4c/source/i18n/measunit.cpp
hugovdm: Did the "one" that you added come from ICU4J as per the normal instructions? Or was it a hand-edit?

I'm wondering if it matters that I'm hand-editing this generated code - I feel I'm asking for trouble, but diffs and unit tests should at least catch that trouble?

icu4c/source/i18n/measunit.cpp
hugovdm: Aha: I see VSCode is also simply failing to resolve this correctly, showing the wrong documentation ("copy constructor"). Looking at the header file myself, I can see it's clearly documented ("Move constructor"), thanks for clarifying!

icu4c/source/i18n/measunit.cpp
hugovdm: #TIL about move assignment operators. :-D

icu4c/source/i18n/measunit_extra.cpp
hugovdm: I also changed SingleUnitImpl to use -1 - I'd have to change that back if we keep it here. My idea was to not confuse default-constructed and identifier-constructed SingleUnitImpl instances by treating them more differently.

Additionally I wasn't sure what the Trie would do with an empty string (UCharsTrieBuilder b; b.add(u"", ...);)

I'll leave this unresolved for now (first dealing with other feedback and sleeping), and either I'll come back to it after reviewing my decisions, or you'll come back to the decision while reviewing other changes related to it. :)

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Agreed - also noticed this just after I had "sent it for review" - will deal with it when I'm fresh in the morning, demonstrating the failures with proper unit tests before fixing it.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -692,6 +772,7 @@ const MeasureUnitImpl& MeasureUnitImpl::forMeasureUnit(
     if (measureUnit.fImpl) {
         return *measureUnit.fImpl;
     } else {
+        // FIXME test this code path.
```
hugovdm, line 775: By this FIXME, I'm reminding myself:

- the parser doesn't accept "" at the moment,
- getIdentifier() can return "" though, for MeasureUnit's not constructed via the parser,
- I'm expecting there to be a code path that fails therefore, of MeasureUnitImpl::forMeasureUnit can be invoked on a default-constructed MeasureUnit();

I need to convince myself that this can't happen (and then document it well enough to convince others), or I need to fix it.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Going the "-1" route has indeed created some pain in serializing. I'm far enough into this that I think I'll finish fixing it before I consider whether it's the wrong tree I'm barking up. :-/ If so, backtrack and restore the "dimensionless" SingleUnitImpl instances as "proper" instances.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Done: my conclusion was the parser needs to support "" - it's used internally by factory functions (they parse the output of getIdentifier).

Marking as resolved.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -270,28 +287,30 @@ class Token {
     enum Type {
         TYPE_UNDEFINED,
         TYPE_SI_PREFIX,
+        // Token type for "-per-", "-", "-and-", as well as "per-".
```
sffc, line 290: Not for `"per-"` any longer

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -395,63 +435,89 @@ class Parser {
...
-            Token token = nextToken(status);
-            if (U_FAILURE(status)) {
-                return;
+        // Tracks the starting offset of the token currently under
+        // consideration.
+        int32_t previ = fIndex;
+        Token token = nextToken(status);
+        if (U_FAILURE(status)) { return; }
+
+        if (previ == 0) {
```
sffc, line 466: I don't think you're actually using `previ` for anything.  Here, you can read `fIndex` instead of `previ`, and there isn't anywhere else in this function where you are reading the value from `previ`.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -483,37 +545,50 @@ class Parser {
                     status = kUnitIdentifierSyntaxError;
                     return;
             }
-        }
 
-        // We ran out of tokens before finding a complete single unit.
-        status = kUnitIdentifierSyntaxError;
+            previ = fIndex;
+            token = nextToken(status);
+            if (U_FAILURE(status)) {
```
sffc, line 551: I'd have a slight preference for the following form:

```cpp
if (!hasNext()) {
  // We ran out of tokens before finding a complete single unit.
  status = kUnitIdentifierSyntaxError;
  return;
}
token = nextToken(status);
if (U_FAILURE(status)) {
  return;
}
```

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -483,37 +545,50 @@ class Parser {
...
         if (U_FAILURE(status)) {
             return;
         }
+        if (fSource.empty()) {
+            // The dimenionless unit: nothing to parse. leave result as is.
+            return;
+        }
         int32_t unitNum = 0;
+        if (!hasNext()) {
+            // We don't support the empty string.
```
sffc, line 571: Not relevant any more.  You could `U_ASSERT(hasNext())` or just remove this if statement.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -483,37 +545,50 @@ class Parser {
...
             }
             if ((++unitNum) >= 2) {
-                UMeasureUnitComplexity complexity = sawAnd
-                    ? UMEASURE_UNIT_MIXED
-                    : UMEASURE_UNIT_COMPOUND;
+                UMeasureUnitComplexity complexity =
+                    sawAnd ? UMEASURE_UNIT_MIXED : UMEASURE_UNIT_COMPOUND;
                 if (unitNum == 2) {
                     U_ASSERT(result.complexity == UMEASURE_UNIT_SINGLE);
                     result.complexity = complexity;
```
sffc, line 594: You catch mixed and/per sooner than this code did before.  Is it still possible for `result.complexity != complexity` on the following line?  If not, maybe remove it or add an assertion.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -702,7 +794,9 @@ MeasureUnitImpl MeasureUnitImpl::forMeasureUnitMaybeCopy(
     if (measureUnit.fImpl) {
         return measureUnit.fImpl->copy(status);
     } else {
-        return Parser::from(measureUnit.getIdentifier(), status).parse(status);
+        const char *identifier = measureUnit.getIdentifier();
+        if (identifier[0] == 0) { return MeasureUnitImpl(); }
```
sffc, line 798: Shouldn't this case be covered by the `if (fSource.empty()) {` on line 565?

icu4c/source/test/intltest/measfmttest.cpp
```diff
@@ -3413,34 +3410,119 @@ void MeasureFormatTest::TestCompoundUnitOperations() {
...
+    assertEquals("dimensionless dimensionality", 1, dimensionality);
+
+    // dimensionless.withDimensionality()
+    modified = dimensionless.withDimensionality(-1, status);
+    status.errIfFailureAndReset("dimensionless.withDimensionality(...)");
+    singles = modified.splitToSingleUnits(count, status);
+    assertEquals("no singles in modified", 0, count);
+    dimensionality = modified.getDimensionality(status);
+    status.errIfFailureAndReset("modified.getDimensionality(...)");
+    assertEquals("modified dimensionality", 1, dimensionality);
```
sffc, line 3473: Should dimensionality be 0 for the dimensionless unit?

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Going with a bool: there's no default constructor for token, and I need the value of fIndex from before the nextToken call.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -597,6 +597,7 @@ class Parser {
                     status = kUnitIdentifierSyntaxError;
                     return;
                 }
+                result.complexity = sawAnd ? UMEASURE_UNIT_MIXED : UMEASURE_UNIT_COMPOUND;
```
sffc, line 600: Do you need this line?  At this point I think you could assert that result.complexity == complexity.

icu4c/source/test/intltest/measfmttest.cpp
sffc: I understand why it returns 1, but I'm thinking that it's actually more correct to return 0.  It also gives code an easy way to detect the base dimensionless unit.

Spend 30 minutes seeing if you can make it work returning 0.  Maybe make SimpleUnitImpl default to 0 instead of 1, and set it to 1 at the top of the parse function after it's seen at least one token?  If you can get it to work, let's add that change.  If not, leave the value at 1 and perhaps file a follow-up ticket.

hugovdm: This gets messy without setters and getters: I would love to maintain the invariant that if (index < 0), then (dimensionality == 0), and vice versa. However in the parsing, the dimensionality gets set first, before we know whether there will be an index to set. If the index does not get set, the dimensionality needs to get set back to 0 if we care about the invariant.

I already added isDimensionless() - I think we should just define "dimensionality" as meaningless (NaN!) when it's the dimensionless unit.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Whoops - that was the one-liner I temporarily replaced the block with. Incomplete rollback. :-)

icu4c/source/test/intltest/measfmttest.cpp
hugovdm: My proposal (as per latest commit): getting explicit with dimensionality being as meaningless for dimensionless units as for compound units.

icu4c/source/i18n/measunit_impl.h
```diff
@@ -41,8 +41,8 @@ struct SingleUnitImpl : public UMemory {
      * considered the same as square-meter.
      *
      * The dimensionless unit generally does not get compared, but if it did, it
-     * would sort before other units by virtue of dimensionality being 0 and
-     * index being < 0.
+     * would sort before other units by virtue of index being < 0 and
+     * dimensionality not being negative.
```
hugovdm, line 45: I verified that dimensionless units never got compared by temporarily adding a set of asserts.

I wasn't sure whether to leave those asserts in there... I chose not to, because it's not fundamentally necessary: code could go compare with them if it wants.

sffc: OK.  Fine to leave as-is

icu4c/source/i18n/measunit_impl.h
```diff
@@ -78,13 +89,20 @@ struct SingleUnitImpl : public UMemory {
     /**
      * Simple unit index, unique for every simple unit, -1 for the dimensionless
      * unit. This is an index into a string list in measunit_extra.cpp.
+     *
+     * The default value is -1, meaning the dimensionless unit:
+     * isDimensionless() will return true, until index is changed.
      */
     int32_t index = -1;
 
     /** SI prefix. **/
```
sffc, line 98: Optional: If you add the comment about dimensionality being meaningless, you should probably also say that SI prefix is meaningless

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -744,11 +838,21 @@ MeasureUnit MeasureUnit::withSIPrefix(UMeasureSIPrefix prefix, UErrorCode& statu
...
+    if (U_FAILURE(status)) { return 0; }
+    if (singleUnit.isDimensionless()) {
+        return 0;
+    }
+    return singleUnit.dimensionality;
 }
 
 MeasureUnit MeasureUnit::withDimensionality(int32_t dimensionality, UErrorCode& status) const {
     SingleUnitImpl singleUnit = SingleUnitImpl::forMeasureUnit(*this, status);
+    if (U_FAILURE(status)) { return MeasureUnit(); }
```
sffc, line 851: Optional: you don't actually need lines 851-855 that you added; I would remove them, but you can keep them if you like how it makes getDimensionality more parallel to withDimensionality

hugovdm: Done.

icu4c/source/i18n/measunit_impl.h
hugovdm: Done.

sffc:
LGTM please squash

PR #1098, opened 2020-04-07
ICU-21060 Fix heap-use-after-free bug.

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21060
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -321,6 +321,14 @@ private:
 
 class Parser {
 public:
+    /**
+     * Factory function for parsing the given identifier.
+     *
+     * @param source The identifier to parse. This function does not make a copy
+     * of source: the underlying string that source points at, must outlive the
+     * parser.
+     * @param status ICU error code.
+     */
     static Parser from(StringPiece source, UErrorCode& status) {
         if (U_FAILURE(status)) {
             return Parser();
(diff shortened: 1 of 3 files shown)


icu4c/source/test/intltest/measfmttest.cpp
```diff
@@ -82,6 +82,7 @@ class MeasureFormatTest : public IntlTest {
     void TestInvalidIdentifiers();
     void TestCompoundUnitOperations();
     void TestIdentifiers();
+    void TestAddressSanitizerProblem();
```
sffc, line 85: Suggested:

```cpp
void Test21060_AddressSanitizerProblem();
```

in case there's another address sanitizer problem in the future.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -321,6 +321,12 @@ class Token {
 
 class Parser {
 public:
+    /**
+     * Factory function for parsing the given identifier.
+     *
+     * @param source The identifier to parse. This function does not make a copy
+     * of source: the data source points at must outlive the parser.
```
sffc, line 328: Check the grammar of this sentence

icu4c/source/test/intltest/measfmttest.cpp
hugovdm: Done (i.e. coming up in my next push). I've also moved the actual body of this code to be below the body of TestIdentifiers, FWIW. In the body, I improved the comments.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Rewrote it, thanks! ("data" was way too ambiguous, looking like an adjective rather than a noun.)

PR #1099, opened 2020-04-07
ICU-5938 Fix inaccurate output on RBNF demo sample

robertgmelo:
ICU-5938 - Fix inaccurate output on RBNF demo sample

- Replace the attribute type of the attribute from double to
BigDecimal in order to format a number by using RBNF rules
without rounding error issues.

- A modification on < and > buttons (multiply/divide by 10)
to also fix rounding error issues.

- @Override annotations automatically added.

### DIFF
--- a/icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
+++ b/icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
@@ -45,6 +45,7 @@ import java.text.ParsePosition;
 import java.util.Locale;
 
 import com.ibm.icu.dev.demo.impl.DemoApplet;
+import com.ibm.icu.math.BigDecimal;
 import com.ibm.icu.text.RuleBasedNumberFormat;
 
 public class RbnfDemo extends DemoApplet {
@@ -66,10 +67,12 @@ public class RbnfDemo extends DemoApplet {
         new RbnfDemo().showDemo();
     }
 
+    @Override
     protected Dimension getDefaultFrameSize(DemoApplet applet, Frame f) {
         return new Dimension(430,270);
     }
 
+    @Override
     protected Frame createDemoFrame(DemoApplet applet) {
         final Frame window = new Frame("Number Spellout Demo");
         window.setSize(800, 600);
@@ -89,7 +92,7 @@ public class RbnfDemo extends DemoApplet {
         populateRuleSetMenu();
         numberFormatter = new DecimalFormat("#,##0.##########");
         parsePosition = new ParsePosition(0);
-        theNumber = 0;
+        theNumber = BigDecimal.ZERO;
 
         numberField = new TextField();
         numberField.setFont(new Font("Serif", Font.PLAIN, 24));
@@ -100,6 +103,7 @@ public class RbnfDemo extends DemoApplet {
         lenientParseButton = new Checkbox("Lenient parse", lenientParse);
 
         numberField.addTextListener(new TextListener() {
+            @Override
             public void textValueChanged(TextEvent e) {
                 if (!numberFieldHasFocus)
                     return;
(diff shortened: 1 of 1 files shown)


markusicu:
cc @grhoten

icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
```diff
@@ -108,37 +112,40 @@ public void textValueChanged(TextEvent e) {
...
                 if (temp == null || parsePosition.getIndex() == 0) {
-                    theNumber = 0;
+                    theNumber = BigDecimal.ZERO;
                     textField.setText("PARSE ERROR");
                 }
                 else {
-                    theNumber = temp.doubleValue();
-                    textField.setText(spelloutFormatter.format(theNumber, ruleSetName));
+                    theNumber = new BigDecimal(fieldText);
+                    textField.setText(spelloutFormatter.format(theNumber.doubleValue(), ruleSetName));
```
grhoten, line 120: I will admit that I initially did this quick hack when doing the demo for the [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp). If you're trying to do this accurately, then you should at least check if it has a fraction. If it doesn't have a fraction, then you should use the long value instead.

You may want to look at the more maintained Number Format Tester here: https://github.com/unicode-org/cldr/blob/master/tools/cldr-apps/WebContent/numbers.jsp

grhoten:
Actually my feedback was that it doesn't look good because it probably won't handle INT64_MAX or Long.MAX_VALUE.  There are tests in ICU to make sure that such values work correctly.

srl295:
OK, so it was not ready to merge.
@robertgmelo can you make a followon PR to address the feedback?

robertgmelo:
I understand your point, @grhoten. 

Do you agree we were already using double type for that? In any case, I agree it's a plus to guarantee more accuracy and we should apply your suggestion. 

So just to confirm: are you referring to the `theNumber.doubleValue()` occurrences, so we should check if it has fraction, then use `double`, otherwise `long`. Is that correct?

grhoten:
To be honest, I can't use this demo. It's hard to get set up and use. The numero demo at http://demo.icu-project.org/icu4jweb/numero.jsp is also hard to use because you don't know which locales are valid. You can't specify different rule types, and it's hard to see ranges of numbers. It's also very slow compared to the one from CLDR.

Instead of having at least 3 different demos for RBNF, I'd rather just collapse it into 1. I'd prefer to delete this entire demo, but perhaps someone is actually using it.

In any case... Yes, when theNumber has a fraction, use doubleValue, otherwise use longValue.

srl295:
> To be honest, I can't use this demo. It's hard to get set up and use.

> The numero demo at http://demo.icu-project.org/icu4jweb/numero.jsp is also hard to use because you don't know which locales are valid. You can't specify different rule types, and it's hard to see ranges of numbers. It's also very slow compared to the one from CLDR.

I wonder why it's slow (well, it's down at the moment. but that's not directly related.) Anyway, I'd agree Numero could be merged into something else and/or just retired.

> Instead of having at least 3 different demos for RBNF, I'd rather just collapse it into 1. I'd prefer to delete this entire demo, but perhaps someone is actually using it.

The benefit of this demo is it's embedded with ICU4J.

What do you think about moving the CLDR demo (which is useful for CLDR but really an ICU demo) into ICU4J's demo space? Then linking to it from CLDR.

> In any case... Yes, when theNumber has a fraction, use doubleValue, otherwise use longValue.

grhoten:
From my experience, the icu4jweb demo has been slow and less maintained. I give the CLDR tool link out very frequently when someone wants to review or create RBNF rules. So I really want the current CLDR link to stay.

srl295:
> From my experience, the icu4jweb demo has been slow 

It's currently proxied, but it does have a direct link which should be quick.

> and less maintained.

Want to help ? : )

> I give the CLDR tool link out very frequently when someone wants to review or create RBNF rules. So I really want the current CLDR link to stay.

If the current CLDR code moved with a redirect, to somewhere you could easily make changes to, would you +1 it?

grhoten:
@srl295 I'd rather just have the status quo. Sorry I'm already overloaded with tasks at this time.

srl295:
Fair enough

PR #1102, opened 2020-04-08
ICU-21069 Fix ucptrie_swap pointer logic

FrankYFTang:
Also include unit tests for 8bits.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21069
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/ucptrietest.c
+++ b/icu4c/source/test/cintltst/ucptrietest.c
@@ -1145,6 +1145,9 @@ TrieTestSet3Initial9(void) {
     testTrieRanges("set3-initial-9", FALSE,
         setRanges3, UPRV_LENGTHOF(setRanges3),
         checkRanges3, UPRV_LENGTHOF(checkRanges3));
+    testTrieRanges("set3-initial-9-clone", TRUE,
+        setRanges3, UPRV_LENGTHOF(setRanges3),
+        checkRanges3, UPRV_LENGTHOF(checkRanges3));
 }
 
 static void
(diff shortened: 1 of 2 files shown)


icu4c/source/test/cintltst/ucptrietest.c
```diff
@@ -1103,6 +1103,47 @@ checkRanges3[]={
     { 0x110000, 6 }
 };
 
+/* set some interesting overlapping ranges */
+/* similar to setRange2 but use only 8 bits value */
```
markusicu, line 1107: typo: setRange2 --> setRanges2 (plural)

Also, most of the other test cases already exercises 8-bit-value tries, so we should have coverage and should not need to add more test cases. That is, I assume that we need not change this file at all.

markusicu: Actually, I think it would be good to change this test code, but by looking at why it didn't catch this problem, and fixing it.

I think I see the problem: The swapping code is only tested when testTrieRanges() is called with withClone=TRUE. This only happens in TrieTestSet2OverlapWithClone() which uses 16-bit values.

I suggest you just change, say, TrieTestSet3Initial9() to also call testTrieRanges() with withClone=TRUE.

FrankYFTang:
PTAL

icu4c/source/test/cintltst/ucptrietest.c
```diff
@@ -1145,6 +1145,9 @@ TrieTestSet3Initial9(void) {
     testTrieRanges("set3-initial-9", FALSE,
         setRanges3, UPRV_LENGTHOF(setRanges3),
         checkRanges3, UPRV_LENGTHOF(checkRanges3));
+    testTrieRanges("set3-initial-9", TRUE,
```
markusicu, line 1148: Please either just *change* the flag in the existing call, or else add a suffix to the test name (e.g., "set3-initial-9-clone").

FrankYFTang: Done

FrankYFTang:
PTAL

PR #1103, opened 2020-04-09
Removes **emphasis** in quoted text

filmil:
Quoted text does not get emphasized with double-asterisk inside a markdown blockquote,
so keeping it there leads to syntactically incorrect MessageFormat.

##### Checklist

- [ ] Issue accepted
- [ ] Tests included


markusicu:
Sorry, it looks like we missed this. The User Guide has been further fixed up, and merged into the master branch.
Feel free to open a new PR against that if there are still problems.

PR #1104, opened 2020-04-09
ICU-5938 Fix inaccurate output on RBNF demo sample (accuracy)

robertgmelo:
ICU-5938 Fix inaccurate output on RBNF demo sample
    
    - In order to guarantee more accuracy on formatting, check
    if number has fraction. If so, use double. Otherwise, use long.

### DIFF
--- a/icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
+++ b/icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
@@ -116,8 +116,9 @@ public class RbnfDemo extends DemoApplet {
                     textField.setText("PARSE ERROR");
                 }
                 else {
-                    theNumber = new BigDecimal(fieldText);
-                    textField.setText(spelloutFormatter.format(theNumber.doubleValue(), ruleSetName));
+                    theNumber = new BigDecimal(temp instanceof Long ? temp.longValue() : temp.doubleValue());
+                    textField.setText(spelloutFormatter.format(
+                            theNumber.scale() == 0 ? theNumber.longValue() : theNumber.doubleValue(), ruleSetName));
                 }
             }
         } );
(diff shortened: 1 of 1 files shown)


icu4j/demos/src/com/ibm/icu/dev/demo/rbnf/RbnfDemo.java
```diff
@@ -116,8 +116,9 @@ public void textValueChanged(TextEvent e) {
                     textField.setText("PARSE ERROR");
                 }
                 else {
-                    theNumber = new BigDecimal(fieldText);
-                    textField.setText(spelloutFormatter.format(theNumber.doubleValue(), ruleSetName));
+                    theNumber = new BigDecimal(temp instanceof Long ? temp.longValue() : temp.doubleValue());
+                    textField.setText(spelloutFormatter.format(
+                            theNumber.scale() == 0 ? theNumber.longValue() : theNumber.doubleValue(), ruleSetName));
```
richgillam, line 121: Why'd you need to change this part?  Does the `BigDecimal` constructor fail to set its scale in such a way that you can tell the difference between an integer and a rational number later on?  Or is there some other special behavior you're getting out of `numberFormatter.parse()` that you don't get out of `BigDecimal`'s constructor?  On the surface, it seems like the code you had before would do what you need and would be more efficient than what you're doing now.

robertgmelo: Sure, I've changed this part because the input text accepts non-number characters (e.g. letters). So in case the user enters something like "1000abc", if we use the text content directly we can observe exceptions.

So I've used the temp formatted number, which ignores non-number info, and combined with Hoten's suggestion, to guarantee the accuracy and avoid undesired exceptions.

Note: I also though in an improvement by filtering the input type to accept only numbers and decimal separators, but as it's not the focus of the issue, so I just consider the inaccurate aspect.

richgillam: Okay, that sounds reasonable to me.  Thanks for the explanation!

grhoten:
Looks good!

robertgmelo:
Thanks for the feedback, Rich and George. I'll squash and wait Steven's review.

PR #1105, opened 2020-04-09
ICU-20926 Adding test to verify fallback rule of DateIntervalFormat

caiolima:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20926
- [x] Issue accepted
- [x] Tests included

Cc @sffc 

I added tests with some skeletons, however I was not able to find the skeleton that would result in the same pattern we get from `DateFormate::createTimeInstance (DateFormat::kFull, DateFormat::kFull, locale);`.  Considering the test as it is, it is passing. After I get green light for it, I'll add Java version of this test.

### DIFF
--- a/icu4c/source/test/intltest/dtifmtts.cpp
+++ b/icu4c/source/test/intltest/dtifmtts.cpp
@@ -60,6 +60,8 @@ void DateIntervalFormatTest::runIndexedTest( int32_t index, UBool exec, const ch
         TESTCASE(11, testCreateInstanceForAllLocales);
         TESTCASE(12, testTicket20707);
         TESTCASE(13, testFormatMillisecond);
+        TESTCASE(14, testTicket20710);
+
         default: name = ""; break;
     }
 }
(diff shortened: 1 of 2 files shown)


FrankYFTang:
Is the Java test already have this? If not, could you also add the same test to ICU4J.

sffc:
Good, this is the first half of what the ticket is requesting that we test.  The second half:

> We should add a test for this behavior: UFIELD_CATEGORY_DATE_INTERVAL_SPAN should be present in the UFormattedValue *if and only if* a date interval was formatted. If the field is not present, then that should imply that no date interval was formatted.

Can you check for UFIELD_CATEGORY_DATE_INTERVAL_SPAN and test that it is present *if and only if* a date interval was formatted?

sffc:
@caiolima Do you have plans to respond to the feedback on this PR?

caiolima:
> @caiolima Do you have plans to respond to the feedback on this PR?

Oops, I feel this got lost in my notification list for quite a while. I'm quite busy on other tasks and I don't think will be able to work there in the near term.

markusicu:
ping

sffc:
I updated the branch and responded to the feedback in https://github.com/unicode-org/icu/pull/2245.

PR #1115, opened 2020-04-18
ICU-21053 Fix compile problems with UCONFIG_NO_FORMATTING = 1

sven-oly:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21053
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/listformatter.cpp
+++ b/icu4c/source/i18n/listformatter.cpp
@@ -348,6 +348,7 @@ const ListFormatInternal* ListFormatter::getListFormatInternal(
     return result;
 }
 
+#if !UCONFIG_NO_FORMATTING
 static const char* typeWidthToStyleString(UListFormatterType type, UListFormatterWidth width) {
     switch (type) {
         case ULISTFMT_TYPE_AND:
@@ -391,6 +392,7 @@ static const char* typeWidthToStyleString(UListFormatterType type, UListFormatte
 
     return nullptr;
 }
+#endif
 
 static const UChar solidus = 0x2F;
 static const UChar aliasPrefix[] = { 0x6C,0x69,0x73,0x74,0x50,0x61,0x74,0x74,0x65,0x72,0x6E,0x2F }; // "listPattern/"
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/listformatter.cpp
```diff
@@ -348,6 +348,7 @@ const ListFormatInternal* ListFormatter::getListFormatInternal(
     return result;
 }
 
+#if !UCONFIG_NO_FORMATTING
```
sffc, line 351: Why do you need to block out this method?  Because it depends on the enum symbols from ulistformatter.h?  It doesn't make sense that this method is blocked but not the others on this class.  Is the more correct fix to pull the enums in ulistformatter.h *out of* UCONFIG_NO_FORMATTING?  I don't know the answer to that.

sffc:
Also, what's up with the base branch?  Why are you merging this into `unicode-org:ICU-21053` instead of master or maint/maint-67?

icu4c/source/i18n/unicode/listformatter.h
```diff
@@ -199,7 +200,8 @@ class U_I18N_API ListFormatter : public UObject{
     static ListFormatter* createInstance(
       const Locale& locale, UListFormatterType type, UListFormatterWidth width, UErrorCode& errorCode);
 #endif  /* U_HIDE_DRAFT_API */
-
+#endif  /* !UCONFIG_NO_FORMATTING */
```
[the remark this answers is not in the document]
sven-oly: Updated.

icu4c/source/i18n/listformatter.cpp
```diff
@@ -511,9 +513,10 @@ ListFormatter* ListFormatter::createInstance(UErrorCode& errorCode) {
 }
 
 ListFormatter* ListFormatter::createInstance(const Locale& locale, UErrorCode& errorCode) {
-    return createInstance(locale, ULISTFMT_TYPE_AND, ULISTFMT_WIDTH_WIDE, errorCode);
+  return createInstance(locale, "standard", errorCode);
```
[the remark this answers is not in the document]
sven-oly: Fixed.

icu4c/source/i18n/listformatter.cpp
sven-oly: Updated to conditionally remove the method and the call. Moving the enums in ulistformatter.h doesn't seem right because everything else is in that file is conditioned on UCONFIG_NO_FORMATTING.

sven-oly: This doesn't compile if UNCONFIG_NO_FORMATTING = 1.  The problem is that this is only used in the previous call using the new interface. I'm updating this to put both the call and this function under UCONFIG_NO_FORMATTING.

PR #1118, opened 2020-04-21
ICU-21085 Fix issue of __has_attribute on Windows

FrankYFTang:
Fix "source\i18n\double-conversion-utils.h(77):
warning C4067: unexpected tokens following preprocessor directive - expected a newline"

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21085
- [ ] Issue accepted
- [ ] Tests included


FrankYFTang:
@jefgen  mentioned the issue is fixed in the upstream in https://github.com/google/double-conversion/pull/131/files

I think we should pull from that instead.

PR #1119, opened 2020-04-21
ICU-21084 Migrating ICU tools to use PathMatcher and refactoring

hagbard:
This is almost the last planned work for the new ICU data conversion tools. Once this is done it should be easy to move the pseudo-locale logic into the CLDR side (which is what Markus wanted).

NOTE: This code should not be committed until the last release branch has been merged back in (and any changes reconciled). The output from this code is identical to the current code (on the master branch) for all CLDR data.

There's some refactoring here to use the new CldrDataProcessor class, which is intended to allow a static, pure functional processor to be built and make a clean split between describing how data is to be processed vs processing the data at runtime. The code looks a bit different in a few cases, but is typically (a) shorter and (b) less error prone (since the new PathMatcher is stricter about how it matches).

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21084
- [ ] Issue accepted
- [x] Tests included


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/CollationMapper.java
```diff
@@ -68,88 +71,76 @@
...
-        icuSpecialData.ifPresent(s -> s.accept(DTD, visitor));
-        cldrData.accept(DTD, visitor);
-        return visitor.icuData;
+        CollationMapper mapper = new CollationMapper(icuData, cldrVersion);
+        icuSpecialData.ifPresent(specialData -> CLDR_PROCESSOR.process(specialData, mapper, DTD));
+        CLDR_PROCESSOR.process(cldrData, mapper, DTD);
+        return icuData;
     }
 
-    final static class CollationVisitor implements PrefixVisitor {
```
hagbard, line 77: Note that in the new code there's no need to directly implement the path visitors anymore. You just bind paths and actions together and let the processor handle visitation. This lets us remove this class and directly implement the mutable collection state in the outer mapper.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/CollationMapper.java
```diff
@@ -68,88 +71,76 @@
...
+                    .map(CollationMapper::removeComment)
+                    .filter(s -> !s.isEmpty())::iterator);
+            icuData.replace(rbPath, rules);
+            icuData.replace(RbPath.of("collations", type, "Version"), cldrVersion);
         }
+    }
 
-        private void collectRules(CldrValue v) {
-            CldrPath p = v.getPath();
-            if (COLLATION_RULE.matchesSuffixOf(p)) {
```
hagbard, line 106: What used to be conditional code with explicit matching logic now becomes 2 different callbacks in the new code.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/mapper/PluralsMapper.java
```diff
@@ -35,15 +34,21 @@
...
+    private static final AttributeKey PLURALS_TYPE = keyOf("plurals", "type");
     private static final AttributeKey RULES_LOCALES = keyOf("pluralRules", "locales");
-
-    private static final PathMatcher RULE = PathMatcher.of("pluralRule[@count=*]");
     private static final AttributeKey RULE_COUNT = keyOf("pluralRule", "count");
 
+    private static final CldrDataProcessor<PluralsMapper> CLDR_PROCESSOR;
+    static {
+        CldrDataProcessor.Builder<PluralsMapper> processor = CldrDataProcessor.builder();
+        processor
```
hagbard, line 45: This is really only processing one path (pluralRule) but it's done as 3 nested actions because of the need to group these rules in the generated data. Each addAction provides a new scope in which the child actions will be processed. This makes is clear that:
* Each `Rule` is collected by a `Rules` instance
* `Rules` are collected by a `Plurals` instance
* The mapper (top level) collects `Plurals`

hagbard:
As per always, github has messed up this PR so it's no longer just my changes. I'll fix it, but I also don't want to lose the comments I added (which will definitely happen).
God I hate github (git is fine, github is awful).

markusicu:
> As per always, github has messed up this PR so it's no longer just my changes. I'll fix it, but I also don't want to lose the comments I added (which will definitely happen).

Ok. Please ping us when the "Files changed" view is cleaned up.

hagbard:
> > As per always, github has messed up this PR so it's no longer just my changes. I'll fix it, but I also don't want to lose the comments I added (which will definitely happen).
> 
> Ok. Please ping us when the "Files changed" view is cleaned up.

Done.

hagbard:
I'm merging now since there's an issue with pseudo-locales that's best fixed on top of these changes.

PR #1122, opened 2020-04-21
ICU-21071 Fix lenient parse rules

robertgmelo:
ICU-21071 Fix lenient parse rules
- Check non-lenient rules before call lenint parsing
- Remove logKnownIssue 9503 from test code
- Adjust TestAllLocales test on ICU4C
- Add lenient checks on ICU4J

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21071
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/nfrule.cpp
+++ b/icu4c/source/i18n/nfrule.cpp
@@ -1297,6 +1297,10 @@ NFRule::prefixLength(const UnicodeString& str, const UnicodeString& prefix, UErr
 #if !UCONFIG_NO_COLLATION
     // go through all this grief if we're in lenient-parse mode
     if (formatter->isLenient()) {
+        // Check if non-lenient rule finds the text before call lenient parsing
+        if (str.startsWith(prefix)) {
+            return prefix.length();
+        }
         // get the formatter's collator and use it to create two
         // collation element iterators, one over the target string
         // and another over the prefix (right now, we'll throw an
@@ -1505,9 +1509,15 @@ NFRule::findText(const UnicodeString& str,
         return str.indexOf(key, startingAt);
     }
     else {
-        // but if lenient parsing is turned ON, we've got some work
-        // ahead of us
-        return findTextLenient(str, key, startingAt, length);
+        // Check if non-lenient rule finds the text before call lenient parsing
+        *length = key.length();
+        int32_t pos = str.indexOf(key, startingAt);
+        if(pos >= 0) {
+            return pos;
+        } else {
+            // but if lenient parsing is turned ON, we've got some work ahead of us
+            return findTextLenient(str, key, startingAt, length);
+        }
     }
 }
 
--- a/icu4c/source/i18n/plurfmt.cpp
+++ b/icu4c/source/i18n/plurfmt.cpp
@@ -549,9 +549,15 @@ void PluralFormat::parseType(const UnicodeString& source, const NFRule *rbnfLeni
 
         UnicodeString currArg = pattern.tempSubString(partStart->getLimit(), partLimit->getIndex() - partStart->getLimit());
         if (rbnfLenientScanner != NULL) {
-            // If lenient parsing is turned ON, we've got some time consuming parsing ahead of us.
-            int32_t length = -1;
-            currMatchIndex = rbnfLenientScanner->findTextLenient(source, currArg, startingAt, &length);
+            // Check if non-lenient rule finds the text before call lenient parsing
+            int32_t tempIndex = source.indexOf(currArg, startingAt);
+            if (tempIndex >= 0) {
+                currMatchIndex = tempIndex;
+            } else {
+                // If lenient parsing is turned ON, we've got some time consuming parsing ahead of us.
+                int32_t length = -1;
+                currMatchIndex = rbnfLenientScanner->findTextLenient(source, currArg, startingAt, &length);
+            }
         }
         else {
             currMatchIndex = source.indexOf(currArg, startingAt);
--- a/icu4c/source/test/intltest/itrbnf.cpp
+++ b/icu4c/source/test/intltest/itrbnf.cpp
@@ -1889,16 +1889,19 @@ IntlTestRBNF::TestAllLocales()
             UErrorCode status = U_ZERO_ERROR;
             RuleBasedNumberFormat* f = new RuleBasedNumberFormat((URBNFRuleSetTag)j, *loc, status);
 
-            if (status == U_USING_DEFAULT_WARNING || status == U_USING_FALLBACK_WARNING) {
-                // Skip it.
-                delete f;
-                break;
-            }
             if (U_FAILURE(status)) {
                 errln(UnicodeString(loc->getName()) + names[j]
                     + "ERROR could not instantiate -> " + u_errorName(status));
                 continue;
             }
+
+            Locale actualLocale = f->getLocale(ULOC_ACTUAL_LOCALE, status);
+            if (actualLocale != *loc) {
+                // Skip the redundancy
+                delete f;
+                break;
+            }
+
 #if !UCONFIG_NO_COLLATION
             for (unsigned int numidx = 0; numidx < UPRV_LENGTHOF(numbers); numidx++) {
                 double n = numbers[numidx];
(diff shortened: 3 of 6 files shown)


jefgen:
Note: The ticket is set for 67.1, but this PR is targeting the `master` branch rather than the `maint/maint-67` branch.

robertgmelo:
> Note: The ticket is set for 67.1, but this PR is targeting the `master` branch rather than the `maint/maint-67` branch.

Oh yeah, sorry. You're right. I did not pay attention on that, Jeff.

I can abort this, and submit a new to maint branch, in case is approved.

robertgmelo:
I totally understand your considerations, @richgillam.

As mentioned in the CR, all solutions I was trying was converging to existing non-lenient code. In some cases, it caused side-effects.

Indeed, by using lenient parsing, I could see the offset on collator iterator was working weirdly, for example, in somes cases, when next() is called, it jumps two positions, which affect the character position of the search results. So I didn't intend to affect other parts of the code. I've tried to be touch in the specific cases, but I agree there may have "hole" as you pointed.

One thing it can makes sense and improve code and reuse non-lenient rule checks before call lenient parsing, but you guys have more property to say.

I also tested locally for other cases, but the class itself allows a bunch of variations for the same locale, imagine for other locales.

Let's hear George.

Thanks for the prompt reply.

icu4c/source/i18n/nfrule.cpp
```diff
@@ -1505,9 +1509,15 @@ NFRule::findText(const UnicodeString& str,
         return str.indexOf(key, startingAt);
     }
     else {
-        // but if lenient parsing is turned ON, we've got some work
-        // ahead of us
-        return findTextLenient(str, key, startingAt, length);
+        // Check if non-lenient rule finds the text before call lenient parsing
+        *length = key.length();
+        int pos = str.indexOf(key, startingAt);
```
grhoten, line 1514: I think most of the time we try to use the exact type, like int32_t instead of the old int type. This is meant to improve portability. Though for strings, it's not that big of a deal these days.

icu4c/source/i18n/plurfmt.cpp
```diff
@@ -549,9 +549,15 @@ void PluralFormat::parseType(const UnicodeString& source, const NFRule *rbnfLeni
 
         UnicodeString currArg = pattern.tempSubString(partStart->getLimit(), partLimit->getIndex() - partStart->getLimit());
         if (rbnfLenientScanner != NULL) {
-            // If lenient parsing is turned ON, we've got some time consuming parsing ahead of us.
-            int32_t length = -1;
-            currMatchIndex = rbnfLenientScanner->findTextLenient(source, currArg, startingAt, &length);
+            // Check if non-lenient rule finds the text before call lenient parsing
+            int tempIndex = source.indexOf(currArg, startingAt);
```
grhoten, line 553: This should probably be changed to int32_t too.

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -1889,16 +1889,19 @@ IntlTestRBNF::TestAllLocales()
...
-            }
             if (U_FAILURE(status)) {
                 errln(UnicodeString(loc->getName()) + names[j]
                     + "ERROR could not instantiate -> " + u_errorName(status));
                 continue;
             }
+
+            Locale actualLocale = f->getLocale(ULOC_ACTUAL_LOCALE, status);
+            if (actualLocale != *loc) {
+                // Skip the redundancy
```
grhoten, line 1900: Yeah, this is probably more accurate these days.

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -1936,7 +1939,7 @@ IntlTestRBNF::TestAllLocales()
                             + UnicodeString(" -> ") + str + UnicodeString(" -> ") + num.getDouble());
                     }
                 }
-                if (!quick && !logKnownIssue("9503") ) {
```
grhoten, line 1939: 👍

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1162,6 +1191,26 @@ public void TestAllLocales() {
...
+                            logln(loc.getName() + names[j] + "success parse: " + s + " -> " + num);
+                        }
+                        if (j != 0) {
+                            // TODO: Fix the ordinal rules.
+                            continue;
+                        }
+                        if (n != num.doubleValue()) {
+                            errors.append("\n" + loc + names[j] + "got " + num + " expected " + n);
+                        }
+                    } catch (ParseException pe) {
```
grhoten, line 1209: This code seems redundant with the previous block. Please try to use a do while loop that just enumerates a lenient mode boolean from false to true?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1207,6 +1255,32 @@ else if (testParsing) {
         }
     }
 
+    void doLenientTest(RuleBasedNumberFormat formatter, String[][] testData) {
+        NumberFormat decFmt = new DecimalFormat("#,###.################");
+        formatter.setLenientParseMode(true);
```
grhoten, line 1260: Please refactor this to not have copypasta (redundant code). It makes it harder to maintain the code when the logic needs to change. You can create a new one function that just defaults the lenient mode argument.

icu4c/source/i18n/nfrule.cpp
```diff
@@ -1297,6 +1297,10 @@ NFRule::prefixLength(const UnicodeString& str, const UnicodeString& prefix, UErr
 #if !UCONFIG_NO_COLLATION
     // go through all this grief if we're in lenient-parse mode
     if (formatter->isLenient()) {
+        // Check if non-lenient rule finds the text before call lenient parsing
+        if (str.startsWith(prefix)) {
+          return prefix.length();
```
grhoten, line 1302: The rest of the code is using  4 space indentation. Please use the same indentation.

grhoten:
FYI if you want to grab my attention, tag me in the pull request, which is @grhoten. It's the same handle that I've used since 2000. I just happened to notice this pull request.

robertgmelo:
Thanks for the review, @grhoten. 

I'll apply the improvements you've mentioned.

icu4c/source/test/intltest/itrbnf.cpp
robertgmelo: @grhoten, let me put here the question I put on the CR: 
- Can we remove the _!quick_ guard as well?
By skipping the redundancy, the test now executes in ~2-3s (against 25~30s with redundancy).

icu4c/source/test/intltest/itrbnf.cpp
robertgmelo: Yeah, with previous check, the test was not being executed as intended.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
robertgmelo: @grhoten,
Hmm... I didn't get. In this case, the method **doLenientTest()** is different from **doTest()**. 

_setLenientParseMode(true)_ is called only in **doLenientTest()**.

For _NumberFormat decFmt = new DecimalFormat("#,###.################");_, I agree it was copy/paste, so it makes sense to refactor to a method.

icu4c/source/test/intltest/itrbnf.cpp
grhoten: I have no objections with removing the quick check.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
grhoten: Typically people use default arguments in such scenarios, like refactoring it the following way:

```
public void doTest(args...)
{
    doTest(args, false);
}
public void doTest(args..., boolean lenient)
{
...
    if (lenient) {
        rbnf.setLenient(true);
    }
...
}
```

robertgmelo: Got it, got it. So we can even reuse the doTest stuff in future, in case we want.

That's fine, George. I'll do an ammend.

robertgmelo:
@grhoten, improvements applied. When your time allows, please double-check.

Thanks in advance,
_Robert._

PR #1125, opened 2020-04-22
ICU-20949 Fix compound unit in "ar", "ne" locales

FrankYFTang:
Do not assume the "one" pattern always contains "{0}"

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20949
- [X] Issue accepted
- [X] Tests included


sffc:
Spec: https://unicode.org/reports/tr35/tr35-general.html#perUnitPatterns

> …get the compoundUnit pattern, and substitute the formatted numerator for `{0}` and the singular form of the denominator for `{1}`, after stripping the `{0}` and trimming spaces.

Before, we unconditionally stripped the `{0}`.  This change means that we only strip the `{0}` if it exists in the string; otherwise, we don't strip anything.

Seems reasonable, but the spec is not 100% clear. @macchiati @hugovdm @younies

sffc:
I approve the change, but it might be worth looking at whether the spec should be improved.

PR #1127, opened 2020-04-22
ICU-21052 Fix two compilations with UCONFIG_NO_CONVERSION option in t…

sven-oly:
…ools.toolutil

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21052
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/tools/toolutil/dbgutil.cpp
+++ b/icu4c/source/tools/toolutil/dbgutil.cpp
@@ -29,7 +29,7 @@ static const UnicodeString&  _fieldString(UDebugEnumType type, int32_t field, Un
     if(str == NULL) {
         return fillin.remove();
     } else {
-        return fillin = UnicodeString(str, ""); // optimize?
+        return fillin = UnicodeString(str, -1, US_INV);
     }
 }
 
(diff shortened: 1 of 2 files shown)


sven-oly:
Note that this does not fix compilation of icu4c/source/tools/tooluti, but only these two files. There is no requirement or expectation that all tools will compile with UCONFIG_NO_CONVERSION set to 1.

icu4c/source/tools/toolutil/dbgutil.cpp
```diff
@@ -29,7 +29,7 @@ static const UnicodeString&  _fieldString(UDebugEnumType type, int32_t field, Un
     if(str == NULL) {
         return fillin.remove();
     } else {
-        return fillin = UnicodeString(str, ""); // optimize?
+      return fillin = UnicodeString(str, -1, US_INV);
```
sffc, line 32: Why did you change the indentation?

sven-oly: Ah, didn't intend to. Updated.

PR #1129, opened 2020-04-23
ICU-21089 Adding check for alt-path consistency.

hagbard:
The will now report when an alt-path mapping exists but for which the target path is missing in that locale. This only triggers at the moment when an alt-path exists, it's not a consistency check over all of CLDR (this would be better as a new tool on the CLDR side using the new API).

One minor additional fix relating to directory creation (which prevented easy debugging).

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21089
- [ ] Issue accepted
- [ ] Tests included


markusicu:
This is a duplicate of PR #1130, right? If so, please close this one.

hagbard:
Not a duplicate, but done when the fixes to Norbert's issues came in.

PR #1131, opened 2020-04-24
ICU-21092 Add #include <cstdlib> for std::abs().

roubert:
The declaration for `std::abs()` can be found in many different header files (which is the reason that this code compiles as-is on many, but not all, systems), but it's only guaranteed to be found in either `<cstdlib>` or `<cmath>`.

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -8,6 +8,8 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include <cstdlib>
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -8,6 +8,8 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include <cstdlib>
```
markusicu, line 11: This include is already on line 17, added by PR #1117.

PR #1132, opened 2020-04-25
ICU-21081 Make U_ASSERT C++14 compatible

sffc:
@jefgen How/where would you recommend adding a C++14 build bot?

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21081
- [x] Issue accepted
- [ ] Tests included


jefgen:
> @jefgen How/where would you recommend adding a C++14 build bot?

Hmm, I think we could likely add another build bot for it -- though we should probably also update to Ubuntu 18.04 as well for it. Let me see...

PR #1134, opened 2020-04-26
ICU-21096 adjust logKnownIssues for ICU rbbitst

pedberg-icu:
Update failing rbbi tests to have a logKnownIssue that points to an open ticket; remove one logKnownIssue that does not produce a test failure.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21096
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/test/intltest/rbbitst.cpp
+++ b/icu4c/source/test/intltest/rbbitst.cpp
@@ -1326,34 +1326,37 @@ UBool RBBITest::testCaseIsKnownIssue(const UnicodeString &testCase, const char *
         const UChar *fString;
     } badTestCases[] = {
         {"10666", "GraphemeBreakTest.txt", u"\u0020\u0020\u0033"},    // Fake example, for illustration.
+        // The following tests were originally for
         // Issue 8151, move the Finnish tailoring of the line break of hyphens to root.
-        // This probably ultimately wants to be resolved by updating UAX-14, but in the mean time
-        // ICU is out of sync with Unicode.
-        {"8151",  "LineBreakTest.txt", u"-#"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\u0023"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u00a7"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\u00a7"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\U00050005"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\U00050005"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0e01"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\u0e01"},
+        // However, that ticket has been closed as fixed but these tests still fail, so
+        // ICU-21097 has been created to investigate and address these remaining issues.
+        {"21097",  "LineBreakTest.txt", u"-#"},
+        {"21097",  "LineBreakTest.txt", u"\u002d\u0308\u0023"},
+        {"21097",  "LineBreakTest.txt", u"\u002d\u00a7"},
+        {"21097",  "LineBreakTest.txt", u"\u002d\u0308\u00a7"},
+        {"21097",  "LineBreakTest.txt", u"\u002d\U00050005"},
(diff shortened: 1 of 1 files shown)


aheninger:
This is a funny situation.

LineBreakTest.txt is a file taken unmodified Unicode, and exactly reflects UAX-14 expected behavior.

ICU tailors UAX-14 behavior by promoting the Finnish tailoring to root. Which was ICU-8151. And is why the there are (expected) failures when checking LineBreakTest.txt. They are not a bug in ICU.

What really wants to happen is for UAX-14 to be updated. I had started to look at that, but it wasn't obvious how to get the desired behavior using the existing UAX rule syntax. I guess the new ICU-21097 could become a ticket for removing the ICU logKnownIssues when/if UAX-14 is udated.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -1326,34 +1326,37 @@ UBool RBBITest::testCaseIsKnownIssue(const UnicodeString &testCase, const char *
...
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\u00a7"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\U00050005"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\U00050005"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0e01"},
-        {"8151",  "LineBreakTest.txt", u"\u002d\u0308\u0e01"},
-
-        // Issue ICU-12017 Improve line break around numbers
-        {"12017", "LineBreakTest.txt", u"\u002C\u0030"},   // ",0"
-        {"12017", "LineBreakTest.txt", u"\u002C\u0308\u0030"},
-        {"12017", "LineBreakTest.txt", u"find .com"},
```
FrankYFTang, line 1344: is this removed accidentally?

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -1326,34 +1326,37 @@ UBool RBBITest::testCaseIsKnownIssue(const UnicodeString &testCase, const char *
...
-        {"12017", "LineBreakTest.txt", u"a.2\u3000\u0033"},
-        {"12017", "LineBreakTest.txt", u"A.1 \uBABB"},
-        {"12017", "LineBreakTest.txt", u"\uBD24\uC5B4\u002E\u0020\u0041\u002E\u0032\u0020\uBCFC"},
-        {"12017", "LineBreakTest.txt", u"\uBD10\uC694\u002E\u0020\u0041\u002E\u0033\u0020\uBABB"},
-        {"12017", "LineBreakTest.txt", u"\uC694\u002E\u0020\u0041\u002E\u0034\u0020\uBABB"},
-        {"12017", "LineBreakTest.txt", u"a.2\u3000\u300C"},
+        // However, that ticket has been closed as fixed but these tests still fail, so
+        // ICU-21097 has been created to investigate and address these remaining issues.
+        {"21097",  "LineBreakTest.txt", u"-#"},
+        {"21097",  "LineBreakTest.txt", u"\u002d\u0308\u0023"},
```
FrankYFTang, line 1334: [optional] since we are using u"" here, should we just put the raw text as UTF8 w/o \u encoding here instead?

icu4c/source/test/intltest/rbbitst.cpp
pedberg-icu: No, on purpose. See the description at the top: "remove one logKnownIssue that does not produce a test failure."

icu4c/source/test/intltest/rbbitst.cpp
pedberg-icu: In this case the object is to match lines in LineBreakTest.txt that are specified in hex, for example
```
× 0023 × 0308 × 002D ÷	#  × [0.3] NUMBER SIGN (AL) × [9.0] COMBINING ...
```
so I think it is more useful to have the hex for comparison with that file.

icu4c/source/test/intltest/rbbitst.cpp
pedberg-icu: My initial goal was to remove logKnownIssue test skips that were no longer relevant. MOst of the logKnownIssue test skips here are still relevant, they just needed to refer to a new still-open ticket. But this test no longer needs to be skipped.

PR #1135, opened 2020-04-27
ICU-21099 udat_toCalendarDateField, handle all UDateFormatFields and out of range

pedberg-icu:
Fix udat_toCalendarDateField to handle all defined UDateFormatField values, and to not read past array bounds for unknown inputs

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21099
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/udat.cpp
+++ b/icu4c/source/i18n/udat.cpp
@@ -82,13 +82,18 @@ static UCalendarDateFields gDateFieldMapping[] = {
     UCAL_ZONE_OFFSET,          // UDAT_TIMEZONE_ISO_FIELD = 32 (also UCAL_DST_OFFSET)
     UCAL_ZONE_OFFSET,          // UDAT_TIMEZONE_ISO_LOCAL_FIELD = 33 (also UCAL_DST_OFFSET)
     UCAL_EXTENDED_YEAR,        // UDAT_RELATED_YEAR_FIELD = 34 (not an exact match)
-    UCAL_FIELD_COUNT,          // UDAT_FIELD_COUNT = 35
+    UCAL_FIELD_COUNT,          // UDAT_AM_PM_MIDNIGHT_NOON_FIELD=35 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_FLEXIBLE_DAY_PERIOD_FIELD=36 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_TIME_SEPARATOR_FIELD = 37 (no match)
+                               // UDAT_FIELD_COUNT = 38 as of ICU 67
     // UCAL_IS_LEAP_MONTH is not the target of a mapping
 };
 
 U_CAPI UCalendarDateFields U_EXPORT2
 udat_toCalendarDateField(UDateFormatField field) {
-  return gDateFieldMapping[field];
+  static_assert(UDAT_FIELD_COUNT == UPRV_LENGTHOF(gDateFieldMapping),
+    "UDateFormatField and gDateFieldMapping should have the same number of entries and be kept in sync.");
+  return (field >= UDAT_ERA_FIELD && field < UPRV_LENGTHOF(gDateFieldMapping))? gDateFieldMapping[field]: UCAL_FIELD_COUNT;
 }
 
 /* For now- one opener. */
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/udat.cpp
```diff
@@ -82,13 +82,16 @@ static UCalendarDateFields gDateFieldMapping[] = {
...
-    UCAL_FIELD_COUNT,          // UDAT_FIELD_COUNT = 35
+    UCAL_FIELD_COUNT,          // UDAT_AM_PM_MIDNIGHT_NOON_FIELD=35 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_FLEXIBLE_DAY_PERIOD_FIELD=36 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_TIME_SEPARATOR_FIELD = 37 (no match)
+                               // UDAT_FIELD_COUNT = 38 as of ICU 67
     // UCAL_IS_LEAP_MONTH is not the target of a mapping
 };
 
 U_CAPI UCalendarDateFields U_EXPORT2
 udat_toCalendarDateField(UDateFormatField field) {
```
jefgen, line 93: It's a bit unfortunate that this API doesn't have an error code parameter to indicate failure... Looking at the API docs it says that it will return back `UCAL_FIELD_COUNT` if an error occurs. However, this doesn't work for binary compat across different versions of ICU, as a new version might start using the previous value of UCAL_FIELD_COUNT for something legitimate. 😕
(Though this fix of checking for the range is still much better than before though!)

markusicu: Right, not good. With the range check, we could modify the docs to suggest calling the function with a bogus value (cast -1 to the enum) and capture the return value as the error value.

icu4c/source/i18n/udat.cpp
```diff
@@ -82,13 +82,16 @@ static UCalendarDateFields gDateFieldMapping[] = {
     UCAL_ZONE_OFFSET,          // UDAT_TIMEZONE_ISO_FIELD = 32 (also UCAL_DST_OFFSET)
     UCAL_ZONE_OFFSET,          // UDAT_TIMEZONE_ISO_LOCAL_FIELD = 33 (also UCAL_DST_OFFSET)
     UCAL_EXTENDED_YEAR,        // UDAT_RELATED_YEAR_FIELD = 34 (not an exact match)
-    UCAL_FIELD_COUNT,          // UDAT_FIELD_COUNT = 35
+    UCAL_FIELD_COUNT,          // UDAT_AM_PM_MIDNIGHT_NOON_FIELD=35 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_FLEXIBLE_DAY_PERIOD_FIELD=36 (no match)
+    UCAL_FIELD_COUNT,          // UDAT_TIME_SEPARATOR_FIELD = 37 (no match)
+                               // UDAT_FIELD_COUNT = 38 as of ICU 67
     // UCAL_IS_LEAP_MONTH is not the target of a mapping
 };
```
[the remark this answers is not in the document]
pedberg-icu: I like that very much, I will add that

icu4c/source/i18n/udat.cpp
pedberg-icu: Rather than have the range check use UDAT_FIELD_COUNT (which can vary release to release and will result in problems if it UDAT_FIELD_COUNT in increased without adding array entries), it should use the highest specific enum value actually in the array. I will make that change.

markusicu: 1. I prefer the range check against `UPRV_LENGTHOF(gDateFieldMapping)` because it reduces the number of lines that someone needs to update in sync when there is a new UDateFormatField.
2. Jeff's concern here is that a caller checking for a constant UCAL_FIELD_COUNT is brittle, hence my suggestion to change the API doc. Note that these "count" constants are `@deprecated ICU 58 The numeric value may change over time, see ICU ticket #12420.`

pedberg-icu: Ah yes, good points, will update

markusicu:
@jefgen are most of the Azure bots down?

jefgen:
> are most of the Azure bots down?

It looks like all of the Azure built bots actually [passed](https://unicode-icu.visualstudio.com/ICU/_build/results?buildId=7969&view=results), but the status on GitHub isn't reporting/reflecting that. 
It looks like there is a known/reported GitHub [outage/issue](https://www.githubstatus.com/incidents/cq3jvp8xp8dw) as of 2020-04-28 16:44 UTC, so I think this might be the why.

jefgen:
Thank you @pedberg !

PR #1137, opened 2020-04-27
ICU-21109 Add support for locale-dependent minimum grouping digits to DecimalFormat

sffc:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21109
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/number_grouping.cpp
+++ b/icu4c/source/i18n/number_grouping.cpp
@@ -64,6 +64,13 @@ Grouper Grouper::forProperties(const DecimalFormatProperties& properties) {
 }
 
 void Grouper::setLocaleData(const impl::ParsedPatternInfo &patternInfo, const Locale& locale) {
+    if (fMinGrouping == -2) {
+        fMinGrouping = getMinGroupingForLocale(locale);
+    } else if (fMinGrouping == -3) {
+        fMinGrouping = static_cast<int16_t>(uprv_max(2, getMinGroupingForLocale(locale)));
+    } else {
+        // leave fMinGrouping alone
+    }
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/number_grouping.cpp
```diff
@@ -64,6 +64,13 @@ Grouper Grouper::forProperties(const DecimalFormatProperties& properties) {
 }
 
 void Grouper::setLocaleData(const impl::ParsedPatternInfo &patternInfo, const Locale& locale) {
+    if (fMinGrouping == -2) {
```
FrankYFTang, line 67: I have hard time to understand how the code jump from DecimalFormat::setMinimumGroupingDigits to here. Could you explain the relationship between  DecimalFormat and Grouper? 
Is there a design doc I can see to understand the internal relationship between these classes/objects?

sffc: Grouper is for NumberFormatter, but also indirectly for DecimalFormat.  The DecimalFormat properties gets turned into NumberFormatter properties via number_mapper.cpp.

sffc: Note that -2 and -3 are already supported internally in NumberFormatter.  I'm just exposing them to DecimalFormat.

I had to move these 7 lines of code to the top of the function in order for them to get evaluated even when fGrouping1 and fGrouping2 are set, as is the case in the DecimalFormat code path.

FrankYFTang:
This is obsoleted by https://github.com/unicode-org/icu/pull/1152

PR #1138, opened 2020-04-27
ICU-21104 Add new build bot: Ubuntu 18.04 with C++14

jefgen:
Note: The Ubuntu 18.04 image doesn't have HarfBuzz, so we need to disable building the layout engine.

(Related PR: #1132).

Also updating/fixing the Ubuntu image name to use the preferred VM Image name.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21104
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -6,7 +6,7 @@ jobs:
   displayName: 'J: Linux OpenJDK (Ubuntu 16.04)'
   timeoutInMinutes: 20
   pool:
-    vmImage: 'Ubuntu 16.04'
+    vmImage: 'ubuntu-16.04'
     demands: ant
   steps:
     - checkout: self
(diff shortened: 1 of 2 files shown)


sffc:
Looks like the C++14 bot is passing.  Can you try to enable `-Winvalid-constexpr` to reproduce the issue purportedly fixed by #1132?

jefgen:
> Looks like the C++14 bot is passing. Can you try to enable `-Winvalid-constexpr` to reproduce the issue

FWIW, I don't see any warnings produced in the Clang build with C++14 enabled and `-Winvalid-constexpr`  set. Maybe it is specific to GCC? I'll try adding a GCC build on Travis.

jefgen:
> > Looks like the C++14 bot is passing. Can you try to enable `-Winvalid-constexpr` to reproduce the issue
> 
> FWIW, I don't see any warnings produced in the Clang build with C++14 enabled and `-Winvalid-constexpr` set. Maybe it is specific to GCC? I'll try adding a GCC build on Travis.

Actually, it seems that GCC doesn't [support](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) the flag `-Winvalid-constexpr`, which is why the Travis build was failing.

PR #1140, opened 2020-04-29
ICU-13786 Fix addLikelySubtags/minimizeSubtags

FrankYFTang:
1. Add test cases in C++ from Java.
2. Add test cases in C++ and Java from examples of UTS35 and
   https://github.com/tc39/test262/blob/master/test/intl402/Locale/prototype/minimize/removing-likely-subtags-first-adds-likely-subtags.js
3. Fix C++ code:
   a) Make _ulocimp_addLikelySubtags propagate matching error so
   minimizeSubtags can just return the no matching case unchanged.
   b) rename the sink variable to outSink and local sink variables to
   avoid confusion and mistake
   c) Correct the implementation of minimizeSubtags by using the lang,
   region, script of the "max" instead of the localeID in the algorithm.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-13786
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -826,7 +826,7 @@ error:
     } \
 } UPRV_BLOCK_MACRO_END
 
-static void
+static UBool
 _uloc_addLikelySubtags(const char* localeID,
                        icu::ByteSink& sink,
                        UErrorCode* err) {
@@ -897,14 +897,21 @@ _uloc_addLikelySubtags(const char* localeID,
         sink.Append(localeID, localIDLength);
     }
 
-    return;
+    return success;
 
 error:
 
     if (!U_FAILURE(*err)) {
         *err = U_ILLEGAL_ARGUMENT_ERROR;
     }
+    return FALSE;
 }
+
+// Add likely subtags to the sink
+// return true if the value in the sink is produced by a match during the lookup
+// return false if the value in the sink is the same as input because there are
+// no match after the lookup.
+static UBool _ulocimp_addLikelySubtags(const char*, icu::ByteSink&, UErrorCode*);
 
 static void
 _uloc_minimizeSubtags(const char* localeID,
@@ -921,6 +928,7 @@ _uloc_minimizeSubtags(const char* localeID,
     const char* trailing = "";
     int32_t trailingLength = 0;
     int32_t trailingIndex = 0;
+    UBool successGetMax = FALSE;
 
     if(U_FAILURE(*err)) {
         goto error;
@@ -961,7 +969,7 @@ _uloc_minimizeSubtags(const char* localeID,
     {
         icu::CharString base;
         {
-            icu::CharStringByteSink sink(&base);
+            icu::CharStringByteSink baseSink(&base);
             createTagString(
                 lang,
                 langLength,
@@ -971,7 +979,7 @@ _uloc_minimizeSubtags(const char* localeID,
                 regionLength,
                 NULL,
                 0,
-                sink,
+                baseSink,
                 err);
         }
 
@@ -980,11 +988,38 @@ _uloc_minimizeSubtags(const char* localeID,
          * from AddLikelySubtags.
          **/
         {
-            icu::CharStringByteSink sink(&maximizedTagBuffer);
-            ulocimp_addLikelySubtags(base.data(), sink, err);
+            icu::CharStringByteSink maxSink(&maximizedTagBuffer);
+            successGetMax = _ulocimp_addLikelySubtags(base.data(), maxSink, err);
         }
     }
 
+    if(U_FAILURE(*err)) {
+        goto error;
+    }
+
+    if (!successGetMax) {
+        /**
+         * If we got here, return the locale ID parameter unchanged.
+         **/
+        const int32_t localeIDLength = (int32_t)uprv_strlen(localeID);
+        sink.Append(localeID, localeIDLength);
+        return;
+    }
+
+    // In the following, the lang, script, region are referring to those in
+    // the maximizedTagBuffer, not the one in the localeID.
+    langLength = sizeof(lang);
+    scriptLength = sizeof(script);
+    regionLength = sizeof(region);
+    parseTagString(
+        maximizedTagBuffer.data(),
+        lang,
+        &langLength,
+        script,
+        &scriptLength,
+        region,
+        &regionLength,
+        err);
     if(U_FAILURE(*err)) {
         goto error;
     }
@@ -995,7 +1030,7 @@ _uloc_minimizeSubtags(const char* localeID,
     {
         icu::CharString tagBuffer;
         {
-            icu::CharStringByteSink sink(&tagBuffer);
+            icu::CharStringByteSink tagSink(&tagBuffer);
             createLikelySubtagsString(
                 lang,
                 langLength,
@@ -1005,14 +1040,15 @@ _uloc_minimizeSubtags(const char* localeID,
                 0,
                 NULL,
                 0,
-                sink,
+                tagSink,
                 err);
         }
 
         if(U_FAILURE(*err)) {
             goto error;
         }
-        else if (!tagBuffer.isEmpty() && uprv_strnicmp(
+        else if (!tagBuffer.isEmpty() &&
+                 uprv_strnicmp(
                     maximizedTagBuffer.data(),
                     tagBuffer.data(),
                     tagBuffer.length()) == 0) {
@@ -1039,7 +1075,7 @@ _uloc_minimizeSubtags(const char* localeID,
 
         icu::CharString tagBuffer;
         {
-            icu::CharStringByteSink sink(&tagBuffer);
+            icu::CharStringByteSink tagSink(&tagBuffer);
             createLikelySubtagsString(
                 lang,
                 langLength,
@@ -1049,14 +1085,15 @@ _uloc_minimizeSubtags(const char* localeID,
                 regionLength,
                 NULL,
                 0,
-                sink,
+                tagSink,
                 err);
         }
 
         if(U_FAILURE(*err)) {
             goto error;
         }
-        else if (uprv_strnicmp(
+        else if (!tagBuffer.isEmpty() &&
+                 uprv_strnicmp(
                     maximizedTagBuffer.data(),
                     tagBuffer.data(),
                     tagBuffer.length()) == 0) {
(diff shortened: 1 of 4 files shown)


icu4c/source/common/loclikely.cpp
```diff
@@ -826,7 +826,7 @@ createLikelySubtagsString(
     } \
 } UPRV_BLOCK_MACRO_END
 
-static void
+static UBool
```
roubert, line 829: I find it confusing to introduce parallel error reporting paths, both the existing `UErrorCode` output parameter and the new `UBool` return value. Why not continue reporting all errors through `UErrorCode`?

icu4c/source/common/loclikely.cpp
```diff
@@ -897,18 +897,21 @@ _uloc_addLikelySubtags(const char* localeID,
...
     }
+    return FALSE;
 }
 
+static UBool _ulocimp_addLikelySubtags(const char*, icu::ByteSink&, UErrorCode*);
+
 static void
 _uloc_minimizeSubtags(const char* localeID,
-                      icu::ByteSink& sink,
+                      icu::ByteSink& outSink,
```
roubert, line 914: I don't understand the purpose of renaming this reference here.

The actual object is allocated on the stack in `uloc_minimizeSubtags()` and is there called `sink`. Then a reference to that object is passed to `ulocimp_minimizeSubtags()` where it still is called `sink`. Then a reference is passed to `_uloc_minimizeSubtags()` where it now suddenly is called `outSink`.

That's confusing to me. I'd find it clearer if the same object continued to be referred to by the same name.

(And regarding that name, the only possible purpose of any Sink object is to provide output, every Sink object is an "out sink", there can be no other kind, so I can't see how adding "out" to the name would do anything but make the name longer.)

icu4c/source/common/loclikely.cpp
```diff
@@ -921,11 +924,12 @@ _uloc_minimizeSubtags(const char* localeID,
     const char* trailing = "";
     int32_t trailingLength = 0;
     int32_t trailingIndex = 0;
+    UBool successGetMax = FALSE;
 
     if(U_FAILURE(*err)) {
         goto error;
     }
-    else if (localeID == NULL) {
+    else if (localeID == nullptr) {
```
roubert, line 932: Changing `NULL` to `nullptr` here would introduce an inconsistency in this file.

At the moment, `NULL` is used consistently throughout the entire file. Plain and simple.

Your proposed change would make this file instead use 20% `nullptr` and 80% `NULL`. I don't see how that would be an improvement and it has absolutely nothing to do with what else this PR is about.

If you want to change `NULL` to `nullptr`, I suggest that you instead do that all at once as a separate change. That'll result in files with consistent usage and an easier to follow change history.

icu4c/source/common/loclikely.cpp
```diff
@@ -961,17 +965,17 @@ _uloc_minimizeSubtags(const char* localeID,
     {
         icu::CharString base;
         {
-            icu::CharStringByteSink sink(&base);
+            icu::CharStringByteSink baseSink(&base);
```
roubert, line 968: See previous comment about consistenly using the same name to refer to the same object.

If I understand your PR description correctly, you think that it might cause confustion that it in two different scopes within this function exist two different objects that both are named `sink`.

Is that really a problem? We have lots and lots for for-loops that all name their loop variable `i`, even though they each have their own separate locally scoped loop variable. That doesn't seem to cause any confusion. Is the situation with the two Sink objects here really different from that?

icu4c/source/common/loclikely.cpp
```diff
@@ -1193,15 +1236,23 @@ uloc_addLikelySubtags(const char* localeID,
     return reslen;
 }
 
-U_CAPI void U_EXPORT2
-ulocimp_addLikelySubtags(const char* localeID,
-                         icu::ByteSink& sink,
-                         UErrorCode* status) {
+static UBool
```
roubert, line 1239: See previous comment about introducing parallel error reporting paths.

icu4c/source/common/loclikely.cpp
FrankYFTang: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1Locale.html#a974d212572d7fc61a0579f820100a8f6 
said
"If this Locale is already in the maximal form, or not valid, or there is no data available for maximization, the Locale will be unchanged."
Notice, the case of "there is no data available for maximization, the Locale will be unchanged." - is not a condition of "error" from this API description.

The implementation of minimizeSubtags need to call addLikelySubtags, but the algorithm in 
http://www.unicode.org/reports/tr35/#Likely_Subtags stated
3. Return
3-1. If there is no match,either return
3-1-1. an error value, or ...

Because the above quoted text, addLikelySubtags won't return any error code for this case, but just return the Locale unchanged, based on the design of that API. That is not the problem, but ....

For minimizeSubtags, it stated:
1. First get max = AddLikelySubtags(inputLocale). If an error is signaled, return it.
to correctly implement the "If an error is signaled, return it." we need to catch this "error" - which is the "no match"
Therefore, I cannot use the UErrorCode to report the same information, because if I put any error code in the lower level, the upper level codes one for the public addLikelySubtags() and one for the internal usage inside minimizeSubtags won't be able to distinguish the error. For example, if I place U_ILLEGAL_ARGUMENT_ERROR into the UErrorCode, then the upper code cannot distinguish the U_ILLEGAL_ARGUMENT_ERROR caused by some other reason from the U_ILLEGAL_ARGUMENT_ERROR caused by "no matching".



minimizeSubtags
--

icu4c/source/common/loclikely.cpp
FrankYFTang: Before my change, the input is sink, then there are several other local are also "sink" which "shadowed" this REAL sink which mean for output. I could change this name back to sink but I really think reusing the same name "sink" is very confusing and problematic while there are 5 sinks here, in particular one got SHADOWED here.

icu4c/source/common/loclikely.cpp
FrankYFTang: I have been asked to change all NULL to nullptr in one function because I touch one line of NULL and someone quote we should always use nullptr in the "new code". I can switch back but it seems now different reviewers has different standard about this.

icu4c/source/common/loclikely.cpp
FrankYFTang: I think it is very confusing. Would you write a function like tis
void Foo(int i) {
 // 50 lines of code
 for (int i = 0; ...) {
     b(i);
 }
 // 50 lines of code
 for (int i = 0; ...) {
     b(i);
 }
 // 50 lines of code
 for (int i = 0; ...) {
     b(i);
 }
 // 50 lines of code
 b(i);
}
Basically that is what happen here. The name "sink" got passed in, and then some other local sink are declare which "shadow" the sink in the parameter for 4-5 times, and then later the real "sink" got called. It is very difficult to track WHICH sink it is at the point of the last call and I think that could easily introduce bug

icu4c/source/common/loclikely.cpp
```diff
@@ -1111,22 +1145,31 @@ _uloc_minimizeSubtags(const char* localeID,
                         langLength,
                         script,
                         scriptLength,
-                        NULL,
+                        nullptr,
                         0,
                         trailing,
                         trailingLength,
-                        sink,
```
FrankYFTang, line 1118: for example, how easy for you to figure out which "sink " this sink is referring to? How easy for you to spot this "sink" is not the "sink" in line 1119, not the "sink" in line 1073, not the sink in line 993 but the "sink" passed in as input parameter.

icu4c/source/common/loclikely.cpp
FrankYFTang: The so called "parallel error reporting paths" pre-exist in createLikelySubtagsString because the API treatment of "no match" in addLikelySubtags and the algorithm for minimizeSubtags calling addLikelySubtags are different. As you can see the current code in _uloc_addLikelySubtags ALSO depend on the return UBool of createLikelySubtagsString() to perform different action and cannot depend on the error code of createLikelySubtagsString because it won't be able to figure out the differences between the  "no match" error from others.

icu4c/source/common/loclikely.cpp
FrankYFTang: "the only possible purpose of any Sink object is to provide output" 
That is not true, this sink is provide output of this FUNCTION _uloc_minimizeSubtags but later on you see there are sink create on the heap NOT for output of this FUNCTION, but rather to gather result from the function it called. I found the parameter name sink is useless, because sink is a name of a type, we won't name a parameter int integer because it is an integer, right? so why we name a icu::ByteSink sink here? Maybe we should rename it to "result" or "output" ? 

Also any function called be called in multiple places and each places could name the local variable differently so there are no reason "the same object continued to be referred to by the same name." because this outSink COULD come from multiple calling functions, right?

FrankYFTang:
PTAL I change the name of "outSink" back to "sink" but keep my other rename of "sink" of local variable as baseSink, tagSink and so on so it is clear which one is the local one and which one is the 'sink' which intend for the output. I also change the nullptr back to NULL as requested by roubert.

icu4c/source/common/loclikely.cpp
FrankYFTang: I added more comment. Notice while there are no match (therefore the return will be false) the value in the sink is always the same as the input, but having the same value of input in the sink could also come with a return value of true because the result of matching. so we cannot simply compare the input/output to get the 'match" or "no match" here. and the condition of "no match" is also defined NOT as an error by the current API.

FrankYFTang:
ping

FrankYFTang:
ping

icu4c/source/common/loclikely.cpp
markusicu: I think renaming the inner sinks is fine.

icu4c/source/common/loclikely.cpp
markusicu: So this returns FALSE when there was an error, or no error and no data. And the caller copies the input locale if this function didn't actually add likely subtags. Seems reasonable to me.

icu4c/source/common/loclikely.cpp
markusicu: Seems reasonable from what Frank says and what I see the call site doing.

PR #1141, opened 2020-05-02
ICU-20545 Detect file separator char from dir

wojnilowicz:
[ICU-20545]

If `udata_create` won't find `U_FILE_SEP_CHAR` at the end of a `dir` variable,
then it appends it. The problem starts if the `dir` variable uses
`U_FILE_ALT_SEP_CHAR` which is not equal to `U_FILE_SEP_CHAR`. Then the
resulting path could look like this
`../data\mappings/cns-11643-1992.ucm`
instead of this
`../data/mappings/cns-11643-1992.ucm`

This patch uses `U_FILE_SEP_CHAR` unless it detects that the `dir` variable
doesn't use it, and uses `U_FILE_ALT_SEP_CHAR` instead.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20545
- [x] Issue accepted
- [ ] Tests included

[ICU-20545]: https://unicode-org.atlassian.net/browse/ICU-20545

### DIFF
--- a/icu4c/source/tools/toolutil/unewdata.cpp
+++ b/icu4c/source/tools/toolutil/unewdata.cpp
@@ -56,15 +56,25 @@ udata_create(const char *dir, const char *type, const char *name,
         *pErrorCode=U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
-    
+
+    char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+    // We may need to append a different directory separator when building for Cygwin or MSYS2.
+    if(dir && *dir) {
+      if(!uprv_strchr(dir, U_FILE_SEP_CHAR) && uprv_strchr(dir, U_FILE_ALT_SEP_CHAR)) {
+          dirSepChar = U_FILE_ALT_SEP_CHAR;
+      }
+    }
+#endif
+
     /* Check that the full path won't be too long */
     length = 0;					/* Start with nothing */
     if(dir != NULL  && *dir !=0)	/* Add directory length if one was given */
     {
     	length += static_cast<int32_t>(strlen(dir));
-	
+
     	/* Add 1 if dir doesn't end with path sep */
-        if (dir[strlen(dir) - 1]!= U_FILE_SEP_CHAR) {
+        if (dir[strlen(dir) - 1]!= dirSepChar) {
             length++;
         }
 	}
@@ -74,7 +84,7 @@ udata_create(const char *dir, const char *type, const char *name,
         length += static_cast<int32_t>(strlen(type));
     }
 
-        
+
      /* LDH buffer Length error check */
     if(length  > ((int32_t)sizeof(filename) - 1))
     {
@@ -82,13 +92,13 @@ udata_create(const char *dir, const char *type, const char *name,
    	    uprv_free(pData);
 	    return NULL;
     }
-   
+
     /* open the output file */
     if(dir!=NULL && *dir!=0) { /* if dir has a value, we prepend it to the filename */
         char *p=filename+strlen(dir);
         uprv_strcpy(filename, dir);
-        if (*(p-1)!=U_FILE_SEP_CHAR) {
-            *p++=U_FILE_SEP_CHAR;
+        if (*(p-1)!=dirSepChar) {
+            *p++=dirSepChar;
             *p=0;
         }
     } else { /* otherwise, we'll output to the current dir */


markusicu:
FYI We have been modernizing some of our code to using the internal class CharString (icu4c/source/common/charstr.h) which has a method appendPathPart(). I am pretty sure that handles both separators, but unlike the proposed fix it will _write_ only the "normal" one.

jefgen:
Thank you for creating a pull-request! :)

I don't have a working Cygwin environment, so I can't actually test anything for this at the moment.

However, in the past I've noticed that Cygwin will "automatically" change path separators to either POSIX-style or Windows/DOS-style based on what seems to be a bunch of random and often unpredictable heuristics.

I'm wondering if this is a case where the Cygwin build with MSVC is confused -- as it is building ICU for Windows with MSVC, but using a Cygwin-style POSIX command line environment.

FWIW, I think the `appendPathPart()` method will only ever append the `U_FILE_SEP_CHAR`.
Whereas I think this patch is trying to dynamically append either a `U_FILE_SEP_CHAR` or `U_FILE_ALT_SEP_CHAR`, depending on if the input dir path uses one or the other.

I'm a bit leery of unconditionally changing the ICU code for all platforms with this.

However, if there is really no other way to address the Cygwin+MSVC build issues at all, then I might suggest trying to constrain this by checking for the special macro `CYGWINMSVC`.
https://github.com/unicode-org/icu/blob/20c29becd69b19c62cb6618665fd773eeb73dab1/icu4c/source/common/unicode/platform.h#L210
This macro is only set when building using MSVC in the Cygwin environment.

One other thing to note:
It also looks like the change as-is fails to compile one several of the CI build bots.

I think you might have meant to use `dir` instead of the variable `text`?

wojnilowicz:
> Thank you for creating a pull-request! :)

You're welcome. This PR does not fix issue ICU-20545 completely, but allows to go a bit further with it. After successfully merging this PR, I will try to propose a new PR.

> FWIW, I think the `appendPathPart()` method will only ever append the `U_FILE_SEP_CHAR`.
> Whereas I think this patch is trying to dynamically append either a `U_FILE_SEP_CHAR` or `U_FILE_ALT_SEP_CHAR`, depending on if the input dir path uses one or the other.

That's right.

> I'm a bit leery of unconditionally changing the ICU code for all platforms with this.
> 
> However, if there is really no other way to address the Cygwin+MSVC build issues at all, then I might suggest trying to constrain this by checking for the special macro `CYGWINMSVC`.
> https://github.com/unicode-org/icu/blob/20c29becd69b19c62cb6618665fd773eeb73dab1/icu4c/source/common/unicode/platform.h#L210

I would not guard it with CYGWINMSVC, because I'm hitting this issue with MSYS2.

With this PR I use heuristics that is already present in [makeconv.cpp](https://github.com/unicode-org/icu/blob/master/icu4c/source/tools/makeconv/makeconv.cpp#L359) for all platforms.

```
            /* Insure the static data name matches the  file name */
            /* Changed to ignore directory and only compare base name
             LDH 1/2/08*/
            char *p;
            p = strrchr(cnvName, U_FILE_SEP_CHAR); /* Find last file separator */

            if(p == NULL)            /* OK, try alternate */
            {
                p = strrchr(cnvName, U_FILE_ALT_SEP_CHAR);
                if(p == NULL)
                {
                    p=cnvName; /* If no separators, no problem */
                }
            }
            else
            {
                p++;   /* If found separator, don't include it in compare */
            }
```

> One other thing to note:
> It also looks like the change as-is fails to compile one several of the CI build bots.
> 
> I think you might have meant to use `dir` instead of the variable `text`?

Thank you. I've corrected this.
It still shows some errors. Any idea why?

icu4c/source/tools/toolutil/unewdata.cpp
```diff
@@ -56,6 +56,10 @@ udata_create(const char *dir, const char *type, const char *name,
         *pErrorCode=U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
+
+    char dirSepChar = U_FILE_SEP_CHAR;
+    if(strchr(dir, U_FILE_SEP_CHAR) == NULL && strchr(dir, U_FILE_ALT_SEP_CHAR) != NULL)
```
jefgen, line 61: >>  It also looks like the change as-is fails to compile one several of the CI build bots.
> It still shows some errors. Any idea why?

It looks like the issue is the `gendict.exe` program is crashing, which causes the build to fail.

It seems the problem is that the input `dir` variable might be `NULL` in some cases. This causes the function `strchr` to deference a `NULL` pointer, causing the crash.

I'd probably suggest moving the check below inside the if block which already checks if `dir` is not `NULL`.

> With this PR I use heuristics that is already present in makeconv.cpp for all platforms.

Interesting... it seems that there is (for better or worse) already many places through the ICU code that does similar checks/hacks. 
Based on how this is done in other places in the ICU code, I'd likely suggest doing something like this instead:

```c++
    /* Check that the full path won't be too long */
    length = 0;					/* Start with nothing */
    if(dir != NULL  && *dir !=0)	/* Add directory length if one was given */
    {
#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
        if(strchr(dir, U_FILE_SEP_CHAR) == NULL && strchr(dir, U_FILE_ALT_SEP_CHAR) != NULL)
        {
            dirSepChar = U_FILE_ALT_SEP_CHAR;
        }
#endif
```

(The use of the preprocessor macro will mean that the change only effects Windows/Cygwin platforms.)

jefgen:
One other question:
> I would not guard it with CYGWINMSVC, because I'm hitting this issue with MSYS2.

How are you building with MSYS2?

We do have a MSYS2 build bot setup here that builds: 
https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-pipelines.yml#L256

I'd be curious to know how you're building ICU with MSYS2. If it's different (I suspect it is) then perhaps we might be able to add another build bot for it (though that would likely be another ticket).

wojnilowicz:
> I'd be curious to know how you're building ICU with MSYS2. If it's different (I suspect it is) then perhaps we might be able to add another build bot for it (though that would likely be another ticket).

I attach minimal test case that fails to build
[icubuild.zip](https://github.com/unicode-org/icu/files/4626568/icubuild.zip)
In order to use it:
1) unpack to single directory
2) run mingw64.exe
3) invoke "bash build.sh"

My configuration:
1) msys2-x86_64-20190524.exe
2) Windows 7, 64 bit
3) pacman --noconfirm -S mingw-w64-x86_64-toolchain make
4) Python 3.8.2
5) Cmake  3.17.1
6) "whereis cmake" returns
cmake: /usr/share/cmake /mingw64/bin/cmake.exe
7) "whereis python3" returns
python3: /usr/bin/python3.8-config /usr/bin/python3.exe /usr/lib/python3.8 /usr/include/python3.8 /mingw64/bin/python3.8-config /mingw64/bin/python3.exe /usr/share/man/man1/python3.1.gz

wojnilowicz:
> Thanks for the PR!
> 
> I have some feedback/changes on it -- but overall I think this might be okay with the additional changes.

Thanks for finding a crash cause. I corrected my PR according to your review. Please merge in the case of no other issues.

icu4c/source/tools/toolutil/unewdata.cpp
```diff
@@ -57,14 +57,24 @@ udata_create(const char *dir, const char *type, const char *name,
         return NULL;
     }
     
+    char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+    // We may need to append a different directory separator when building for Cygwin or MSYS2.
+    if(dir != NULL  && *dir !=0) {
+      if(strchr(dir, U_FILE_SEP_CHAR) == NULL && strchr(dir, U_FILE_ALT_SEP_CHAR) != NULL) {
```
jefgen, line 64: Feedback from Markus was that we should be using the `uprv_strstr` instead of directly using `strchr`.

Additionally, he suggested that in new code `nullptr` should be used instead of `NULL` (though it looks like this file uses `NULL` everywhere, so this is more of a nit) .

jefgen:
> I attach minimal test case that fails to build

Thanks!

It looks like you're using CMake to build ICU, but it's not really clear to me how. I think it just uses the existing makefiles, but with CMake somehow?

I wonder if that is different enough to cause the MSYS2/Cygwin auto-magic path heuristics to get confused, leading to the paths being different.

> Thanks for finding a crash cause. I corrected my PR according to your review. Please merge in the case of no other issues.

Markus had some feedback/suggestions here: https://github.com/unicode-org/icu/pull/1141#discussion_r428193130
I think once those could be addressed then this looks good and can be merged!

srl295:
There are existing old tickets and branches proposing a move to CMake, it stalled due to lack of interest and feedback.

wojnilowicz:
> It looks like you're using CMake to build ICU, but it's not really clear to me how. I think it just uses the existing makefiles, but with CMake somehow?

I think, you should study the case. It goes from zero to a built library (or in the current state to a build failure)
For me it's the case of bootstraping an environment with libraries, so that they can be packaged latter, and CMake is used here to do it, regardless of the build system you (or any other library) use.

wojnilowicz:
Thanks for the review. Please review again, and merge in the case of no more issues.

icu4c/source/tools/toolutil/unewdata.cpp
```diff
@@ -56,15 +56,25 @@ udata_create(const char *dir, const char *type, const char *name,
         *pErrorCode=U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
-    
+
+    char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+    // We may need to append a different directory separator when building for Cygwin or MSYS2.
+    if(dir && *dir) {
+      if(!uprv_strchr(dir, U_FILE_SEP_CHAR) && uprv_strchr(dir, U_FILE_ALT_SEP_CHAR)) {
```
jefgen, line 64: Note: There might be a preference to be explicit and check for `!= nullptr` and `== nullptr` here.

However, the ICU code is somewhat [inconsistent](https://github.com/unicode-org/icu/blob/master/icu4c/source/common/locid.cpp#L402) about this, so I don't have strong opinion on it

PR #1143, opened 2020-05-11
ICU-20963 Add .clang-format file to icu4c/source

younies:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20963
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/.clang-format
+++ b/icu4c/source/.clang-format
@@ -1,0 +1,11 @@
+# © 2020 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+---
+Language: Cpp
+BasedOnStyle: LLVM
+IndentWidth: 4
+ColumnLimit: 105
+AllowShortBlocksOnASingleLine: false 
+AllowShortIfStatementsOnASingleLine: true 
+...


hugovdm:
I've been using these clang-format settings too for a while, so have some experience with how it compares to typical ICU style. I do manual formatting (via vscode). We're not recommending any automatic reformatting, this is mostly for use for new code.

Question: do we want to document this somewhere? E.g. in https://github.com/unicode-org/icu/blob/userguide-migration/docs/userguide/dev/codingguidelines.md ?

hugovdm:
I see there's a "git-clang-format" which could be used to format one's changed code only. I suggest we do update documentation, but in a separate pull request - then I can experiment a little bit more, write up what I've discovered, and "socialize" the suggestions and doc update? :)

younies:
> Please fix the lint error (add a copyright notice to this file)

Done!

icu4c/source/.clang-format
```diff
@@ -0,0 +1,11 @@
+# © 2020 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+---
+Language: Cpp
+BasedOnStyle: LLVM
+IndentWidth: 4
+ColumnLimit: 105
+AllowShortBlocksOnASingleLine: false 
+AllowShortIfStatementsOnASingleLine: true
```
hugovdm, line 10: I propose adding this:
AllowShortFunctionsOnASingleLine: false

younies: Okay, take your time :)

hugovdm: The code has lots of examples of one-line {} function bodies (I didn't try to compare to others), so I withdraw my request to add this here. :) LGTM.

sffc:
@markusicu ?

hugovdm:
@markusicu - taken from https://github.com/unicode-org/icu/pull/1250 :

LLVM is the style we're currently using, I don't know how this was chosen. I notice LLVM, Microsoft and GNU have `UErrorCode &status`, whereas the other styles have `UErrorCode& status`. A quick grep survey shows both styles are present in ICU, with significant numbers, though "& " is more popular than " &":

```
$ grep "UErrorCode& status" */*.h | wc -l
448
$ grep "UErrorCode& status" */*.cpp | wc -l
1034
$ grep "UErrorCode &status" */*.h | wc -l
230
$ grep "UErrorCode &status" */*.cpp | wc -l
661
```

Here are examples of clang-format output for a handful of styles - all of them place the ":" at the beginning of the next line. We're using "AllowShortBlocksOnASingleLine: false" because clang-format can't distinguish between just "return" and more interesting statements, and the latter we don't want to have on just one line. Samples:

BasedOnStyle: LLVM

    CompactUnicodeString(const UnicodeString &text, UErrorCode &status)
        : fBuffer(text.length() + 1, status) {
        if (U_FAILURE(status)) {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

BasedOnStyle: Google as well as Chromium

    CompactUnicodeString(const UnicodeString& text, UErrorCode& status)
        : fBuffer(text.length() + 1, status) {
        if (U_FAILURE(status)) {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

BasedOnStyle: Mozilla

    CompactUnicodeString(const UnicodeString& text, UErrorCode& status)
      : fBuffer(text.length() + 1, status)
    {
        if (U_FAILURE(status)) {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

BasedOnStyle: WebKit

    CompactUnicodeString(const UnicodeString& text, UErrorCode& status)
        : fBuffer(text.length() + 1, status)
    {
        if (U_FAILURE(status)) {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

BasedOnStyle: Microsoft

    CompactUnicodeString(const UnicodeString &text, UErrorCode &status)
        : fBuffer(text.length() + 1, status)
    {
        if (U_FAILURE(status))
        {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

BasedOnStyle: GNU

    CompactUnicodeString (const UnicodeString &text, UErrorCode &status)
        : fBuffer (text.length () + 1, status)
    {
        if (U_FAILURE (status))
            {
                return;
            }
        uprv_memcpy (fBuffer.getAlias (), text.getBuffer (), sizeof (UChar) * text.length ());
        fBuffer[text.length ()] = 0;
    }

Looking at https://clang.llvm.org/docs/ClangFormatStyleOptions.html I see there's a "BreakConstructorInitializers" option. Specifying "BreakConstructorInitializers: AfterColon" gets the colon on the previous line, but the indents are still undesireable - Google style:

    CompactUnicodeString(const UnicodeString& text, UErrorCode& status) :
        fBuffer(text.length() + 1, status) {
        if (U_FAILURE(status)) {
            return;
        }
        uprv_memcpy(fBuffer.getAlias(), text.getBuffer(), sizeof(UChar) * text.length());
        fBuffer[text.length()] = 0;
    }

*My proposal*: LLVM, Google, or Chromium, leaving the ":" on the beginning of the next line, but perhaps adjusting the indent? (Considering the default indent for multiple styles has the ":" initializer string on the same indentation level as the rest of the code, so it's the presence of the ": " that helps distinguish it from the function body.)

sffc:
On `UErrorCode &status` versus `UErrorCode& status`, I've always preferred the `&` attached to the type instead of the variable name, because I think of the type of the variable being a reference.  @aheninger preferred it attached to the variable, because it more closely to the semantics of the language.  We're definitely not consistent.

markusicu:
> On `UErrorCode &status` versus `UErrorCode& status`, I've always preferred the `&` attached to the type instead of the variable name, because I think of the type of the variable being a reference. @aheninger preferred it attached to the variable, because it more closely to the semantics of the language. We're definitely not consistent.

We haven't cared to be very consistent so far.
Like Andy, I prefer the asterisk and ampersand attached to the variable because that's how the language works.
```
int *a, b, *c[3][5];
```
If you attached the first `*` to the "int" you might be tempted to think that both a and b are pointers. b isn't. c illustrates that, to read the type of the variable, you have to read attached punctuation from the inside out. My first C book (something like "C for Unix programmers" from the 80s) had an annotated variable declaration with circle arrows showing the inside-out reading order.

Therefore, it makes no sense to me to attach select parts of that to the leaf type name.

We tend to paper over this a bit nowadays by requiring each variable declaration on its own line, so we have to think less about it. But since you have to put the array brackets after the variable, and we always keep them adjacent, it's still inconsistent to attach (some of?) the front matter to the leaf type name.

I suspect that I am fighting windmills on this, and I haven't really "fought" -- I haven't asked people to change this stuff when reviewing code, except when new code was grossly out of sync with old code in the same file.

Also, Hugo's grep lines were a bit narrow, looking only at references and only where it was "status" not "errorCode" or "pErrorCode". If you broaden it a bit, you see a small majority of "classic" style. From the root of icu4c:
```
$ egrep 'UErrorCode[*&] ' --include='*.h' -r | wc -l
1607
$ egrep 'UErrorCode[*&] ' --include='*.cpp' -r | wc -l
1737

$ egrep 'UErrorCode [*&]' --include='*.h' -r | wc -l
1945
$ egrep 'UErrorCode [*&]' --include='*.cpp' -r | wc -l
2577
```

markusicu:
> Indenting colon+2 is common after the colon, as I'd expect - but there are a couple of other interesting solutions too, including three spaces after the colon, and moving the comma to the next line to align with the colon... :-)

Yeah... I explained elsewhere why I have started to put the colon at the end of the constructor head line:
https://github.com/unicode-org/icu/pull/1250#discussion_r474873988

Actually, let me repeat it here for an easier to read record:

I can live with the colon on the second line.

I don't prefer it that way because when you have to wrap the initializer list, you get a conflict on the following line between wanting to line up the initializers at colon+2 vs. standard indent of 4. This is not a conflict in Google code style because the standard indent there is 2. In ICU it looks jarring.

It also feels like the colon is more like the open curly brace which we put at the end of the function head/if/loop... which I guess started to make code look more like in Pascal, with fewer lines that have nothing but curly braces.

Finally, in Python the colon is at the end of the if line etc.

markusicu:
Braces: When I first learned C I put them like Microsoft, but I liked the LLVM etc. style better because it's more compact and makes it look more like Pascal or Modula-2 which are good things.

sffc:
> And I am not using an IDE yet for ICU where I could invoke clang-format.

You can invoke it from the command line, too.

hugovdm:
Useful, formatting only touched code:

`$ git clang-format`

Or with more frills:

`$ git clang-format -p --commit <commit>`

Does formatting only for changed code. Changed since most recent commit, or changed since <commit> with `--commit`. -p makes it interactive: selectively choosing which of its suggestions to apply. So `-p` is great. :)

I've been experimenting with an informational git presubmit hook, which can be rather confusing until one is used to it: in `~/git/icu/.git/hooks/pre-commit` I simply have "git clang-format --diff -v". (I've seen interactive hooks, but they've got some problems, depending on how you use git.) As a pre-commit, the commit simply goes ahead after this has run successfully, hence "just informational, essentially post-hoc". Which means if I see some formatting that I missed, I might then do a `git clang-format -p --commit HEAD^` or similar, then `git commit --amend` to fold it in. (Generally only amending commits when I haven't yet pushed.)

PR #1148, opened 2020-05-16
ICU-20545 Ensure that path ends with detected file separator

wojnilowicz:
`CharString`, when asked, appends `U_FILE_SEP_CHAR` at the end of the string
it holds, if it won't find `U_FILE_SEP_CHAR` or `U_FILE_ALT_SEP_CHAR` there.
The problem starts if the `dir` variable uses
`U_FILE_ALT_SEP_CHAR` which is not equal to `U_FILE_SEP_CHAR`. Then the
resulting path could look like this
`../data\`
instead of this
`../data/`

This patch uses `U_FILE_SEP_CHAR` unless it detects that the `dir` variable
doesn't use it, and uses `U_FILE_ALT_SEP_CHAR` instead.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20545
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/charstr.cpp
+++ b/icu4c/source/common/charstr.cpp
@@ -197,7 +197,7 @@ CharString &CharString::appendPathPart(StringPiece s, UErrorCode &errorCode) {
     }
     char c;
     if(len>0 && (c=buffer[len-1])!=U_FILE_SEP_CHAR && c!=U_FILE_ALT_SEP_CHAR) {
-        append(U_FILE_SEP_CHAR, errorCode);
+        append(getDirSepChar(), errorCode);
     }
     append(s, errorCode);
     return *this;
@@ -207,9 +207,19 @@ CharString &CharString::ensureEndsWithFileSeparator(UErrorCode &errorCode) {
     char c;
     if(U_SUCCESS(errorCode) && len>0 &&
             (c=buffer[len-1])!=U_FILE_SEP_CHAR && c!=U_FILE_ALT_SEP_CHAR) {
-        append(U_FILE_SEP_CHAR, errorCode);
+        append(getDirSepChar(), errorCode);
     }
     return *this;
+}
+
+char CharString::getDirSepChar() const {
+    char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+    // We may need to return a different directory separator when building for Cygwin or MSYS2.
+    if(len>0 && !uprv_strchr(data(), U_FILE_SEP_CHAR) && uprv_strchr(data(), U_FILE_ALT_SEP_CHAR))
+        dirSepChar = U_FILE_ALT_SEP_CHAR;
+#endif
+    return dirSepChar;
 }
 
 U_NAMESPACE_END
--- a/icu4c/source/common/charstr.h
+++ b/icu4c/source/common/charstr.h
@@ -141,13 +141,13 @@ public:
 
     /**
      * Appends a filename/path part, e.g., a directory name.
-     * First appends a U_FILE_SEP_CHAR if necessary.
+     * First appends a U_FILE_SEP_CHAR or U_FILE_ALT_SEP_CHAR if necessary.
      * Does nothing if s is empty.
      */
     CharString &appendPathPart(StringPiece s, UErrorCode &errorCode);
 
     /**
-     * Appends a U_FILE_SEP_CHAR if this string is not empty
+     * Appends a U_FILE_SEP_CHAR or U_FILE_ALT_SEP_CHAR if this string is not empty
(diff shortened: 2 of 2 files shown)


jefgen:
Thanks for creating this PR!

I think Markus had some feedback that was similar to the feedback on the PR #1141, about using `uprv_strchr` and `nullptr`.

But otherwise, this looks good in principle.

icu4c/source/common/charstr.cpp
```diff
@@ -197,7 +197,14 @@ CharString &CharString::appendPathPart(StringPiece s, UErrorCode &errorCode) {
     }
     char c;
     if(len>0 && (c=buffer[len-1])!=U_FILE_SEP_CHAR && c!=U_FILE_ALT_SEP_CHAR) {
-        append(U_FILE_SEP_CHAR, errorCode);
+        char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+        // We may need to append a different directory separator when building for Cygwin or MSYS2.
+        if(strchr(data(), U_FILE_SEP_CHAR) == NULL && strchr(data(), U_FILE_ALT_SEP_CHAR) != NULL) {
```
markusicu, line 203: This change is ok in principle.

1. Internally we don't use strchr() and such directly but instead use wrappers with a uprv_ prefix, like uprv_strchr().
2. However, we don't forbid embedded NULs, and we have the length, so we should use uprv_memchr() with pointer+length.
3. Please use nullptr instead of NULL.

icu4c/source/common/charstr.cpp
```diff
@@ -197,7 +197,14 @@ CharString &CharString::appendPathPart(StringPiece s, UErrorCode &errorCode) {
     }
     char c;
     if(len>0 && (c=buffer[len-1])!=U_FILE_SEP_CHAR && c!=U_FILE_ALT_SEP_CHAR) {
-        append(U_FILE_SEP_CHAR, errorCode);
+        char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
```
markusicu, line 201: optional: i would remove the parentheses

icu4c/source/common/charstr.cpp
```diff
@@ -207,7 +214,14 @@ CharString &CharString::ensureEndsWithFileSeparator(UErrorCode &errorCode) {
     char c;
     if(U_SUCCESS(errorCode) && len>0 &&
             (c=buffer[len-1])!=U_FILE_SEP_CHAR && c!=U_FILE_ALT_SEP_CHAR) {
-        append(U_FILE_SEP_CHAR, errorCode);
+        char dirSepChar = U_FILE_SEP_CHAR;
```
markusicu, line 217: optional: maybe consider making a private helper function like `char getDirSepChar() const` ?

icu4c/source/common/charstr.cpp
wojnilowicz: ad 2. I think, that's then an invalid path, and it doesn't matter if we return U_FILE_SEP_CHAR or U_FILE_ALT_SEP_CHAR.

wojnilowicz:
Thanks for the review. Please review again, and merge in the case of no more issues.

icu4c/source/common/charstr.cpp
```diff
@@ -207,9 +207,19 @@ CharString &CharString::ensureEndsWithFileSeparator(UErrorCode &errorCode) {
...
+        append(getDirSepChar(), errorCode);
     }
     return *this;
 }
 
+char CharString::getDirSepChar() const {
+    char dirSepChar = U_FILE_SEP_CHAR;
+#if (U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)
+    // We may need to return a different directory separator when building for Cygwin or MSYS2.
+    if(len>0 && !uprv_strchr(data(), U_FILE_SEP_CHAR) && uprv_strchr(data(), U_FILE_ALT_SEP_CHAR))
```
jefgen, line 219: Note: There might be a preference to be explicit and check for `!= nullptr` and `== nullptr` here.

However, the ICU code is somewhat [inconsistent](https://github.com/unicode-org/icu/blob/master/icu4c/source/common/locid.cpp#L402) about this, so I don't have strong opinion on it.

icu4c/source/common/charstr.cpp
jefgen: FWIW, I think most of the other places that use/check for `(U_FILE_SEP_CHAR != U_FILE_ALT_SEP_CHAR)` do have parentheses though.

jefgen:
I think GitHub is blocking merging this PR due to @markusicu requesting changes.
Though I think most of the comments/feedback were addressed in the most recent update.
So perhaps @markusicu could PTAL to see if this is ready. Thanks!

PR #1149, opened 2020-05-18
Update uk.txt

IahThoth:
##### Checklist

- [ ] Issue accepted
- [ ] Tests included


FrankYFTang:
Please fix data in CLDR but not in ICU. ICU will take data flow from CLDR.

jefgen:
Thanks for creating a PR!

As noted by Frank, the data in ICU comes from the CLDR project, so this should be fixed there.
Otherwise, when ICU updates to a newer version of CLDR, this change will be overwritten.

Can you please file a bug for CLDR, following the steps here:
http://cldr.unicode.org/index/bug-reports

Thanks!

PR #1150, opened 2020-05-20
ICU-21029 LocaleMatcher: add option to turn off default locale

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21029


markusicu:
This API was approved in today's team meeting.

Mark or others, could you please review soon? I need this for some users.

I plan to ignore the one failing build bot. It's a MinGW that fails to configure ICU4C with "configure: error: no acceptable C compiler found in $PATH" @jefgen

jefgen:
> MinGW that fails to configure ICU4C with "configure: error: no acceptable C compiler found in $PATH"

I'm not sure what happened, as it looks like the MinGW install succeeded. I'll try re-running the failed bot...

jefgen:
> I'll file a ticket later, but for now I would just ignore the failed bot.

Sadly this looks like this a widespread issue in upstream MSYS2... :(
Filed this for tracking... https://unicode-org.atlassian.net/browse/ICU-21131

PR #1153, opened 2020-05-22
ICU-21032 Backport to 66.x:  ICU-20958 Prevent SEGV_MAPERR in append

srl295:
ICU-21032 backport of ICU-20958 to 66.x

See #971

(cherry picked from commit b7d08bc04a4296982fcef8b6b8a354a9e4e7afca)

### DIFF
--- a/icu4c/source/common/unistr.cpp
+++ b/icu4c/source/common/unistr.cpp
@@ -1563,7 +1563,11 @@ UnicodeString::doAppend(const UChar *srcChars, int32_t srcStart, int32_t srcLeng
   }
 
   int32_t oldLength = length();
-  int32_t newLength = oldLength + srcLength;
+  int32_t newLength;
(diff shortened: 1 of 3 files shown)


srl295:
#1063 is the same thing for 61.x

srl295:
The testcase crashes before and after applying the patch for me on a mac. I may need something with more swap.

jefgen:
> The testcase crashes before and after applying the patch for me on a mac. I may need something with more swap.

IIRC, one of the reports noted that you needed >=16GB of RAM.

srl295:
> > The testcase crashes before and after applying the patch for me on a mac. I may need something with more swap.
> 
> IIRC, one of the reports noted that you needed ~ >=10-16GB of RAM.

I thought I did…

srl295:
that would explain it… oops

PR #1157, opened 2020-05-26
ICU-21135 Fix pseudo locales to filter only non-root paths and avoid aliases.

hagbard:
Hopefully a fix to the issue of pseudo-locales generating both values and aliases to values (which is not expected by ICU resource bundle generation, even though they are distinct CLDR paths).

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21135
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
+++ b/tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
@@ -45,7 +45,7 @@ public class AlternateLocaleDataTest {
 
         FakeDataSupplier src = new FakeDataSupplier()
             .addLocaleData("xx", target, source, other)
-            .addInheritedData("xx", inherited);
+            .addLocaleData("root", inherited);
         CldrDataSupplier transformed =
             AlternateLocaleData.transform(
                 src, ImmutableMap.of(target.getPath(), source.getPath()), ImmutableTable.of());
@@ -97,10 +97,10 @@ public class AlternateLocaleDataTest {
         CldrData unresolved = transformed.getDataForLocale("xx", UNRESOLVED);
         CldrData resolved = transformed.getDataForLocale("xx", RESOLVED);
 
-        // Even though the missing target is not matched (so no change there) the source is always
-        // removed from the transformed data.
-        assertValuesUnordered(unresolved, other);
-        assertValuesUnordered(resolved, other);
+        // If there's no target the alt-path mapping is incomplete and we do nothing (this matches
+        // the old CLDR tool behaviour and reasonable but can hide inconsistencies in CLDR data).
+        assertValuesUnordered(unresolved, source, other);
+        assertValuesUnordered(resolved, source, other);
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


srl295:
^ please fix the commit message to `ICU-21135 Fix pseudo locales…` or immilar

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
```diff
@@ -97,10 +97,10 @@ public void testMissingTarget() {
         CldrData unresolved = transformed.getDataForLocale("xx", UNRESOLVED);
         CldrData resolved = transformed.getDataForLocale("xx", RESOLVED);
 
-        // Even though the missing target is not matched (so no change there) the source is always
-        // removed from the transformed data.
-        assertValuesUnordered(unresolved, other);
-        assertValuesUnordered(resolved, other);
+        // If there's no target the alt-path mapping is incomplete and we do nothing (this matches
+        // the old CLDR tool behaviour and reasonable but can hide inconsistencies in CLDR data).
+        assertValuesUnordered(unresolved, source, other);
```
hagbard, line 102: This was technically unrelated but broke in the previous merge (which effectively undid some of the alt-path fixes I made earlier). The code itself is (and was) correct, it's just this one test that got muddled up.
This reverts the comment to previous (we no longer expect a change) and gets the expectations correct by including the original source.

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/AlternateLocaleDataTest.java
```diff
@@ -45,7 +45,7 @@ public void testLocaleData() {
 
         FakeDataSupplier src = new FakeDataSupplier()
             .addLocaleData("xx", target, source, other)
-            .addInheritedData("xx", inherited);
```
hagbard, line 48: The tests are being changed because I removed the very poor API that let you specify "inherited" data directly rather than adding data to a different locale and having inheritance happen. The original API was prone to all sorts of accidental misuse since you could create very inconsistent data easily.

The "root" locale is hard-coded to be the parent of every parent-less locale, but later on you'll see places where other parent locales are set explicitly.

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/PseudoLocalesTest.java
```diff
@@ -75,12 +79,12 @@ public void testBidi() {
             .addLocaleData("en",
                 value(included, "{Hello} {0} {World} 100x"),
                 value(pattern, "'plus' HH:mm; 'minus' HH:mm"))
-            .addInheritedData("en",
-                value(inherited, "UTC"));
+            .addLocaleData("en_001", value(inherited, "UTC"))
```
hagbard, line 82: the "inherited" value appears as part of the pseudo locale data, but the "root" data doesn't (this is even true for normal values that could be pseudo-localized). One day we might revisit this and allow some root values to be pseudo-localized, but this is pretty good now (and the generated data doesn't include aliases any more).

tools/cldr/cldr-to-icu/src/test/java/org/unicode/icu/tool/cldrtoicu/mapper/LocaleMapperTest.java
```diff
@@ -65,7 +65,7 @@ public void testInheritedValuesNotIncludedByDefault() {
             ldml("units/durationUnit[@type=\"foo\"]/durationUnitPattern", "Bar"),
             simpleResult("/durationUnits/foo", "Bar"));
         //ldml/localeDisplayNames/keys/key[@type="(%A)"] ; /Keys/$1
-        addInheritedMapping("xx",
+        addRootMapping(
```
hagbard, line 68: I'm a bit unhappy with "root" here, it's only important that it's a parent locale, not the root itself.
Happy to change if you want.

PR #1161, opened 2020-05-29
ICU-21140 Fix cldr-to-icu tooling to work on Windows

daniel-ju:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21140
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
+++ b/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
@@ -100,8 +100,9 @@ public final class CleanOutputDirectoryTask extends Task {
     }
 
     @SuppressWarnings("unused")
-    public void setRoot(Path root) {
-        this.root = root;
+    public void setRoot(String root) {
+        // Use String here since on some systems Ant doesn't support automatically converting Path instances.
+        this.root = Paths.get(root);
     }
 
     @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -100,8 +100,8 @@ public void init() throws BuildException {
     }
 
     @SuppressWarnings("unused")
-    public void setRoot(Path root) {
-        this.root = root;
+    public void setRoot(String root) {
```
hagbard, line 103: I understand the charset change (sorry, that was very lax of me).

However I'm interested to know why the Path -> String changes were made, since unless I know what's going on I'm inevitably going to make the same mistake again (since I cannot test on Windows).

Can you add a one-line comment here (and in the other places where you've made changes) explaining it please?

E.g.
// We can't use Path as a parameter type because <insert reason here...>

I assume it's an Ant version thing, not a Windows thing per se, but I could be wrong.

daniel-ju: Thanks for taking a look!

Right, this is an Ant issue, although I'm unsure if it's due to a different version. Ant doesn't seem to perform conversions for `java.nio.file.Path` objects. Attempting to build results in the following error:

`build-icu-data.xml:339: outputDirectories doesn't support the "root" attribute`

See "Conversions Ant will perform for attributes" here: https://ant.apache.org/manual/develop.html

I'll push an update to add comments explaining this.

hagbard: Maybe say " ...because on some systems Ant doesn't ..."
That way, if someone does try it and it works for them they won't think they can "fix" it.
If you want to keep it on one line, you can probably just have something like:
"Use String here since on some systems Ant doesn't support automatically converting Path instances."

daniel-ju:
> Reply once your happy with the comment and I can merge.

Thanks! (The ticket hasn't been accepted yet though, I can merge once it's been accepted.)

hagbard:
I've asked to take the ticket, so once it's assigned I can merge immediately if you want.
I'm good with all the changes now; thanks for this!

daniel-ju:
Great, feel free to merge once the ticket is accepted.
Thanks for your feedback as well!

PR #1162, opened 2020-06-02
ICU-21142 Moving CLDR jar file to common location

hagbard:
This paves the way to easily add new utilities in ICU which depend upon the CLDR API.
Any new tools can trivially be built using Maven by simply inheriting the CLDR API pom file.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21142
- [ ] Issue accepted
- [ ] Tests included


tools/cldr/lib/install-cldr-jars.sh
```diff
@@ -97,19 +92,6 @@ run_with_logging mvn -B install:install-file \
   -DlocalRepositoryPath=. \
   -Dfile="${CLDR_TOOLS_DIR}/cldr.jar"
 
-echo "Installing CLDR utilities JAR file..."
```
hagbard, line 100: Not needing this utilities JAR is a requirement for expanding who can use the CLDR API in ICU (since the code in this utility JAR comes from ICU originally and creates the potential for duplicate (incompatible) classes at build time.

srl295:
Is this a CLDR or ICU ticket?  This is ICU, but the subject line and commit has a CLDR-21142 ticket number, but the PR body has a link to a ICU-21142

markusicu:
Making the CLDR API library available in a shared place sounds good.

For a place in ICU, maybe we could make a "lib" subfolder under https://github.com/unicode-org/icu/tree/master/icu4j/main/shared

However, should this library not live in the CLDR repo? As an export there, rather than an import here?

hagbard:
> For a place in ICU, maybe we could make a "lib" subfolder under https://github.com/unicode-org/icu/tree/master/icu4j/main/shared

Why in icu4j? Since this can only apply to offline tools, it feels odd for it to be there, rather than in the top-level tools/ directory. To me icu4j feels like it should be for code used in the end-user libraries, not as part of data generation that *just happens to be in Java*.

> However, should this library not live in the CLDR repo? As an export there, rather than an import here?

No, Maven needs to have a consistent relative path to this directory from other POM files in ICU.
The *absolutely* better thing to do would be to publish the artifact to Maven and just *use* it as normal rather that messing about with local repositories.

hagbard:
As discussed in the TC meeting, I think this fine for now.
We can come back to this once it's in and I can get to work on the locale distance stuff.

hagbard:
Still waiting for a review on this.

PR #1163, opened 2020-06-02
ICU-21140 Make UTF-8 explicit for all file access.

hagbard:
I found one more case where UTF-8 had to be specified and I made it explicit in all the remaining file accesses I could find. While I'm glad UTF-8 is default in Files.newBufferedReader(), the fact that Java has for so many years assumed that "no charset" implies "system default" means the people reading the code are potentially going to be confused if it's omitted, so I just added it consistently everywhere.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21140
- [x] Issue accepted
- [ ] Tests included


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -274,7 +275,7 @@ private static void checkBuild(boolean condition, String message, Object... args
 
     private static ImmutableList<String> readLinesFromResource(String name) {
         try (InputStream in = CleanOutputDirectoryTask.class.getResourceAsStream(name)) {
-            return ImmutableList.copyOf(CharStreams.readLines(new InputStreamReader(in)));
```
hagbard, line 277: This was the one place that actually needed it, the rest of the changes are "no ops" but I think improve readability.

PR #1170, opened 2020-06-14
ICU-21149 Adding a helper to allow simpler debugging of mappers.

hagbard:
Small class so it's easy to run mappers directly instead of as part of the main tool-chain.
Note that this is going to best work for the supplemental data mappers, since they typically don't need any config from the Ant build file.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21149
- [x] Issue accepted
- [ ] Tests included


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/DebugWriter.java
```diff
@@ -0,0 +1,126 @@
...
+     * Writes the IcuData generated by the given function using the default {@code CLDR_DIR}
+     * system property.
+     *
+     * <p>This is a helper method to make it easy for each mapper to have its own main
+     * method for debugging, and it should not be used directly by {@code LdmlConverter}.
+     */
+    public static void writeMultipleForDebugging(
+            String[] args, Function<CldrDataSupplier, Collection<IcuData>> fn)
+            throws IOException {
+        String cldrPath = System.getProperty("CLDR_DIR", System.getenv("CLDR_DIR"));
```
markusicu, line 65: Should this not call whatever CLDR code typically calls for reading properties/env vars?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/DebugWriter.java
```diff
@@ -0,0 +1,126 @@
...
+                "java.util.logging.ConsoleHandler.formatter = " + LogFormatter.class.getName(),
+                "",
+                PACKAGE_ROOT + ".level  = " + logLevel);
+        LogManager.getLogManager()
+                .readConfiguration(new ByteArrayInputStream(loggerConfig.getBytes(UTF_8)));
+
+        Files.createDirectories(outDir);
+        CldrDataSupplier src = CldrDataSupplier.forCldrFilesIn(Paths.get(cldrPath));
+        for (IcuData icuData : fn.apply(src)) {
+            IcuTextWriter.writeToFile(icuData, outDir, ImmutableList.of("Debug output"), true);
```
markusicu, line 85: pull the ImmutableList.of() out of the loop?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/DebugWriter.java
hagbard: It can't really, that's not an available API. This reads a property and falls back to the environment variable, which is essentially the same behaviour.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/DebugWriter.java
hagbard: True.

PR #1172, opened 2020-06-15
ICU-21078 Improve instructions and gitignore file for cldr-to-icu

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21078
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/.gitignore
+++ b/tools/cldr/.gitignore
@@ -3,5 +3,5 @@
 !/lib/README.txt
 
 # Ignore the default Maven target directory.
-/target
+/cldr-to-icu/target
 
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ perf-*.xml
 test-*.xml
 
 # Directories
+.idea/
 .vs/
 .vscode/
 ARM/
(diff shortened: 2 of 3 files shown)


tools/cldr/.gitignore
```diff
@@ -3,5 +3,5 @@
 !/lib/README.txt
 
 # Ignore the default Maven target directory.
-/target
```
hagbard, line 6: Good pont. Can you also add any occurrence of directory /.idea/ here?
It's what IntelliJ uses.

hugovdm: Where do these directories appear? If I make it "/.idea", I think it means only ".idea" in the same directory as the gitignore file. If I make it ".idea", it's also in the subdirectories.

hugovdm: Added `.idea/` in the top-level .gitignore file, next to `.vscode/`.

hugovdm:
I squashed commits, this being standard procedure in CLDR. It looks like this looses me my approval though?
"@hagbard approved this pull request."

PR #1174, opened 2020-06-16
ICU-21169 Add SimpleUnitImpl::getSimpleUnitID() method

hugovdm:
Also swap a UCharsTrie for a BytesTrie, providing us with a set of char* values we can also return through getIdentifier().

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21169
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -12,18 +12,18 @@
 // Helpful in toString methods and elsewhere.
 #define UNISTR_FROM_STRING_EXPLICIT
 
-#include <cstdlib>
 #include "cstring.h"
 #include "measunit_impl.h"
 #include "uarrsort.h"
 #include "uassert.h"
 #include "ucln_in.h"
 #include "umutex.h"
+#include "unicode/bytestrie.h"
+#include "unicode/bytestriebuilder.h"
 #include "unicode/errorcode.h"
 #include "unicode/localpointer.h"
 #include "unicode/measunit.h"
-#include "unicode/ucharstrie.h"
-#include "unicode/ucharstriebuilder.h"
+#include <cstdlib>
 
 #include "cstr.h"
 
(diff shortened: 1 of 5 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -777,6 +775,7 @@ MeasureUnit SingleUnitImpl::build(UErrorCode& status) const {
     return std::move(temp).build(status);
 }
 
+const char *SingleUnitImpl::getIdentifier() const { return gSimpleUnits[index]; }
```
hugovdm, line 778: Multi-line: check/change clang-format options...

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -230,37 +230,36 @@ UBool U_CALLCONV cleanupUnitExtras() {
...
-    b.add(u"p13-", POWER_PART_P13, status);
-    b.add(u"p14-", POWER_PART_P14, status);
-    b.add(u"p15-", POWER_PART_P15, status);
+    b.add("-per-", COMPOUND_PART_PER, status);
+    b.add("-", COMPOUND_PART_TIMES, status);
+    b.add("-and-", COMPOUND_PART_AND, status);
+    b.add("per-", INITIAL_COMPOUND_PART_PER, status);
+    b.add("square-", POWER_PART_P2, status);
+    b.add("cubic-", POWER_PART_P3, status);
+    b.add("p2-", POWER_PART_P2, status);
```
younies, line 249: I think it is better to have also "p1-"

icu4c/source/i18n/measunit_impl.h
```diff
@@ -32,6 +32,8 @@ struct SingleUnitImpl : public UMemory {
     /** Transform this SingleUnitImpl into a MeasureUnit, simplifying if possible. */
     MeasureUnit build(UErrorCode& status) const;
 
+    const char *getIdentifier() const;
```
younies, line 35: Optional: more documentation please

icu4c/source/i18n/measunit_extra.cpp
hugovdm: The specification specifically supports "pow2" to "pow15":
http://unicode.org/reports/tr35/tr35-general.html#Unit_Identifiers

A bigger concern: that specification says "pow#", not "p#"...
I find `pow4-second-square-ampere-per-kilogram-square-meter` in the unitQuantities resource.
We may need to do a s/p/pow/ - but it's also out of scope for this PR.

icu4c/source/i18n/measunit_impl.h
sffc: It would be useful to say that the returned identifier points to static memory and does not need to be cleaned up.

hugovdm: getIdentifier was misleading because this won't return "square-kilometer", even though that would be the identifier for a MeasureUnit containing only this SingleUnit. I've renamed it to "getSimpleUnitID" as inspired by terminology from
https://unicode-org.github.io/cldr/ldml/tr35-general.html#Unit_Identifiers
- and I've improved the documentation.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Reformatted.
The .clang_format option for this is:

    AllowShortFunctionsOnASingleLine: false

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -271,14 +273,13 @@ void U_CALLCONV initUnitExtras(UErrorCode& status) {
...
-    b.buildUnicodeString(USTRINGTRIE_BUILD_FAST, result, status);
+    StringPiece result = b.buildStringPiece(USTRINGTRIE_BUILD_FAST, status);
     if (U_FAILURE(status)) { return; }
 
     // Copy the result into the global constant pointer
-    size_t numBytes = result.length() * sizeof(char16_t);
-    kSerializedUnitExtrasStemTrie = static_cast<char16_t*>(uprv_malloc(numBytes));
-    uprv_memcpy(kSerializedUnitExtrasStemTrie, result.getBuffer(), numBytes);
+    size_t numBytes = result.length();
+    kSerializedUnitExtrasStemTrie = static_cast<char *>(uprv_malloc(numBytes));
```
jefgen, line 281: Technically unrelated to this specific change/PR, but `uprv_malloc`/`malloc` can fail for some platforms (due to out-of-memory conditions) and return `NULL`/`nullptr`.

If you don't mind, I think it would be good to check for `nullptr` here, and set the error code to `U_MEMORY_ALLOCATION_ERROR`. :)

Perhaps something like this:
```c++
if (kSerializedUnitExtrasStemTrie == nullptr) {
    return;
}
uprv_memcpy(kSerializedUnitExtrasStemTrie, result.data(), numBytes);
```

sffc: Yeah, good catch

hugovdm: Done!

hugovdm:
Squashing!

icu4c/source/i18n/measunit_extra.cpp
jefgen: Thanks!

hugovdm:
Sorry, I somehow managed to not add line breaks to the one-liner function.
880b0e1 added the nullptr check.
701128b added two line-breaks, and hasn't been reviewed, but I'm confident...
I'm boldly squashing again without waiting, to save a round-trip.

hugovdm:
My conclusion is that my latest action is indeed highly un-recommended and should be strongly discouraged. What I learned: the Jira-GitHub PR Checker Bot claims "no changes!" - but that's because it was comparing 7b30cd7 to 701128b, with 701128b thereby flying under GitHub PR Radar.

The actions were:
880b0e1 added the nullptr check, and was approved.
I squashed to: bdd2ae8.
I pushed another commit: 701128b.
I squashed to 7b30cd7 not waiting for review.

Probably quite non-trivial to improve automation for this. History can be investigated via links in "[person] force-pushed <...>" log entries. Particularly if someone pushes more than one commit before squashing again... (one has to click on 701128b and then check its parent if one wants to verify there were no other commits between it and bdd2ae8).

sffc:
> Probably quite non-trivial to improve automation for this. History can be investigated via links in "[person] force-pushed <...>" log entries. Particularly if someone pushes more than one commit before squashing again... (one has to click on 701128b and then check its parent if one wants to verify there were no other commits between it and bdd2ae8).

This is precisely what an astute reviewer should be doing.

PR #1176, opened 2020-06-18
ICU-20465 Calendar/DateFormat listen to tz extension

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20465
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -870,7 +870,7 @@ Calendar::createInstance(const TimeZone& zone, UErrorCode& success)
 Calendar* U_EXPORT2
 Calendar::createInstance(const Locale& aLocale, UErrorCode& success)
 {
-    return createInstance(TimeZone::createDefault(), aLocale, success);
+    return createInstance(TimeZone::forLocaleOrDefault(aLocale), aLocale, success);
 }
 
 // ------------------------------------- Adopting
--- a/icu4c/source/i18n/unicode/timezone.h
+++ b/icu4c/source/i18n/unicode/timezone.h
@@ -318,6 +318,17 @@ public:
     static TimeZone* U_EXPORT2 createDefault(void);
 
     /**
+     * If the locale contains the timezone keyword, creates a copy of that TimeZone.
+     * Otherwise, create the default timezone.
+     *
+     * @param locale a locale which may contains 'timezone' keyword/value.
+     * @return   A TimeZone. Clients are responsible for deleting the time zone
+     *           object returned.
+     * @internal
+     */
+    static TimeZone* U_EXPORT2 forLocaleOrDefault(const Locale& locale);
+
+    /**
      * Sets the default time zone (i.e., what's returned by createDefault()) to be the
      * specified time zone.  If NULL is specified for the time zone, the default time
      * zone is set to the default host time zone.  This call adopts the TimeZone object
--- a/icu4j/main/classes/core/src/com/ibm/icu/util/CECalendar.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/util/CECalendar.java
@@ -76,7 +76,7 @@ abstract class CECalendar extends Calendar {
      * @param aLocale The locale for the new calendar.
      */
     protected CECalendar(Locale aLocale) {
-        this(TimeZone.getDefault(), aLocale);
+        this(TimeZone.forLocaleOrDefault(aLocale), aLocale);
     }
 
     /**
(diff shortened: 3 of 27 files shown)


FrankYFTang:
@yumaoka ping

FrankYFTang:
@yumaoka  Do you have time to code review this PR? Thanks

icu4c/source/i18n/calendar.cpp
```diff
@@ -870,7 +870,28 @@ Calendar::createInstance(const TimeZone& zone, UErrorCode& success)
...
+// ------------------------------------- Adopting
+
+void
+Calendar::adoptTimeZoneFromLocale(const Locale& aLocale, UErrorCode &status) {
+    if (U_FAILURE(status)) {
+        return;
+    }
+    char buffer[ULOC_KEYWORDS_CAPACITY] = "";
+    int32_t count = aLocale.getKeywordValue("timezone", buffer, sizeof(buffer), status);
+    if (U_FAILURE(status) || status == U_STRING_NOT_TERMINATED_WARNING) {
```
sffc, line 887: it is better practice to make `localStatus`, rather than setting `status` and unsetting it later, for two reasons:

1. Resetting `status` to `U_ZERO_ERROR` might delete warning statuses
2. People have debuggers that look for changes to their `status` variable, and touching it will trigger their debugger

icu4c/source/i18n/calendar.cpp
```diff
@@ -870,7 +870,28 @@ Calendar::createInstance(const TimeZone& zone, UErrorCode& success)
 Calendar* U_EXPORT2
 Calendar::createInstance(const Locale& aLocale, UErrorCode& success)
 {
-    return createInstance(TimeZone::createDefault(), aLocale, success);
+    Calendar* res = createInstance(TimeZone::createDefault(), aLocale, success);
```
sffc, line 873: Did you consider making a function that returns the locale's time zone if it is present, or `TimeZone::createDefault()` if it is not, instead of `adoptTimeZoneFromLocale`?  You could also use it in `smpdtfmt.cpp`.  For example:

```cpp
TimeZone* TimeZone::forLocaleOrDefault(const Locale& locale) {
  if (/* locale has time zone keyword */) {
    // return locale time zone
  } else {
    return TimeZone::createDefault();
  }
}
```

icu4c/source/i18n/calendar.cpp
FrankYFTang: DONE

icu4c/source/i18n/calendar.cpp
FrankYFTang: DONE

FrankYFTang:
PTAL

FrankYFTang:
@sffc  ping

icu4c/source/i18n/unicode/timezone.h
```diff
@@ -317,6 +317,17 @@ class U_I18N_API TimeZone : public UObject {
      */
     static TimeZone* U_EXPORT2 createDefault(void);
 
+    /**
+     * If the locale contains the timezone keyword, creates a copy of that TimeZone.
+     * Otherwise, create the defaul timezone.
```
sffc, line 322: s/defaul/default/

Also, I would rather see this as `@draft ICU 68` than internal.  The methods above and below are stable, so why not this one?

icu4j/main/classes/core/src/com/ibm/icu/util/CECalendar.java
```diff
@@ -86,7 +86,10 @@ protected CECalendar(Locale aLocale) {
      * @param locale The locale for the new calendar.
      */
     protected CECalendar(ULocale locale) {
-        this(TimeZone.getDefault(), locale);
+        this((locale.getKeywordValue("timezone") == null)
```
sffc, line 89: Why didn't you make the `forLocaleOrDefault` change in Java, too?

icu4c/source/i18n/unicode/timezone.h
FrankYFTang: then we need to make an API proposal, right? But why do we need a public API for this?

icu4j/main/classes/core/src/com/ibm/icu/util/CECalendar.java
FrankYFTang: ok added.

FrankYFTang:
PTAL @sffc

icu4c/source/i18n/unicode/timezone.h
sffc: I think it would be a useful public API; it makes it easier to use the `u-tz` extension, which many applications are probably currently ignoring.  However, you can check in this PR with it as `@internal` and open a follow-up ticket to promote it to `@draft` later.

sffc:
LGTM please squash

FrankYFTang:
@sffc PTAL

PR #1178, opened 2020-06-18
ICU-21160 Fix the length return by preflight

FrankYFTang:
When the function is called during preflight,
return the max of the two possible results.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21160
- [X] Issue accepted
- [X] Tests included


FrankYFTang:
@srl295 @younies  ping

FrankYFTang:
PTAL

sffc:
LGTM please squash

PR #492, opened 2019-02-26
ICU-20438 Adds fuzzer directory to testinternalheaders.sh

gnrunge:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20438
- [ ] Updated PR title and link in previous line to include Issue number
- [ ] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #1181, opened 2020-06-18
ICU-20697 Delete now unused files and documentation for the old ICU LDML tooling.

hagbard:
I think this covers the files that can be deleted here, but there might be more docs I am unaware of.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20697
- [ ] Issue accepted
- [ ] Tests included


sffc:
LGTM for the removed XML files. I'd like Peter to review the changes to the instructions file.

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -97,34 +97,6 @@
...
-#                <target name="clean" ...> to clean the correct set of files.
-#
-#                If there are new CLDR resource bundle types, you may need to
-#                updated the <remapper> sections to put these in the correct
-#                data subfolder for ICU.
-#
-#    $ICU4C_DIR/source/data/xml/      - If you are adding a new locale, break
-#                iterator, collation tailoring, or rule-based number formatter,
-#                you may need to add a corresponding xml file in (respectively)
-#                the main/, brkitr/, collation/, or rbnf/ subdirectory here.
```
pedberg-icu, line 126: In the new tools, where is the correspondence now established between a particular locale and break type (such as locale "ja" with break type 'line alt="loose"') with a particular rules file (like line_loose_cj.txt source, or line_loose_cj.brk compiled)?

hagbard: Hmm, this is not something I know anything about. It's not something which was encoded anywhere in the CLDR tooling I converted. Is it part of the CLDR-to-ICU conversion or a post-processing step?

Can you show me an example of the difference this setting would produce?

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -166,39 +138,12 @@ export ICU4C_DIR=$HOME/icu-myfork/icu4c
 export ICU4J_ROOT=$HOME/icu-myfork/icu4j
 export TOOLS_ROOT=$HOME/icu-myfork/tools
 
+# 2. Build an install the CLDR jar
```
robertgmelo, line 141: Typo here. I suppose it's **and**

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -239,10 +184,10 @@ ant proddata 2>&1 | tee /tmp/cldr-newData-proddataLog.txt
 # be ones that it would generate, except for pasts listed in <retain> elements such as
 # coll/de__PHONEBOOK.txt, coll/de_.txt, etc.
 #
-# Before this step, make any necessary changes to
-# build-icu-data.xml$TOOLS_ROOT/cldr/cldr-to-icu/ to add new locales, etc.
+# Before running Ant to regenerate the data, make any necessary changes to the
```
robertgmelo, line 187: I suppose we also should remove the commands showed after 4a. instructions `(ant cleanprod, andt setup, ant proddata)`, correct? Or are we planning to remove/move build.xml?
@sffc, you were right (tks in advance). I've spent more time analyzing the error, and this PR may answer the question. Indeed, CLDRBuild is one of the files removed on CLDR-13829 (#492).

icu4c/source/data/cldr-icu-readme.txt
hagbard: Done

icu4c/source/data/cldr-icu-readme.txt
hagbard: No, that has to stay I think. That relates to the fact that CLDR data in the main repository must now be pre-processed before it can be consumed by the new tools I've written. This change happened at about the same time, but is separate from it. The CLDR-to-ICU tool reads data from the "staging" repository, which corresponds to the "production" data mentioned there.

hagbard:
Are people still interested in doing this, or should I abandon this PR?

icu4c/source/data/cldr-icu-readme.txt
robertgmelo: Got it. My point is just because though this will not be part of the data generation anymore, in case someone tries to run one of those commands (cleanprod, setup, proddata), errors will be triggered since the class CLDRBuild was dropped from the project, so I was wondering that maybe we could avoid those errors.

hagbard: There must be something I don't understand then. The "tools/java/org/unicode/cldr/ant/CLDRBuild.java" was removed from the CLDR project, but the instructions in (4a) refer to an Ant file in "$ICU4C_DIR/source/data". Does this Ant file rely on the removed LDML-to-ICU tools then?

Ahh ha!
https://github.com/unicode-org/icu/blob/5722aa45a004c53791768585227f0e555b9d44c8/icu4c/source/data/build.xml#L129
So I didn't know that was a thing (a bit poor from my side).

Okay, yes, it sounds like most (maybe all) of that build.xml file can go as well, or it could in theory be made to call the new converter instead. Having not looked at it yet I don't know enough to be sure yet. I'll dig into it today.

Thanks for pointing it out!

hagbard: Ahh it seems that "cleanprod", "setup" and "proddata" don't use the tools that were removed, so should continue to work fine (and are a very necessary step that must happen before running the conversion tools).
So these docs must stay, but the XML file can have the unused bits referencing the old tools removed from it.

robertgmelo: Last time I ran, the error was happening on "setup":
https://github.com/unicode-org/icu/blob/5722aa45a004c53791768585227f0e555b9d44c8/icu4c/source/data/build.xml#L59

I can try again.

hagbard: That might just be a dependency that needs removing then (i.e. a failing step that shouldn't even be there).
I'll have a look now.

hagbard: I think I fixed this in the latest push, it just means deleting the unused targets.

hagbard:
I think this is ready to go in now, but happy to wait if the timing is an issue (re. releases).

PR #1182, opened 2020-06-18
ICU-21171 Remove sgn-* from REDUNDANT

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21171
- [X] Issue accepted
- [X] Tests included


FrankYFTang:
ICU-TC decide to let CLDR-TC to add those into the files.

PR #1183, opened 2020-06-19
ICU-21174 Add a const version of MaybeStackVector::getAlias().

hugovdm:
(Also makes a tiny tweak to appendAll() documentation.)

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -779,6 +779,10 @@ public:
         return this->fPool.getAlias();
     }
 
+    const T *const *getAlias() const {
+        return this->fPool.getAlias();
+    }
+
     /**
      * Array item access (read-only).
      * No index bounds check.
@@ -800,7 +804,7 @@ public:
     }
 
     /**
-     * Append all the items from another MaybeStackVector to this one.
+     * Append copies of all the items from another MaybeStackVector to this one.
      */
     void appendAll(const MaybeStackVector& other, UErrorCode& status) {
         for (int32_t i = 0; i < other.fCount; i++) {


icu4c/source/common/cmemory.h
```diff
@@ -728,6 +735,28 @@ class MemoryPool : public UMemory {
...
 
+    /**
+     * Takes ownership of ptr, adding it to the pool.
+     *
+     * @param ptr Pointer to take ownership of.
+     * @param status Is set to U_MEMORY_ALLOCATION_ERROR if allocation of the
+     * needed pool capacity fails.
+     */
+    template<typename... Args>
+    void adoptBack(T* ptr, UErrorCode& status) {
```
sffc, line 746: I would make the MemoryPool version "adopt", since MemoryPool has no concept of order.

icu4c/source/common/cmemory.h
```diff
@@ -728,6 +735,27 @@ class MemoryPool : public UMemory {
...
     }
 
+    /**
+     * Takes ownership of ptr, adding it to the pool.
+     *
+     * @param ptr Pointer to take ownership of.
+     * @param status Is set to U_MEMORY_ALLOCATION_ERROR if allocation of the
+     * needed pool capacity fails.
+     */
+    template <typename... Args> void adopt(T *ptr, UErrorCode &status) {
```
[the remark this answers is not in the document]
hugovdm: Done - thereby also consistent with the rest of the file, thanks for catching it!

hugovdm:
Squashed and force-pushed based on sffc's approval 17 hours ago. (Click through to pre-force-push 46cf976 to review any individual commits since.)

icu4c/source/common/cmemory.h
```diff
@@ -323,6 +323,11 @@ class MaybeStackArray {
      * @return the array pointer
      */
     T *getAlias() const { return ptr; }
+    /**
+     * Access without ownership change.
+     * @return the array pointer
+     */
+    const T *getConstAlias() const { return ptr; }
```
roubert, line 330: As the `getAlias()` method already is marked `const`, these two statements would have the same result:

```
const T* p1 = msa.getAlias();
const T* p1 = msa.getConstAlias();
```

As would these:

```
const auto* p1 = msa.getAlias();
const auto* p1 = msa.getConstAlias();
```

Do you have a use-case where this new `getConstAlias()` method makes it possible to do something that itsn't already possible to do with the existing `getAlias()` method?

It is a bit unusual that the existing `getAlias()` method is `const` but returns a pointer to a non-`const` object. The PR description and the associated ticket are both very vague on what the purpose of these new methods are, but they both mention _const correctness_.

Might it be that what you really would want to do here is to change the existing `getAlias()` method to make it non-`const` and then add an overload that is `const` and returns a pointer to a `const` object?

That'd be a rather common pattern used to ensure _const correctness_ by requiring that the contents of the container be modified only through a non-`const` reference to the container object:

```
T *getAlias() { return ptr; }
const T *getAlias() const { return ptr; }
```

icu4c/source/common/cmemory.h
```diff
@@ -728,6 +735,28 @@ class MemoryPool : public UMemory {
...
 
+    /**
+     * Takes ownership of ptr, adding it to the pool.
+     *
+     * @param ptr Pointer to take ownership of.
+     * @param status Is set to U_MEMORY_ALLOCATION_ERROR if allocation of the
+     * needed pool capacity fails.
+     */
+    template <typename... Args>
+    void adopt(T *ptr, UErrorCode &status) {
```
roubert, line 746: What is the use-case for this method? Is there a use-case where `MemoryPool` is a suitable data structure to use but the `create()` method isn't possible to use?

The reason for why `MemoryPool` currently has a `create()` method but no `adopt()` method is that this makes memory management both simple and unambiguous. When the `MemoryPool` object goes out-of-scope, it deletes all objects that it has created and that's all there is to it.

This proposed `adopt()` method does instead require rather intricate error handling for memory allocation errors:

```
MemoryPool<MyType> pool;

MyType* o1 = new MyType(1, 2, 3);
if (o1 != nullptr) {
  ErrorCode status;
  pool.adopt(o1, status);
  if (status == U_MEMORY_ALLOCATION_ERROR) {
    delete o1;
    o1 = nullptr;
  }
}
```

Compare that to the example in the class comment above that calls the `create()` method.

This looks discouragingly error-prone to me so I wonder what it might be that otherwise isn't possible that this would make possible.

icu4c/source/common/cmemory.h
```diff
@@ -800,7 +844,8 @@ class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
     }
 
     /**
-     * Append all the items from another MaybeStackVector to this one.
+     * Append all the items from another MaybeStackVector to this one via
+     * copy-construction.
```
roubert, line 848: I'm not really sure what it is that you want to convey with your edit to this comment. Is maybe "Append copies of all items from another MaybeStackVector to this one." what you want to say?

icu4c/source/common/cmemory.h
```diff
@@ -771,14 +800,29 @@ class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
...
+    void adoptBack(T *ptr, UErrorCode &status) {
+        this->adopt(ptr, status);
+    }
+
     int32_t length() const {
         return this->fCount;
     }
 
-    T** getAlias() {
+    T** getAlias() const {
```
roubert, line 818: Why are you making this method `const`?

icu4c/source/common/cmemory.h
```diff
@@ -771,14 +800,29 @@ class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
...
     }
 
+    /**
+     * Takes ownership of ptr, adding it to the back of the vector.
+     *
+     * @param ptr Pointer to take ownership of.
+     * @param status Is set to U_MEMORY_ALLOCATION_ERROR if allocation of the
+     * needed pool capacity fails.
+     */
+    void adoptBack(T *ptr, UErrorCode &status) {
```
roubert, line 810: This method has the same error handling issues as `MemoryPool::adopt()` so it raises the same questions.

icu4c/source/common/cmemory.h
hugovdm: My use-case is to have MaybeStackVector hold pointers that were created by a factory method. My usage looks a bit like this:

        fLongNameHandlers.adoptBack(
            LongNameHandler::forMeasureUnit(
                loc, *units[i], MeasureUnit(), width, rules, NULL, status),
            status);

I'm also sad about making MemoryPool more error prone (hence my uncomfortable comment change above). Does it make sense to make it a private method though, for use by MaybeStackVector::adoptBack()?

icu4c/source/common/cmemory.h
hugovdm: I want to call it from a method that is marked const (accessing a member instance of the MaybeStackVector class).

getAlias is const in the sense that it doesn't modify the fPool pointer, but the contents of the pool may of course be modified: I now recognise, from what you wrote, that that might be rather confusing, and "const" methods on container types should perhaps be "deeply const". I've now created an overloaded getAlias(), how does this look?

icu4c/source/common/cmemory.h
hugovdm: Yes, that sounds much better, thanks!

icu4c/source/common/cmemory.h
hugovdm: (Let's have this discussion on the MemoryPool::adopt() thread?)

icu4c/source/common/cmemory.h
roubert: You can't do it like that, your example code snippet will leak the object created by `forMeasureUnit()` if the call to `adoptBack()` fails.

Let's take a step back and look at how you'd like to do memory management in the code you're writing.

If these `LongNameHandler` objects are cheap to copy or move, one option that would simplify memory management considerably would be to let `forMeasureUnit()` return the newly created object by value and then invoke the copy or move constructor through `fLongNameHandlers.emplaceBack()`.

If it isn't possible to make the copy or move constructor cheap enough, an old commonly used pattern is to first create an empty object and then pass a pointer to this object to the factory function that then fills in the data. Like this:

```
LongNameHandler* obj = fLongNameHandlers.emplaceBack();
if (obj != nullptr) {
  LongNameHandler::forMeasureUnit(
              obj, loc, *units[i], MeasureUnit(), width, rules, NULL, status),
}
```

I guess you've seen such code before. It's a bit cumbersome, but it's a very simple way to avoid creating an object that then must be deleted again because it can't be stored anywhere.

And there are other possible solutions, all depending on your bigger picture here.

hugovdm: Thanks - I'm tackling a hierarchy of move and copy constructors now (maybe I only need the move constructors).

sffc: Another pattern might be if `forMeasureUnit` returned the LongNameHandler by value, and then you can move-construct it into the emplaceBack container.

hugovdm: No new self-defined constructors necessary, it does seem to work:

        auto *lnh = LongNameHandler::forMeasureUnit(...);
        result->fLongNameHandlers.emplaceBack(*lnh);

And I guess this might do move construction? - But it's still not something I'm confident about:

        result->fLongNameHandlers.emplaceBack(std::move(*lnh));

sffc: Two problems:

1. The first code example is probably implicitly using the implicitly defined copy constructor.  (This is one of my least favorite things about C++: copy constructor should be explicit, but it's not.)
2. Even if you use `std::move` out of the pointer, like in the second example, you are still doing 2 heap allocations.  Better is to change the signature of `forMeasureUnit` to return by value, such that you can do,

```cpp
result->fLongNameHandlers.emplaceBack(std::move(LongNameHandler::forMeasureUnit(...)));
```

I think the `std::move` above might not be necessary, but it can make it more explicit what you are trying to do.

hugovdm: This PR is now so small, I'm wondering if it is still worth submitting/merging separately. :)

hugovdm: @sffc I guess I should apply this style principle to LongNameHandler:
https://google.github.io/styleguide/cppguide.html#Copyable_Movable_Types
I'll try to do that in my "Glue Code PR" this afternoon (CEST).

roubert: > The first code example is probably implicitly using the implicitly defined copy constructor.

As of C++11 this copy constructor can be deleted by using the `= deleted` syntax. Even in situations where you want the class to have a copy constructor this can be used during development if you at some point want to verify that the copy constructor isn't called.

Another way to easily verify during development that the intended constructor really is called is to add assertion failures to all other constructors, then a simple run of the unit tests immediately shows exactly which constructor was invoked in any situation where this was unclear.

You can use that to verify that the use of `std::move()` in a situation like this (an object returned by value passed to a move constructor) is a no-op that can (and probably should) be omitted.

icu4c/source/common/cmemory.h
```diff
@@ -779,6 +779,10 @@ class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
         return this->fPool.getAlias();
     }
 
+    const T *const *getAlias() const {
```
roubert, line 782: This new `getAlias()` method now looks fine and makes perfect sense in this context. Please squash your commits and write a commit message that describes what this now actually does.

hugovdm: Done.

icu4c/source/common/cmemory.h
hugovdm: Copy or move is not possible, because in the process of constructing the instance, other objects are given pointers to it. When it is moved or copied, those pointers fail.

So if we don't add "pointer adoption" to our containers, I'll try adding a fillIn parameter to the factory functions, such that if not null, the fill-in is used, otherwise a new instance is allocated.

hugovdm: Refactoring the client code in question to cope with a fillIn: 3a575c9510

PR #1185, opened 2020-06-21
ICU-21142 Hopefully fixing the install script (bash only)

hagbard:
A few tweaks to the install jar script to make it less dependent on Unix/Linux tools.
I think that (perhaps with the exception of 'which') everything is now a command that exists in the Bash man-page.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21142
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/lib/install-cldr-jars.sh
+++ b/tools/cldr/lib/install-cldr-jars.sh
@@ -15,6 +15,12 @@
 # Usage (from the directory of this script):
 #
 # ./install-cldr-jars.sh <CLDR-root-directory>
+#
+# Note to maintainers: This script cannot be assumed to run on a Unix/Linux
+# based system, and while a Posix compliant bash shell is required, any
+# assumptions about auxiliary Unix tools should be minimized (e.g. things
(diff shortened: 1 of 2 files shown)


tools/cldr/lib/install-cldr-jars.sh
```diff
@@ -67,9 +66,10 @@ which mvn > /dev/null || die "Cannot find Maven executable 'mvn' in the current
 (( $# == 1 )) && [[ -d "$1" ]] || die "Usage: ./install-cldr-jars.sh <CLDR-root-directory>"
 
 # Set up a log file (and be nice about tidying it up).
-LOG_FILE="$(tempfile)" || die "Cannot create temporary file!"
-trap "rm -f -- '${LOG_FILE}'" EXIT
```
hagbard, line 71: I got rid of the tidyup since now the log file isn't a random name, it seems fine to leave it in /tmp and avoid "polluting" the lib/ directory with a "last_log" file. I'm easy either way so if anyone prefers the copy and tidyup behaviour, just let me know.

PR #1191, opened 2020-07-01
ICU-13590 RBBI, improve handling of concurrent look-ahead rules.

aheninger:
Change the mapping from rule number to boundary position to use a simple array
instead of a linear search lookup map.

Look-ahead rules have a preceding context, a boundary position, and following context.
In the implementation, when the preceding context matches, the potential boundary
position is saved. Then, if the following context proves to match, the saved boundary is
returned as an actual boundary.

Look-ahead rules are numbered, and the implementation maintains a map from
rule number to the tentative saved boundary position.

In an earlier improvement to the rule builder, the rule numbering was changed to be a
contiguous sequence, from the original sparse numbering. In anticipation of
changing the mapping from number to position to use a simple array.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-13590
- [x] Issue accepted
- [x] Tests included


FrankYFTang:
sorry I just back from vacation. Will take a look at this soon.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1386,6 +1380,7 @@ void RBBITableBuilder::exportTable(void *where) {
 
     table->fNumStates = fDStates->size();
     table->fDictCategoriesStart = fRB->fSetBuilder->getDictCategoriesStart();
+    table->fLookAheadResultsSize = fLASlotsInUse == ACCEPTING_UNCONDITIONAL ? 0 : fLASlotsInUse + 1;
```
FrankYFTang, line 1383: Would it be easier to read by the following?
table->fLookAheadResultsSize = (fLASlotsInUse == ACCEPTING_UNCONDITIONAL) ? 0 : (fLASlotsInUse + 1);

PR #1192, opened 2020-07-01
ICU-21179 Clean up unused code in the RBBI dictbe

mikamikem:
The wordsMatched conditional gave the misleading impression that only the first "first word" candidate that was followed by a dictionary word would be marked, but because the wordsMatched declaration is inside the while loop, every "first word" candidate that's followed by a dictionary word is marked (resulting in the shortest and not the longest first candidate that's followed by a second dictionary word).

After making the change, the comment felt like it made more sense, so I didn't modify it in this change.  If anyone feels like it should change or if clarification should be made somewhere of what the result of this loop is, I'm open to making additional changes to clarify.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21179
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/dictbe.cpp
+++ b/icu4c/source/common/dictbe.cpp
@@ -265,13 +265,9 @@ ThaiBreakEngine::divideUpDictionaryRange( UText *text,
                 goto foundBest;
             }
             do {
-                int32_t wordsMatched = 1;
                 if (words[(wordsFound + 1) % THAI_LOOKAHEAD].candidates(text, fDictionary, rangeEnd) > 0) {
-                    if (wordsMatched < 2) {
-                        // Followed by another dictionary word; mark first word as a good candidate
-                        words[wordsFound%THAI_LOOKAHEAD].markCurrent();
-                        wordsMatched = 2;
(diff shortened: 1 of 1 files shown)


icu4c/source/common/dictbe.cpp
```diff
@@ -265,13 +265,9 @@ ThaiBreakEngine::divideUpDictionaryRange( UText *text,
                 goto foundBest;
             }
             do {
-                int32_t wordsMatched = 1;
```
aheninger, line 268: The proposed change certainly does the same thing as the existing code, but I don't really understand this code, so I don't know what the original intent of the wordsMatched test was. Maybe we have some other bug that ought to be fixed.

Also, the same code exists in ICU4J, and whatever we do here needs to be done there as well.
https://github.com/unicode-org/icu/blob/master/icu4j/main/classes/core/src/com/ibm/icu/text/ThaiBreakEngine.java#L137

sven-oly (approved):
I suggest that this change is fine, with no ill effects.

PR #1193, opened 2020-07-02
ICU-21182 Adding UNumberRangeFormatter.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21182
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/unumberrangeformatter.h
+++ b/icu4c/source/i18n/unicode/unumberrangeformatter.h
@@ -1,0 +1,432 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+#ifndef __UNUMBERRANGEFORMATTER_H__
+#define __UNUMBERRANGEFORMATTER_H__
+
+#include "unicode/utypes.h"
+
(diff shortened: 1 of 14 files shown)


sffc:
Squashed, and I also caught and fixed that the commit was misticketted!

sffc:
Can I have an approval for the squashed commit, please? :smiley:

icu4c/source/i18n/unicode/unumberrangeformatter.h
```diff
@@ -0,0 +1,431 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+#include "unicode/utypes.h"
+
+#if !UCONFIG_NO_FORMATTING
+#ifndef __UNUMBERRANGEFORMATTER_H__
```
markusicu, line 7: The guard should be between the copyright and the first include, like in other files. Remember to move the corresponding endif+comment to the last line, too.

sffc: Seems that we're not consistent here (I spot several files with my order), but fixed via commit amendment.

PR #1199, opened 2020-07-17
ICU-21066 Copy an includelist of CLDR testData when updating ICU resources

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21066
- [X] Issue accepted
- [ ] Tests included


hugovdm:
Azure Pipelines failed with this:

```
BUILD FAILED
/home/vsts/work/1/s/icu4j/build.xml:454: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:439: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:598: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:348: test failed: charset
```

Quick naive glance at the path makes me think it's not related to this PR... I can revisit later.

jefgen:
FWIW, it looks like there are CI failures on other PRs that aren't touching anything ICU4J related.

From taking a quick look at the logs, it seems like this is the failure:

```
Testsuite: com.ibm.icu.dev.test.charset.TestConversion
Tests run: 292, Failures: 263, Errors: 0, Skipped: 0, Time elapsed: 12.548 sec

Testcase: [0] TestDataPair (conversionTest) took 0.002 sec
	FAILED
UTF-16BE was not found
junit.framework.AssertionFailedError: UTF-16BE was not found
	at com.ibm.icu.dev.test.AbstractTestLog.errln(AbstractTestLog.java:80)
	at com.ibm.icu.dev.test.charset.TestConversion.FromUnicodeCase(TestConversion.java:258)
	at com.ibm.icu.dev.test.charset.TestConversion.TestFromUnicode(TestConversion.java:225)
	at com.ibm.icu.dev.test.charset.TestConversion.conversionTest(TestConversion.java:121)
	at junitparams.internal.InvokeParameterisedMethod.evaluate(InvokeParameterisedMethod.java:234)
	at junitparams.internal.ParameterisedTestMethodRunner.runMethodInvoker(ParameterisedTestMethodRunner.java:47)
	at junitparams.internal.ParameterisedTestMethodRunner.runTestMethod(ParameterisedTestMethodRunner.java:40)
	at junitparams.internal.ParameterisedTestClassRunner.runParameterisedTest(ParameterisedTestClassRunner.java:146)
	at junitparams.JUnitParamsRunner.runChild(JUnitParamsRunner.java:417)
	at junitparams.JUnitParamsRunner.runChild(JUnitParamsRunner.java:386)
```

> UTF-16BE was not found

Could something have changed recently with JUnit maybe on the build machines?

hugovdm:
> Could something have changed recently with JUnit maybe on the build machines?

I have no insights into this yet and can't investigate right now. I'm hoping there's someone "closer to ICU4J" who can shed some light on this issue? Maybe we should ask on icu-core?

sffc:
I think we should make an includelist of files to copy, and when someone needs a new file, they should add it to that list. Copying all 245 KB of test data unnecessarily increases the download size of the ICU repository.

.cpyskip.txt
```diff
@@ -107,6 +107,11 @@ docs/ide4c/vscode/*.json
 tools/trac/IcuCodeTools/*
 tools/unicodetools/*
 #
+# CLDR - TODO(review): verify that _readme.txt copyright notice is sufficient?
+#
+testdata/cldr-testData/transforms/*.txt
+testdata/cldr-testData/segmentation/graphemeCluster/*.txt
+#
```
hugovdm, line 114: Since we're now using an include-list, these txt files aren't copied anymore. I'll remove this file from the PR before squashing into a single commit.

testdata/cldr-testData/units/_readme.txt
```diff
@@ -0,0 +1,6 @@
+# Test data for units: conversions and preferences
```
hagbard, line 1: FMI: What's with the leading underscore?

tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -332,6 +335,18 @@
         </convert>
     </target>
 
+    <target name="copy-cldrtestdata" depends="init-args">
+        <copy todir="${testDataDir}/cldr-testData">
+            <fileset dir="${cldrDir}/common/testData">
+                <include name="units/**"/>
```
hagbard, line 341: Is this where you want people to add more directories?
If so, please comment as such.
Maybe also comment what the inclusions are intended for (so one day some could even be removed).

tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -332,6 +335,18 @@
...
+            </fileset>
+        </copy>
+    </target>
+
+    <target name="clean-cldrtestdata" depends="init-args">
+        <delete dir="${testDataDir}/cldr-testData"/>
+    </target>
+
```
hagbard, line 349: Why doesn't "clean" depend on clean-cldrtestdata ? I think I'd expect if I run "clean" then it's back to the same state as when I download the repository.

tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -332,6 +335,18 @@
         </convert>
     </target>
 
+    <target name="copy-cldrtestdata" depends="init-args">
```
hagbard, line 338: Surely copy should depend on clean, otherwise you risk inconsistent test data (which is a confusing state to be in).

tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -15,7 +15,7 @@
       inconsistent state when the conversion is cancelled). -->
 <project name="Convert" default="all" basedir=".">
 
-    <target name="all" depends="init-args, prepare-jar, clean, convert"/>
+    <target name="all" depends="init-args, prepare-jar, clean, convert, clean-cldrtestdata, copy-cldrtestdata"/>
```
hagbard, line 18: Why does "conversion" need test data at all? I really don't understand this.

testdata/cldr-testData/units/_readme.txt
hugovdm: I took a look around, couldn't find an answer, just an observation:

- All readme.txt files in the common/testData tree have a leading underscore.
- No other readme files in CLDR have a leading underscore.

tools/cldr/cldr-to-icu/build-icu-data.xml
hugovdm: unitsTest.txt:
- Tests that conversions that should be possible on account of CLDR conversion data are indeed possible in the ICU implementation. (In theory this could be used for other implementations too.) Converting anything to anything that is compatible requires a combination of the right conversion factors.
- Can also be used to test precision: the ICU implementation is not at "arbitrary precision" at the moment, we can increase test accuracy and use these tests to ensure we are doing arbitrary precision. (In the constants used in these conversions, I think it's only "PI" and "G" that are approximate, the rest should be accurate to the appropriate number of decimal digits.)

unitPreferencesTest.txt:
An integration test that confirms we're also finding the right localized preferences for the specified usages.

As new units, conversion rates, and unit preferences are added to CLDR, adding them to the CLDR-defined data-driven test suite can check, end-to-end, that the desired output is produced.

tools/cldr/cldr-to-icu/build-icu-data.xml
hugovdm: Thanks, good point - done!

tools/cldr/cldr-to-icu/build-icu-data.xml
hugovdm: Done.

tools/cldr/cldr-to-icu/build-icu-data.xml
hugovdm: Done and tested.

hagbard:
Let me know when you've answered my comments and I'll look again.

testdata/cldr-testData/units/_readme.txt
hagbard: Hmm, okay. Seems a bit janky, but there's probably a reason (e.g. visual file ordering in Windows or something).

tools/cldr/cldr-to-icu/build-icu-data.xml
hagbard: So is this "test data" we're talking about "data for unit/regression tests" or something else?

Data for unit/regression tests shouldn't be needed when running the tool normally.
There should be a separate test target of some kind for running tests separately.

hugovdm:
I think all comments have been addressed - I think it is ready to squash again?

hugovdm:
Ah no, found one more discussion thread under way.

tools/cldr/cldr-to-icu/build-icu-data.xml
hugovdm: It is a painful mix of both, so I found myself trying to go both ways:

- If CLDR unit preference data is changed, the tests will change (e.g. we find an incorrect preference and want to change its thresholds, or something else, in CLDR). For this, we want the test data in sync with the tests.

- If new preferences are added though, without impact on existing test cases, and there's a bug in ICU, or something missing, we would perhaps *not* want to update the tests too, as it would cause ICU test breakage upon CLDR data import, which the importer would not want to care about? (An example: suppose we add compound units inside mixed units in CDLR, when in ICU the mixed units code only supports built-in units.)

Desireable action, as I see it currently: file a ticket for the broken tests, and comment them out of the test data files?

@macchiati do you have an opinion on handling the CLDR test data for this kind of ICU-impacting "breaking" change?

hugovdm: I have started an email thread for this, with subject line "ICU PR 1199 and CLDR Test Data". I'm considering this resolved for now, so I'm squashing the commits again.

hugovdm:
Sorry about one more round-trip: I had to add a copyright notice to the README, meaning another squashing action. Squashed.

hugovdm:
I think you reviewed at c5c8cbb, because it looked like you reviewed just before squashing. Is there a way to check this though (after squashing)?

sffc:
I think we should check this in to unblock progress on the units work. If we get more clarity later about a better place to do this step, we can make the change at that time.

hugovdm:
OK, my conclusion is this:
- adding more manual steps right now are not good. There's a proliferation of documents and READMEs which people can't always find. Any google-internal notes someone has on the process is definitely insufficient in my opinion.
- the conceptual fit here is "CLDR data moved into a state where ICU can make use of it". CLDR to ICU tool and build file.
  - Input: It's still coming from cldr-staging/production/common/ , as does all the XML files.
  - Output: we're writing to testDataDir="${basedir}/../../../testdata/...", while other outputs are to outDir="${basedir}/../../../icu4c/source/data/..." and specialsDir="${basedir}/../../../icu4c/source/data/xml".
  - Before I discovered ant has a cross-platform <copy> command, I even considered that I could use Java: could add an "Identity transform" to the cldr-to-icu jar that could copy files that don't need XML-to-ICU-Resource transformations. (Smellier? ;-) )
- Maybe one of these options is an improvement:
  - (A) I can change testDataDir to be relative to outDir, but what about specialsDir then, should that also be relative to outDir?
  - (B) I can file a JIRA ticket for a better automated build process. Or a ticket for somehow teaching ICU4C's Makefile about CLDR directories and have that call ant and whatever else needs to be done. And then I can add a TODO referencing this ticket in build-icu-data.xml - so we can clean this up later.
  - (C) I can submit this with Shane and Mark's approval and insistence that we move on, and stop driving people nuts with indecisiveness?

If I do B, or perhaps A, I need to squash again and need another approval. @hagbard would you like me to do A or B, would you approve it after squash? Or shall I go ahead with C? (I'll let this sit until Monday.) As you also pointed out, I tend to go too far when I see small low-priority broken things, trying to make everything too clean and perfect: from Units' perspectives and the next release's perspective, and the perspective of what I must get done, changing the build process is one of those lower priority broken things. But I'm enough of a stickler for cleanliness that if we want such a ticket (option B above), I'm willing to assign it to myself and later try to work with those who do the build/integration/release process on making it more streamlined/automated and less manual.

I write too much. SGTY, @hagbard ?

hagbard:
The things I still think are problematic in this PR are:

1) There's no update explaining anything the top level documentation in the build-icu-data.xml file. This is a fundamental shift in who should use this file, and when, and that needs to be documented.

2) The project name "Convert" is conceptually wrong and misleading now, that's no longer what the project does.

3) The Ant file name is conceptually wrong and misleading now (it doesn't just "build icu data" anymore).

4) The "all" target should probably not be copying the data, I think that since data copying is a "one off" compared to regenerating the ICU data (or parts of the ICU data) it doesn't make sense to roll-up copying loads of data into a 5 second regeneration of, say, plurals data. A rethink of other targets is needed if we want "all" to copy data, and an update to the existing readme files to change how we tell people to invoke this.

5) None of the command line flags related to controlling what's converted have any impact on changing what's copied, which is weird (and undocumented). Right now if you supply a flag it's used, because "conversion" is all the build file does. If this becomes a generic "do everything" build file, there'll be confusion about which flags get used by what targets.

While we could address all of these and give this build file a much bigger responsibility (effectively rewriting it into a "busy box"), I still feel it would be much cleaner in the long run to just have a parent build file (or makefile) which knows how to invoke this with default arguments and also knows how to copy test data etc. If someone then wants to do "interesting" rebuilds of the data, they can just use the targeted build file for conversion with whatever arguments/config they need.

Otherwise in 12 months we'll probably have ended up stuffing loads more functionality into this Ant file (because it already does several disconnected things, so why not) and make it overly complex to use for the already non-trivial things related to its original core requirements (e.g. customizing alt-paths for Android).

If others want to LGTM this "as is" then fine, but I can't do that. It just doesn't "look good" to me as it stands.
However, I'm also happy to defer to people like Norbert, Peter and Victor who get to use these tools during a release.

hugovdm:
Would this fit the directory? A new ant build file with a new name,
something like this:

tools/cldr/cldr-to-icu/copy-cldr-files.xml

Then the manual instructions change would be minimal:

$ # Prepare Environment Variables
$ ant -f tools/cldr/cldr-to-icu/build-icu-data.xml
$ ant -f tools/cldr/cldr-to-icu/copy-cldr-files.xml

Versus using a different build system in a different directory:

$ # Prepare Environment Variables
$ ant -f tools/cldr/cldr-to-icu/build-icu-data.xml
$ cd testdata ; make

I'll take a look at top-level documentation.

Thanks,
Hugo

hugovdm:
Apologies for this being scattered over multiple threads.

If we go for separate build files, the parent directory makes sense,
something like this:

$ ant -f tools/cldr/import-cldr.xml

Which could then essentially call "ant -f
tools/cldr/cldr-to-icu/build-icu-data.xml" if it's easy for me to figure
out how to do that properly.

hugovdm:
New idea, to be implemented when I'm working again (taking a break from now):

Have the copying taken care of by:
$ ant -f tools/cldr/build.xml

Ideally have this build.xml file also invoke tools/cldr/cldr-to-icu/build-icu-data.xml - if it can transparently handle everything it needs to handle and pass through, then instructions and process need just a small change.

If it doesn't make sense to somehow pass `-D<flag-name>=<flag-value>` through from cldr/build.xml to cldr-to-icu/build-icu-data.xml, then the README.md in that directory (to be added) will give instructions on how to narrow things down like that, for special needs.

sffc:
`tools/cldr/import-cldr.xml` that copies the test data and also delegates to `tools/cldr/cldr-to-icu/build-icu-data.xml` sounds fine to me.  Does that work for you @hagbard?

hagbard:
build.xml is the idiomatic name for a default Ant build file (which is why I deliberately didn't use it in the cldr-to-icu directory, I'm only using Ant there as a means to capture config information for the conversion process, and I'd use something else if I could, since Ant is so messy).

However, I've got no strong opinions on anything above the /cldr-to-icu directory. It could even be a Makefile that calls down or something else entirely. I don't mind.

For now, if Hugo wants to submit this as-is with your approval then I've got no problem with that. I just cannot in all good faith say that I think it's the right long term thing to be doing.

For Hugo's benefit with Ant, calling Ant sub-tasks probably wants:
  https://ant.apache.org/manual/Tasks/subant.html
There's also "antcall", but I think that's only for tasks in the same build file.

sffc:
I'd say if Hugo can spend 40 minutes to get a new build.xml file in tools/cldr to delegate successfully to the XML file in the child directory, then do it.  If it will take more than 40 minutes, I don't think the status quo is bad enough to justify the additional time investment, since we have a lot to do before ICU 68.

hagbard:
We should raise a ticket to do it after the release in the latter case I think.

hugovdm:
I believe it will be less than 30 minutes - I trust ant subtasks and ant
documentation. :) I'll do it on Monday at the latest. Thanks!

markusicu:
About the path "testdata/cldr-testData/..." -- this was probably discussed elsewhere...
1. It seems odd to have both a dash separator in the name, and a camelCase 'D' later. Seems prone to typos.
2. It seems redundant with "testdata" twice in the path.

Why not just "testdata/cldr/..."?

hugovdm:
> About the path "testdata/cldr-testData/..." -- this was probably discussed elsewhere...
> 
> 1. It seems odd to have both a dash separator in the name, and a camelCase 'D' later. Seems prone to typos.
> 2. It seems redundant with "testdata" twice in the path.
> 
> Why not just "testdata/cldr/..."?

Done. The original name was a pretty arbitrary decision and I was hoping someone would raise an opinion. :-D (I still have a description for the rationale in the PR description, will change that later.)

Just did the basics first, now looking at "subant" and trying to be flexible with it, will indeed make this take a bit longer than 30 minutes if I'm very thorough.

hugovdm:
Sorry, pausing this PR for a little bit: given stress from other things I was supposed to have finished by this evening (i.e. in the next 2 hours), I can't productively focus on English right now. (So my focus shifts back to C++ until I can relax better.) There are anyway source-bundling concerns that need to also be resolved before the results of this PR can be made use of...

At the moment it works, the "-D<name>=<value>" flags get passed through correclty too, but looking over the README requires a bunch of head-scratching about what else could be moved into the new build.xml file and where the delineation needs to be.

hugovdm:
FYI: Given the work needed to make everything consistent, we may automate
the copying of testdata we need, but not yet to a common testdata
directory: I think we must not do half a job on such a move, and we have
insufficient time (too much work), and thus we rather punt the common
testdata directory to the next release.

That said, agreed on testdata/cldr/, I made that change in my previous
proposed PR, and will resume that next quarter.

Thanks,
Hugo

hugovdm:
Updated the now-existing build.xml file to do the copying to two destinations (one in ICU4C, one in ICU4J).

Wrote a short README.md in the tools/cldr directory.

PTAL? (I'm not squashing yet, in case someone likes seeing the list of commits that describe what I'm doing.)

tools/cldr/build.xml
```diff
@@ -39,9 +39,31 @@
         </condition>
         <fail unless="cldrDir"
              message="Set the CLDR_DIR environment variable (or cldrDir property) to the top level CLDR source directory (containing 'common')."/>
+        <property name="testDataDir4C" value="${basedir}/../../icu4c/source/test/testdata/cldr"/>
+        <property name="testDataDir4J" value="${basedir}/../../icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr"/>
     </target>
 
-    <!-- TODO: Add properties and targets for copying test data. -->
+    <!-- Copies CLDR test data -->
```
hagbard, line 46: Maybe say:
    
`<!-- Copies CLDR test data directories (deleting previous contents first to prevent inconsistent state) -->`

Ant uses timestamps and won't copy "old" files over "new" ones, but this is prone to all sorts of fun, so it's very important that the directory is "cleaned" first. The alternative is using the "overwrite" flag, but you can still leave old versions around with that. So what you're doing is definitely the only sensible approach, it's just that it takes a bit of thinking to appreciate that.

It also makes the comment "Add directories here..." important since you don't want an explicit list of files, because when that changes you'd potentially leave old files un-deleted.

Your directory-at-a-time approach is safe and good, and I think the only reasonable solution, so maybe change:
`<!-- The set of CLDR test data files to be copied into ICU. -->`
to:
`<!-- CLDR test data directories to be copied into ICU (do not list individual files here). -->`

hagbard (approved):

tools/cldr/build.xml
hugovdm: Done. I've changed the README too, to emphasize directories / de-emphasize files.

However are directories that much better? If someone removes a directory from the list, same problem would occur, it wouldn't get deleted? So I'm emphasizing directories but not banning files... does this seem reasonable?

(Not squashing yet, for ease of reviewing the latest changes.)

hagbard (approved):

hugovdm:
Squashing!

hagbard (approved):

hugovdm:
> Should `ant clean` depend on `clean-icu-data`?
> 
> Should we update icu4c/data/cldr-readme.txt?

I think since, at this time, `ant` defaults to `ant install` which doesn't touch `build-icu-data`, `ant clean` should not touch `clean-icu-data`.

I think since this PR only touches source/test/testdata and not source/data, that source/data/cldr-icu-readme.txt is not impacted by this.

I think perhaps I should add a README in icu4c/source/test/testdata/cldr/ though? And perhaps also in icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/?

markusicu:
> > Should we update icu4c/data/cldr-readme.txt?
> 
> I think since, at this time, `ant` defaults to `ant install` which doesn't touch `build-icu-data`, `ant clean` should not touch `clean-icu-data`.
> 
> I think since this PR only touches source/test/testdata and not source/data, that source/data/cldr-icu-readme.txt is not impacted by this.
> 
> I think perhaps I should add a README in icu4c/source/test/testdata/cldr/ though? And perhaps also in icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/?

https://github.com/unicode-org/icu/blob/master/icu4c/source/data/cldr-icu-readme.txt covers the whole ICU data generation and already mentions testdata. Let's keep it together.

sffc:
In particular, cldr-icu-readme.txt says

```
cd $TOOLS_ROOT/cldr/cldr-to-icu
ant -f build-icu-data.xml -DcldrDataDir="$CLDR_TMP_DIR/production" | tee /tmp/cldr-newData-builddataLog.txt
```

You should change that to also run your test data copy routine.

hugovdm:
Thanks for pushing on this point: I do deeply appreciate good doc-updating practices.

I'm now tweaking the `./install-cldr-jars.sh` step, replacing it by simply an "ant" invocation in the tools/cldr dir. This takes care of invoking install-cldr-jars.sh as well as copying the testData files.

I'm also actually including the hereby copied testData files in this PR. I think that makes sense.

hugovdm:
Sorry, missed a "-a" on a commit. No diffs relative to Shane's latest review.

- sffc reviewed 8b0d23d.
- Commit 9731154 was a partial squash (missed the earlier merge commit).
- Commit a564467c28 is a proper squashing.
- No diffs:
  `$ git diff 8b0d23d 9731154fbd`
  `$ git diff 9731154fbd a564467c28`

See the commit history from 8b0d23d if you'd like to examine individual commits.

Or just look at the noteworthy change since @hagbard last reviewed/approved: icu4c/source/data/cldr-icu-readme.txt

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -135,12 +135,14 @@ export ICU4C_DIR=$HOME/icu-myfork/icu4c
...
 export TOOLS_ROOT=$HOME/icu-myfork/tools
 
-# 2. Build and install the CLDR jar
+# 2. Build and install the CLDR jar, as well as installing some CLDR
+# testData files needed by ICU4C and ICU4J tests.
 
-cd $TOOLS_ROOT/cldr/lib
-./install-cldr-jars.sh "$CLDR_DIR"
+cd $TOOLS_ROOT/cldr
+ant
```
hagbard, line 142: So this won't run the actual CLDR-to-ICU tools at this point.
Since the default action is not "do everything" I'd be tempted to give an explicit target here (e.g. "ant prepare-cldr" or "cldr-ant init"). Default actions are only good when they're the only thing you ever normally need.

icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/units/_readme.txt
```diff
@@ -0,0 +1,6 @@
+# Test data for units: conversions and preferences
```
hagbard, line 1: Won't this get nuked when someone runs the clean target?

icu4c/source/data/cldr-icu-readme.txt
hugovdm: If the default action isn't recommended, should it be disabled somehow? (Is that possible?)

Is "ant install" appropriate? I've gone with that, removing one step of "default" indirection. But perhaps, considering "clean" and "install"'s usual meaning, maybe you don't want this as "install"?

```
<project name="CLDR" **default="install"** basedir=".">
[...]
    <target name="install" depends="clean, copy-cldr-testdata, **install-cldr-libs**"/>
[...]
```

icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/units/_readme.txt
hugovdm: Yes. So if I read you right, you'd like "clean" to leave the repository in the desired commit state. Do we need a "clean" target at all presently?

hugovdm: Deleted "clean" and "install" targets for now, they're not useful at this time.

Left `clean-cldr-testdata` target, to me it looks like it behaves similarly to `clean-icu-data`.

icu4c/source/data/cldr-icu-readme.txt
hugovdm: How do you feel about invoking the script, vs invoking the install-cldr-libs ant target (one more indirection step, but a step towards consolidation)?

icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/units/_readme.txt
hagbard: SGTM

icu4c/source/data/cldr-icu-readme.txt
hagbard: If the Ant script works as well (or better) then it seems like a step forward. Using "install"  feels fine for now, that's more a question for Peter et al.

As for default tasks, if there's no way to disable one, we can make a default action which just does a "help page".

hagbard (approved):

markusicu:
Hugo please squash.

PR #1200, opened 2020-07-23
ICU-21158 Fix doc of UDISPCTX_NO_SUBSTITUTE

FrankYFTang:
Add document to mention U_ILLEGAL_ARGUMENT_ERROR would be set to error
code when there are no data for UDISPCTX_NO_SUBSTITUTE.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21158
- [X] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/unicode/udisplaycontext.h
+++ b/icu4c/source/common/unicode/udisplaycontext.h
@@ -156,7 +156,8 @@ enum UDisplayContext {
     UDISPCTX_SUBSTITUTE = (UDISPCTX_TYPE_SUBSTITUTE_HANDLING<<8) + 0,
     /**
      * A possible setting for SUBSTITUTE_HANDLING:
-     * Returns a null value when no data is available.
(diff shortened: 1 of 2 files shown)


FrankYFTang:
PTAL

sffc:
LGTM please squash

The CI is failing on Java on Azure, and this change doesn't touch Java. @jefgen

jefgen:
> The CI is failing on Java on Azure, and this change doesn't touch Java.

There's a thread on icu-core, and so far it looks like it is caused by the recent Java update from version `1.8.0_252` to `1.8.0_262`.
(The failure reproduces elsewhere with the same Java version, so it isn't specific to the build bot setup.)

From the test failure messages, it seems that the tests now have issues detecting/finding the UTF-16BE BOM on files. However, I haven't had any more time to investigate what is going on though, or why the newer version of Java might have changed/broken things...

PR #1202, opened 2020-07-28
ICU-20088 Fixing pages under transforms/general

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20088
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [ ] Tests included
- [x] Documentation is changed or added

PR #1201, opened 2020-07-23
ICU-20088 Userguide migration - cleaning up formatting for some of the documents

sven-oly:
https://unicode-org.atlassian.net/browse/ICU-20088

PR #1205, opened 2020-07-31
ICU-20088 more User Guide md fixes

markusicu:
[ICU-20088] / #919 for the merge to master

# This is not a merge onto master.
DISABLE_JIRA_ISSUE_MATCH=true
ALLOW_MANY_COMMITS=true

[ICU-20088]: https://unicode-org.atlassian.net/browse/ICU-20088


markusicu:
All changes from Craig's PR #1201, rebased onto and merged with the latest migration branch (merged some changes with Shane's PR #1202). I will take another look too.

docs/userguide/strings/properties.md
```diff
@@ -82,100 +90,100 @@ words, there are two ICU properties for the same Unicode property, one
 delivering single values (for per-code point lookup) and the other delivering
 sets of values (for use with value aliases and UnicodeSet).
 
-| UCD Name(see PropertyAliases.txt) | Type |  | ICU4C uchar.hICU4J UCharacter | UCD File (.txt) |
```
sffc, line 85: Why did you remove the fifth column "UCD File (.txt)"?

sffc: I see the comment in the email thread:

> * strings/properties.md. The big properties table is too wide, and the last column is not visible. I intend to change this to a series of short sections.

markusicu: I worked on the table to make it less wide, and ended up with a half-visible last column. Then I decided that it was both tedious and unnecessary to document *on this page* which property has data in which Unicode file. Instead, I add a short paragraph pointing to UAX 44 which has all that and is carefully kept in sync with the properties for each Unicode release.

PR #1207, opened 2020-08-01
ICU-20725 Fix stack overflow of u_unescapeAt

FrankYFTang:
When the u_unescape is called with repeated escaped surrogate lead,
it calls u_unescapeAt and recursively calls u_unescapeAt and
cause stack overflow. Limit the length to the end of the escape of
surrogate tail to avoid repeat recursion calls.

The problem case is a long list of surrogate lead in escaped form follow by each other, which would cause recursion for every one of them.

For example:

\ud800\ud800\ud800 ...... \ud800 (repeat several thousands)

To prevent that, we only recursively call the u_unescapeAt function with max to length 5 (the u and the 4 hex) from the current position after we found the surrogate lead and a '\'  . In this way, we guarantee the stack depth is only 2 .

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20725
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/ustring.cpp
+++ b/icu4c/source/common/ustring.cpp
@@ -1294,7 +1294,15 @@ u_unescapeAt(UNESCAPE_CHAR_AT charAt,
             int32_t ahead = *offset + 1;
             c = charAt(*offset, context);
             if (c == 0x5C /*'\\'*/ && ahead < length) {
-                c = (UChar) u_unescapeAt(charAt, &ahead, length, context);
+                // Calling u_unescapeAt recursively may cause a stack overflow if
+                // we have repeated surrogate lead after that. Limit the
+                // length to 5 ('u' and 4 hex) after ahead.
+                int32_t tailLimit = ahead + 5;
+                if (tailLimit > length) {
+                    tailLimit = length;
+                }
+                c = (UChar) u_unescapeAt(charAt, &ahead, tailLimit,
+                                         context);
             }
             if (U16_IS_TRAIL(c)) {
                 *offset = ahead;
--- a/icu4c/source/test/cintltst/custrtst.c
+++ b/icu4c/source/test/cintltst/custrtst.c
@@ -22,6 +22,7 @@
 #include "unicode/ucnv.h"
 #include "unicode/uiter.h"
 #include "cintltst.h"
+#include "cstring.h"
 #include "cmemory.h"
 #include <string.h>
 
@@ -36,6 +37,7 @@ static void TestStringFunctions(void);
 static void TestStringSearching(void);
 static void TestSurrogateSearching(void);
 static void TestUnescape(void);
+static void TestUnescapeRepeatedSurrogateLead20725(void);
 static void TestCountChar32(void);
 static void TestUCharIterator(void);
 
@@ -48,6 +50,8 @@ void addUStringTest(TestNode** root)
     addTest(root, &TestStringSearching, "tsutil/custrtst/TestStringSearching");
     addTest(root, &TestSurrogateSearching, "tsutil/custrtst/TestSurrogateSearching");
     addTest(root, &TestUnescape, "tsutil/custrtst/TestUnescape");
+    addTest(root, &TestUnescapeRepeatedSurrogateLead20725,
+            "tsutil/custrtst/TestUnescapeRepeatedSurrogateLead20725");
     addTest(root, &TestCountChar32, "tsutil/custrtst/TestCountChar32");
     addTest(root, &TestUCharIterator, "tsutil/custrtst/TestUCharIterator");
 }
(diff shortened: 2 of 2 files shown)


icu4c/source/common/ustring.cpp
```diff
@@ -1294,7 +1294,15 @@ u_unescapeAt(UNESCAPE_CHAR_AT charAt,
             int32_t ahead = *offset + 1;
             c = charAt(*offset, context);
             if (c == 0x5C /*'\\'*/ && ahead < length) {
-                c = (UChar) u_unescapeAt(charAt, &ahead, length, context);
+                // Calling u_unescapeAt recursively may cause stack overlow if
```
markusicu, line 1297:
```suggestion
                // Calling u_unescapeAt recursively may cause a stack overflow if
```

icu4c/source/common/ustring.cpp
```diff
@@ -1294,7 +1294,15 @@ u_unescapeAt(UNESCAPE_CHAR_AT charAt,
...
+                // we have repeated surrogate lead after that. Limit the
+                // length to 5 ('u' and 4 hex) after ahead.
+                int32_t recursive_length = ahead + 5;
```
markusicu, line 1300: it's not a length, it's a position, right? therefore, i would call it something with "limit" in the name. how about "trailLimit"? (we camelCase variable names)

icu4c/source/test/cintltst/custrtst.c
```diff
@@ -1124,6 +1128,32 @@ TestUnescape() {
...
+        log_err("failure in u_unescape()");
+    }
+    for (int32_t i = 0; i < repeat; i++) {
+      if (dest[i] != 0xd841) {
+        log_err("failure in u_unescape() return value");
+      }
+    }
+    free(src);
+    free(dest);
+
```
markusicu, line 1154: Please add a few simple test cases to make sure that the code recovers properly. For example:
*   `"\\ud841\\x5A"` --> `u"\ud841Z"`
*   `"\\ud841\\U00050005"` --> `u"\ud841\U00050005"`
*   `"\\ud841\\xXX"` --> `u"\ud841\\xXX"`

The last one is ill-formed. I am just guessing at the output for that one.
Hopefully UTF-16 string literals -- u"..." -- work in C test code on our platforms.

FrankYFTang: ill-formed u"" literal does not work in C, clang give me a "invalid universal character" error
```
custrtst.c:1155:26: error: invalid universal character
    if (u_strcmp(dest, u"\ud841Z")!=0) {
                         ^~~~~~
```

FrankYFTang:
PTAL

icu4c/source/common/ustring.cpp
```diff
@@ -1294,7 +1294,15 @@ u_unescapeAt(UNESCAPE_CHAR_AT charAt,
             int32_t ahead = *offset + 1;
             c = charAt(*offset, context);
             if (c == 0x5C /*'\\'*/ && ahead < length) {
-                c = (UChar) u_unescapeAt(charAt, &ahead, length, context);
+                // Calling u_unescapeAt recursively may cause a stack overflow if
+                // we have repeated surrogate lead after that. Limit the
+                // length to 5 ('u' and 4 hex) after ahead.
+                int32_t tailLimit = ahead + 5;
+                if (tailLimit > length) {
+                  tailLimit = length;
```
markusicu, line 1302:
```suggestion
                    tailLimit = length;
```

icu4c/source/test/cintltst/custrtst.c
```diff
@@ -1124,6 +1128,52 @@ TestUnescape() {
...
+        log_err("u_unescape() should return u\"\\ud841Z\" but got %s", dest);
+    }
+
+    u_unescape("\\ud841\\U00050005", dest, repeat);
+    const UChar expected2[] = {0xd841, 0xd900, 0xdc05, 0};
+    if (u_strcmp(dest, expected2)!=0) {
+        log_err("u_unescape() should return u\"\\ud841\\ud900\\udc05\" but got %s", dest);
+    }
+
+    u_unescape("\\ud841\\x41", dest, repeat);
```
markusicu, line 1167: This is basically the same as the `"\\ud841\\x5A"` case above. My suggestion was to test `"\\ud841\\xXX"`, that is, with an ill-formed escape sequence. Please try that. (As I said, I am not sure what the output will be. I am guessing it will be surrogate, backslash, x, X, X.)

FrankYFTang: I see.

FrankYFTang: it return empty string because "If an escape sequence is ill-formed, this method returns an empty string. "
see https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/ustring_8h.html#a7aad56a2a6f7cbe3b2fb802a218d239b

FrankYFTang:
PATL

PR #1211, opened 2020-08-04
ICU-21059 Load simple unit IDs from convertUnits.

hugovdm:
In a discussion with Mark, I think we have consensus that it's fine to pull this from convertUnits for now. It's the cleanest list of simple unit identifiers we have.

Follow-up work should help us trust the convertUnits list to always be complete: I propose I add a test that simply ensures we can parse all of the units that show up in the supplementalData/idValidity/unit/regular" list.

(This was previously reviewed as: https://github.com/icu-units/icu/pull/41)

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21059
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -12,20 +12,24 @@
 // Helpful in toString methods and elsewhere.
 #define UNISTR_FROM_STRING_EXPLICIT
 
-#include "cstring.h"
+#include "charstr.h"
+#include "cmemory.h"
 #include "measunit_impl.h"
+#include "resource.h"
 #include "uarrsort.h"
 #include "uassert.h"
 #include "ucln_in.h"
 #include "umutex.h"
 #include "unicode/bytestrie.h"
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -110,122 +114,72 @@ const struct SIPrefixStrings {
...
 icu::UInitOnce gUnitExtrasInitOnce = U_INITONCE_INITIALIZER;
 
+// Array of simple unit IDs.
+//
+// The array memory itself is owned by this pointer, but the individual char* in
+// that array point at static memory. (Note that these char* are also returned
+// by SingleUnitImpl::getSimpleUnitID().)
+const char **gSimpleUnits = nullptr;
+
 char *kSerializedUnitExtrasStemTrie = nullptr;
```
sffc, line 176: Optional: this variable should probably have a g suffix instead of k

hugovdm: Done. And I went ahead and squashed.

PR #1215, opened 2020-08-05
ICU-21216 Windows: Ensure directory exists before running rc

madscientist:
When building in parallel it's possible that the rc command is invoked
before other rules get around to creating the output directory.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21216
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/data/Makefile.in
+++ b/icu4c/source/data/Makefile.in
@@ -241,6 +241,7 @@ ifeq ($(ENABLE_SO_VERSION_DATA),1)
 ifeq ($(PKGDATA_MODE),dll)
 SO_VERSION_DATA = $(OUTTMPDIR)/icudata.res
 $(SO_VERSION_DATA) : $(MISCSRCDIR)/icudata.rc
+	mkdir -p $(@D)
 ifeq ($(MSYS_RC_MODE),1)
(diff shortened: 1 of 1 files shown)


madscientist:
It is being created... by some other make target.  In a serial build the order in which targets are run is well-defined and always in the same order so even without a prerequisite relationship you know which targets will run before which other targets.

But when parallel builds are enabled you can't guarantee the order in which targets are run _unless_ they have a prerequisite relationship.  The .rc target doesn't list any target that creates the directory as a prerequisite, and so it's possible that during parallel builds the .rc target will be built before the target that creates the directory.

Just as a point of interest: in my parallel build that failed I did see the directory being created in the output... but it happened after the .rc file failed to build.

sffc:
Okay; most rules have their dependencies set correctly, but I guess this one slipped through.  It's possible that there are more steps that don't have all of the dependencies they need, but it's hard to find out what they are.  This one in particular probably wasn't noticed because there aren't any team members or CI bots that routinely run a multicore build on MSYS.

jefgen:
Echoing Shane's earlier comment, I think the change here seems mostly harmless. If it helps to unblock your build(s), then it feels like we should take it.

Thank you for taking the time to file a ticket and create a PR!

PR #1216, opened 2020-08-05
ICU-21217 Windows: Fix link command for extra/scrptrun

madscientist:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21217
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/extra/scrptrun/Makefile.in
+++ b/icu4c/source/extra/scrptrun/Makefile.in
@@ -33,7 +33,7 @@ ifeq ($(ENABLE_RPATH),YES)
 RPATHLDFLAGS = $(LD_RPATH)$(LD_RPATH_PRE)$(libdir)
 endif
 LDFLAGS = @LDFLAGS@ $(RPATHLDFLAGS)
-LIBS = $(LIBICUUC) @LIBS@ @LIB_M@
+LIBS = $(LIBICUUC) $(LIB_M)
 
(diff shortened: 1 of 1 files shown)


madscientist:
Just wanted to point out that there are a lot of other Makefile.in files that have this same issue (just search for LINK.cc).  Also, I think many makefiles set variables "incorrectly", based on @FOO@ to be replaced at configure time instead of using the $(FOO) variables which are set at make invocation time.

However, this patch was all I needed to get my Windows build to complete so the others, while probably should be cleaned up, apparently don't impact Windows.

jefgen:
> Just wanted to point out that there are a lot of other Makefile.in files that have this same issue (just search for LINK.cc). 
> ...
> However, this patch was all I needed to get my Windows build to complete so the others, while probably should be cleaned up, apparently don't impact Windows.

Thanks! -- Taking a quick look, it seems like most of ones that directly use `-o` instead of `$(OUTOPT)` are in the folders under `icu4c/source/test/perf/*`. FWIW, I'm not sure if the ICU4C perf tests ever been used/tested/etc. on Windows with Cygwin/MSYS, and they also aren't built as part of the regular CI builds either.

So it seems that while it still be a good idea to clean them up at some point, I think that could probably be done as a separate ticket. This PR looks fine as-is for now.

Thank you for creating this PR and for filing a bug!

PR #1220, opened 2020-08-06
ICU-21174 Add a CopyableMaybeStackArray class.

hugovdm:
##### Checklist

- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -297,9 +297,6 @@ public:
      * Automatically allocates the heap array if the argument is larger than the stack capacity.
      * Intended for use when an approximate capacity is known at compile time but the true
      * capacity is not known until runtime.
-     *
-     * WARNING: does not report errors upon memory allocation failure, after
-     * which capacity will be less than expected!
      */
     MaybeStackArray(int32_t newCapacity) : MaybeStackArray() {
         if (capacity < newCapacity) { resize(newCapacity); }
@@ -402,11 +399,11 @@ public:
 
   private:
     /* No comparison operators with other MaybeStackArray's. */
-    bool operator==(const MaybeStackArray & /*other*/) = delete;
-    bool operator!=(const MaybeStackArray & /*other*/) = delete;
+    bool operator==(const MaybeStackArray & /*other*/) {return FALSE;}
+    bool operator!=(const MaybeStackArray & /*other*/) {return TRUE;}
     /* No ownership transfer: No copy constructor, no assignment operator. */
-    MaybeStackArray(const MaybeStackArray & /*other*/) = delete;
-    void operator=(const MaybeStackArray & /*other*/) = delete;
+    MaybeStackArray(const MaybeStackArray & /*other*/) {}
+    void operator=(const MaybeStackArray & /*other*/) {}
 };
 
 /**
@@ -522,7 +519,7 @@ template<typename T, int32_t stackCapacity>
 inline T *MaybeStackArray<T, stackCapacity>::resize(int32_t newCapacity, int32_t length) {
     if(newCapacity>0) {
 #if U_DEBUG && defined(UPRV_MALLOC_COUNT)
-        ::fprintf(::stderr, "MaybeStackArray (resize) alloc %d * %lu\n", newCapacity, sizeof(T));
+      ::fprintf(::stderr,"MaybeStacArray (resize) alloc %d * %lu\n", newCapacity,sizeof(T));
 #endif
         T *p=(T *)uprv_malloc(newCapacity*sizeof(T));
         if(p!=NULL) {
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -387,8 +387,6 @@ public:
   protected:
     T *ptr;
     int32_t capacity;
-
-  private:
     UBool needToRelease;
     T stackArray[stackCapacity];
     void releaseArray() {
@@ -401,9 +399,11 @@ public:
         capacity=stackCapacity;
         needToRelease=FALSE;
     }
+
+  private:
     /* No comparison operators with other MaybeStackArray's. */
-    bool operator==(const MaybeStackArray & /*other*/) {return FALSE;}
-    bool operator!=(const MaybeStackArray & /*other*/) {return TRUE;}
+    bool operator==(const MaybeStackArray & /*other*/) = delete;
+    bool operator!=(const MaybeStackArray & /*other*/) = delete;
     /* No ownership transfer: No copy constructor, no assignment operator. */
     MaybeStackArray(const MaybeStackArray & /*other*/) = delete;
     void operator=(const MaybeStackArray & /*other*/) = delete;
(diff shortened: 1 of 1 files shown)


hugovdm:
I'm also particularly interested how we might handle the allocation errors better - I haven't given it enough thought yet myself. We could document that code calling resize() must check the return value, but it doesn't help with constructors. The main problem is the lack of bounds checking - but what would be the right behaviour *with* bounds checking? Assertion failure?

I could add a U_ASSERT for debug mode bounds checking. It's going to be a rare bug though: only when this malloc call fails, so debug mode isn't going to reliably catch it.

sffc:
I'm not sure about the use case for this.

hugovdm:
Use case is in icu4c/source/i18n/number_microprops.h in this commit:
https://github.com/icu-units/icu/pull/52/commits/8a3b58ff273baad04c8769cca478441d7509bc8a

If you prefer we make this microprops-specific, I could move it out of cmemory.h and into number_microprops.h?

icu4c/source/common/cmemory.h
```diff
@@ -399,8 +405,49 @@ class MaybeStackArray {
...
+    // Assignment operator
+    void operator=(const CopyableMaybeStackArray &rhs) {
+        int32_t length = rhs.capacity;
+        this->resize(length, length);
+        // If memory allocation failed, capacity will be at least stackCapacity.
+        if (length > this->capacity) {
+            length = this->capacity;
+        }
+        uprv_memcpy(rhs.ptr, this->ptr, (size_t)length * sizeof(T));
+    };
```
sffc, line 450: I think you might have to implement the move constructor and move assignment, or at least explicitly default them, to avoid the copy constructor always being used.

icu4c/source/common/cmemory.h
```diff
@@ -399,8 +405,49 @@ class MaybeStackArray {
...
+
+    // Assignment operator
+    void operator=(const CopyableMaybeStackArray &rhs) {
+        int32_t length = rhs.capacity;
+        this->resize(length, length);
+        // If memory allocation failed, capacity will be at least stackCapacity.
+        if (length > this->capacity) {
+            length = this->capacity;
+        }
+        uprv_memcpy(rhs.ptr, this->ptr, (size_t)length * sizeof(T));
```
sffc, line 449: Isn't this backwards? The signature of memcpy is

`memcpy ( void * destination, const void * source, size_t num )`

icu4c/source/common/cmemory.h
```diff
@@ -395,12 +399,85 @@ class MaybeStackArray {
         capacity=stackCapacity;
         needToRelease=FALSE;
     }
+
+  private:
     /* No comparison operators with other MaybeStackArray's. */
-    bool operator==(const MaybeStackArray & /*other*/) {return FALSE;}
-    bool operator!=(const MaybeStackArray & /*other*/) {return TRUE;}
+    bool operator==(const MaybeStackArray & /*other*/) = delete;
```
roubert, line 405: These changes to `MaybeStackArray`are not in any way coupled to the creation of `CopyableMaybeStackArray` and ought to be done in a separate PR, with a commit message that explains them.

icu4c/source/common/cmemory.h
```diff
@@ -435,7 +512,7 @@ template<typename T, int32_t stackCapacity>
 inline T *MaybeStackArray<T, stackCapacity>::resize(int32_t newCapacity, int32_t length) {
     if(newCapacity>0) {
 #if U_DEBUG && defined(UPRV_MALLOC_COUNT)
-      ::fprintf(::stderr,"MaybeStacArray (resize) alloc %d * %lu\n", newCapacity,sizeof(T));
+        ::fprintf(::stderr, "MaybeStackArray (resize) alloc %d * %lu\n", newCapacity, sizeof(T));
```
roubert, line 515: It's not a good idea to do a trivial bugfix to existing code like this in the same commit as other changes for if problems are encountered with those other changes that lead to them being reverted, the trival bugfix will then be reverted too and that wouldn't benefit anyone.

roubert:
> Use case is in icu4c/source/i18n/number_microprops.h in this commit:
> [icu-units@8a3b58f](https://github.com/icu-units/icu/commit/8a3b58ff273baad04c8769cca478441d7509bc8a)

I don't understand why any kind of `MaybeStackArray` would be a suitable data structure to use here.

Would it not be more appropriate to use some kind of normal vector data structure instead, which keeps track of its own length and allocates more memory itself when needed, instead of having to keep the separate `mixedMeasuresCount` and call `resize()` explicitly from the client code?

hugovdm:
Having it keep track of its own size is nice, but I didn't think it necessary. (I see I'm even making it one bigger than necessary already.)

Having it do so automatically is less efficient than telling it in advance how many will be needed (so an explicit "ensureCapacity(N)" call seems worthwhile).

Do we have any such a class in ICU though? Is it worthwhile to add one that tracks its own size? I could adjust this new class to do that... (and figure out a new name) - but do we want to do bounds checking then? With bounds checking, automatic resize to double the capacity could be done, keeping separate conceptions of "length" and "capacity"...

icu4c/source/common/cmemory.h
hugovdm: Good point. How would you track trivial bug fixes with ICU tickets? A ticket per little fix, or set of fixes? Do you feel it is reasonable enough to use ticket ICU-21174, "Improvements to ICU internal tools for units work", as a bucket for these?

icu4c/source/common/cmemory.h
hugovdm: Thanks, fixed!

icu4c/source/common/cmemory.h
hugovdm: Done.

hugovdm: (And manually tested/explored with some printf'ing and a debugger, FWIW. I finally think I'm not confused by std::move anymore - it is as straight-forward as I once thought before I got confused. :)

icu4c/source/common/cmemory.h
hugovdm: Is it reasonable to combine these with the trivial bugfix?

hugovdm: https://github.com/unicode-org/icu/pull/1221

roubert:
To determine what the most suitable data structure to use in `struct MicroProps` would be, I suggest that you start out by assuming that you have every possible data structure readily available while considering which one to choose.

(As you're already proposing to add a new data structure implementation, there's no reason to limit yourself to thinking about what data structures are already available in ICU while making this choice.)

Then, when that question has been answered, we can look at what data structures already exist in ICU and what should be added.

For example, if you come the conclusion that pre-allocating a fixed-size chunk of memory and then keeping track of its size and resizing it when needed would be the kind of data structure best suited for the problem you're trying to solve here, then I would suggest considering doing simply that, calling `uprv_malloc()` and `uprv_realloc()` directly without any intermediate layer.

But if you instead would come to the conclusion that being able to simply add items without concerns about memory management while still having random access into the structure would be a better fit here, then I would suggest using some kind of vector data structure, eg. using the existing `icu::UVector` data structure or adding some new ICU vector data structure with a more idiomatic C++ API.

Or maybe there is a maximum number of possible entries small enough that it'd be easier to just allocate a buffer of this fixed size directly within the struct?

And there are certainly also other aspects worth considering here.

hugovdm:
@sffc A.resize does pointless copying from A's old location to A's new, following which I would overwrite it all with B's content anyway. Trade-off between double-copying and double code?

hugovdm:
@roubert Basic need: a structure which does copy-construction and copy-assignment. So I can wrap a class around malloc and free, to get just that. I noticed MaybeStackArray does almost exactly that, "if only it were also Copyable".

If it feels too half-baked to add like this, I propose I add it directly in number_microprops.h to have a special-case minimal container for now, with a TODO to flesh it out into a fancy and thorough new ICU container type in the future?

sffc:
MaybeStackArray is strictly better than malloc/realloc, because the data
gets to live on the stack.

It's a little odd that MaybeStackArray has a capacity but not a length, but
we have other types in a similar situation, like LocalArray. I think it's a
good idea to make CopyableMaybeStackArray have the same feature set as
MaybeStackArray.

hugovdm:
Do either of you know when U_NOEXCEPT is required? I'm inconsistent, and it seems the rest of this file might also be inconsistent - though maybe there is some rule I just don't recognise yet (under which one sees the consistency).

sffc:
> Do either of you know when U_NOEXCEPT is required? I'm inconsistent, and it seems the rest of this file might also be inconsistent - though maybe there is some rule I just don't recognise yet (under which one sees the consistency).

It's required for all move constructors and move assignment operators, and not generally anywhere else.

hugovdm:
Taking a look, the U_NOEXCEPT's seem right to me, and in this case if they're missing they cause compilation failures anyway.

There are no cmemory.h-specific unit tests yet. I've started writing some - I can add them to this PR later, maybe after #1222 is merged, so that I can simply cover all the functionality in one go?

hugovdm:
Status: discussed with @sffc the quirky behaviour resulting from my trying to emulate MaybeStackArray's quirky (and not entirely safe) behaviour. Still to consider: if we can do better by not following MaybeStackArray's example.

hugovdm:
We have decided to drop this from cmemory.h and just do something simple as needed by our units-related code.

(And at some later time, I'll try to review usages of MaybeStackArray's int32_t accepting constructor to see if their error handling should be improved.)

sffc:
> (And at some later time, I'll try to review usages of MaybeStackArray's int32_t accepting constructor to see if their error handling should be improved.)

Can you file a follow-up ticket, either in the icu-units repo or in Jira?

hugovdm:
Filed a Jira ticket for it: https://unicode-org.atlassian.net/browse/ICU-21230

PR #602, opened 2019-04-06
ICU-20536 Japanese era Reiwa support in ICU 4.8

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20536
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [x] Tests included
- [ ] Documentation is changed or added

ALLOW_MANY_COMMITS=true

PR #1221, opened 2020-08-07
ICU-21174 Delete not-to-be-used methods. Add a constructor warning. Fix a debug msg.

hugovdm:
##### Checklist

- [X] Issue accepted
- [ ] Tests included

ALLOW_MANY_COMMITS=true


hugovdm:
(Two commits so I could easily split them up into separate PRs if that's better?)

roubert:
> (Two commits so I could easily split them up into separate PRs if that's better?)

I think it's good to have this as two separate commits, as one of them doesn't change behaviour in any way while the other one does (and therefore possibly could end up being reverted).

To tell GitHub that this PR consists of more than one commit intentionally, you can add this magic string to the PR description:

ALLOW_MANY_COMMITS=true

See here for how I know about that feature:

https://github.com/unicode-org/icu/pull/602#issuecomment-481916856

hugovdm:
Ooh, neat! "With great power..." I've been wanting something like this, nice to know there's already a "proper way to do it".

PR #1222, opened 2020-08-07
ICU-21174 Add "AndCheckError" versions of "emplaceBack()" and "create()"

hugovdm:
I intend to squash this and attribute the whole PR to @younies, with his approval - as I really just copied his code to move it one step up the class hierarchy.

##### Checklist

- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -728,6 +728,18 @@ public:
         return fPool[fCount++] = new T(std::forward<Args>(args)...);
     }
 
+    template <typename... Args>
+    T* createAndCheckErrorCode(UErrorCode &status, Args &&... args) {
+        if (U_FAILURE(status)) {
+            return nullptr;
+        }
+        T *pointer = this->create(args...);
+        if (U_SUCCESS(status) && pointer == nullptr) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+        }
+        return pointer;
+    }
+
     /**
      * @return Number of elements that have been allocated.
      */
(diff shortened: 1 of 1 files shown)


icu4c/source/common/cmemory.h
```diff
@@ -771,6 +783,19 @@ class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
         return this->create(args...);
     }
 
+    template <typename... Args>
+    T *emplaceBackAndCheckErrorCode(UErrorCode &status, Args &&... args) {
+        if (U_FAILURE(status)) {
```
sffc, line 788: Optional: Down here you can probably get away with

```
return this->createAndCheckErrorCode(status, args...);
```

hugovdm: Done. I intend to squash into the first commit, leaving it attributed to @younies

sffc: I think you changed it in the wrong place.

Is there a unit test dedicated to these classes? If there is, you should add a unit test.

hugovdm: Oh great. Partial attention bites again. :-( Sorry!

There are no direct tests for MemoryPool or MaybeStackVector yet. What I had failed to do was merge into the PR which uses it and run that.

hugovdm: I see the trie tests are also under test/intltest/ - so it seems the right location for tests for source/common/.

hugovdm: (I wonder why it is called "cmemory" anyway, when it's C++ code?)

hugovdm:
- Tested with my local version of https://github.com/icu-units/icu/pull/52
- Squashed
- Half way through adding some unit tests, but I think there are good reasons for adding them as part of https://github.com/unicode-org/icu/pull/1220 instead of this PR.

So: post-squash, one more approve needed, then I can merge this, and update our development branches.

PR #1223, opened 2020-08-10
ICU-20808 Add test for the C locale to default to en_US_POSIX

erik0686:
##### Checklist

This PR adds a test to make sure that the `C` and `C.UTF-8` LANG setting will cause ICU to have `en_US_POSIX` as the default locale. It seems that in some previous versions this was changed and we now want to make sure that it remains the same or a test fails to let us know it changed so we can investigate further. 

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20808
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -37,6 +37,12 @@ jobs:
       env:
         CC: clang
         CXX: clang++
+    - script: |
+        cd icu4c/source/test/cintltst && LANG=C LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./cintltst /tsutil/cloctst/TestCDefaultLocale
+      displayName: 'Test C Default locale'
+    - script: |
+        cd icu4c/source/test/cintltst && LANG=C.UTF-8 LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./cintltst /tsutil/cloctst/TestCDefaultLocale
+      displayName: 'Test C.UTF-8 Default locale'
 #-------------------------------------------------------------------------
 - job: ICU4C_Clang_Ubuntu_1604_WarningsAsErrors
   displayName: 'C: Linux Clang WarningsAsErrors (Ubuntu 16.04)'
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -261,6 +261,7 @@ void addLocaleTest(TestNode** root)
     TESTCASE(TestToLanguageTag);
     TESTCASE(TestBug20132);
     TESTCASE(TestBug20149);
+    TESTCASE(TestCDefaultLocale);
     TESTCASE(TestForLanguageTag);
     TESTCASE(TestLangAndRegionCanonicalize);
     TESTCASE(TestTrailingNull);
@@ -408,6 +409,10 @@ static void TestNullDefault() {
             log_err("Failure returned from uloc_setDefault - \"%s\"\n", u_errorName(status));
         }
         
+    }
+    uloc_setDefault(original, &status);
+    if (U_FAILURE(status)) {
+        log_err("Failed to change the default locale back to %s\n", original);
     }
     
 }
@@ -6812,5 +6817,19 @@ static void TestUsingDefaultWarning() {
         u_UCharsToChars(buff, errorOutputBuff, length+1);
         log_err("ERROR: in uloc_getDisplayKeywordValue %s %s return len:%d %s with status %d %s\n",
                 keyword_value, keyword, length, errorOutputBuff, status, myErrorName(status));
+      }
+}      
+// Test case for ICU-20575
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that both C and C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL) {
+      log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
+      return;
+    }
+    if ((strcmp(env_var, "C") == 0 || strcmp(env_var, "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
+      log_err("The default locale for LANG=%s should be en_US_POSIX, not %s\n", env_var, defaultLocale);
     }
 }
--- a/icu4c/source/test/cintltst/cloctst.h
+++ b/icu4c/source/test/cintltst/cloctst.h
@@ -136,6 +136,8 @@ static void TestToLegacyKey(void);
 static void TestToUnicodeLocaleType(void);
 static void TestToLegacyType(void);
 static void TestBug20149(void);
+static void TestCDefaultLocale(void);
+
 
 /**
  * U_USING_DEFAULT_WARNING


icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,11 @@ static void TestBug20149() {
         log_err("ERROR: in uloc_forLanguageTag %s return %s\n", myErrorName(status), locale);
     }
 }
+
+// For the C locale, ICU should have a default locale of en_US_POSIX
+static void TestCLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    if ( (strcmp(getenv("LANG"), "C") == 0 || strcmp(getenv("LANG"), "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
```
jefgen, line 6750: `getenv("LANG")` can return `NULL` if the variable isn't set.
We should only call `getenv("LANG")` once, and assign it to variable, then check if that is NULL first.
Reusing the variable also means we don't call `getenv("LANG")` up to 3 times, which is wasteful.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,11 @@ static void TestBug20149() {
         log_err("ERROR: in uloc_forLanguageTag %s return %s\n", myErrorName(status), locale);
     }
 }
+
+// For the C locale, ICU should have a default locale of en_US_POSIX
```
jefgen, line 6746: Please mention the previous ticket number here (ICU-20575) as a breadcumb for others.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -259,6 +259,7 @@ void addLocaleTest(TestNode** root)
     TESTCASE(TestToLanguageTag);
     TESTCASE(TestBug20132);
     TESTCASE(TestBug20149);
+    TESTCASE(TestCLocale);
```
jefgen, line 262: nit: I'd probably suggest mentioning the word "default" in the test case name maybe.

icu4c/source/test/cintltst/cloctst.c
jefgen: Please also add some comments that this test checks if the environment variable LANG is set, and if so ensures that C/C.UTF-8 cause ICU's default locale to be `en_US_POSIX`.

jefgen:
Thanks for creating this PR!

A few other comments/notes:
- All commits must be prefixed with the ICU ticket number (`ICU-20808`).
- All commits will need to be squashed in your fork before merging. However, for just getting review/feedback on a changes it's fine to leave this until later on after the changes have been approved.
- There's also some more info on the page here: http://site.icu-project.org/repository/gitdev
- Please uncheck the "Documentation" entry in the PR description, as this PR doesn't effect the public API docs, or the user guide, etc.
- We'll also want to add/modify one of CI pipelines to run this new test case with the `LANG=C` environment variable set.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,17 @@ static void TestBug20149() {
...
 }
+
+// Test case for ICU-20575
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that C/C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL){
+      log_err( "Enviromental variable \'LANG\' not set");
```
jefgen, line 6753: If the environment variable isn't set, then it shouldn't be an error.
Otherwise, this will make all the tests fail in the CI pipeline (which is what they are doing now. :) ).

Instead, if it isn't set, then we should just skip the test.
You could add a log message with `log_verbose` to say something like:
```c
   log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
```

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,17 @@ static void TestBug20149() {
...
+    }
+    if ( (strcmp(env_var, "C") == 0 || strcmp(env_var, "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
```
jefgen, line 6755: If the `env_var` is NULL, then we shouldn't be calling `strcmp` on it.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,17 @@ static void TestBug20149() {
         log_err("ERROR: in uloc_forLanguageTag %s return %s\n", myErrorName(status), locale);
     }
 }
+
+// Test case for ICU-20575
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that C/C.UTF-8 cause ICU's default locale to be en_US_POSIX.
```
jefgen, line 6748: nit: I'd probably suggest making it a bit more explicit:
"ensures that C/C.UTF-8" --> "ensures that both "C" and "C.UTF-8" cause"

.ci-builds/.azure-valgrind.yml
```diff
@@ -61,6 +61,10 @@ jobs:
         cd icu4c/source/test/cintltst && LD_LIBRARY_PATH=../../lib:../../stubdata:../../tools/ctestfw:$LD_LIBRARY_PATH valgrind --tool=memcheck --error-exitcode=1 --leak-check=full --show-reachable=yes ./cintltst
       displayName: 'Valgrind cintltst'
       timeoutInMinutes: 15
+    - script: |
```
jefgen, line 64: The valgrind CI pipeline doesn't run on normal PRs and is really only for testing with Valgrind.
I'd suggest adding this to the "regular" CI pipeline instead.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,17 @@ static void TestBug20149() {
...
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that both C and C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL){
+      log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
+    }
+    else if( (strcmp(env_var, "C") == 0 || strcmp(env_var, "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
+      log_err( "The default locale for LANG=C should be en-US_POSIX, not %s\n", defaultLocale );
```
jefgen, line 6756: Hm, can we output the LANG variable here to check what it is?
(Right now this hard-codes LANG=C, but why not LANG=%s ?)

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6742,17 @@ static void TestBug20149() {
...
 }
+
+// Test case for ICU-20575
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that both C and C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL){
+      log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
```
jefgen, line 6753: It think it might be cleaner to just add a `return` here, and remove the `else if`.

jefgen:
It looks like there is a test case in the same file named `TestNullDefault` that changes the default locale to `en_US`. 

I wonder if this might be causing the CI builds to fail. However, I'm not sure why it only effects the Debug build though...

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6746,18 @@ static void TestBug20149() {
...
     }
 }
+
+// Test case for ICU-20575
+// This test checks if the environment variable LANG is set, 
+// and if so ensures that both C and C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL){
```
jefgen, line 6756: I think the ICU coding style is to have a space after the closing parenthesis for an if statement.
So this should be: `if (env_var == NULL) {`

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6746,18 @@ static void TestBug20149() {
...
+      log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
+      return;
+    }
+    if( (strcmp(env_var, "C") == 0 || strcmp(env_var, "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
```
jefgen, line 6760: coding style: space after `if` and no space between the two parens
`if( (strcmp` --> `if ((strcmp`

.ci-builds/.azure-pipelines.yml
```diff
@@ -37,6 +37,9 @@ jobs:
       env:
         CC: clang
         CXX: clang++
+    - script: |
+        cd icu4c/source/test/cintltst && LANG=C LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./cintltst /tsutil/cloctst/TestCDefaultLocale
```
jefgen, line 41: I think it might be good to test for both cases -- so let's add another script entry for LANG=C.UTF-8 and also change the displayName too.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6741,3 +6746,18 @@ static void TestBug20149() {
...
+// and if so ensures that both C and C.UTF-8 cause ICU's default locale to be en_US_POSIX.
+static void TestCDefaultLocale(){
+    const char *defaultLocale = uloc_getDefault();
+    char *env_var = getenv("LANG");
+    if (env_var == NULL) {
+      log_verbose("Skipping TestCDefaultLocale test, as the LANG variable is not set.");
+      return;
+    }
+    if ((strcmp(env_var, "C") == 0 || strcmp(env_var, "C.UTF-8") == 0) && strcmp(defaultLocale, "en_US_POSIX") != 0) {
+      log_err( "The default locale for LANG=%s should be en-US_POSIX, not %s\n", env_var, defaultLocale );
```
jefgen, line 6761: a few more nits...
- no space between the opening paren and the double quote: `log_err( "` --> `log_err("`
- no space at the end closing paren.
- the locale name should be `en_US_POSIX`

jefgen:
Note: The failing Travis CI build is due to a regression caused by the change in PR #1235.
See the comment I added here: https://github.com/unicode-org/icu/pull/1235#issuecomment-675628832

jefgen:
Created a PR to fix the leaks here: PR #1242

jefgen:
Hey @erik0686, it looks like there's a merge conflict in the file `icu4c/source/test/cintltst/cloctst.c` which prevents rebasing and merging the PR.

It looks like for some reason my change to fix the failing Travis CI build in PR 1242 is showing as a conflict on the last few lines of the file. (It's a bit odd since there really shouldn't have been any conflict though... maybe the git diff is confused somehow...)

Can you rebase your changes on top of the latest from `master` and then force-push? This should also help to fix the Travis CI build on the PR too. ;)

PR #1229, opened 2020-08-13
ICU-21212 full range check for Punycode digits

markusicu:
Since I had to add another range check, I changed the code from using a table for Latin-1 code points to have an all-range-checks function more like [decode_digits in the RFC sample code](https://tools.ietf.org/html/rfc3492#page-25), but with a hardcoded binary search.

https://unicode-org.atlassian.net/browse/ICU-21212

### DIFF
--- a/icu4c/source/test/intltest/uts46test.cpp
+++ b/icu4c/source/test/intltest/uts46test.cpp
@@ -39,6 +39,7 @@ public:
     void runIndexedTest(int32_t index, UBool exec, const char *&name, char *par=NULL);
     void TestAPI();
     void TestNotSTD3();
+    void TestInvalidPunycodeDigits();
     void TestSomeCases();
     void IdnaTest();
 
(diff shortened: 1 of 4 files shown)


icu4c/source/test/intltest/uts46test.cpp
```diff
@@ -245,6 +247,20 @@ void UTS46Test::TestNotSTD3() {
...
+    if(errorCode.isFailure()) {
+        return;
+    }
+    UnicodeString input(u"xn--pleѐ");  // ends with non-ASCII U+0450
+    UnicodeString result;
+    IDNAInfo info;
+    idna->nameToUnicode(input, result, info, errorCode);
+    assertTrue("nameToUnicode() should detect non-ASCII",
+               (info.getErrors()&UIDNA_ERROR_PUNYCODE)!=0);
+}
```
FrankYFTang, line 262: since some of the ASCII which would caused basicToDigit[ch] to get -1 will now get value < -1, could we add some test for those chars? such as "]", "[" and ">" "<". I know from the current code it is only used in one place right after that but I think adding some test of that which assert it will fail will be nice. Something like u"xn--ple[" and u"xn--p<le"

markusicu: Good idea, done. Will force-push an amended commit.

PR #1230, opened 2020-08-14
ICU-20164 Make NoUnit a zero-cost abstraction over MeasureUnit.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20164
- [ ] Issue accepted
- [x] Tests included


icu4c/source/i18n/unicode/nounit.h
```diff
@@ -32,77 +32,43 @@ U_NAMESPACE_BEGIN
  * @see NumberFormatter
  * @draft ICU 60
  */
-class U_I18N_API NoUnit: public MeasureUnit {
-public:
+namespace NoUnit {
```
markusicu, line 35: lgtm but enough of a change to warrant an API proposal

icu4j/main/classes/core/src/com/ibm/icu/util/NoUnit.java
```diff
@@ -10,44 +10,34 @@
...
  * @provisional This API might change or be removed in a future release.
  */
-public class NoUnit extends MeasureUnit {
-    private static final long serialVersionUID = 2467174286237024095L;
-
+public final class NoUnit {
     /**
      * Constant for the base unit (dimensionless and no scaling).
      *
+     * Since ICU 68, this constant equals null.
```
markusicu, line 17: A change like this should get the API moved to `@draft ICU 68`.

For PERCENT and PERMILLE I guess it's debatable.

icu4c/source/i18n/unicode/nounit.h
```diff
@@ -32,77 +32,43 @@ U_NAMESPACE_BEGIN
...
     /**
      * Returns an instance for the base unit (dimensionless and no scaling).
      *
-     * @return               a NoUnit instance
+     * Since ICU 68, this function returns the same value as the default MeasureUnit constructor.
```
markusicu, line 39: See comment in the Java version: I think at least this one should go to `@draft ICU 68`.

icu4c/source/i18n/nounit.cpp
```diff
@@ -8,33 +8,6 @@
 
 U_NAMESPACE_BEGIN
```
markusicu, line 9: why not delete this whole file?

icu4c/source/i18n/unicode/nounit.h
sffc: I sent the API proposal this morning.

icu4j/main/classes/core/src/com/ibm/icu/util/NoUnit.java
sffc: I have an open question about this in the ICU 68 API Proposal Status doc.

icu4c/source/i18n/unicode/nounit.h
```diff
@@ -32,77 +32,43 @@ U_NAMESPACE_BEGIN
...
+     *
+     * @return               a MeasureUnit instance
      * @draft ICU 60
      */
-    static NoUnit U_EXPORT2 base();
+    static inline MeasureUnit U_EXPORT2 base() {
```
markusicu, line 44: The class had U_I18N_API but the stand-alone functions only have U_EXPORT2. Sufficient? @jefgen

markusicu:
would be good to finish by sep09 feature freeze

icu4c/source/i18n/unicode/nounit.h
markusicu: I was looking for another non-class-member function and found this in unistr.h, outside of the UnicodeString class definition but still inside the icu namespace:
```
U_COMMON_API UnicodeString U_EXPORT2
operator+ (const UnicodeString &s1, const UnicodeString &s2);
```
So please add U_I18N_API to the front of these function definitions. Unsure exactly where, I guess between "inline" and the return type.

icu4j/main/classes/core/src/com/ibm/icu/util/NoUnit.java
```diff
@@ -7,47 +7,43 @@
...
  * Dimensionless unit for percent and permille.
  * @see NumberFormatter
- * @draft ICU 60
+ * @draft ICU 68
  * @provisional This API might change or be removed in a future release.
  */
-public class NoUnit extends MeasureUnit {
-    private static final long serialVersionUID = 2467174286237024095L;
-
+public final class NoUnit {
```
markusicu, line 13: I believe that this class is still trivially instantiable. (You could experiment with that.)
If so, please add a private, empty default constructor.

sffc: Done.

icu4c/source/i18n/unicode/nounit.h
jefgen: Sorry, I missed this comment/ping before.

> The class had U_I18N_API but the stand-alone functions only have U_EXPORT2. Sufficient? @jefgen

FWIW, since this is now a "header-only" implementation I don't *think* the `U_I18N_API` macro is really needed for it.

The `U_I18N_API` macro marks things for dllexport, and before it was needed since the actual implementation for the functions was inside the .cpp file, so the translation unit would need them to be marked exported. The implementation is now only in the header, so anyone including the header already has access to the full implementation, and there's nothing to export.

For the UnicodeString `operator+`, it looks like implementation of the function is actually in the `unistr.cpp` file, so in that case it needs to be marked exported, which is likely why the header has `U_I18N_API` on it.

The `U_EXPORT2` macro only adds the `__cdecl` calling convention, but I'm less sure about that though.
Since this is header-only, I would think that it might not really matter much, as the consumers of the header would be calling the `MeasureUnit` functions directly.

markusicu:
Jeff, thanks for chiming in. From my side, Shane could squash now for final approval.

PR #1232, opened 2020-08-14
ICU-20774 Add option to icupkg to allow missing deps between resource files.

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20774
- [ ] Issue accepted
- [ ] Tests included


icu4c/source/tools/icupkg/icupkg.cpp
```diff
@@ -198,6 +203,8 @@ static UOption options[]={
 
     UOPTION_DEF("matchmode", 'm', UOPT_REQUIRES_ARG),
 
+    UOPTION_DEF("ignore-deps", '\1', UOPT_NO_ARG),
```
FrankYFTang, line 206: sorry, what does '\1' means here?

sffc: I copied it from elsewhere. I think it means "no short form for this flag".

PR #1233, opened 2020-08-14
ICU-20826 Updating C++ ListFormatter impl to be more like Java

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20826
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/simpleformatter.cpp
+++ b/icu4c/source/common/simpleformatter.cpp
@@ -263,6 +263,8 @@ UnicodeString SimpleFormatter::getTextWithNoArguments(
             sb.append(compiledPattern + i, n);
             i += n;
         } else if (n < offsetsLength) {
+            // TODO(ICU-20406): This does not distinguish between "{0}{1}" and "{1}{0}".
+            // Consider removing this function and replacing it with an iterator interface.
             offsets[n] = sb.length();
         }
     }
--- a/icu4c/source/i18n/formattedval_impl.h
+++ b/icu4c/source/i18n/formattedval_impl.h
@@ -69,6 +69,9 @@ U_NAMESPACE_BEGIN
 
 /**
  * Implementation of FormattedValue using FieldPositionHandler to accept fields.
+ *
+ * TODO(ICU-20897): This class is unused. If it is not needed when fixing ICU-20897,
+ * it should be deleted.
  */
 class FormattedValueFieldPositionIteratorImpl : public UMemory, public FormattedValue {
 public:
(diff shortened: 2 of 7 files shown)


sffc:
I re-implemented ICU4C ListFormatter using the logic from ICU4J.  This should result in performance improvements as it avoids a quadratic sort.

@jefgen, the MSVC bots are complaining:

```
icu4c\source\i18n\formattedval_impl.h(156,0): Error C4251: 'icu_67::FormattedValueStringBuilderImpl::spanIndices': class 'icu_67::MaybeStackArray<int32_t,8>' needs to have dll-interface to be used by clients of class 'icu_67::FormattedValueStringBuilderImpl' (compiling source file formattedval_iterimpl.cpp)
```

I could export MaybeStackArray, but that's a widely used class.  Suggestions?

jefgen:
> I re-implemented ICU4C ListFormatter using the logic from ICU4J. This should result in performance improvements as it avoids a quadratic sort.

Nice!

> @jefgen, the MSVC bots are complaining:
> 
> ```
> icu4c\source\i18n\formattedval_impl.h(156,0): Error C4251: 'icu_67::FormattedValueStringBuilderImpl::spanIndices': class 'icu_67::MaybeStackArray<int32_t,8>' needs to have dll-interface to be used by clients of class 'icu_67::FormattedValueStringBuilderImpl' (compiling source file formattedval_iterimpl.cpp)
> ```
> 
> I could export MaybeStackArray, but that's a widely used class. Suggestions?

There's already various specializations of `MaybeStackArray` conditionally exported in various places in the ICU4C code, so this wouldn't be anything new I suppose.

You could do something like what is done here:
https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/collationiterator.h#L34

So perhaps something that looks like this (in the header `formattedval_impl.h`): 
```c++
#if U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN
template class U_I18N_API MaybeStackArray<int32_t, 8>;
#endif
```

Or perhaps as an alternative, could try something like what `LocalePriorityList` does, and wrap it with a struct that inherits from `UMemory`:

https://github.com/unicode-org/icu/blob/master/icu4c/source/common/localeprioritylist.cpp#L98

Though would mean that the `MaybeStackArray` is no longer directly part of the class `FormattedValueStringBuilderImpl`, and would be dynamically allocated. Maybe for perf having it be part of the class is better though (ie: one less heap memory allocation and also a pointer to deference rather frequently).

sffc:
Perfect; I adopted the pattern from https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/collationiterator.h#L34

jefgen:
Ugh, I don't think it likes the `typedef` unfortunately. :(

FrankYFTang (approved):
LGTM

PR #1236, opened 2020-08-15
ICU-21157 Fix incorrect null termination.

FrankYFTang:
While the buffer is the exact length of the output string, the
function should not null terminate it but set status to
U_STRING_NOT_TERMINATED_WARNING.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21157
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -48,6 +48,7 @@ static void TestLocDataWithRgTag(void);
 static void TestLanguageExemplarsFallbacks(void);
 static void TestDisplayNameBrackets(void);
 static void TestIllegalArgumentWhenNoDataWithNoSubstitute(void);
+static void Test21157CorrectTerminating(void);
 
 static void TestUnicodeDefines(void);
 
@@ -242,6 +243,7 @@ void addLocaleTest(TestNode** root)
     TESTCASE(TestGetLocale);
 #endif
     TESTCASE(TestDisplayNameWarning);
+    TESTCASE(Test21157CorrectTerminating);
     TESTCASE(TestNonexistentLanguageExemplars);
     TESTCASE(TestLocDataErrorCodeChaining);
     TESTCASE(TestLocDataWithRgTag);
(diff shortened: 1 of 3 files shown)


icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,155 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+              locale, inLocale, status, myErrorName(status), cbuff);
+  }
+  if (buff[length-1] == sentinal1) {
+      log_err("uloc_getDisplayName(%s, %s) w/ maxResultSize=length+1 does not "
+              "change memory in the end of buffer while it should. "
+              "Returns %s\n",
+              locale, inLocale, cbuff);
+  }
+  if (buff[length] != 0x0000) {
+      log_err("uloc_getDisplayName(%s, %s) w/ maxResultSize=length+1 chould "
```
gnrunge, line 6716: chould --> should

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,155 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+static void checkTerminating(
+    const char* locale, const char* inLocale) {
+  UErrorCode status = U_ZERO_ERROR;
+  int32_t preflight_length = uloc_getDisplayName(
+      locale, inLocale, NULL, 0, &status);
+  if (status != U_BUFFER_OVERFLOW_ERROR) {
+      log_err("uloc_getDisplayName(%s, %s) preflight failed", locale, inLocale);
+  }
+  UChar buff[256];
+  const UChar sentinal1 = 0x6C38; // 永- a Han unicode as sentinal.
```
markusicu, line 6679: s/sentinal/sentinel/g

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,155 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+  buff[preflight_length-1] = sentinal1;
+  // Set sentinal2 in the buff[preflight_length] to check it will be replaced
+  // by null.
+  buff[preflight_length] = sentinal2;
+  // It should be properly null terminated at buff[preflight_length].
+  status = U_ZERO_ERROR;
+  int32_t length = uloc_getDisplayName(
+      locale, inLocale, buff, preflight_length + 1, &status);
+  UErrorCode convert_status = U_ZERO_ERROR;
+  // Convert the actual result to UTF8 for debugging output.
```
markusicu, line 6694: UTF-8 does not work for platforms where that is not the system encoding. Please look at other test code for what function to use to convert to the system charset.

Alternatively, log the locale IDs and the output length but not the string itself.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,155 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+              locale, inLocale, buff[length], cbuff);
+  }
+  if (buff[preflight_length - 2] == 0x0000) {
+      log_err("uloc_getDisplayName(%s, %s) w/ maxResultSize=length-1 null "
+              "terminated while it should not. Return '%s'\n",
+              locale, inLocale, cbuff);
+  }
+}
+
+static void Test21157CorrectTerminating(void) {
```
markusicu, line 6807: FYI (really, no need to change)

I find it more convenient to write unit tests in C++ (intltest). As long as we have full *coverage* of C API in cintltst, making sure everything can be used/called from plain C, we can test further functionality from C++.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,155 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+  buff[preflight_length-1] = sentinal1;
+  // Set sentinal2 in the buff[preflight_length] to check it won't be replaced
+  // by null.
+  buff[preflight_length] = sentinal2;
+  status = U_ZERO_ERROR;
+  length = uloc_getDisplayName(
+      locale, inLocale, buff, preflight_length - 1, &status);
+
+  convert_status = U_ZERO_ERROR;
+  // Convert the actual result to UTF8 for debugging output.
```
markusicu, line 6776: When you get an error code, then the output buffer might not contain anything useful, so don't try to read it.

icu4c/source/test/cintltst/cloctst.c
FrankYFTang: > Alternatively, log the locale IDs and the output length but not the string itself.

I prefer we print out the output in debugging so we can see what happen. The position of the null terminator is the key issue of this bug and seeing where the string got truncated is a very helpful information about the error.

FrankYFTang: > UTF-8 does not work for platforms where that is not the system encoding. Please look at other test code for what function to use to convert to the system charset.

I cannot figure out which C api I can use for that. Could we just keep it in UTF8 and let the Linux version help us figure out what happen.

jefgen: I think you'd want to use something like `u_austrcpy`

FrankYFTang:
Changed. PTAL

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,145 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
   }
 }
 
+static void checkTerminating(
+    const char* locale, const char* inLocale) {
```
jefgen, line 6671: nit: FWIW, it looks like the *majority* of this file uses 4 spaces for indentation. For some reason the functions above seem to use only 2 spaces. 
The [ICU User Guide](https://unicode-org.github.io/icu/userguide/dev/codingguidelines.html#tabs-and-indentation) seems to says 4 spaces -- but I don't think that all the code actually does that though...

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,145 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+  UErrorCode status = U_ZERO_ERROR;
+  int32_t preflight_length = uloc_getDisplayName(
+      locale, inLocale, NULL, 0, &status);
+  if (status != U_BUFFER_OVERFLOW_ERROR) {
+      log_err("uloc_getDisplayName(%s, %s) preflight failed", locale, inLocale);
```
jefgen, line 6676: If the test fails here, I wonder if it should just `return` right away?
(Same with the other `log_err` below)

markusicu: Makes sense, but I would accept it either way.

icu4c/source/test/cintltst/cloctst.c
markusicu: Good catch!
Yes, Frank, please fix the indent to 4 spaces.
Looks like you just need to increase the indentation by another 2 on all new lines since the further relative indents appear to be fine already.

For this line itself, I would just move the arguments to the end of the previous line.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6665,6 +6667,145 @@ static void test_unicode_define(const char *namech, char ch, const char *nameu,
...
+  }
+  if (buff[preflight_length - 2] == 0x0000) {
+      log_err("uloc_getDisplayName(%s, %s) w/ maxResultSize=length-1 null "
+              "terminated while it should not. Return '%s'\n",
+              locale, inLocale, cbuff);
+  }
+}
+
+static void Test21157CorrectTerminating(void) {
+  checkTerminating("fr", "fr");
```
markusicu, line 6798: indent 4

markusicu:
Travis Linux/asan shows memory leaks from the test. Could you please take a look? https://travis-ci.org/github/unicode-org/icu/jobs/719141116

jefgen:
> Travis Linux/asan shows memory leaks from the test. 

I have a fix in PR #1242 for it. :)

markusicu:
I restarted the Travis build bot.

PR #1239, opened 2020-08-17
ICU-20088 fix code, tables, typos, broken links, add code styling

robertgmelo:
- Fix typos
- Fix broken links
- Add anchor links
- Adjust table on icu4j-locale-service-provider.md
- Add code styling for source code keywords (class, method, etc.)

### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20088
- [x] Issue accepted


jefgen:
Thanks for working on the User Guide!

Quick FYI/Note: You might want to rebase your changes, as the PR #1228 (which was just merged) has some merge conflicts.
(ex: the `sitemap.md` file no longer exists.)

robertgmelo:
np @jefgen,
I'll rebase here, and try to submit more improvements later today.

robertgmelo:
@jefgen I did the rebase and applied more changes. Could you please take a look?

**Note:** I saw your explanation about reference Markdown files without `.md`. Do you think we should apply for all references? Please note here I've kept *.md*.

docs/userguide/boundaryanalysis/index.md
```diff
@@ -237,7 +237,7 @@ French, Italian and Portuguese.
 
 ICU's break iterators are based on the default boundary rules described in the
 Unicode Standard Annexes [14](https://www.unicode.org/reports/tr14/) and
-[29](https://www.unicode.org/unicode/reports/tr29/) . These are relatively
+[29](https://unicode.org/reports/tr29/) . These are relatively
```
jefgen, line 240: Hmm, it looks like both of these URLs work:
https://www.unicode.org/reports/tr29/
https://unicode.org/reports/tr29/

I wonder which might be better...? (I'm not sure TBH.) 
Though the link as-is doesn't work, so fixing it is a good catch!

docs/userguide/datetime/universaltimescale.md
```diff
@@ -123,31 +123,31 @@ information about a particular timescale. In all of these routines, the
...
-| Precision            | UTSV_UNITS_VALUE        | UNITS_VALUE        |
-| Epoch offet          | UTSV_EPOCH_OFFSET_VALUE | EPOCH_OFFSET_VALUE |
-| Minimum "from" value | UTSV_FROM_MIN_VALUE     | FROM_MIN_VALUE     |
-| Maximum "from" value | UTSV_FROM_MAX_VALUE     | FROM_MAX_VALUE     |
-| Minimum "to" value   | UTSV_TO_MIN_VALUE       | TO_MIN_VALUE       |
-| Maximum "to" value   | UTSV_TO_MAX_VALUE       | TO_MAX_VALUE       |
+| Value                | ICU4C                      | ICU4J                |
+| -------------------- | -------------------------- | -------------------- |
+| Precision            | `UTSV_UNITS_VALUE`         | `UNITS_VALUE`        |
+| Epoch offset          | `UTSV_EPOCH_OFFSET_VALUE` | `EPOCH_OFFSET_VALUE` |
```
jefgen, line 146: nit: the spacing for the second column is slightly off. ;)

docs/userguide/howtouseicu.md
```diff
@@ -220,5 +220,5 @@ Any additional arguments will be passed to the program.
 
 *   `icurun test1.cpp *args...*`
 
-*This feature is a work in progress. Please give feedback at [Ticket
-#8481](https://unicode-org.atlassian.net/browse/ICU-8481).*
+*This feature is a work in progress. Please give feedback at
```
jefgen, line 223: FWIW, this ticket is closed as "Done" and there is nobody watching it -- so if somebody did give feedback/comments on the ticket nobody would notice. :(
I think it might be better to say something like:
"Please give feedback to the `icu-support` mailing list, and refer to Ticket #8481..." instead.

jefgen:
> **Note:** I saw your explanation about reference Markdown files without `.md`. Do you think we should apply for all references? Please note here I've kept _.md_.

Good question...

It looks like Jekyll will automatically handle the `.md` Markdown extension and remove it without any issue. However, it only does this as long as there is nothing else after the extension. If there's a link to a specific anchor then it doesn't work.

For example, the existing link to the `calendar.md` file here (https://github.com/unicode-org/icu/blob/master/docs/userguide/datetime/index.md#L134) gets converted to "https://unicode-org.github.io/icu/userguide/datetime/calendar/"

Whereas if there is a link to a page with an anchor (ex: `concepts.md#sortkeys-vs-comparison`) then Jekyll doesn't do anything and the link kept as-is, so it is broken on the live site. (For example, the link "Sortkeys vs
Comparison" on the page here is broken: https://unicode-org.github.io/icu/userguide/collation/api)

So I think we actually *can* have the `.md` extension on links -- as long as that there's nothing else afterwards. :)
(Which is kind of nice as it means that they would also work in the GitHub renderer as well).

docs/userguide/services.md
```diff
@@ -91,7 +91,7 @@ separate file instead of being embedded in the lines of the program.*
 ### Locales and Services
 
 The interaction between locales and services is fundamental to ICU. Please refer
-to [Locales and Services](./locale/index.md#Locales_and_Services).
+to [Locales and Services](./locale/index.md#locales-and-services).
```
jefgen, line 94: Since this is link with an anchor, I think we'll need to remove the `.md` part, otherwise it'll be broken.

docs/userguide/services.md
```diff
@@ -118,16 +118,20 @@ ICU provides the following [classes](datetime/index.md) to support calendars and
 time zones:
 
 *   [Calendar](datetime/calendar/index.md#calendar)
```
jefgen, line 120: Since this is link with an anchor, I think we'll need to remove the `.md` part, otherwise it'll be broken.
Same with the ones below...
(Idea: You could probably search and replace "`.md#`" with empty string.)

docs/userguide/unicode.md
```diff
@@ -138,7 +138,7 @@ more complexity into the Unicode standard, but far less than managing a large
 number of different encodings.
 
 Starting with Unicode 2.0 (published in 1996), the Unicode standard began
-assigning numbers from 0 to 10ffff16, which requires 21 bits but does not use
+assigning numbers from 0 to 10ffff₁₆, which requires 21 bits but does not use
```
jefgen, line 141: Just as a FYI: You can do `<sup>16</sup>` in Markdown as well.

docs/userguide/boundaryanalysis/index.md
robertgmelo: Let's keep *www* for consistency with other links. I just copied from browser without get full link (copy-paste exception 😅 ). 
Fixed.

docs/userguide/datetime/universaltimescale.md
robertgmelo: Done.

docs/userguide/howtouseicu.md
robertgmelo: Make sense. Done.

docs/userguide/services.md
robertgmelo: Done.

docs/userguide/services.md
robertgmelo: Done. I've checked all "`.md#`" references.

docs/userguide/unicode.md
robertgmelo: Awesome. I also noticed `<sub>16</sub>` can be used, which is the case. Replacing all references with this approach.

docs/userguide/boundaryanalysis/index.md
jefgen: Sounds good, thanks!

docs/userguide/howtouseicu.md
jefgen: thanks!

docs/userguide/services.md
jefgen: Thanks!

PR #1240, opened 2020-08-17
ICU-20088 User Guide migration: Fix links on formatting page.

jefgen:
Fix links on the "Formatting and Parsing" page in the User Guide.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20088
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/docs/userguide/format_parse/index.md
+++ b/docs/userguide/format_parse/index.md
@@ -179,47 +179,47 @@ and messages:
 
 **Formatting Numbers**
 
-*   [NumberFormat](formatparse/numbers/index.md) (§)
+*   [NumberFormat](numbers/legacy-numberformat) (§)
     The abstract superclass that provides the basic fields and methods for
     formatting Number objects and number primitives to localized strings and
     parsing localized strings to Number objects.
 
-*   [DecimalFormat](formatparse/numbers/index.md) (§)
+*   [DecimalFormat](numbers/legacy-numberformat#decimalformat) (§)
     A concrete class for formatting Number objects and number primitives to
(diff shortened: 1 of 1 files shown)


docs/userguide/format_parse/index.md
```diff
@@ -179,47 +179,47 @@ and messages:
 
 **Formatting Numbers**
 
-*   [NumberFormat](formatparse/numbers/index.md) (§)
+*   [NumberFormat](numbers/legacy-numberformat) (§)
```
daniel-ju, line 182: Do these work without the `.md` extension? Clicking the links from the rich diff on Github gives 404 page not found.

jefgen: Yes, the extension-less URLs work on the "fancy" rendered output site, but the ones with the `.md` extension do not.

For example, this doesn't work:
https://unicode-org.github.io/icu/userguide/format_parse/numbers/legacy-numberformat.md

But these do work:
https://unicode-org.github.io/icu/userguide/format_parse/numbers/legacy-numberformat
https://unicode-org.github.io/icu/userguide/format_parse/numbers/legacy-numberformat.html

It is rather unfortunate that GitHub doesn't handle the extension-less URLs... FWIW though they *do* work in my local Markdown editor [VS Code].

So if we want the links to work on the "site", then they need to either be extension-less, or use the .`html` extension. I though using the extension-less URLs seems slightly better as it is (1) a bit clean, and (2) works in local markdown editors as well.  :)

daniel-ju: Thanks! :)

PR #1241, opened 2020-08-18
ICU-13727 limit length of input for Punycode encode/decode

markusicu:
https://unicode-org.atlassian.net/browse/ICU-13727

See the bug ticket for a discussion of the length limits.

Using the new input-too-long error/exception.

The unit tests deliberately use much longer input strings. I do not want them to be tighter than necessary.

### DIFF
--- a/icu4c/source/common/punycode.cpp
+++ b/icu4c/source/common/punycode.cpp
@@ -168,15 +168,23 @@ adaptBias(int32_t delta, int32_t length, UBool firstTime) {
     return count+(((BASE-TMIN+1)*delta)/(delta+SKEW));
 }
 
-#define MAX_CP_COUNT    200
+namespace {
 
-U_CFUNC int32_t
+// ICU-13727: Limit input length for n^2 algorithm
+// where well-formed strings are at most 59 characters long.
+constexpr int32_t ENCODE_MAX_CODE_UNITS=1000;
+constexpr int32_t DECODE_MAX_CHARS=2000;
+
+}  // namespace
+
+// encode
+U_CAPI int32_t
 u_strToPunycode(const UChar *src, int32_t srcLength,
                 UChar *dest, int32_t destCapacity,
                 const UBool *caseFlags,
                 UErrorCode *pErrorCode) {
 
-    int32_t cpBuffer[MAX_CP_COUNT];
+    int32_t cpBuffer[ENCODE_MAX_CODE_UNITS];
     int32_t n, delta, handledCPCount, basicLength, destLength, bias, j, m, q, k, t, srcCPCount;
     UChar c, c2;
 
@@ -187,6 +195,10 @@ u_strToPunycode(const UChar *src, int32_t srcLength,
 
     if(src==NULL || srcLength<-1 || (dest==NULL && destCapacity!=0)) {
         *pErrorCode=U_ILLEGAL_ARGUMENT_ERROR;
+        return 0;
+    }
+    if (srcLength>ENCODE_MAX_CODE_UNITS) {
+        *pErrorCode=U_INPUT_TOO_LONG_ERROR;
         return 0;
     }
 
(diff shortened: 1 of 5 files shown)


icu4c/source/common/punycode.cpp
```diff
@@ -385,6 +390,12 @@ u_strFromPunycode(const UChar *src, int32_t srcLength,
     if(srcLength==-1) {
         srcLength=u_strlen(src);
     }
+    if (srcLength > 2000) {
```
FrankYFTang, line 393: Declare this 2000 as a const with a meaningful name for it.

icu4j/main/classes/core/src/com/ibm/icu/impl/Punycode.java
```diff
@@ -124,6 +124,12 @@ public static StringBuilder encode(CharSequence src, boolean[] caseFlags) throws
         int n, delta, handledCPCount, basicLength, bias, j, m, q, k, t, srcCPCount;
         char c, c2;
         int srcLength = src.length();
+        if (srcLength > 1000) {
```
FrankYFTang, line 127: Define this 1000 as a name private static . Also, why such check is not in the C++ version?

icu4j/main/classes/core/src/com/ibm/icu/impl/Punycode.java
```diff
@@ -263,6 +269,11 @@ private static boolean isSurrogate(int ch){
     public static StringBuilder decode(CharSequence src, boolean[] caseFlags)
                                throws StringPrepParseException{
         int srcLength = src.length();
+        if (srcLength > 2000) {
```
FrankYFTang, line 272: Define this 2000 as a name private static w/ a meaningful name please.

icu4c/source/common/punycode.cpp
markusicu: Seems unnecessary since we use it exactly once, and it has a comment.

icu4j/main/classes/core/src/com/ibm/icu/impl/Punycode.java
markusicu: 1. ditto
2. see the PR description and last ticket comment

icu4j/main/classes/core/src/com/ibm/icu/impl/Punycode.java
markusicu: ditto

FrankYFTang:
Please fix the MSVC build breakage first.

jefgen:
> Please fix the MSVC build breakage first.

Taking a super quick look, here is the failure:

```
uts46test.obj : error LNK2001: unresolved external symbol u_strToPunycode_67 [D:\a\1\s\icu4c\source\test\intltest\intltest.vcxproj]
uts46test.obj : error LNK2001: unresolved external symbol u_strFromPunycode_67 [D:\a\1\s\icu4c\source\test\intltest\intltest.vcxproj]
```

I think the function might need to change from `U_CFUNC` to `U_CAPI`, as the (internal) functions are now being used outside the common library.

icu4c/source/common/punycode.cpp
markusicu: now done

markusicu:
> I think the function might need to change from `U_CFUNC` to `U_CAPI`, as the (internal) functions are now being used outside the common library.

Yes, thanks Jeff!
I debated keeping the Punycode functions purely internal, but testing them would be more awkward via the IDNA API, and would not let me check the actual error code.

FrankYFTang:
Please squash

markusicu:
I just proposed adding a new error code / exception for "input too long" and plan to use those when they are in.

markusicu:
Travis failure: Pre-existing, has a pending PR.

AppVeyor failure:
@jefgen @sffc CygWin *compilation* seems to have crashed. I plan to ignore this for this PR.
```
g++ -D_REENTRANT  -DU_HAVE_ELF_H=1 -DU_HAVE_STRTOD_L=1 -DU_HAVE_XLOCALE_H=0  -DWINVER=0x0601 -D_WIN32_WINNT=0x0601 -I. -I../common   -DU_ATTRIBUTE_DEPRECATED= -DU_I18N_IMPLEMENTATION -O3  -W -Wall -pedantic -Wpointer-arith -Wwrite-strings -Wno-long-long -std=c++11    -c -DPIC  -o number_formatimpl.o number_formatimpl.cpp
make[1]: *** [../config/mh-cygwin64:76: number_formatimpl.o] Segmentation fault
```
https://ci.appveyor.com/project/unicode-org/icu/builds/34950160

jefgen:
> AppVeyor failure: ... CygWin _compilation_ seems to have crashed. 
> 

FWIW, I've seen the AppVeyor builds be flaky like this on occasion. I think ignoring it is fine.
(I can re-queue the build for the PR, though it will perhaps take a bit given how slow AppVeyor is -- but I don't think that should block merging.)

PR #1235, opened 2020-08-14
ICU-21159 Document U_USING_DEFAULT_WARNING in .h

FrankYFTang:
Document the fact
uloc_getDisplay(Language|Script|Country|Variant|Keyword|KeywordValue)
would fallback with the code, case canonicalied in same cases, and
set the status to U_USING_DEFAULT_WARNING.

No change to the implementation behavior. Only complete the missing
comments and tweak line wrap, remove double spaces and add test to
validate this pre-existing behavior that I added the documents now.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21159
- [X] Updated PR title and link in previous line to include Issue number
- [X] Issue accepted
- [X] Tests included
- [X] Documentation is changed or added

PR #1242, opened 2020-08-18
ICU-21159 Fix memory leaks in cintltst TestUsingDefaultWarning.

jefgen:
Fix memory leaks in the test code in PR #1235

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21159
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -6747,70 +6747,70 @@ static void TestBug20149() {
 
 static void TestUsingDefaultWarning() {
     UChar buff[256];
-    char cbuff[256];
+    char errorOutputBuff[256];
     UErrorCode status = U_ZERO_ERROR;
     const char* language = "jJj";
     int32_t length = uloc_getDisplayLanguage(language, "de", buff, 256, &status);
     if (status != U_USING_DEFAULT_WARNING ||
-        u_strcmp(buff, CharsToUChars("jjj")) != 0 ||
+        u_strcmp(buff, u"jjj") != 0 ||
         length != 3) {
-        u_UCharsToChars(buff, cbuff, length+1);
+        u_UCharsToChars(buff, errorOutputBuff, length+1);
         log_err("ERROR: in uloc_getDisplayLanguage %s return len:%d %s with status %d %s\n",
-                language, length, cbuff, status, myErrorName(status));
+                language, length, errorOutputBuff, status, myErrorName(status));
     }
 
     status = U_ZERO_ERROR;
     const char* script = "und-lALA";
     length = uloc_getDisplayScript(script, "de", buff, 256, &status);
     if (status != U_USING_DEFAULT_WARNING ||
-        u_strcmp(buff, CharsToUChars("Lala")) != 0 ||
+        u_strcmp(buff, u"Lala") != 0 ||
         length != 4) {
-        u_UCharsToChars(buff, cbuff, length+1);
+        u_UCharsToChars(buff, errorOutputBuff, length+1);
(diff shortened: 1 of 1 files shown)


icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6750,67 +6750,79 @@ static void TestUsingDefaultWarning() {
     char cbuff[256];
     UErrorCode status = U_ZERO_ERROR;
     const char* language = "jJj";
+    const UChar* languageUChar = CharsToUChars("jjj");
```
daniel-ju, line 6753: It seems like we'd want to remove the `const` here, as the CI pipelines fail due to discarding qualifiers.

jefgen: Oh, good point.
Hmm, I think I might actually change this to use `u_UCharsToChars` instead.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6747,70 +6747,83 @@ static void TestBug20149() {
...
+    UChar expectedBuff[256];
+    char errorOutputBuff[256];
     UErrorCode status = U_ZERO_ERROR;
     const char* language = "jJj";
+    const char expectedLanguage[] = "jjj";
+    u_charsToUChars(expectedLanguage, expectedBuff, UPRV_LENGTHOF(expectedLanguage));
     int32_t length = uloc_getDisplayLanguage(language, "de", buff, 256, &status);
     if (status != U_USING_DEFAULT_WARNING ||
-        u_strcmp(buff, CharsToUChars("jjj")) != 0 ||
+        u_strcmp(buff, expectedBuff) != 0 ||
```
markusicu, line 6758: Can we not use `u"jjj"` in cintltst? As long as we avoid using char16_t?

FrankYFTang: compiler of C won't take u""

jefgen:
> Thanks. Sorry for causing the issue.

No worries! Thanks for taking a look at the PR! :)

icu4c/source/test/cintltst/cloctst.c
jefgen: Oh, that would be nice... but as Frank hints at, support for the `char16_t` type (and the `u""` prefix for string literals) was added in C11. We can't use it, since the ICU code must be C99.

markusicu: My memory is that `u"abc"` works but we either cannot use the type `char16_t` or cannot include the relevant header file, or both, because I think the Apple clang compiler doesn't support it.

So then I grepped for `u"` in cintltst/*.c files and found dozens of examples where we already use UTF-16 string literals in the C unit test code. For example: https://github.com/unicode-org/icu/blob/master/icu4c/source/test/cintltst/cdateintervalformattest.c#L328

So please use that for string literals instead of calling conversion functions.

jefgen: Interesting... AFAICT, the `char16_t` type and the `u""` literals are C11 only. (Ex: https://en.cppreference.com/w/c/language/string_literal , https://en.wikipedia.org/wiki/C11_(C_standard_revision)#Changes_from_C99 ).
However, if we're *already* using them in the test code, then I definitely agree that would be *much* cleaner than having to call the conversion functions!

(I guess it sort of seems like we sort of have a mix of C99 and C11 code... Perhaps the `u""` literals are allowed in test code maybe?)

I'll force-push an update to remove the conversion functions in a moment...

PR #1243, opened 2020-08-18
ICU-21184 rephrase docs/comments using the term grandfathered

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21184

At first mention in a given context I use

> legacy language tags (marked as “Type: grandfathered” in BCP 47)

or similar. I then use just “legacy” or remove unnecessary comments.

Note that since this is a “term of art” in BCP 47, we cannot remove it completely.


icu4c/source/common/locid.cpp
```diff
@@ -1030,7 +1030,8 @@ Locale::forLanguageTag(StringPiece tag, UErrorCode& status)
     // uloc_forLanguageTag() to parse it if it somehow is able to detect that
     // the string actually is BCP-47. This works well for things like strings
     // using BCP-47 extensions, but it does not at all work for things like
-    // BCP-47 grandfathered tags (eg. "en-GB-oed") which are possible to also
+    // legacy language tags (marked as “Type: grandfathered” in BCP 47, e.g., "en-GB-oed")
```
roubert, line 1033: I think it'd be nicer if you kept the line wrapping of the block of text you're editing intact, instead of introducing one single line that now is longer than all the others.

roubert, line 1033: I'd very much prefer if you used the same quotation marks consistently, either “xxx” or "xxx", but not mixing them, especially not on the same line for no obvious reason. The exising ICU4C source code uses the latter, "ASCII quotes", almost exclusively so it might be a good idea to simply stick with that.

roubert, line 1033: Note that the lines directly before and after this line write "BCP-47" instead of "BCP 47".

icu4c/source/i18n/calendar.cpp
```diff
@@ -266,7 +266,7 @@ static ECalType getCalendarTypeForLocale(const char *locid) {
     //TODO: ULOC_FULL_NAME is out of date and too small..
     char canonicalName[256];
 
-    // canonicalize, so grandfathered variant will be transformed to keywords
+    // Canonicalize, so that an old-style variant will be transformed to keywords.
```
roubert, line 269: Is _old-style variant_ really the right expression to use here, wouldn't these be _legacy variants_?

roubert (changes requested):

icu4c/source/common/locid.cpp
markusicu: I would like to keep the whole "legacy ... marked as grandfathered ..." on one line for easier grepping for these things.
I am splitting the example to the next line though.

markusicu: Almost all quotes in ICU docs are ASCII quotes because we were limited to ASCII + HTML entity refs until recently.
I would like to use real quotes for human quoting when I think of them, and ASCII quotes when it's more like a code string literal so someone could copy & paste it into real source code.

markusicu: Fixed. https://www.rfc-editor.org/info/bcp47 uses a space, not a dash.

icu4c/source/i18n/calendar.cpp
markusicu: "grandfathered" was misleading here. Variants like TRADITIONAL predate even our early ICU-syntax '@' key-value extension mechanism, not to mention BCP 47. In fact, the profusion of these kinds of orthogonal tweaks expressed as unusable variants drove us to inventing the '@' syntax first and then to add extensions to BCP 47.

I don't want to use "legacy" here because it would be misunderstood as code for what BCP 47 calls "grandfathered".

PR #1244, opened 2020-08-19
ICU-21237 Removing the obsolete license information for all ICU data files

hagbard:
If this goes in, I think it means that the next data generation run someone does will need to handle deleting and overwriting files manually (since the cleanup code uses the header to determine if a file can safely be removed during data regeneration).

See the ticket for an alternative idea.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21237
- [ ] Issue accepted

### DIFF
--- a/tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_header.txt
+++ b/tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_header.txt
@@ -1,2 +1,1 @@
 © 2016 and later: Unicode, Inc. and others.
-License & terms of use: http://www.unicode.org/copyright.html#License


tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_header.txt
```diff
@@ -1,2 +1 @@
 © 2016 and later: Unicode, Inc. and others.
-License & terms of use: http://www.unicode.org/copyright.html#License
```
markusicu, line 2: Don't remove the whole line, only the #License part as discussed -- and please adjust the PR title+description to match.
I am fine with adding a "generated" line if that helps. I wouldn't add "Auto-" (unusual & redundant).
In total that would be:
```
© 2016 and later: Unicode, Inc. and others.
License & terms of use: http://www.unicode.org/copyright.html

Generated using tools/cldr/cldr-to-icu/build-icu-data.xml
```

hagbard: Oh, once again I misread your original point. You kept talking about the "License" thing, so it seemed like you meant the whole line.

Since the lack of an anchor just puts people at the top of the page anyway, and changing it will be a pain to anyone doing a release (since it breaks the logic for how to detect files that are safe to delete and requires manual file tidyup). How about doing this in 2 phases:

1) Make the file deletion logic not need to care about exact header info (by adding the generated line)
2) Then change the license URL when it won't cause people any grief.

markusicu: I thought I was pretty explicit about the `#License` *anchor* -- see https://unicode-org.atlassian.net/browse/ICU-21028?focusedCommentId=157658

Two phases is fine.

PR #1245, opened 2020-08-19
ICU-21219 Fix for Java version number overflow problem

yumaoka:
Internal API VersionInfo.javaVersion() maps Java version number to  4 integer fields. Each field must be up to 255. However, recent OpenJDK 8 update exceed this range.

Luckily, we have only one reference in our code base for checking Java version. CharsetUTF16 uses maxBytePerChar = 4 for Java 5 and older, maxBytePerChar = 2 for newer Java version. Because we no longer support Java 5 runtime, we don't need this conditional check.

We don't have any other uses of VersionInfo.javaVersion(). Java's version range is not what we can control, so I decided to delete the internal use only API completely.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21219
- [x] Issue accepted
- [ ] Tests included


markusicu:
... and the Java CI bot is happy! “J: Linux OpenJDK (Ubuntu 16.04) succeeded”

PR #1246, opened 2020-08-19
ICU-21237 Improving how file deletion and cleanup works by adding a build label.

hagbard:
Adds a new line to the header to more accurately identify files generated by the tool.
Also makes file cleanup able to distinguish the new label or fallback to old behaviour (this is important or else there will be confusion when people come to regenerate files).

Tested by generating new data "as is" (which adds the new label) and then modifying the header and regenerating again to show the files are recognized even when the header changes.

Once this is in and the files have been regenerated once for the next release, the #License anchor in the copyright URL can just be removed without causing anyone doing regeneration any issues.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21237
- [x] Issue accepted
- [x] Tests included


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -222,23 +244,54 @@ private boolean wasAutoGenerated(Path path) {
...
+                    // generated label) but for older data files, that's fine.
+                    return true;
+                }
+                // Check the next header line (not including the trailing generated label).
+                isLenientHeaderMatchSoFar = matchesToEndOfLine(line, headerStart, headerLines.get(n));
             }
-            return true;
+            // This is actually an unusual case. It corresponds to a file which:
+            // * has a leading comment section at least as long as MAX_HEADER_CHECK_LINES
+            // * does no contain WAS_GENERATED_LABEL anywhere in the first MAX_HEADER_CHECK_LINES
```
markusicu, line 281:
```suggestion
            // * does not contain WAS_GENERATED_LABEL anywhere in the first MAX_HEADER_CHECK_LINES
```

hagbard:
Figured this could do with tests, so I added some. Obviously squashing etc. nukes lgtm.

PR #1248, opened 2020-08-20
ICU-21028 Adding ICU version file generation to tools

hagbard:
Note that in this version the CLDR version is (by default) from CLDRFile, but could be overwritten.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21028
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/cldr-to-icu/build-icu-data.xml
+++ b/tools/cldr/cldr-to-icu/build-icu-data.xml
@@ -46,7 +46,14 @@
         <!-- The directory in which the additional ICU XML data is stored. -->
         <property name="specialsDir" value="${basedir}/../../../icu4c/source/data/xml"/>
 
-        <!-- An override for the CLDR version string to be used in generated data. -->
+        <!-- Default value for ICU version (icuver.txt). Update this for each release. -->
+        <property name="icuVersion" value="67.1.0.0"/>
+
+        <!-- Default value for ICU data version (icuver.txt). Update this for each release. -->
+        <property name="icuDataVersion" value="67.1.0.0"/>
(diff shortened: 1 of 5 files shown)


tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -46,7 +46,14 @@
         <!-- The directory in which the additional ICU XML data is stored. -->
         <property name="specialsDir" value="${basedir}/../../../icu4c/source/data/xml"/>
 
-        <!-- An override for the CLDR version string to be used in generated data. -->
+        <!-- Default value for ICU version (icuver.txt). Update this for each release. -->
```
jefgen, line 49: Oh, one quick thought -- do we want to also add this to the steps on the page here as well:
http://site.icu-project.org/processes/release/tasks/versions

markusicu: Yes. The section "icu4c/source/data/misc/icuver.txt needs to be updated ..." needs to be updated to say that this xml file here needs to be updated, and that the .txt file is now generated.

markusicu: I updated the release task description. Please fix if I got it wrong.

jefgen: Thanks Markus!
FWIW, it looks good to me -- thanks for updating it!
Also, cc: @erik0686 and @daniel-ju, since I think they'll be working on the version number updates.

PR #1249, opened 2020-08-20
ICU-21035 Replace backward compatibility ulocimp_getLanguage() overload.

roubert:

By always calling the dynamic memory allocation implementation directly
instead, the fixed memory buffer boundary gets pushed one step further
towards the edges.

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -464,8 +464,18 @@ parseTagString(
         goto error;
     }
 
-    subtagLength = ulocimp_getLanguage(position, lang, *langLength, &position);
-    u_terminateChars(lang, *langLength, subtagLength, err);
+    {
+        icu::CharString result = ulocimp_getLanguage(position, &position, *err);
+        if (U_FAILURE(*err)) {
+            goto error;
+        }
+
+        subtagLength = result.length();
+        if (subtagLength <= *langLength) {
+            uprv_memcpy(lang, result.data(), subtagLength);
+        }
+        u_terminateChars(lang, *langLength, subtagLength, err);
+    }
 
     /*
      * Note that we explicit consider U_STRING_NOT_TERMINATED_WARNING
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -1148,7 +1148,7 @@ uloc_getCurrentLanguageID(const char* oldID){
  *
  * TODO try to use this in Locale
  */
-static CharString
+CharString U_EXPORT2
 ulocimp_getLanguage(const char *localeID,
                     const char **pEnd,
                     UErrorCode &status) {
@@ -1191,20 +1191,6 @@ ulocimp_getLanguage(const char *localeID,
     }
 
     return result;
-}
-
-U_CFUNC int32_t
-ulocimp_getLanguage(const char *localeID,
-                    char *language, int32_t languageCapacity,
-                    const char **pEnd) {
-    ErrorCode status;
-    CharString result = ulocimp_getLanguage(localeID, pEnd, status);
-    if (status.isFailure()) {
-        return 0;
-    }
-    int32_t reslen = result.length();
-    uprv_memcpy(language, result.data(), std::min(reslen, languageCapacity));
-    return reslen;
 }
 
 static CharString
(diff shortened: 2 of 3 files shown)


icu4c/source/common/loclikely.cpp
```diff
@@ -464,8 +466,16 @@ parseTagString(
...
-    subtagLength = ulocimp_getLanguage(position, lang, *langLength, &position);
-    u_terminateChars(lang, *langLength, subtagLength, err);
+    {
+        icu::CharString result = ulocimp_getLanguage(position, &position, *err);
+        if (U_FAILURE(*err)) {
+            goto error;
+        }
+
+        subtagLength = result.length();
+        uprv_memcpy(lang, result.data(), std::min(*langLength, subtagLength));
```
markusicu, line 476: We don't usually write to an output buffer if only part of the result fits.
```
if (subtagLength <= *langLength) {
    uprv_memcpy(lang, result.data(), subtagLength);
}
```

icu4c/source/common/uloc.cpp
```diff
@@ -1755,8 +1744,14 @@ uloc_getLanguage(const char*    localeID,
...
 
-    i=ulocimp_getLanguage(localeID, language, languageCapacity, NULL);
-    return u_terminateChars(language, languageCapacity, i, err);
+    CharString result = ulocimp_getLanguage(localeID, NULL, *err);
+    if (U_FAILURE(*err)) {
+        return 0;
+    }
+
+    int32_t reslen = result.length();
+    uprv_memcpy(language, result.data(), std::min(languageCapacity, reslen));
```
markusicu, line 1753: ditto

We should really add a CharString method `extract(dest, capacity, errorCode)` like https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeString.html#a6d9b3ff497c7f0abf84cea6946b1633b

I can make a PR after yours for that.

icu4c/source/common/uloc.cpp
```diff
@@ -1486,7 +1472,11 @@ uloc_openKeywords(const char* localeID,
     }
 
     /* Skip the language */
-    ulocimp_getLanguage(tmpLocaleID, NULL, 0, &tmpLocaleID);
+    ulocimp_getLanguage(tmpLocaleID, &tmpLocaleID, *status);
+    if (U_FAILURE(*status)) {
```
markusicu, line 1476: optional: It's ok (and a little less clutter) to write this as a one-liner (but still with the braces): `if (U_FAILURE(*status)) { return 0; }`

icu4c/source/common/uloc.cpp
roubert: > ditto

Done.

> We should really add a CharString method `extract(dest, capacity, errorCode)`

I hope that when all of this is done one day, such a method won't really be needed then.

icu4c/source/common/loclikely.cpp
```diff
@@ -464,8 +464,22 @@ parseTagString(
...
-    u_terminateChars(lang, *langLength, subtagLength, err);
+    {
+        icu::CharString result = ulocimp_getLanguage(position, &position, *err);
+        if (U_FAILURE(*err)) {
+            goto error;
+        }
+
+        subtagLength = result.length();
+        if (subtagLength > *langLength) {
+            *lang = '\0';
```
markusicu, line 475: This should not be necessary.

The following should really suffice, as suggested before:
```
if (subtagLength <= *langLength) {
    uprv_memcpy(lang, result.data(), subtagLength);
}
```

The already then-following u_terminateChars() will set the buffer overflow error, and the next statement takes care of the `goto error;`.

Keep it simple.

icu4c/source/common/uloc.cpp
```diff
@@ -1755,8 +1744,22 @@ uloc_getLanguage(const char*    localeID,
...
-    return u_terminateChars(language, languageCapacity, i, err);
+    CharString result = ulocimp_getLanguage(localeID, NULL, *err);
+    if (U_FAILURE(*err)) {
+        return 0;
+    }
+
+    int32_t reslen = result.length();
+    if (reslen > languageCapacity) {
+        if (language != NULL) {
+            *language = '\0';
```
markusicu, line 1755: ditto

```
if (reslen <= languageCapacity) {
    uprv_memcpy(language, result.data(), reslen);
}
return u_terminateChars(language, languageCapacity, reslen, err);
```

icu4c/source/common/loclikely.cpp
roubert: Done.

icu4c/source/common/uloc.cpp
roubert: Done.

markusicu:
“All checks have passed” :-)

PR #1250, opened 2020-08-20
ICU-21230 Add status parameter to MaybeStackArray(int32_t newCapacity) constructor

hugovdm:
It’s probably a very unlikely failure: how often might memory allocation fail for something like this?

But “more correct” is probably not bad, and the “toString” that’s hereby getting a “status” parameter is probably mostly for debugging?

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21230
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -297,12 +297,16 @@ public:
      * Automatically allocates the heap array if the argument is larger than the stack capacity.
      * Intended for use when an approximate capacity is known at compile time but the true
      * capacity is not known until runtime.
-     *
-     * WARNING: does not report errors upon memory allocation failure, after
-     * which capacity will be stackCapacity, not the requested newCapacity.
      */
-    MaybeStackArray(int32_t newCapacity) : MaybeStackArray() {
-        if (capacity < newCapacity) { resize(newCapacity); }
+    MaybeStackArray(int32_t newCapacity, UErrorCode status) : MaybeStackArray() {
+        if (U_FAILURE(status)) {
+            return;
+        }
+        if (capacity < newCapacity) {
+            if (resize(newCapacity) == nullptr) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+            }
+        }
     }
     /**
      * Destructor deletes the array (if owned).
--- a/icu4c/source/i18n/number_decimalquantity.cpp
+++ b/icu4c/source/i18n/number_decimalquantity.cpp
@@ -20,6 +20,7 @@
 #include "charstr.h"
 #include "number_utils.h"
 #include "uassert.h"
+#include "util.h"
 
 using namespace icu;
 using namespace icu::number;
@@ -634,7 +635,10 @@ void DecimalQuantity::toDecNum(DecNum& output, UErrorCode& status) const {
 
     // Use the BCD constructor. We need to do a little bit of work to convert, though.
     // The decNumber constructor expects most-significant first, but we store least-significant first.
-    MaybeStackArray<uint8_t, 20> ubcd(precision);
+    MaybeStackArray<uint8_t, 20> ubcd(precision, status);
+    if (U_FAILURE(status)) {
+        return;
+    }
     for (int32_t m = 0; m < precision; m++) {
         ubcd[precision - m - 1] = static_cast<uint8_t>(getDigitPos(m));
     }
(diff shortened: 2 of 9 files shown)


icu4c/source/common/cmemory.h
```diff
@@ -297,12 +297,16 @@ class MaybeStackArray {
      * Automatically allocates the heap array if the argument is larger than the stack capacity.
      * Intended for use when an approximate capacity is known at compile time but the true
      * capacity is not known until runtime.
-     *
-     * WARNING: does not report errors upon memory allocation failure, after
-     * which capacity will be stackCapacity, not the requested newCapacity.
      */
-    MaybeStackArray(int32_t newCapacity) : MaybeStackArray() {
-        if (capacity < newCapacity) { resize(newCapacity); }
+    MaybeStackArray(int32_t newCapacity, UErrorCode status) : MaybeStackArray() {
```
markusicu, line 301: This `: MaybeStackArray()` is new syntax to me. Looks like it invokes the default constructor before doing its specific work? That would be nice... otherwise this constructor would leave its fields undefined if the status is already an error.

icu4c/source/common/cmemory.h
```diff
@@ -297,12 +297,16 @@ class MaybeStackArray {
...
+        if (U_FAILURE(status)) {
+            return;
+        }
+        if (capacity < newCapacity) {
+            if (resize(newCapacity) == NULL) {
```
markusicu, line 306: Please use nullptr in new code.

icu4c/source/i18n/numparse_types.h
```diff
@@ -64,8 +64,9 @@ class CompactUnicodeString {
         fBuffer[0] = 0;
     }
 
-    CompactUnicodeString(const UnicodeString& text)
-            : fBuffer(text.length() + 1) {
+    CompactUnicodeString(const UnicodeString& text, UErrorCode& status)
+        : fBuffer(text.length() + 1, status) {
```
markusicu, line 68: Needs indentation. My preference:
```
    CompactUnicodeString(const UnicodeString& text, UErrorCode& status) :
            fBuffer(text.length() + 1, status) {
        ...
```

icu4c/source/common/cmemory.h
markusicu: Self-education: https://en.cppreference.com/w/cpp/language/constructor#Delegating_constructor -- since C++11 -- very nice.

icu4c/source/i18n/number_decimalquantity.cpp
```diff
@@ -1323,8 +1326,8 @@ bool DecimalQuantity::operator==(const DecimalQuantity& other) const {
     }
 }
 
-UnicodeString DecimalQuantity::toString() const {
-    MaybeStackArray<char, 30> digits(precision + 1);
+UnicodeString DecimalQuantity::toString(UErrorCode& status) const {
+    MaybeStackArray<char, 30> digits(precision + 1, status);
```
sffc, line 1330: Optional: remove the `UErrorCode&` parameter that you added to the function, and then write

```cpp
MaybeStackArray<char, 30> digits(precision + 1, localStatus);
if (U_FAILURE(localStatus)) {
    return ICU_Utility::makeBogusString();
}
```

icu4c/source/i18n/numparse_types.h
hugovdm: Can I go with the old formatting of this line, i.e. going for minimal diffs? (I've done this already.)

Maybe a good time to discuss https://github.com/unicode-org/icu/pull/1143 - Younies and I: are taking guidance from clang-format for new code. I manually override to keep things consistent with other code in the same file, but for new files I basically just follow clang-format's recommendations.

I wrote a bunch more here, but I'm moving that to #1143.

icu4c/source/i18n/number_decimalquantity.cpp
hugovdm: Done.

icu4c/source/i18n/numparse_types.h
sffc: If given a choice, I prefer the `:` on the line below the constructor signature, but I most prefer to defer to clang-tidy.

markusicu:
> @markusicu what's your usual approach to the sequence of reviewing and squashing?

I find it hard enough in the GitHub review UI to track changes vs. comments as it is, so I prefer squashing at the end, unless the change is really small and simple. Normally I just send a comment like "lgtm pse squash" and only after that do I approve.

markusicu:
> CI failure: "The job running on agent Hosted VS2017 4 ran longer than the maximum time of 60 minutes." - I clicked on "re-run".

I have seen that on another PR. Pretty sure it's not from yours. I have alerted Jeff, but it looks like this configuration hangs in one of Frank's tests. We will need to tackle this one separately.

PR #1252, opened 2020-08-20
ICU-21035 Replace backward compatibility ulocimp_getScript() overload.

roubert:
By always calling the dynamic memory allocation implementation directly
instead, the fixed memory buffer boundary gets pushed one step further
towards the edges.

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -1181,7 +1181,7 @@ ulocimp_getLanguage(const char *localeID,
     return result;
 }
 
-static CharString
+CharString U_EXPORT2
 ulocimp_getScript(const char *localeID,
                   const char **pEnd,
                   UErrorCode &status) {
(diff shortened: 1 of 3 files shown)


icu4c/source/common/uloc.cpp
```diff
@@ -1746,9 +1739,13 @@ uloc_getScript(const char*    localeID,
     }
 
     if(_isIDSeparator(*localeID)) {
-        i=ulocimp_getScript(localeID+1, script, scriptCapacity, NULL);
+        return ulocimp_getScript(localeID+1, NULL, *err).extract(script, scriptCapacity, *err);
     }
-    return u_terminateChars(script, scriptCapacity, i, err);
+
+    if (script != NULL && scriptCapacity > 0) {
```
markusicu, line 1745: The old code set a not-terminated-warning if the capacity is 0. How about keeping `return u_terminateChars(script, scriptCapacity, 0, err);` instead of this replacement if-terminate-return?

roubert: Done.

markusicu:
Please check the other similar functions whether we overlooked calling u_terminateChars() there too in recent changes.

roubert:
> Please check the other similar functions whether we overlooked calling u_terminateChars() there too in recent changes.

The other such function would be `uloc_getLanguage()` which returns unconditionally through `extract()` and therefore through `u_terminateChars()`.

PR #1253, opened 2020-08-20
ICU-21035 add & use CharString::extract(dest buffer)

markusicu:

Inspired by PR #1249 and similar to UnicodeString::extract(dest buffer).


FrankYFTang:
That is my fault. The test code is not correctly null terminiated the string before I call u_austrcpy in some cases. I was using u_strToUTF8 before which passed in the length but then I forgot to terminate the buffer after jeff suggest me to switch to use u_austrcpy. Sorry. Will work on a fix soon.

FrankYFTang:
Fix for the loop issue jeff mentioned is here https://github.com/unicode-org/icu/pull/1256/ sorry

markusicu:
> That is my fault.
> Sorry.

It's not just your fault. I didn't look at what was used elsewhere in these tests, and glanced over the issues when the build bots *seemed* happy.

I am sorry, too.

jefgen:
Well, FWIW, I think maybe one idea is that we could try to avoid some of the issues by running the Valgrind check on PRs. :)
I filed a ticket here to consider the idea: https://unicode-org.atlassian.net/browse/ICU-21244 
(I think it might only add ~15 mins to the current set of checks.)

PR #1196, opened 2020-07-08
ICU-21191 Improve configurability of testtagsguards.sh

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21191
- [x] Updated PR title and link in previous line to include Issue number
- [ ] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #1254, opened 2020-08-21
ICU-21236 Correct Locale canonicalization

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21236
- [X] Issue accepted
- [X] Tests included

Bugs fixed by this PR

* https://unicode-org.atlassian.net/browse/ICU-21236
* https://unicode-org.atlassian.net/browse/ICU-13899
* https://unicode-org.atlassian.net/browse/ICU-21139
* https://unicode-org.atlassian.net/browse/ICU-21115
* https://unicode-org.atlassian.net/browse/ICU-7656


FrankYFTang:
Read for review now!

FrankYFTang:
Rebased on ad043349ed92b850ab27c159e29481877558e3f8 and force pushed.

FrankYFTang:
Mark incomplete after discuss with Markus. The class we use for the Unique String is not design for some of the usage and I need to replace the code of how I use it.

FrankYFTang:
ready for review. @markusicu

markusicu:
One of the Travis test failures is the dependency checker complaining that something in locid.cpp is using global operator new. Please fix.
```
$ ( cd ../icu4c/source/test/depstest && ./depstest.py ../../../../build/ )
...
Error: in group resourcebundle: common/locid.o imports operator new(unsigned long) - is this a new system symbol?
...
Error: There were errors, please fix them and re-run. Processing may have terminated abnormally.
The command "( cd ../icu4c/source/test/depstest && ./depstest.py ../../../../build/ )" exited with 1.
```

markusicu:
@sffc one of the Travis tests here is failing, and it looks related to your PR #1196 -- does the build bot (sometimes??) pick an incompatible shell?

Troublingly, the build does not seem to fail due to this. (There is another piece also failing.) @gnrunge could you please look at making sure that testtagsguards.sh also returns with a non-zero error code when there is a failure? Or did this properly report errors when it got past the initialization?

https://travis-ci.org/github/unicode-org/icu/jobs/727243445
```
9.04s$ ( cd .. && source/test/hdrtst/testtagsguards.sh )
source/test/hdrtst/testtagsguards.sh: line 9: DEF:=-DU_NO_DEFAULT_INCLUDE_UTF_HEADERS=1: command not found
source/test/hdrtst/testtagsguards.sh: line 10: INCL:=-Isource/common -Isource/i18n -Isource/io: No such file or directory
source/test/hdrtst/testtagsguards.sh: line 11: TMPDIR:=/tmp/tmp.f0RSsiuhLF: No such file or directory
source/test/hdrtst/testtagsguards.sh: line 12: DIFF:=diff -u --minimal: command not found
source/test/hdrtst/testtagsguards.sh: line 13: CXX:=clang++: command not found
*** testtagsguards.sh TMPDIR=
source/common/unicode/appendable.h
source/test/hdrtst/testtagsguards.sh: line 20: /ht-appendable.h.cpp: Permission denied
clang-7: error: no such file or directory: '/ht-appendable.h.cpp'
clang-7: error: no input files
...
pass
The command "( cd .. && source/test/hdrtst/testtagsguards.sh )" exited with 0.
```

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1071 @@ Locale::operator==( const   Locale& other) const
...
+        ures_getByKey(metadataAlias.getAlias(), "territory", nullptr, &status));
+    LocalUResourceBundlePointer variantAlias(
+        ures_getByKey(metadataAlias.getAlias(), "variant", nullptr, &status));
+
+    int32_t languagesLength = 0, scriptLength = 0, territoryLength = 0,
+            variantLength = 0;
+
+    // Read the languageAlias into languageTypes, languageReplacementIndexes
+    // and strings
+    LocalPointer<UniqueCharStrings> strings(new UniqueCharStrings(status));
```
markusicu, line 898: This is where global "new" is used because UniqueCharStrings does not extend UMemory. Please stack-allocate the "strings" object, not heap-allocate it.

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1071 @@ Locale::operator==( const   Locale& other) const
...
+                        status);
+    }
+
+    // copy hashtables
+    return new AliasData(
+        languageMap.orphan(),
+        scriptMap.orphan(),
+        territoryMap.orphan(),
+        variantMap.orphan(),
+        strings.orphan());
```
markusicu, line 984: ... and then call `strings.orphanCharStrings()` and store the CharString, and then let the UniqueCharStrings object go out of scope.

(Like in loclikelysubtags.cpp)

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1071 @@ Locale::operator==( const   Locale& other) const
...
+
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+
+    // We can only use strings after freeze it.
+    strings->freeze();
+
+    // Build the languageMap from languageTypes & languageReplacementIndexes
+    LocalPointer<CharStringMap> languageMap(new CharStringMap(490, status));
```
markusicu, line 947: I would stack-allocate these CharStringMap's too (like in my code), and in the end `std::move` them into the runtime data.

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1071 @@ Locale::operator==( const   Locale& other) const
...
+
+/**
+ * Read the languageAlias data from alias to strings+types+replacementIndexes.
+ * Allocate length itmes for types, to store the type field. Allocate length
+ * itmes for replacementIndexes, to store the index in the strings for the
+ * replacement language.
+ */
+void
+AliasDataBuilder::readLanguageAlias(
+        UResourceBundle* alias,
```
markusicu, line 699: This works, but I find it much cleaner to work with ures_getValueWithFallback() and ResourceDataValue etc.

Is there a reason you didn't copy that kind of data-loading code from loclikelysubtags.cpp? It should have been possible to copy more of that code, incl. readStrings() etc.

Using the public API will work, and there is no time to change this now.

Maybe a TODO?

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1071 @@ Locale::operator==( const   Locale& other) const
...
+
+    // Gather fields and generate locale ID into out.
+    CharString& outputToString(CharString& out, UErrorCode status);
+
+    // Generate the lookup key.
+    CharString& generateKey(const char* language, const char* region,
+                            const char* variant, CharString& out,
+                            UErrorCode status);
+
+    void parseLangaugeReplacement(const char* replacement,
```
markusicu, line 1045: typo --> parseLanguageReplacement

icu4c/source/test/intltest/loctest.h
```diff
@@ -122,6 +122,8 @@ class LocaleTest: public IntlTest {
     void TestBug13554();
     void TestBug20410();
     void TestBug20900();
+    void TestLocaleCanonicalizationFromFile();
+    void TestKnownCanonicalizatedListCorrect();
```
markusicu, line 126: typo --> TestKnownCanonicalizedListCorrect

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ULocaleTest.java
```diff
@@ -5175,21 +5177,20 @@ public void TestCanonical() {
         // also test with script, variants and extensions
         Assert.assertEquals("fa-Cyrl-AF-1009-u-ca-roc", canonicalTag("prs-Cyrl-1009-u-ca-roc"));
 
-        if (!logKnownIssue("21236", "skip some canonicalization tests until code fixed")) {
-            //  language _ country -> language _ script _ country
-            Assert.assertEquals("pa-Guru-IN", canonicalTag("pa-IN"));
-        }
+        //  language _ country -> language _ script _ country
```
markusicu, line 5180: This comment does not match the tested behavior. Please update or remove the comment.

Do you know/remember why we expected canonicalization to add script subtags?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ULocaleTest.java
```diff
@@ -5175,21 +5177,20 @@ public void TestCanonical() {
...
-            Assert.assertEquals("pa-Guru-IN-5678-u-ca-hindi", canonicalTag("pa-IN-5678-u-ca-hindi"));
+        // also test with variants and extensions
+        Assert.assertEquals("pa-IN-5678-u-ca-hindi", canonicalTag("pa-IN-5678-u-ca-hindi"));
 
-            //  language _ script _ country -> language _ country
-            Assert.assertEquals("ky-KG", canonicalTag("ky-Cyrl-KG"));
-            // also test with variants and extensions
-            Assert.assertEquals("ky-KG-3456-u-ca-roc", canonicalTag("ky-Cyrl-KG-3456-u-ca-roc"));
-        }
+        //  language _ script _ country -> language _ country
```
markusicu, line 5187: This comment does not match either.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+                this.region = replacedRegion;
+                if (searchVariant != null && !searchVariant.isEmpty()) {
+                    if (replacedVariant != null && !replacedVariant.isEmpty()) {
+                        this.variants.set(variantIndex, replacedVariant);
+                    } else {
+                        this.variants.remove(variantIndex);
+                    }
+                }
+                if (replacedExtensions != null) {
+                    this.extensions = replacedExtensions;
```
markusicu, line 1452: This clobbers all existing extensions... What are the possible values here? Is it just for -u-va-posix? Should we try to "mix that in"? TODO?

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1239,147 +1563,54 @@ public static String canonicalize(String localeID){
...
-                            parser.getVariant()))
-                        .append(parser.getName().substring(parser.getBaseName().length()))
-                        .toString());
-                } catch (MissingResourceException e) {
-                }
+        boolean knownCanonicalized = false;
+        String name = parser.getName();
+        if (!isKnownCanonicalizedLocale(name)) {
+            AliasReplacer replacer = new AliasReplacer(
+                parser.getLanguage(), parser.getScript(), parser.getCountry(), parser.getVariant().toLowerCase(),
```
markusicu, line 1570: not `toLowerCase()` -- that does more work than we need for ASCII, and with the implicit default locale, it will fail on Turkish etc. systems

I feel sure that we have one or more helper classes for ASCII lowercasing.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1239,147 +1563,54 @@ public static String canonicalize(String localeID){
...
+                parser.getName().substring(parser.getBaseName().length()));
```
markusicu, line 1571: getName() is in ICU syntax, so if there are extensions, this will start with `@`. intentional? harmless? comment? tested?

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1239,147 +1563,54 @@ public static String canonicalize(String localeID){
...
-            } catch (MissingResourceException e) {
-            }
+        return parser.getName();
+    }
+
+    private static boolean isKnownCanonicalizedLocale(String name) {
+        if (name.equals("c") || name.equals("en") || name.equals("en_US")) {
+            return true;
         }
+        if (gKnownCanonicalizedCases == null) {
```
markusicu, line 1585: Remember to declare the function "synchronized", or synchronize(getClass()) a block surrounding this lazy init.

markusicu (commented):
I finally made it through the Java code... I plan to stop here, wait for you to make changes, and once we agree on Java please make equivalent changes in C++ before I take a deeper look at locid.cpp. Or if you like, you can make parallel changes in Java & C++ at the same time.

sffc:
I commented regarding testtagsguards.sh in #1196.

icu4c/source/test/testdata/localeCanonicalization.txt
```diff
@@ -0,0 +1,1648 @@
+# File copied from cldr common/testData/localeIdentifiers/localeCanonicalization.txt
```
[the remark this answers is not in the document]
FrankYFTang: yes, I am not sure what you like me to do in this PR for that. 
I believe common/testData/localeIdentifiers/localeCanonicalization.txt is generated by mark but I believe there are no code hook up to automatically generate that inside CLDR when anyone change that file.

icu4c/source/common/locid.cpp
FrankYFTang: Because I copy code from somewhere else (Region::loadRegionData i18n/region.cpp. That file read from the same resource file so I copy the code from there instead from loclikelysubtags.cpp which read a different resource file.) No way I can tell which one is better than the other the moment I copied the code, but it is clear copy from i18n/region.cpp is easier since it read the same resource file.

icu4c/source/common/locid.cpp
FrankYFTang: ok I will do it, but it is just a shock we cannot use new in C++.

icu4c/source/common/locid.cpp
FrankYFTang: ok. Well do.

icu4c/source/common/locid.cpp
FrankYFTang: May be my C++ knowledge is too old, but I am still not sure why we want to move from the pointer w/ new / delete pattern to this stack allocate and std::move pattern.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ULocaleTest.java
FrankYFTang: I start to write the code and test tests when Mark and I reach the understanding of how the algorithm should work but before he change the data in CLDR. After I wrote these test based on the CLDR shipped with ICU 67, he figured some of the CLDR data were there fore addLikelySubtag and do not need to be there so he removed them. So now this part won't change since those entries got removed but I forgot to change the comments.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+                return false;
+            }
+            assert this.script != replacement;
+            this.script = replacement;
+            // The script is changed by data in script alias.
+            return true;
+        }
+
+        private boolean replaceVariant() {
+            if (variants == null) return false;
```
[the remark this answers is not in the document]
FrankYFTang: ok. then I need to set the variants to null if the variants is empty after we remove one.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+                String language, String region, String variant) {
+            StringBuilder buf = new StringBuilder();
+            if (language != null) {
+                buf.append(language);
+            }
+            if (region != null && !region.isEmpty()) {
+                buf.append(UNDERSCORE);
+                buf.append(region);
+            }
+            if (variant != null && !variant.isEmpty()) {
```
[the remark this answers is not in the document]
FrankYFTang: well. I have to filter that out outside this function otherwise, the logic would mass up of the replacement, because the outter function may believe the  ill-formed variant is in the type but it is not in the type while we access the key. for example, if we have "nb_NO_NY", if we use >= 4 the outter function will think the type is nb_NY while the key generated is really nb only, then when it find the replacement for "no" for "nb", it will see there are no variant in the replacement but believe the NY is in the type and therefore remove the NY . So if we do not want to search for nb_NY, I have to ignore the NY outside this function so the outter function think the key is only "nb" and therefore won't try to remove "NY" from it.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+            UResourceBundle languageAlias = metadataAlias.get("language");
+            UResourceBundle scriptAlias = metadataAlias.get("script");
+            UResourceBundle territoryAlias = metadataAlias.get("territory");
+            UResourceBundle variantAlias = metadataAlias.get("variant");
+
+            for ( int i = 0 ; i < languageAlias.getSize(); i++ ) {
+                UResourceBundle res = languageAlias.get(i);
+                String aliasFrom = res.getKey();
+                String aliasTo = res.get("replacement").getString();
+                languageAliasMap.put(aliasFrom, aliasTo);
```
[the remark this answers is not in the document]
FrankYFTang: I just added the check

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+                }
+
+                replacedScript = deleteOrReplace(script, null, replacedScript);
+                replacedRegion = deleteOrReplace(region, searchRegion, replacedRegion);
+                replacedVariant = deleteOrReplace(searchVariant, searchVariant, replacedVariant);
+
+                if (same(this.language, replacedLanguage) &&
+                    same(this.script, replacedScript) &&
+                    same(this.region, replacedRegion) &&
+                    same(searchVariant, replacedRegion) &&
```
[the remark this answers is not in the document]
FrankYFTang: yea. Thanks

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
FrankYFTang: We currently have four entries which have extension in langaugeAlias :
```
            i_default{
                reason{"legacy"}
                replacement{"en_x_i_default"}
            }
            i_enochian{
                reason{"legacy"}
                replacement{"und_x_i_enochian"}
            }
            i_mingo{
                reason{"legacy"}
                replacement{"see_x_i_mingo"}
            }
            zh_min{
                reason{"legacy"}
                replacement{"nan_x_zh_min"}
            }
```

The type (which is the key in the resources) are not legal under uts35 unicode identifier syntax, so the only time we will get into problem is if we have something like
zh-min-u-ca-chinese
or
i-enochian-u-kf-upper

@macchiati 
My understanding is according to Mark's new spec (https://docs.google.com/document/d/1x14W-jBfra3Y-Lg-UfnippIMmU29VISI8dSAAUCcRhE/edit#) , these entries are considered as part of "BCP47 LegacyRules" But his draft spec stated:
"If the source is identical to one of the types in the BCP47 LegacyRules, replace the entire source by the replacement value.
"
It seems ULocale, before the code reach this point, already change 
zh-min to nan@x=zh-min
and change
zh-min-u-ca-chinese to nan@x=zh-min-u-ca-chinese

I am not sure we should do a "mix that in" unless Mark's spec said we should do that.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
FrankYFTang: good catch AsciiUtil.toLowerString() should do it.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
FrankYFTang: parser.getName().substring(parser.getBaseName().length()) 
would get the part of of the locale starting "@" to the AliasReplacer, and the the replace() method will append them back. 
There are test cases cover that already

FrankYFTang:
I believe I changed the code to address all Mark's comment. PTAL @markusicu

FrankYFTang:
friendly ping

icu4c/source/common/locid.cpp
markusicu: We can and do use operator new on subclasses of UMemory. If needed, we can make UniqueCharStrings be such a subclass, but I don't think we need it.

https://unicode-org.github.io/icu/userguide/dev/codingguidelines.html#memory-allocation

icu4c/source/common/locid.cpp
markusicu: fewer heap allocations

icu4c/source/test/testdata/localeCanonicalization.txt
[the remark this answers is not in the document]
markusicu: As I said, I think this should be documented for now. Not sure if we have some place where people actually look, but at a quick glance cldr-icu-readme.txt seemed to make sense because that's where people look who import data from CLDR.

And for the future (via a ticket), it would be better to automate this copying, which would probably be an addition to the cldr-to-icu converter or a sibling tool. (Like Hugo was looking into for unit formatting test data.)

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
markusicu: ULocale.toString() uses the old `@` syntax for compatibility.
ULocale.toLanguageTag() uses `-x-...`.
Please submit a ticket and add a TODO here to revisit this.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1386,53 +1442,59 @@ private boolean replaceLanguage(boolean checkLanguage,
...
 
-                if (same(this.language, replacedLanguage) &&
-                    same(this.script, replacedScript) &&
-                    same(this.region, replacedRegion) &&
-                    same(searchVariant, replacedRegion) &&
-                    replacedExtensions == null) {
+                if (    this.language.equals(replacedLanguage) &&
+                        this.script.equals(replacedScript) &&
+                        this.region.equals(replacedRegion) &&
+                        searchVariant.equals(replacedVariant) &&
```
markusicu, line 1496: can be null, right? if so, then please use `Objects.equals(searchVariant, replacedVariant)` to avoid a crash and handle nulls

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1446,10 +1508,21 @@ private boolean replaceLanguage(boolean checkLanguage,
...
+                if (replacedExtensions != null && !replacedExtensions.isEmpty()) {
+                    // DO NOTHING
+                    // UTS35 does not specifiy what should we do if we have extensions in the
+                    // replacement. Currently we know only the following 4 "BCP47 LegacyRules" have
+                    // extensions in them:
+                    //  i_default => en_x_i_default
+                    //  i_enochian => und_x_i_enochian
+                    //  i_mingo => see_x_i_mingo
+                    //  zh_min => nan_x_zh_min
+                    // But all of them are already change my LanguageTag before hitting this code.
```
markusicu, line 1525: I don't know what this line means. Please clarify.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1459,37 +1532,32 @@ private boolean replaceLanguage(boolean checkLanguage,
...
-                        // language [script] is one of the country in the
-                        // replacement, use that one.
-                        index = i;
-                        break;
-                    }
-                }
+                replacedRegion = replacement.contains(regionOfLanguageAndScript) ?
+                    regionOfLanguageAndScript : replacement.get(0);
+            } else {
+                replacedRegion = replacement.get(index);
```
markusicu, line 1551: Just use `replacement.get(0)` and remove the index variable.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1203,6 +1205,328 @@ public static String getKeywordValue(String localeID, String keywordName) {
...
+                return false;
+            }
+            int variantIndex = 0;
+            int variantSize = checkVariants ? variants.size() : 1;
+            // Since we may have more than one variant, we need to loop through
+            // them.
+            String searchLanguage = checkLanguage ? language : UNDEFINED_LANGUAGE;
+            String searchRegion = checkRegion ? region : null;
+            String searchVariant = null;
+            while (variantIndex < variantSize) {
```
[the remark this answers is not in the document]
markusicu: ping

actually best to also pull the variable declaration in here:
```
for (int variantIndex = 0; variantIndex < variantSize; ++variantIndex) {
```

markusicu (commented):
Just a couple more Java things and comments/tickets.

Now please also make changes in locid.cpp parallel to review changes you made in ULocale.java. Then I will take another look at that.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
FrankYFTang: ok https://unicode-org.atlassian.net/browse/ICU-21292

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
FrankYFTang: Code in LanguageTag ( see https://github.com/unicode-org/icu/blob/9971c663ff89e1bf78f9c4805daa407af8a023cb/icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LanguageTag.java#L130 ) would be run before calling into this replace function. So the code inside the replace function will never see these legacy tags.

markusicu: Ok, so it's just typos. --> But all of them are already changed by class LanguageTag before hitting this code.

FrankYFTang:
OK, I revise both the Java and C++ code per Markus' comment today. PTAL

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+    script = locale.getScript();
+    if (!notEmpty(script)) {
+        script = nullptr;
+    }
+    region = locale.getCountry();
+    if (!notEmpty(region)) {
+        region = nullptr;
+    }
+    const char* variantsStr = locale.getVariant();
+    const char* extensionsStr = uprv_strchr(locale.getName(), '@');
```
markusicu, line 1467: This will not work in EBCDIC because different EBCDIC codepages use different byte values for the at sign.
I know there is a helper function in the uloc*.cpp files for this. Please look and call that.

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+        // Nothing to search.
+        return false;
+    }
+    int32_t variant_size = checkVariants ? variants.size() : 1;
+    // Since we may have more than one variant, we need to loop through them.
+    const char* searchLanguage = checkLanguage ? language : "und";
+    const char* searchRegion = checkRegion ? region : nullptr;
+    const char* searchVariant = nullptr;
+    for (int32_t variant_index = 0;
+         variant_index < variant_size;
```
markusicu, line 1212: indent +8 not +5

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+        // We need to check which one match based on the language.
+        Locale l(language, nullptr, script);
+        l.addLikelySubtags(status);
+        const char* likelyRegion = l.getCountry();
+        CharString* item = nullptr;
+        if (likelyRegion != nullptr && uprv_strlen(likelyRegion) > 0) {
+            size_t len = uprv_strlen(likelyRegion);
+            const char* foundInReplacement = uprv_strstr(replacement, likelyRegion);
+            if (    foundInReplacement != nullptr &&
+                    (*foundInReplacement + len) != ' ' &&
```
markusicu, line 1324: I don't think this does what you intended. `(*foundInReplacement + len)` is the first character of the region code, and you add `len` to that character code.

I think you can just not check if the substring is a complete match, assuming that the likely-subtags data and the alias data both only contain well-formed region codes.

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+          out.append(SEP_CHAR, status);
+        }
+        variants.sort([](UElement e1, UElement e2) -> int8_t {
+            return uprv_strcmp(
+                (const char*)e1.pointer, (const char*)e2.pointer);
+        }, status);
+        for (int32_t i = 0; i < variants.size(); i++) {
+             U_ASSERT(((const char*)(
+                 (UVector*)variants.elementAt(i)))[0] != NULL_CHAR);
+             char uppercaseVariant[256];
```
markusicu, line 1424: This seems more fragile than necessary. Counter-proposal:
- before the variants loop, store `int32_t variantsStart = out.length();`
- in the loop, just append each variant subtag as is
- after the variants loop, `T_CString_toUpperCase(out.data() + variantsStart);`

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+    // Replace by using territoryAlias.
+    bool replaceTerritory(UVector& toBeFreed, UErrorCode& status);
+
+    // Replace by using scriptAlias.
+    bool replaceScript(UErrorCode& status);
+
+    // Replace by using variantAlias.
+    bool replaceVariant(UErrorCode& status);
+
+    // cleanup should be called before replace() return;
```
markusicu, line 1085: I think this is unnecessary. You only make a replacer as a temp object, and then it goes out of scope, right? The UVector should automatically removeAllElements in its destructor, so you should not need this explicit function. (Looks like you are not always calling it anyway.)

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+    if (!notEmpty(region)) {
+        region = nullptr;
+    }
+    const char* variantsStr = locale.getVariant();
+    const char* extensionsStr = uprv_strchr(locale.getName(), '@');
+    if (U_FAILURE(status)) {
+        cleanup();
+        return false;
+    }
+    char varaintsBuff[256];
```
markusicu, line 1472: typo --> variantsBuff

Also, please use a CharString not a raw buffer.

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+    outputToString(out, status);
+    if (extensionsStr != nullptr) {
+        out.append(extensionsStr, status);
+    }
+    if (U_FAILURE(status)) {
+        cleanup();
+        return false;
+    }
+    // If the tag is not changed, return.
+    if (uprv_strcmp(out.data(), locale.getName()) == 0) {
```
markusicu, line 1548: Does this work with extensions? The extensionsStr is in BCP syntax -- right? -- while extensions in getName() are in legacy syntax.

Maybe rare enough to just add a comment? Or add a TODO?

icu4c/source/common/locid.cpp
```diff
@@ -504,34 +508,1104 @@ Locale::operator==( const   Locale& other) const
...
+    U_ASSERT(gKnownCanonicalized != nullptr);
+    return uhash_geti(gKnownCanonicalized, locale) != 0;
+}
+
 }  // namespace
 
+// Function for testing.
+U_CAPI const char* const*
+ulocimp_getKnownCanonicalizedLocaleForTest(
+    int32_t* length) {
```
markusicu, line 1594: please pull this up on the previous line

icu4c/source/common/locid.cpp
FrankYFTang: The extension part of out.data() came from     
``` 
  if (extensionsStr != nullptr) {
        out.append(extensionsStr, status);
    }
```
and extensionsStr came from 
    const char* extensionsStr = uprv_strchr(locale.getName(), '@');
so both are in legacy syntax.

FrankYFTang:
PTAL

icu4c/source/common/locid.cpp
markusicu: Thanks for using a CharString, but please also fix the typo.

FrankYFTang: sorry, didn't catch that. Updated

icu4c/source/common/locid.cpp
FrankYFTang: ok I use ULOC_KEYWORD_SEPARATOR  but '@' is everywhere. If you really care about fixing for EBCDIC then please file a separate bug for that. I don't think we should snowball the required fix for that into this PR.

icu4c/source/common/locid.cpp
FrankYFTang: ok

icu4c/source/common/locid.cpp
FrankYFTang: use locale_getKeywordsStart now

markusicu:
changes lgtm, please squash

markusicu (approved):
lgtm

PR #1255, opened 2020-08-22
ICU-21242 rephrase documentation using term master 

sven-oly:
https://unicode-org.atlassian.net/browse/ICU-21242

### DIFF
--- a/icu4c/source/data/cldr-icu-readme.txt
+++ b/icu4c/source/data/cldr-icu-readme.txt
@@ -226,7 +226,7 @@ make check 2>&1 | tee /tmp/icu4c-newData-makeCheck.txt
 # Repeat steps 4-7 until there are no errors.
 
 # 9. You can also run the make check tests in exhaustive mode (these will also
-# be run automatically on the master branch after the merge resulting from this
+# be run automatically on the default branch after the merge resulting from this
 # integration):
 
 cd $ICU4C_DIR/source
--- a/icu4c/source/i18n/unicode/rbnf.h
+++ b/icu4c/source/i18n/unicode/rbnf.h
@@ -297,7 +297,7 @@ enum URBNFRuleSetTag {
  *   </tr>
  *   <tr>
  *     <td>x.0:</td>
- *     <td>The rule is a <em>master rule</em>. If the full stop in
+ *     <td>The rule is a <em>default rule</em>. If the full stop in
  *     the middle of the rule name is replaced with the decimal point
  *     that is used in the language or DecimalFormatSymbols, then that rule will
  *     have precedence when formatting and parsing this rule. For example, some
@@ -332,9 +332,9 @@ enum URBNFRuleSetTag {
  * algorithms: If the rule set is a regular rule set, do the following:
  *
  * <ul>
- *   <li>If the rule set includes a master rule (and the number was passed in as a <tt>double</tt>),
- *     use the master rule.&nbsp; (If the number being formatted was passed in as a <tt>long</tt>,
- *     the master rule is ignored.)</li>
+ *   <li>If the rule set includes a default rule (and the number was passed in as a <tt>double</tt>),
+ *     use the default rule.&nbsp; (If the number being formatted was passed in as a <tt>long</tt>,
+ *     the default rule is ignored.)</li>
  *   <li>If the number is negative, use the negative-number rule.</li>
  *   <li>If the number has a fractional part and is greater than 1, use the improper fraction
  *     rule.</li>
@@ -393,7 +393,7 @@ enum URBNFRuleSetTag {
  *   </tr>
  *   <tr>
  *     <td></td>
- *     <td>in fraction or master rule</td>
+ *     <td>in fraction or default rule</td>
  *     <td>Isolate the number's fractional part and format it.</td>
  *   </tr>
  *   <tr>
@@ -425,7 +425,7 @@ enum URBNFRuleSetTag {
  *   </tr>
  *   <tr>
  *     <td></td>
- *     <td>in fraction or master rule</td>
+ *     <td>in fraction or default rule</td>
  *     <td>Isolate the number's integral part and format it.</td>
  *   </tr>
  *   <tr>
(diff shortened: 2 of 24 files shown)


markusicu:
Craig, I don't know how or why, but you are targeting a bad branch again. Please make the PR go to unicode-org:master, not to unicode-org:ICU-21242.

icu4c/source/i18n/unicode/rbnf.h
```diff
@@ -297,7 +297,7 @@ enum URBNFRuleSetTag {
  *   </tr>
  *   <tr>
  *     <td>x.0:</td>
- *     <td>The rule is a <em>master rule</em>. If the full stop in
+ *     <td>The rule is a <em>primary rule</em>. If the full stop in
```
markusicu, line 300: I would like @grhoten to chime in on renaming "master rule" in RBNF. Maybe even @richgillam has an opinion.

markusicu:
I assume that there will also be Java files and Markdown files to modify.

icu4c/source/i18n/unicode/rbnf.h
grhoten: Can we replace "master" with "default" rule in this entire header? It's not really master nor primary. We don't really have tertiary types of rules, which is why I don't want to say primary nor secondary.

grhoten: Another reason to not say primary is that the other rules are the primary ones and not this one.

richgillam: Wow, this brings back memories.  Are we talking about the `.:` rule that gets used to split the number into integral and fractional parts?  If so, I'm fine with changing it from "master", and I agree with George that "primary" doesn't feel right either.  Maybe because this is a very specialized rule that's just used to split the number into integral and fractional parts, we should give it a specific name, such as "decimal point rule."

Of course, if I've misunderstood what we're talking about here, feel free to ignore me.  :-)

grhoten: @richgillam actually your information is a little out of date. This is in reference to selecting alternate decimal points. So you can say “point” or “comma” without duplicating the rules into a bunch of locales.

The full stop is for all locales, unless an alternate one matches, like a comma. This is important for Spanish. I added this bit of functionality.

richgillam: Thanks for bringing me up to date.  Based on what you're saying, "master rule" and "primary rule" both sound even worse and "decimal point rule" sounds better.  But I'll go with whatever.

grhoten: Well these are decimal point rules. Just one of them is the default (this one with the full stop) in the absence of a localized one.

richgillam: Okay, that makes sense.  Thanks for the clarifications.

markusicu: So you recommend `The rule is a <em>default rule</em>.` ?

grhoten: Yes, "default rule" is a good name for now.

sven-oly: Done. Thx for your suggestion.

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedNumberFormat.java
```diff
@@ -275,7 +275,7 @@
  *   <tr>
  *     <td style="width: 5%; vertical-align: top;"></td>
  *     <td style="width: 8%; vertical-align: top;">x.0:</td>
- *     <td valign="top">The rule is a <em>master rule</em>. If the full stop in
+ *     <td valign="top">The rule is a <em>primary rule</em>. If the full stop in
```
grhoten, line 278: My previous comments about rbnf.h applies to all of the RBNF Java code too. The documentation and changes should be kept in sync. It should be "default" and not "master" nor "primary".

sven-oly: Done. Thx.

grhoten:
Let me be clearer. RBNF (Rule Based Number Format) covers the following files. Please use "default" instead of "primary" in RBNF so that there isn't any confusion about mismatched terminology.

```
nfrs.cpp
nfrule.cpp
nfrule.h
nfsubs.cpp
rbnf.h
NFRule.java
NFRuleSet.java
RuleBasedNumberFormat.java
```

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedNumberFormat.java
sven-oly: OK, I think I've handled this better now. Thanks for your patience explanation.

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -226,7 +226,7 @@ make check 2>&1 | tee /tmp/icu4c-newData-makeCheck.txt
 # Repeat steps 4-7 until there are no errors.
 
 # 9. You can also run the make check tests in exhaustive mode (these will also
-# be run automatically on the master branch after the merge resulting from this
+# be run automatically on the primary branch after the merge resulting from this
```
markusicu, line 229: It's still the "master branch" until GitHub gives us better tools to change that.

However, what happens in git/GitHub is that they will allow us to change the name of the default branch more easily, and I think we could use that:
`# be run automatically on the default branch after the merge resulting from this`

icu4c/source/test/testdata/rbbitst.txt
```diff
@@ -32,7 +32,7 @@
 #   [ICU4C]   source/test/testdata/rbbitst.txt
 #   [ICU4J]   main/tests/core/src/com/ibm/icu/dev/test/rbbi/rbbitst.txt
 #
-# ICU4C's copy is the master.  If any changes are made to ICU4J's copy, make sure they
+# ICU4C's copy is the primary.  If any changes are made to ICU4J's copy, make sure they
```
markusicu, line 35: please add the word "one" --> "ICU4C's copy is the primary one."

icu4c/source/data/cldr-icu-readme.txt
sven-oly: Good suggestion. Done.

icu4c/source/test/testdata/rbbitst.txt
sven-oly: Done

PR #1256, opened 2020-08-24
ICU-21157 Fix test code

FrankYFTang:
Change from using u_austrcpy to aescstrdup

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21157
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -6707,28 +6707,28 @@ static void checkTerminating(const char* locale, const char* inLocale)
     status = U_ZERO_ERROR;
     int32_t length = uloc_getDisplayName(
         locale, inLocale, buff, preflight_length + 1, &status);
-    char cbuff[256];
-    u_austrcpy(cbuff, buff);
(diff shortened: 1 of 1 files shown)


jefgen:
Thanks you for the really quick fix! :)

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -6742,7 +6742,7 @@ static void checkTerminating(const char* locale, const char* inLocale)
     status = U_ZERO_ERROR;
     length = uloc_getDisplayName(
         locale, inLocale, buff, preflight_length, &status);
-    u_austrcpy(cbuff, buff);
+    u_austrncpy(cbuff, buff, length);
```
markusicu, line 6745: The other problem with all of these is that the test code is reading from `buff` even when `U_FAILURE(status)` which means that there may be nothing but garbage in the buffer. It would be better to do something like

```
const char *cbuff = U_SUCCESS(status) ? aescstrdup(buff, length) : "(undefined when failure)";
```

PR #1257, opened 2020-08-24
ICU-21246 Handle kilogram SI prefix correctly

hugovdm:
Fix kilogram parsing: ignore 'kilogram' as a stem, we have 'gram'.

Failures in the added unit test before the fix: withSIPrefix resulted
in 'microkilogram', and kilogram's prefix was considered to be
"ONE" (i.e. 10^0).

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21246
- [ ] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -14,6 +14,7 @@
 
 #include "charstr.h"
 #include "cmemory.h"
+#include "cstring.h"
 #include "measunit_impl.h"
 #include "resource.h"
 #include "uarrsort.h"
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -240,6 +241,13 @@ void U_CALLCONV initUnitExtras(UErrorCode& status) {
     TableKeysSink identifierSink(gSimpleUnits, simpleUnitsCount);
     ures_getAllItemsWithFallback(unitsBundle.getAlias(), "convertUnits", identifierSink, status);
     for (int i = 0; i < simpleUnitsCount; i++) {
+        if (uprv_strcmp(gSimpleUnits[i], "kilogram") == 0) {
```
sffc, line 244: I have a slight preference for putting this check inside of the resource sink (TableKeysSink), such that `"kilogram"` doesn't get added to gSimpleUnits in the first place.

hugovdm: "TableKeysSink" being conceived of as generic and reusable ("just collect the keys from a table"), this would be making it more "custom" (and less likely to be promoted as a possibly shared utility). I'll think on this a bit longer - it would at least need a new narrower name.

PR #1260, opened 2020-08-25
ICU-21035 Replace backward compatibility locale_getKeywords() function.

roubert:
By updating the last 3 callers to do dynamic memory allocation instead,
the fixed memory buffer function becomes obsolete.

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -102,12 +102,6 @@ typedef enum ELocalePos {
     eMAX_LOCALES
 } ELocalePos;
 
-U_CFUNC int32_t locale_getKeywords(const char *localeID,
-            char prev,
-            char *keywords, int32_t keywordCapacity,
-            UBool valuesToo,
-            UErrorCode *status);
-
 U_CDECL_BEGIN
 //
 // Deleter function for Locales owned by the default Locale hash table/
@@ -1416,8 +1410,6 @@ UnicodeKeywordEnumeration::~UnicodeKeywordEnumeration() = default;
 StringEnumeration *
 Locale::createKeywords(UErrorCode &status) const
 {
-    char keywords[256];
-    int32_t keywordCapacity = sizeof keywords;
     StringEnumeration *result = NULL;
 
     if (U_FAILURE(status)) {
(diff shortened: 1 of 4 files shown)


icu4c/source/common/locid.cpp
```diff
@@ -102,11 +102,12 @@ typedef enum ELocalePos {
     eMAX_LOCALES
 } ELocalePos;
 
-U_CFUNC int32_t locale_getKeywords(const char *localeID,
-            char prev,
-            char *keywords, int32_t keywordCapacity,
-            UBool valuesToo,
-            UErrorCode *status);
+void U_EXPORT2
+_getKeywords(const char *localeID,
```
markusicu, line 106: Please keep the "U_CFUNC" for a function that is shared across .cpp files. I don't remember if it *also* needs U_EXPORT2 -- looks like it worked without that.

https://unicode-org.github.io/icu/userguide/dev/codingguidelines.html#function-declarations

icu4c/source/common/uloc.cpp
```diff
@@ -598,7 +592,7 @@ compareKeywordStructs(const void * /*context*/, const void *left, const void *ri
     return uprv_strcmp(leftString, rightString);
 }
 
-static void
+void U_EXPORT2
```
markusicu, line 595: U_CFUNC, and maybe not U_EXPORT2. Be consistent decl vs impl.

icu4c/source/common/uloc.cpp
```diff
@@ -598,7 +592,7 @@ compareKeywordStructs(const void * /*context*/, const void *left, const void *ri
...
 _getKeywords(const char *localeID,
```
markusicu, line 596: For a function that is no longer file-local, please also give it a uloc_ prefix.

icu4c/source/common/locid.cpp

icu4c/source/common/uloc.cpp
roubert: Done.

icu4c/source/common/uloc.cpp
roubert: Done.

PR #1261, opened 2020-08-25
ICU-21247 Move #include "unicode/localpointer.h" into U_SHOW_CPLUSPLUS_API block

gvictor:
https://unicode-org.atlassian.net/browse/ICU-21247

### DIFF
--- a/icu4c/source/common/unicode/ubidi.h
+++ b/icu4c/source/common/unicode/ubidi.h
@@ -21,7 +21,10 @@
 
 #include "unicode/utypes.h"
 #include "unicode/uchar.h"
+
+#if U_SHOW_CPLUSPLUS_API
 #include "unicode/localpointer.h"
+#endif   // U_SHOW_CPLUSPLUS_API
 
 /**
  *\file
--- a/icu4c/source/common/unicode/ubiditransform.h
+++ b/icu4c/source/common/unicode/ubiditransform.h
@@ -21,7 +21,10 @@
 #include "unicode/utypes.h"
 #include "unicode/ubidi.h"
 #include "unicode/uchar.h"
+
+#if U_SHOW_CPLUSPLUS_API
 #include "unicode/localpointer.h"
+#endif   // U_SHOW_CPLUSPLUS_API
 
 /**
  * \file
--- a/icu4c/source/common/unicode/ubrk.h
+++ b/icu4c/source/common/unicode/ubrk.h
@@ -13,7 +13,10 @@
 #include "unicode/utypes.h"
 #include "unicode/uloc.h"
 #include "unicode/utext.h"
+
+#if U_SHOW_CPLUSPLUS_API
 #include "unicode/localpointer.h"
+#endif   // U_SHOW_CPLUSPLUS_API
 
 /**
  * A text-break iterator.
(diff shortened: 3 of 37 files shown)


markusicu:
I don't like these fairly extensive changes for what seems like an aesthetic issue. I added a comment on the ticket.

jefgen:
Hmm, maybe I am missing something, but shouldn't the contents of the header file `localpointer.h` just be empty if `U_SHOW_CPLUSPLUS_API` is false?

The whole file is guarded with `#if U_SHOW_CPLUSPLUS_API`:
https://github.com/unicode-org/icu/blob/master/icu4c/source/common/unicode/localpointer.h#L43

gvictor:
Re @jefgen: As I mentioned in the ticket, it forces me to export the 500-line localpointer.h though the header does nothing on a C compiler.

Initially, I didn't plan to report this, but I found that it's fairly common to guard the #include with a U_SHOW_CPLUSPLUS_API flag in the code base, e.g.
https://github.com/unicode-org/icu/blob/master/icu4c/source/common/unicode/utext.h#L144

btw, it does not cause me a big trouble or anything, but took me (and probably other developers) a few minutes to understand why the headers need to exist in the exported header folder...

markusicu:
Discussed in the team, agreed to do this. Will review.

jefgen:
If this is for all the _public headers_ then it seems like we should also do the ones in the `io` library as well.

It looks like there is one in `ustdio.h`:
```
icu4c/source/io/unicode/ustdio.h:
  32  #include "unicode/utrans.h"
  33: #include "unicode/localpointer.h"
  34  #include "unicode/unum.h"
```

One other thought:
It would be nice if we could have some automated way to check for this...
If someone adds a new header file in the future, then it might not have the `#include "unicode/localpointer.h"` guarded with the check for `#if U_SHOW_CPLUSPLUS_API`.
(Manually checking could be a BRS task I supposed -- but it would be a bit tedious.)

gvictor:
I found this issue manually. 

Regarding automation idea, to export headers, I use to clang to traverse the #include dependency with U_SHOW_CPLUSPLUS_API=0 for each non-C++ header. I exclude C++ headers by checking the list icu4c/source/test/hdrtst/cxxfiles.txt. If I check the dependency against cxxfiles.txt again, I will probably discover this issue.  

If I remember correctly, icu4c/source/test/hdrtst/cxxfiles.txt has been removed in master?

markusicu:
> Regarding automation idea, to export headers, ...
> If I remember correctly, icu4c/source/test/hdrtst/cxxfiles.txt has been removed in master?

Right, we don't have such a list file any more. We would need to bring it back and maintain it again (ugh), or have a test see if the header file says `\brief C API` vs. C++. @jefgen I am satisfied with this PR as is -- if you think we should add an automated test for compiling all public C headers with C++ headers not present, could you please submit a new ticket?

jefgen:
Thanks @gvictor for updating the io header too.

> We would need to bring it back and maintain it again (ugh), or have a test see if the header file says `\brief C API` vs. C++. @jefgen I am satisfied with this PR as is -- if you think we should add an automated test for compiling all public C headers with C++ headers not present, could you please submit a new ticket?

I was thinking that maybe we could add an automated test for just the header file `localpointer.h` -- but I think that you're right we'd probably want to do it for *all* C vs C++ headers. I don't know if it's worth really adding/maintaining such an automated test though, so this is likely fine as-is.

PR #1264, opened 2020-08-26
ICU-21232 Improve performance of DecimalFormat#parse on long numbers

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21232
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/test/intltest/numfmtst.cpp
+++ b/icu4c/source/test/intltest/numfmtst.cpp
@@ -247,6 +247,7 @@ void NumberFormatTest::runIndexedTest( int32_t index, UBool exec, const char* &n
   TESTCASE_AUTO(Test20961_CurrencyPluralPattern);
   TESTCASE_AUTO(Test21134_ToNumberFormatter);
   TESTCASE_AUTO(Test13733_StrictAndLenient);
+  TESTCASE_AUTO(Test21232_ParseTimeout);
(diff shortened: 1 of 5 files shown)


icu4c/source/test/intltest/numfmtst.cpp
```diff
@@ -10033,4 +10034,29 @@ void NumberFormatTest::Test13733_StrictAndLenient() {
...
+        assertEquals("Somewhat long input of digits", 1280, input.length());
+    } else {
+        for (int32_t i = 0; i < 12; i++) {
+            input.append(input);
+        }
+        assertEquals("Very long input of digits", 163840, input.length());
+    }
+    Formattable result;
+    df.parse(input, result, status);
+    // Should not hang
```
gnrunge, line 10059: How to verify 'Should not hang'? The issue from the ticket was a timeout (after 60 secs). Can the code really hang here or is the real error symptom that processing takes unexpectedly long? Would it make sense to measure the time and raise an issue after 60 secs of elapse time? Somewhat imprecise, of course, depends on machine and machine load.

sffc: I'm not sure of the best way to enforce this.  In exhaustive mode, on my machine, the test takes ~30 seconds before the change, and ~1 second after the change.

PR #1267, opened 2020-08-28
ICU-21240 Removing unused makefile files and updating docs.

hagbard:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21240
- [ ] Issue accepted
- [ ] Tests included


hagbard:
If this isn't in by Monday someone else will have to do it (I'm off for 30 days).

PR #1270, opened 2020-08-28
ICU-21251 release icu4j.jar/utilities.jar to GH Maven

srl295:
ICU-21251

- on a release (created from Github), publish to
<https://github.com/unicode-org/icu/packages>
- also fix a doc bug in DecimalFormat
- this is intended for CLDR use to support CLDR-11585 maven

No effect on current build systems, no effect on commit or PRs.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -1,0 +1,40 @@
+name: Publish icu4j.jar/utilities.jar to GH Maven
+on:
+  release:
+    types: [created]
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          lfs: true
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: icu4j ant releaseCLDR
+        run: ant -noinput releaseCLDR -f icu4j/build.xml
+      - name: deploy it
+        run: |
+          MYVERSION=$(sed -n -e 's%^\s*maven.pom.ver[ =]*%%p' < icu4j/build.properties)-$(echo ${GITHUB_REF} | cut -d/ -f3- | tr ./ _-)
+          echo Github Ref ${GITHUB_REF} @ ${GITHUB_SHA}, version ${MYVERSION}
+          mvn deploy:deploy-file -DgroupId=com.ibm.icu \
+            -DartifactId=icu4j-for-cldr \
+            -Dversion=${MYVERSION} \
+            -Dpackaging=jar \
+            -Dfile=icu4j/release_cldr/icu4j.jar \
+            -Dsources=icu4j/release_cldr/icu4j-src.jar \
+            -DrepositoryId=github \
+            -Durl=https://maven.pkg.github.com/${GITHUB_REPOSITORY}
+          mvn deploy:deploy-file -DgroupId=com.ibm.icu \
+            -DartifactId=utilities-for-cldr \
+            -Dversion=${MYVERSION} \
+            -Dpackaging=jar \
+            -Dfile=icu4j/release_cldr/utilities.jar \
+            -Dsources=icu4j/release_cldr/utilities-src.jar \
+            -DrepositoryId=github \
+            -Durl=https://maven.pkg.github.com/${GITHUB_REPOSITORY}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+# Note: can use 0.0.0-${GITHUB_SHA} as the version for hash-based.
+# i.e. build on every commit
--- a/icu4j/.gitignore
+++ b/icu4j/.gitignore
@@ -1,0 +1,1 @@
+/release_cldr
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormat.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormat.java
@@ -2015,10 +2015,10 @@ public class DecimalFormat extends NumberFormat {
   }
 
   /**
-   * {@icu} Constant for {@link #setMinimumGroupingDigits()} to specify display
+   * {@icu} Constant for {@link #setMinimumGroupingDigits(int)} to specify display
    * grouping using the default strategy for all locales.
    *
-   * @see #setMinimumGroupingDigits
+   * @see #setMinimumGroupingDigits(int)
    * @see #MINIMUM_GROUPING_DIGITS_MIN2
    * @category Separators
    * @provisional This API might change or be removed in a future release.
@@ -2027,11 +2027,11 @@ public class DecimalFormat extends NumberFormat {
   public static final int MINIMUM_GROUPING_DIGITS_AUTO = -2;
 
   /**
-   * {@icu} Constant for {@link #setMinimumGroupingDigits()} to specify display
+   * {@icu} Constant for {@link #setMinimumGroupingDigits(int)} to specify display
    * grouping using locale defaults, except do not show grouping on values smaller than
    * 10000 (such that there is a minimum of two digits before the first separator).
    *
-   * @see #setMinimumGroupingDigits
+   * @see #setMinimumGroupingDigits(int)
    * @see #MINIMUM_GROUPING_DIGITS_AUTO
    * @category Separators
    * @provisional This API might change or be removed in a future release.


srl295:
If you want to kick the tires a bit, check out: https://github.com/srl295/icu/packages

.github/workflows/maven.yaml
```diff
@@ -0,0 +1,40 @@
...
+          java-version: 1.8
+      - name: icu4j ant releaseCLDR
+        run: ant -noinput releaseCLDR -f icu4j/build.xml
+      - name: deploy it
+        run: |
+          MYVERSION=$(echo ${GITHUB_REF} | cut -d/ -f3 | tr . _)
+          echo Github Ref ${GITHUB_REF} @ ${GITHUB_SHA}, version ${MYVERSION}
+          mvn deploy:deploy-file -DgroupId=com.ibm.icu \
+            -DartifactId=icu4j \
+            -Dversion=0.0.0-${MYVERSION} \
```
echeran, line 23: Based on the intention described in the [Jira comment](https://unicode-org.atlassian.net/browse/ICU-21251?focusedCommentId=157956), I think the ["typical Maven way to do it"](https://maven.apache.org/guides/getting-started/index.html#What_is_a_SNAPSHOT_version) is for the development version number to be "<semver>-SNAPSHOT", and the release artifact version number is <semver>.

FWIW, in a previous workplace which used git, one of the teams introduced the git hash into the Maven artifact version numbers, which required downstream teams to deal with this convention.  Many of the senior developers found the benefits to be minimal and outweighed by the hassle caused during development -- every time they made a new git commit locally and wanted to re-upload the resulting binary to test out, they had to manually find the new binary file name and upload, and their build output directory would be cluttered with several development snapshot build files.  They grumbled that the git hash version numbering came into existence as a band-aid solution to immediate problems instead of properly solving the root cause by decomplecting a complex spaghetti architecture.  That example is much more extreme than what is going on here, though, so not sure how relevant that feels.

.github/workflows/maven.yaml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          java-version: 1.8
+      - name: icu4j ant releaseCLDR
+        run: ant -noinput releaseCLDR -f icu4j/build.xml
+      - name: deploy it
+        run: |
+          MYVERSION=$(echo ${GITHUB_REF} | cut -d/ -f3 | tr . _)
+          echo Github Ref ${GITHUB_REF} @ ${GITHUB_SHA}, version ${MYVERSION}
+          mvn deploy:deploy-file -DgroupId=com.ibm.icu \
+            -DartifactId=icu4j \
```
echeran, line 22: If we aren't able to directly use the "vanilla" version of ICU4J directly in CLDR -- meaning that CLDR's ICU4J has any sort of differences that need to be maintained -- then I think it would be cleaner to consider this as a distinct entity in Maven, which means using a distinct value in `artifactId`.  (ex: `cldr-icu4j`).

Extending the version number as is currently configured is certainly one way to do it, too.  I feel like it could be slightly error-prone (complexity on user side to due to required mental overhead) -- see subsequent comment.

Using the [classifier field of a Maven artifact](https://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html) is a third option, which could be useful is you want to keep the connection to ICU4J and treat this as a variant.  This feature isn't commonly used, but maybe it makes sense here if you want to preserve the semantic relationship to ICU4J both in name and in version number.

.github/workflows/maven.yaml
```diff
@@ -0,0 +1,40 @@
...
+          lfs: true
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: icu4j ant releaseCLDR
+        run: ant -noinput releaseCLDR -f icu4j/build.xml
+      - name: deploy it
+        run: |
+          MYVERSION=$(echo ${GITHUB_REF} | cut -d/ -f3 | tr . _)
+          echo Github Ref ${GITHUB_REF} @ ${GITHUB_SHA}, version ${MYVERSION}
```
echeran, line 20: Where does `MYVERSION` get set?  I may have forgotten it if I saw it before.  I'm curious about how it persists / propagates across GH Action workflows.

srl295: ☝️  it gets set on the prior line 19.  It's literally the git tag (ref) but with `.` changed to `_`

so if you tag `cldr-9.8` GITHUB_REF will be something like `refs/tags/cldr-9.8` (don't quote me on the refs/tags part) and this sets `MYVERSION=cldr-9_8`

(PS thanks for looking at this. This is a greenfield here, so appreciate the input)

.github/workflows/maven.yaml
srl295: > If we aren't able to directly use the "vanilla" version of ICU4J directly in CLDR -- meaning that CLDR's ICU4J has any sort of differences that need to be maintained 

It's not CLDR's icu4j (no code change), but it is an arbitrary snapshot. We absolutely use the vanilla version of ICU4J in CLDR. However, consider what has to happen right now, September 8th - such as this PR https://github.com/unicode-org/cldr/pull/702/files#diff-42a78bd8fb3858e8442d2c9f16efb057  by @pedberg-icu  - that is the target for this ICU PR.

CLDR and ICU are both revving their versions.  ICU will wait for new CLDR data, and CLDR needs a new ICU that supports the new CLDR features so that CLDR can use that ICU…  yes, interlock.

So what happens in this PR is that an arbitrary snapshot of ICU4J (including the utilities.jar 'extras') is checked into CLDR's repo. Now, at some future time, a release version of ICU4J could be checked in (plus the right version of utilities.jar).  There's no special ICU version, other than than being a special snapshot.

  This build would get kicked off on a 'normal' release tag as well.   There is an accompanying utilities.jar, though, which is not part of a normal build.

So, when ICU 68.1 is released, there will be a `release-68-1` tag and official binaries (uploaded to maven by other means).  But this workflow would also create an icu4j.jar which would be the equivalent of the official 68.1 release (though perhaps unsigned). 

**I could see using the classifier**, to mark all of this stuff as pertaining to CLDR and not intended for general use.  

I'll reply about the version in your next comment.

.github/workflows/maven.yaml
srl295: (^ I added some backticks to your comment)
> `<semver…>` 

Yes, but there's no semver that makes a lot of sense here. It really is an arbitrary snapshot. We could make a completely unrelated semver just for this purpose I suppose.

> git hash into the Maven artifact version numbers, which required downstream teams to deal with this convention

in this case, the same person (such as @pedberg-icu  ) is cutting the ICU4J version on the ICU side… and referencing it on the CLDR side. A git hash could work.

Maybe a structured tag could make some sense.  I assume the 'on: release: … ` can take a regex or glob? Then on the ICU side we create releases named:`cldr/38.0.0` or `cldr/38.0.1` etc…  and title it "snapshot 0 for CLDR 38…". 

Then, this workflow can pickup the maven version from there:  "38.0.0-SNAPSHOT" with the "cldr" classifier. I think that will make it not interfere with the regular artifacts, even if we had CLDR v68. 

ICU folks, any thoughts about the tag naming structure?

icu4j/.gitignore
```diff
@@ -0,0 +1 @@
+/release_cldr
```
srl295, line 1: this build creates this dir, i think it's an oversight that this wasn't included..

icu4j/main/classes/core/src/com/ibm/icu/text/DecimalFormat.java
```diff
@@ -2015,10 +2015,10 @@ public synchronized void setMinimumGroupingDigits(int number) {
   }
 
   /**
-   * {@icu} Constant for {@link #setMinimumGroupingDigits()} to specify display
+   * {@icu} Constant for {@link #setMinimumGroupingDigits(int)} to specify display
    * grouping using the default strategy for all locales.
    *
-   * @see #setMinimumGroupingDigits
+   * @see #setMinimumGroupingDigits(int)
```
srl295, line 2021: this was broken by something else… I fix it here so that the javadoc build passes.

srl295:
@pedberg-icu @markusicu and others… My presumption in this PR is that there could be a 'release' to create a tag for the version of ICU. Something like `cldr/38.0.0` or similar?

srl295:
@echeran and all: See https://github.com/unicode-org/cldr/pull/704 for a use case, a PR for moving CLDR to add a parallel Maven build.

pedberg-icu:
Sorry, I know nothing about maven or yaml...

srl295:
OK, reworking this.
- the classifier is a great idea, but doesn't let you use the autogenerated pom.xml
- instead, i changed the artifacts to `icu4j-for-cldr` and `utilities-for-cldr`— should make it celar
- I'm now using ICU's `maven.pom.ver = 68.1-SNAPSHOT` from build.properties as the basis for the version number. So the version will have some relation to the ICU4J reality.
- example: https://github.com/srl295/icu/releases/tag/cldr%2F2020-00-01 - tag `cldr/2020-00-01`
> Github Ref refs/tags/cldr/2020-00-03 @ cedf0bbb8296f84c1f354db19e2f7ef3228355ae, version 68.1-SNAPSHOT-cldr-2020-00-03
- assets: icu4j-for-cldr and utilities-for-cldr at https://github.com/srl295/icu/packages

echeran (approved):
LGTM.  I like the artifact name changes to reduce the chance that a external inadvertently starts using them without knowing the internal-focused purpose.

srl295:
@markusicu @macchiati @yumaoka @pedberg ? Can we move this forward?

markusicu (approved):
lgtm

FrankYFTang (approved):

srl295:
@pedberg-icu Now that this is merged ( 🎉  ) let's get it actually used in CLDR.  Do you have a projected next cut of ICU to go into CLDR? Or could we make an arbitrary one ( such as tagging current ICU HEAD as `cldr/2020-09-21`  ?)

pedberg-icu:
> Do you have a projected next cut of ICU to go into CLDR?

I was kinda waiting for https://github.com/unicode-org/icu/pull/1349 to be merged, then after that I thought I would update the ICU4J libs in CLDR again.

srl295:
> > Do you have a projected next cut of ICU to go into CLDR?
> 
> I was kinda waiting for #1349 to be merged, then after that I thought I would update the ICU4J libs in CLDR again.

I assume you are going to update cldr 38-maint also?

OK, once #1349 is merged, would you be able to create the ICU tag `cldr/2020-09-__` with whatever the date is? And then do your usual update.  Or, you can just do your regular update, and I'll create a tag from the hash you end up with.

pedberg-icu:
> I assume you are going to update cldr 38-maint also?

Yes, that would be part of the next CLDR-ICU integration (alpha3 or beta).

> OK, once #1349 is merged, would you be able to create the ICU tag `cldr/2020-09-__` with whatever the date is? And then do your usual update.

Sure.

srl295:
> > I assume you are going to update cldr 38-maint also?
> 
> Yes, that would be part of the next CLDR-ICU integration (alpha3 or beta).
> 
> > OK, once #1349 is merged, would you be able to create the ICU tag `cldr/2020-09-__` with whatever the date is? And then do your usual update.
> 
> Sure.

Thank you. I will do the parallel update for maven on the CLDR side, it will be different the first time, but afterwards will only be changing the version string.

echeran:
As you may have noticed, PR #1349 is only for the Java side of things.  If that looks good and gets merged, I'll work next on the C++ side of things.  Out of curiosity, does the ICU-CLDR integration work depend mostly on ICU4J, or will the lack of C++ port also be blocking your integration testing?

srl295:
> As you may have noticed, PR #1349 is only for the Java side of things. If that looks good and gets merged, I'll work next on the C++ side of things. Out of curiosity, does the ICU-CLDR integration work depend mostly on ICU4J, or will the lack of C++ port also be blocking your integration testing?

CLDR's tooling is entirely Java, and therefor only consumes icu4j. But, ICU4C's tests still ought to pass when the integration happens.

pedberg-icu:
OK, I tagged ICU @3d8ae5eb6 as "cldr/2020-09-22" and have a PR for integrating the corresponding ICU4J libs into CLDR: https://github.com/unicode-org/cldr/pull/731

srl295:
Thanks!

>

PR #1274, opened 2020-08-29
ICU-21225 Conflicting documentation for stand alone quarter

robertgmelo:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21225
- [x] Issue accepted


robertgmelo:
I've addressed the conflict between regular vs stand-alone form.

In the issue, the user also asked if documentation could also explain why stand-alone matter and examples. Since we know stand-alone may varies from language to language, I I've added a link referencing to CLDR TR35, which brings who have stand-alone, and explain the distinction, instead of include more information in the guide.

With that, we may satisfy both requests.

Once approved, we provide the link of new User Guide (github).

markusicu (approved):
lgtm
Thanks Robert!

PR #1265, opened 2020-08-28
ICU-21051 Add Locale::createCanonical performance test

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21051
- [X] Updated PR title and link in previous line to include Issue number
- [X] Issue accepted
- [X] Tests included
- [X] Documentation is changed or added

This is the first step toward fixing the performance issue- by adding a simple performance test so we can measure the two builds.

PR #1276, opened 2020-08-29
ICU-13416 change Armenian (hy) uppercase/titlecase of և ligature ech-…

markusicu:
…yiwn

https://unicode-org.atlassian.net/browse/ICU-13416
See especially my comments there from 2020 with links to feedback, discussions, and references.

Note that the data builder removes the full case mappings when one of them is conditional. I could override that to minimize the change in the data file, but the mappings would no longer be used.

The runtime code hardcodes all mappings for a character that has the conditional-mappings flag, except when there is no mapping (as here for lowercasing) or when the default is a simple (1:1) case mapping (not the case here).


markusicu:
FYI The copyright test failure is from PR #1265

PR #1277, opened 2020-08-29
ICU-21253 remove std from max_align_t

neheb:
Causes compilation failure under CentOS 7 whose GCC (4.8.2) has a
cstddef header that does not place max_align_t under the std namespace.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

https://unicode-org.atlassian.net/browse/ICU-21253

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21253
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/uarrsort.cpp
+++ b/icu4c/source/common/uarrsort.cpp
@@ -37,7 +37,7 @@ enum {
 };
 
 static constexpr int32_t sizeInMaxAlignTs(int32_t sizeInBytes) {
-    return (sizeInBytes + sizeof(std::max_align_t) - 1) / sizeof(std::max_align_t);
+    return (sizeInBytes + sizeof(max_align_t) - 1) / sizeof(max_align_t);
 }
 
 /* UComparator convenience implementations ---------------------------------- */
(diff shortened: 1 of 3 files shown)


jefgen:
Thanks for filing a ticket and creating this PR.

ICU4C requires a minimum of C++11 and C99 in order to build as of ICU version 59.1.
(See the page here for the C++11 change notes: http://site.icu-project.org/download/59#TOC-ICU4C-Specific-Changes ).

Looking at the change, it removes the `std::` prefix from the `max_align_t`.

This means that the code is no longer is using the C++11 definition and instead is now using the definition from C.

However, `max_align_t` is actually not present and/or defined in C99 as it was only added to C in the C11 standard.

https://en.cppreference.com/w/c/types/max_align_t
The C99 standard (draft): http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf
The C11 standard (draft): http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf

Unfortunately, this change would mean that ICU4C is now depending on C11 features -- which would break the C99 only requirement. (Sort of as an aside, this change is also somewhat odd in the sense that it's depending on a C11 feature in order to work around incomplete support for C++11.)

Due to this, the discussion from the ICU-TC meeting was to not accept the change.

neheb:
Looking at cstddef in GCC 10, I see:

```
#if __cplusplus >= 201103L
namespace std
{
  // We handle size_t, ptrdiff_t, and nullptr_t in c++config.h.
  using ::max_align_t;
}
#endif // C++11
```

So it's relying on C headers anyway.

GCC 4.8.2 has max_align_t in stddef.h . It does not have max_align_t in cstddef.

There's a way to hack around this by installing LLVM's libcxx and using that. Version 3.8.0 in CentOS 7 has:

```
_LIBCPP_BEGIN_NAMESPACE_STD

using ::ptrdiff_t;
using ::size_t;

#if defined(__CLANG_MAX_ALIGN_T_DEFINED) || defined(_GCC_MAX_ALIGN_T)
// Re-use the compiler's <stddef.h> max_align_t where possible.
using ::max_align_t;
#else
typedef long double max_align_t;
#endif
```

I can't say I like this.

PR #1275, opened 2020-08-29
ICU-21010 MeasureUnit extension in Java

younies:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21010
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [x] Tests included
- [ ] Documentation is changed or added

PR #1279, opened 2020-08-30
ICU-20568 Implementation of UnitConverter, ComplexUnitConverter and UnitsRouter

younies:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20568
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
@@ -1,0 +1,128 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+
+package com.ibm.icu.impl.units;
+
+
+import com.ibm.icu.util.Measure;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Converts from single or compound unit to single, compound or mixed units.
+ * For example, from `meter` to `foot+inch`.
+ * <p>
+ * DESIGN:
+ * This class uses `UnitConverter` in order to perform the single converter (i.e. converters from a
+ * single unit to another single unit). Therefore, `ComplexUnitsConverter` class contains multiple
+ * instances of the `UnitConverter` to perform the conversion.
+ */
+public class ComplexUnitsConverter {
+    public static final BigDecimal EPSILON = BigDecimal.valueOf(Math.ulp(1.0));
+    public static final BigDecimal EPSILON_MULTIPLIER = BigDecimal.valueOf(1).add(EPSILON);
+    private ArrayList<UnitConverter> unitConverters_;
+    private ArrayList<MeasureUnitImpl> units_;
+
+    /**
+     * Constructor of `ComplexUnitsConverter`.
+     * NOTE:
+     * - inputUnit and outputUnits must be under the same category
+     * - e.g. meter to feet and inches --> all of them are length units.
+     *
+     * @param inputUnit   represents the source unit. (should be single or compound unit).
+     * @param outputUnits represents the output unit. could be any type. (single, compound or mixed).
+     */
+    public ComplexUnitsConverter(MeasureUnitImpl inputUnit, MeasureUnitImpl outputUnits,
+                                 ConversionRates conversionRates) {
+        units_ = outputUnits.extractIndividualUnits();
+        assert (!units_.isEmpty());
+
+        // Sort the units in a descending order.
+        Collections.sort(
+                this.units_,
+                Collections.reverseOrder(new MeasureUnitImpl.MeasureUnitImplComparator(conversionRates)));
+
+
+        // If the `outputUnits` is `UMEASURE_UNIT_MIXED` such as `foot+inch`. Thus means there is more than one unit
+        //  and In this case we need more converters to convert from the `inputUnit` to the first unit in the
+        //  `outputUnits`. Then, a converter from the first unit in the `outputUnits` to the second unit and so on.
+        //      For Example:
+        //          - inputUnit is `meter`
+        //          - outputUnits is `foot+inch`
+        //              - Therefore, we need to have two converters:
+        //                      1. a converter from `meter` to `foot`
+        //                      2. a converter from `foot` to `inch`
+        //          - Therefore, if the input is `2 meter`:
+        //              1. convert `meter` to `foot` --> 2 meter to 6.56168 feet
+        //              2. convert the residual of 6.56168 feet (0.56168) to inches, which will be (6.74016
+        //              inches)
+        //              3. then, the final result will be (6 feet and 6.74016 inches)
+        unitConverters_ = new ArrayList<>();
+        for (int i = 0, n = units_.size(); i < n; i++) {
+            if (i == 0) { // first element
+                unitConverters_.add(new UnitConverter(inputUnit, units_.get(i), conversionRates));
+            } else {
+                unitConverters_.add(new UnitConverter(units_.get(i - 1), units_.get(i), conversionRates));
+            }
+        }
+    }
+
+    /**
+     * Returns true if the specified `quantity` of the `inputUnit`, expressed in terms of the biggest
+     * unit in the MeasureUnit `outputUnit`, is greater than or equal to `limit`.
+     * <p>
+     * For example, if the input unit is `meter` and the target unit is `foot+inch`. Therefore, this
+     * function will convert the `quantity` from `meter` to `foot`, then, it will compare the value in
+     * `foot` with the `limit`.
+     */
+    public boolean greaterThanOrEqual(BigDecimal quantity, BigDecimal limit) {
+        assert !units_.isEmpty();
+
+        // NOTE: First converter converts to the biggest quantity.
+        return unitConverters_.get(0).convert(quantity).multiply(EPSILON_MULTIPLIER).compareTo(limit) >= 0;
+    }
+
+    /**
+     * Returns outputMeasures which is an array with the corresponding values.
+     * - E.g. converting meters to feet and inches.
+     * 1 meter --> 3 feet, 3.3701 inches
+     * NOTE:
+     * the smallest element is the only element that could have fractional values. And all
+     * other elements are floored to the nearest integer
+     */
+    public List<Measure> convert(BigDecimal quantity) {
+        List<Measure> result = new ArrayList<>();
+
+        for (int i = 0, n = unitConverters_.size(); i < n; ++i) {
+            quantity = (unitConverters_.get(i)).convert(quantity);
+
+            if (i < n - 1) {
+                // The double type has 15 decimal digits of precision. For choosing
+                // whether to use the current unit or the next smaller unit, we
+                // therefore nudge up the number with which the thresholding
+                // decision is made. However after the thresholding, we use the
+                // original values to ensure unbiased accuracy (to the extent of
+                // double's capabilities).
+                BigDecimal newQuantity = quantity.multiply(EPSILON_MULTIPLIER).setScale(0, RoundingMode.FLOOR);
+
+                result.add(new Measure(newQuantity, units_.get(i).build()));
+
+                // Keep the residual of the quantity.
+                //   For example: `3.6 feet`, keep only `0.6 feet`
+                quantity = quantity.subtract(newQuantity);
+                if (quantity.compareTo(BigDecimal.ZERO) == -1) {
+                    quantity = BigDecimal.ZERO;
+                }
+            } else { // LAST ELEMENT
+                result.add(new Measure(quantity, units_.get(i).build()));
+            }
+        }
+
+        return result;
+    }
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
@@ -1,0 +1,234 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+
+package com.ibm.icu.impl.units;
+
+import com.ibm.icu.impl.ICUData;
+import com.ibm.icu.impl.ICUResourceBundle;
+import com.ibm.icu.impl.UResource;
+import com.ibm.icu.util.MeasureUnit;
+import com.ibm.icu.util.UResourceBundle;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class ConversionRates {
+
+    /**
+     * Map from any simple unit (i.e. "meter", "foot", "inch") to its basic/root conversion rate info.
+     */
+    private HashMap<String, ConversionRateInfo> mapToConversionRate;
+
+    public ConversionRates() {
+        // Read the conversion rates from the data (units.txt).
+        ICUResourceBundle resource;
+        resource = (ICUResourceBundle) UResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME, "units");
+        ConversionRatesSink sink = new ConversionRatesSink();
+        resource.getAllItemsWithFallback(UnitsData.Constants.CONVERSION_UNIT_TABLE_NAME, sink);
+        this.mapToConversionRate = sink.getMapToConversionRate();
+    }
+
+    /**
+     * Extracts the factor from a `SingleUnitImpl` to its Basic Unit.
+     *
+     * @param singleUnit
+     * @return
+     */
+    private UnitConverter.Factor getFactorToBase(SingleUnitImpl singleUnit) {
+        int power = singleUnit.getDimensionality();
+        MeasureUnit.SIPrefix siPrefix = singleUnit.getSiPrefix();
+        UnitConverter.Factor result = UnitConverter.Factor.processFactor(mapToConversionRate.get(singleUnit.getSimpleUnit()).getConversionRate());
+
+        return result.applySiPrefix(siPrefix).power(power); // NOTE: you must apply the SI prefixes before the power.
+    }
+
+    public UnitConverter.Factor getFactorToBase(MeasureUnitImpl measureUnit) {
+        UnitConverter.Factor result = new UnitConverter.Factor();
+        for (SingleUnitImpl singleUnit :
+                measureUnit.getSingleUnits()) {
+            result = result.multiply(getFactorToBase(singleUnit));
+        }
+
+        return result;
+    }
+
+    protected BigDecimal getOffset(MeasureUnitImpl source, MeasureUnitImpl target, UnitConverter.Factor
+            sourceToBase, UnitConverter.Factor targetToBase, UnitConverter.Convertibility convertibility) {
+        if (convertibility != UnitConverter.Convertibility.CONVERTIBLE) return BigDecimal.valueOf(0);
+        if (!(checkSimpleUnit(source) && checkSimpleUnit(target))) return BigDecimal.valueOf(0);
+
+        String sourceSimpleIdentifier = source.getSingleUnits().get(0).getSimpleUnit();
+        String targetSimpleIdentifier = target.getSingleUnits().get(0).getSimpleUnit();
+
+        BigDecimal sourceOffset = this.mapToConversionRate.get(sourceSimpleIdentifier).getOffset();
+        BigDecimal targetOffset = this.mapToConversionRate.get(targetSimpleIdentifier).getOffset();
+        return sourceOffset
+                .subtract(targetOffset)
+                .divide(targetToBase.getConversionRate(), MathContext.DECIMAL128);
+
+
+    }
+
+    public MeasureUnitImpl extractCompoundBaseUnit(MeasureUnitImpl measureUnit) {
+        ArrayList<SingleUnitImpl> baseUnits = this.extractBaseUnits(measureUnit);
+
+        MeasureUnitImpl result = new MeasureUnitImpl();
+        for (SingleUnitImpl baseUnit :
+                baseUnits) {
+            result.appendSingleUnit(baseUnit);
+        }
+
+        return result;
+    }
+
+    public ArrayList<SingleUnitImpl> extractBaseUnits(MeasureUnitImpl measureUnitImpl) {
+        ArrayList<SingleUnitImpl> result = new ArrayList<>();
+        ArrayList<SingleUnitImpl> singleUnits = measureUnitImpl.getSingleUnits();
+        for (SingleUnitImpl singleUnit :
+                singleUnits) {
+            result.addAll(extractBaseUnits(singleUnit));
+        }
+
+        return result;
+    }
+
+    /**
+     * @param singleUnit
+     * @return The bese units in the `SingleUnitImpl` with applying the dimensionality only and not the SI prefix.
+     * <p>
+     * NOTE:
+     * This method is helpful when checking the convertibility because no need to check convertibility.
+     */
+    public ArrayList<SingleUnitImpl> extractBaseUnits(SingleUnitImpl singleUnit) {
+        String target = mapToConversionRate.get(singleUnit.getSimpleUnit()).getTarget();
+        MeasureUnitImpl targetImpl = MeasureUnitImpl.UnitsParser.parseForIdentifier(target);
+
+        // Each unit must be powered by the same dimension
+        targetImpl.applyDimensionality(singleUnit.getDimensionality());
+
+        // NOTE: we do not apply SI prefixes.
+
+        return targetImpl.getSingleUnits();
+    }
+
+    /**
+     * Checks if the `MeasureUnitImpl` is simple or not.
+     *
+     * @param measureUnitImpl
+     * @return true if the `MeasureUnitImpl` is simple, false otherwise.
+     */
+    private boolean checkSimpleUnit(MeasureUnitImpl measureUnitImpl) {
+        if (measureUnitImpl.getComplexity() != MeasureUnit.Complexity.SINGLE) return false;
+        SingleUnitImpl singleUnit = measureUnitImpl.getSingleUnits().get(0);
+
+        if (singleUnit.getSiPrefix() != MeasureUnit.SIPrefix.ONE) return false;
+        if (singleUnit.getDimensionality() != 1) return false;
+
+        return true;
+    }
+
+    public static class ConversionRatesSink extends UResource.Sink {
+        /**
+         * Map from any simple unit (i.e. "meter", "foot", "inch") to its basic/root conversion rate info.
+         */
+        private HashMap<String, ConversionRateInfo> mapToConversionRate = new HashMap<>();
+
+        @Override
+        public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
+            assert (UnitsData.Constants.CONVERSION_UNIT_TABLE_NAME.equals(key.toString()));
+
+            UResource.Table conversionRateTable = value.getTable();
+            for (int i = 0; conversionRateTable.getKeyAndValue(i, key, value); i++) {
+                assert (value.getType() == UResourceBundle.TABLE);
+
+                String simpleUnit = key.toString();
+
+                UResource.Table simpleUnitConversionInfo = value.getTable();
+                String target = null;
+                String factor = null;
+                String offset = "0";
+                for (int j = 0; simpleUnitConversionInfo.getKeyAndValue(j, key, value); j++) {
+                    assert (value.getType() == UResourceBundle.STRING);
+
+
+                    String keyString = key.toString();
+                    String valueString = value.toString().replaceAll(" ", "");
+                    if ("target".equals(keyString)) {
+                        target = valueString;
+                    } else if ("factor".equals(keyString)) {
+                        factor = valueString;
+                    } else if ("offset".equals(keyString)) {
+                        offset = valueString;
+                    } else {
+                        assert false : "The key must be target, factor or offset";
+                    }
+                }
+
+                // HERE a single conversion rate data should be loaded
+                assert (target != null);
+                assert (factor != null);
+
+                mapToConversionRate.put(simpleUnit, new ConversionRateInfo(simpleUnit, target, factor, offset));
+            }
+
+
+        }
+
+        public HashMap<String, ConversionRateInfo> getMapToConversionRate() {
+            return mapToConversionRate;
+        }
+    }
+
+    public static class ConversionRateInfo {
+
+        private final String simpleUnit;
+        private final String target;
+        private final String conversionRate;
+        private final BigDecimal offset;
+
+        public ConversionRateInfo(String simpleUnit, String target, String conversionRate, String offset) {
+            this.simpleUnit = simpleUnit;
+            this.target = target;
+            this.conversionRate = conversionRate;
+            this.offset = forNumberWithDivision(offset);
+        }
+
+        private static BigDecimal forNumberWithDivision(String numberWithDivision) {
+            String[] numbers = numberWithDivision.split("/");
+            assert (numbers.length <= 2);
+
+            if (numbers.length == 1) {
+                return new BigDecimal(numbers[0]);
+            }
+
+            return new BigDecimal(numbers[0]).divide(new BigDecimal(numbers[1]), MathContext.DECIMAL128);
+        }
+
+        /**
+         * @return the base unit.
+         * <p>
+         * For example:
+         * ("meter", "foot", "inch", "mile" ... etc.) have "meter" as a base/root unit.
+         */
+        public String getTarget() {
+            return this.target;
+        }
+
+        /**
+         * @return The offset from this unit to the base unit.
+         */
+        public BigDecimal getOffset() {
+            return this.offset;
+        }
+
+        /**
+         * @return The conversion rate from this unit to the base unit.
+         */
+        public String getConversionRate() {
+            return conversionRate;
+        }
+    }
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
@@ -4,8 +4,14 @@
 
 package com.ibm.icu.impl.units;
 
-import com.ibm.icu.util.*;
+import com.ibm.icu.util.BytesTrie;
+import com.ibm.icu.util.CharsTrie;
+import com.ibm.icu.util.CharsTrieBuilder;
+import com.ibm.icu.util.ICUCloneNotSupportedException;
+import com.ibm.icu.util.MeasureUnit;
+import com.ibm.icu.util.StringTrieBuilder;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -112,6 +118,17 @@ public class MeasureUnitImpl {
 
         result.add(this.clone());
         return result;
+    }
+
+    /**
+     * Applies dimensionality to all the internal single units.
+     * For example: <b>square-meter-per-second</b>, when we apply dimensionality -2, it will be <b>square-second-per-p4-meter</b>
+     */
+    public void applyDimensionality(int dimensionality) {
+        for (SingleUnitImpl singleUnit :
+                singleUnits) {
+            singleUnit.setDimensionality(singleUnit.getDimensionality() * dimensionality);
+        }
     }
 
     /**
@@ -728,7 +745,21 @@ public class MeasureUnitImpl {
         }
     }
 
-    class SingleUnitComparator implements Comparator<SingleUnitImpl> {
+    static class MeasureUnitImplComparator implements Comparator<MeasureUnitImpl> {
+        private ConversionRates conversionRates;
+
+        public MeasureUnitImplComparator(ConversionRates conversionRates) {
+            this.conversionRates = conversionRates;
+        }
+
+        @Override
+        public int compare(MeasureUnitImpl o1, MeasureUnitImpl o2) {
+            UnitConverter fromO1toO2 = new UnitConverter(o1, o2, conversionRates);
+            return fromO1toO2.convert(BigDecimal.valueOf(1)).compareTo(BigDecimal.valueOf(1));
+        }
+    }
+
+    static class SingleUnitComparator implements Comparator<SingleUnitImpl> {
         @Override
         public int compare(SingleUnitImpl o1, SingleUnitImpl o2) {
             return o1.compareTo(o2);
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/SingleUnitImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/SingleUnitImpl.java
@@ -1,7 +1,6 @@
 // © 2020 and later: Unicode, Inc. and others.
 // License & terms of use: http://www.unicode.org/copyright.html
 
-
 package com.ibm.icu.impl.units;
 
 import com.ibm.icu.util.MeasureUnit;
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
@@ -1,0 +1,311 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+
+package com.ibm.icu.impl.units;
+
+import com.ibm.icu.util.MeasureUnit;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.regex.Pattern;
+
+import static java.math.MathContext.DECIMAL128;
+
+public class UnitConverter {
+    private BigDecimal conversionRate;
+    private BigDecimal offset;
+
+    /**
+     * Constructor of `UnitConverter`.
+     * NOTE:
+     * - source and target must be under the same category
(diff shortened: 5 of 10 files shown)


hugovdm (changes requested):
**_Status: RESOLVED._**


You've got two different versions of MeasureUnitImpl with no common history, making it hard to review and maintain.

For your try-rebase-java branch, it was added in 8a914628a1d3f5697519fb81a81423adcb9ab27d and 8a914628a1d3f5697519fb81a81423adcb9ab27d (merged in 45d7de15e7e73025aae761d636d59a7fd6ad39e2).

If you want #1275 reviewed, then further work should be based on top of it. E.g. this upstream-unit-converter-java branch should contain commit 45d7de15e7e73025aae761d636d59a7fd6ad39e2.

- If you're going to rebase this branch, I would suggest you might as well rebase it onto 6d0e83ff789033d812466f1450a43b2d4a5facc9 - the latest commit of try-rebase-java.

- [I don't recommend this option:] If you're going to maintain these separately, and not use git commit history to manage the diffs, then I'd perhaps hold off on sending this for review until try-rebase-java is merged - alternatively make a PR into the try-rebase-java branch so we can more easily see what further changes you're making on MeasureUnitImpl.

I currently think I should leave this PR for later, and instead review #1275 with urgency? I could also help guide rebasing this onto 1275. (In short, I'd create a new branch on on this branch's HEAD, then rebase that onto "try-rebase-java", and do a "git diff" to prove that the result is identical.) Technically a merge is also possible, but without the linear history, I find they're much, much harder to review.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param inputUnit   represents the source unit. (should be single or compound unit).
+     * @param outputUnits represents the output unit. could be any type. (single, compound or mixed).
+     */
+    public ComplexUnitsConverter(MeasureUnitImpl inputUnit, MeasureUnitImpl outputUnits,
+                                 ConversionRates conversionRates) {
+        units_ = outputUnits.extractIndividualUnits();
+        Assert.assrt(!units_.isEmpty());
+
+        // Sort the units in a descending order.
+        MeasureUnitImpl.MeasureUnitImplComparator.setConversionRates(conversionRates);
```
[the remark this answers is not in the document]
younies: because I want it to be static, but I think there is no point for doing this right now.  Good catch,  I am going to change it.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,133 @@
...
+     * - inputUnit and outputUnits must be under the same category
+     * - e.g. meter to feet and inches --> all of them are length units.
+     *
+     * @param inputUnit   represents the source unit. (should be single or compound unit).
+     * @param outputUnits represents the output unit. could be any type. (single, compound or mixed).
+     */
+    public ComplexUnitsConverter(MeasureUnitImpl inputUnit, MeasureUnitImpl outputUnits,
+                                 ConversionRates conversionRates) {
+        units_ = outputUnits.extractIndividualUnits();
+        Assert.assrt(!units_.isEmpty());
```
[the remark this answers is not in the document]
younies: Done.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,133 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html#License
+/*
+ *******************************************************************************
+ * Copyright (C) 2004-2020, Google Inc, International Business Machines
```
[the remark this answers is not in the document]
younies: In all the files, the notice are as I pasted it. 

Good catch for dropping `#Licence`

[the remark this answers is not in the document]
sffc: You don't need the International Business Machines copyright in new files.  In new files, you only need

```
// © 2020 and later: Unicode, Inc. and others.
```

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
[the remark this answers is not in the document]
sffc: Yeah, in Java it is more conventional to use the `assert` keyword.  It doesn't look like you've made this change.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -0,0 +1,315 @@
...
+
+        /**
+         * Returns a single `BigDecimal` that represent the conversion rate after substituting all the constants.
+         *
+         * @return
+         */
+        public BigDecimal getConversionRate() {
+            Factor resultCollector = this.clone();
+
+            resultCollector.substitute(BigDecimal.valueOf(0.3048), this.CONSTANT_FT2M);
```
sffc, line 183: Here and below: contrary to intuition, `new BigDecimal("0.3048")` is better than `BigDecimal.valueOf(0.3048)` because the string is more exact

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -0,0 +1,756 @@
+// © 2020 and later: Unicode, Inc. and others.
```
sffc, line 1: As Hugo said, once #1275 is checked in, make sure MeasureUnitImpl.java and related files are removed from this PR (or is updated based on what you need).

sffc (commented):
At a high level it looks good.  I was trying to compare it with the C++ implementation, but the two had differences that were large enough that I had trouble lining them up.  In general we try to keep Java and C++ implementations consistent.

I would like @hugovdm to review the business logic in the areas of the code that differ.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,134 @@
...
+     * <p>
+     * For example, if the input unit is `meter` and the target unit is `foot+inch`. Therefore, this
+     * function will convert the `quantity` from `meter` to `foot`, then, it will compare the value in
+     * `foot` with the `limit`.
+     */
+    public boolean greaterThanOrEqual(BigDecimal quantity, BigDecimal limit) {
+        assert !units_.isEmpty() ;
+
+        // NOTE: First converter converts to the biggest quantity.
+        return unitConverters_.get(0).convert(quantity).multiply(BigDecimal.valueOf(1 + EPSILON)).compareTo(limit) >= 0;
```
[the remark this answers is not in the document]
younies: yes, even big decimal is more accurate. Still, there is very very ting deterioration .

icu4j/main/tests/core/src/com/ibm/icu/dev/data/units/unitsTest.txt
```diff
@@ -0,0 +1,197 @@
+# This file is a copy of common/testData/units/unitsTest.txt from CLDR.
+# WIP/TODO(hugovdm): determine a good update procedure and document it.
```
hugovdm, line 2: We have a location now for automatically updated tests, please use this:
icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/units/unitsTest.txt

icu4j/main/tests/core/src/com/ibm/icu/dev/data/units/unitPreferencesTest.txt
```diff
@@ -0,0 +1,453 @@
+# This file is a copy of common/testData/units/unitPreferencesTest.txt from CLDR.
+# WIP/TODO(hugovdm): determine a good update procedure and document it.
```
hugovdm, line 2: Please use icu4j/main/tests/core/src/com/ibm/icu/dev/data/cldr/units/unitPreferencesTest.txt

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,134 @@
...
+                // double's capabilities).
+                Number newQuantity = Math.floor(quantity.doubleValue() * (1 + EPSILON));
+
+                // NOTE: Measure would own its MeasureUnit. // TODO?
+                //  result.add(new Measure(newQuantity, units_.get(i).build()));
+                tempResult.add(Pair.of(units_.get(i), BigDecimal.valueOf(newQuantity.doubleValue())));
+
+                // Keep the residual of the quantity.
+                //   For example: `3.6 feet`, keep only `0.6 feet`
+                quantity = quantity.subtract(BigDecimal.valueOf(newQuantity.longValue()));
```
hugovdm, line 123: It's possible we will still need the "if ((quantity - roundedQuantity) / quantity < DBL_EPSILON) {" logic, depending on how accurate we make our arbitrary precision calculations.

Can you add ComplexUnitConverterTest.java? Unit tests in "void UnitsTest::testComplexUnitsConverter() {" exercises the precision logic relatively well. (There might be a "TEST_EPSILON" which we can tweak to see how it relates to BigDecimal precision.)

hugovdm (commented):
Initial request: use the latest .txt files and add ComplexUnitConverter unit tests, so we can see if it can trigger calculation precision issues.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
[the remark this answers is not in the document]
younies: done

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
younies: is it possible to just compare if it is less than zero, then make it zero

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitConverterTest.java
```diff
@@ -0,0 +1,166 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+/*
+ *******************************************************************************
+ * Copyright (C) 2004-2020, Google Inc, International Business Machines
+ * Corporation and others. All Rights Reserved.
+ *******************************************************************************
+ */
+
```
younies, line 9: Remove

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitRouterTest.java
```diff
@@ -0,0 +1,116 @@
+// © 2020 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+/*
+ *******************************************************************************
+ * Copyright (C) 2004-2020, Google Inc, International Business Machines
+ * Corporation and others. All Rights Reserved.
+ *******************************************************************************
+ */
```
younies, line 8: Revome

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,134 @@
...
+            quantity = (unitConverters_.get(i)).convert(quantity);
+
+            if (i < n - 1) {
+                // The double type has 15 decimal digits of precision. For choosing
+                // whether to use the current unit or the next smaller unit, we
+                // therefore nudge up the number with which the thresholding
+                // decision is made. However after the thresholding, we use the
+                // original values to ensure unbiased accuracy (to the extent of
+                // double's capabilities).
+                Number newQuantity = Math.floor(quantity.doubleValue() * (1 + EPSILON));
```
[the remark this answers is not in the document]
sffc: I still think we can do more of this math with BigDecimal.  Let's discuss.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitRouterTest.java
```diff
@@ -0,0 +1,116 @@
...
+            if (line.isEmpty() || line.startsWith("#")) continue;
+            tests.add(new TestCase(line));
+        }
+
+        for (TestCase testCase :
+                tests) {
+            UnitsRouter router = new UnitsRouter(testCase.inputUnit.second, testCase.region, testCase.usage);
+            List<Measure> measures = router.route(testCase.input).measures;
+
+            Assert.assrt(measures.size() == testCase.expectedInOrder.size()
```
sffc, line 103: Since you are in a test, you should use `assertEquals`.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -0,0 +1,210 @@
...
+package com.ibm.icu.impl.units;
+
+import com.ibm.icu.impl.ICUData;
+import com.ibm.icu.impl.ICUResourceBundle;
+import com.ibm.icu.impl.UResource;
+import com.ibm.icu.util.UResourceBundle;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.TreeMap;
```
sffc, line 14: HashMap, unless you need it to be sorted

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
younies: For me, I think lets leave this part when we are doing the glue code. I am not sure if we needed it in this pull request.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitRouterTest.java
younies: done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitRouterTest.java
younies: done

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
younies: I use only HashMap

sffc (dismissed):

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
hugovdm: In my opinion it shouldn't be much more work to add testComplexUnitsConverter() to this PR instead of the glue code PR. Redoing a squash should be less than a minute, losing Shane's approve isn't a problem since I can approve on his behalf.

Me reviewing the validity of this "on behalf of the unit test" means I have to do more work, on behalf of the unit test. Meaning only adding the unit test in the glue code PR instead of this PR because "it's more work" means you would rather give me more work than you.

I can remove my review from this PR and just review the Glue Code PR instead.

hugovdm: The logic looks reasonable. The unit-sorting logic is missing though, I'll start a new thread for that.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -0,0 +1,126 @@
...
+
+                // Keep the residual of the quantity.
+                //   For example: `3.6 feet`, keep only `0.6 feet`
+                quantity = quantity.subtract(newQuantity);
+            } else { // LAST ELEMENT
+                result.add(new Measure(quantity, units_.get(i).build()));
+            }
+        }
+
+        return result;
```
hugovdm, line 123: FYI: this commit is already in C++ upstream:

https://github.com/icu-units/icu/pull/86/commits/602e41b080fa5d18ad9491ce299938f094a13b9c

It takes care of returning "inch-and-foot", when that was the requested output unit. I don't mind whether this is part of this PR or part of a later PR, but please remember to also take care of this? I've made the change in UnitsTest::testComplexUnitConverterSorting(), I've not tested for it in new unit test, so adding only the new tests won't catch this.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/MeasureUnitTest.java
```diff
@@ -3103,12 +3103,12 @@ static void generateCXXConstants() {
                     throw new IllegalStateException();
                 }
                 checkForDup(seen, name, unit);
-                System.out.printf("MeasureUnit *MeasureUnit::create%s(UErrorCode &status) {\n", name);
-                System.out.printf("    return MeasureUnit::create(%d, %d, status);\n",
+                System.out.printf("MeasureUnit *MeasureUnit.create%s(UErrorCode &status) {\n", name);
+                System.out.printf("    return MeasureUnit.create(%d, %d, status);\n",
```
hugovdm, line 3107: This is breaking the C++ output again.

hugovdm: Please revert changes to MeasureUnitTest.java

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -0,0 +1,276 @@
...
+
+        ComplexUnitsConverter complexConverter = new ComplexUnitsConverter(source, target, conversionRates);
+        List<Measure> measures = complexConverter.convert(BigDecimal.valueOf(10.0));
+
+        assertEquals(measures.size(), 2);
+        assertEquals("inch-and-foot unit 0", "foot", measures.get(0).getUnit().getIdentifier());
+        assertEquals("inch-and-foot unit 1", "inch", measures.get(1).getUnit().getIdentifier());
+
+        assertTrue("inch-and-foot value 0", compareTwoBigDecimal(BigDecimal.valueOf(32), BigDecimal.valueOf(measures.get(0).getNumber().doubleValue()), BigDecimal.valueOf(0.0001)));
+        assertTrue("inch-and-foot value 1", compareTwoBigDecimal(BigDecimal.valueOf(9.7008), BigDecimal.valueOf(measures.get(1).getNumber().doubleValue()), BigDecimal.valueOf(0.0001)));
```
hugovdm, line 54: Given that changes are anyway needed for MeasureUnitTest.java, I think you might as well make the Java equivalent of this fix a part of this PR: https://github.com/icu-units/icu/commit/602e41b080fa5d18ad9491ce299938f094a13b9c - I think it's less work?

Given that Shane has approved up to now, I'm happy/confident approving any further changes made, then you should be able to submit without first waiting for Shane to review the additional changes too.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -0,0 +1,309 @@
...
+            this.conversionRate = sourceToBase.multiply(targetToBase).getConversionRate();
+        }
+
+        // calculate the offset
+        this.offset = conversionRates.getOffset(source, target, sourceToBase, targetToBase, convertibility);
+    }
+
+    static public Convertibility extractConvertibility(MeasureUnitImpl source, MeasureUnitImpl target, ConversionRates conversionRates) {
+        ArrayList<SingleUnitImpl> sourceSingleUnits = conversionRates.getBasicUnitsWithoutSIPrefix(source);
+        ArrayList<SingleUnitImpl> targetSingleUnits = conversionRates.getBasicUnitsWithoutSIPrefix(target);
```
hugovdm, line 49: FYI / optional:

In C++ this is called extractCompoundBaseUnit. I would probably have gone with the same name, but if I wanted to rename it to "get*", I'd go with "getCompoundBaseUnit".

Alternatively, if I think the names in C++ also needs changing, I would make the names the same for now, with a TODO to rename, and then do a rename of both C++ and Java in the same PR: that makes review much easier. Otherwise each review requires some manual work trying to find the equivalent C++/Java code for the reviewer to compare and ensure they stay in sync.

hugovdm (commented):
I'm leaving some optional comments, where we can discuss what we expect of future C++/Java sync work. I'm very happy to do that technical debt work after the release, I just want to make sure we're tracking it well, that we don't forget what we still wanted to clean up.

One reason I've chosen to care more about code review feedback not left for future PRs is because of past experience: in https://github.com/sffc/icu/pull/21 I made some requests on which I was lenient. You wrote: "Yes, I prefer to submit this PR and add following PRs later" - I thought "Sure go ahead and fix it later" - but I had to come and fix some of these things myself later.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * @param singleUnit
+     * @return The bese units in the `SingleUnitImpl` with applying the dimensionality only and not the SI prefix.
+     * <p>
+     * NOTE:
+     * This method is helpful when checking the convertibility because no need to check convertibility.
+     */
+    public ArrayList<SingleUnitImpl> getBasicUnitsWithoutSIPrefix(SingleUnitImpl singleUnit) {
+        String target = mapToConversionRate.get(singleUnit.getSimpleUnit()).getTarget();
```
hugovdm, line 106: This getTarget() might return "kilogram", which does have an SI prefix. Hence the C++ names can be confusing.

The concept of an "SI Base Unit" is a well-known one, see https://en.wikipedia.org/wiki/Base_unit_(measurement) - so some thought went into "extractCompoundBaseUnit". This doens't mean it's the best name ever, but it does mean it can be worthwhile to be mindful about the reasons you choose to change the name.

And names do matter: I apologise if I've at times been too nitpicky about names in previous pull requests, I'm trying to find the right balance. Hwoever I think the fact that you're using different names in this PR than we have in C++, even for code that you wrote yourself, is proof that the C++ names may have benefited from some more thought.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
```diff
@@ -0,0 +1,243 @@
...
+        this.mapToConversionRate = sink.getMapToConversionRate();
+    }
+
+    /**
+     * Extracts the factor from a `SingleUnitImpl` to its Basic Unit.
+     *
+     * @param singleUnit
+     * @return
+     */
+    private UnitConverter.Factor getFactorToBase(SingleUnitImpl singleUnit) {
```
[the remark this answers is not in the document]
hugovdm: I'm stricter than Shane: I think it makes sense, I have more time for Units.

If we want to make C++ look like this code in the future, then I need this code to be at least as good as, or better than, the C++ code. This also makes me additionally nitpicky on the naming. (One example: "Basic Unit" is not a concept on Wikipedia. "Base Unit" is.)

If getting this clean now is too much work for the release deadline, and we rather want to rush it out the door with technical debt to be fixed next quarter (which is more work than getting it done now, but does help us meet the deadline), then update the ticket to not say "make C++ look like Java" - it would be about revisiting and cleaning up both C++ and Java.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -0,0 +1,309 @@
...
+
+        insertInMap(dimensionMap, sourceSingleUnits, 1);
+        insertInMap(dimensionMap, targetSingleUnits, -1);
+
+        if (areDimensionsZeroes(dimensionMap)) return Convertibility.CONVERTIBLE;
+
+        insertInMap(dimensionMap, targetSingleUnits, 2);
+        if (areDimensionsZeroes(dimensionMap)) return Convertibility.RECIPROCAL;
+
+        return Convertibility.INCONVERTIBLE;
```
hugovdm, line 61: Was using "inconvertible" instead of "unconvertible" a conscious choice? Both inconvertible (here) and unconvertible (C++) are valid English.

"1. unconvertible - used especially of currencies; incapable of being exchanged for or replaced by another currency of equal value. inconvertible, unexchangeable."

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -0,0 +1,309 @@
...
+
+    /**
+     * Responsible for all the Factor operation
+     * NOTE:
+     * This class is immutable
+     */
+    static class Factor {
+        private BigDecimal factorNum;
+        private BigDecimal factorDen;
+        /* FACTOR CONSTANTS */
```
hugovdm, line 105: This comment makes it sound like these ints should be the factors themselves, e.g. pi being 3.14159... - rather than as the powers. I commented on this in May, but not too clearly, and I reckoned we could leave it for later.

Do you want to fix the documentation now, or do you want to leave it for more work later (i.e. icu-units#84)? If you'd like to leave it for later, how about we add a TODO at the top of this file for improving the documentation?

The point of documentation is to make it easier for future readers of the code to understand. When the documentation only manages to confuse, no documentation is better.

In C++ we have `int32_t constants[CONSTANTS_COUNT] = {};` and we index them via constants.

Here in Java, you're doing this: `result.CONSTANT_GRAVITY = this.CONSTANT_GRAVITY * power;`
So we're redefining gravity as gravity^power. I didn't like "constants" as a variable name in C++, here it's simply worse.

I know finding good names isn't easy. "POWER_GRAVITY" and "POWER_PI" feel weird too, but it's at least closer to what these values mean, the power to which these constant factors are raised, rather than it looking like we're redefining universal constants.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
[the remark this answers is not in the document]
hugovdm: @sffc Sorry that I took so long to review this PR thoroughly!
@younies Please don't hesitate to tell me "please review my PR as soon as you get the chance". Don't worry about creating pressure for me: for me, I think I do better with that pressure.

(Marking as "unresolved" because I continued this conversation.)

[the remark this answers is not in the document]
hugovdm: (I also wasn't sure if it's best to get all code review feedback in one go, or whether it's better to iteratively go for a few rounds. I'm beginning to believe "all in one go" is best if/when the reviewer has sufficient capacity, while incremental makes sense if initial feedback is more urgently desired.)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -36,6 +37,95 @@ public static boolean compareTwoBigDecimal(BigDecimal expected, BigDecimal actua
...
+                measures.get(0).getUnit().getIdentifier());
+
+        assertTrue("1.9999: measures[1] value", compareTwoBigDecimal(BigDecimal.valueOf(11.9988),
+                BigDecimal.valueOf(measures.get(1).getNumber().doubleValue()), BigDecimal.valueOf(0.0001)));
+        assertEquals("1.9999: measures[1] unit", MeasureUnit.INCH.getIdentifier(),
+                measures.get(1).getUnit().getIdentifier());
+
+        // TODO: consider factoring out the set of tests to make this function more
+        // data-driven, *after* dealing appropriately with the memory leaks that can
+        // be demonstrated by this code.
```
hugovdm, line 63: "... memory leaks that can be demonstrated by the **ICU4C version of** this code."

hugovdm: (I haven't checked whether reuse makes any sense in Java, do we ever modify instances - I think not? Given objects are always allocated on the heap, never on the stack, there's less reason to attempt to reuse?)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/MeasureUnitTest.java
younies: Good catch, this file needs to be reverted

younies: Done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -0,0 +1,366 @@
...
+                new TestData("square-meter-per-square-hour", "hectare-per-square-second", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("hertz", "revolution-per-second", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("millimeter", "meter", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("yard", "meter", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("ounce-troy", "kilogram", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("percent", "portion", UnitConverter.Convertibility.CONVERTIBLE),
+                new TestData("ofhg", "kilogram-per-square-meter-square-second", UnitConverter.Convertibility.CONVERTIBLE),
+
+                new TestData("second-per-meter", "meter-per-second", UnitConverter.Convertibility.RECIPROCAL),
+        };
```
hugovdm, line 173: Please add these test cases:

        {"kilometer", "foot", CONVERTIBLE},                                          //
        {"hectare", "square-foot", CONVERTIBLE},                                     //
        {"kilometer-per-second", "second-per-meter", RECIPROCAL},                    //
        {"square-meter", "square-foot", CONVERTIBLE},                                //
        {"kilometer-per-second", "foot-per-second", CONVERTIBLE},                    //
        {"square-hectare", "pow4-foot", CONVERTIBLE},                                //
        {"square-kilometer-per-second", "second-per-square-meter", RECIPROCAL},      //
        {"cubic-kilometer-per-second-meter", "second-per-square-meter", RECIPROCAL}, //

I will attempt to make ICU4J and ICU4C unit testing much more data-driven in the future (so we can share tests more easily).

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
younies: I had a pull request before to change `UNCONVERTIBLE` in C++ to `INCONVERTIBLE` , but I do not see it.

Yes, it was a conscious choice. :)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -0,0 +1,366 @@
...
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.math.MathContext;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.Assert.*;
+
+
+public class UnitsTest {
```
hugovdm, line 28: Optional: consider sorting the tests in the same order as in units_test.cpp? Otherwise I would sort them myself next time I touch these files (and then "git blame" gets updated).

I noticed this PR does not have:

    void testUnitConstantFreshness();
    void testSiPrefixes();
    void testMass();
    void testArea();

I recognise the last three can be folded together with the temperature test:

    public void testConverterForTemperature() {

Do you think we should sync up the tests later, rather than now? If so, as part of which ticket?

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ConversionRates.java
younies: Done!

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
younies: done

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
younies: but flipped to `UNCONVERTIBLE`

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/MeasureUnitTest.java
hugovdm: Thanks! (Awaiting push of the commit.)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: This is the best I've come up with so far, and I'd be much more satisfied with it than CONSTANT_*: EXPONENT_GRAVITY, EXPONENT_PI, ...

younies: I agree with you, I have some ideas about that, for now, I have just added a TODO, because of the hard deadlines

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -728,10 +745,24 @@ private Type calculateType(int fMatch) {
...
+    }
+
+    static class SingleUnitComparator implements Comparator<SingleUnitImpl> {
         @Override
         public int compare(SingleUnitImpl o1, SingleUnitImpl o2) {
             return o1.compareTo(o2);
         }
     }
-}
+}
```
hugovdm, line 768: Optional: newline at end of file.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/MeasureUnitTest.java
```diff
@@ -3863,4 +3863,4 @@ private void verifyMixedUnit(
                     subUnits.get(i).getIdentifier());
         }
     }
-}
+}
```
hugovdm, line 3866: Optional: end with a new line to have this file not be touched by this PR?

hugovdm (commented):
I think I've now been thorough enough! (I think the parts that I've not been as thorough reviewing, will probably be reconsidered in icu-units#84.) Meaning if we have a decision on each of the open discussions, I think it should be fine if I approve.

(And then I think not necessary to wait for Shane to re-approve if there aren't noteworthy changes in terms of Java-specific things or ICU4J style - though I'll keep the commits tagged, so he could look at the diffs if he wants to.)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: Great, thanks!

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: Thanks!

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
younies: this is should be part of the glue code PR.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
younies: this will be synced later

hugovdm (approved):
We've agreed to merge in order to get the features in by tomorrow - on condition that fixes will then be done in follow-up PRs after tomorrow.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -0,0 +1,311 @@
...
+     * - source and target must be under the same category
+     * - e.g. meter to mile --> both of them are length units.
+     *
+     * @param source          represents the source unit.
+     * @param target          represents the target unit.
+     * @param conversionRates contains all the needed conversion rates.
+     */
+    public UnitConverter(MeasureUnitImpl source, MeasureUnitImpl target, ConversionRates conversionRates) {
+        Convertibility convertibility = extractConvertibility(source, target, conversionRates);
+        assert (convertibility == Convertibility.CONVERTIBLE || convertibility == Convertibility.RECIPROCAL);
```
hugovdm, line 32: This assert is failing for me I do get this assert failing for a global `ant check`.

I don't get an assert failure for this:
$ ant checkTest -Dtestclass='com.ibm.icu.dev.test.impl.UnitConverterTest'

I hope asserts are enabled for both? But then why is my "check" broken but my "checkTest" not?

I don't see this assert in C++ either, there's quite a bit of difference between the constructors.

hugovdm: Turns out I had an old copy of unitsTest.txt in ./main/tests/core/out/bin/com/ibm/icu/dev/data/units/unitsTest.txt,
ant did not pick up the newer ./main/tests/core/src/com/ibm/icu/dev/data/cldr/units/unitsTest.txt

So that resolves my "ant check" problem. :-)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
hugovdm: It wasn't in the glue code PR. I've filed https://github.com/icu-units/icu/issues/103 to track this.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
hugovdm: I'm fixing this as part of unicode-org#1337

icu4j/main/classes/core/src/com/ibm/icu/impl/units/SingleUnitImpl.java
```diff
@@ -1,7 +1,6 @@
 // © 2020 and later: Unicode, Inc. and others.
 // License & terms of use: http://www.unicode.org/copyright.html
 
-
```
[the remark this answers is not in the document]
hugovdm: Saw most java classes have zero blank lines between the copyright notcie and the package declaration. It's moot, anyway. Resolving.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
hugovdm: Do we have a ticket or TODO to track this syncing?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -0,0 +1,366 @@
...
+                measures4.get(0).getUnit().getIdentifier());
+        assertTrue("light-year test: measures[1] value", compareTwoBigDecimal(BigDecimal.valueOf(10),
+                BigDecimal.valueOf(measures4.get(1).getNumber().doubleValue()),
+                BigDecimal.valueOf(1)));
+        assertEquals("light-year test: measures[1] unit", MeasureUnit.METER.getIdentifier(),
+                measures4.get(1).getUnit().getIdentifier());
+
+        // 2e-16 light years is 1.892146 meters. We consider this in the noise, and
+        // thus expect a 0. (This test fails when 2e-16 is increased to 4e-16.)
+        List<Measure> measures5 = converter3.convert(BigDecimal.valueOf(1.0 + 2e-17));
```
hugovdm, line 115: FYI: C++ and the comment here talks about 2e-16 - but the unit test has 2e-17? Why is it 2e-17? (I missed this in my review, sorry. Busy fixing it now.)

1.0 + 2e-17, at double precision, is basically just 1.0. What makes fixing this a bit less straight-forward:
Java is using BigDecimal. Why? As in: if we're using something more precise than double, particularly arbitrary precision, we need to decide how precise we want the result to be.

I'll see what I can figure out as a good place to draw the line. (If I had plenty of time, I'd benchmark to see how much slower BigDecimal is, and perhaps then argue to just stick to double precision, for Java too.)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
hugovdm: I have taken care of this in https://github.com/unicode-org/icu/pull/1421

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
hugovdm: Taken care of in https://github.com/unicode-org/icu/pull/1421

hugovdm:
> We've agreed to merge in order to get the features in by tomorrow - on condition that fixes will then be done in follow-up PRs after tomorrow.

My PR https://github.com/unicode-org/icu/pull/1421, merged on October 27th, took care of some of my requests for improvements to the unit tests.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: I've created https://github.com/unicode-org/icu/pull/1464 to address this.

PR #1280, opened 2020-08-31
ICU-21256 Fix MeasureUnitImpl assert failure

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21256
- [ ] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -683,6 +683,11 @@ void serialize(MeasureUnitImpl& impl, UErrorCode& status) {
     if (U_FAILURE(status)) {
         return;
     }
+    if (!impl.identifier.isEmpty()) {
+        // We already have an identifier, trust other methods maintained it or
+        // discarded it.
+        return;
+    }
     U_ASSERT(impl.identifier.isEmpty());
     if (impl.units.length() == 0) {
         // Dimensionless, constructed by the default constructor: no appending
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -688,7 +688,6 @@ void serialize(MeasureUnitImpl& impl, UErrorCode& status) {
         // discarded it.
         return;
     }
-    U_ASSERT(impl.identifier.isEmpty());
     if (impl.units.length() == 0) {
         // Dimensionless, constructed by the default constructor: no appending
         // to impl.identifier, we wish it to contain the zero-length string.
--- a/icu4c/source/i18n/measunit_impl.h
+++ b/icu4c/source/i18n/measunit_impl.h
@@ -22,7 +22,7 @@ static const char kDefaultCurrency8[] = "XXX";
 /**
  * A struct representing a single unit (optional SI prefix and dimensionality).
  */
-struct SingleUnitImpl : public UMemory {
+struct U_I18N_API SingleUnitImpl : public UMemory {
     /**
      * Gets a single unit from the MeasureUnit. If there are multiple single units, sets an error
      * code and returns the base dimensionless unit. Parses if necessary.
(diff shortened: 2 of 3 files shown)


hugovdm:
(force-pushed to rebase onto latest upstream/master, fixing conflicting unit tests.)

icu4c/source/i18n/measunit_impl.h
```diff
@@ -120,12 +119,14 @@ struct SingleUnitImpl : public UMemory {
     int32_t dimensionality = 1;
 };
 
+// For Windows: enable DLL export of a struct that uses this template.
+template class U_I18N_API MaybeStackVector<SingleUnitImpl>;
```
sffc, line 123: Put it inside an ifdef, like this:

https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/collationiterator.h#L34

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -683,7 +683,11 @@ void serialize(MeasureUnitImpl& impl, UErrorCode& status) {
     if (U_FAILURE(status)) {
         return;
     }
-    U_ASSERT(impl.identifier.isEmpty());
+    if (!impl.identifier.isEmpty()) {
```
sffc, line 686: This is fine, but what code triggered the assertion failure?

hugovdm: Some code in "MixedUnits part 2" - want to see it? I'll tidy up that branch for an initial review. (Still have to deal with skeleton generation failure.)

icu4c/source/i18n/measunit_impl.h
hugovdm: Not yet sufficient:

```
icu4c\source\common\cmemory.h(759,0): Error C4251: 'icu_67::MemoryPool<icu_67::SingleUnitImpl,8>::fPool': class 'icu_67::MaybeStackArray<T *,8>' needs to have dll-interface to be used by clients of class 'icu_67::MemoryPool<icu_67::SingleUnitImpl,8>'
```

hugovdm: Looks like I missed a "*".

hugovdm: Or no, also a `MaybeStackArray<T*, stackCapacity>` for the private `fPool` member of MemoryPool. Adding that and the ifdef.

hugovdm: OT (or tangential), I wonder if this solution found in number_decimfmtprops.h could be resolved by using a similar "template class" instantiation:

```
// Exported as U_I18N_API because it is a public member field of exported DecimalFormatProperties
// Using this wrapper is rather unfortunate, but is needed on Windows platforms in order to allow
// for DLL-exporting an fully specified template instantiation.
class U_I18N_API CurrencyPluralInfoWrapper {
```

Leaving it for now. (Is it worth filing a [low-priority] ticket for cruft / cleanup?)

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Ooh, ouch: looks like I might have triggered this assertion failure only via my debug output, so removing that, I don't need this? :-)

    const MeasureUnitImpl &outputUnit = MeasureUnitImpl::forMeasureUnit(unit, temp, status);
    const MeasureUnitImpl *inputUnit = &outputUnit;
    ...  // If inputUnit is mixed, replace by the first unit
    // printf("inputUnit: %s\n", std::move(inputUnit->copy(status)).build(status).getIdentifier());
    // printf("outputUnit: %s\n", std::move(outputUnit.copy(status)).build(status).getIdentifier());

hugovdm:
Perhaps best I close this PR ("not currently needed") and re-open it when I get back to it.

PR #1282, opened 2020-09-01
ICU-21267 stop defining & using FALSE & TRUE macros in public headers

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21267

In public header files, and in a few internal ones included by tests and tools:
- Stop defining macros FALSE & TRUE unless U_DEFINE_FALSE_AND_TRUE=1.
- Deprecate FALSE & TRUE.
- Replace FALSE & TRUE with false & true.

### DIFF
--- a/docs/userguide/dev/codingguidelines.md
+++ b/docs/userguide/dev/codingguidelines.md
@@ -439,11 +439,29 @@ includes other header files. The most common types are `uint8_t`, `uint16_t`,
 `uint32_t`, `int8_t`, `int16_t`, `int32_t`, `char16_t`,
 `UChar` (same as `char16_t`), `UChar32` (signed, 32-bit), and `UErrorCode`.
 
-The language built-in type bool and constants true and false may be used
+The language built-in type `bool` and constants `true` and `false` may be used
 internally, for local variables and parameters of internal functions. The ICU
 type `UBool` must be used in public APIs and in the definition of any persistent
-data structures. `UBool` is guaranteed to be one byte in size and signed; bool is
+data structures. `UBool` is guaranteed to be one byte in size and signed; `bool` is
 not.
+
+Traditionally, ICU4C has defined its own `FALSE`=0 / `TRUE`=1 macros for use with `UBool`.
+Starting with ICU 68 (2020q4), we no longer define these in public header files
+(unless `U_DEFINE_FALSE_AND_TRUE`=1),
+in order to avoid name collisions with code outside ICU defining enum constants and similar
+with these names.
+
+Instead, the versions of the C and C++ standards we require now do define type `bool`
+and values `false` & `true`, and we and our users can use these values.
+
+As of ICU 68, we are not changing ICU4C API from `UBool` to `bool`.
+Doing so in C API, or in structs that cross the library boundary,
+would break binary compatibility.
+Doing so only in other places in C++ could be confusingly inconsistent.
+We may revisit this.
+
+Note that the details of type `bool` (e.g., `sizeof`) depend on the compiler and
+may differ between C and C++.
 
 #### File Names (.h, .c, .cpp, data files if possible, etc.)
 
--- a/icu4c/source/common/unicode/umachine.h
+++ b/icu4c/source/common/unicode/umachine.h
@@ -49,6 +49,7 @@
  * ANSI C headers:
  * stddef.h defines wchar_t
  */
+#include <stdbool.h>
 #include <stddef.h>
 
 /*==========================================================================*/
@@ -170,11 +171,11 @@
 
 /**
  * \def UPRV_BLOCK_MACRO_END
- * Defined as "while (FALSE)" by default.
+ * Defined as "while (false)" by default.
  * @internal
  */
 #ifndef UPRV_BLOCK_MACRO_END
-#define UPRV_BLOCK_MACRO_END while (FALSE)
+#define UPRV_BLOCK_MACRO_END while (false)
 #endif
 
 /*==========================================================================*/
@@ -257,18 +258,59 @@
 /* Boolean data type                                                        */
 /*==========================================================================*/
 
-/** The ICU boolean type @stable ICU 2.0 */
+/**
+ * The ICU boolean type, a signed-byte integer.
+ * ICU-specific for historical reasons: The C and C++ standards used to not define type bool.
+ * Also provides a fixed type definition, as opposed to
+ * type bool whose details (e.g., sizeof) may vary by compiler and between C and C++.
+ *
+ * @stable ICU 2.0
+ */
 typedef int8_t UBool;
 
+/**
+ * \def U_DEFINE_FALSE_AND_TRUE
+ * Normally turns off defining macros FALSE=0 & TRUE=1 in public ICU headers.
+ * These obsolete macros sometimes break compilation of other code that
+ * defines enum constants or similar with these names.
+ * C++ has long defined bool/false/true.
+ * C99 also added definitions for these, although as macros; see stdbool.h.
+ *
+ * You may transitionally define U_DEFINE_FALSE_AND_TRUE=1 if you need time to migrate code.
+ *
+ * @internal ICU 68
+ */
+#ifdef U_DEFINE_FALSE_AND_TRUE
+    // Use the predefined value.
+#elif defined(U_COMBINED_IMPLEMENTATION) || \
+        defined(U_COMMON_IMPLEMENTATION) || defined(U_I18N_IMPLEMENTATION) || \
+        defined(U_IO_IMPLEMENTATION) || defined(U_LAYOUTEX_IMPLEMENTATION) || \
+        defined(U_TOOLUTIL_IMPLEMENTATION)
+    // Inside ICU: Keep FALSE & TRUE available.
+#   define U_DEFINE_FALSE_AND_TRUE 1
+#else
+    // Outside ICU: Avoid collision with non-macro definitions of FALSE & TRUE.
+#   define U_DEFINE_FALSE_AND_TRUE 0
+#endif
+
+#if U_DEFINE_FALSE_AND_TRUE || defined(U_IN_DOXYGEN)
 #ifndef TRUE
-/** The TRUE value of a UBool @stable ICU 2.0 */
+/**
+ * The TRUE value of a UBool.
+ *
+ * @deprecated ICU 68 Use standard "true" instead.
+ */
 #   define TRUE  1
 #endif
 #ifndef FALSE
-/** The FALSE value of a UBool @stable ICU 2.0 */
+/**
+ * The FALSE value of a UBool.
+ *
+ * @deprecated ICU 68 Use standard "false" instead.
+ */
 #   define FALSE 0
 #endif
-
+#endif  // U_DEFINE_FALSE_AND_TRUE
 
 /*==========================================================================*/
 /* Unicode data types                                                       */
--- a/icu4c/source/tools/ctestfw/unicode/testtype.h
+++ b/icu4c/source/tools/ctestfw/unicode/testtype.h
@@ -39,3 +39,9 @@
     #define T_CTEST_EXPORT_API T_CTEST_IMPORT
 #endif
 
+#ifndef TRUE
+#   define TRUE  1
+#endif
+#ifndef FALSE
+#   define FALSE 0
+#endif
--- a/icu4c/source/tools/toolutil/toolutil.h
+++ b/icu4c/source/tools/toolutil/toolutil.h
@@ -23,6 +23,12 @@
 
 #include "unicode/utypes.h"
 
+#ifndef TRUE
+#   define TRUE  1
+#endif
+#ifndef FALSE
+#   define FALSE 0
+#endif
 
 #ifdef __cplusplus
 
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -292,7 +292,7 @@ public:
     /**
      * Default constructor initializes with internal T[stackCapacity] buffer.
      */
-    MaybeStackArray() : ptr(stackArray), capacity(stackCapacity), needToRelease(FALSE) {}
+    MaybeStackArray() : ptr(stackArray), capacity(stackCapacity), needToRelease(false) {}
     /**
      * Automatically allocates the heap array if the argument is larger than the stack capacity.
      * Intended for use when an approximate capacity is known at compile time but the true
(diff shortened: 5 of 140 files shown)


sffc:
> Replace FALSE & TRUE in code with 0 & 1.

I don't think this is good.  I would rather see `U_TRUE` or `UPRV_TRUE` than `1`.  I know 1 means true, but that's not always evident from reading code, and it's confusing if you come from environments where 0 means true.

sffc:
> Replace FALSE & TRUE in API docs with false & true.

But this is not true.  Those functions *don't* return false or true.  As I've stated before, some compilers, like MSVC, won't let you treat `uint8_t` as if it were `bool`.

richgillam:
>I don't think this is good. I would rather see U_TRUE or UPRV_TRUE than 1. I know 1 means true, but that's not always evident from reading code, and it's confusing if you come from environments where 0 means true.

I'm with Shane on this one.  We really, really, don't want to use `0` and `1` to represent Boolean values in our code; that's a huge step backward in readability and maintainability.  I thought the whole point of this PR was to switch to using `bool`, `true`, and `false` in the spots where we were using `UBool`, `TRUE`, and `FALSE`, at least internally where it wouldn't cause API breakage.  If we can't do that, we either shouldn't do this at all or should define new constants that don't conflict with other libraries (such as the `U_TRUE` and `U_FALSE` constants Shane was suggesting).

pedberg-icu:
Personally I really like U_TRUE and U_FALSE, seems like the simplest solution.

markusicu:
> > Replace FALSE & TRUE in code with 0 & 1.
> 
> I don't think this is good. I would rather see `U_TRUE` or `UPRV_TRUE` than `1`. I know 1 means true, but that's not always evident from reading code,

So far I have done what we discussed a few weeks ago in a team meeting.

I agree that 0 and 1 are much less readable, but it also seems a little weird in this day and age to introduce new macro definitions for false & true.

> and it's confusing if you come from environments where 0 means true.

Shell scripts? Not C or C++ (nor Python etc.).

markusicu:
> > Replace FALSE & TRUE in API docs with false & true.
> 
> But this is not true. Those functions _don't_ return `false` or `true`.

Right, they don't really, but those values are at least convertible.

> As I've stated before, some compilers, like MSVC, won't let you treat `uint8_t` as if it were `bool`.

I believe that @jefgen said that that was a bug, and that it has been fixed.

markusicu:
> I thought the whole point of this PR was to switch to using `bool`, `true`, and `false` in the spots where we were using `UBool`, `TRUE`, and `FALSE`,

We discussed the type in the 2020-06-17 team meeting and had consensus to "Keep using UBool in old & new API.".

We also discussed my ticket and agreed for me to work on it.

When we no longer define FALSE and TRUE by default, then it no longer makes sense to mention them in the API docs.

It also seems a little weird to define new constants now that C and C++ each define bool/false/true. Unfortunately, C and C++ define them differently, 

At least false and true convert to 0 and 1 in C++ (and in C they are macros with those values), and you can use a UBool in a conditional expression.

I could change the 0/1 to false/true in call sites. In C code, that would require also including `<stdbool.h>`.

> at least internally where it wouldn't cause API breakage.

Changing internal usage is a non-goal for this ticket.

What I am trying to do is avoid compiler errors when newly including an ICU header for the first time, and having to hack around the name collision between the ICU macros and other people's enum constants.

> If we can't do that, we either shouldn't do this at all or should define new constants that don't conflict with other libraries (such as the `U_TRUE` and `U_FALSE` constants Shane was suggesting).

I added this to tomorrow's agenda.

markusicu:
> > As I've stated before, some compilers, like MSVC, won't let you treat `uint8_t` as if it were `bool`.
> 
> I believe that @jefgen said that that was a bug, and that it has been fixed.

Found it. Jeff wrote in a comment on the ticket: “this is only with VS 2015. The warning is no longer generated or issued in VS 2017.”

sffc:
> Found it. Jeff wrote in a comment on the ticket: “this is only with VS 2015. The warning is no longer generated or issued in VS 2017.”

If we're okay breaking VS 2015, then I would be okay with replacing `TRUE` and `FALSE` with `true` and `false`.  That's better than replacing them with `1` and `0`.

richgillam:
>If we're okay breaking VS 2015, then I would be okay with replacing TRUE and FALSE with true and false. That's better than replacing them with 1 and 0.

+1

jefgen:
> Jeff wrote in a comment on the ticket: “this is only with VS 2015. The warning is no longer generated or issued in VS 2017.”

Unfortunately, the situation seems to be slightly more complicated now it seems... :/

In VS2015, the compiler would issue warning `C4800` anywhere a type was implicitly converted to the type `bool`.
For example:
```
'int32_t': forcing value to bool 'true' or 'false' (performance warning)
```

In VS2017 the warning was removed completely. The compiler will never issue it, and there is no way to enable any such warning.

However, the warning was actually brought back in VS2019, though it is off by default. The output message is also changed slightly as well. It now looks like this:
```
Implicit conversion from 'int32_t' to bool. Possible information loss`
```

So if we use `0`/`1` then it is possible that people might try to build ICU4C with the warning enabled -- using `/Wall` does enable the warning for example. (Or perhaps someone might selectively enable just this warning). If you build with warnings as errors (using `/WX`) then it would be an error. So it's conceivable that we could (perhaps) get future bug reports about it.

--

One other comment regarding the `UBool` typedef comment:

```
/**
 * The ICU boolean type, a signed-byte integer.
 * ICU-specific for historical reasons.
 * The C and C++ standards used to not define a bool type.
 *
 * @stable ICU 2.0
 */
typedef int8_t UBool;
```

FWIW, I think that this should perhaps mention that the type is ICU-specific for more than just historical reasons.

The problem is that `sizeof(bool)` is implementation defined in C++, and so it **can't** be used on any public API surface where binary compatibility is a concern. (This unfortunately means that everyone is forced to come up with their own fixed-size type for boolean on public APIs...)

markusicu:
Jeff, thanks for digging.

> However, the warning was actually brought back in VS2019, though it is off by default. The output message is also changed slightly as well. It now looks like this:
> 
> ```
> Implicit conversion from 'int32_t' to bool. Possible information loss`
> ```

So if we and/or our users have
```
UBool uicu_function(...);
bool result = uicu_function(...);
```
that second line is liable to get a warning for information loss.
Which means that a call site like this either needs an explicit cast or needs to use type UBool to match the actual return type of the function.

This seems like an issue that we need to document, but less of a problem for the kind of change we are discussing.

It should continue to work to use a UBool in a conditional expression, as in `if (uicu_function(...)) { ...`

More importantly for this change, I assume that the opposite direction -- using a bool value (especially constants false & true) to initialize a UBool variable or argument -- will not get a warning because there is no information loss. (It widens, instead of narrowing.)

> FWIW, I think that this should perhaps mention that the type is ICU-specific for more than just historical reasons.
> 
> The problem is that `sizeof(bool)` is implementation defined in C++, and so it **can't** be used on any public API surface where binary compatibility is a concern. (This unfortunately means that everyone is forced to come up with their own fixed-size type for boolean on public APIs...)

Yes, good point. Although less important for C++ where we don't promise binary compatibility, but would still prevent us from using bool in a struct across the API. Bigger problem for C; I don't see anything about `sizeof(_Bool)` there.

srl295:
Yes 0 means true in shell scripts but also things like `ioctl()` http://refspecs.linux-foundation.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/baselib-ioctl-2.html

Seeing bare 0/1 is very strange and non obvious. I'd prefer true false or U_something.

markusicu:
@jefgen could you please try this on Windows / VS2019, with the warning about narrowing to bool enabled?

markusicu:
> I could change the 0/1 to false/true in call sites. In C code, that would require also including `<stdbool.h>`.

done

richgillam (approved):
For my part, I think this looks _much_ better, and since using the standard `true` and `false` doesn't seem to be causing any problems, I'm wholeheartedly behind this.

sffc:
Both of @jefgen's situations can be fixed by using `UBool` instead of `bool`.  I think the idea was to stop using `TRUE` and `FALSE`, but to use `UBool` as the type.

markusicu:
@jefgen correct me if I am wrong -- since I didn't change any UBool to bool, all of the bool types are already in ICU before my PR, right? And these are pre-existing warnings? As such, I think they should be cleaned up separately (general warnings swat or specific ticket) rather than complicate this PR here.

@sffc Yes, we could change bool back to UBool but I think many contributors want to move to a more normal way of writing C++. I would accept either going back to UBool or using a C-style cast (`bool whiteSpace = (bool)u_hasBinaryProperty(...);`) which is harmless and adds less visual clutter. I would leave it up to the contributor+reviewer for now. (If they prefer a C++ cast that's ok too of course.)

We did settle on keeping UBool in public API (possibly pending Fredrik's experiment with `operator==`). We have played it loose so far on bool vs. UBool in library code.

icu4c/source/common/unicode/umachine.h
```diff
@@ -257,18 +258,59 @@
...
+ * \def U_DEFINE_FALSE_AND_TRUE
+ * Normally turns off defining macros FALSE=0 & TRUE=1 in public ICU headers.
+ * These obsolete macros sometimes break compilation of other code that
+ * defines enum constants or similar with these names.
+ * C++ has long defined bool/false/true.
+ * C99 also added definitions for these, although as macros; see stdbool.h.
+ *
+ * You may transitionally define U_DEFINE_FALSE_AND_TRUE=1 if you need time to migrate code.
+ *
+ * @internal ICU 68
```
sffc, line 281: Why internal?

markusicu: Same as for switches like U_HIDE_DRAFT_API.
Not real API, and not promising that these will be around forever.

icu4c/source/tools/ctestfw/unicode/testtype.h
```diff
@@ -39,3 +39,9 @@
     #define T_CTEST_EXPORT_API T_CTEST_IMPORT
 #endif
 
+#ifndef TRUE
```
sffc, line 42: Why?  Weren't we going to use a C-specific header from the standard library (I forgot which one)?

icu4c/source/tools/toolutil/toolutil.h
```diff
@@ -23,6 +23,12 @@
 
 #include "unicode/utypes.h"
 
+#ifndef TRUE
```
sffc, line 26: Same question here

icu4c/source/tools/ctestfw/unicode/testtype.h
markusicu: I am not offering to change all of the implementation/samples/tools/test code in this ticket+PR. Therefore, the macros continue to be defined when compiling the ICU libraries themselves, and I added equivalent definitions into a few strategic places so that I need not change another few hundred files.

Only in simple cases that don't benefit from the shared toolutil or test header files did I change the call sites to use false/true.

icu4c/source/tools/toolutil/toolutil.h
markusicu: ditto

sffc (approved):

jefgen:
> @jefgen correct me if I am wrong -- since I didn't change any UBool to bool, all of the bool types are already in ICU before my PR, right? And these are pre-existing warnings? As such, I think they should be cleaned up separately (general warnings swat or specific ticket) rather than complicate this PR here.
> 

Good point. Yes, these would all be pre-existing warnings in the code before this change. I'd agree that it would be better to not add more churn to this PR -- I think the idea of using a general warnings clean-up ticket for addressing them later perhaps.

I tested building the changes in this PR with VS2015 as well, and the good news that with the latest version of the change it doesn't break/regress support for using VS2015. (Note: I did this in a branch in my fork [here](https://github.com/jefgen/icu/tree/test-pr-1282-with-vs2015). The logs are [here](https://dev.azure.com/ms/icu/_build/results?buildId=107208&view=logs&j=3bf652e4-2d1d-5980-4b40-cf8edb9ca05a) in case anyone is curious.)

Thanks for also adding info to the Coding Guidelines in the ICU User Guide as well.

docs/userguide/dev/codingguidelines.md
```diff
@@ -445,6 +445,24 @@ type `UBool` must be used in public APIs and in the definition of any persistent
 data structures. `UBool` is guaranteed to be one byte in size and signed; bool is
 not.
 
+Traditionally, ICU4C has defined its own FALSE=0 / TRUE=1 macros for use with UBool.
+Starting with ICU 68 (fall 2020), we no longer define these in public header files
```
jefgen, line 449: nit: It might be a bit better to use the term "autumn" instead of "fall", as fall is only for the Northern Hemisphere. :)

docs/userguide/dev/codingguidelines.md
```diff
@@ -445,6 +445,24 @@ type `UBool` must be used in public APIs and in the definition of any persistent
...
+(unless U_DEFINE_FALSE_AND_TRUE=1),
```
jefgen, line 450: super nit: could use back-tick (\`) for macro names (similar to the above `UBool`) for syntax highlighting in the rendered markdown output.

jefgen (approved):
The changes LGTM. Please squash the commits.

docs/userguide/dev/codingguidelines.md
markusicu: done

docs/userguide/dev/codingguidelines.md
markusicu: Good point, but "autumn" has the same problem. Changed to "2020q4".

jefgen (dismissed):
lgtm, thanks!

markusicu:
Squashed. Please re-approve.

jefgen (approved):
lgtm, thanks!

markusicu:
FYI re Steven's comment:

> Yes 0 means true in shell scripts but also things like `ioctl()` http://refspecs.linux-foundation.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/baselib-ioctl-2.html

That does not return a boolean value. "On success, 0 is returned. An ioctl() may use the return value as an output parameter and return a non-negative value on success. On error, -1 is returned and the global variable errno is set appropriately."

So you would test for `if (ioctl(...) >= 0) { success }`.

It does not make sense for a C function to use 0 as a boolean value because the C language treats 0 as "false" in conditions.

PR #1284, opened 2020-09-01
ICU-20568 Add .unit().usage() support to ICU4C NumberFormatter

hugovdm:
* https://github.com/unicode-org/icu/pull/1289 is a subset of this PR: just the public API changes in numberformatter.h. The API matches icu-design review.
* Code here has been written by @younies and @hugovdm, with some mutual code review, but with primary review by @sffc.
* Tag hugovdm_upstreaming_step24 points at a 10-commit squashed version of this PR. (Same content, but only 10 commits with alternating authors).

Closes: https://github.com/icu-units/icu/issues/42

Squashing examples:
* split up 697dd8c6c67ef7071a1268b2ea0e6cb602c04c95 and squashed it into 749dd2811fa96de501a6d19bf1ea770585d6caab and 297d831eee066591ba33a6c330f150db596979b9.
* squashed cdb85649677cec4938788dedb26f28c541da4e4b into 5cb4ca52b76550a53e473914a291083b64de88cd.
* squashed dcad0b6a4d2e56f6d401414391e27653801aee25 into 749dd2811fa96de501a6d19bf1ea770585d6caab.
* Successive commits by the same author can be squashed together. Some reordering was possible for even further squashing.

End result: 10 commits.

* This branch originally had a commit for each Pull Request into the old units-staging development branch. Some of the early PRs have been squashed together already though.
  * Each commit's message contains links to the original units-staging PRs and the commit ID at which those PRs were first merged into units-staging.
* A sequence of tags maintains history for this branch, e.g. a tag before every rebase (and sometimes tags right after a rebase too). The tags I created are documented in https://github.com/icu-units/icu/issues/42.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20568
- [x] Issue accepted
- [x] Tests included

ALLOW_MANY_COMMITS=true


younies (dismissed):

hugovdm:
Forgot to push 842cb10384 before pushing db36b9fdc7. These two have identical content, except for a chmod on .clang_format.
From 3961d07b9e to 842cb10384 is a couple of new commits.
My rewriting/rebasing steps are documented in https://github.com/icu-units/icu/issues/42

hugovdm:
I've not fixed this yet:

```
icu4c\source\common\cmemory.h(738,39): Error C2280: 'icu_68::units::ConverterPreference::ConverterPreference(const icu_68::units::ConverterPreference &)': attempting to reference a deleted function (compiling source file number_fluent.cpp)
```

Looks pretty hairy. Here's line 738:

```
        return fPool[fCount++] = new T(std::forward<Args>(args)...);
```

So is that trying to match a slightly different copy constructor on MSVC? :thinking: 

I've fixed all the MSVC / DLL export template instantiations at least (d347dc2). Marking this as no longer draft, because possibly some reviewer might have an MSVC tip? :-)

hugovdm:
I'm upstreaming (to CLDR!) the "Fix local copy of unitsTest.txt to match new CLDR data." change as https://github.com/unicode-org/cldr/pull/693.

jefgen:
Unfortunately, I don't have time to look into this more at the moment -- but in case it helps somewhat, here is the full error message output:

```
\source\common\cmemory.h(738,39): error C2280: 'icu_68::units::ConverterPreference::ConverterPreference(const icu_68::units::ConverterPreference &)': attempting to reference a deleted function (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\i18n\unitsrouter.h(78): message : compiler has generated 'icu_68::units::ConverterPreference::ConverterPreference' here (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\i18n\unitsrouter.h(78,1): message : 'icu_68::units::ConverterPreference::ConverterPreference(const icu_68::units::ConverterPreference &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'icu_68::units::ComplexUnitsConverter::ComplexUnitsConverter(const icu_68::units::ComplexUnitsConverter &)' (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\i18n\complexunitsconverter.h(75,1): message : 'icu_68::units::ComplexUnitsConverter::ComplexUnitsConverter(const icu_68::units::ComplexUnitsConverter &)': function was implicitly deleted because a data member invokes a deleted or inaccessible function 'icu_68::MaybeStackVector<icu_68::units::UnitConverter,8>::MaybeStackVector(const icu_68::MaybeStackVector<icu_68::units::UnitConverter,8> &)' (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\common\cmemory.h(845,1): message : 'icu_68::MaybeStackVector<icu_68::units::UnitConverter,8>::MaybeStackVector(const icu_68::MaybeStackVector<icu_68::units::UnitConverter,8> &)': function was implicitly deleted because a base class invokes a deleted or inaccessible function 'icu_68::MemoryPool<icu_68::units::UnitConverter,8>::MemoryPool(const icu_68::MemoryPool<icu_68::units::UnitConverter,8> &)' (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\common\cmemory.h(708,5): message : 'icu_68::MemoryPool<icu_68::units::UnitConverter,8>::MemoryPool(const icu_68::MemoryPool<icu_68::units::UnitConverter,8> &)': function was explicitly deleted (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\common\cmemory.h(793): message : see reference to function template instantiation 'T *icu_68::MemoryPool<T,8>::create<const T&>(const T &)' being compiled [\source\i18n\i18n.vcxproj]
          with
          [
              T=icu_68::units::ConverterPreference
          ] (compiling source file number_formatimpl.cpp)
\source\common\cmemory.h(793): message : see reference to function template instantiation 'T *icu_68::MemoryPool<T,8>::create<const T&>(const T &)' being compiled [\source\i18n\i18n.vcxproj]
          with
          [
              T=icu_68::units::ConverterPreference
          ] (compiling source file number_formatimpl.cpp)
\source\common\cmemory.h(838): message : see reference to function template instantiation 'T *icu_68::MaybeStackVector<T,8>::emplaceBack<const T&>(const T &)' being compiled [\source\i18n\i18n.vcxproj]
          with
          [
              T=icu_68::units::ConverterPreference
          ] (compiling source file number_formatimpl.cpp)
\source\common\cmemory.h(838): message : see reference to function template instantiation 'T *icu_68::MaybeStackVector<T,8>::emplaceBack<const T&>(const T &)' being compiled [\source\i18n\i18n.vcxproj]
          with
          [
              T=icu_68::units::ConverterPreference
          ] (compiling source file number_formatimpl.cpp)
\source\common\cmemory.h(836): message : while compiling class template member function 'void icu_68::MaybeStackVector<icu_68::units::ConverterPreference,8>::appendAll(const icu_68::MaybeStackVector<icu_68::units::ConverterPreference,8> &,UErrorCode &)' (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
\source\i18n\unitsrouter.h(89): message : see reference to class template instantiation 'icu_68::MaybeStackVector<icu_68::units::ConverterPreference,8>' being compiled (compiling source file number_formatimpl.cpp) [\source\i18n\i18n.vcxproj]
```

It looks as though the crux of the complaint is that `MemoryPool` cannot be copy-constructed, only move-constructed:
https://github.com/unicode-org/icu/blob/master/icu4c/source/common/cmemory.h#L705
```c++
    MemoryPool(const MemoryPool&) = delete;
```
There are a number of classes/structs that use and/or build upon the `MemoryPool`, but ultimately it looks as though this is trying to invoke a copy-construction with: `ConverterPreference::ConverterPreference(const icu_68::units::ConverterPreference &)`.

hugovdm:
**Status:** I'm late and have to run. On Linux I've found a `AddressSanitizer: attempting free on address which was not malloc()-ed` with my latest commit, need to dig in and see if it might be a MeasureUnitImpl memory leak I keep on bumping into but just haven't had time to hunt down yet, and/or whether I'm still missing something else or doing something else wrong.

hugovdm:
**Status:** My last commit causes ugly failures (ICU-20568 Functions may return MeasureUnitImpl&&, not by-value MeasureUnitImpl). I have created a minimal test case against master that does the same, so I can hunt this down better later (tomorrow).

That said: I would appreciate a/some sanity check review/s of the PR in the meantime.

hugovdm:
@jefgen 2f4561c contains the fix, for which I also have a separate PR (https://github.com/unicode-org/icu/pull/1306) to highlight the change rather than letting it sneak through in this big PR.

icu4c/source/i18n/complexunitsconverter.cpp
```diff
@@ -0,0 +1,148 @@
...
+                                             const ConversionRates &ratesInfo, UErrorCode &status)
+    : units_(outputUnits.extractIndividualUnits(status)) {
+    if (U_FAILURE(status)) {
+        return;
+    }
+
+    U_ASSERT(units_.length() != 0);
+
+    // NOTE:
+    //  This comparator is used to short the units in a descending order. Therefore, we return -1 if
```
younies, line 34: sort not short

icu4c/source/i18n/number_usageprefs.h
```diff
@@ -0,0 +1,126 @@
...
+ * MeasureUnit appropriate for a particular localized usage: see
+ * NumberFormatterSettings::usage().
+ */
+class U_I18N_API UsagePrefsHandler : public MicroPropsGenerator, public UMemory {
+  public:
+    UsagePrefsHandler(const Locale &locale, const MeasureUnit &inputUnit, const StringPiece usage,
+                      const MicroPropsGenerator *parent, UErrorCode &status);
+
+    /**
+     * Obtains the appropriate output value, MeasurementUnit and
```
younies, line 40: MeasureUnit or MeasurementUnit ?

icu4c/source/i18n/number_usageprefs.h
```diff
@@ -0,0 +1,126 @@
...
+     *     MIXED unit, the input MeasureUnit will be just the biggest unit of
+     *     the sequence.
+     * @param parent The parent MicroPropsGenerator.
+     * @param status Receives status.
+     */
+    UnitConversionHandler(const MeasureUnit &unit, const MicroPropsGenerator *parent,
+                          UErrorCode &status);
+
+    /**
+     * Obtains the appropraite output values from the Unit Converter.
```
younies, line 111: typo: appropriate

icu4c/source/test/depstest/dependencies.txt
```diff
@@ -979,6 +980,8 @@ group: number_output
     number_representation format formatted_value_sbimpl
     # PluralRules internals:
     unifiedcache
+    # Unit Formatting
```
sffc, line 983: This comment isn't really correct.  The dependency is because FormattedNumber now holds a MeasureUnit by value.  It doesn't necessarily have to do with formatting.

icu4c/source/test/depstest/dependencies.txt
```diff
@@ -987,16 +990,41 @@ group: numberformatter
...
     numrange_fluent.o numrange_impl.o
   deps
-    decnumber double_conversion formattable units
-    number_representation number_output
+    decnumber double_conversion formattable units unitsformatter
+    listformatter number_representation number_output
+    numsys
+    number_usageprefs
     uclean_i18n common
+    number_symbolswrapper
```
sffc, line 1005: I think numberformatter should depend directly on number_rounding?

icu4c/source/i18n/sources.txt
```diff
@@ -32,6 +32,7 @@ collationsettings.cpp
 collationtailoring.cpp
 collationweights.cpp
 compactdecimalformat.cpp
+complexunitsconverter.cpp
```
sffc, line 35: Very very optional (feel free to ignore): It would be nice if complexunitsconverter.cpp were sorted closer to the other units code.  It's up here by itself.  I would consider naming the files:

- units_complexconverter.cpp
- units_converter.cpp
- units_data.cpp
- units_router.cpp

sffc (dismissed):
Basically LGTM based on previous reviews.  Just some small optional comments.

icu4c/source/i18n/number_fluent.cpp
```diff
@@ -506,123 +520,6 @@ LocalizedNumberFormatter UnlocalizedNumberFormatter::locale(const Locale& locale
...
-    return fPtr.dfs;
-}
-
-const NumberingSystem* SymbolsWrapper::getNumberingSystem() const {
-    U_ASSERT(fType == SYMPTR_NS);
-    return fPtr.ns;
-}
-
-
 FormattedNumber LocalizedNumberFormatter::formatInt(int64_t value, UErrorCode& status) const {
```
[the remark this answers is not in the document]
hugovdm: It is related - was necessary for resolving dependencies issues (e.g. dependencies.txt).

It's something that could be refactored out into an earlier PR, but not as a later PR. I think it unnecessary to make a separate PR though.

icu4c/source/i18n/complexunitsconverter.cpp
hugovdm: Done.

icu4c/source/test/depstest/dependencies.txt
hugovdm: Done.

icu4c/source/test/depstest/dependencies.txt
hugovdm: Done.

icu4c/source/i18n/number_usageprefs.h
hugovdm: Done, thanks!

icu4c/source/i18n/number_usageprefs.h
hugovdm: Whoops, done!

icu4c/source/i18n/sources.txt
hugovdm: Done.

sffc (commented):
LGTM.  Please wait for the meeting tomorrow, and then squash as discussed.

icu4c/source/i18n/number_usageprefs.cpp
```diff
@@ -0,0 +1,233 @@
...
+    blueprint_helpers::parseIncrementOption(segment, macros, status);
+    return macros.precision;
+}
+
+UnitConversionHandler::UnitConversionHandler(const MeasureUnit &unit, const MicroPropsGenerator *parent,
+                                             UErrorCode &status)
+    : fOutputUnit(unit), fParent(parent) {
+    MeasureUnitImpl temp;
+    const MeasureUnitImpl &outputUnit = MeasureUnitImpl::forMeasureUnit(unit, temp, status);
+    const MeasureUnitImpl *inputUnit = &outputUnit;
```
younies, line 191: why this is a pointer ?

icu4c/source/i18n/number_usageprefs.cpp
```diff
@@ -0,0 +1,233 @@
...
+    MaybeStackVector<MeasureUnitImpl> singleUnits;
+    U_ASSERT(outputUnit.complexity == UMEASURE_UNIT_MIXED);
+    // When we wish to support unit conversion, replace the above assert with this if:
+    // if (outputUnit.complexity == UMEASURE_UNIT_MIXED) {
```
younies, line 195: I think this is too much details for the future use.

hugovdm: I don't mind if this is not in the Java version. My own preference was to simply support conversions, the comment was a compromise for removing that support.

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: To not unnecessarily copy units.

hugovdm:
Squashed version is good to go (10 commits). I did push it, and then undid the push, since there were still comments to respond to and Shane asked me to wait.

icu4c/source/i18n/number_usageprefs.cpp
younies: it can be reference.

icu4c/source/i18n/number_usageprefs.cpp
younies: okay, no problem :)

icu4c/source/i18n/number_microprops.h
```diff
@@ -36,27 +86,71 @@ struct MicroProps : public MicroPropsGenerator {
...
     MicroProps() = default;
 
     MicroProps(const MicroProps& other) = default;
 
     MicroProps& operator=(const MicroProps& other) = default;
 
-    void processQuantity(DecimalQuantity&, MicroProps& micros, UErrorCode& status) const U_OVERRIDE {
+    /**
+     * As MicroProps is the "base instance", this implementation of
+     * MicoPropsGenerator::processQuantity() just ensures that the output
```
younies, line 139: typo: Micro not Mico

icu4c/source/i18n/number_usageprefs.cpp
```diff
@@ -0,0 +1,233 @@
...
+        singleUnits = outputUnit.extractIndividualUnits(status);
+        U_ASSERT(singleUnits.length() > 0);
+        inputUnit = singleUnits[0];
+    }
+    // TODO: this should become an initOnce thing? Review with other
+    // ConversionRates usages.
+    ConversionRates conversionRates(status);
+    if (U_FAILURE(status)) {
+        return;
+    }
```
younies, line 206: We should pass the conversionRates.

icu4c/source/i18n/number_usageprefs.cpp
```diff
@@ -0,0 +1,233 @@
...
+    U_ASSERT(precisionSkeleton[kSkelPrefixLen - 1] == u'/');
+    StringSegment segment(precisionSkeleton, false);
+    segment.adjustOffset(kSkelPrefixLen);
+    MacroProps macros;
+    blueprint_helpers::parseIncrementOption(segment, macros, status);
+    return macros.precision;
+}
+
+UnitConversionHandler::UnitConversionHandler(const MeasureUnit &unit, const MicroPropsGenerator *parent,
```
younies, line 186: This confusing me, I think it is better to have an inputUnit and outPutUnit parameters.

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -1127,6 +1125,61 @@ class U_I18N_API Scale : public UMemory {
...
+    Usage& operator=(Usage&& src) U_NOEXCEPT;
+
+    /** @internal */
+    ~Usage();
+
+    /** @internal */
+    int16_t length() const { return fLength; }
+
+    /** @internal
+     * Makes a copy of value.
```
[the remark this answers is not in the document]
hugovdm: The main thing is we need to save the "usage" string in MacroProps instances. So this field is the main need, and in Java I'm quite confident it can indeed just be a plain String instance:

```
struct U_I18N_API MacroProps : public UMemory {
// [...]
    /** @internal */
    Usage usage;  // = Usage();  (no usage)
// [...]
```

In Java, a string is fine - we don't need to track error codes when we can throw exceptions, and we don't even need to copy strings (because they're immutable and garbage collected).

In C++, I think I added fError in order to be able to propagate errors happening while dealing with the Usage field. Looking at it now though, it looks like I've not done this correctly: I'm not checking for memory allocation errors, for example. I think it's mostly a bug that I'm not properly using fError - i.e. I still think it necessary to have a UErrorCode embedded with the Usage string (with the quartet of default copy/move construction in C++ being to blame).

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: We could add a second constructor that takes separate input and output units, and we could move the "take the first unit within a mixed unit sequence" logic to the callsite in `NumberFormatterImpl::macrosToMicroGenerator` to drop this single-unit constructor. The macrosToMicroGenerator code is fairly complicated anyway. I could set up a units-staging PR for this suggestion?

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: The question for the TODO is from where though. Where is the right place to load the conversion rates in the first place? Once we know this, we can figure out how to plumb it through in the neatest fashion.

icu4c/source/i18n/complexunitsconverter.cpp
hugovdm: (Commit 41d7ee071e270a030af5091f6e0fd8de197cbb87, found in the history of tag "hugovdm_upstreaming_step20".)

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: I've created a PR implementing your suggested improvement - looks a lot nicer! Please take a look: https://github.com/icu-units/icu/pull/90 (PR into units-staging, whereby we know the code was separately reviewed, then I can include it in this PR too).

icu4c/source/i18n/number_microprops.h
hugovdm: Done: fixed via e411135445ed8c50af2a0c02ec2d2526b5d661ad and squashed.

sffc (approved):

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: PR https://github.com/unicode-org/icu/pull/1322 improves this.

icu4c/source/i18n/number_usageprefs.cpp
hugovdm: This is now https://github.com/unicode-org/icu/pull/1322

icu4c/source/i18n/measunit_impl.h
```diff
@@ -189,6 +196,16 @@ struct U_I18N_API MeasureUnitImpl : public UMemory {
      */
     MeasureUnitImpl copy(UErrorCode& status) const;
```
hugovdm, line 197: I noticed this PR (with the change originating in https://github.com/icu-units/icu/pull/15) is adding a copy constructor, in conflict with this copy() method's intent to keep copying explicit: "Don't use copy constructor to make this explicit." I filed https://github.com/icu-units/icu/issues/113

PR #1285, opened 2020-09-01
ICU-21261 Adds 'units' data to pseudolocale data generation. Also cor…

gnrunge:
…rects the documentation in a few places.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21261
- [x] Issue accepted
- [ ] Tests included


hugovdm (approved):
Me learning about this code:
I looked into PseudoLocales.java, I understand this to be for the locale-specific data (e.g. found in data/unit/en.txt), this change making this units data accessible via the en_XA and ar_XB pseudolocales, which are used for debugging.

Please let me know if my understanding is lacking / if there's something more I should read!

PR #1290, opened 2020-09-02
ICU-21035 Pass ByteSink from Locale::getKeywordValue() to uloc_getKeywordValue()

roubert:
This eliminates the need for a scratch buffer in Locale::getKeywordValue()
and also the need for counting bytes required in uloc_getKeywordValue(),
something that ByteSink will now handle correctly.

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -719,14 +719,38 @@ uloc_getKeywordValue(const char* localeID,
                      char* buffer, int32_t bufferCapacity,
                      UErrorCode* status)
 {
-    if (buffer != nullptr) {
-        buffer[0] = '\0';
+    if (U_FAILURE(*status)) {
+        return 0;
     }
+
+    CheckedArrayByteSink sink(buffer, bufferCapacity);
+    ulocimp_getKeywordValue(localeID, keywordName, sink, status);
+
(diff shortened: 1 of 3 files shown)


icu4c/source/common/uloc.cpp
```diff
@@ -719,27 +719,51 @@ uloc_getKeywordValue(const char* localeID,
...
-    if (buffer != nullptr) {
-        buffer[0] = '\0';
+    if (U_FAILURE(*status)) {
+        return 0;
+    }
+
+    CheckedArrayByteSink sink(buffer, bufferCapacity);
+    ulocimp_getKeywordValue(localeID, keywordName, sink, status);
+
+    int32_t reslen = sink.NumberOfBytesAppended();
```
markusicu, line 729: FYI -- The rest of the function from here might be a candidate for a helper function in bytesinkutil.h, assuming we do this in more than one place. Food for thought.

markusicu (approved):
lgtm

icu4c/source/common/uloc.cpp
roubert: Once again I hope that when all of this is done one day, this will only be needed in a select few places (and at that point in time it'll also be clearer precisely what the most useful helper function would be).

PR #1291, opened 2020-09-02
ICU-21035 Delete meaningless calls to uloc_getKeywordValue().

roubert:
These two calls to uloc_getKeywordValue() write to the buffer "id" which
then immediately after is overwritten by calls to idForLocale() so they
can simply be removed without any loss of functionality.


markusicu (approved):
lgtm

I wonder if this was intended to do validation, leaving an error in localStatus, but I think at most it would indicate an overflow, and there is no comment explaining why this is here.

PR #1294, opened 2020-09-02
ICU-21035 Update locale implementation to use ulocimp_getKeywordValue().

roubert:


markusicu (approved):
lgtm nice tnx

PR #1296, opened 2020-09-03
ICU-21258 Refactor code and tests for compact data known issue

sffc:
First step: I found the place in library code that needs to change, and I refactored the tests accordingly.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21258
- [x] Issue accepted
- [x] Tests included


sffc:
This is an initial first step that makes it such that we don't throw an exception in this case, and we instead fall back on the old behavior.  I'll work on actually using the new forms next week, but at least now we can consume the new data.

pedberg-icu (approved):
Nice cleanup, thanks

pedberg-icu:
I am going to go ahead and merge this so it is in before I start the next CLDR-ICU integration pass shortly...

PR #1301, opened 2020-09-04
ICU-21257 remove #License fragment from license URLs

markusicu:
The copyright.html page has no such anchor, so `#License` goes nowhere special.

We don't want to point to just the license, we are pointing to the "license & terms of use". Pointing to the whole copyright.html page looks right for that.

https://unicode-org.atlassian.net/browse/ICU-21257


yumaoka (approved):
LGTM

srl295 (approved):
LGTM (spot checked…)

PR #1303, opened 2020-09-04
ICU-21264 Re-enable optimizations for various functions for MSVC with VS2019 and above

jefgen:
A number of functions had the optimizer turned off due to optimizer bugs that were present in VS2015 and VS2017.

Since we can now build with VS 2019 (see PR #1288), we can enable optimization for these functions for MSVC when building with VS2019 or higher.

Note: Unfortunately, it seems that the MSVC compiler (using the latest VS2019) is *still* having issues with the following function:
`void ExtraData::writeExtraData(UChar32 c, Norm &norm)`
https://github.com/unicode-org/icu/blob/master/icu4c/source/tools/gennorm2/extradata.cpp#L189
So we’ll need to keep this one turned off for now.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21264
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -1988,11 +1988,12 @@ _appendPrivateuseToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool 
 #define PRIV 0x0080
 
 /**
- * Ticket #12705 - Visual Studio 2015 Update 3 contains a new code optimizer which has problems optimizing
- * this function. (See https://blogs.msdn.microsoft.com/vcblog/2016/05/04/new-code-optimizer/ )
- * As a workaround, we will turn off optimization just for this function on VS2015 Update 3 and above.
+ * Ticket #12705 - The optimizer in Visual Studio 2015 Update 3 has problems optimizing this function.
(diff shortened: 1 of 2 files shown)


daniel-ju:
Can we also re-enable optimizations in for `ExtraData::writeExtraData` in `icu4c/source/tools/gennorm2/extradata.cpp`?

jefgen:
> Can we also re-enable optimizations in for `ExtraData::writeExtraData` in `icu4c/source/tools/gennorm2/extradata.cpp`?

Good question. :)
I actually tried this, and added a comment to the ICU ticket, I probably should have put it in the PR description as well.
Copy/pasting it here:

Unfortunately, it seems that the MSVC compiler (using the latest VS2019) is *still* having issues with the following function:
`void ExtraData::writeExtraData(UChar32 c, Norm &norm)`
https://github.com/unicode-org/icu/blob/master/icu4c/source/tools/gennorm2/extradata.cpp#L189
So we’ll need to keep this one turned off for now. :(

daniel-ju (approved):
LGTM, thanks!

axelandrejs (approved):

PR #1304, opened 2020-09-04
ICU-21043 Fix calendar problems

FrankYFTang:
This PR fixes
ICU-21043 Erroneous date display in indian calendar of all dates prior to 0001-01-01.
ICU-21044 Hebrew Calendar calculation is incorrect when the year < 1
ICU-21045 Erroneous date display in islamic and islamic-rgsa calendars of all dates prior to 0622-07-18.
ICU-21046 Erroneous date display in islamic-umalqura calendar of all dates prior to  -195366-07-23.

The problem in the indiancal.cpp is
* ICU-21043 the gregorian/julain convesion is wrong. Swith to use
i18n/gregoimp.h fix the problem.

The problem in the hebrwcal.cpp is
* ICU-21044 the use of bulit in / is wrong when the year or month could be < 1.
Fix it by switching to use ClockMath::floorDivide in i18n/gregoimp.h .

The problem in the islamcal.cpp:
* ICU-21045: The math of % negative number for year and month is wrong.
* ICU-21046: Not use int64_t methods and caused int32_t overflow while the year is a
huge negative number. Cast to int64_t to force using the int64_t version
for the math fix the problem.

Also add tests to exhaust test 8000 years for all calendar. In quick
mode, only test 2.5 years.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21043
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/indiancal.cpp
+++ b/icu4c/source/i18n/indiancal.cpp
@@ -83,7 +83,6 @@ static const int32_t LIMITS[UCAL_FIELD_COUNT][4] = {
     {/*N/A*/-1,/*N/A*/-1,/*N/A*/-1,/*N/A*/-1}, // IS_LEAP_MONTH
 };
 
-static const double JULIAN_EPOCH = 1721425.5;
 static const int32_t INDIAN_ERA_START  = 78;
 static const int32_t INDIAN_YEAR_START = 80;
 
@@ -96,7 +95,7 @@ int32_t IndianCalendar::handleGetLimit(UCalendarDateFields field, ELimitType lim
  */
 static UBool isGregorianLeap(int32_t year)
 {
-    return ((year % 4) == 0) && (!(((year % 100) == 0) && ((year % 400) != 0))); 
+    return Grego::isLeapYear(year);
 }
   
 //----------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


FrankYFTang:
Not ready for review, still need to add test for ICU-21046  and still want to fix ICU-21044 together. Please wait.

FrankYFTang:
Currently the test failed on case 36
https://github.com/unicode-org/icu/blame/778d0a6d1d46faa724ead19613bda84621794b72/icu4c/source/test/testdata/calendar.txt
```
               {
                    "en_US@calendar=hebrew",         
                    "MILLIS=-180799750799999",
                    "add",
                    "YEAR=100000", // year + 100000
                    "MILLIS=2974930006022001",
               },
```
which I think is spit out by the following code
https://github.com/unicode-org/icu/blob/ce7e060d50a9af8f631e4684d151cccd5fbbc8c2/icu4c/source/test/intltest/dadrcal.cpp
```
        if(useDate){
        	if(!(toCalendar->getTime(status)==toDate) || U_FAILURE(status)){
        		errln("FAIL: "+caseString+" Match operation had an error: "
        		                    +u_errorName(status));
        	}else{
```

I think that is just the fact my fix will produce a different value than what coded in the test data file for what the value of ms should be after adding 100000 years in Hebrew calendar.

sffc:
When it's ready, convert the PR back to a regular PR.  I also suggest Peter as a better reviewer for this code.

FrankYFTang:
How to remove the DRAFT and into a normal PR? 
IT is actually very simple to review the PR. There are no change of math here. All the fix fall into one of the following three
1. when using 32 bits multiple something and the divide , cast it to 64bits first and cast down to 32 bits after the multiplication and division. this prevent overflow.
2. Replace a / b to ClockMath::floorDivid(a, b)
3. Replace hand code gregorian / julian date conversion logic to a library call.

sffc (approved):
The library changes look good.  I scanned over the tests but didn't review them in great detail.

icu4c/source/i18n/indiancal.cpp
```diff
@@ -141,52 +140,18 @@ int32_t IndianCalendar::handleGetYearLength(int32_t eyear) const {
...
-      (-uprv_floor((year - 1) / 100)) +
-      uprv_floor((year - 1) / 400) +
-      uprv_floor((((367 * month) - 362) / 12) +
-            ((month <= 2) ? 0 :
-             (isGregorianLeap(year) ? -1 : -2)
-            ) +
-            date);
-
-   return julianDay;
+   return Grego::fieldsToDay(year, month - 1, date) + kEpochStartAsJulianDay - 0.5;
```
pedberg-icu, line 143: Isn't month already 0-based here? Should the parameter just be month instead of month-1?

pedberg-icu: (Not completely sure...)

jefgen:
> Also add tests to exhaust test 8000 years for all calendar. 

FWIW, you can trigger an Exhaustive test run by commenting with: `/azp run CI-Exhaustive` :)

icu4c/source/i18n/indiancal.cpp
FrankYFTang: I switch to 0 based month in that file now.

FrankYFTang:
@pedberg-icu  could you take a look. It should be easy to review. There are really no heavy math here. Just replace some 32bits * and / to first cast to 64 bits to avoid overflow, change the month from 1 based to 0 based in internal routine and replace some of the code to other better test utility class.

pedberg-icu (approved):
Thanks, looks like many good cleanups.

PR #1306, opened 2020-09-05
ICU-21174 Drop MaybeStackVector::appendAll()

hugovdm:
Rationale: appendAll() requires T to have a copy constructor. For Units work, we're using MaybeStackVector to store things that cannot be copy-constructed. While Clang succeeds, we have MSVC failures when we export the required symbols.

This was cherry-picked out of the big ICU4C Units-upstreaming PR: https://github.com/unicode-org/icu/pull/1284/commits/2f4561c7a76794b90c800768b407977a2336df5a / https://github.com/unicode-org/icu/pull/1284.

##### Checklist

- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -826,19 +826,6 @@ public:
     T* operator[](ptrdiff_t i) {
         return this->fPool[i];
     }
-
-    /**
-     * Append copies of all the items from another MaybeStackVector to this one.
-     */
-    void appendAll(const MaybeStackVector& other, UErrorCode& status) {
-        for (int32_t i = 0; i < other.fCount; i++) {
-            T* item = emplaceBack(*other[i]);
-            if (!item) {
-                status = U_MEMORY_ALLOCATION_ERROR;
-                return;
-            }
-        }
-    }
 };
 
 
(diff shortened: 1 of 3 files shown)


sffc (approved):
measunit.cpp is the correct place to move the MeasureUnitImpl::copy function to satisfy dependencies.txt requirements. :+1:

hugovdm:
Thanks for the thumbs-up / approval.

For the record, from my perspective: in the context of this PR, I can't reason why MeasureUnitImpl::copy needs to be in measunit.cpp. A brief examination of dependencies.txt leaves me thinking it could go in measunit_extra.cpp. However the problem comes up for 1284, and this was the solution I found with lots of poking and probing. I'm happy to continue with it like this, and potentially have it reconsidered whether there's another viable factoring of this code, after 1284 is merged and working. :-)

sffc:
> Thanks for the thumbs-up / approval.
> 
> For the record, from my perspective: in the context of this PR, I can't reason why MeasureUnitImpl::copy needs to be in measunit.cpp. A brief examination of dependencies.txt leaves me thinking it could go in measunit_extra.cpp. However the problem comes up for 1284, and this was the solution I found with lots of poking and probing. I'm happy to continue with it like this, and potentially have it reconsidered whether there's another viable factoring of this code, after 1284 is merged and working. :-)

It's because the MeasureUnit copy constructor maybe needs to copy a MeasureUnitImpl, which is independent from the rest of the code in measunit_extra.cpp that deals with identifier parsing.

hugovdm:
Wonder if that's worth adding a comment for...

PR #1307, opened 2020-09-07
ICU-21106 avoid buffer overflow in u_fopen_u

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21106
- [x] Issue accepted
- [x] Tests included

This avoids possible buffer overflow in u_fopen_u for long filenames due to use of u_austrcpy with fixed-size buffer. Instead, this uses UnicodeString.extract for conversion to platform codepage, with buffer allocation if result would overflow fixed buffer.

It adds a test opening a file with a very long name. I noticed that existing u_file open tests generally *created* files for writing instead of opening existing files for reading. @grhoten, @jefgen, @markusicu: Do we (still) have platform compatibility issues that might cause problems adding a file with a long name (in the first version of this the filename itself is 254 characters, but the path passed to u_fopen_u is >256) with a specified content encoding (UTF-8)?

### DIFF
--- a/icu4c/source/io/ufile.cpp
+++ b/icu4c/source/io/ufile.cpp
@@ -40,6 +40,7 @@
 #include "unicode/ures.h"
 #include "unicode/ucnv.h"
 #include "unicode/ustring.h"
+#include "unicode/unistr.h"
 #include "cstring.h"
 #include "cmemory.h"
 
@@ -142,18 +143,42 @@ u_fopen(const char    *filename,
     return result; /* not a file leak */
 }
 
+// FILENAME_BUF_MAX represents the largest size that we are willing to use for a
+// stack-allocated buffer to contain a file name or path. If PATH_MAX (POSIX) or MAX_PATH
+// (Windows) are defined and are smaller than this we will use their defined value;
+// otherwise, we will use FILENAME_BUF_MAX for the stack-allocated buffer, and dynamically
+// allocate a buffer for any file name or path that is that length or longer.
+#define FILENAME_BUF_MAX 296
+#if defined PATH_MAX && PATH_MAX < FILENAME_BUF_MAX
+#define FILENAME_BUF_CAPACITY PATH_MAX
+#elif defined MAX_PATH && MAX_PATH < FILENAME_BUF_MAX
+#define FILENAME_BUF_CAPACITY MAX_PATH
+#else
+#define FILENAME_BUF_CAPACITY FILENAME_BUF_MAX
+#endif
+
 U_CAPI UFILE* U_EXPORT2
 u_fopen_u(const UChar   *filename,
         const char    *perm,
         const char    *locale,
         const char    *codepage)
 {
-    UFILE     *result;
-    char buffer[256];
+    UFILE *result;
+    char buffer[FILENAME_BUF_CAPACITY];
+    char *filenameBuffer = buffer;
 
-    u_austrcpy(buffer, filename);
+    icu::UnicodeString filenameString(true, filename, -1); // readonly aliasing, does not allocate memory
+    // extract with conversion to platform default codepage, return full length (not including 0 termination)
+    int32_t filenameLength = filenameString.extract(0, filenameString.length(), filenameBuffer, FILENAME_BUF_CAPACITY);
+    if (filenameLength >= FILENAME_BUF_CAPACITY) { // could not fit (with zero termination) in buffer
+        filenameBuffer = static_cast<char *>(uprv_malloc(++filenameLength)); // add one for zero termination
+        if (!filenameBuffer) {
+            return nullptr;
+        }
+        filenameString.extract(0, filenameString.length(), filenameBuffer, filenameLength);
+    }
 
-    result = u_fopen(buffer, perm, locale, codepage);
+    result = u_fopen(filenameBuffer, perm, locale, codepage);
 #if U_PLATFORM_USES_ONLY_WIN32_API
     /* Try Windows API _wfopen if the above fails. */
     if (!result) {
@@ -161,19 +186,24 @@ u_fopen_u(const UChar   *filename,
         wchar_t wperm[40] = {};
         size_t  retVal;
         mbstowcs_s(&retVal, wperm, UPRV_LENGTHOF(wperm), perm, _TRUNCATE);
-        FILE *systemFile = _wfopen((const wchar_t *)filename, wperm);
+        FILE *systemFile = _wfopen(reinterpret_cast<const wchar_t *>(filename), wperm); // may return NULL for long filename
         if (systemFile) {
             result = finit_owner(systemFile, locale, codepage, TRUE);
         }
-        if (!result) {
+        if (!result && systemFile) {
             /* Something bad happened.
-               Maybe the converter couldn't be opened. */
+               Maybe the converter couldn't be opened.
+               Bu do not fclose(systemFile) if systemFile is NULL. */
             fclose(systemFile);
         }
     }
 #endif
+    if (filenameBuffer != buffer) {
+        uprv_free(filenameBuffer);
+    }
     return result; /* not a file leak */
 }
+
 
 U_CAPI UFILE* U_EXPORT2
 u_fstropen(UChar *stringBuf,
--- a/icu4c/source/test/iotest/filetst.c
+++ b/icu4c/source/test/iotest/filetst.c
@@ -15,6 +15,7 @@
  */
 
 #include "cmemory.h"
+#include "cstring.h"
 #include "iotest.h"
 #include "unicode/ustdio.h"
 #include "unicode/ustring.h"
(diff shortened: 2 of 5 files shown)


pedberg-icu:
> Do we (still) have platform compatibility issues that might cause problems adding a file with a long name (the filename itself is 254 characters, but the path passed to u_fopen_u is >256) ...

Well, Appveyor answered that one right away:
```
git checkout -qf FETCH_HEAD
6error: unable to create file icu4c/source/test/testdata/iotest_long_filename_234567_30_234567_40_234567_50_234567_60_234567_70_234567_80_234567_90_23456_100_23456_110_23456_120_23456_130_23456_140_23456_150_23456_160_23456_170_23456_180_23456_190_23456_200_23456_210_23456_220_23456_230_23456_240_23456_250.txt: Filename too long
7Command exited with code 1
8APPVEYOR_SAVE_CACHE_ON_ERROR is set to 'true' - cache will be updated
```

pedberg-icu:
OK, adjusted this to handle filesystems that may not support names > 255 chars.

pedberg-icu:
OK, just added a few log_infos to help narrow down crash in C: MSVC x86 CI tests.

pedberg-icu:
Adding some temporary printf logging for debugging the CI MSVC crash

icu4c/source/io/ufile.cpp
```diff
@@ -142,39 +143,56 @@ u_fopen(const char    *filename,
     return result; /* not a file leak */
 }
 
+enum { kFilenameBufCapacity = 256 };
```
grhoten, line 146: Shouldn't this be PATH_MAX instead of something arbitrary? I think it's from limits.h. I think putil.cpp uses that value.

grhoten (changes requested):

icu4c/source/io/ufile.cpp
pedberg-icu: > Shouldn't this be PATH_MAX instead of something arbitrary?
Well:
1. The old code just had "256" so this is certainly no worse than that.
2. PATH_MAX is not necessarily available on all platforms where this code needs to work. In common/putil.cpp is use is conditionalized on
```
#if !UCONFIG_NO_FILE_IO && ((U_PLATFORM_IS_DARWIN_BASED && (U_PLATFORM != U_PF_IPHONE || defined(U_TIMEZONE))) || U_PLATFORM_IS_LINUX_BASED || U_PLATFORM == U_PF_BSD || U_PLATFORM == U_PF_SOLARIS)
```
or (ICU_DATA_DIR_PREFIX_ENV_VAR is set mainly for Apple platform builds)
```
#if defined(ICU_DATA_DIR_PREFIX_ENV_VAR)
```
3. PATH_MAX is not that useful anyway, since the limits really depend on filesystem implementations; see e.g. https://eklitzke.org/path-max-is-tricky

jefgen:
Thanks for fixing this Peter!

> > Do we (still) have platform compatibility issues that might cause problems adding a file with a long name (in the first version of this the filename itself is 254 characters, but the path passed to u_fopen_u is >256) ...
> 
> Well, Appveyor answered that one right away:
> 
> ```
> git checkout -qf FETCH_HEAD
> 6error: unable to create file icu4c/source/test/testdata/iotest_long_filename_234567_30_234567_40_234567_50_234567_60_234567_70_234567_80_234567_90_23456_100_23456_110_23456_120_23456_130_23456_140_23456_150_23456_160_23456_170_23456_180_23456_190_23456_200_23456_210_23456_220_23456_230_23456_240_23456_250.txt: Filename too long
> 7Command exited with code 1
> 8APPVEYOR_SAVE_CACHE_ON_ERROR is set to 'true' - cache will be updated
> ```
> 
> Guess I will redo this to try writing the file instead, as with the other tests.

Unfortunately, `MAX_PATH` (value of 260) is still a thing on [Windows platforms](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#paths)...
Though there have been some recent efforts to work towards removing this limitation, but it uses an *opt-in* model, so by default `MAX_PATH` is still a concern.

icu4c/source/io/ufile.cpp
jefgen: FWIW, I think it might be worth adding a brief comment in the code that this is just an arbitrary sized buffer though, and isn't meant to represent any platform or file-system specific size limitation.

icu4c/source/io/ufile.cpp
```diff
@@ -142,39 +143,56 @@ u_fopen(const char    *filename,
...
+    result = u_fopen(filenameBuffer, perm, locale, codepage);
 #if U_PLATFORM_USES_ONLY_WIN32_API
     /* Try Windows API _wfopen if the above fails. */
     if (!result) {
         // TODO: test this code path, including wperm.
         wchar_t wperm[40] = {};
         size_t  retVal;
         mbstowcs_s(&retVal, wperm, UPRV_LENGTHOF(wperm), perm, _TRUNCATE);
-        FILE *systemFile = _wfopen((const wchar_t *)filename, wperm);
+        FILE *systemFile = _wfopen((const wchar_t *)filename, wperm); // may return NULL for filename > 255 chars
```
jefgen, line 177: nit: I know it's existing code... but this function now mixes C style casts with C++ style casts. :) 
Would you mind changing this to a `reinterpret_cast<>` ?

icu4c/source/io/ufile.cpp
grhoten: You could just check if it's defined. If it's defined, use it.  If it's undefined, default to the current value.

jefgen: Well if we're going to do that, then it feels like maybe it should be conditional on `U_PLATFORM_USES_ONLY_WIN32_API`, and `MAX_PATH` should be used for Windows then...

jefgen (dismissed):
FWIW, the changes lgtm.
(just minor nits about comment and the cast though).

icu4c/source/io/ufile.cpp
pedberg-icu: My point was that I _don't_ want to use MAX_PATH even if it _is_ defined, since it could be defined to be a large number (like 4096) that makes no sense for a stack-allocated buffer like this. This buffer size is not meant to be tied to any notion of MAX_PATH; rather, it is only meant to be a not-too-big value (for stack allocation) that will still be larger than almost all filenames/paths in practice; if the name is larger we will dynamically allocate the buffer. We could make it smaller if we wanted to. I can certainly add a comment to this effect.

pedberg-icu: How about this: If MAX_PATH is defined and is a reasonable value for stack allocation (like <= 296) I will use that, otherwise I will use 296 (and explain with comments).

jefgen: FWIW, I'd be happy with just leaving it as an arbitrary fixed-size value, but with a comment to that effect.

However, if @grhoten prefers to have it be tied to `PATH_MAX` / `MAX_PATH`, then I like your suggestion of checking to make sure it's not an overly large value.

pedberg-icu:
OK, I believe I have addressed the review comments.
1. Changed the old `(const wchar_t *) filename` to `reinterpret_cast<const wchar_t *>(filename)`
2. #defined FILENAME_BUF_MAX 296; if PATH_MAX is defined and smaller we will use that instead (the resolved capacity is in  FILENAME_BUF_CAPACITY). Added comments to explain the reasoning.

icu4c/source/test/iotest/filetst.c
```diff
@@ -315,17 +329,28 @@ static void TestFileFromICU(UFILE *myFile) {
...
 static void TestFile(void) {
-
-    log_verbose("Testing u_fopen\n");
-    TestFileFromICU(u_fopen(STANDARD_TEST_FILE, "w", STANDARD_TEST_LOCALE, NULL));
+    UChar ufilename[296];
+ 
+    log_verbose("Testing u_fopen with STANDARD_TEST_FILE\n");
+    TestFileFromICU(u_fopen(STANDARD_TEST_FILE, "w", STANDARD_TEST_LOCALE, NULL), "u_fopen STANDARD");
+
+    u_uastrcpy(ufilename, MEDIUMNAME_TEST_FILE);
```
jefgen, line 337: I wonder if it would be safer to use `u_uastrncpy` with `UPRV_LENGTHOF(ufilename)` for this, ex:

```c++
u_uastrncpy(ufilename, MEDIUMNAME_TEST_FILE, UPRV_LENGTHOF(ufilename)); 
```
Just in the (rare) case somebody ever adds another test case, or perhaps if for some reason the conversion to the host code page needs more room, so that it doesn't overflow the buffer.

pedberg-icu: Good idea, thanks, I addressed this in a slightly different way that also ensures 0 termination of the converted buf.

icu4c/source/io/ufile.cpp
grhoten: Meh, I guess the current changes are fine.

grhoten (approved):

icu4c/source/io/ufile.cpp
```diff
@@ -142,39 +143,68 @@ u_fopen(const char    *filename,
...
+    result = u_fopen(filenameBuffer, perm, locale, codepage);
 #if U_PLATFORM_USES_ONLY_WIN32_API
     /* Try Windows API _wfopen if the above fails. */
     if (!result) {
         // TODO: test this code path, including wperm.
         wchar_t wperm[40] = {};
         size_t  retVal;
         mbstowcs_s(&retVal, wperm, UPRV_LENGTHOF(wperm), perm, _TRUNCATE);
-        FILE *systemFile = _wfopen((const wchar_t *)filename, wperm);
+        FILE *systemFile = _wfopen(reinterpret_cast<const wchar_t *>(filename), wperm); // may return NULL for long filename
```
grhoten, line 189: Out of curiosity, was this meant to be a static cast like the malloc above?

jefgen: Oh good point. Since the type is originally `void *` this could be a `static_cast` instead of `reinterpret_cast`.
(Sorry, I think I was the one that suggested using `reinterpret_cast` here. I saw the `wchar_t*` type, and I guess I'm too used to casting to/from `char16_t*` <-> `wchar_t*` on Windows.)

jefgen (approved):

icu4c/source/io/ufile.cpp
pedberg-icu: Well actually, the original type here is not `void*`, it is `const UChar* filename` i.e. `const char16_t*`, so I think we _do_ want `reinterpret_cast` here.

pedberg-icu:
Thanks for all of the review feedback!

PR #1309, opened 2020-09-08
ICU-21190 Adding PluralRules select for number ranges

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21190
- [ ] Issue accepted
- [ ] Tests included


icu4c/source/i18n/unicode/upluralrules.h
```diff
@@ -179,6 +180,29 @@ uplrules_selectFormatted(const UPluralRules *uplrules,
...
+/**
+ * Given a formatted number range, returns the overall plural form of the
+ * range. For example, "3-5" returns "other" in English.
+ *
+ * To get a UFormattedNumberRange, see UNumberRangeFormatter.
+ *
+ * @param uplrules The UPluralRules object specifying the rules.
+ * @param range The number range onto which the rules will be applied.
+ * @param keyword The destination buffer for the keyword of the rule that
+ *         applies to number.
```
jefgen, line 193: a very minor nit: perhaps "applies to number" -> "applies to the number range".

icu4j/main/classes/core/src/com/ibm/icu/impl/PluralRulesLoader.java
```diff
@@ -28,7 +29,8 @@
  * Loader for plural rules data.
  */
 public class PluralRulesLoader extends PluralRules.Factory {
-    private final Map<String, PluralRules> rulesIdToRules;
+    // Key is locale + "/" + type
```
sffc, line 32: Previously, the key was the "rule set".  Locales that share the same rule set used the same PluralRules instance.  As written, my change will have the result of making the cache substantially larger.  Since plural ranges are also defined by their set, I can improve this a bit by caching on the pair of plural rule set and plural range set.  Alternatively, I can cache the parsed plural rules separately from the parsed plural ranges.

markusicu: If you have time, please either cache range rules separately or build a key from both rule-set strings.

icu4c/source/i18n/plurrule.cpp
```diff
@@ -286,6 +310,33 @@ PluralRules::select(const IFixedDecimal &number) const {
...
+}
+
+UnicodeString
+PluralRules::select(const number::impl::UFormattedNumberRangeData* impl, UErrorCode& status) const {
+    if (fStandardPluralRanges == nullptr) {
+        // Happens if PluralRules was constructed via createRules()
+        status = U_UNSUPPORTED_ERROR;
+        return ICU_Utility::makeBogusString();
+    }
+    if (U_FAILURE(status)) {
```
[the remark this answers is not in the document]
markusicu: I agree, please move the `fStandardPluralRanges == nullptr` check into third place.

icu4c/source/i18n/number_output.cpp
```diff
@@ -47,6 +48,42 @@ void FormattedNumber::getDecimalQuantity(impl::DecimalQuantity& output, UErrorCo
...
+    UPRV_FORMATTED_VALUE_METHOD_GUARD(ICU_Utility::makeBogusString())
+    return fData->quantity1.toScientificString();
+}
+
+UnicodeString FormattedNumberRange::getSecondDecimal(UErrorCode& status) const {
+    UPRV_FORMATTED_VALUE_METHOD_GUARD(ICU_Utility::makeBogusString())
+    return fData->quantity2.toScientificString();
+}
+
+void FormattedNumberRange::getDecimalNumbers(ByteSink& sink1, ByteSink& sink2, UErrorCode& status) const {
```
markusicu, line 65: You will need to rebase again to pick up your recent API addition, right? (After which this here should no longer show up as an addition.)

sffc: My branch is up-to-date.  You are seeing it as added because I moved the definition of FormattedNumberRange from numrange_fluent.cpp to number_output.cpp.  This function is deleted from the first and added to the second.

icu4c/source/i18n/numrange_impl.h
```diff
@@ -105,6 +76,11 @@ class NumberRangeFormatterImpl : public UMemory {
 };
 
 
+/** Helper function used in upluralrules.cpp */
+const UFormattedNumberRangeData* validateUFormattedNumberRange(
```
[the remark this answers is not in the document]
sffc: It's implemented in numrange_capi.cpp.  You left a comment on this function: "indent +4"

icu4c/source/i18n/pluralranges.cpp
```diff
@@ -0,0 +1,138 @@
...
+    PluralRangesDataSink(StandardPluralRanges& output) : fOutput(output) {}
+
+    void put(const char* /*key*/, ResourceValue& value, UBool /*noFallback*/, UErrorCode& status) U_OVERRIDE {
+        ResourceArray entriesArray = value.getArray(status);
+        if (U_FAILURE(status)) { return; }
+        fOutput.setCapacity(entriesArray.getSize());
+        for (int i = 0; entriesArray.getValue(i, value); i++) {
+            ResourceArray pluralFormsArray = value.getArray(status);
+            if (U_FAILURE(status)) { return; }
+            pluralFormsArray.getValue(0, value);
```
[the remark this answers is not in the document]
sffc: I bail if the status is ever a failure.  If the data is ill-formed, the function that should fail is `value.getArray(status)` or `value.getUnicodeString(status)`, right?  By the time I call `getValue()`, the data is already validated.  Right?

icu4c/source/i18n/pluralranges.cpp
```diff
@@ -0,0 +1,138 @@
...
+            if (U_FAILURE(status)) { return; }
+            fOutput.addPluralRange(first, second, result);
+        }
+    }
+
+  private:
+    StandardPluralRanges& fOutput;
+};
+
+void getPluralRangesData(const Locale& locale, StandardPluralRanges& output, UErrorCode& status) {
```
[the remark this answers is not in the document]
sffc: I like the style where all data loading code is inside the anonymous namespace, and the implementation class calls a black-box function to load the data.

icu4c/source/i18n/pluralranges.h
```diff
@@ -0,0 +1,67 @@
...
+    StandardPluralRanges copy(UErrorCode& status) const;
+
+    /** Create an object (called on an rvalue) */
+    LocalPointer<StandardPluralRanges> toPointer(UErrorCode& status) && noexcept;
+
+    /** Select rule based on the first and second forms */
+    StandardPlural::Form resolve(StandardPlural::Form first, StandardPlural::Form second) const;
+
+    /** Used for data loading. */
+    void addPluralRange(
```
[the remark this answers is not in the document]
sffc: Mmm, not quite.  The functions are called from the data sink, which is not a member of the class.  It also cannot be friended, because it is defined in an anonymous namespace.

I think this is not a big deal since this is an internal class.

sffc:
Review comments applied.

I also noticed that there is an `@internal` class in Java named com.ibm.icu.text.PluralRanges.  It looks like a work-in-progress by Mark that was backed by fake static data.  Should I delete it?  (If there are any users, they can switch to StandardPluralRanges, which is public in the impl package)

markusicu:
> Review comments applied.
> 
> I also noticed that there is an `@internal` class in Java named com.ibm.icu.text.PluralRanges. It looks like a work-in-progress by Mark that was backed by fake static data. Should I delete it? (If there are any users, they can switch to StandardPluralRanges, which is public in the impl package)

Don't delete it without checking if it's used in CLDR code. And probably asking Mark, too.

icu4j/main/classes/core/src/com/ibm/icu/text/PluralRanges.java
```diff
@@ -24,7 +24,7 @@
  * @deprecated This API is ICU internal only.
  */
 @Deprecated
-public final class PluralRanges implements Freezable<PluralRanges>, Comparable<PluralRanges> {
+final class PluralRanges implements Freezable<PluralRanges>, Comparable<PluralRanges> {
```
markusicu, line 27: FYI  I just checked the CLDR code, and I don't see any use of "ibm.icu.text.PluralRanges".

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PluralRangesTest.java
```diff
@@ -73,54 +74,48 @@ public void TestRangePattern() {
...
-            assertEquals(i + " Formatting unit", expected, actual);
+            FormattedNumberRange actual = NumberRangeFormatter.with()
+                .numberFormatterBoth(NumberFormatter.with().unit(unit).unitWidth(unitWidth))
+                .locale(locale)
+                .formatRange(low, high);
+            assertEquals(i + " Formatting unit", expected, actual.toString());
         }
     }
 
+    /*
```
markusicu, line 118: Commented out on purpose?

icu4c/source/i18n/pluralranges.cpp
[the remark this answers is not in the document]
markusicu: You get an error from getArray() if it's not an array, and from getUnicodeString() if it's not a string.

However, getValue(i) just returns false if the index i is not 0..getSize()-1, without modifying the value object.
So if you have an array of just two values and you try to read a third one, the value object will still have the second value and you just use that again.

We usually call getValue(i) in a loop over i and stop when it's false. That's why it does not take/set a UErrorCode.

For this code, I think you should check `getSize()>=3` as soon as you have the entriesArray (just before resizing fOutput), and decide what to do if it's too short -- probably set an error (U_RESOURCE_TYPE_MISMATCH?).

Please also check the Java code to see if it has the equivalent problem.

markusicu:
ping

markusicu:
lgtm please squash

markusicu (approved):
lgtm tnx

PR #1311, opened 2020-09-09
ICU-21268 Adds check of string value to prevent segmentation fault

gnrunge:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21268
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/unumberrangeformattertst.c
+++ b/icu4c/source/test/cintltst/unumberrangeformattertst.c
@@ -60,6 +60,9 @@ static void TestExampleCode() {
     int32_t len;
     const UChar* str = ufmtval_getString(unumrf_resultAsValue(uresult, &ec), &len, &ec);
     assertSuccessCheck("There should not be a failure in the example code", &ec, TRUE);
+    if (str == NULL) {
+        str = u"(null)";
+    }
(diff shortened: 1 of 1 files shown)


icu4c/source/test/cintltst/unumberrangeformattertst.c
```diff
@@ -59,6 +59,10 @@ static void TestExampleCode() {
     // Get the result string:
     int32_t len;
     const UChar* str = ufmtval_getString(unumrf_resultAsValue(uresult, &ec), &len, &ec);
+    if (str == NULL) {
```
markusicu, line 62: This bypasses the following assertSuccessCheck.

I think a better fix is probably to put a nullptr check into assertUEquals(). Something like
```
if (str == nullptr) {
    str = u"(null)";
}
```
For each of the expected/actual string pointers.

markusicu: I see that you changed *this* code to use that check. But as I said before, I think it would be better (= help other call sites) to do this *inside* assertUEquals(), for each of the two pointer arguments.

Also please make sure to use ICU indentation.

gnrunge:
Closed and replaced by PR#1312.

PR #1318, opened 2020-09-10
ICU-21196 Clarify API docs for input strings in ucoleiter.h and usearch.h

jefgen:
This change adds notes to the API documentation in `ucoleiter.h` and `usearch.h` that the input string arguments must not be deleted and/or changed while the ICU objects are still potentially using them.

I tried to use wording for this that is similar to existing wording on other APIs in order to have consistency in the library.

Some existing examples:
https://github.com/unicode-org/icu/blob/master/icu4c/source/common/unicode/brkiter.h#L178
```
     * The BreakIterator will retain a reference to the supplied string.
     * The caller must not modify or delete the text while the BreakIterator
     * retains the reference.
```
https://github.com/unicode-org/icu/blob/master/icu4c/source/common/unicode/ubrk.h#L364
```
 * The break iterator retains a pointer to the supplied text.
 * The caller must not modify or delete the text while the BreakIterator
 * retains the reference.
```

This change also fixes a few minor typos in the headers as well.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21196
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/ucoleitr.h
+++ b/icu4c/source/i18n/unicode/ucoleitr.h
@@ -23,7 +23,7 @@
 #if !UCONFIG_NO_COLLATION
 
 /**  
- * This indicates an error has occured during processing or if no more CEs is 
+ * This indicates an error has occurred during processing or if no more CEs is 
  * to be returned.
  * @stable ICU 2.0
  */
@@ -66,14 +66,14 @@ typedef struct UCollationElements UCollationElements;
  * .      UCollationElements *c;
  * .      UCollatorOld *coll;
  * .      UErrorCode success = U_ZERO_ERROR;
- * .      s=(UChar*)malloc(sizeof(UChar) * (strlen("This is a test")+1) );
- * .      u_uastrcpy(s, "This is a test");
+ * .      str=(UChar*)malloc(sizeof(UChar) * (strlen("This is a test")+1) );
+ * .      u_uastrcpy(str, "This is a test");
  * .      coll = ucol_open(NULL, &success);
  * .      c = ucol_openElements(coll, str, u_strlen(str), &status);
  * .      order = ucol_next(c, &success);
  * .      ucol_reset(c);
  * .      order = ucol_prev(c, &success);
- * .      free(s);
+ * .      free(str);
  * .      ucol_close(coll);
  * .      ucol_closeElements(c);
  * .  }
@@ -102,6 +102,10 @@ typedef struct UCollationElements UCollationElements;
 /**
  * Open the collation elements for a string.
  *
+ * The UCollationElements retains a pointer to the supplied text.
+ * The caller must not modify or delete the text while the UCollationElements
+ * object is used to iterate over this text.
+ *
  * @param coll The collator containing the desired collation rules.
  * @param text The text to iterate over.
  * @param textLength The number of characters in text, or -1 if null-terminated
@@ -114,7 +118,6 @@ ucol_openElements(const UCollator  *coll,
                   const UChar      *text,
                         int32_t    textLength,
                         UErrorCode *status);
-
 
 /**
  * get a hash code for a key... Not very useful!
@@ -153,7 +156,7 @@ ucol_reset(UCollationElements *elems);
  * @param elems The UCollationElements containing the text.
  * @param status A pointer to a UErrorCode to receive any errors.
  * @return The next collation elements ordering, otherwise returns UCOL_NULLORDER 
- *         if an error has occured or if the end of string has been reached
+ *         if an error has occurred or if the end of string has been reached
  * @stable ICU 2.0
  */
 U_CAPI int32_t U_EXPORT2 
@@ -168,7 +171,7 @@ ucol_next(UCollationElements *elems, UErrorCode *status);
  *               a U_BUFFER_OVERFLOW_ERROR is returned if the internal stack
  *               buffer has been exhausted.
  * @return The previous collation elements ordering, otherwise returns 
- *         UCOL_NULLORDER if an error has occured or if the start of string has 
+ *         UCOL_NULLORDER if an error has occurred or if the start of string has 
  *         been reached.
  * @stable ICU 2.0
  */
@@ -194,6 +197,11 @@ ucol_getMaxExpansion(const UCollationElements *elems, int32_t order);
  * Property settings for collation will remain the same.
  * In order to reset the iterator to the current collation property settings,
  * the API reset() has to be called.
+ *
+ * The UCollationElements retains a pointer to the supplied text.
+ * The caller must not modify or delete the text while the UCollationElements
+ * object is used to iterate over this text.
+ *
  * @param elems The UCollationElements to set.
  * @param text The source text containing the collation elements.
  * @param textLength The length of text, or -1 if null-terminated.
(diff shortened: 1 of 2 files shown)


markusicu:
FYI  The AppVeyor failure looks like a cygwin+make threading issue again.

icu4c/source/i18n/unicode/ucoleitr.h
```diff
@@ -102,6 +102,10 @@ typedef struct UCollationElements UCollationElements;
 /**
  * Open the collation elements for a string.
  *
+ * The UCollationElements retains a pointer to the supplied text.
```
markusicu, line 105: Misplaced I think. Please move this to ucol_openElements(). We normally document the text lifetime on functions that take text inputs.

icu4c/source/i18n/unicode/ucoleitr.h
```diff
@@ -102,6 +102,10 @@ typedef struct UCollationElements UCollationElements;
...
+ * The caller must not modify or delete the text while the UCollationElements
+ * retains the reference, until after calling <tt>ucol_setText</tt>.
```
markusicu, line 107: also until ucol_closeElements()

Maybe make it more generic? "... while the UCollationElements object is used to iterate over this text"?

Similar for the other places.

I prefer `<code>` over `<tt>`.

icu4c/source/i18n/unicode/usearch.h
```diff
@@ -280,9 +280,15 @@ typedef enum {
 /* open and close ------------------------------------------------------ */
 
 /**
-* Creating a search iterator data struct using the argument locale language
+* Create a String Search iterator data struct using the argument locale language
```
markusicu, line 283: While and where you change docs please use third person "Creates" etc.

daniel-ju (dismissed):

icu4c/source/i18n/unicode/usearch.h
```diff
@@ -280,9 +280,15 @@ typedef enum {
...
 * rule set. A collator will be created in the process, which will be owned by
-* this search and will be deleted in <tt>usearch_close</tt>.
+* this String Search and will be deleted in <tt>usearch_close</tt>.
+*
+* The UStringSearch retains a pointer to both the pattern and text strings.
```
markusicu, line 287: This might be true... but is it actually true about the pattern string?

icu4c/source/i18n/unicode/usearch.h
```diff
@@ -597,6 +614,10 @@ U_CAPI void U_EXPORT2 usearch_setCollator(      UStringSearch *strsrch,
 * Sets the pattern used for matching.
 * Internal data like the Boyer Moore table will be recalculated, but the 
 * iterator's position is unchanged.
+*
+* The UStringSearch retains a pointer to pattern string. The caller must not
```
markusicu, line 618:
```suggestion
* The UStringSearch retains a pointer to the pattern string. The caller must not
```

icu4c/source/i18n/unicode/usearch.h
```diff
@@ -536,11 +549,15 @@ U_CAPI const UBreakIterator * U_EXPORT2 usearch_getBreakIterator(
...
     
 #endif
-    
+
 /**
 * Set the string text to be searched. Text iteration will hence begin at the 
 * start of the text string. This method is useful if you want to re-use an 
 * iterator to search for the same pattern within a different body of text.
+*
+* The UStringSearch retains a pointer to text string. The caller must not
```
markusicu, line 558:
```suggestion
* The UStringSearch retains a pointer to the text string. The caller must not
```

jefgen:
> FYI The AppVeyor failure looks like a cygwin+make threading issue again.

From the logs:
```
ln -fs libicutu68.dll.a ../../lib/libicutu.dll.a
ln -fs libicutu68.dll.a ../../lib/libicutu.dll.a
ln: cannot remove '../../lib/libicutu.dll.a': No such file or directory
```
Indeed it does look like another one...

icu4c/source/i18n/unicode/ucoleitr.h
jefgen: Hmm... I think maybe the diff on GitHub might be confused or perhaps it doesn't show the full comment block(?).
If I expand the diff view on GitHub it looks like this is on the API `ucol_openElements`...

I think if I fix up some of the other changes that you suggested (thanks!), then maybe the diff might be larger, and thus easier to view.

icu4c/source/i18n/unicode/ucoleitr.h
jefgen: Good idea: If it's more generic text then we don't need to update the comments if more APIs are added/etc.

Re: `<code>` vs `<tt>`: I like the `<code>` too, but used `<tt>` since the file seems to only use that. If you like I'll change the whole file to use `<code>` instead.

icu4c/source/i18n/unicode/usearch.h
jefgen: Thanks, will rephrase things.

icu4c/source/i18n/unicode/usearch.h
jefgen: Fair question. :)

I created a simple/hacky test program to double-check here:
https://gist.github.com/jefgen/261424866469a1ad3ec6c35dd16a72e6

The short summary is:
- For the `pattern` input string, if the caller deletes the string then the UStringSearch object will still continue to "work" and give valid results.
- However, for the `text` input string, if the caller deletes the string then the UStringSearch object won't crash -- but it will start giving incorrect or invalid results (as it will be searching the deleted memory).
- That said, for both of them:
If the caller deletes/frees the strings then the two APIs `usearch_getPattern` and `usearch_getText` will give back a pointer to the free'd strings as the result.

Although, perhaps the last one might be expected or even perhaps acceptable maybe...?
But on the other hand, if the caller assumed that the `usearch_open*` API(s) made copies of the input strings, then they might expect to still get back a valid string(s).

icu4c/source/i18n/unicode/usearch.h
jefgen: thanks!

icu4c/source/i18n/unicode/usearch.h
jefgen: thanks!

icu4c/source/i18n/unicode/ucoleitr.h
markusicu: Sorry! I must have been confused by what was shown vs. hidden.

icu4c/source/i18n/unicode/usearch.h
markusicu: Thanks for checking. Let's play it safe.

icu4c/source/i18n/unicode/usearch.h
```diff
@@ -280,9 +280,14 @@ typedef enum {
...
 /**
-* Creating a search iterator data struct using the argument locale language
+* Creates a String Search iterator data struct using the argument locale language
 * rule set. A collator will be created in the process, which will be owned by
-* this search and will be deleted in <tt>usearch_close</tt>.
+* this String Search and will be deleted in <code>usearch_close</code>.
+*
+* The UStringSearch retains a pointer to both the pattern and text strings.
+* The caller must not modify or delete them while using the UStringSearch
+* is using them.
```
markusicu, line 289: I suggest you just remove this line and terminate the previous one with a period.

markusicu (commented):
one more, and please squash

icu4c/source/i18n/unicode/usearch.h
jefgen: Thanks!

jefgen: (I did line and the other one as well.)

markusicu (approved):
lgtm tnx

daniel-ju (approved):

PR #1319, opened 2020-09-10
ICU-21267 stop using FALSE & TRUE macros in most library-internal hea…

markusicu:
…ders

- use standard false & true
- in library-internal headers except some generated ones
- also make the internal-header-test script say it's "inside ICU" so that the generated headers continue to work
  (after which it is probably less necessary to replace FALSE/TRUE but it's a step forward)
- change a few test/tools headers to use false/true as well for the header test to work there

https://unicode-org.atlassian.net/browse/ICU-21267

### DIFF
--- a/icu4c/source/common/bmpset.h
+++ b/icu4c/source/common/bmpset.h
@@ -101,7 +101,7 @@ private:
      */
     UBool latin1Contains[0x100];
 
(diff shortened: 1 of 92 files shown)


gnrunge (approved):
LGTM

hugovdm (commented):
(Post-merge) LGTM!

PR #1323, opened 2020-09-11
ICU-20651 add setContext/getContext for [U]DateIntervalFormat

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20651
- [x] Issue accepted
- [x] Tests included

This adds setContext and getContext to [U]DateIntervalFormat, to control capitalization of the formatted result depending on the context in which it will be used. This API addition was approved by TC 2019-Jun-12 (originally for ICU 65). Because DateIntervalFormat already uses a SimpleDateFormat (in fDateFormat) to format the various parts of the pattern and because SimpleDateFormat already supports setContext, all we need to do is make sure that the context set for DateIntervalFormat is used for the call to SimpleDateFormat that formats the first part of the interval pattern and then no context is set for any subsequent calls to SimpleDateFormat for the remainder of the interval pattern.

### DIFF
--- a/icu4c/source/i18n/dtitvfmt.cpp
+++ b/icu4c/source/i18n/dtitvfmt.cpp
@@ -23,6 +23,7 @@
 #include "unicode/dtptngen.h"
 #include "unicode/dtitvinf.h"
 #include "unicode/simpleformatter.h"
+#include "unicode/udisplaycontext.h"
 #include "cmemory.h"
 #include "cstring.h"
 #include "dtitv_impl.h"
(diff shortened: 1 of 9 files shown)


richgillam (approved):
Looks good.  No complaints from me.

pedberg-icu:
Sorry, updated DateIntervalFormat.java readObject at line 2250 to only force-set fCapitalizationSetting if it is null (i.e. not deserialized from an older version of the object). 
```
-        fCapitalizationSetting = DisplayContext.CAPITALIZATION_NONE;
+        if (fCapitalizationSetting == null) {
+            fCapitalizationSetting = DisplayContext.CAPITALIZATION_NONE;
+        }
```

richgillam (approved):
Still looks good.

sffc (approved):

PR #1324, opened 2020-09-11
ICU-21280 Correct source bytes counting

stima:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21280
- [x] Issue accepted
- [x] Tests included


markusicu (approved):
lgtm
Thanks, good catch!
(I am surprised that no one has complained before.)

I believe that there is another bug that will cause an assertion failure. After merging this PR, I will write another test case and see.

markusicu:
> I believe that there is another bug that will cause an assertion failure.

Nope, everything else I could think of is covered, so never mind, and again thanks!

PR #1325, opened 2020-09-11
ICU-21248 Adds internal header check to Travis Continued Integration.

gnrunge:
Updates test script to return an non zero exit code in case of a
compilation error.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21248
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -186,3 +186,12 @@ matrix:
       cache:
         directories:
         - $HOME/.m2
+
+    # Check compilation of internal headers.
+    - name: "internal header compilation check"
+      language: cpp
+      os:       linux
+      before_script:
+        - cd icu4c/source
+      script:
+        - test/hdrtst/testinternalheaders.sh
--- a/icu4c/source/test/hdrtst/testinternalheaders.sh
+++ b/icu4c/source/test/hdrtst/testinternalheaders.sh
@@ -9,6 +9,8 @@
 CC=clang
 CXX=clang++
 
+ERROR_EXIT=0
+
 # Runtime libraries
 
 for file in `ls common/*.h`; do
@@ -16,6 +18,9 @@ for file in `ls common/*.h`; do
     echo '#include "'$file'"' > ht_temp.cpp ;
     echo 'void noop() {}' >> ht_temp.cpp ;
     $CXX -c -std=c++11 -I common -DU_COMMON_IMPLEMENTATION -O0 ht_temp.cpp ;
+    if [ $? != 0 ] ; then
+        ERROR_EXIT=1
+    fi
 done ;
 
 for file in `ls i18n/*.h`; do
@@ -23,6 +28,9 @@ for file in `ls i18n/*.h`; do
     echo '#include "'$file'"' > ht_temp.cpp ;
     echo 'void noop() {}' >> ht_temp.cpp ;
     $CXX -c -std=c++11 -I common -I i18n -DU_I18N_IMPLEMENTATION -O0 ht_temp.cpp ;
+    if [ $? != 0 ] ; then
+        ERROR_EXIT=1
+    fi
 done ;
 
 for file in `ls io/*.h`; do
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -186,3 +186,12 @@ matrix:
       cache:
         directories:
         - $HOME/.m2
+
+    # Check compilation of internal headers.
+    - name: "internal header compilation check"
+      language: cpp
+      os:       linux
+      before_srcipt:
```
markusicu, line 194: This has got to be a typo, right? --> before_script

icu4c/source/test/hdrtst/testinternalheaders.sh
```diff
@@ -9,27 +9,38 @@
...
+ERROR_EXIT=0
+
 # Runtime libraries
 
 for file in `ls common/*.h`; do
     echo $file
     echo '#include "'$file'"' > ht_temp.cpp ;
     echo 'void noop() {}' >> ht_temp.cpp ;
     $CXX -c -std=c++11 -I common -DU_COMMON_IMPLEMENTATION -O0 ht_temp.cpp ;
+    if [ $? != 0 ] ; then
```
markusicu, line 21: Plausible. (I am no .sh expert. I assume you know this stuff or looked it up.)

Out of curiosity, could you achieve the same thing with replacing the semicolon on the previous line with `|| ERROR_EXIT=1`  ?

markusicu: > I think the most compact syntax would be
> $CXX -c -std=c++11 -I common -DU_COMMON_IMPLEMENTATION -O0 ht_temp.cpp || ERROR_EXIT=1

That's what I had in mind. I wasn't sure if you considered it and didn't like it.

> I can change it thus if you prefer.

I will leave it to you to decide which one is clearer/simpler/readable.

> As for if [ $? != 0 ] || ERROR_EXIT=1

That's not at all what I had in mind.

markusicu (approved):
lgtm

.travis.yml
```diff
@@ -186,3 +186,12 @@ matrix:
...
         - $HOME/.m2
+
+    # Check compilation of internal headers.
+    - name: "internal header compilation check"
+      language: cpp
+      os:       linux
+      before_script:
+        - cd icu4c/source
+      script:
+        - test/hdrtst/testinternalheaders.sh
```
jefgen, line 197: nit: Does this want to start with a `./` (similar to the scripts above)?
(ex: `./test/hdrtst/testinternalheaders.sh`) but I don't think it matters though.

icu4c/source/test/hdrtst/testinternalheaders.sh
```diff
@@ -9,27 +9,38 @@
 CC=clang
 CXX=clang++
 
+ERROR_EXIT=0
```
jefgen, line 12: One thought is that we changed the script to use bash instead of sh then we could perhaps use `set -x -e`, so that the script would exit if any command fails with a non-zero exit status.
However, that would mean the script stops right away on the first error -- whereas I think the intent of the change here is to have *all* header checks run. (That way you don't end up with fixing one error, only to hit another, and so on, etc.)
Maybe a comment about this might be worth adding perhaps?

icu4c/source/test/hdrtst/testinternalheaders.sh
jefgen: I think many (most?) bash-like shells (bash, zsh, etc.) support the special variable `$?`.
Though perhaps one thought is that by using the pipe operator that might not depend on the particular shell implementation as much.

.travis.yml
srl295: Looks ok to me

icu4c/source/test/hdrtst/testinternalheaders.sh
srl295: I think we want to run all tests and get a deferred error. Otherwise you'd have to iterate over every header that's broken.

icu4c/source/test/hdrtst/testinternalheaders.sh
srl295: I think $? As it is is more readable because it makes it clear that there's a control flow, not just an order of commands. $? is the common way to test the prior status, and had the advantage that you can check for a specific status not just nonzero.

srl295 (approved):
Lgtm 

FWIW this is a long standing script, I think the way this was done makes it clear a minimal change is made for automation.

icu4c/source/test/hdrtst/testinternalheaders.sh
jefgen: Right, I was just wonder if it might be worth adding a comment about as such, but perhaps it's obvious from the script.

jefgen (approved):
thanks for automating this. 👍

PR #1330, opened 2020-09-15
ICU-21021 Changing MeasureUnit::splitToSingleUnits to return std::pair

sffc:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21021
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/number_usageprefs.cpp
+++ b/icu4c/source/i18n/number_usageprefs.cpp
@@ -94,14 +94,14 @@ void mixedMeasuresToMicros(const MaybeStackVector<Measure> &measures, DecimalQua
         U_ASSERT(micros->outputUnit.getComplexity(status) == UMEASURE_UNIT_MIXED);
         U_ASSERT(U_SUCCESS(status));
         // Check that we received measurements with the expected MeasureUnits:
-        int32_t singleUnitsCount;
-        LocalArray<MeasureUnit> singleUnits =
-            micros->outputUnit.splitToSingleUnits(singleUnitsCount, status);
+        MeasureUnitImpl temp;
+        const MeasureUnitImpl& impl = MeasureUnitImpl::forMeasureUnit(micros->outputUnit, temp, status);
         U_ASSERT(U_SUCCESS(status));
-        U_ASSERT(measures.length() == singleUnitsCount);
+        U_ASSERT(measures.length() == impl.units.length());
         for (int32_t i = 0; i < measures.length(); i++) {
-            U_ASSERT(measures[i]->getUnit() == singleUnits[i]);
+            U_ASSERT(measures[i]->getUnit() == impl.units[i]->build(status));
         }
+        (void)impl;
 #endif
         // Mixed units: except for the last value, we pass all values to the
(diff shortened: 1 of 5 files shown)


markusicu (dismissed):
lgtm

markusicu:
Oh wait, lots of build bots are unhappy. Please check.

icu4c/source/i18n/number_usageprefs.cpp
```diff
@@ -94,13 +94,12 @@ void mixedMeasuresToMicros(const MaybeStackVector<Measure> &measures, DecimalQua
         U_ASSERT(micros->outputUnit.getComplexity(status) == UMEASURE_UNIT_MIXED);
         U_ASSERT(U_SUCCESS(status));
         // Check that we received measurements with the expected MeasureUnits:
-        int32_t singleUnitsCount;
-        LocalArray<MeasureUnit> singleUnits =
-            micros->outputUnit.splitToSingleUnits(singleUnitsCount, status);
+        MeasureUnitImpl temp;
+        const MeasureUnitImpl& impl = MeasureUnitImpl::forMeasureUnit(micros->outputUnit, temp, status);
```
hugovdm, line 98: Hmm, what's the best resolution for this:

```
number_usageprefs.cpp:98:32: error: unused variable 'impl' [-Werror,-Wunused-variable]
        const MeasureUnitImpl& impl = MeasureUnitImpl::forMeasureUnit(micros->outputUnit, temp, status);
```

This impl is only used in U_ASSERT, which is why I placed the whole block in a U_DEBUG. However I now see U_DEBUG can be defined while U_ASSERT is skipped:

```
#if U_DEBUG
#   include <assert.h>
#   define U_ASSERT(exp) assert(exp)
#elif U_CPLUSPLUS_VERSION
#   define U_ASSERT(exp) (void)0
#else
#   define U_ASSERT(exp)
#endif
```

I must have missed a warning for `singleUnits` then?

Can one have multiple statements together in a U_ASSERT? Perhaps drop the U_ASSERT and rely only on the U_DEBUG guards?

hugovdm:
```
            unitMeasure {
            FAIL: :0: Mixed unit: Unsafe Path (U_INTERNAL_PROGRAM_ERROR)
            FAIL: Mixed unit: Unsafe Path; got ""; expected "3 yd, 1 ft, 11.4 in"
            FAIL: :0: Mixed unit: Safe Path (U_INTERNAL_PROGRAM_ERROR)
            FAIL: Mixed unit: Safe Path; got ""; expected "3 yd, 1 ft, 11.4 in"
            assertFormatSingle destructor: expected success but got error: U_INTERNAL_PROGRAM_ERROR scope: Mixed unit
```

(I can't quickly imagine off the top of my head what the error might be.)

sffc:
I think I must be letting some memory get corrupted somewhere with my new code in the units router. I'll take a closer look in Valgrind this afternoon.

sffc:
OK, let's see if the above amendment fixes the issue.

markusicu (approved):
lgtm

PR #1334, opened 2020-09-15
ICU-21249 Adds #if !UCONFIG_NO_FORMATTING around all code in number_s…

gnrunge:
…ymbolswrapper.cpp.

If UCONFIG_NO_FORMATTING is set to 1 in uconfig.h nothing in
number_symbolswrapper.cpp compiles. Note, for example, that the entirety
of the included numberformatter.h header file is inclosed in
!UCONFIG_NO_FORMATTING.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21249
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/number_symbolswrapper.cpp
+++ b/icu4c/source/i18n/number_symbolswrapper.cpp
@@ -1,6 +1,10 @@
 // © 2020 and later: Unicode, Inc. and others.
 // License & terms of use: http://www.unicode.org/copyright.html
 
+#include "unicode/utypes.h"
+
+#if !UCONFIG_NO_FORMATTING
+
 #include "number_microprops.h"
 #include "unicode/numberformatter.h"
 
@@ -123,3 +127,5 @@ const NumberingSystem *SymbolsWrapper::getNumberingSystem() const {
     U_ASSERT(fType == SYMPTR_NS);
     return fPtr.ns;
 }
+
+#endif /* #if !UCONFIG_NO_FORMATTING */


hugovdm (approved):
Oops, thanks! Double-checked against number_fluent.cpp where I moved this from.

hugovdm:
Is this an easy enough catch that we don't need a CI test for it? (People don't make this mistake often enough, I suspect.)

gnrunge:
In fact, people make the mistake of not considering the UCONFIG_NO_XXX
configuration flags fairly often, and the uconfig header check then mops it
all up.
I am also working on putting the uconfig check into CI, still have to write
a script.

icu4c/source/i18n/number_symbolswrapper.cpp
```diff
@@ -4,6 +4,8 @@
 #include "number_microprops.h"
 #include "unicode/numberformatter.h"
 
+#if !UCONFIG_NO_FORMATTING
```
sffc, line 7: I think what we normally do is

```
// © 2020 and later: Unicode, Inc. and others.

#include "unicode/utypes.h"

#if !UCONFIG_NO_FORMATTING

#include "number_microprops.h"
#include "unicode/numberformatter.h"

// ...

#endif /* #if !UCONFIG_NO_FORMATTING */
```

Basically, put the opening `#if !UCONFIG_NO_FORMATTING` before any other imports besides utypes.h

gnrunge:
Re-arranged the #if construct as suggested.

sffc (approved):

PR #1335, opened 2020-09-15
ICU-21287 Enable asserts for 'ant checkTest'

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21287
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/build.xml
+++ b/icu4j/build.xml
@@ -267,6 +267,8 @@
             <mkdir dir="${junit.out.dir}/checkTest"/>
 
             <junit printsummary="yes" fork="yes">
+                <jvmarg value="-ea"/>
+                <jvmarg value="-Djava.awt.headless=true"/>
                 <jvmarg value="-DICU.exhaustive=10"/>
                 <classpath>
                     <path refid="junit.jars"/>


icu4j/build.xml
```diff
@@ -267,6 +267,8 @@
             <mkdir dir="${junit.out.dir}/checkTest"/>
 
             <junit printsummary="yes" fork="yes">
+                <jvmarg value="-ea"/>
+                <jvmarg value="-Djava.awt.headless=true"/>
```
srl295, line 271: Maybe i'm missing something, but why isn't checkTest implemented in terms of ` <icu-junit  … ` but with an expanded fileset ? seems like that would solve this without keeping duplicate code. 

And related, Why would you use `ant checkTest` instead of `ant check … ` with parameters selecting just one subtest?

srl295: `checkTest` seems to be "new" with ICU-13273, not clear why a separate target was needed rather than adding methods="${testnames}" to `check`?

At any rate, i see `-ea` duplicated in the file, might be time to move it into a variable ( defaultTestJvmOpts or something) so that its common andnot missed.

markusicu: @srl295 I suspect that the people involved didn't know better... and http://site.icu-project.org/setup/java/ant#TOC-Test-arguments-and-running-just-one-test-or-the-tests-of-just-one-test-class was adjusted back then, too.

If there is a better/simpler way to do this, please suggest & help document.

srl295: I might be thinking of the equivalent CLDR items.

srl295 (approved):
I think it is a good improvement as is - LGTM with a suggestion

icu4j/build.xml
srl295: I'm probably missing something.

LGTM with a comment: I did notice there were 'user-jvm-options' and similar under the `icu-junit` macro I mentioned (link below). Should these be included as well? As above, `-ea` and the headless arg might be good common variables (repeated a few times in this file).

https://github.com/unicode-org/icu/blob/cb662b8725815e266c3e930e5ed980f75b0c3422/icu4j/build.xml#L324-L329

yumaoka (approved):
LGTM

markusicu:
The AppVeyor failure is the "usual" cygwin makefile threading bug, please ignore.

hugovdm:
I think we agreed I can go ahead and merge without doing the "common arguments" thing at this time. I like the suggestions, but I don't have time for them right now. And I benefit from this commit cherry-picked into branches when reviewing Java PRs.

markusicu:
> I think we agreed I can go ahead and merge without ...

yes

PR #1343, opened 2020-09-17
ICU-21248 Adds source file check (UTF-8 and absence of BOM) to

gnrunge:
Travis Continued Integration.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21248
- [x] Issue accepted
- [ ] Tests included

Screen Shot 2020-09-17 at 3 33 31 PM

### DIFF
--- a/tools/scripts/icu-file-utf8-check.py
+++ b/tools/scripts/icu-file-utf8-check.py
@@ -22,6 +22,8 @@
 #  The tool operates recursively on the directory from which it is run.
 #  Only files from the ICU github repository are checked.
 #  No changes are made to the repository; only the working copy will be altered.
+#  The script checks all source files and returns a non-zero exit code if any of
+#  the checked files contain a non-UTF-8 character.
 
 from __future__ import print_function
 
@@ -31,6 +33,10 @@ import os.path
 import re
 import getopt
 
+
+# List of directories to check for UTF-8 and BOM. Currently covers
+# all of icu/. Modify as needed.
+icu_directories_to_be_scanned = ["."]
 
 def runCommand(cmd):
     output_file = os.popen(cmd);
@@ -45,13 +51,14 @@ def runCommand(cmd):
 def usage():
     print("usage: " + sys.argv[0] + " [-h | --help]")
 
-    
+
 #
 #  File check.         Check source code files for UTF-8 and all except text files for not containing a BOM
 #    file_name:        name of a text file.
 #    is_source:        Flag, set to True if file is a source code file (.c, .cpp, .h, .java).
 #
 def check_file(file_name, is_source):
+    rc = 0
     f = open(file_name, 'rb')
     bytes = f.read()
     f.close()
@@ -61,16 +68,19 @@ def check_file(file_name, is_source):
             bytes.decode("UTF-8")
         except UnicodeDecodeError:
             print("Error: %s is a source code file but contains non-utf-8 bytes." % file_name)
-    
+            rc = 1
+
     if bytes[0] == 0xef:
         if not (file_name.endswith(".txt") or file_name.endswith(".sln")
-                    or file_name.endswith(".targets")
-                    or ".vcxproj" in file_name):
+                    or file_name.endswith(".targets") or ".vcxproj" in file_name):
             print("Warning: file %s contains a UTF-8 BOM: " % file_name)
+            rc = 1
 
-    return
+    return rc
 
 def main(argv):
+    exit_status = 0
+
     try:
         opts, args = getopt.getopt(argv, "h", ("help"))
     except getopt.GetoptError:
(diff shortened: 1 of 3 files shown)


gnrunge:
Question: ICU also has the icu/tools/scripts/ directory, which may be a better location for this script. Any comments?

icu4c/source/tools/icu-file-utf8-check.py
```diff
@@ -22,6 +22,8 @@
 #  The tool operates recursively on the directory from which it is run.
 #  Only files from the ICU github repository are checked.
 #  No changes are made to the repository; only the working copy will be altered.
+#  The script checks all source files and keeps a non-zero exit code if any of
```
jefgen, line 25: nit: For the phrase, "and keeps a non-zero exit code", would "and returns a non-zero exit code" maybe be a slightly more common way to say it? I'm not sure...
(I see the script does have a rc variable, so it does "keep" track of the return code, but a caller of the program might not care I guess.)

icu4c/source/tools/icu-file-utf8-check.py
```diff
@@ -61,16 +67,20 @@ def check_file(file_name, is_source):
             bytes.decode("UTF-8")
         except UnicodeDecodeError:
             print("Error: %s is a source code file but contains non-utf-8 bytes." % file_name)
-    
+            rc = 1
+
     if bytes[0] == 0xef:
         if not (file_name.endswith(".txt") or file_name.endswith(".sln")
-                    or file_name.endswith(".targets")
+                    or file_name.endswith(".targets") or file_name.endswith(".md")
```
jefgen, line 74: Are the markdown files all UTF-8 with BOM? I wonder if we might want to remove the BOM on them too...

icu4c/source/tools/icu-file-utf8-check.py
```diff
@@ -84,23 +94,30 @@ def main(argv):
...
-        check_file(f, source_file)
+        sys.exit(2)
+
+    git_cmd = "git ls-files DIR"
+    for dir in icu_directories_to_be_scanned:
+        print('Scanning ' + dir)
+        output = runCommand(git_cmd.replace("DIR", dir))
+        file_list = output.splitlines()
+
+        source_file_re = re.compile(".*((?:\\.c$)|(?:\\.cpp$)|(?:\\.h$)|(?:\\.java$))")
```
jefgen, line 105: I wonder if this regex could be moved outside of the for loop.
(Not that perf is really that big of a concern here though...)

jefgen:
> Question: ICU also has the icu/tools/scripts/ directory, which may be a better location for this script. Any comments?

Good question. 
FWIW, I think having this under `icu4c/source` seems a bit odd since it's used for checking `icu4j` as well.

The Git/Jira commit checker script lives under `icu/tools/commit-checker/`.
I wonder if perhaps something like `icu/tools/utf8-checker/` ?

markusicu:
> > Question: ICU also has the icu/tools/scripts/ directory, which may be a better location for this script. Any comments?
> 
> Good question.
> FWIW, I think having this under `icu4c/source` seems a bit odd since it's used for checking `icu4j` as well.
> 
> The Git/Jira commit checker script lives under `icu/tools/commit-checker/`.
> I wonder if perhaps something like `icu/tools/utf8-checker/` ?

https://github.com/unicode-org/icu/tree/master/tools/scripts is definitely a better place.
As a single script I don't think it needs a new subfolder.

icu4c/source/tools/icu-file-utf8-check.py
markusicu: Hopefully the Markdown files do *not* start with a BOM.

jefgen: Right -- I don't know if we want to start skipping the `.md` files.

gnrunge: The script flagged one MD file:
icu4c/source/test/testdata/break_rules/README.md
I thought it's close enough to a .txt file and added .md to the filter. Apparently MD files are not that close after all.

icu4c/source/tools/icu-file-utf8-check.py
gnrunge: I like your wording - done.

icu4c/source/tools/icu-file-utf8-check.py
gnrunge: Valid point -done.

gnrunge:
Please take a look again. The BOM is gone from the README.md, albeit not visible in the diff view.
Thx.

srl295 (dismissed):

tools/scripts/icu-file-utf8-check.py
```diff
@@ -32,6 +34,9 @@
 import getopt
 
 
+# List of directories to check for UTF-8 and BOM. Extend as needed.
+icu_directories_to_be_scanned = ["icu4c/source", "icu4j"]
```
markusicu, line 38: Does it work if we scan all files? Or at least "tools"? Or is there too much amiss right now?

markusicu (dismissed):
lgtm

gnrunge:
No findings reported when executed from the root icu/ directory on all the
9,267 files listed by git ls-files.

markusicu:
> No findings reported when executed from the root icu/ directory on all the 9,267 files listed by git ls-files.

Well, maybe change the list of directories to just `["."]`? Keep it as a list, so we can go back to specific folders if needed.

gnrunge:
Coverage of UTF-8/BOM check now all of icu/.

markusicu (approved):
lgtm tnx!

jefgen (approved):
Thanks @gnrunge! :)

PR #1349, opened 2020-09-22
ICU-21270 Update FixedDecimal in Java to support exponent

echeran:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21270
- [X] Issue accepted
- [X] Tests included

@sffc Let me know if this is reasonable or at least if it is on the right track.  FYI, the test data came from Mark's commit at https://github.com/unicode-org/cldr/pull/712/files.  Also, in case it's useful, Mark's PR solves the problem alternatively by subclassing FixedDecimal.

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/PluralRules.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/PluralRules.java
@@ -547,6 +547,8 @@ public class PluralRules implements Serializable {
 
         final boolean isNegative;
 
+        final int exponent;
+
         private final int baseFactor;
 
         /**
@@ -639,9 +641,10 @@ public class PluralRules implements Serializable {
          * @param v number of digits to the right of the decimal place. e.g 1.00 = 2 25. = 0
          * @param f Corresponds to f in the plural rules grammar.
          *   The digits to the right of the decimal place as an integer. e.g 1.10 = 10
+         * @param e Suppressed exponent for scientific and compact notation
          */
         @Deprecated
-        public FixedDecimal(double n, int v, long f) {
+        public FixedDecimal(double n, int v, long f, int e) {
             isNegative = n < 0;
             source = isNegative ? -n : n;
             visibleDecimalDigitCount = v;
@@ -649,6 +652,7 @@ public class PluralRules implements Serializable {
             integerValue = n > MAX
                     ? MAX
                             : (long)n;
+            exponent = e;
             hasIntegerValue = source == integerValue;
             // check values. TODO make into unit test.
             //
@@ -677,6 +681,24 @@ public class PluralRules implements Serializable {
                 visibleDecimalDigitCountWithoutTrailingZeros = trimmedCount;
             }
             baseFactor = (int) Math.pow(10, v);
+        }
+
+        /**
+         * @internal CLDR
+         * @deprecated This API is ICU internal only.
+         */
+        @Deprecated
+        public FixedDecimal(double n, int v, long f) {
+            this(n, v, f, 0);
+        }
+
+        /**
+         * @internal CLDR
+         * @deprecated This API is ICU internal only.
+         */
+        @Deprecated
+        public static FixedDecimal createWithExponent(double n, int v, int e) {
+            return new FixedDecimal(n,v,getFractionalDigits(n, v), e);
         }
 
         /**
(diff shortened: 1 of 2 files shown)


icu4j/main/classes/core/src/com/ibm/icu/text/PluralRules.java
```diff
@@ -639,16 +643,19 @@ public int getBaseFactor() {
...
+        public FixedDecimal(double n, int v, long f, int e, boolean hasE) {
             isNegative = n < 0;
             source = isNegative ? -n : n;
             visibleDecimalDigitCount = v;
             decimalDigits = f;
             integerValue = n > MAX
                     ? MAX
                             : (long)n;
+            exponent = e;
+            hasExponent = hasE;
```
sffc, line 658: Why do you need the boolean?  If `e` is not present, it default to 0 per the spec.  There is no difference between "1" and "1E0" in the eyes of the spec, unless I am mistaken.

echeran: The reason that I needed the boolean here was because I needed to update `toString()` in order to support the test helper function's assertion statements.

The need for `toString()` was not present in the previous user-facing work for exponent support, which occurred in DecimalQuantity, etc., since we only need to support operand value computation (as a means for plural keyword selection).

sffc: Are there cases where you are printing "1E0" where you can't just print "1"?

echeran: Yeah, although the problem is actually the inverse (I would be printing "1" where it should be "1e0", for example).  It's happening in the test helper method in PluralRulesTest for testing plural rule samples called [`checkNewSamples`](https://github.com/unicode-org/icu/blob/master/icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PluralRulesTest.java#L202), which I'm reusing here in my tests.  One of the things you pass to it is a `PluralRules` object, then the method gets the samples samples from the rules object.  And then one of the assert statements in the method tests whether the `.toString()` of the samples object is equal to the provided samples string.

I tried taking away the `hasExponent` field to confirm the error again, and this is the error message.  It is complaining specifically that `"@integer 0, 1, 1e5"` is not equal to `"@integer 0, 1, 1"`:

```
java.lang.AssertionError: : samples; one: i = 0,1 @integer 0, 1, 1e5 @decimal 0.0~1.5, 1.1e5; many: e = 0 and i != 0 and i % 1000000 = 0 and v = 0 or e != 0..5 @integer 1000000, 2e6, 3e6, 4e6, 5e6, 6e6, 7e6, … @decimal 2.1e6, 3.1e6, 4.1e6, 5.1e6, 6.1e6, 7.1e6, …; other:  @integer 2~17, 100, 1000, 10000, 100000, 2e5, 3e5, 4e5, 5e5, 6e5, 7e5, … @decimal 2.0~3.5, 10.0, 100.0, 1000.0, 10000.0, 100000.0, 1000000.0, 2.1e5, 3.1e5, 4.1e5, 5.1e5, 6.1e5, 7.1e5, …, INTEGER: expected "@integer 0, 1, 1e5", got "@integer 0, 1, 1"
	at org.junit.Assert.fail(Assert.java:88)
	at com.ibm.icu.dev.test.AbstractTestLog.errln(AbstractTestLog.java:80)
	at com.ibm.icu.dev.test.TestFmwk.handleAssert(TestFmwk.java:666)
	at com.ibm.icu.dev.test.TestFmwk.handleAssert(TestFmwk.java:647)
	at com.ibm.icu.dev.test.TestFmwk.assertEquals(TestFmwk.java:599)
	at com.ibm.icu.dev.test.format.PluralRulesTest.checkNewSamples(PluralRulesTest.java:234)
	at com.ibm.icu.dev.test.format.PluralRulesTest.testSamplesWithExponent(PluralRulesTest.java:206)
``` 

Of course, 1e5 and 1 are different numbers, but the above stacktrace is only showing how the testing helper method and its assert statements are working.  More to the point you were raising, in a previous iteration of the code, without knowing that the exponent is set, then the `.toString()` of a FixedDecimal of something like 1.0e6 would come out as "1000000.0", so the sample strings wouldn't match as expected.

echeran: I talked offline with @sffc .  He said that in other places in our codebase, when e == 0, we do not print 'e0' in the formatted number.  Based on that and the lack of samples with an explicit 0 exponent, I removed the `hasExponent` boolean.

srl295 (approved):
Oh, and changes LGTM too

srl295 (dismissed):
RSLGTM (unchanged since my last LGTM)

icu4j/main/classes/core/src/com/ibm/icu/text/PluralRules.java
```diff
@@ -639,16 +641,18 @@ public int getBaseFactor() {
          * @param v number of digits to the right of the decimal place. e.g 1.00 = 2 25. = 0
          * @param f Corresponds to f in the plural rules grammar.
          *   The digits to the right of the decimal place as an integer. e.g 1.10 = 10
+         * @param e Suppressed exponent for scientific and compact notation
          */
         @Deprecated
-        public FixedDecimal(double n, int v, long f) {
+        public FixedDecimal(double n, int v, long f, int e, boolean hasE) {
```
sffc, line 647: You don't need the `hasE` argument any more

sffc (changes requested):

srl295 (approved):
Change lgtm

sffc (approved):

PR #1350, opened 2020-09-22
ICU-20590 Ensure time format consistency for SHORT vs 'j' patterns if no locale data

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20590
- [x] Issue accepted
- [x] Tests included

When there is no specific data for a locale, ensure that standard time formats (primarily the SHORT version) at least match the same formats that would be obtained using DateFormat.getInstanceForSkeleton or DateTimePatternGenerator.getBestPattern using skeletons with 'j' such as "jmm"; this ensures that the time formats at least have the correct hour cycle for the specified or inferred region.

To do this, ICU4C SimpleDateFormat and ICU4J Calendar call DateTimePatternGenerator to generate the appropriate time patterns when there is no locale data. To prevent circular dependencies from DateTimePatternGenerator on SimpleDateFormat/Calendar, the DateTimePatternGenerator instantiated from SimpleDateFormat/Calendar is a slightly stripped-down version that still provides what is needed for the purpose.

This is basically what Apple ICU has been doing up through the first half of 2020. For releases in fall 2020 it has a more comprehensive approach to handling locales without data, which will also be proposed for open ICU (but is a more extensive change).

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -311,6 +311,16 @@ DateTimePatternGenerator::createInstance(const Locale& locale, UErrorCode& statu
     return U_SUCCESS(status) ? result.orphan() : nullptr;
 }
 
+DateTimePatternGenerator* U_EXPORT2
+DateTimePatternGenerator::createInstanceNoStdPat(const Locale& locale, UErrorCode& status) {
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+    LocalPointer<DateTimePatternGenerator> result(
+            new DateTimePatternGenerator(locale, status, true), status);
+    return U_SUCCESS(status) ? result.orphan() : nullptr;
+}
+
 DateTimePatternGenerator*  U_EXPORT2
 DateTimePatternGenerator::createEmptyInstance(UErrorCode& status) {
     if (U_FAILURE(status)) {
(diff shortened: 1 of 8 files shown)


richgillam (approved):
LGTM.

Quick historical question: With the advent of skeletons, why didn't ICU go back and use them for the standard full-long-medium-short date and time patterns?  Seems like that would have made a lot of this code much simpler.

pedberg-icu:
> Quick historical question: With the advent of skeletons, why didn't ICU go back and use them for the standard full-long-medium-short date and time patterns? Seems like that would have made a lot of this code much simpler.

There are tickets (and plans) to do precisely that (I think Shane filed one of them). One of the issues is that in locales for which we do have data, there are many cases in which the actual standard patterns do not match what we would get from mapping skeletons using availableFormats data. That is probably something that needs to be fixed in CLDR before we can do what you suggest.

mihnita:
My best guess is a combination of these:* Tried to stay backwards compatible* Tried to stay compatible with JDK, which does not have skeletons* For some locales it might make sense to have differences between what fields to show
  (for instance time zones are useful / common for multi-time zone counties, but not for single time zone countries)* Consistency with the OS (although that was lost, if it was ever possible)  Most major OSes allow the user to customize a long / short date format  https://www.windowscentral.com/sites/wpcentral.com/files/styles/xlarge/public/field/image/2016/12/formats-additional-settings.jpg
  https://www.macobserver.com/wp-content/uploads/2017/08/Default_Date_Settings.jpg  Would be nice to respect these settings. I would like if ICU would format a "long" date the way I defined the long date for my OS.  It would require platform specific code, would only, apply to the current locale, so it would be difficult to implement properly.

Mihai
    On Tuesday, September 22, 2020, 11:09:09 AM PDT, Peter Edberg <notifications@github.com> wrote:  
 
 



Quick historical question: With the advent of skeletons, why didn't ICU go back and use them for the standard full-long-medium-short date and time patterns? Seems like that would have made a lot of this code much simpler.


There are tickets (and plans) to do precisely that (I think Shane filed one of them). One of the issues is that in locales for which we do have data, there are many cases in which the actual standard patterns do not match what we would get from mapping skeletons using availableFormats data. That is probably something that needs to be fixed in CLDR before we can do what you suggest.

PR #1353, opened 2020-09-22
ICU-21249 BRS68RC: Fixes and changes for the ICU4C Samples.

jefgen:
This PR makes changes/fixes for the ICU 68 BRS task for "ICU4C Samples", in order to get the samples to compile with some of the changes made in ICU 68. (The change to use C++/C99 `true`/`false` instead of of the custom ICU macros for them).

Changes in this PR include:
- Change to use `true`/`false` instead of the custom `TRUE`/`FALSE` macros in order to fix compilation errors.
- Adding casts to prevent/silence warnings.
- Some clean-up to the "ufortune" and "ugrep" Visual Studio project `.vcxproj` files.
- Setting the codepage to UTF-8 in `samplecheck.bat` in order to prevent garbled output when running the samples.

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/samples/coll/coll.cpp
+++ b/icu4c/source/samples/coll/coll.cpp
@@ -47,13 +47,13 @@ const char gHelpString[] =
  */
 char * opt_locale     = "en_US";
 char * opt_rules      = 0;
-UBool  opt_help       = FALSE;
-UBool  opt_norm       = FALSE;
-UBool  opt_french     = FALSE;
-UBool  opt_shifted    = FALSE;
-UBool  opt_lower      = FALSE;
-UBool  opt_upper      = FALSE;
-UBool  opt_case       = FALSE;
+UBool  opt_help       = false;
+UBool  opt_norm       = false;
+UBool  opt_french     = false;
+UBool  opt_shifted    = false;
+UBool  opt_lower      = false;
+UBool  opt_upper      = false;
+UBool  opt_case       = false;
 int    opt_level      = 0;
 char * opt_source     = "abc";
 char * opt_target     = "abd";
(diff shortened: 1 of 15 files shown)


icu4c/source/samples/coll/coll.cpp
```diff
@@ -244,12 +244,12 @@ UBool processCollator()
  */
 int main(int argc, const char** argv) 
 {
-    if (processOptions(argc, argv, opts) != TRUE || opt_help) {
+    if (static_cast<bool>(processOptions(argc, argv, opts)) != true || static_cast<bool>(opt_help)) {
```
daniel-ju, line 247: Kind of a nit, but do we still need to compare equal/not equal to true here, versus using/not using `!` (same in other places/files)?

jefgen: Oh good point, since these are `bool`s now, we should be able to just use an if statement. :) 
I'll force-push an update later.

daniel-ju (approved):

icu4c/source/samples/ucnv/convsamp.cpp
```diff
@@ -342,7 +342,7 @@ UErrorCode convsample_05()
         
         ucnv_toUnicode(conv, &target, targetLimit, 
                        &source, sourceLimit, NULL,
-                       feof(f)?TRUE:FALSE,         /* pass 'flush' when eof */
+                       feof(f)?true:false,         /* pass 'flush' when eof */
```
erik0686, line 345: Could this be the same as Daniel mentioned before? removing the condition? Same for cases below?

jefgen: Hmm, good question...

The documentation for the function `feof` only says that it returns a "nonzero value":
> The feof function returns a nonzero value if a read operation has attempted to read past the end of the file; it returns 0 otherwise.

However, the function `ucnv_toUnicode` takes a `UBool`, which is a `typedef` for `int8_t`. This means that it **only** wants a value of `1` for things that are "true". So because of this I think we need to keep the condition here, as we need to possibly handle cases where `feof` returns back something like 2.
In other words, we need it to work like this: ` feof returns 2 --> true --> 1`.
So I think this needs to stay as-is.

erik0686: Ah that is a good point :D Thanks!

erik0686 (approved):

PR #1354, opened 2020-09-22
ICU-21028 Modifies ICU data generator so that the CLDR version is

gnrunge:
no longer added to the .txt data files with the exception of the
root.txt files. This will drastically limit the number of changed
files during a version upgrade because the version change is the
only change in many of the data files.

ICU-21028 Fixed an indent.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21028
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
+++ b/tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
@@ -348,7 +348,9 @@ public final class LdmlConverter {
                 });
 
                 if (!splitData.getPaths().isEmpty() || isBaseLanguage || dir.includeEmpty()) {
-                    splitData.setVersion(cldrVersion);
+                    if (id.equals("root")) {
+                        splitData.setVersion(cldrVersion);
+                    }
                     write(splitData, outDir, false);
                     writtenLocaleIds.put(dir, id);
                 }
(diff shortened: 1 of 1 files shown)


tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -348,7 +348,9 @@ private void processLdml() {
                 });
 
                 if (!splitData.getPaths().isEmpty() || isBaseLanguage || dir.includeEmpty()) {
-                    splitData.setVersion(cldrVersion);
+                    if (id.equals("root")) {
+                        splitData.setVersion(cldrVersion);
+                  }
```
markusicu, line 353: indent +2

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/LdmlConverter.java
```diff
@@ -558,10 +560,6 @@ private void writeEmptyFile(String id, Path dir, Collection<String> aliasTargets
         if (aliasTargets.contains(id)) {
             icuData.setFileComment("generated alias target");
             icuData.add(RB_EMPTY_ALIAS, "");
-        } else {
```
markusicu, line 561: Please keep the else and the comment and just delete the setVersion() line.

If it bothers you that there is nothing being done, you could add another comment line like
```
// No need to add data: Just write a resource bundle with an empty top-level table.
```

gnrunge: Done.

markusicu (approved):
lgtm tnx

PR #1357, opened 2020-09-23
ICU-20846 remove header guard from .cpp file

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20846


FrankYFTang (approved):
ok, common/unicode/localematcher.h is defining __LOCALEMATCHER_H__ not __LOCMATCHER_H__ so this should be ok.

PR #1361, opened 2020-09-24
ICU-21297 Fix crash when ICU runs with stubdata only

FrankYFTang:
Reproduce procedure
1. build and test normally
2. cd icu4c/source
   cp stubdata/libicudata.so.60.1 lib/
   cd test/intltest
   INTLTEST_OPTS=-w make check
   cd ../cintltst
   CINTLTST_OPTS=-w make check

Fixed 4 issues:
   In locid.cpp return memory error if the singleton is nullptr
   In test code, stop testing after checked failure.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21297
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -633,7 +633,11 @@ private:
  */
 class AliasData : public UMemory {
 public:
-    static const AliasData* singleton(UErrorCode status) {
+    static const AliasData* singleton(UErrorCode& status) {
+        if (U_FAILURE(status)) {
+            // Do not get into loadData if the status already has error.
+            return nullptr;
+        }
         umtx_initOnce(AliasData::gInitOnce, &AliasData::loadData, status);
         return gSingleton;
     }
@@ -881,9 +885,6 @@ AliasData::loadData(UErrorCode &status)
     UDate end = uprv_getRawUTCtime();
     printf("AliasData::loadData took total %f ms\n", end - start);
 #endif  // LOCALE_CANONICALIZATION_DEBUG
-    if (gSingleton == nullptr) {
-        status = U_MEMORY_ALLOCATION_ERROR;
-    }
 }
 
 /**
@@ -904,6 +905,9 @@ AliasDataBuilder::build(UErrorCode &status) {
     LocalUResourceBundlePointer variantAlias(
         ures_getByKey(metadataAlias.getAlias(), "variant", nullptr, &status));
 
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
     int32_t languagesLength = 0, scriptLength = 0, territoryLength = 0,
             variantLength = 0;
 
@@ -989,13 +993,22 @@ AliasDataBuilder::build(UErrorCode &status) {
                        status);
     }
 
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+
     // copy hashtables
-    return new AliasData(
+    auto *data = new AliasData(
         std::move(languageMap),
         std::move(scriptMap),
         std::move(territoryMap),
         std::move(variantMap),
         strings.orphanCharStrings());
+
+    if (data == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+    }
+    return data;
 }
 
 /**
@@ -1460,6 +1473,7 @@ AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
     if (U_FAILURE(status)) {
         return false;
     }
+    U_ASSERT(data != nullptr);
     out.clear();
     language = locale.getLanguage();
     if (!notEmpty(language)) {
--- a/icu4c/source/test/cintltst/unumberrangeformattertst.c
+++ b/icu4c/source/test/cintltst/unumberrangeformattertst.c
@@ -65,9 +65,10 @@ static void TestExampleCode() {
     // Get the result string:
     int32_t len;
     const UChar* str = ufmtval_getString(unumrf_resultAsValue(uresult, &ec), &len, &ec);
-    assertSuccessCheck("There should not be a failure in the example code", &ec, TRUE);
-    assertUEquals("Should produce expected string result", u"$3 – $5", str);
-    assertIntEquals("Length should be as expected", u_strlen(str), len);
+    if (assertSuccessCheck("There should not be a failure in the example code", &ec, TRUE)) {
+        assertUEquals("Should produce expected string result", u"$3 – $5", str);
+        assertIntEquals("Length should be as expected", u_strlen(str), len);
+    }
 
     // Cleanup:
     unumrf_close(uformatter);
--- a/icu4c/source/test/intltest/plurults.cpp
+++ b/icu4c/source/test/intltest/plurults.cpp
@@ -1160,6 +1160,11 @@ void PluralRulesTest::testSelectTrailingZeros() {
         status.setScope(message);
         Locale locale(cas.localeName);
         LocalPointer<PluralRules> rules(PluralRules::forLocale(locale, status));
+        if (U_FAILURE(status)) {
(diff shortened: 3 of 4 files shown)


icu4c/source/common/locid.cpp
```diff
@@ -1457,7 +1460,8 @@ bool
 AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
 {
     data = AliasData::singleton(status);
-    if (U_FAILURE(status)) {
+    if (U_FAILURE(status) || data == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
```
markusicu, line 1464: Don't clobber an existing failure.
Better:
```
if (U_FAILURE(status)) {
    return false;
}
if (data == nullptr) {
    return false;
}
```

Alternatively, and a bit cleaner I think, would be to check for nullptr after "new" at the end of the singleton() function, and set the memory error there. Then you need not change this call site. Basically, singleton() should either return with "success" and a non-nullptr, or a failure and nullptr.

markusicu: PS: Before `return new AliasData(` please check for failure and return nullptr.

markusicu: So... I realize that singleton() is not the right place. The "new" is in `AliasDataBuilder::build()`. And once that behaves as it should, you should then remove the `if (gSingleton == nullptr) {` check from `AliasData::loadData()` so that another error code does not get clobbered.

Sorry for not paying more attention during my earlier reviews.

FrankYFTang: I somehow missed a & in the loadData and cause issue and didn't realize till now.
Also we should return earlier if the status has already error so we won't call the init once code with status is error because that code will not be run again. 
We also need to set the error after the line of calling the init once call and set the status to memory error because the loadData will not be called the second time. 

PTAL

icu4c/source/common/locid.cpp
```diff
@@ -633,8 +633,21 @@ class AliasDataBuilder {
...
 class AliasData : public UMemory {
 public:
-    static const AliasData* singleton(UErrorCode status) {
+    static const AliasData* singleton(UErrorCode& status) {
+        if (U_FAILURE(status)) {
+            // Do not get into loadData if the status already has error.
+            return nullptr;
+        }
         umtx_initOnce(AliasData::gInitOnce, &AliasData::loadData, status);
+        if (U_FAILURE(status)) {
```
markusicu, line 642: optional: This works, but if loadData() returns nullptr if-and-only-if U_SUCCESS(status) then this singleton() function should be able to return gSingleton without doing anything else.

icu4c/source/common/locid.cpp
```diff
@@ -881,6 +894,9 @@ AliasData::loadData(UErrorCode &status)
     UDate end = uprv_getRawUTCtime();
     printf("AliasData::loadData took total %f ms\n", end - start);
 #endif  // LOCALE_CANONICALIZATION_DEBUG
+    if (U_FAILURE(status)) {
```
markusicu, line 897: optional: Same thing here -- if build() is well-behaved, then you could remove this failure check and the following nullptr check.

icu4c/source/common/locid.cpp
```diff
@@ -904,6 +920,9 @@ AliasDataBuilder::build(UErrorCode &status) {
     LocalUResourceBundlePointer variantAlias(
         ures_getByKey(metadataAlias.getAlias(), "variant", nullptr, &status));
 
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
     int32_t languagesLength = 0, scriptLength = 0, territoryLength = 0,
             variantLength = 0;
```
markusicu, line 928: (Note: GitHub won't let me comment directly on a line far from where you have made changes already.)

Please do change the end of this build() function from
```
    // copy hashtables
    return new AliasData(
        std::move(languageMap),
        std::move(scriptMap),
        std::move(territoryMap),
        std::move(variantMap),
        strings.orphanCharStrings());
```
to
```
    if (U_FAILURE(status)) {
        return nullptr;
    }

    // copy hashtables
    auto *data = new AliasData(
        std::move(languageMap),
        std::move(scriptMap),
        std::move(territoryMap),
        std::move(variantMap),
        strings.orphanCharStrings());
    if (data == nullptr) {
    }
    return data;
```

icu4c/source/common/locid.cpp
FrankYFTang: the first time the singleton() got called, it will cause AliasData::loadData to be called, and AliasData::loadData may set the status to error and gSingleton to nullptr. if we do not check the error and return here, then the next line will clobber the error to U_MEMORY_ALLOCATION_ERROR but the error might be some other error.

icu4c/source/common/locid.cpp
FrankYFTang: no, I cannot, if I skip this status check, then the error would be set to U_MEMORY_ALLOCATION_ERROR while it is point to a different error, set inside builder.build()

icu4c/source/common/locid.cpp
markusicu: What I mean is, after your latest commit (thanks!), where the build() function becomes well-behaved, you should be able to shorten this function here to just:
```
    static const AliasData* singleton(UErrorCode& status) {
        if (U_FAILURE(status)) {
            // Do not get into loadData if the status already has error.
            return nullptr;
        }
        umtx_initOnce(AliasData::gInitOnce, &AliasData::loadData, status);
        return gSingleton;
    }
```

icu4c/source/common/locid.cpp
markusicu: What I mean is, after your latest commit ... you should be able to shorten this function here to just:
```
void U_CALLCONV
AliasData::loadData(UErrorCode &status)
{
#ifdef LOCALE_CANONICALIZATION_DEBUG
    UDate start = uprv_getRawUTCtime();
#endif  // LOCALE_CANONICALIZATION_DEBUG
    ucln_common_registerCleanup(UCLN_COMMON_LOCALE_ALIAS, cleanup);
    AliasDataBuilder builder;
    gSingleton = builder.build(status);
#ifdef LOCALE_CANONICALIZATION_DEBUG
    UDate end = uprv_getRawUTCtime();
    printf("AliasData::loadData took total %f ms\n", end - start);
#endif  // LOCALE_CANONICALIZATION_DEBUG
}
```

icu4c/source/common/locid.cpp
FrankYFTang: I don't understand how could I do this.
let's say this got called twice, the first time it come in, umtx_initOnce called AliasData::loadData and caused error, the status is set to error and gSingleton is set to nullptr
now the second time it is called, the umtx_initOnce skip calling AliasData::loadData because it has been called once, the status is still no error and gSingleton is still nullptr
and we return gSingleton, which mean we return nullptr

Then... in AliasReplacer::replace, I check the status and return if it is error, but the status is not error because singleton neither call  AliasData::loadData nor set the error as what you suggested above, so the code continue, and the next line is U_ASSERT(data != nullptr)
then it will bark if we run under a debug build but will continue if it is on a release build and then hit the same problem we have before.

FrankYFTang: ok, I just realize umtx_initOnce would cache the status from the first load and set it to the later calls even it skip the calling for loadData. Now it make sense to me.

markusicu: The failure code from the first load is cached in the UInitOnce object and umtx_initOnce() will set it again on subsequent calls. You need not do anything special to remember the failure.

markusicu (commented):
lgtm please squash

markusicu (approved):
lgtm tnx

PR #1364, opened 2020-09-24
ICU-21269 ParagraphLayout complexTable: init & access based on actual values

markusicu:
Let the `complexTable[]` length be determined by the compiler, and check the array index against the actual array length.

https://unicode-org.atlassian.net/browse/ICU-21269

FYI  I also see where scriptCodeCount is used but not where it's defined.

### DIFF
--- a/icu4c/source/layoutex/ParagraphLayout.cpp
+++ b/icu4c/source/layoutex/ParagraphLayout.cpp
@@ -137,7 +137,7 @@ le_int32 StyleRuns::getRuns(le_int32 runLimits[], le_int32 styleIndices[])
  * process, rather for all scripts which require
  * complex processing for correct rendering.
  */
-static const le_bool complexTable[scriptCodeCount] = {
+static const le_bool complexTable[] = {
     FALSE , /* Zyyy */
     FALSE,  /* Qaai */
     TRUE,   /* Arab */
@@ -974,7 +974,7 @@ le_int32 ParagraphLayout::getLanguageCode(const Locale *locale)
 
 le_bool ParagraphLayout::isComplex(UScriptCode script)
 {
-    if (script < 0 || script >= (UScriptCode) scriptCodeCount) {
+    if (script < 0 || script >= ARRAY_SIZE(complexTable)) {
         return FALSE;
     }
 


srl295 (approved):

srl295:
scriptCodeCount is written by `ScriptIDModuleWriter.java` which should be run as a BRS task

markusicu:
> `scriptCodeCount` is in https://github.com/harfbuzz/icu-le-hb/blob/b8a3163a4e7309c8684f74edaf416a96f7cd6ac8/src/LEScripts.h#L208 (the harfbuzz shim).

Right, I just got there, too. Shorter link: https://github.com/harfbuzz/icu-le-hb/blob/master/src/LEScripts.h

What's strange is that the current file there has `scriptCodeCount = 175`. The layoutex code should compile fine with *fewer* initializers than the declared array length. However, the bug report is due to a *smaller* scriptCodeCount = 130 which fails to compile because there are more initializers than that. I wonder where Jai picked up an even older header file. My changes here should make it "work" anyway.

More interestingly, ICU no longer has a header file that defines this constant or the layout engine script codes. That got farmed out to the HarfBuzz sub-project. But ICU's layoutex code still defines a table with initializers per script which we haven't kept up to date. This ICU code should probably defer the "isComplex()" determination to the underlying HarfBuzz, via the icu-le-hb layer.

FYI  "isComplex(string)" is stable public API (so should be kept) but appears unused inside layoutex.

Looking at my unidata/changes.txt log, the last time I updated the layout engine script information was for Unicode 9.0/ICU 58. The layout engine itself was deleted between that and the next Unicode update for ICU.
- https://github.com/unicode-org/icu/commits/master/icu4c/source/layout/LEScripts.h
- ICU-12708 "remove deprecated LayoutEngine" for ICU 58
- The last version there already had `scriptCodeCount = 175`
- Current ICU 67 (Unicode 13) has `USCRIPT_CODE_LIMIT    = 193`
- A scriptCodeCount of 130 would have to come from an LEScripts.h from ICU 4.0 or 4.2

khaledhosny:
> This ICU code should probably defer the "isComplex()" determination to the underlying HarfBuzz

What is this function used for? HarfBuzz does not have a notion complex and non-complex scripts.

khaledhosny:
Based on the documentation, this is a misguided API and should be a no-op that always returns true:
https://github.com/unicode-org/icu/blob/1c8bc8078939f01ec9763db2586627d69bc3cea1/icu4c/source/layoutex/layout/ParagraphLayout.h#L433-L445

markusicu:
> Based on the documentation, this is a misguided API and should be a no-op that always returns true:

Most likely it relates to how ICU's former layout engine worked.
Feel free to submit an ICU ticket proposing to deprecate this API and make it always return true or whatever seems right.

I don't know who might be calling this function and thus might be affected. Maybe look for other open source code that's using layoutex? Might be safest to just deprecate it without changing what it does.

khaledhosny:
Since the list is not kept up to date, any code depending on this to decide whether a script needs a real layout engine or can be hacked by “just mapping the characters to glyphs and rendering them in order” is already broken for a few scripts, and even some of the existing entries are questionable.

PR #1366, opened 2020-09-25
ICU-21249 restore tests of Version resources

markusicu:
... which PR #1365 commented out to work around ICU-21028 removing the Version string resource from most locale data bundles. The root bundles in each locale data tree still have a Version string, so this PR restores the test cases with a little more leniency.

https://unicode-org.atlassian.net/browse/ICU-21249 (ICU 68rc BRS)


yumaoka (approved):
LGTM

PR #1370, opened 2020-09-26
Icu 8144

jowilco:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-8144
- [ ] Issue accepted
- [ ] Tests included


sffc (changes requested):
Thank you, and sorry for the very much delayed review!

We need to think more about the lifecycle of the CurrencyDisplayNames objects.  Who owns them?  If you request one from the same locale multiple times in a row, are they saved in a cache?  Although I don't generally like caches, I think this is a case where we might want to consider pulling in the UnifiedCache.

sffc:
(Closing and opening the PR to trigger fresh CI)

sffc:
You currently have the following build error:

```
../../lib/libicuuc.so: undefined reference to `icu_69::CurrencyDisplayNames::getName(char16_t const*, UCurrNameStyle, UErrorCode&) const'
../../lib/libicui18n.so: undefined reference to `icu_69::CurrencyDisplayNames::getPluralName(char16_t const*, icu_69::PluralMapBase::Category, UErrorCode&) const'
../../lib/libicui18n.so: undefined reference to `icu_69::CurrencyDisplayNames::getInstance(icu_69::Locale const*, UErrorCode&)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:81: recipe for target '../../bin/makeconv' failed
```

That error usually means either you haven't defined the functions, or you haven't told the compiler where to find them.  I think you're in the second case.  You added "currencydisplaynames.cpp" to the vcxproj files, but you also need to add it to the file *icu4c/source/common/sources.txt* for Unixy systems.

PR #1371, opened 2020-09-28
ICU-21249 ICU4J 68 API status updates

yumaoka:
See https://docs.google.com/spreadsheets/d/1OIezmSf2mN-hiRnIrpwugCHTzCFIm_-GyeA3ng4QFP8

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
@@ -28,7 +28,6 @@ import com.ibm.icu.text.TimeZoneFormat.Style;
 import com.ibm.icu.text.TimeZoneFormat.TimeType;
 import com.ibm.icu.text.TimeZoneNames;
 import com.ibm.icu.text.TimeZoneNames.NameType;
-import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.ULocale.Category;
 
 /**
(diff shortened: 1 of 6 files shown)


icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
```diff
@@ -924,7 +923,9 @@ public static String getEquivalentID(String id, int index) {
      * @return A <code>TimeZone</code>. Clients are responsible for deleting the
      *   <code>TimeZone</code> object returned.
      * @internal
+     * @deprecated This API is ICU internal only.
      */
+    @Deprecated
     public static TimeZone forULocaleOrDefault(ULocale locale) {
```
sffc, line 929: I think this is a useful API to propose to be draft.  @FrankYFTang

sffc (approved):

icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
yumaoka: It was marked as @internal. This change was made as a part of locale keyword "timezone" support. When you create a new Calendar object, you should be able to specify a locale with "timezone" keyword.

It does not make much sense to create a TimeZone with a locale object. This method added in this release just check if Locale object has "timezone" keyword specified. If not, returns default time zone. So users directly creating TimeZone object has no benefit to create a TimeZone object in this way.

I personally think it can be a part of abstract Calendar class, not necessarily be in TimeZone class.

FrankYFTang: I have no strong opinion either way. I think it should be a private method that needed for some other classes inside ICU to call it. I won't propose an API change to include it for public usage but won't stop anyone else to do that.

yumaoka: Anyway, we did not have any API discussion about this - so I'd leave it `@internal`.

This is basically TimeZone.getTimeZone(uloc.getKeywordValue("timezone")). So it's quite straight forward for ICU4J users even without this one to be exposed.

markusicu (approved):
lgtm

PR #1373, opened 2020-09-29
ICU-21222 Fix Era in Interval format

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21222
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/dtitvfmt.cpp
+++ b/icu4c/source/i18n/dtitvfmt.cpp
@@ -863,6 +863,14 @@ DateIntervalFormat::initializePattern(UErrorCode& status) {
                 setPatternInfo(UCAL_DATE, nullptr, &pattern, fInfo->getDefaultOrder());
                 setPatternInfo(UCAL_MONTH, nullptr, &pattern, fInfo->getDefaultOrder());
                 setPatternInfo(UCAL_YEAR, nullptr, &pattern, fInfo->getDefaultOrder());
+
+                timeSkeleton.insert(0, CAP_G);
+                pattern = DateFormat::getBestPattern(
+                        locale, timeSkeleton, status);
+                if ( U_FAILURE(status) ) {
+                    return;
+                }
+                setPatternInfo(UCAL_ERA, nullptr, &pattern, fInfo->getDefaultOrder());
             } else {
                 // TODO: fall back
             }
@@ -889,15 +897,23 @@ DateIntervalFormat::initializePattern(UErrorCode& status) {
         setPatternInfo(UCAL_DATE, nullptr, &pattern, fInfo->getDefaultOrder());
         setPatternInfo(UCAL_MONTH, nullptr, &pattern, fInfo->getDefaultOrder());
         setPatternInfo(UCAL_YEAR, nullptr, &pattern, fInfo->getDefaultOrder());
+
+        timeSkeleton.insert(0, CAP_G);
+        pattern = DateFormat::getBestPattern(
+                locale, timeSkeleton, status);
+        if ( U_FAILURE(status) ) {
+            return;
+        }
+        setPatternInfo(UCAL_ERA, nullptr, &pattern, fInfo->getDefaultOrder());
     } else {
         /* if both present,
-         * 1) when the year, month, or day differs,
+         * 1) when the era, year, month, or day differs,
          * concatenate the two original expressions with a separator between,
          * 2) otherwise, present the date followed by the
          * range expression for the time.
          */
         /*
-         * 1) when the year, month, or day differs,
+         * 1) when the era, year, month, or day differs,
          * concatenate the two original expressions with a separator between,
          */
         // if field exists, use fall back
@@ -916,6 +932,11 @@ DateIntervalFormat::initializePattern(UErrorCode& status) {
             // then prefix skeleton with 'y'
             skeleton.insert(0, LOW_Y);
             setFallbackPattern(UCAL_YEAR, skeleton, status);
+        }
+        if ( !fieldExistsInSkeleton(UCAL_ERA, dateSkeleton) ) {
+            // then prefix skeleton with 'G'
+            skeleton.insert(0, CAP_G);
+            setFallbackPattern(UCAL_ERA, skeleton, status);
         }
 
         /*
--- a/icu4c/source/test/intltest/dtifmtts.cpp
+++ b/icu4c/source/test/intltest/dtifmtts.cpp
@@ -45,25 +45,27 @@
 
 void DateIntervalFormatTest::runIndexedTest( int32_t index, UBool exec, const char* &name, char* /*par*/ ) {
     if (exec) logln("TestSuite DateIntervalFormat");
-    switch (index) {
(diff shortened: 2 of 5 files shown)


icu4c/source/i18n/dtitvfmt.cpp
```diff
@@ -863,6 +863,7 @@ DateIntervalFormat::initializePattern(UErrorCode& status) {
                 setPatternInfo(UCAL_DATE, nullptr, &pattern, fInfo->getDefaultOrder());
                 setPatternInfo(UCAL_MONTH, nullptr, &pattern, fInfo->getDefaultOrder());
                 setPatternInfo(UCAL_YEAR, nullptr, &pattern, fInfo->getDefaultOrder());
+                setPatternInfo(UCAL_ERA, nullptr, &pattern, fInfo->getDefaultOrder());
```
fabalbon, line 866: The pattern you are adding here does not include the era. You should probably call getBestPattern() after prepending "G" to the timeSkeleton (similar to what you are doing in line 894).

icu4c/source/test/intltest/dtifmtts.cpp
```diff
@@ -2135,4 +2137,212 @@ void DateIntervalFormatTest::testTicket20707() {
...
+                 formatted.toString(status));
+
+    formatted = g->formatToValue(bothBC, status);
+    assertEquals("Gregorian - calendar both dates in BC",
+                 u"4/5/124, 6 AM \u2013 4/5/123, 6 AM",
+                 formatted.toString(status));
+
+    formatted = g->formatToValue(BCtoAD, status);
+    assertEquals("Gregorian - BC to AD",
+                 u"4 5, 123 BC, 6 AM \u2013 4 5, 124 AD, 6 AM",
```
richgillam, line 2219: How come the format changed here?  I would have expected "4/5/123 BC", for example.  Do we not have that pattern?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/DateIntervalFormatTest.java
```diff
@@ -2303,4 +2305,153 @@ public void testFormatMillisecond() {
...
+        cal.set(Calendar.MONTH, Calendar.JUNE);
+        cal.set(Calendar.DAY_OF_MONTH, 5);
+        Date date2019Jun5 = cal.getTime();
+
+        DateInterval bothBeforeRawa = new DateInterval(date2019Mar2.getTime(), date2019Apr3.getTime());
+        DateInterval beforeAfterRawa = new DateInterval(date2019Mar2.getTime(), date2019May4.getTime());
+        DateInterval bothAfterRawa = new DateInterval(date2019May4.getTime(), date2019Jun5.getTime());
+
+        FormattedDateInterval formatted = japanese.formatToValue(bothAfterRawa);
+        assertEquals("japanese calendar - both dates in Rawa",
```
richgillam, line 2440: Shouldn't that be "Reiwa"?

richgillam (dismissed):
Looks good, I think, although I did have a couple questions.

icu4c/source/test/intltest/dtifmtts.cpp
FrankYFTang: >  I would have expected "4/5/123 BC", for example. Do we not have that pattern?

we do NOT have
 GyMd{"M/d/y G"}
but we have
 GyMMMd{"MMM d, y G"}

Here is what I understand why we got this, 
for bothAD and bothBC, we need to display yMdBh while the largest different unit is y between the two time. notice the format is "h" 
since there are no pattern for yMdh in intervalFormats
It has to use fallback{"{0} – {1}"} inside intervalFormats to glue the two different date separated formatted by SimpleDateFormat
so now SimpleDateFormat try to format each of the time for yMdh
but there are no yMdh in availableFormats
so it has to use the the glue pattern in DateTimePatterns ("{1}, {0}") to glue the formatted value of yMd and the formatted value of Bh.
now, we found pattern for yMd as yMd{"M/d/y"} so the formatted result is "4/5/124" and "4/5/123"
and we found pattern for h  as h{"h a"} so the formatted result is "6 AM"
so the DateTimePatterns ("{1}, {0}") to glue the start time to "4/5/124, 6AM" and the end time to "4/5/123, 6AM" 
and fallback{"{0} – {1}"} glue them as "4/5/124, 6AM – 4/5/123, 6AM" 
(The fact that we do not display "BC" is another separated issue)

now, for the case of BCtoAD
we need to display GyMdBh while the largest different unit is G between the two time. notice the format is "h" 
since there are no pattern for GyMdBh in intervalFormats
It has to use fallback{"{0} – {1}"} inside intervalFormats to glue the two different date separated formatted by SimpleDateFormat (this is the same as the case of bothBC and both AD)

so now SimpleDateFormat try to format each of the time for GyMdh
but there are no GyMdh in availableFormats
so it has to use the the glue pattern in DateTimePatterns ("{1}, {0}") to glue the formatted value of GyMd and the formatted value of h.
now, we found pattern for GyMd from  GyMMMd{"MMM d, y G"}  so the formatted result is "4 5, 123 BC" and "4 5, 124 AD".   [hum.... we should get "April 5, 123 BC" and "April 5, 124 AD" in that case I think....)
and we found  pattern for h  as h{"h a"} so the formatted result is "6 AM"
so the DateTimePatterns ("{1}, {0}") to glue the start time to "4 5, 123 BC, 6AM" and the end time to "4 5, 124 AD, 6AM" 
and fallback{"{0} – {1}"} glue them as "4 5, 123 BC, 6AM – 4 5, 124 AD, 6AM"

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/DateIntervalFormatTest.java
FrankYFTang: good catch!

FrankYFTang (commented):
PTAL

icu4c/source/test/intltest/dtifmtts.cpp
richgillam: I think the thing that's throwing me here is that the month names in the `GyMd` pattern (actually, I guess it's the `GyMMMd` pattern, based on what you're saying) are coming out as numerals instead of abbreviated month names.  If this is English with the Gregorian calendar, "4 5, 123 BC" is an extremely unnatural way to render the data.  I'm assuming we get that because we're matching the `GyMMMd` pattern and then algorithmically changing `MMM` to `M` to match what was originally requested?  But if the original pattern passed in was just `h`, who's supplying the extra fields and deciding what lengths they should have?  Something seems off here.

I think _your_ code is working correctly, so I have no problem approving this PR, but I'm wondering if it's worth filing a ticket to make the output here more consistent.

richgillam (dismissed):

FrankYFTang:
squashed, please approval again. Thanks

richgillam (approved):

PR #1379, opened 2020-09-30
ICU-20568 Improve negative measure handling for mixed units

hugovdm:
Without this PR, negative number formatting for mixed units is several kinds of screwy.

With this PR: "-1 foot, 3 inches". First number gets the negative, the remaining numbers are formatted positively. (TBD: whether this is correct for all locales. It's definitely a huge improvement for common European languages and languages with similar writing systems for numbers.)

Closes: https://github.com/icu-units/icu/issues/63.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20568
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/number_longnames.cpp
+++ b/icu4c/source/i18n/number_longnames.cpp
@@ -455,6 +455,8 @@ const Modifier *MixedUnitLongNameHandler::getMixedUnitModifier(DecimalQuantity &
         status = U_UNSUPPORTED_ERROR;
         return &micros.helpers.emptyWeakModifier;
     }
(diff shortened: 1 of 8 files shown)


sffc (commented):
Please undo the fifth commit and squash commits 1-4

hugovdm:
Clarified this offline:

> Please undo the fifth commit

Shane's happy either way ("I think #1379 is fine either way, with or without the last commit"), Mark and I are convinced by English examples that negative on all units is too bizarre, so I'm keeping it.

Squashing!

sffc (approved):

PR #1382, opened 2020-09-30
ICU-21249 Fixed some coding problems found by Spotbugs

yumaoka:
ICU-21249 clone to copy

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
@@ -128,7 +128,7 @@ public class UnitConverter {
             assert (!factor.isEmpty());
 
             // Remove all spaces in the factor
-            factor.replaceAll("\\s+", "");
+            factor = factor.replaceAll("\\s+", "");
 
             String[] fractions = factor.split("/");
             assert (fractions.length == 1 || fractions.length == 2);
@@ -153,10 +153,9 @@ public class UnitConverter {
         }
 
         /**
-         * Clone this <code>Factor</code>.
+         * Copy this <code>Factor</code>.
          */
-        @Override
-        protected Factor clone() {
+        protected Factor copy() {
             Factor result = new Factor();
             result.factorNum = this.factorNum;
             result.factorDen = this.factorDen;
@@ -175,7 +174,7 @@ public class UnitConverter {
          * Returns a single `BigDecimal` that represent the conversion rate after substituting all the constants.
          */
         public BigDecimal getConversionRate() {
-            Factor resultCollector = this.clone();
+            Factor resultCollector = this.copy();
 
             resultCollector.substitute(new BigDecimal("0.3048"), this.CONSTANT_FT2M);
             resultCollector.substitute(new BigDecimal("411557987.0").divide(new BigDecimal("131002976.0"), DECIMAL128), this.CONSTANT_PI);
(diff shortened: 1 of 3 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -128,7 +128,7 @@ public static Factor processFactor(String factor) {
             assert (!factor.isEmpty());
 
             // Remove all spaces in the factor
-            factor.replaceAll("\\s+", "");
```
yumaoka, line 131: This is obviously a coding error. `replaceAll()` is not changing `String factor` in place (String is immutable anyway).

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -153,9 +153,9 @@ private static Factor processFactorWithoutDivision(String factorWithoutDivision)
         }
 
         /**
-         * Clone this <code>Factor</code>.
+         * Copy this <code>Factor</code>.
          */
-        protected Factor clone() {
```
yumaoka, line 158: If method name is `clone()`, it overrides `Object.clone()`. Returning covariant type in subclass is OK since Java 5.

When you implement clone() method, you should start with `super.clone()` which creates a new instance of the object in optimized way. Calling a constructor of the class is off from the convention. For this specific implementation, the end result will be equivalent, but such implementation does not work if parent is not `Object`.

If you really want to override `Object#clone()`, you should add `implements Cloneable` to the class. Again, this is not a strict contract, but it's out of standard convention. Static code analysis tool like Spotbugs tend to produce a warning.

I could change this method to standard `clone()` implementation, and add `implements Cloneable`. But after checking call site, changing the method name to `copy()` would be the less impact change.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsData.java
```diff
@@ -168,7 +168,7 @@ public CategoriesSink() {
 
         @Override
         public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
-            assert (key.toString() == Constants.CATEGORY_TABLE_NAME);
```
yumaoka, line 171: This does not currently cause assertion failure. As long as the String is interned, `==` operation works. However, it looks `toString()` here seems to call `substring()`, which may create a new String instance.

Anyway, I don't think this assertion is not expecting String object's identity. We may hit this assertion unexpectedly with `==` through a certain code path, or future code changes. I believe this assertion is for String value equality, so we should use `equals()` method here. Again, static code analysis tools don't like such code.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1544,7 +1544,7 @@ private boolean replaceRegion() {
             } else {
                 replacedRegion = replacement.get(0);
             }
-            assert this.region != replacedRegion;
```
yumaoka, line 1547: Same as above - we should use `String#equals()`

markusicu (dismissed):
lgtm

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: Thanks for the explanation/education!

hugovdm (approved):

sffc (dismissed):

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsData.java
sffc: Yeah, good catch; I didn't see when reviewing this code that these were strings.

markusicu (approved):
lgtm

PR #1393, opened 2020-10-01
ICU-21249 BRS68RC Remaining updates for Readmes for ICU4C and ICU4J

echeran:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21249
- [X] Issue accepted
- [ ] Tests included

These are the remaining changes for updating the Readme files for the BRS for 68 RC.

It seems like many of the changes for the ICU4C Readme were previously done by @erik0686 and @daniel-ju during the work in https://github.com/unicode-org/icu/commit/ba3b6acd5c2599466c3b3fb460212c438f556768 (thanks!)

### DIFF
--- a/icu4c/readme.html
+++ b/icu4c/readme.html
@@ -47,7 +47,7 @@
     <p class="note only-rc">This is a release candidate version of ICU4C.
       It is not recommended for production use.</p>
 
-    <p>Last updated: 2020-Aug-27<br/>
+    <p>Last updated: 2020-Oct-01<br/>
       Copyright &copy; 2016 and later: Unicode, Inc. and others. License &amp; terms of use:
       <a href="http://www.unicode.org/copyright.html">http://www.unicode.org/copyright.html</a><br/>
       Copyright &copy; 1997-2016 International Business Machines Corporation and others.
--- a/icu4j/readme.html
+++ b/icu4j/readme.html
@@ -14,23 +14,23 @@ h3.doc { text-decoration: underline }
 <body style="background-color: rgb(255, 255, 255);" lang="EN-US"
  link="#0000ff" vlink="#800080">
 <h1>International Components for Unicode for Java (ICU4J)</h1>
-<h2>Read Me for ICU4J 67.1</h2>
-(Last Update: 2020-Apr-1)
+<h2>Read Me for ICU4J 68.1</h2>
+(Last Update: 2020-Oct-1)
 <hr size="2" width="100%">
 
 <p>
-<b>Note:</b> This is major release of ICU4J. It contains bug fixes and adds implementations
+<b>Note:</b> This is a major release of ICU4J. It contains bug fixes and adds implementations
 of inherited API and introduces new API or functionality.
-<!-- <b>Note:</b> This is a preview release of ICU4J 67.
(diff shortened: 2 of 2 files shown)


markusicu (approved):
lgtm tnx

jefgen (dismissed):
Looks good, thanks @echeran!

icu4c/readme.html
```diff
@@ -23,8 +23,8 @@
         milestone - if the release note is a milestone release
 -->
 
-  <body>
-  <!-- <body class="rc"> -->
+  <!-- <body> -->
+  <body class="rc">
```
daniel-ju, line 27: I don't have a strong opinion either way, but I think in the BRS for 67 there was mention that we would go straight to the "final" readme instead of marking it as RC (if we don't plan on making changes to the readmes between releases), so that we wouldn't need another update for GA.

daniel-ju (approved):

icu4c/readme.html
markusicu: You are right, we said that for 67: “just "67", not rc vs. final  updated instructions for this"”

However, I don't see the [instructions](https://sites.google.com/site/icusite/processes/release/tasks/publish#TOC-Update-readme) having been updated.

Leaving it up to @echeran whether he wants to merge as is and do a GA update later, or make another commit here to go straight to GA language.

daniel-ju: The instructions were updated with this: "If the readme should remain the same between milestones, we can skip directly to the GA description. Otherwise, pick the right one for the release type."

But I agree, either is fine. :)

echeran: Done.  If had a reasoning for what you did for the last BRS, then I may as well be consistent here, so I also changed it to the "final" readme

icu4j/readme.html
```diff
@@ -14,24 +14,24 @@
...
+<h2>Read Me for ICU4J 68.1</h2>
+(Last Update: 2020-Oct-1)
 <hr size="2" width="100%">
 
 <p>
-<b>Note:</b> This is major release of ICU4J. It contains bug fixes and adds implementations
-of inherited API and introduces new API or functionality.
-<!-- <b>Note:</b> This is a preview release of ICU4J 67.
+<!-- <b>Note:</b> This is major release of ICU4J. It contains bug fixes and adds implementations
+of inherited API and introduces new API or functionality. -->
```
daniel-ju, line 23: If we make the icu4c readme the "final" version, then we should do the same for the icu4j readme here to be consistent.

echeran: Done.  Thanks for catching that.

daniel-ju (approved):
lgtm, thanks! :)

markusicu (approved):
lgtm tnx

icu4c/readme.html
markusicu: > The instructions were updated with this: "If the readme should remain the same between milestones, we can skip directly to the GA description. Otherwise, pick the right one for the release type."

I updated the instructions now. http://site.icu-project.org/processes/release/tasks/publish#TOC-Update-readme
Hopefully this is clearer going forward.

PR #1394, opened 2020-10-01
ICU-21249 add unumberrangeformatter.h to API docs main page

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21249


sffc (approved):

jefgen (commented):
post-merge lgtm, thanks for fixing this!

PR #1398, opened 2020-10-07
ICU-20677 Fixing cygwin/MSVC build

Artemkth:
Dear Maintainers

I have fixed multiple failures preventing normal building with cygwin/MSVC. Could you consider also merging those into next release alpha?

- Fixed genrb (and all other utils) not recognizing alternative file separators on windows, which previously prevented generation of data in out-of-source build. Officially distributed binaries have the same issue. f402ff9; Now checks for last occurrence of both the separator and alternative separator are run before returning the one which is furthest into the string.
- Added few more fixes to the makefile of a new component  (extra/scrptrun), it had a redundant inculde of platform setup duplicating suffices in names of libraries (i.e. requiring compiler to find 'libicuucd**d**.lib' along with some other possible ill side-effects) 065f559;
- Changed earlier fix to depend on already built-in directory creation script, instead of doing it one more time. 8dc2c9c

Both debug and release were compiled on windows and 'check-exhaustive' was run successfully. Build on linux is tested from debian/testing running on WSL in win10. Both logs are appended.

Kind regards,
Artem

[icu_build.log](https://github.com/unicode-org/icu/files/5342653/icu_build.log)
[icu-build-linux.log](https://github.com/unicode-org/icu/files/5342656/icu-build-linux.log)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20677
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/tools/toolutil/toolutil.cpp
+++ b/icu4c/source/tools/toolutil/toolutil.cpp
@@ -166,14 +166,11 @@ findBasename(const char *filename) {
     const char *basename=uprv_strrchr(filename, U_FILE_SEP_CHAR);
 
 #if U_FILE_ALT_SEP_CHAR!=U_FILE_SEP_CHAR
-#if !(U_PLATFORM == U_PF_CYGWIN && U_PLATFORM_USES_ONLY_WIN32_API)
-    if(basename==NULL)
-#endif
-    {
-        /* Use lenient matching on Windows, which can accept either \ or /
-           This is useful for environments like Win32+CygWin which have both.
-        */
-        basename=uprv_strrchr(filename, U_FILE_ALT_SEP_CHAR);
+    //be lenient about pathname separators on Windows, like official implementation of C++17 std::filesystem in MSVC
+    //would be convenient to merge this loop with the one above, but alas, there is no such solution in the standard library
+    const char *alt_basename=uprv_strrchr(filename, U_FILE_ALT_SEP_CHAR);
+    if(alt_basename>basename) {
+        basename=alt_basename;
     }
 #endif
 
--- a/icu4c/source/data/Makefile.in
+++ b/icu4c/source/data/Makefile.in
@@ -240,8 +240,7 @@ include $(top_builddir)/$(subdir)/rules.mk
 ifeq ($(ENABLE_SO_VERSION_DATA),1)
 ifeq ($(PKGDATA_MODE),dll)
 SO_VERSION_DATA = $(OUTTMPDIR)/icudata.res
-$(SO_VERSION_DATA) : $(MISCSRCDIR)/icudata.rc
-	mkdir -p $(@D)
+$(SO_VERSION_DATA) : $(MISCSRCDIR)/icudata.rc | $(TMP_DIR)/dirs.timestamp
 ifeq ($(MSYS_RC_MODE),1)
 	rc.exe -i$(srcdir)/../common -i$(top_builddir)/common -fo$@ $(CPPFLAGS) $<
 else
--- a/icu4c/source/extra/scrptrun/Makefile.in
+++ b/icu4c/source/extra/scrptrun/Makefile.in
@@ -12,9 +12,6 @@ top_builddir = ../..
 
 include $(top_builddir)/icudefs.mk
 
-## Platform-specific setup
-include @platform_make_fragment@
-
 ## Build directory information
 subdir = extra/scrptrun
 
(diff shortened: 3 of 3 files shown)


jefgen:
Thank you very kindly for creating this PR and also for signing the CLA! :)

icu4c/source/tools/toolutil/toolutil.cpp
```diff
@@ -166,15 +166,11 @@ findBasename(const char *filename) {
...
-    if(basename==NULL)
-#endif
-    {
-        /* Use lenient matching on Windows, which can accept either \ or /
-           This is useful for environments like Win32+CygWin which have both.
-        */
-        basename=uprv_strrchr(filename, U_FILE_ALT_SEP_CHAR);
-    }
+    //be lenient about pathname separators on Windows, like official implementation of C++17 std::filesystem in MSVC
+    //would be convinient to merge this loop with the one above, but alas, there is no such solution in the standard library
```
jefgen, line 170: nit: typo: convinient --> convenient

jefgen, line 170: When you say "loop" in the comment, do you mean the call to `uprv_strrchr`?

icu4c/source/tools/toolutil/toolutil.cpp
```diff
@@ -166,15 +166,11 @@ findBasename(const char *filename) {
...
+    const char *alt_basename=uprv_strrchr(filename, U_FILE_ALT_SEP_CHAR);
+    if(alt_basename>basename)
```
jefgen, line 172:
```suggestion
    if(alt_basename>basename) {
        basename=alt_basename;
     }
```

The code in this file doesn't really follow the ICU style guides... but would you mind using the parenthesis for the if statement block though? (The ICU style guide says to always enclose the block anyways.)

jefgen (changes requested):
LGTM, just a minor typo to fix.

Thank you!

icu4c/source/tools/toolutil/toolutil.cpp
Artemkth: Internals of both calls to `uprv_strrchr`, they loop over whole string twice when it can be done in a single pass. I checked and there was only a variant returning first match [`strbrk`](http://www.cplusplus.com/reference/cstring/strpbrk/).

jefgen:
Note for other ICU-TC members: The code freeze for ICU 68.1 was technically supposed to be ~1 week ago or so -- but since this fix is important to unblock building I'd vote to consider putting this into the 68 RC.

icu4c/source/tools/toolutil/toolutil.cpp
Artemkth: How do you suggest changing the suggestion? I am confused.
You forgot to include the line below and now it has a duplicate of `basename=alt_basename;`

icu4c/source/tools/toolutil/toolutil.cpp
```diff
@@ -166,15 +166,11 @@ findBasename(const char *filename) {
...
+        basename=alt_basename;
```
Artemkth, line 173:
```suggestion
    if(alt_basename>basename) {
        basename=alt_basename;
    }
```

icu4c/source/tools/toolutil/toolutil.cpp
jefgen: Sorry for the confusion... I think the GitHub suggestion got messed up as I started the comment by only commenting on one line.
Your suggested change here https://github.com/unicode-org/icu/pull/1398#discussion_r501288905 looks good. :)

icu4c/source/tools/toolutil/toolutil.cpp
Artemkth, line 170:
```suggestion
    //would be convenient to merge this loop with the one above, but alas, there is no such solution in the standard library
```

icu4c/source/tools/toolutil/toolutil.cpp
Artemkth: I squashed your suggestions into my initial commit^

jefgen:
Thanks for making the edits and squashing them!

It looks like the AppVeyor build failed with a segfault. However, I think this is just flakiness. (The AppVeyor builds have historically been rather flaky with transient failures). 

```
make[1]: *** [../config/mh-cygwin64:76: region.o] Segmentation fault
make[1]: Leaving directory '/cygdrive/c/projects/icu/icu4c/source/i18n'
make: *** [Makefile:153: all-recursive] Error 2
Command exited with code 2
```

I'll re-queue another build, as I think it should likely pass next time.

jefgen (approved):
LGTM, thank you very much!

PR #1374, opened 2020-09-29
ICU-21249 ICU 68 BRS - Clean up import statements

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21249
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #1402, opened 2020-10-09
ICU-21308 Fix units parser thread safety issue

younies:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21308
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
@@ -4,7 +4,6 @@ package com.ibm.icu.impl.units;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Comparator;
 
 import com.ibm.icu.util.BytesTrie;
@@ -35,7 +34,7 @@ public class MeasureUnitImpl {
      * <p>
      * The "dimensionless" <code>MeasureUnitImpl</code> has an empty <code>singleUnits</code>.
      */
-    private ArrayList<SingleUnitImpl> singleUnits;
+    private final ArrayList<SingleUnitImpl> singleUnits;
 
     public MeasureUnitImpl() {
         singleUnits = new ArrayList<>();
@@ -147,8 +146,7 @@ public class MeasureUnitImpl {
 
         // Find a similar unit that already exists, to attempt to coalesce
         SingleUnitImpl oldUnit = null;
-        for (int i = 0, n = this.singleUnits.size(); i < n; i++) {
-            SingleUnitImpl candidate = this.singleUnits.get(i);
+        for (SingleUnitImpl candidate : this.singleUnits) {
             if (candidate.isCompatibleWith(singleUnit)) {
                 oldUnit = candidate;
                 break;
@@ -226,7 +224,7 @@ public class MeasureUnitImpl {
         }
         if (this.complexity == MeasureUnit.Complexity.COMPOUND) {
             // Note: don't sort a MIXED unit
-            Collections.sort(this.getSingleUnits(), new SingleUnitComparator());
+            this.getSingleUnits().sort(new SingleUnitComparator());
         }
 
         StringBuilder result = new StringBuilder();
@@ -372,34 +370,32 @@ public class MeasureUnitImpl {
     public static class UnitsParser {
         // This used only to not build the trie each time we use the parser
         private volatile static CharsTrie savedTrie = null;
-        private final String[] simpleUnits;
+
         // This trie used in the parsing operation.
-        private CharsTrie trie;
+        private final CharsTrie trie;
+        private final String fSource;
         // Tracks parser progress: the offset into fSource.
         private int fIndex = 0;
         // Set to true when we've seen a "-per-" or a "per-", after which all units
         // are in the denominator. Until we find an "-and-", at which point the
         // identifier is invalid pending TODO(CLDR-13700).
         private boolean fAfterPer = false;
-        private String fSource;
         // If an "-and-" was parsed prior to finding the "single
         //     * unit", sawAnd is set to true. If not, it is left as is.
         private boolean fSawAnd = false;
 
         private UnitsParser(String identifier) {
-            this.simpleUnits = UnitsData.getSimpleUnits();
             this.fSource = identifier;
 
-            if (UnitsParser.savedTrie != null) {
-                try {
-                    this.trie = UnitsParser.savedTrie.clone();
-                } catch (CloneNotSupportedException e) {
-                    throw new ICUCloneNotSupportedException();
-                }
-                return;
+            try {
+                this.trie = UnitsParser.savedTrie.clone();
+            } catch (CloneNotSupportedException e) {
+                throw new ICUCloneNotSupportedException();
             }
+        }
 
-            // Building the trie.
+        static {
+            // Build Units trie.
             CharsTrieBuilder trieBuilder;
             trieBuilder = new CharsTrieBuilder();
 
@@ -432,6 +428,7 @@ public class MeasureUnitImpl {
             }
 
             // Add simple units
+            String[] simpleUnits = UnitsData.getSimpleUnits();
             for (int i = 0; i < simpleUnits.length; i++) {
                 trieBuilder.add(simpleUnits[i], i + UnitsData.Constants.kSimpleUnitOffset);
 
@@ -439,14 +436,7 @@ public class MeasureUnitImpl {
 
             // TODO: Use SLOW or FAST here?
             UnitsParser.savedTrie = trieBuilder.build(StringTrieBuilder.Option.FAST);
-
-            try {
-                this.trie = UnitsParser.savedTrie.clone();
-            } catch (CloneNotSupportedException e) {
-                throw new ICUCloneNotSupportedException();
-            }
         }
-
 
         /**
          * Construct a MeasureUnit from a CLDR Unit Identifier, defined in UTS 35.
@@ -607,7 +597,7 @@ public class MeasureUnitImpl {
                         break;
 
                     case TYPE_SIMPLE_UNIT:
-                        result.setSimpleUnit(token.getSimpleUnitIndex(), simpleUnits);
+                        result.setSimpleUnit(token.getSimpleUnitIndex(), UnitsData.getSimpleUnits());
                         return result;
 
                     default:
@@ -746,7 +736,7 @@ public class MeasureUnitImpl {
     }
 
     static class MeasureUnitImplComparator implements Comparator<MeasureUnitImpl> {
-        private ConversionRates conversionRates;
+        private final ConversionRates conversionRates;
 
         public MeasureUnitImplComparator(ConversionRates conversionRates) {
             this.conversionRates = conversionRates;


icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -266,6 +262,11 @@ public void serialize() {
         this.identifier = result.toString();
     }
 
+    @Override
+    public String toString() {
+        return "MeasureUnitImpl [" + build().getIdentifier() + "]";
+    }
+
```
hugovdm, line 269: Optional: because this is an urgent bug fix, I would not do any formatting changes.

More verbosely - my suggestions for future PRs (if I wrote this Pull Request, I would do this):

- have one commit that does reformatting only, an one commit that does the logic changes. The reviewer then does not have to figure out what moved code might have changed, and what didn't.
- I would perhaps even do one PR that just does reformatting (no logic changes), perhaps on all related files, then follow that with a separate PR with logic changes. But this makes sense only if the second PR isn't urgent.
- For this urgent PR, after ICU has already been tagged with a release candidate, I would do only logic changes, no formatting changes. This is an "urgent burden" on the reviewers, when we want to make minimal bugfix changes to the Release Candidate.

I can't review formatting changes yet, since I'm not yet familiar with ICU4J style (and haven't started using Eclipse).

hugovdm: Actually, prior to this PR, I had touched only toString and three more lines. Yoshito touched the import statements in b2edfb89ccb534ff882954328c0d65d6721bd688 - fixing them according to Eclipse-configured style? This PR is *undoing* those changes again, so clearly IntelliJ is *not* following Eclipse style.

@younies please undo the formatting changes that aren't related to the logic changes. Having warring IDEs will be bad for every PR (next time Eclipse formats this, these changes get undone again).

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -226,7 +222,7 @@ public void serialize() {
         }
         if (this.complexity == MeasureUnit.Complexity.COMPOUND) {
             // Note: don't sort a MIXED unit
-            Collections.sort(this.getSingleUnits(), new SingleUnitComparator());
+            this.getSingleUnits().sort(new SingleUnitComparator());
```
hugovdm, line 225: I'm getting this error, which may be a Java 7 vs Java 8 thing? :

The method sort(MeasureUnitImpl.SingleUnitComparator) is undefined for the type ArrayList<SingleUnitImpl>	MeasureUnitImpl.java	/icu4j-core/src/com/ibm/icu/impl/units	line 227	Java Problem

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: For easier review, this diff is the same code as commit 6a15cad, but shuffled around to make it minimal:
https://github.com/unicode-org/icu/compare/master...icu-units:minimal_diff?expand=1

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -607,7 +600,7 @@ private SingleUnitImpl nextSingleUnit() {
                         break;
 
                     case TYPE_SIMPLE_UNIT:
-                        result.setSimpleUnit(token.getSimpleUnitIndex(), simpleUnits);
+                        result.setSimpleUnit(token.getSimpleUnitIndex(), UnitsData.getSimpleUnits());
```
hugovdm, line 603: getSimpleUnits() looks slow, in which case this will make nextSingleUnit() and UnitsParser.parse() slow.

How about restoring `String[] simpleUnits` as static, and keeping the results in the class?

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -2,29 +2,18 @@
...
 
-public class MeasureUnitImpl {
-
-    /**
-     * The full unit identifier.  Null if not computed.
-     */
-    private String identifier = null;
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.Comparator;
```
hugovdm, line 14: Move these imports above the "import com.*" imports. That's how Eclipse is configured to save them.

(For the future: figure out how to make IntelliJ get this right, to not end up in an edit war with Eclipse.)

hugovdm: Yoshito's comment in https://github.com/unicode-org/icu/pull/1374 :

```
@younies It looks files touched by you uses different import statement ordering. We set up our preferences in Eclipse project files. Could you please honor the setting? Easiest way to handle this is to import ICU projects to Eclipse, and let Eclipse to do the work.
```

I've seen I can change the import order in clang-format too.

hugovdm: Looking into Eclipse, the desired order of imports is: ['java', 'javax', 'org', 'com'].

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: (I recommend simply not changing this line - it seems fine as it was, and Java 7 compatible?)

hugovdm:
(Pushed my "Minimal Diff Version of ICU PR 1402" commit to this branch, following a conversation with Younies.)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: (FYI: pushed to this branch/PR.)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: My commit changed the import order back, so this is done.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: I reopened this in Eclipse to take another look, and this error is no longer showing up. I don't know enough about Eclipse and Java to speculate what might have been different when I saw this error before.

I'll mark as resolved.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
younies: why it is slow ? 
it will return just a reference to simpleUnits ??

younies: simpleUnits will be stored in UnitsData, it is not like the other data such as conversion rates.

hugovdm: :facepalm: - looks like I wasn't reading thoroughly? I see this now, which I must have skipped last week, sorry:

        if (simpleUnits != null) {
            return simpleUnits;
        }

hugovdm: Marking as resolved.

hugovdm:
Let me get bolder. @younies looks good to me, please squash!

hugovdm (approved):

hugovdm:
I've done a force-push simply to rewrite the author from hugovdm to younies.
- It seems Squash Bot uses the author of the last commit, not the first.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -372,34 +370,32 @@ public int getValue() {
...
-                return;
+            try {
+                this.trie = UnitsParser.savedTrie.clone();
+            } catch (CloneNotSupportedException e) {
+                throw new ICUCloneNotSupportedException();
             }
+        }
 
-            // Building the trie.
+        static {
```
sffc, line 397: I think `static {}` blocks run when the class gets loaded.  Is that more often than we want?  If someone isn't doing unit formatting, we don't want to have them wait for this trie to be built and take up their memory.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -432,22 +428,16 @@ private UnitsParser(String identifier) {
...
 
             // Add simple units
+            String[] simpleUnits = UnitsData.getSimpleUnits();
             for (int i = 0; i < simpleUnits.length; i++) {
                 trieBuilder.add(simpleUnits[i], i + UnitsData.Constants.kSimpleUnitOffset);
 
             }
 
             // TODO: Use SLOW or FAST here?
             UnitsParser.savedTrie = trieBuilder.build(StringTrieBuilder.Option.FAST);
```
sffc, line 438: Basically what you want is a compare-and-swap.  I think you can do this safely with the old code but with an extra synchronized block:

```java
synchronized {
    if (UnitsParser.savedTrie == null) {
        UnitsParser.savedTrie = trieBuilder.build(...);
    }
}
```

See https://stackoverflow.com/questions/7420504/threading-lazy-initialization-vs-static-lazy-initialization

hugovdm (approved):
From what I read, it looks like static initializer blocks are run when the first instance of a class is created, or a static member of the class is first accessed. This does sound like the correct behaviour?

hugovdm:
Understood from the ICU-TC meeting today: changes intended for the v68 release should go straight into the maintenance branch.

sffc (changes requested):
Setting review bit

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: s/unit formatting/unit identifier parsing/ ?

I think this initialization code is needed whenever this class (UnitsParser) is used?

So the question is just whether nested classes initialize in the way I think they do? Instantiating MeasureUnitImpl but not parsing identifiers (not calling MeasureUnitImpl.forIdentifier) should result in this not being run? We could test with a quick "system.out.println" in a unit test that does this?

younies: I agree with @hugovdm

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
younies: I think static block works fine too.

sffc (approved):

hugovdm (commented):
Thanks for fixing! / LGTM

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: Turns out the error was indeed a problem, and this line needs to be changed back.

PR #1404, opened 2020-10-12
ICU-21250 BRS68GA Update version numbers for 68GA and regenerate jar files

erik0686:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21250
- [X] Issue accepted
- [ ] Tests included


jefgen:
FYI: I think the PR should be going into the `maint/maint-68` branch, rather than `master`.

daniel-ju:
It seems like the full JAR files are being added, instead of using Git LFS. Do you have Git LFS configured?

daniel-ju (approved):

jefgen (approved):
Thanks!

yumaoka:
You don't need to regenerate ICU4J data files with the change, but it's not harmful.

We somewhat agreed that deprecating the process for putting RC/Release Candidate to API doc, etc. I think I did not document this  in BRS description.

PR #1408, opened 2020-10-15
ICU-21336 Fix format string to use long instead of int in common/wintz.cpp

jefgen:
This is related to PR #1362. 

Thanks to @FrankYFTang for catching this issue, and sorry for the build break.

The variable `utcOffsetMins` is type `long` (via a typedef for `LONG`), so the calculation of `-utcOffsetMins/60` is also a `long` as well. This leads to an error as the format string uses `d` which is only for `int`.

(Aside: A possible alternate fix would be to use `static_cast<int>(-utcOffsetMins/60)` in order to make it an `int` as the value should never require the full range of `long`.)

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21336
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/wintz.cpp
+++ b/icu4c/source/common/wintz.cpp
@@ -86,7 +86,7 @@ uprv_detectWindowsTimeZone()
             // Note '-' before 'utcOffsetMin'. The timezone ID's sign convention
             // is that a timezone ahead of UTC is Etc/GMT-<offset> and a timezone
             // behind UTC is Etc/GMT+<offset>.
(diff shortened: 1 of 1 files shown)


jefgen:
Interestingly I don't see any warnings generated at all for this when building with MSVC, for both x64 and x86 builds. I think this warning might only be generated when using the clang compiler to build -- which would explain why the warnings-as-errors build bots for ICU didn't catch this.

FrankYFTang (approved):

daniel-ju (approved):

PR #1409, opened 2020-10-15
ICU-20941 Support formatting joule-per-furlong (builtin-per-builtin).

hugovdm:
Small step in the direction of ICU-20941 - contributes to icu-units#27.

With this change:
* LongNameHandler constructors no longer take "unit" and "perUnit" - they take only one parameter.
  * Combining of "perUnit" into "unit" thus moved outside of the class, and
  * splitting of "unit" into numerator and denominator moved into the class.
* Skeleton handling is simpler, because we no longer care about "perUnit".

Joule per Furlong was already supported when specified separately. It's just when given as a single unit that it didn't automatically get split.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20941
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/number_formatimpl.cpp
+++ b/icu4c/source/i18n/number_formatimpl.cpp
@@ -389,8 +389,12 @@ NumberFormatterImpl::macrosToMicroGenerator(const MacroProps& macros, bool safe,
                 fMixedUnitLongNameHandler.getAlias(), status);
             chain = fMixedUnitLongNameHandler.getAlias();
         } else {
+            MeasureUnit unit = macros.unit;
+            if (!utils::unitIsBaseUnit(macros.perUnit)) {
+                unit = unit.product(macros.perUnit.reciprocal(status), status);
+            }
             fLongNameHandler.adoptInsteadAndCheckErrorCode(new LongNameHandler(), status);
-            LongNameHandler::forMeasureUnit(macros.locale, macros.unit, macros.perUnit, unitWidth,
+            LongNameHandler::forMeasureUnit(macros.locale, unit, unitWidth,
                                             resolvePluralRules(macros.rules, macros.locale, status),
                                             chain, fLongNameHandler.getAlias(), status);
             chain = fLongNameHandler.getAlias();
--- a/icu4c/source/i18n/number_longnames.cpp
+++ b/icu4c/source/i18n/number_longnames.cpp
@@ -218,38 +218,38 @@ UnicodeString getPerUnitFormat(const Locale& locale, const UNumberUnitWidth &wid
 } // namespace
 
 void LongNameHandler::forMeasureUnit(const Locale &loc, const MeasureUnit &unitRef,
-                                     const MeasureUnit &perUnit, const UNumberUnitWidth &width,
-                                     const PluralRules *rules, const MicroPropsGenerator *parent,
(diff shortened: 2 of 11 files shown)


macchiati:
Haven't looked at the code yet, but just wanted to make sure: the perUnit pattern should be used if it is available.

icu4c/source/i18n/number_formatimpl.cpp
```diff
@@ -389,8 +389,12 @@ NumberFormatterImpl::macrosToMicroGenerator(const MacroProps& macros, bool safe,
                 fMixedUnitLongNameHandler.getAlias(), status);
             chain = fMixedUnitLongNameHandler.getAlias();
         } else {
+            MeasureUnit unit = macros.unit;
+            if (uprv_strcmp(macros.perUnit.getType(), "none") != 0) {
```
sffc, line 393: Prefer `utils::unitIsBaseUnit()` rather than a string comparison to `"none"`.  (I see you had copied this from pre-existing code)

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -218,38 +218,37 @@ UnicodeString getPerUnitFormat(const Locale& locale, const UNumberUnitWidth &wid
...
-        } else {
-            // No simplified form is available.
-            forCompoundUnit(loc, unit, perUnit, width, rules, parent, fillIn, status);
-            return;
+    if (uprv_strcmp(unitRef.getType(), "") == 0) {
+        // Not a built-in unit. Split it up, since we can already format
+        // "builtin-per-builtin".
+        // TODO(ICU-20941): support more generic case than builtin-per-builtin.
+        MeasureUnitImpl fullUnit = MeasureUnitImpl::forMeasureUnitMaybeCopy(unitRef, status);
+        MeasureUnit unit;
```
sffc, line 235: As usual, please use `MeasureUnitImpl` style code in the implementation.  `MeasureUnitImpl` has an `append` function that you can use here, and then you call `.build()` at the end to turn it into the MeasureUnit.

Alternatively, add a function directly on MeasureUnitImpl named something like "splitPerUnit" which modifies self, removes the denominator, and returns the denominator as a new MeasureUnitImpl.

sffc (commented):
Nice cleanup!  Some suggestions.

icu4c/source/i18n/number_longnames.cpp
hugovdm: Done - not going with a new class member function for now, might revisit the idea as I add full support for arbitrary unit formatting.

icu4c/source/i18n/number_formatimpl.cpp
hugovdm: Done!

hugovdm:
> Haven't looked at the code yet, but just wanted to make sure: the perUnit pattern should be used if it is available.

With this PR, the perUnit pattern is picked up by line 290 in number_longnames.cpp - i.e. it looks correct to me. Or in other terms: "LongNameHandler::forCompoundUnit()" is already good and correct, this refactoring just provides more entry points into that factory method (rather than returning an error).

sffc (approved):
LGTM please squash

sffc (approved):

PR #1412, opened 2020-10-16
ICU-21340 Add tests verify getNextPosition()

FrankYFTang:
getNextPosition() for n items should have element exactly n times.

This is just intended for the one who is going to fix (Shane?) to see the expectation of test we need in v8.


sffc:
This test is included in #1425.

PR #1413, opened 2020-10-16
ICU-21339 Add regression test

FrankYFTang:
Just a PR to add tests to show the problem in ICU-21339 for Hugo to debug


hugovdm:
#1414 merged, containing this test. Shall we close it?

FrankYFTang:
obsoleted.

PR #1414, opened 2020-10-16
ICU-21339 toSkeleton: treat percent and permille correctly

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21339
- [x] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/number_skeletons.cpp
+++ b/icu4c/source/i18n/number_skeletons.cpp
@@ -732,6 +732,7 @@ skeleton::parseStem(const StringSegment& segment, const UCharsTrie& stemTrie, Se
 
         case STEM_CURRENCY:
             CHECK_NULL(seen, unit, status);
+            CHECK_NULL(seen, perUnit, status);
             return STATE_CURRENCY_UNIT;
 
         case STEM_INTEGER_WIDTH:
@@ -1500,32 +1501,33 @@ bool GeneratorHelpers::notation(const MacroProps& macros, UnicodeString& sb, UEr
 }
 
 bool GeneratorHelpers::unit(const MacroProps& macros, UnicodeString& sb, UErrorCode& status) {
-    if (utils::unitIsCurrency(macros.unit)) {
+    MeasureUnit unit = macros.unit;
+    if (!utils::unitIsBaseUnit(macros.perUnit)) {
+        if (utils::unitIsCurrency(macros.unit) || utils::unitIsCurrency(macros.perUnit)) {
+            status = U_UNSUPPORTED_ERROR;
+            return false;
+        }
+        unit = unit.product(macros.perUnit.reciprocal(status), status);
+    }
+
+    if (utils::unitIsCurrency(unit)) {
         sb.append(u"currency/", -1);
-        CurrencyUnit currency(macros.unit, status);
+        CurrencyUnit currency(unit, status);
         if (U_FAILURE(status)) {
             return false;
         }
         blueprint_helpers::generateCurrencyOption(currency, sb, status);
         return true;
-    } else if (utils::unitIsBaseUnit(macros.unit)) {
+    } else if (utils::unitIsBaseUnit(unit)) {
         // Default value is not shown in normalized form
         return false;
-    } else if (utils::unitIsPercent(macros.unit)) {
+    } else if (utils::unitIsPercent(unit)) {
         sb.append(u"percent", -1);
         return true;
-    } else if (utils::unitIsPermille(macros.unit)) {
+    } else if (utils::unitIsPermille(unit)) {
         sb.append(u"permille", -1);
         return true;
     } else {
-        MeasureUnit unit = macros.unit;
-        if (utils::unitIsCurrency(macros.perUnit)) {
-            status = U_UNSUPPORTED_ERROR;
-            return false;
-        }
-        if (!utils::unitIsBaseUnit(macros.perUnit)) {
-            unit = unit.product(macros.perUnit.reciprocal(status), status);
-        }
         sb.append(u"unit/", -1);
         sb.append(unit.getIdentifier());
         return true;
--- a/icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
@@ -620,6 +620,7 @@ class NumberSkeletonImpl {
                 case STATE_INCREMENT_PRECISION:
                 case STATE_MEASURE_UNIT:
                 case STATE_PER_MEASURE_UNIT:
+                case STATE_IDENTIFIER_UNIT:
                 case STATE_UNIT_USAGE:
                 case STATE_CURRENCY_UNIT:
                 case STATE_INTEGER_WIDTH:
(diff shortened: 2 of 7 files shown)


sffc (dismissed):

icu4c/source/i18n/number_skeletons.cpp
```diff
@@ -1500,32 +1500,33 @@ bool GeneratorHelpers::notation(const MacroProps& macros, UnicodeString& sb, UEr
 }
 
 bool GeneratorHelpers::unit(const MacroProps& macros, UnicodeString& sb, UErrorCode& status) {
-    if (utils::unitIsCurrency(macros.unit)) {
+    MeasureUnit unit = macros.unit;
+    if (!utils::unitIsBaseUnit(macros.perUnit)) {
+        if (utils::unitIsCurrency(macros.unit) || utils::unitIsCurrency(macros.perUnit)) {
```
FrankYFTang, line 1505: Isn't the !utils::unitIsBaseUnit(macros.perUnit) && utils::unitIsCurrency(macros.unit)) => U_UNSUPPORTED_ERROR / false
a new behavior?

FrankYFTang (commented):
Since you are changing the code around currency, could you add unit test to test the condition
1. check the currency -per- meter
2. check meter -per- currency
Also, see does the expectation changed if you run that on 67. IF so, is the change reasonable?

icu4c/source/i18n/number_skeletons.cpp
hugovdm: You're right that I've hereby changed the toSkeleton output. I'm not sure what the best thing to do is - is it just a matter of needing to document the behaviour change?

When constructing a NumberFormatter, what happens (in number_formatimpl.cpp, macrosToMicroGenerator FWIW) is if unit is set to currency, perUnit was simply ignored in 67, and is still ignored in 68. Is it better to preserve this behaviour, simply ignoring perUnit?

hugovdm: I'll add the suggested unit tests. Thoughts on the question above ("is the change reasonable?"):

- It is indeed a behaviour change. I do think the change is reasonable, though arguably we should give an error already at the formatter construction, not only at the toSkeleton() call... and I may have given insufficient thought to what we might end up breaking with this? Do you think it's risky?

- For the remaining part of ICU-20941 for ICU 69, we *could* consider adding support for $CURRENCY-per-meter, but this doesn't fall into the unit identifier spec, so we may need to carefully consider the right way to deal with this case (e.g. also in the specs).

sffc: We already announced that skeleton behavior for units is changing in ICU 68.  The RC version of that behavior was buggy, as Frank found.  I think it is 110% in scope to fix behavior between RC and GA according to requests from clients like Frank.

sffc (dismissed):

icu4c/source/i18n/number_skeletons.cpp
hugovdm: Turns out forSkeleton doesn't support currency: blueprint_helpers::parseMeasureUnitOption returns an U_INTERNAL_PROGRAM_ERROR because there are more than 40 currencies - it supports only unit types with up to 40 subtypes.

The new toSkeleton behaviour can still be triggered by constructing a NumberFormatter with the fluent API.

sffc: "currency-per-meter" requires more work, not only in skeletons but also in the formatting step.

I like the currently proposed fix because it fixes "percent-per-meter", for which we have formatting data, without touching compound currency units, for which we do not have formatting data.

hugovdm: Changes made:
- for currencies, reject `perUnit` at `forSkeleton` time, just as `unit` is rejected.
- unit test demonstrating toSkeleton fails when fluent-API is used to construct a currency/unit conflict.

For ICU 69, we may need to revisit what is produced by the formatter: possibly an error, rather than ignoring perUnit, or possibly supporting it, after figuring out how the skeletons will be handled for currency-vs-unit.

icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
```diff
@@ -1484,25 +1484,25 @@ private static boolean notation(MacroProps macros, StringBuilder sb) {
...
+                }
+                unit = unit.product(macros.perUnit.reciprocal());
+            }
+            if (unit instanceof Currency) {
                 sb.append("currency/");
-                BlueprintHelpers.generateCurrencyOption((Currency) macros.unit, sb);
+                BlueprintHelpers.generateCurrencyOption((Currency)unit, sb);
                 return true;
-            } else if (macros.unit == MeasureUnit.PERCENT) {
+            } else if (unit == MeasureUnit.PERCENT) {
```
younies, line 1499: unit.equals(MeasureUnit.PERCENT)

icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
```diff
@@ -1484,25 +1484,25 @@ private static boolean notation(MacroProps macros, StringBuilder sb) {
...
                 sb.append("percent");
                 return true;
-            } else if (macros.unit == MeasureUnit.PERMILLE) {
+            } else if (unit == MeasureUnit.PERMILLE) {
```
younies, line 1502: unit.equals(MeasureUnit. PERMILLE)

younies (changes requested):

icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
hugovdm: Done. Thanks!

icu4j/main/classes/core/src/com/ibm/icu/number/NumberSkeletonImpl.java
hugovdm: Done.

icu4c/source/i18n/number_skeletons.cpp
```diff
@@ -732,6 +732,7 @@ skeleton::parseStem(const StringSegment& segment, const UCharsTrie& stemTrie, Se
 
         case STEM_CURRENCY:
             CHECK_NULL(seen, unit, status);
+            CHECK_NULL(seen, perUnit, status);
```
sffc, line 735: Can you make the equivalent change in Java?

sffc (changes requested):

icu4c/source/i18n/number_skeletons.cpp
FrankYFTang: I have no strong opinion about CURRENCY-per-other unit as well as 
1. you add a unit test for that
2. you understand the impact of the changes 
3. V8 will not hit that case and it will not impact the v8/chrome integration work.
I will leave it to you and shane to decide is the changed behavior is acceptable to 68 or not. But we need to have Unit test to test whatever behavior you and shane agree to.

icu4c/source/i18n/number_skeletons.cpp
hugovdm: Done. Required a slightly different implementation because CHECK_NULL in C++ sets the seen flag, whereas in Java checkNull only checks for references.

FrankYFTang (dismissed):
Thanks you for fixing this promptly

hugovdm:
Squashed. The commit that @sffc has not reviewed yet: b837dad (in case he wants to see).

younies (approved):

hugovdm:
I'm waiting a bit longer to see if @FrankYFTang wants to rubber-stamp the squash (as I have pre-squash approval), and/or whether @sffc wants to review the final changes. I'm otherwise considering merging when the ICU meeting starts.

sffc (approved):

PR #1418, opened 2020-10-19
ICU-21337 Remove U_I18N_API from enum.

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21337
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/units_converter.h
+++ b/icu4c/source/i18n/units_converter.h
@@ -94,7 +94,7 @@ struct U_I18N_API ConversionRate : public UMemory {
         : source(std::move(source)), target(std::move(target)) {}
 };
 
-enum U_I18N_API Convertibility {
+enum Convertibility {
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/units_converter.h
```diff
@@ -94,7 +94,7 @@ struct U_I18N_API ConversionRate : public UMemory {
         : source(std::move(source)), target(std::move(target)) {}
 };
 
-enum U_I18N_API Convertibility {
+enum Convertibility {
```
younies, line 97: I hope this will not break the building on MacOSX

younies: it looks building on MacOSX has passed, it should be okay

FrankYFTang (approved):

PR #1422, opened 2020-10-21
ICU-21344 Fix wrong passing of script in Locale.

FrankYFTang:
Also add same unit test in Java which have no problem.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21344
- [X] Updated PR title and link in previous line to include Issue number
- [X] Issue accepted
- [X] Tests included
- [X] Documentation is changed or added

PR #1424, opened 2020-10-22
ICU-21342 Made sure adjustFieldWidth() got called on the duplicated h…

richgillam:
…our pattern used for UCAL_AM_PM in 24-hour patterns.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21342
- [x] Issue accepted
- [x] Tests included

If the skeleton specifies a 24-hour format, the UCAL_HOUR pattern is copied into the UCAL_AM_PM slot as well, but we weren't calling adjustFieldWidth() on it to make sure the appropriate hour metacharacters were substituted into the pattern.  In this case, that meant that the `HH` coming from the resource bundles wasn't correctly being converted into the `kk` requested by the caller.


icu4c/source/i18n/dtitvfmt.cpp
```diff
@@ -1422,7 +1422,11 @@ DateIntervalFormat::setIntervalPattern(UCalendarDateFields field,
         if ( field == UCAL_AM_PM ) {
             fInfo->getIntervalPattern(*bestSkeleton, UCAL_HOUR, pattern,status);
             if ( !pattern.isEmpty() ) {
-                setIntervalPattern(field, pattern);
+                UBool suppressDayPeriodField = fSkeleton.indexOf(CAP_J) != -1;
+                UnicodeString adjustIntervalPattern;
+                adjustFieldWidth(*skeleton, *bestSkeleton, pattern, differenceInfo,
+                                 suppressDayPeriodField, adjustIntervalPattern);
+                setIntervalPattern(field, adjustIntervalPattern);
```
pedberg-icu, line 1429: I see this is all similar to what is done later in the same function after `if ( !pattern.isEmpty() )` ...

pedberg-icu (approved):
Looks good, thanks

icu4c/source/i18n/dtitvfmt.cpp
richgillam: Yup, it's the same logic.  Since it's only a couple lines, I didn't think it was worth it to factor it out into a separate function.

pedberg-icu: Agree, I was just noting that as a way of verifying correctness

pedberg-icu:
Note, the travis-ci build failure in depstest is caused by a different PR: https://github.com/unicode-org/icu/pull/1422#issuecomment-714578382

pedberg-icu (approved):

FrankYFTang (approved):
LGTM

PR #1430, opened 2020-10-23
ICU-21250 integrate CLDR release-38-final0 to ICU trunk

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21250
- [x] Issue accepted
- [x] Tests included

Integrate CLDR release data tagged "release-38-final0" into ICU maint/maint-68 branch. While this is not the final CLDR 38 release, I am expecting it to be the last one that has any data changes that affect ICU. There is only one, reverting a localized name change in zh_Hant for metazone Indochina.


markusicu (approved):
rslgtm

PR #1432, opened 2020-10-23
ICU-21352 Windows time zone detection can fail in RDP sessions with 3rd party software and older RDP clients

jefgen:
#### Description
The time zone detection logic in ICU uses the Win32 API `GetDynamicTimeZoneInformation` (which is available since Vista) to get the current time zone info, as this API returns a _non-localized_ time zone name which can be then mapped to an ICU time zone.

However, in some RDP/terminal services situations, this struct isn't always fully complete, and the `TimeZoneKeyName` field of the struct might be `NULL`.

This can happen in some RDP/terminal services sessions where the 'Time Zone Redirection' feature is enabled, as some 3rd party RDP clients (and also older versions of the RDP protocol), don't send the newer `TimeZoneKeyName` information, and only send the `StandardName` and `DaylightName`.

Since these 3rd party clients and older RDP clients only send the pre-Vista time zone information to the server, this means that we need to fallback on using the pre-Vista methods to determine the time zone. This unfortunately requires examining the registry directly in order to try and determine the current time zone.

Note that this can however still fail in some cases though if the client and server are using different languages, as the `StandardName` that is sent by client is _localized_ in the client's language. However, we must compare this to the names that are on the server, which are localized in registry using the server's language. Despite that, this is the best we can do as fallback for this RDP scenario.

One thing to note is that this fallback method won't work for the UWP version however, as we can't use the registry APIs in UWP.

### Testing
I've tested this locally using a 3rd party RDP client which I've confirmed does *not* send the TimeZoneKeyName information to the server. I setup two VMs (a client and a server setup) with the 'Time Zone Redirection' feature enabled, and set different time zones on each system. This setup causes the struct returned by `GetDynamicTimeZoneInformation` to be incomplete, which causes the time zone to detection to fail.

Before the timezone that was reported by the `icuinfo.exe` program when run inside the RDP session was the (incorrect) string "`Pacific Standard Time`".

With this change, the `icuinfo.exe` program now reports the string "`America/Vancouver`", which is the expected/correct result when the user's region is set to "CA".

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21352
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/wintz.cpp
+++ b/icu4c/source/common/wintz.cpp
@@ -36,17 +36,58 @@
 
 U_NAMESPACE_BEGIN
 
+// Note these constants and the struct are only used when dealing with the fallback path for RDP sesssions.
+
+// This is the location of the time zones in the registry on Vista+ systems.
+// See: https://docs.microsoft.com/windows/win32/api/timezoneapi/ns-timezoneapi-dynamic_time_zone_information
+#define WINDOWS_TIMEZONES_REG_KEY_PATH L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Time Zones"
+
+// Max length for a registry key is 255. +1 for null.
+// See: https://docs.microsoft.com/windows/win32/sysinfo/registry-element-size-limits
+#define WINDOWS_MAX_REG_KEY_LENGTH 256
+
+#if U_PLATFORM_HAS_WINUWP_API == 0
+
+// This is the layout of the TZI binary value in the registry.
+// See: https://docs.microsoft.com/windows/win32/api/timezoneapi/ns-timezoneapi-time_zone_information
+typedef struct _REG_TZI_FORMAT {
+    LONG Bias;
+    LONG StandardBias;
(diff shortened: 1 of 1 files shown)


axelandrejs (approved):

yumaoka:
Ha. It looks this is a workaround for the design issue in RDP. I remember I personally wrote something like this 20+ years ago when Microsoft translated registry keys in some (not all) localized version of Windows...

yumaoka (approved):
thanks

daniel-ju (approved):

jefgen:
I haven't merged yet as I'm still waiting for the Travis CI build to start:
https://travis-ci.org/github/unicode-org/icu/builds/738425098

I'm not sure what's causing all of the delays with Travis. For example it took 2+ hours on another PR before the build even started. The build for this PR hasn't even started yet either.

jefgen:
It looks like the Travis CI build finally ran eventually, and it was green. Thanks for merging @yumaoka!

PR #1433, opened 2020-10-26
ICU-21349 Improve MeasureUnit comparators

younies:
Compare two units using their conversion rates to the base instead of using a complete converter

##### Checklist

- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/units_complexconverter.cpp
+++ b/icu4c/source/i18n/units_complexconverter.cpp
@@ -37,28 +37,18 @@ ComplexUnitsConverter::ComplexUnitsConverter(const MeasureUnitImpl &inputUnit,
         outputUnits_.emplaceBackAndCheckErrorCode(status, units_[i]->copy(status).build(status));
     }
 
-    // NOTE:
-    //  This comparator is used to sort the units in a descending order. Therefore, we return -1 if
-    //  the left is bigger than right and so on.
+    // Sorts units in descending order. Therefore, we return -1 if
+    // the left is bigger than right and so on.
     auto descendingCompareUnits = [](const void *context, const void *left, const void *right) {
         UErrorCode status = U_ZERO_ERROR;
 
         const auto *leftPointer = static_cast<const MeasureUnitImpl *const *>(left);
         const auto *rightPointer = static_cast<const MeasureUnitImpl *const *>(right);
 
-        UnitConverter fromLeftToRight(**leftPointer,                                  //
-                                      **rightPointer,                                 //
-                                      *static_cast<const ConversionRates *>(context), //
-                                      status);
-
-        double rightFromOneLeft = fromLeftToRight.convert(1.0);
-        if (std::abs(rightFromOneLeft - 1.0) < 0.0000000001) { // Equals To
-            return 0;
-        } else if (rightFromOneLeft > 1.0) { // Greater Than
-            return -1;
-        }
-
-        return 1; // Less Than
+        return -1 * UnitConverter::compareTwoUnits(**leftPointer,                                  //
+                                                   **rightPointer,                                 //
+                                                   *static_cast<const ConversionRates *>(context), //
+                                                   status);
     };
 
     uprv_sortArray(units_.getAlias(),                                                                  //
--- a/icu4c/source/i18n/units_converter.cpp
+++ b/icu4c/source/i18n/units_converter.cpp
@@ -502,6 +502,52 @@ UnitConverter::UnitConverter(const MeasureUnitImpl &source, const MeasureUnitImp
                        ratesInfo, status);
 }
 
+int32_t UnitConverter::compareTwoUnits(const MeasureUnitImpl &firstUnit,
+                                       const MeasureUnitImpl &secondUnit,
+                                       const ConversionRates &ratesInfo, UErrorCode &status) {
+    if (U_FAILURE(status)) {
+        return 0;
+    }
+
+    if (firstUnit.complexity == UMeasureUnitComplexity::UMEASURE_UNIT_MIXED ||
+        secondUnit.complexity == UMeasureUnitComplexity::UMEASURE_UNIT_MIXED) {
+        status = U_INTERNAL_PROGRAM_ERROR;
+        return 0;
+    }
+
+    Convertibility unitsState = extractConvertibility(firstUnit, secondUnit, ratesInfo, status);
+    if (U_FAILURE(status)) {
+        return 0;
+    }
+
+    if (unitsState == Convertibility::UNCONVERTIBLE || unitsState == Convertibility::RECIPROCAL) {
+        status = U_INTERNAL_PROGRAM_ERROR;
+        return 0;
+    }
+
+    // Represents the conversion factor from the firstUnit to the base unit that specified in the
+    // conversion data which is considered as the root of the firstUnit and the secondUnit.
+    Factor firstUnitToBase = loadCompoundFactor(firstUnit, ratesInfo, status);
+    Factor secondUnitToBase = loadCompoundFactor(secondUnit, ratesInfo, status);
+
+    firstUnitToBase.substituteConstants();
+    secondUnitToBase.substituteConstants();
+
+    double firstUnitToBaseConversionRate = firstUnitToBase.factorNum / firstUnitToBase.factorDen;
+    double secondUnitToBaseConversionRate = secondUnitToBase.factorNum / secondUnitToBase.factorDen;
+
+    double diff = firstUnitToBaseConversionRate - secondUnitToBaseConversionRate;
+    if (diff > 0) {
+        return 1;
+    }
+
+    if (diff < 0) {
+        return -1;
+    }
+
+    return 0;
+}
+
 double UnitConverter::convert(double inputValue) const {
     double result =
         inputValue + conversionRate_.sourceOffset; // Reset the input to the target zero index.
--- a/icu4c/source/i18n/units_converter.h
+++ b/icu4c/source/i18n/units_converter.h
@@ -144,6 +144,16 @@ class U_I18N_API UnitConverter : public UMemory {
                   const ConversionRates &ratesInfo, UErrorCode &status);
 
     /**
+     * Compares two single units and returns 1 if the first one is greater, -1 if the second
+     * one is greater and 0 if they are equal.
+     *
+     * NOTE:
+     *  Compares only single units that are convertible.
+     */
+    static int32_t compareTwoUnits(const MeasureUnitImpl &firstUnit, const MeasureUnitImpl &SecondUnit,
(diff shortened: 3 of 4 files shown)


icu4c/source/i18n/units_converter.h
```diff
@@ -143,6 +143,16 @@ class U_I18N_API UnitConverter : public UMemory {
     UnitConverter(const MeasureUnitImpl &source, const MeasureUnitImpl &target,
                   const ConversionRates &ratesInfo, UErrorCode &status);
 
+    /**
+     * Compares two single units and returns and returns 1 if the first one is greater, -1 if the second
+     * one is greater and 0 if they are equal.
```
hugovdm, line 148: "and returns" appears twice.

icu4c/source/i18n/units_complexconverter.cpp
```diff
@@ -37,28 +37,19 @@ ComplexUnitsConverter::ComplexUnitsConverter(const MeasureUnitImpl &inputUnit,
...
-
-        double rightFromOneLeft = fromLeftToRight.convert(1.0);
-        if (std::abs(rightFromOneLeft - 1.0) < 0.0000000001) { // Equals To
-            return 0;
-        } else if (rightFromOneLeft > 1.0) { // Greater Than
-            return -1;
-        }
-
-        return 1; // Less Than
+        // Return -ve the result because we are sorting in descending order.
```
hugovdm, line 48: I don't understand "-ve" ?

hugovdm: Perhaps this comment is unnecessary given the comment on lines 40-41?

hugovdm (commented):
Looking good to me - are you satisfied with the unit tests impacted by this?

hugovdm (commented):
I would suggest shortening the pull request title, "ICU-21349 Improve measure unit comparisons", but I consider this optional.

A common English confusion though: there vs their. («There means the opposite of here; “at that place.” Their means “belongs to them.”»)

sffc (dismissed):
OK, but can you add additional unit tests for this behavior please, or point to where we have existing unit tests for it?

hugovdm:
> I would suggest shortening the pull request title, "[ICU-21349](https://unicode-org.atlassian.net/browse/ICU-21349) Improve measure unit comparisons", but I consider this optional.
> 
> A common English confusion though: there vs their. («There means the opposite of here; “at that place.” Their means “belongs to them.”»)

I've gone ahead and changed the pull request name. And I agree with Shane's request:

> can you add additional unit tests for this behavior please, or point to where we have existing unit tests for it?

Please consider doing this, even if we didn't explicitly "request changes".

hugovdm (changes requested):
Looking at the code review feedbac again, I decided to make "request changes" explicit.

icu4c/source/i18n/units_complexconverter.cpp
younies: done

icu4c/source/i18n/units_converter.h
younies: good catch.

younies (commented):
@sffc , @hugovdm.  there is no change in the behaviour, the change in the performance and there is testing for that.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -754,8 +754,9 @@ public MeasureUnitImplComparator(ConversionRates conversionRates) {
 
         @Override
         public int compare(MeasureUnitImpl o1, MeasureUnitImpl o2) {
-            UnitConverter fromO1toO2 = new UnitConverter(o1, o2, conversionRates);
-            return fromO1toO2.convert(BigDecimal.valueOf(1)).compareTo(BigDecimal.valueOf(1));
+            BigDecimal factor1 = this.conversionRates.getFactorToBase(o1).getConversionRate();
+            BigDecimal factor2 = this.conversionRates.getFactorToBase(o2).getConversionRate();
+            return factor1.compareTo(factor2);
```
hugovdm, line 759: Found the tests by changing `return -factor1.compareTo(factor2);`
causes failures in testComplexUnitConverterSorting, testComplexUnitsConverter, and testUnitPreferencesWithCLDRTests in UnitsTest.
In NumberFormatterApiTest, there are also nice failures (in unitMeasure and unitUsage).

I don't find any failures in com.ibm.icu.dev.test.format.MeasureUnitTest, which does mean MeasureUnitImpl is not tested thoroughly in MeasureUnitTest.

hugovdm: Probably makes sense: MeasureUnitImpl isn't generally tested, and the comparator is only used by complex conversion. In C++, measfmttest.cpp has only one test touching on MeasureUnitImpl, and it's the one I added to check on the leak.

icu4c/source/i18n/units_converter.cpp
```diff
@@ -502,6 +502,45 @@ UnitConverter::UnitConverter(const MeasureUnitImpl &source, const MeasureUnitImp
...
+    Factor secondUnitToBase = loadCompoundFactor(secondUnit, ratesInfo, status);
+
+    firstUnitToBase.substituteConstants();
+    secondUnitToBase.substituteConstants();
+
+    double firstUnitToBaseConversionRate = firstUnitToBase.factorNum / firstUnitToBase.factorDen;
+    double secondUnitToBaseConversionRate = secondUnitToBase.factorNum / secondUnitToBase.factorDen;
+
+    double diff = firstUnitToBaseConversionRate - secondUnitToBaseConversionRate;
+    if (std::abs(diff) <= DBL_EPSILON) {
```
hugovdm, line 533: This check doesn't do much. DBL_EPSILON is relevant relative to the value 1.0, while the conversion rates are probably far from 1.0. No unit tests fail when I delete this if. I think you can remove it.

hugovdm:
> OK, but can you add additional unit tests for this behavior please, or point to where we have existing unit tests for it?

As I commented in another thread, this is covered by tests in UnitsTest and NumberFormatterApiTest. It is not covered by MeasureUnitTest (Java) or MeasureFormatTest (C++) but I think that's fine: it's not a MeasureUnit API (and we don't generally test MeasureUnitImpl directly).

(Relevant tests in UnitsTest: testComplexUnitConverterSorting, testComplexUnitsConverter, testUnitPreferencesWithCLDRTests. Relevant tests in NumberFormatterApiTest: unitMeasure and unitUsage.)

> here is no change in the behaviour, the change in the performance and there is testing for that.

I think you mean there's testing for the unchanged behaviour (i.e. there isn't testing for performance).

So this looks good to me, just one more thing I caught: the DBL_EPSILON check can be removed, it's not doing much (it's not really relative to 1.0 - and we'll probably revisit our other epsilon checks as part of icu-units#108).

icu4c/source/i18n/units_converter.cpp
hugovdm: Rephrased: I do really like the intent of the logic, but it would probably require division, not subtraction, to check if the ratio is within epsilon of each other.

Considering division is relatively slow - and we probably don't care too much about sort stability for units considered to be the same, I think comparing diff to 0 makes sense.

(Consider two conversion rates that are both really tiny: DBL_EPSILON will start being a significant number.)

hugovdm: E.g. attometer to meter conversion rate is 1e-18, that's smaller than DBL_EPSILON. So I suspect comparing attometer to attoinch and this comparison will claim they're the same - we could prove it with a unit test. :-)

sffc (commented):
I will be happy with this PR once Hugo gives LGTM.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/MeasureUnitImpl.java
hugovdm: I'm marking this as resolved, I've repeated myself a bit in a comment below.

icu4c/source/i18n/units_converter.cpp
hugovdm: parsec to light-year, for example:

parsec = 3.086e+16 meters
factor1 -> 3.2404407e-17

light-year = 9.461e+15 meters
factor2 -> 1.0569707e-16

factor2 - factor1 = 1.0569707e-16 - 3.2404407e-17 = 7.3292663e-17
DBL_EPSILON = 2.2204460492503131e-16

However factor2/factor1 = 1.0569707e-16 / 3.2404407e-17 = 3.26181158013
That's what's relevant. (About 3.26 light-years to a parsec.)

So either compare the ratio to something like (1+DBL_EPSILON), or just don't worry about it and compare to zero.

younies: for this line
```
factor2 - factor1 = 1.0569707e-16 - 3.2404407e-17 = 7.3292663e-17
DBL_EPSILON = 2.2204460492503131e-16 // Bigger than factor2-factor1!
```

Do you think double can represent 7.3292663e-17 ?

by the way, I will just compare it to zero.

hugovdm (commented):
Looks good to me, just two minor requests - brace style, status-handling style - and an answer to your question.

I suggest pushing a commit, then squashing, then I'll be happy to approve the squashed commit.

icu4c/source/i18n/units_converter.cpp
hugovdm: Just FYI:

"Accurate to at least 15 digits", when wanting to simply express that constant value. But when doing this calculation, more precision is lost - that constant is itself actually already inaccurate.

Consider: using "bc", which can handle arbitrary precision, asking it to do 50 decimal places, and using the values for parsec and lightyear from units.txt:

```
scale=50
parsec=30856780000000000
1/parsec
.00000000000000003240778849899438632287620419240115
lightyear=9460730000000000
1/lightyear
.00000000000000010570008868237440451212538567319857
1/lightyear - 1/parsec
.00000000000000007329230018338001818924918148079742
```

If I'm doing this correctly, the more accurate answer should be nearer 7.329230018338e-17, not 7.3292663e-17 - so whatever I used to do the calculation before (Google?) already caused an error in the sixth decimal place. Hmmm.

hugovdm: Ah, logical: above, I rounded parsec and light-year to 4 significant digits, and did my further calculations with that. The point was only to get answer accurate enough to show that the result is much smaller than DBL_EPSILON.

hugovdm (dismissed):
Looks good to me, please squash!

hugovdm (approved):
We experimented: I was able to squash. I suggest PR authors can give permission to reviewers to squash, saving a round-trip.

PR #1435, opened 2020-10-27
ICU-21349 Fix mixed in C++

younies:
##### Checklist

- [x] Issue accepted
- [ ] Tests included


hugovdm:
I think this PR is not yet ready for review. It came up in the ICU meeting because it's in the queue: maybe mark it as "draft" until it is ready?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -55,6 +39,21 @@
...
+import java.lang.reflect.Method;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.text.FieldPosition;
+import java.text.Format;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
```
hugovdm, line 56: Import order is still changing. Please fix the import order, maybe sync with David on getting IntelliJ to follow ICU4J style?

hugovdm (changes requested):
The pull request title is unclear: what are you trying to fix? Something like "Fix parsing of mixed units in C++" perhaps?

You're currently touching 5 Java files and 6 C++ files - and the Java files have significant edits. This looks unaligned with the pull request title too: you're not only fixing C++. Consider how much you want to have as a part of this PR, and how big it might get.

- If you want to focus on getting C++ aligned with Java, keep this branch focused on that. It will make for a smaller pull request that is easier to review.
- If you want to do both, adjust the pull request title to better describe what you're intending to fix with it.

hugovdm (changes requested):
Before you fix the mixed unit case in C++ is a perfect time to create a unit test that shows the problem. I recommend you write a unit test as soon as possible, so you can see it fail, then when you fix the code, you've got some very clear evidence that it is an improvement, even before the reviewer analysis the logic to understand the improvement.

PR #1437, opened 2020-10-28
ICU-21356 Fix memory handling in MemoryPool::operator=()

hugovdm:
**Question to reviewer:** indirectly testing this via MeasureUnitImpl: sufficient? Or should I produce a unit test specifically for a MemoryPool instance?

The old array can't simply be deallocated, their pointers still need to be deallcoated too.

Using swap() ensures that the destructor can take care of it when `other` is deallocated.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21356
- [ ] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -725,9 +725,14 @@ public:
     }
 
     MemoryPool& operator=(MemoryPool&& other) U_NOEXCEPT {
-        fCount = other.fCount;
-        fPool = std::move(other.fPool);
-        other.fCount = 0;
+        // Since `this` may contain instances that need to be deleted, we can't
+        // just throw them away and replace them with `other`. The normal way of
+        // dealing with this in C++ is to swap `this` and `other`, rather than
+        // simply overwrite: the destruction of `other` can then take care of
+        // running MemoryPool::~MemoryPool() over the still-to-be-deallocated
+        // instances.
+        std::swap(fCount, other.fCount);
+        std::swap(fPool, other.fPool);
         return *this;
     }
 
@@ -796,9 +801,6 @@ protected:
 template<typename T, int32_t stackCapacity = 8>
 class MaybeStackVector : protected MemoryPool<T, stackCapacity> {
 public:
-    using MemoryPool<T, stackCapacity>::MemoryPool;
-    using MemoryPool<T, stackCapacity>::operator=;
-
     template<typename... Args>
     T* emplaceBack(Args&&... args) {
         return this->create(args...);
--- a/icu4c/source/test/intltest/measfmttest.cpp
+++ b/icu4c/source/test/intltest/measfmttest.cpp
@@ -17,6 +17,9 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include "charstr.h"
+#include "cstr.h"
+#include "measunit_impl.h"
 #include "unicode/decimfmt.h"
 #include "unicode/measfmt.h"
 #include "unicode/measure.h"
@@ -25,8 +28,6 @@
 #include "unicode/tmunit.h"
 #include "unicode/plurrule.h"
 #include "unicode/ustring.h"
-#include "charstr.h"
-#include "cstr.h"
 #include "unicode/reldatefmt.h"
 #include "unicode/rbnf.h"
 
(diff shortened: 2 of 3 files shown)


icu4c/source/common/cmemory.h
```diff
@@ -725,9 +725,14 @@ class MemoryPool : public UMemory {
...
     MemoryPool& operator=(MemoryPool&& other) U_NOEXCEPT {
-        fCount = other.fCount;
-        fPool = std::move(other.fPool);
-        other.fCount = 0;
+        // Since `this` may contain instances that need to be deleted, we can't
+        // just throw them away and replace them with `other`. The normal way of
+        // dealing with this in C++ is to swap `this` and `other`, rather than
+        // simply overwrite: the destruction of `other` can then take care of
+        // running MemoryPool::~MemoryPool() over the still-to-be-deallocated
+        // instances.
```
hugovdm, line 733: Excessive documentation for the codebase? I could have simply given this explanation in the code review itself.

icu4c/source/common/cmemory.h
```diff
@@ -725,6 +725,12 @@ class MemoryPool : public UMemory {
     }
 
     MemoryPool& operator=(MemoryPool&& other) U_NOEXCEPT {
+        // Since `this` may contain instances that need to be deleted, we can't
```
roubert, line 728: Ah, this is an important change, the original assumption here was that the object being assigned to would always be empty.

icu4c/source/common/cmemory.h
roubert: I think this is a place where a lot of documentation is good to have. It would probably have avoided confusion if the current version of this code had explicitly spelled out that the object being assigned to was expected to be empty.

icu4c/source/common/cmemory.h
```diff
@@ -101,7 +101,7 @@ uprv_calloc(size_t num, size_t size) U_MALLOC_ATTR U_ALLOC_SIZE_ATTR2(1,2);
   *    Clears any user heap functions from u_setMemoryFunctions()
   *    Does NOT deallocate any remaining allocated memory.
   */
-U_CFUNC UBool
```
roubert, line 104: The ICU4C code base has thousands of lines with trailing whitespace, making one random clean-up to one of them here on a line unrelated to the rest of the changes in the commit is just noise. Please revert this line.

icu4c/source/common/cmemory.h
```diff
@@ -725,9 +725,8 @@ class MemoryPool : public UMemory {
     }
 
     MemoryPool& operator=(MemoryPool&& other) U_NOEXCEPT {
-        fCount = other.fCount;
-        fPool = std::move(other.fPool);
-        other.fCount = 0;
+        std::swap(this->fCount, other.fCount);
```
roubert, line 734: For consistency with the surrounding code, I suggest omitting `this->` here.

roubert (dismissed):
I think this change makes sense, please just clean up the commits and make sure that all tests pass.

icu4c/source/common/cmemory.h
hugovdm: Thanks for answering the implied question: automatically trimming whitespace of edited files is thus not a useful IDE setting. I'll change it!

icu4c/source/common/cmemory.h
hugovdm: Done.

icu4c/source/common/cmemory.h
hugovdm: Done.

hugovdm:
I'm doing the usual leaving of squashing until after a reviewer is happy: an extra round trip, but makes review comments and reviewing diffs more streamlined.

roubert (commented):
The changes to `cmemory.h` look good to me now, I'll leave it to @sffc to review the changes to the units code.

hugovdm:
Yes, yes it does. :sigh:
Thus: *first* set up serving, *then* push to the branch.

sffc (approved):

hugovdm:
appveyor timed out: "Build execution time has reached the maximum allowed time for your plan (60 minutes)."

Merging!

PR #1443, opened 2020-11-04
ICU-21366 testConverter: allow some double-precision error for 0.0

hugovdm:
Java not included in this PR because Java doesn't have the same unit tests yet: icu-units#92 tracks that.
Closes: icu-units#112

This also takes the testConverter "maxDelta" from `1e-4*expected` to `1e-6*expected`. (Tests start failing for me at 1e-7.)

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21366
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/test/intltest/units_test.cpp
+++ b/icu4c/source/test/intltest/units_test.cpp
@@ -247,9 +247,12 @@ void UnitsTest::testConverter() {
             continue;
         }
 
+        double maxDelta = 1e-6 * uprv_fabs(testCase.expectedValue);
+        if (testCase.expectedValue == 0) {
+            maxDelta = 1e-12;
+        }
         assertEqualsNear(UnicodeString("testConverter: ") + testCase.source + " to " + testCase.target,
-                         testCase.expectedValue, converter.convert(testCase.inputValue),
-                         0.0001 * uprv_fabs(testCase.expectedValue));
+                         testCase.expectedValue, converter.convert(testCase.inputValue), maxDelta);
     }
 }
 


sffc (approved):
Wait for @younies approval as well.

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -247,9 +247,12 @@ void UnitsTest::testConverter() {
             continue;
         }
 
+        double maxDelta = 1e-6 * uprv_fabs(testCase.expectedValue);
+        if (testCase.expectedValue == 0) {
+            maxDelta = 1e-12;
+        }
         assertEqualsNear(UnicodeString("testConverter: ") + testCase.source + " to " + testCase.target,
```
younies, line 254: I suggest to make the behaviour of `assertEqualsNear` should be changed to compare the diff percentage with delta not just the diff

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -247,9 +247,12 @@ void UnitsTest::testConverter() {
             continue;
         }
 
+        double maxDelta = 1e-6 * uprv_fabs(testCase.expectedValue);
```
younies, line 250: We need to stick with specific level of accuracy. For example, if we decided that the deviation  could be 1e-4, then we should make the test cases follow that.

younies (approved):

icu4c/source/test/intltest/units_test.cpp
hugovdm: In this commit: https://github.com/unicode-org/icu/commit/65bbf92f782b1be6da7c4deb56512f9f33a2094e - I changed assertEqualsNear to match the behaviour of JUnit's assertEquals(double, double, double), because the ICU testing functions are modelled on ICU4J. This makes it easier to port testing functions between C++ and Java.

In the rebasing and squashing of commits, originaging in the sffc repository and then the icu-units repository, tracing this back to an original pull request isn't just a few clicks unfortunately.

Here's a pull request in which I originally provide an assertEquals that matches JUnit's behaviour, and we settled on assertEqualsNear (differing from JUnit): https://github.com/sffc/icu/pull/40
(Your review was "I think assertEqualsNear name is more expressive".)

The javadoc for this:
http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.String,%20double,%20double,%20double)

An absolute delta gives unit tests more control. But I need not argue that it's better, I just need to argue we should be consistent with JUnit and they already made this decision.

icu4c/source/test/intltest/units_test.cpp
hugovdm: 1e-4 is a terrible accuracy. I was already disappointed that this seems to have a delta bigger than 1e-7, but I'm indeed not digging into it yet, not considering it a part of this PR.

I did take this opportunity to show how accurate our test currently are, so we could also catch regressions when anything gets worse than this.

PR #1448, opened 2020-11-09
ICU-21349 Remove a potential 0/0 (=NaN). Align C++ & Java better.

hugovdm:
This improves some unit tests and makes C++ and Java code more similar, with an edge-case improvement in C++ complex number conversion accuracy.

I was considering tackling icu-units#108, but I think we should discuss what precision-handling differences we want between C++ and Java.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21349
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/units_complexconverter.cpp
+++ b/icu4c/source/i18n/units_complexconverter.cpp
@@ -132,24 +132,20 @@ MaybeStackVector<Measure> ComplexUnitsConverter::convert(double quantity,
     for (int i = 0, n = unitConverters_.length(); i < n; ++i) {
         quantity = (*unitConverters_[i]).convert(quantity);
         if (i < n - 1) {
-            // The double type has 15 decimal digits of precision. For choosing
-            // whether to use the current unit or the next smaller unit, we
-            // therefore nudge up the number with which the thresholding
-            // decision is made. However after the thresholding, we use the
-            // original values to ensure unbiased accuracy (to the extent of
-            // double's capabilities).
-            int64_t roundedQuantity = floor(quantity * (1 + DBL_EPSILON));
-            intValues[i] = roundedQuantity;
+            // If quantity is at the limits of double's precision from an
+            // integer value, we take that integer value.
+            int64_t flooredQuantity = floor(quantity * (1 + DBL_EPSILON));
+            intValues[i] = flooredQuantity;
 
             // Keep the residual of the quantity.
             //   For example: `3.6 feet`, keep only `0.6 feet`
-            //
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/units_complexconverter.cpp
```diff
@@ -132,24 +132,20 @@ MaybeStackVector<Measure> ComplexUnitsConverter::convert(double quantity,
...
-            // whether to use the current unit or the next smaller unit, we
-            // therefore nudge up the number with which the thresholding
-            // decision is made. However after the thresholding, we use the
-            // original values to ensure unbiased accuracy (to the extent of
-            // double's capabilities).
-            int64_t roundedQuantity = floor(quantity * (1 + DBL_EPSILON));
-            intValues[i] = roundedQuantity;
+            // If quantity is at the limits of double's precision from an
+            // integer value, we take that integer value.
+            int64_t flooredQuantity = floor(quantity * (1 + DBL_EPSILON));
```
hugovdm, line 137: Now that we have support for a `RoundingImpl *rounder` parameter, we could probably drop this DBL_EPSILON nudge by setting a default `rounder` instead. I'm happy to add it to this PR if you prefer, otherwise I can do this in a later PR.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/ComplexUnitsConverter.java
```diff
@@ -129,15 +129,17 @@ public boolean greaterThanOrEqual(BigDecimal quantity, BigDecimal limit) {
...
+                BigDecimal flooredQuantity =
                     quantity.multiply(EPSILON_MULTIPLIER).setScale(0, RoundingMode.FLOOR);
-                intValues.add(roundedQuantity);
+                intValues.add(flooredQuantity);
 
                 // Keep the residual of the quantity.
                 //   For example: `3.6 feet`, keep only `0.6 feet`
-                quantity = quantity.subtract(roundedQuantity);
-                if (quantity.compareTo(BigDecimal.ZERO) == -1) {
+                BigDecimal remainder = quantity.subtract(flooredQuantity);
```
hugovdm, line 138: Looking at this again, I do notice I'm adding a new variable without much reason - I'd be happy to change this back to just using quantity directly if someone prefers?
(`quantity = quantity.subtract(flooredQuantity);` for J, and `quantity -= flooredQuantity;` for C.)

sffc (approved):

PR #1450, opened 2020-11-10
ICU-21383 Fix memory problem in FormattedStringBuilder

sffc:
The fix to the reported issue, a list with more than 8 elements, is the addition of an argument to the `.resize()` call in `appendSpanInfo`.  The other fix, at the bottom of `nextPositionImpl`, came up in Valgrind during testing and is triggered when calling `.nextPosition()` repeatedly after it has returned false.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21383
- [ ] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/formattedval_sbimpl.cpp
+++ b/icu4c/source/i18n/formattedval_sbimpl.cpp
@@ -219,19 +219,35 @@ bool FormattedValueStringBuilderImpl::nextPositionImpl(ConstrainedFieldPosition&
     }
 
     U_ASSERT(currField == kUndefinedField);
+    // Always set the position to the end so that we don't revisit previous sections
+    cfpos.setState(
+        cfpos.getCategory(),
+        cfpos.getField(),
+        fString.fLength,
+        fString.fLength);
     return false;
 }
 
-void FormattedValueStringBuilderImpl::appendSpanInfo(int32_t spanValue, int32_t length) {
-    if (spanIndices.getCapacity() <= spanValue) {
-        spanIndices.resize(spanValue * 2);
+void FormattedValueStringBuilderImpl::appendSpanInfo(int32_t spanValue, int32_t length, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    U_ASSERT(spanIndices.getCapacity() >= spanValue);
+    if (spanIndices.getCapacity() == spanValue) {
+        if (!spanIndices.resize(spanValue * 2, spanValue)) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+            return;
+        }
     }
     spanIndices[spanValue] = {spanValue, length};
 }
 
-void FormattedValueStringBuilderImpl::prependSpanInfo(int32_t spanValue, int32_t length) {
-    if (spanIndices.getCapacity() <= spanValue) {
-        spanIndices.resize(spanValue * 2);
+void FormattedValueStringBuilderImpl::prependSpanInfo(int32_t spanValue, int32_t length, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    U_ASSERT(spanIndices.getCapacity() >= spanValue);
+    if (spanIndices.getCapacity() == spanValue) {
+        if (!spanIndices.resize(spanValue * 2, spanValue)) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+            return;
+        }
     }
     for (int32_t i = spanValue - 1; i >= 0; i--) {
         spanIndices[i+1] = spanIndices[i];
--- a/icu4c/source/test/intltest/listformattertest.cpp
+++ b/icu4c/source/test/intltest/listformattertest.cpp
@@ -551,6 +551,60 @@ void ListFormatterTest::TestFormattedValue() {
             expectedFieldPositions,
             UPRV_LENGTHOF(expectedFieldPositions));
     }
+
+    {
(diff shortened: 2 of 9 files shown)


sffc:
FYI @hugovdm: a bug one can make when using MaybeStackVector: don't forget to pass the optional length parameter to the resize function!

icu4c/source/test/intltest/listformattertest.cpp
```diff
@@ -551,6 +551,60 @@ void ListFormatterTest::TestFormattedValue() {
...
+            {UFIELD_CATEGORY_LIST_SPAN, 8, 24, 25},
+            {UFIELD_CATEGORY_LIST, ULISTFMT_ELEMENT_FIELD, 24, 25},
+            };
+        checkMixedFormattedValue(
+            message,
+            result,
+            expectedString,
+            expectedFieldPositions,
+            UPRV_LENGTHOF(expectedFieldPositions));
+    }
```
hugovdm, line 607: Should these tests not also be in ListFormatterTest.java ?

hugovdm (commented):
Looks good to me, except that the C++ and Java unit tests have falling out of sync?

sffc:
OK, I added the corresponding Java tests via amend.  This was an ICU4C-specific bug, but it doesn't hurt to add the ICU4J tests, since it helps to keep the two unit test suites in sync, as you pointed out.

icu4c/source/i18n/formattedval_sbimpl.cpp
```diff
@@ -219,12 +219,19 @@ bool FormattedValueStringBuilderImpl::nextPositionImpl(ConstrainedFieldPosition&
...
+        cfpos.getField(),
+        fString.fLength,
+        fString.fLength);
     return false;
 }
 
 void FormattedValueStringBuilderImpl::appendSpanInfo(int32_t spanValue, int32_t length) {
-    if (spanIndices.getCapacity() <= spanValue) {
-        spanIndices.resize(spanValue * 2);
+    U_ASSERT(spanIndices.getCapacity() >= spanValue);
```
FrankYFTang, line 232: U_ASSERT( A >= B) follow by a if ( A == B) does not look right to me. You should change the assert to U_ASSERT(spanIndices.getCapacity() >= spanValue) I think . If it is always >=  as the invariant of the code then the if == won't happen.

FrankYFTang: I believe it better be 
U_ASSERT(spanIndices.getCapacity() > spanValue);
if (spanIndices.getCapacity() <= spanValue) {

icu4c/source/i18n/formattedval_sbimpl.cpp
```diff
@@ -219,12 +219,19 @@ bool FormattedValueStringBuilderImpl::nextPositionImpl(ConstrainedFieldPosition&
...
+    if (spanIndices.getCapacity() == spanValue) {
```
FrankYFTang, line 233: why should you change it to ==? Shouldn't we keep it as <= ?

icu4c/source/i18n/formattedval_sbimpl.cpp
FrankYFTang, line 232: Should this be > only?

FrankYFTang: if you keep this as >= in the U_ASSERT then I think your debug build will bark when you run the test when spaceValue is 8

FrankYFTang (commented):
Please change the U_ASSERT condition to >

icu4c/source/i18n/formattedval_sbimpl.cpp
sffc: spanValue is never greater than getCapacity. The <= before was misleading; it should have always been ==. I added the U_ASSERT to emphasize that it's safe and equivalent to replace the <= with a ==.

Please note that the comparison is not always in the same order; sometimes it is A-B and other times B-A.

hugovdm (approved):
Looks good to me. (The assert also made sense to me.)

The MSYS2 failure looks unrelated to this PR:

    configure: error: in `/d/a/1/s/icu4c/source':
    configure: error: no acceptable C compiler found in $PATH

icu4c/source/i18n/formattedval_sbimpl.cpp
FrankYFTang: sorry, I must be too tired when I review your code last night. Yea you are right

FrankYFTang (dismissed):

macchiati (approved):

jefgen:
> The MSYS2 failure looks unrelated to this PR:
> 
> ```
> configure: error: in `/d/a/1/s/icu4c/source':
> configure: error: no acceptable C compiler found in $PATH
> ```

Hmm.. I'll try to re-run the build again...

markusicu (dismissed):
lgtm

sffc:
Thanks for all the reviews.  However, I caught two mistakes which I fixed in another amendment:

1. I need to change this behavior in `prependSpanInfo` too, not just `appendSpanInfo`
2. `.resize()` returns `nullptr` if it fails to allocate memory, but I'm not handling that case

Please re-review.

sffc:
*Note: The Jira-GitHub bot seems to have been down for a time. I manually re-sent the webhook request, which is why it was delayed.*

FrankYFTang (approved):

PR #1460, opened 2020-11-10
ICU-21295 Fix for Java version number overflow problem in 58

yumaoka:
Merging the fix for ICU-21219 (PR #1245) to ICU 58.

Internal API VersionInfo.javaVersion() maps Java version number to  4 integer fields. Each field must be up to 255. However, recent OpenJDK 8 update exceed this range.

Luckily, we have only one reference in our code base for checking Java version. CharsetUTF16 uses maxBytePerChar = 4 for Java 5 and older, maxBytePerChar = 2 for newer Java version. Because we no longer support Java 5 runtime, we don't need this conditional check.

We don't have any other uses of VersionInfo.javaVersion(). Java's version range is not what we can control, so I decided to delete the internal use only API completely.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21295
- [x] Issue accepted
- [ ] Tests included


yumaoka:
Close it for now - may try it later if necessary.

PR #1461, opened 2020-11-10
ICU-21385 Fix assertion when setKeywordValue w/ long value.

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21385
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -2457,9 +2457,13 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
     if (U_FAILURE(status)) {
         return;
     }
+    if (status == U_STRING_NOT_TERMINATED_WARNING) {
+        status = U_ZERO_ERROR;
+    }
     int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
     int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
                                              bufferLength, &status) + 1;
+    U_ASSERT(status != U_STRING_NOT_TERMINATED_WARNING);
     /* Handle the case the current buffer is not enough to hold the new id */
     if (status == U_BUFFER_OVERFLOW_ERROR) {
         U_ASSERT(newLength > bufferLength);
@@ -2476,6 +2480,7 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
         fullName = newFullName;
         status = U_ZERO_ERROR;
         uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status);
+        U_ASSERT(status != U_STRING_NOT_TERMINATED_WARNING);
     } else {
         U_ASSERT(newLength <= bufferLength);
     }
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -877,6 +877,9 @@ uloc_setKeywordValue(const char* keywordName,
     if(U_FAILURE(*status)) {
         return -1;
     }
+    if (*status == U_STRING_NOT_TERMINATED_WARNING) {
+        *status = U_ZERO_ERROR;
+    }
     if (keywordName == NULL || keywordName[0] == 0 || bufferCapacity <= 1) {
         *status = U_ILLEGAL_ARGUMENT_ERROR;
         return 0;
@@ -914,6 +917,7 @@ uloc_setKeywordValue(const char* keywordName,
     startSearchHere = (char*)locale_getKeywordsStart(buffer);
     if(startSearchHere == NULL || (startSearchHere[1]==0)) {
         if(keywordValueLen == 0) { /* no keywords = nothing to remove */
+            U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
             return bufLen;
         }
 
@@ -933,6 +937,7 @@ uloc_setKeywordValue(const char* keywordName,
         startSearchHere += keywordNameLen;
         *startSearchHere++ = '=';
         uprv_strcpy(startSearchHere, keywordValueBuffer);
+        U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
         return needLen;
     } /* end shortcut - no @ */
 
@@ -1047,13 +1052,27 @@ uloc_setKeywordValue(const char* keywordName,
     if (!handledInputKeyAndValue || U_FAILURE(*status)) {
         /* if input key/value specified removal of a keyword not present in locale, or
          * there was an error in CharString.append, leave original locale alone. */
+        U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
         return bufLen;
     }
 
     // needLen = length of the part before '@'
     needLen = (int32_t)(startSearchHere - buffer);
-    return needLen + updatedKeysAndValues.extract(
+    // Check to see can we fit the startSearchHere, if not, return
+    // U_BUFFER_OVERFLOW_ERROR without copy updatedKeysAndValues into it.
+    // We do this because this API function does not behave like most others:
+    // It promises never to set a U_STRING_NOT_TERMINATED_WARNING.
+    // When the contents fits but without the terminating NUL, in this case we need to not change
+    // the buffer contents and return with a buffer overflow error.
+    int32_t appendLength = updatedKeysAndValues.length();
+    if (appendLength >= bufferCapacity - needLen) {
+        *status = U_BUFFER_OVERFLOW_ERROR;
+        return needLen + appendLength;
+    }
+    needLen += updatedKeysAndValues.extract(
                          startSearchHere, bufferCapacity - needLen, *status);
+    U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
+    return needLen;
 }
 
 /* ### ID parsing implementation **************************************************/
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -2245,12 +2245,7 @@ static void TestKeywordSetError(void)
         strcpy(buffer,kwSetTestCases[i].l);
         status = U_ZERO_ERROR;
         res = uloc_setKeywordValue(kwSetTestCases[i].k, kwSetTestCases[i].v, buffer, blen, &status);
-        if(res == blen) {
-            if(status != U_STRING_NOT_TERMINATED_WARNING) {
-                log_err("expected not terminated warning on buffer %d got %s, len %d (%s + [%s=%s])\n", blen, u_errorName(status), res, kwSetTestCases[i].l, kwSetTestCases[i].k, kwSetTestCases[i].v);
-                return;
-            }
-        } else if(status != U_BUFFER_OVERFLOW_ERROR) {
+        if(status != U_BUFFER_OVERFLOW_ERROR) {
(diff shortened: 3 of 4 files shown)


icu4c/source/common/locid.cpp
```diff
@@ -2461,21 +2461,27 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
...
         }
         fullName = newFullName;
-        status = U_ZERO_ERROR;
-        uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status);
+        if (status == U_BUFFER_OVERFLOW_ERROR) {
+            status = U_ZERO_ERROR;
+            uloc_setKeywordValue(keywordName, keywordValue, fullName, newLength, &status);
+        }
     } else {
         U_ASSERT(newLength <= bufferLength);
```
markusicu, line 2486: You can tighten this to `<` now.

markusicu (commented):
Longer-term (not for this maintenance fix) it would be good if we could ByteSink-ify this code.

icu4c/source/common/locid.cpp
```diff
@@ -2461,21 +2461,27 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
     int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
                                              bufferLength, &status) + 1;
     /* Handle the case the current buffer is not enough to hold the new id */
-    if (status == U_BUFFER_OVERFLOW_ERROR) {
+    if (status == U_BUFFER_OVERFLOW_ERROR || status == U_STRING_NOT_TERMINATED_WARNING) {
         U_ASSERT(newLength > bufferLength);
         char* newFullName = (char *)uprv_malloc(newLength);
```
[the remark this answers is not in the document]
FrankYFTang: notice there was a "+ 1" in line 2462

icu4c/source/common/locid.cpp
```diff
@@ -2461,21 +2461,27 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
...
         if (newFullName == nullptr) {
             status = U_MEMORY_ALLOCATION_ERROR;
             return;
         }
-        uprv_strcpy(newFullName, fullName);
+        uprv_strncpy(newFullName, fullName, newLength-1);
```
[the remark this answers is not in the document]
FrankYFTang: > Please don't use strncpy(), it's unintuitive and inefficient. You are also dropping the last character with newLength-1.

Notice the newLength was already "+ 1" of the result from uloc_setKeywordValue so newLength - 1 is the same value as the value returned by uloc_setKeywordValue.

icu4c/source/common/locid.cpp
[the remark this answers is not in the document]
markusicu: Ah!
Well, easy to overlook, and very unusual.
I guess you could make it work however you want, but I would much prefer if you took off that `+ 1` there and adjusted the rest of the code to behave as expected.

icu4c/source/common/locid.cpp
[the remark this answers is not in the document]
FrankYFTang: > If buffer-overflow, then we don't know anything about the destination buffer, so don't copy anything.

We need to copy the content from fullName to newFullName, there were content already in fullName before calling the uloc_setKeywordValue, the uloc_setKeywordValue do not have enough memory in fullName to APPEND the keyword and value but once we get buffer overflow and reallocate memory, we need to copy what WAS in the fullName to the newFullName otherwise the value of the Locale will be lost.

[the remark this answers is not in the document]
FrankYFTang: > I would move this logic to after setting fullName:

 I don't think I can move the logic like this. I need to copy the content from fullName to newFullName in BOTH cases.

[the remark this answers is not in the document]
markusicu: Hm. This is all very unusual. Normally when a function sets an error code you can't rely on anything in the destination buffer.

I just looked up the API docs for the C function that is called here:
https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/uloc_8h.html#aee57778b4fdc2233f278bb6359205d4e

> NOTE: Unlike almost every other ICU function which takes a buffer, this function will NOT truncate the output text, and will not update the buffer with unterminated text setting a status of U_STRING_NOT_TERMINATED_WARNING. If a BUFFER_OVERFLOW_ERROR is received, it means a terminated version of the updated locale ID would not fit in the buffer, and the original buffer is untouched. This is done to prevent incorrect or possibly even malformed locales from being generated and used.

So... if this code here fails because it didn't consider a not-terminated warning, then the real bug is that we changed the behavior of uloc_setKeywordValue() to set such a warning when we promised we wouldn't do so. I assume that that's a change from @roubert that I approved, both of us assuming normal destination-buffer semantics.

Therefore,
- please revert your implementation changes here
- instead assert that there is no not-terminated warning
  - you should check for one before calling the C function, and reset to zero-error if we were called with the warning
- modify the C function implementation to not modify the destination buffer when length==capacity and set a buffer-overflow error instead; with a comment saying that the API docs promise this behavior
- the test case should be good as is

icu4c/source/common/locid.cpp
FrankYFTang: I don't think so the newLength is has a "+1" on top of what got returned by uloc_setKeywordValue in there already. (before my change)

FrankYFTang:
PTAL - Markus I think you missed the fact that the code has a +1 in the newLength already and the locale is already in the fullName before calling to this function.

FrankYFTang:
ok, I take out that + 1 and change the code, please review it again.

icu4c/source/common/locid.cpp
```diff
@@ -2459,25 +2459,34 @@ Locale::setKeywordValue(const char* keywordName, const char* keywordValue, UErro
     }
     int32_t bufferLength = uprv_max((int32_t)(uprv_strlen(fullName) + 1), ULOC_FULLNAME_CAPACITY);
     int32_t newLength = uloc_setKeywordValue(keywordName, keywordValue, fullName,
-                                             bufferLength, &status) + 1;
+                                             bufferLength, &status);
     /* Handle the case the current buffer is not enough to hold the new id */
-    if (status == U_BUFFER_OVERFLOW_ERROR) {
-        U_ASSERT(newLength > bufferLength);
-        char* newFullName = (char *)uprv_malloc(newLength);
+    if (status == U_BUFFER_OVERFLOW_ERROR || status == U_STRING_NOT_TERMINATED_WARNING) {
```
markusicu, line 2464: Hi Frank, I think you missed my big comment from this morning: https://github.com/unicode-org/icu/pull/1461#discussion_r521212124

In short, please change the implementation of uloc_setKeywordValue() back (supposedly) to never setting a not-terminated warning, and then here *assert* that there is no not-terminated warning coming out of that function.

In this function here, other than a new assertion, you can decide whether to just revert it to what it was, or whether to go forward from where you are with cleaning it up a bit. But we should never see a not-terminated warning here and thus need not handle it with special code.

Please see the details in my earlier comment.

FrankYFTang: ok, that make sense. I will work on that and let you know once I repush

sffc:
I am holding #1462 until this PR is merged.  However, I will move forward with #1462 on Friday if this PR is not merged by that time.

FrankYFTang:
PTAL

icu4c/source/common/uloc.cpp
```diff
@@ -914,6 +914,7 @@ uloc_setKeywordValue(const char* keywordName,
     startSearchHere = (char*)locale_getKeywordsStart(buffer);
     if(startSearchHere == NULL || (startSearchHere[1]==0)) {
         if(keywordValueLen == 0) { /* no keywords = nothing to remove */
+            U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
```
markusicu, line 917: The assertions in this function just before it returns are potentially harmful because you didn't check at function entry that the caller didn't give you a not-terminated warning. This is public API, so could be called from anyone, and a previous ICU function call may have left that warning. (That's one of the problems with our warnings, and why we stopped using them in newer API / for newer purposes.)

Please remove these new assertions from this function.

icu4c/source/common/uloc.cpp
```diff
@@ -933,6 +934,7 @@ uloc_setKeywordValue(const char* keywordName,
         startSearchHere += keywordNameLen;
         *startSearchHere++ = '=';
         uprv_strcpy(startSearchHere, keywordValueBuffer);
+        U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
```
markusicu, line 937: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1047,13 +1049,23 @@ uloc_setKeywordValue(const char* keywordName,
     if (!handledInputKeyAndValue || U_FAILURE(*status)) {
         /* if input key/value specified removal of a keyword not present in locale, or
          * there was an error in CharString.append, leave original locale alone. */
+        U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
```
markusicu, line 1052: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1047,13 +1049,23 @@ uloc_setKeywordValue(const char* keywordName,
...
+    // Check to see can we fit the startSearchHere, if not, return
+    // U_BUFFER_OVERFLOW_ERROR without copy updatedKeysAndValues into it.
+    int32_t appendLength = updatedKeysAndValues.length();
+    if (appendLength >= bufferCapacity - needLen) {
+        *status = U_BUFFER_OVERFLOW_ERROR;
+        return needLen + appendLength;
+    }
+    needLen += updatedKeysAndValues.extract(
                          startSearchHere, bufferCapacity - needLen, *status);
+    U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
```
markusicu, line 1067: Except I believe that *this assertion is ok* as is because I believe that .extract() resets an incoming not-terminated warning. So do keep this one.

icu4c/source/common/uloc.cpp
```diff
@@ -1047,13 +1049,23 @@ uloc_setKeywordValue(const char* keywordName,
...
+        U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
         return bufLen;
     }
 
     // needLen = length of the part before '@'
     needLen = (int32_t)(startSearchHere - buffer);
-    return needLen + updatedKeysAndValues.extract(
+    // Check to see can we fit the startSearchHere, if not, return
+    // U_BUFFER_OVERFLOW_ERROR without copy updatedKeysAndValues into it.
+    int32_t appendLength = updatedKeysAndValues.length();
```
markusicu, line 1060: I suggest you add more comment lines like:
```
// We do this because this API function does not behave like most others:
// It promises never to set a U_STRING_NOT_TERMINATED_WARNING.
// When the contents fits but without the terminating NUL, in this case we need to not change
// the buffer contents and return with a buffer overflow error.

File: icu4c/source/common/uloc.cpp [hunk 10]
```diff
@@ -877,6 +877,9 @@ uloc_setKeywordValue(const char* keywordName,
     if(U_FAILURE(*status)) {
         return -1;
     }
+    if (*status == U_STRING_NOT_TERMINATED_WARNING) {
+        *status == U_ZERO_ERROR;
```
markusicu, line 881: `=` not `==`

markusicu (commented):
One more thing -- please use `git commit --amend` to keep a single commit.

markusicu (approved):
lgtm tnx

PR #1464, opened 2020-11-12
ICU-21349 Fix confusing names of exponent variables.

hugovdm:
Addresses feedback given on unicode-org#1279 and sffc#21.

(These names were still confusing me - every time I dealt with the Java code in particular. On #1279 we had agreement that they needed to be fixed, so I'm fixing them.)

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21349
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/units_converter.cpp
+++ b/icu4c/source/i18n/units_converter.cpp
@@ -26,7 +26,7 @@ void U_I18N_API Factor::multiplyBy(const Factor &rhs) {
     factorNum *= rhs.factorNum;
     factorDen *= rhs.factorDen;
     for (int i = 0; i < CONSTANTS_COUNT; i++) {
-        constants[i] += rhs.constants[i];
+        constantExponents[i] += rhs.constantExponents[i];
     }
 
     // NOTE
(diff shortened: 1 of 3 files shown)


younies (dismissed):

sffc (approved):
I have a slight preference for "constants" over "exponents" but if you and Younies like this then I'm good with it

sffc:
Maybe "constantExponents" or "exponentsOfConstants"?

The reason I like "constants" better than "exponents" is because when I see "constants" I know exactly what it means, but when I see "exponents" I need to check if they are constant exponents, the exponent of individual single units, etc.

hugovdm:
> Maybe "constantExponents" or "exponentsOfConstants"?
> 
> The reason I like "constants" better than "exponents" is because when I see "constants" I know exactly what it means, but when I see "exponents" I need to check if they are constant exponents, the exponent of individual single units, etc.

I've gone with "exponentConstants" - I do find that fairly clear.

Variable naming vs variable documentation: if I see "constants", I certainly don't know exactly what that means. :-) Maybe until I've read the variable documentation and managed to associate "constants" with "exponents". (Otherwise it looks like we're changing the gravity constant.)

For Java, I'm leaving it as is? Otherwise the variable names become a bit long...

sffc (dismissed):
Much better than the first edition; thanks!

sffc (approved):

PR #1326, opened 2020-09-11
ICU-20568 Add .unit().usage() support to ICU4J NumberFormatter

younies:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20568
- [X] Updated PR title and link in previous line to include Issue number
- [X] Issue accepted
- [X] Tests included
- [X] Documentation is changed or added

ALLOW_MANY_COMMITS=true

PR #1465, opened 2020-11-12
ICU-21392 Fix Windows Time Zone detection when more than one possible time zone.

jefgen:
This PR is a fix for the regression caused by the change in PR #1326.

The list of possible time zones returned is a space-delimited list in some cases.
The change in PR #1362 was only stopping at the null-terminator. Instead we need to stop at the first space.

I missed this when reviewing the changes in PR https://github.com/unicode-org/icu/pull/1297.

The issue is due to the removal of the check here on this line which was stopping at the null or space:
https://github.com/unicode-org/icu/pull/1297/files#diff-04dfe6b718cf7d1421df2b6ff3aa89d0b95a7e2d468f5eb323e3698557a3a936L106

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21392
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/wintz.cpp
+++ b/icu4c/source/common/wintz.cpp
@@ -280,18 +280,29 @@ uprv_detectWindowsTimeZone()
     int regionCodeLen = GetGeoInfoW(geoId, GEO_ISO2, regionCodeW, UPRV_LENGTHOF(regionCodeW), 0);
 
     const UChar *icuTZ16 = nullptr;
-    int32_t tzLen;
+    int32_t tzListLen = 0;
 
     if (regionCodeLen != 0) {
         for (int i = 0; i < UPRV_LENGTHOF(regionCodeW); i++) {
             regionCode[i] = static_cast<char>(regionCodeW[i]);
         }
-        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), regionCode, &tzLen, &status);
+        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), regionCode, &tzListLen, &status);
     }
     if (regionCodeLen == 0 || U_FAILURE(status)) {
         // fallback to default "001" (world)
         status = U_ZERO_ERROR;
-        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), "001", &tzLen, &status);
+        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), "001", &tzListLen, &status);
+    }
+
+    // Note: We want the first entry in the string returned by ures_getStringByKey.
+    // However this string can be a space delimited list of timezones:
+    //  Ex: "America/New_York America/Detroit America/Indiana/Petersburg ..."
+    // We need to stop at the first space, so we pass tzLen (instead of tzListLen) to appendInvariantChars below.
+    int32_t tzLen = 0;
+    if (tzListLen > 0) {
+        while (!(icuTZ16[tzLen] == u'\0' || icuTZ16[tzLen] == u' ')) {
+            tzLen++;
+        }
     }
 
     // Note: cloneData returns nullptr if the status is a failure, so this


markusicu (dismissed):
lgtm

FrankYFTang (approved):
We need to think about how to unit test these codes....

jefgen:
> We need to think about how to unit test these codes....

FIWW, I think that unit testing is a good idea -- but it's somewhat tricky as this function directly calls platform APIs like the registry and such.

If we had something like [Detours](https://www.microsoft.com/en-us/research/project/detours/) to support mocking the Win32 platform APIs directly, then it would be easier. But I don't know if we want to use another library just to test one function.

daniel-ju (dismissed):

jefgen:
Thanks for the reviews!
Should I wait for the CI builds (Travis & AppVeyor) to finish -- or should I merge now anyways?

icu4c/source/common/wintz.cpp
```diff
@@ -280,18 +280,30 @@ uprv_detectWindowsTimeZone()
...
         status = U_ZERO_ERROR;
-        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), "001", &tzLen, &status);
+        icuTZ16 = ures_getStringByKey(winTZBundle.getAlias(), "001", &tzListLen, &status);
+    }
+
+    // Note: We want the first entry in the string returned by ures_getStringByKey.
+    // However this string can be a space delimited list of timezones:
+    //  Ex: "America/New_York America/Detroit America/Indiana/Petersburg ..."
+    // We need to stop at the first space, so we pass tzLen (instead of tzListLen) to appendInvariantChars below.
+    int32_t tzLen = tzListLen;
```
axelandrejs, line 301: Can this number be negative? If not, you could achieve the same thing by assigning 0 directly to tzlen and not need the inner assignment.

jefgen: Good point. Looking at the code, I don't think the length returned by `ures_getStringByKey` would/could be negative (even though it uses an `int32_t`). So I think your suggestion would probably be a bit cleaner. Thanks!

I'll force push an update in a moment...

axelandrejs (approved):

daniel-ju (approved):

FrankYFTang (approved):

sffc:
CC @mj1856

FrankYFTang:
It seems we see new problem coming in related to this when we start to roll out chrome. Which version of Windows have you test your code on? Would it work on older version of Windows?  https://bugs.chromium.org/p/chromium/issues/detail?id=1168528

jefgen:
> Which version of Windows have you test your code on? Would it work on older version of Windows?

Yes, it should work on older versions of Windows as well.

FWIW, I tested the change on the following versions:
- Windows 7 SP1 build 6.1.7601
- Windows 10 19H1 build 18363.418
- Windows 10 20H2 build 19042.631

I can't repo the issue anymore with the fix. I'll add a comment to the Cr bug, but I think we'll probably need more info in order to understand what might be the issue.

jefgen:
> https://bugs.chromium.org/p/chromium/issues/detail?id=1168528

Just to add an update on this PR:
For the Chromium bug, it's not related to this PR. Rather it was an issue in the changes for ICU-13845. There is a fix and more details in PR #1538.

markusicu:
> Rather it was an issue in the changes for [ICU-13845](https://unicode-org.atlassian.net/browse/ICU-13845). There is a fix and more details in PR #1538.

PR #1538 is unrelated. I assume you mean PR #1543?

jefgen:
> PR #1538 is unrelated. I assume you mean PR #1539 / PR #1543?

Yes, sorry for the confusion. I'll edit the comment above to fix it. Thanks.

jefgen:
Actually, it seems that editing the comment does not remove the link on GitHub... I'm going to delete the comment instead.

PR #1467, opened 2020-11-12
ICU-21395 Fix UBSan breakage on 8bit of rbbi

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21395
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/rbbitblb.cpp
+++ b/icu4c/source/common/rbbitblb.cpp
@@ -1402,12 +1402,13 @@ void RBBITableBuilder::exportTable(void *where) {
             U_ASSERT (sd->fAccepting <= 255);
             U_ASSERT (sd->fLookAhead <= 255);
             U_ASSERT (0 <= sd->fTagsIdx && sd->fTagsIdx <= 255);
-            row->r8.fAccepting = sd->fAccepting;
-            row->r8.fLookAhead = sd->fLookAhead;
-            row->r8.fTagsIdx   = sd->fTagsIdx;
+            RBBIStateTableRow8 *r8 = (RBBIStateTableRow8*)row;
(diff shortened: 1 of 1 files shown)


gnrunge (approved):

sffc (approved):
I don't understand why accessing the data via the union field was undefined behavior

jefgen:
> I don't understand why accessing the data via the union field was undefined behavior

I'm not sure either... maybe the compiler couldn't determine if it was the 8 or 16 bit table in the union that was being used?

aheninger:
I don't understand why it fails either. The original code should be OK.

Maybe the UBSan tool just has one required alignment for types, so "union RBBIStateTableRow {...}" gets stuck with requiring an alignment of 16 bits. Then the access of "row->" gets flagged, without determining that only the eight bit side is actually used.

At any rate, the fix is harmless, and if it makes the failure go away, I'm ok with it.

markusicu (approved):
lgtm tnx

I agree, weird. Might be worth reporting back to the UBSan people. I assume we have a contact in Google.

aheninger:
I think it may actually be undefined behavior. That a union must be aligned such that it could correctly store with any of its constituent members. So, in this case, having a pointer to the union type that will only work with the 8 bit types becomes problematic, even if the 16 bit types aren't actually used.

But I'm not certain, and haven't been able to find a clear reference with a little web searching.

gnrunge:
Is this discussion still open? Or is feedback to more learned authorities
in order?

aheninger:
I think this can be closed. The undefined behavior question is only of academic interest; the situation is removed from the code.

PR #1470, opened 2020-11-17
ICU-21401 Fix C++ canonicalize cel-gaulish to xtg

FrankYFTang:
Java is already doing so. Just add the same test to also assure it.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21401
- [x] Issue accepted
- [X] Tests included


markusicu (approved):
lgtm

Can we remove more of the LEGACY table?

sffc:
Both CI failures are flakes:

Travis:

```
There was an error in the .travis.yml file from which we could not recover.
Unfortunately, we do not know much about this error.
```

AppVeyor: (timed out after 1hr)

So it should be safe to merge.

PR #1471, opened 2020-11-18
ICU-21388 Use a read-only alias string in ucol_openElements and ucol_setText

jefgen:
The ICU APIs `ucol_openElements` and `ucol_setText` both are documented as requiring the caller to not modify or change the input string while the respective object is being used to iterate over the text.
(The API docs say "The caller must not modify or delete the text..." while they are being used. See https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/unicode/ucoleitr.h#L106 for example.)

In other words, they are documented as using a read-only alias of the input string from the caller. 

Both functions already use the `UnicodeString`'s read-only aliasing constructor for the input string arguments. However, some of the implementation methods that they call actually end up unfortunately making full copies of the input string, which hurts performance.

This change adds internal read-only alias methods to the `CollationElementIterator` and `RuleBasedCollator` classes, so that we can avoid making copies of the input string for the `ucol_openElements` and `ucol_setText` APIs.

This helps to improve the performance of not just the `ucol_openElements` and `ucol_setText` APIs, but also the `usearch_*` API, as it relies on the two previous APIs.

----

This PR is for this commit: https://github.com/microsoft/icu/commit/e6301ffbf704ec7b7fb20b006618f0e843c70754
(I thought that splitting up the commits into separate PRs would be easier to review the changes.)

[cc: @tarekgh as FYI.]

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21388
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/coleitr.h
+++ b/icu4c/source/i18n/unicode/coleitr.h
@@ -242,6 +242,19 @@ public:
     */
     void setText(const UnicodeString& str, UErrorCode& status);
 
+#ifndef U_HIDE_INTERNAL_API
+    /**
+    * Sets the source string. Same as the above setText method, but if readOnly
+    * is true then it uses a read-only alias of the input str string.
+    * The str string must outlive this object's lifetime.
+    * @param str the source string.
+    * @param readOnly use a read-only alias of the str string.
+    * @param status the error code status.
+    * @internal
+    */
+    void setText(const UnicodeString &str, UBool readOnly, UErrorCode &status);
+#endif
+
     /**
     * Sets the source string.
     * @param str the source character iterator.
@@ -322,6 +335,19 @@ private:
     */
     CollationElementIterator(const UnicodeString& sourceText,
         const RuleBasedCollator* order, UErrorCode& status);
+
+   /**
+    * CollationElementIterator read-only alias constructor. Same as the above constructor,
+    * but if readOnly is true then it uses a read-only alias of the sourceText string.
+    * The input source string must outlive this object's lifetime.
+    * @param sourceText    the source string.
+    * @param readOnly      if true, uses a read-only alias of sourceText.
+    * @param order         the collation object.
+    * @param status        the error code status.
+    */
+    CollationElementIterator(const UnicodeString& sourceText, UBool readOnly,
+        const RuleBasedCollator* order, UErrorCode& status);
+
     // Note: The constructors should take settings & tailoring, not a collator,
     // to avoid circular dependencies.
     // However, for operator==() we would need to be able to compare tailoring data for equality
--- a/icu4c/source/i18n/unicode/tblcoll.h
+++ b/icu4c/source/i18n/unicode/tblcoll.h
@@ -234,8 +234,8 @@ public:
 
     /**
      * Creates a collation element iterator for the source string. The caller of
-     * this method is responsible for the memory management of the return
-     * pointer.
+     * this method is responsible for the memory management of the returned pointer.
+     * 
      * @param source the string over which the CollationElementIterator will
      *        iterate.
      * @return the collation element iterator of the source string using this as
@@ -244,6 +244,25 @@ public:
      */
     virtual CollationElementIterator* createCollationElementIterator(
                                            const UnicodeString& source) const;
+
+#ifndef U_HIDE_INTERNAL_API
+     /**
+     * Creates a collation element iterator for the source string. The caller of
+     * this method is responsible for the memory management of the returned pointer.
+     * 
+     * This method is like the above createCollationElementIterator, but instead uses
+     * a read-only alias of the input source string.
+     * The input source string must outlive the returned object's lifetime.
+     * 
+     * @param source the string over which the CollationElementIterator will
+     *        iterate.
+     * @return the collation element iterator of the source string using this as
+     *         the based Collator.
+     * @internal
+     */
+    virtual CollationElementIterator* createCollationElementIteratorReadOnlyAlias(
+                                           const UnicodeString& source) const;
+#endif
 
     /**
      * Creates a collation element iterator for the source. The caller of this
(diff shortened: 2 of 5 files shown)


tarekgh:
@jefgen does this optimization can affect the scenario when using the Collator in multiple threads in same time?

jefgen:
> does this optimization can affect the scenario when using the Collator in multiple threads in same time?

The `CollationElementIterator` internally uses a `const` Collator:
https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/unicode/coleitr.h#L366

And the `usearch_openFromCollator` API also uses a `const` Collator as well:
https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/unicode/usearch.h

IIUC, I believe that `const` Collators can indeed be shared/used across multiple threads -- so long as only `const` methods are used on the Collators. i.e.: You can't share a Collator with other threads and and also mutate it at the same time.
(Based on reading the User Guide here: https://unicode-org.github.io/icu/userguide/design.html#thread-safe-const-apis).

So I think this optimization would apply regardless of if the Collator is used by itself, or with multiple threads.
(I think that hopefully answers your question (?), but please let me know if I might have missed it).

icu4c/source/i18n/unicode/coleitr.h
```diff
@@ -242,6 +242,19 @@ class U_I18N_API CollationElementIterator U_FINAL : public UObject {
...
+    /**
+    * Sets the source string. Same as the above setText method, but if readOnly
+    * is true then it uses a read-only alias of the input str string.
+    * The str string must outlive this object's lifetime.
+    * @param str the source string.
+    * @param readOnly use a read-only alias of the str string.
+    * @param status the error code status.
+    * @internal
+    */
+    void setText(const UnicodeString &str, UBool readOnly, UErrorCode &status);
```
markusicu, line 255: Please rename this function to `internalSetText(...)`.

Actually, I think it would be nicer to have separate functions, `setText()` which continues to make a copy, and `internalSetTextReference(string, status)` which does the fastCopyFrom() keeping the alias. More readable than a boolean parameter.

icu4c/source/i18n/unicode/coleitr.h
```diff
@@ -322,6 +335,19 @@ class U_I18N_API CollationElementIterator U_FINAL : public UObject {
...
+   /**
+    * CollationElementIterator read-only alias constructor. Same as the above constructor,
+    * but if readOnly is true then it uses a read-only alias of the sourceText string.
+    * The input source string must outlive this object's lifetime.
+    * @param sourceText    the source string.
+    * @param readOnly      if true, uses a read-only alias of sourceText.
+    * @param order         the collation object.
+    * @param status        the error code status.
+    */
+    CollationElementIterator(const UnicodeString& sourceText, UBool readOnly,
```
markusicu, line 348: Why not just add the readOnly parameter to the existing constructor? Too many call sites to modify? Might be a good opportunity to review them to see which needs what...

Another possibility is to not change the constructor nor add a new one at all. Instead, call sites could construct with an empty string and then call the new setter.

icu4c/source/i18n/unicode/tblcoll.h
```diff
@@ -245,6 +245,25 @@ class U_I18N_API RuleBasedCollator : public Collator {
...
+     * a read-only alias of the input source string.
+     * The input source string must outlive the returned object's lifetime.
+     * 
+     * @param source the string over which the CollationElementIterator will
+     *        iterate.
+     * @return the collation element iterator of the source string using this as
+     *         the based Collator.
+     * @internal
+     */
+    virtual CollationElementIterator* createCollationElementIteratorReadOnlyAlias(
```
markusicu, line 263: Please prefix the function name with "internal".

markusicu, line 263: Does it have to be virtual? If it does, you can't HIDE it because you would create an incompatible vtable.

I think a good alternative is to not add anything here. Change call sites to pass in an empty string and then call the new text setter. We can look at optimizing a CEI with no string / an empty string.

markusicu:
FYI good catch for finding the unintended copy (even if we have to live with it in existing C++ API) -- thanks!

markusicu:
Please modify this PR to use https://unicode-org.atlassian.net/browse/ICU-21533
Getting this into ICU 70 would be lovely. Maybe even early?

markusicu:
@jefgen ping :-)

> Please modify this PR to use https://unicode-org.atlassian.net/browse/ICU-21533
> Getting this into ICU 70 would be lovely. Maybe even early?

Note that you have another PR for ICU-21533. If that PR gets into ICU 70 but this PR slips past it, then we need to split the ticket again.

jefgen:
Thank you for all the feedback on this PR!

> Note that you have another PR for [ICU-21533](https://unicode-org.atlassian.net/browse/ICU-21533). If that PR gets into ICU 70 but this PR slips past it, then we need to split the ticket again.

Thanks Markus. Unfortunately haven't had time to work on this... I'll need to split the ticket and create a new one for ICU 71.

markusicu:
Hi @jefgen 

> Thank you for all the feedback on this PR!
> 
> > Note that you have another PR for [ICU-21533](https://unicode-org.atlassian.net/browse/ICU-21533). If that PR gets into ICU 70 but this PR slips past it, then we need to split the ticket again.
> 
> Thanks Markus. Unfortunately haven't had time to work on this... I'll need to split the ticket and create a new one for ICU 71.
> 
> EDIT: Created a new ticket here: https://unicode-org.atlassian.net/browse/ICU-21760

Could you please retarget this PR for ICU-21760 and finish it for 71?

jefgen:
> Could you please retarget this PR for [ICU-21760](https://unicode-org.atlassian.net/browse/ICU-21760) and finish it for 71?

Unfortunately, I don't think I'll have time to get this in for ICU 71.

Instead of keeping this PR open and punting it from release to release, I'm going to close this PR for now. There are no other commits on the ticket ICU-21760, so I'll move it "future", until there is more time to work on this.

PR #1472, opened 2020-11-18
ICU-21388 Remove test case that does nothing so RuleBasedCollator can be marked as final.

jefgen:
The `RuleBasedCollator` class isn't marked as `U_FINAL`.

It appears as though the only reason the class isn't marked as final is due to test code in the `intltest` suite which subclasses the `RuleBasedCollator` class.
This is despite the fact that the `RuleBasedCollator` is documented as "not to be subclassed". (See: https://github.com/unicode-org/icu/blob/master/icu4c/source/i18n/unicode/tblcoll.h#L112).

However, the test code calling the test-only subclass has actually been commented out since 1999... see: https://github.com/unicode-org/icu/commit/256f5b22ad48ebcf6e1ec9350d4c7b31579176b9

It appears as though the only intent of the test case was to ensure that the `RuleBasedCollator` called `createCollationElementIterator` (and it did this by adding a `count` member), but the comment from 1999 notes that `RuleBasedCollator` doesn't call `createCollationElementIterator` anymore.

Removing this unused test code allows us to mark the `RuleBasedCollator` class as `U_FINAL` -- which can help the compiler to better optimize things by providing more opportunities for devirtualization.

----

This PR is for this commit: https://github.com/microsoft/icu/commit/45cf36afec69d101b591632ee3d0ed9ba75d3943

I thought that splitting up the commits into separate PRs would be easier to review the changes.
Link to other PRs with perf changes: #1471 #1473

[cc: @tarekgh as FYI.]

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21388
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/unicode/tblcoll.h
+++ b/icu4c/source/i18n/unicode/tblcoll.h
@@ -112,7 +112,7 @@ class UVector64;
  * Note, RuleBasedCollator is not to be subclassed.
  * @see        Collator
  */
-class U_I18N_API RuleBasedCollator : public Collator {
+class U_I18N_API RuleBasedCollator U_FINAL : public Collator {
 public:
     /**
      * RuleBasedCollator constructor. This takes the table rules and builds a
--- a/icu4c/source/test/intltest/regcoll.cpp
+++ b/icu4c/source/test/intltest/regcoll.cpp
@@ -984,96 +984,6 @@ void CollationRegressionTest::Test4139572(/* char* par */)
 
     delete col;
 }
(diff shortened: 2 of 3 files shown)


markusicu (approved):
lgtm tnx

markusicu:
ready to merge?

jefgen:
> ready to merge?

Yes, sorry was swamped with other things and haven't had time to look at the ICU perf related changes yet. However, this one is ready to merge.

Thank you for taking a look and reviewing!

PR #1473, opened 2020-11-18
ICU-21533 Lazily create the internal break iterator used in StringSearch & improve error handling

jefgen:
The `usearch_*` API has an internal character break iterator that it uses for handling combining characters or sequences.

However, the internal break iterator isn't always strictly needed or used though -- for example when matching with pure ASCII text. Creating the break iterator is a somewhat non-trivial operation though...

I did some profiling using ICU 64 on Windows with a test program that called the usearch APIs in a loop (calling `usearch_open`, then `usearch_first`), and on average, anywhere from 7-10% of the time was spent on just creating the break iterator (which wasn't actually ever used).

By lazily creating the break iterator when needed we can improve the performance of the `usearch_*` API for cases where it isn't used.

Additionally, I noticed that there were several places in the code that performed memory allocation(s) without any error checking for OOM. This PR also adds/improves the error checking as well.

----

This PR is for this commit: https://github.com/microsoft/icu/commit/5e4e9b5a2914015710b21647b40ec028dec8a808

I thought that splitting up the commits into separate PRs would be easier to review the changes.
Link to other PRs with perf changes: #1471 #1472

[cc: @tarekgh as FYI.]

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21533
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/usearch.cpp
+++ b/icu4c/source/i18n/usearch.cpp
@@ -32,7 +32,7 @@ U_NAMESPACE_USE
 #define SECOND_LAST_BYTE_SHIFT_  8
 #define SUPPLEMENTARY_MIN_VALUE_ 0x10000
 
-static const Normalizer2Impl *g_nfcImpl = NULL;
+static const Normalizer2Impl *g_nfcImpl = nullptr;
 
 // internal methods -------------------------------------------------
 
@@ -44,11 +44,11 @@ static const Normalizer2Impl *g_nfcImpl = NULL;
 */
 static
 inline void setColEIterOffset(UCollationElements *elems,
-                      int32_t             offset)
+                              int32_t offset,
+                              UErrorCode &status)
 {
     // Note: Not "fast" any more after the 2013 collation rewrite.
     // We do not want to expose more internals than necessary.
-    UErrorCode status = U_ZERO_ERROR;
     ucol_setOffset(elems, offset, &status);
 }
 
@@ -75,7 +75,7 @@ inline uint32_t getMask(UCollationStrength strength)
 U_CDECL_BEGIN
 static UBool U_CALLCONV
 usearch_cleanup(void) {
-    g_nfcImpl = NULL;
+    g_nfcImpl = nullptr;
     return TRUE;
 }
 U_CDECL_END
@@ -89,7 +89,7 @@ U_CDECL_END
 static
 inline void initializeFCD(UErrorCode *status)
 {
-    if (g_nfcImpl == NULL) {
+    if (g_nfcImpl == nullptr) {
         g_nfcImpl = Normalizer2Factory::getNFCImpl(*status);
         ucln_i18n_registerCleanup(UCLN_I18N_USEARCH, usearch_cleanup);
     }
@@ -152,18 +152,18 @@ inline int32_t getCE(const UStringSearch *strsrch, uint32_t sourcece)
 }
 
 /**
-* Allocate a memory and returns NULL if it failed.
+* Allocate a memory and returns nullptr if it failed.
 * Internal method, status assumed to be a success.
 * @param size to allocate
 * @param status output error if any, caller to check status before calling
 *               method, status assumed to be success when passed in.
-* @return newly allocated array, NULL otherwise
+* @return newly allocated array, nullptr otherwise
 */
 static
 inline void * allocateMemory(uint32_t size, UErrorCode *status)
 {
     uint32_t *result = (uint32_t *)uprv_malloc(size);
-    if (result == NULL) {
+    if (result == nullptr) {
         *status = U_MEMORY_ALLOCATION_ERROR;
     }
     return result;
@@ -174,7 +174,7 @@ inline void * allocateMemory(uint32_t size, UErrorCode *status)
 * Creates a new array if we run out of space. The caller will have to
 * manually deallocate the newly allocated array.
 * Internal method, status assumed to be success, caller has to check status
-* before calling this method. destination not to be NULL and has at least
+* before calling this method. destination not to be nullptr and has at least
 * size destinationlength.
 * @param destination target array
 * @param offset destination offset to add value
@@ -199,7 +199,7 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
         int32_t *temp = (int32_t *)allocateMemory(
                                          sizeof(int32_t) * newlength, status);
         if (U_FAILURE(*status)) {
-            return NULL;
+            return nullptr;
         }
         uprv_memcpy(temp, destination, sizeof(int32_t) * (size_t)offset);
         *destinationlength = newlength;
@@ -214,7 +214,7 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
 * Creates a new array if we run out of space. The caller will have to
 * manually deallocate the newly allocated array.
 * Internal method, status assumed to be success, caller has to check status
-* before calling this method. destination not to be NULL and has at least
+* before calling this method. destination not to be nullptr and has at least
 * size destinationlength.
 * @param destination target array
 * @param offset destination offset to add value
@@ -240,7 +240,7 @@ inline int64_t * addTouint64_tArray(int64_t    *destination,
                                          sizeof(int64_t) * newlength, status);
 
         if (U_FAILURE(*status)) {
-            return NULL;
+            return nullptr;
         }
 
         uprv_memcpy(temp, destination, sizeof(int64_t) * (size_t)offset);
@@ -273,7 +273,7 @@ inline void initializePatternCETable(UStringSearch *strsrch, UErrorCode *status)
     uint32_t  patternlength      = pattern->textLength;
     UCollationElements *coleiter = strsrch->utilIter;
 
-    if (coleiter == NULL) {
+    if (coleiter == nullptr) {
         coleiter = ucol_openElements(strsrch->collator, pattern->text,
                                      patternlength, status);
         // status will be checked in ucol_next(..) later and if it is an
@@ -332,7 +332,7 @@ inline void initializePatternCETable(UStringSearch *strsrch, UErrorCode *status)
 */
 static
 inline void initializePatternPCETable(UStringSearch *strsrch,
-                                          UErrorCode    *status)
+                                      UErrorCode    *status)
 {
     UPattern *pattern            = &(strsrch->pattern);
     uint32_t  pcetablesize       = INITIAL_ARRAY_SIZE_;
@@ -340,11 +340,11 @@ inline void initializePatternPCETable(UStringSearch *strsrch,
     uint32_t  patternlength      = pattern->textLength;
     UCollationElements *coleiter = strsrch->utilIter;
 
-    if (coleiter == NULL) {
+    if (coleiter == nullptr) {
         coleiter = ucol_openElements(strsrch->collator, pattern->text,
                                      patternlength, status);
-        // status will be checked in ucol_next(..) later and if it is an
-        // error UCOL_NULLORDER the result of ucol_next(..) and 0 will be
+        // status will be checked in nextProcessed(..) later and if it is an error
+        // then UCOL_PROCESSED_NULLORDER is returned by nextProcessed(..), so 0 will be
         // returned.
         strsrch->utilIter = coleiter;
     } else {
@@ -354,7 +354,7 @@ inline void initializePatternPCETable(UStringSearch *strsrch,
         return;
     }
 
-    if (pattern->pces != pcetable && pattern->pces != NULL) {
+    if (pattern->pces != pcetable && pattern->pces != nullptr) {
         uprv_free(pattern->pces);
     }
 
@@ -366,7 +366,7 @@ inline void initializePatternPCETable(UStringSearch *strsrch,
     // ** Should processed CEs be signed or unsigned?
     // ** (the rest of the code in this file seems to play fast-and-loose with
     // **  whether a CE is signed or unsigned. For example, look at routine above this one.)
-    while ((pce = iter.nextProcessed(NULL, NULL, status)) != UCOL_PROCESSED_NULLORDER &&
+    while ((pce = iter.nextProcessed(nullptr, nullptr, status)) != UCOL_PROCESSED_NULLORDER &&
            U_SUCCESS(*status)) {
         int64_t *temp = addTouint64_tArray(pcetable, offset, &pcetablesize,
                               pce,
@@ -421,12 +421,12 @@ inline void initializePattern(UStringSearch *strsrch, UErrorCode *status)
     }
 
     // ** HACK **
-    if (strsrch->pattern.pces != NULL) {
+    if (strsrch->pattern.pces != nullptr) {
         if (strsrch->pattern.pces != strsrch->pattern.pcesBuffer) {
             uprv_free(strsrch->pattern.pces);
         }
 
-        strsrch->pattern.pces = NULL;
+        strsrch->pattern.pces = nullptr;
     }
 
     initializePatternCETable(strsrch, status);
@@ -444,21 +444,57 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
     initializePattern(strsrch, status);
 }
 
+#if !UCONFIG_NO_BREAK_ITERATION
+// If the caller provided a character breakiterator we'll return that,
+// otherwise we lazily create the internal break iterator. 
+static UBreakIterator* getBreakIterator(UStringSearch *strsrch, UErrorCode &status)
+{
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+
+    if (strsrch->search->breakIter != nullptr) {
+        return strsrch->search->breakIter;
+    }
+
+    if (strsrch->search->internalBreakIter != nullptr) {
+        return strsrch->search->internalBreakIter;
+    }
+
+    // Need to create the internal break iterator.
+    strsrch->search->internalBreakIter = ubrk_open(UBRK_CHARACTER,
+        ucol_getLocaleByType(strsrch->collator, ULOC_VALID_LOCALE, &status),
+        strsrch->search->text, strsrch->search->textLength, &status);
+
+    return strsrch->search->internalBreakIter;
+}
+#endif
+
 /**
-* sets match not found
+* Sets the match result to "not found", regardless of the incoming error status.
+* If an error occurs while setting the result, it is reported back.
+* 
 * @param strsrch string search data
+* @param status  for output errors, if they occur.
 */
 static
-inline void setMatchNotFound(UStringSearch *strsrch)
+inline void setMatchNotFound(UStringSearch *strsrch, UErrorCode &status)
 {
-    // this method resets the match result regardless of the error status.
+    UErrorCode localStatus = U_ZERO_ERROR;
+
     strsrch->search->matchedIndex = USEARCH_DONE;
     strsrch->search->matchedLength = 0;
     if (strsrch->search->isForwardSearching) {
-        setColEIterOffset(strsrch->textIter, strsrch->search->textLength);
+        setColEIterOffset(strsrch->textIter, strsrch->search->textLength, localStatus);
     }
     else {
-        setColEIterOffset(strsrch->textIter, 0);
+        setColEIterOffset(strsrch->textIter, 0, localStatus);
+    }
+
+    // If an error occurred while setting the result to not found (ex: OOM),
+    // then we want to report that error back to the caller.
+    if (U_SUCCESS(status) && U_FAILURE(localStatus)) {
+        status = localStatus;
     }
 }
 
@@ -482,8 +518,7 @@ inline UBool isOutOfBounds(int32_t textlength, int32_t offset)
 * @return TRUE if identical match is found
 */
 static
-inline UBool checkIdentical(const UStringSearch *strsrch, int32_t start,
-                                  int32_t    end)
+inline UBool checkIdentical(const UStringSearch *strsrch, int32_t start, int32_t end)
 {
     if (strsrch->strength != UCOL_IDENTICAL) {
         return TRUE;
@@ -512,12 +547,12 @@ U_CAPI UStringSearch * U_EXPORT2 usearch_open(const UChar *pattern,
                                           UErrorCode     *status)
 {
     if (U_FAILURE(*status)) {
-        return NULL;
+        return nullptr;
     }
 #if UCONFIG_NO_BREAK_ITERATION
-    if (breakiter != NULL) {
+    if (breakiter != nullptr) {
         *status = U_UNSUPPORTED_ERROR;
-        return NULL;
+        return nullptr;
     }
 #endif
     if (locale) {
@@ -528,11 +563,11 @@ U_CAPI UStringSearch * U_EXPORT2 usearch_open(const UChar *pattern,
                                               patternlength, text, textlength,
                                               collator, breakiter, status);
 
-        if (result == NULL || U_FAILURE(*status)) {
+        if (result == nullptr || U_FAILURE(*status)) {
             if (collator) {
                 ucol_close(collator);
             }
-            return NULL;
+            return nullptr;
         }
         else {
             result->ownCollator = TRUE;
@@ -540,7 +575,7 @@ U_CAPI UStringSearch * U_EXPORT2 usearch_open(const UChar *pattern,
         return result;
     }
     *status = U_ILLEGAL_ARGUMENT_ERROR;
-    return NULL;
+    return nullptr;
 }
 
 U_CAPI UStringSearch * U_EXPORT2 usearch_openFromCollator(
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/usearch.cpp
```diff
@@ -565,16 +565,26 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
...
 * @param start target text start offset
 * @param end target text end offset
 */
 static
 void checkBreakBoundary(const UStringSearch *strsrch, int32_t * /*start*/,
-                               int32_t *end)
+                               int32_t *end, UErrorCode *status)
 {
 #if !UCONFIG_NO_BREAK_ITERATION
+    // Lazily create the internal break iterator.
```
aheninger, line 578: This is inside `#if BOYER_MOORE`, which is hardcoded to 0 at line 31.
Also, wherever the lazy creation ends up, would want to check for a caller-supplied break iterator, `strsrch->search->breakIter`, before creating an internal one.

icu4c/source/i18n/usearch.cpp
```diff
@@ -3668,6 +3689,16 @@ static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
     UBreakIterator *breakiterator = strsrch->search->breakIter;
 
     if (breakiterator == NULL) {
+        // Lazily create the internal break iterator.
```
aheninger, line 3692: It may make sense to put this creation logic in a little `getInternalBreakIterator()` function, to avoid repeating it multiple times. Or maybe it could be `getBreakIterator()`, and include the test for a user supplied one.

jefgen: Thanks Andy! That's a good idea!
I think I'll add your suggested `getBreakIterator()` function, and then replace the repeated/redundant code with it.

icu4c/source/i18n/usearch.cpp
jefgen: Thanks for taking a look!

Right, this is inside the `BOYER_MOORE` parts of the code that don't seem to be used at all anymore. However, I thought that it would be better to try and change all the places in the file that used the break iterator anyways...

Looking at the history in git, it seems that `usearch.cpp` hasn't actually compiled with the macro `BOYER_MOORE` set to `1` since the Collation V2 rewrite in ~2014. (It tries to use functions that were removed/changed since then). So perhaps this is a pointless change -- but I at least didn't want to make things any worse off in case somebody in the future wanted to try and get the `BOYER_MOORE` code working again.

Good point about checking for the caller-supplied break iterator first!
I'll force-push an update for this in a bit.

icu4c/source/i18n/usearch.cpp
```diff
@@ -562,6 +562,26 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
     strsrch->pattern.defaultShiftSize = 0;
 }
 
+// If the caller provided a character breakiterator we'll return that,
+// otherwise we lazily create the internal break iterator. 
+static UBreakIterator* getBreakIterator(UStringSearch *strsrch, UErrorCode *status)
+{
+    if (strsrch->search->breakIter != NULL) {
```
markusicu, line 569: optional: Would you mind using nullptr in new code?

icu4c/source/i18n/usearch.cpp
```diff
@@ -48,11 +48,11 @@ static const Normalizer2Impl *g_nfcImpl = NULL;
 */
 static
 inline void setColEIterOffset(UCollationElements *elems,
-                      int32_t             offset)
+                              int32_t offset,
+                              UErrorCode& status)
```
markusicu, line 52: optional: I have a preference for `UErrorCode &status` with the ampersand adjacent to the parameter name, like how C types were intended and how the code here uses. But I understand that that's falling out of vogue.

icu4c/source/i18n/usearch.cpp
```diff
@@ -572,11 +592,11 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
 */
 static
 void checkBreakBoundary(const UStringSearch *strsrch, int32_t * /*start*/,
-                               int32_t *end)
+                               int32_t *end, UErrorCode *status)
```
markusicu, line 595: reference please for the status, not a pointer

icu4c/source/i18n/usearch.cpp
```diff
@@ -562,6 +562,26 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
     strsrch->pattern.defaultShiftSize = 0;
 }
 
+// If the caller provided a character breakiterator we'll return that,
+// otherwise we lazily create the internal break iterator. 
+static UBreakIterator* getBreakIterator(UStringSearch *strsrch, UErrorCode *status)
```
markusicu, line 567: reference please for the status, not a pointer

icu4c/source/i18n/usearch.cpp
```diff
@@ -562,6 +562,26 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
...
+{
```
markusicu, line 568: Before you do anything else, if U_FAILURE then exit with nullptr.

icu4c/source/i18n/usearch.cpp
```diff
@@ -572,11 +592,11 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
...
 static
 void checkBreakBoundary(const UStringSearch *strsrch, int32_t * /*start*/,
-                               int32_t *end)
+                               int32_t *end, UErrorCode *status)
 {
 #if !UCONFIG_NO_BREAK_ITERATION
-    UBreakIterator *breakiterator = strsrch->search->internalBreakIter;
-    if (breakiterator) {
+    UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
+    if (breakiterator != NULL && U_SUCCESS(status)) {
```
markusicu, line 599: It looks like it should suffice to check for success but not also for nullptr.
Why did the old code check the pointer? Should it be possible to be nullptr without an error condition? Or just defensive coding against usage of an error-constructed StringSearch object?

icu4c/source/i18n/usearch.cpp
```diff
@@ -602,13 +622,12 @@ void checkBreakBoundary(const UStringSearch *strsrch, int32_t * /*start*/,
...
-UBool isBreakUnit(const UStringSearch *strsrch, int32_t start,
-                               int32_t    end)
+UBool isBreakUnit(const UStringSearch *strsrch, int32_t start, int32_t end, UErrorCode *status)
 {
 #if !UCONFIG_NO_BREAK_ITERATION
-    UBreakIterator *breakiterator = strsrch->search->breakIter;
+    UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
     //TODO: Add here.
-    if (breakiterator) {
+    if (breakiterator != NULL && U_SUCCESS(status)) {
```
markusicu, line 630: ditto

icu4c/source/i18n/usearch.cpp
```diff
@@ -765,16 +784,16 @@ inline int32_t shiftForward(UStringSearch *strsrch,
 * @param strsrch string search data
 */
 static
-inline void setMatchNotFound(UStringSearch *strsrch)
+inline void setMatchNotFound(UStringSearch *strsrch, UErrorCode& status)
 {
```
markusicu, line 788: Before you do anything else, if U_FAILURE then exit immediately.

icu4c/source/i18n/usearch.cpp
```diff
@@ -3626,7 +3656,7 @@ U_NAMESPACE_END
  * has an external break iterator, use that. Otherwise use the internal character
  * break iterator.
  */
-static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
+static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex, UErrorCode *status) {
 #if 0
     const UChar *text = strsrch->search->text;
```
markusicu, line 3661: In this branch, if U_FAILURE then return immediately; maybe return startIndex?

icu4c/source/i18n/usearch.cpp
```diff
@@ -3665,17 +3695,12 @@ static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
...
-    UBreakIterator *breakiterator = strsrch->search->breakIter;
-
-    if (breakiterator == NULL) {
-        breakiterator = strsrch->search->internalBreakIter;
-    }
-
-    if (breakiterator != NULL) {
-        return ubrk_following(breakiterator, startIndex);
+    UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
+    if (breakiterator == NULL || U_FAILURE(*status)) {
```
markusicu, line 3699: check status only, as above?

icu4c/source/i18n/usearch.cpp
```diff
@@ -3688,7 +3713,7 @@ static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
  * has an external break iterator, test using that, otherwise test
  * using the internal character break iterator.
  */
-static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index) {
+static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index, UErrorCode *status) {
 #if 0
     const UChar *text = strsrch->search->text;
```
markusicu, line 3718: In this branch, if U_FAILURE then return immediately; maybe return false?

icu4c/source/i18n/usearch.cpp
```diff
@@ -3716,13 +3741,12 @@ static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index) {
     UBool combining =  !(gcProperty==U_GCB_CONTROL || gcProperty==U_GCB_LF || gcProperty==U_GCB_CR);
     return !combining;
 #elif !UCONFIG_NO_BREAK_ITERATION
-    UBreakIterator *breakiterator = strsrch->search->breakIter;
-
-    if (breakiterator == NULL) {
-        breakiterator = strsrch->search->internalBreakIter;
+    UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
+    if (breakiterator == NULL || U_FAILURE(*status)) {
```
markusicu, line 3745: check status only, as above?

icu4c/source/i18n/usearch.cpp
```diff
@@ -3733,9 +3757,8 @@ static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index) {
 static UBool onBreakBoundaries(const UStringSearch *strsrch, int32_t start, int32_t end)
 {
 #if !UCONFIG_NO_BREAK_ITERATION
-    UBreakIterator *breakiterator = strsrch->search->breakIter;
-
-    if (breakiterator != NULL) {
+    UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
+    if (breakiterator != NULL && U_SUCCESS(status)) {
```
markusicu, line 3761: check status only, as above?

icu4c/source/i18n/usearch.cpp
```diff
@@ -3733,9 +3757,8 @@ static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index) {
 static UBool onBreakBoundaries(const UStringSearch *strsrch, int32_t start, int32_t end)
```
markusicu, line 3757: add status parameter?

icu4c/source/i18n/usearch.cpp
```diff
@@ -4189,8 +4232,11 @@ U_CAPI UBool U_EXPORT2 usearch_searchBackwards(UStringSearch  *strsrch,
      * consider for the match.
      */
     if (startIdx < strsrch->search->textLength) {
-        UBreakIterator *bi = strsrch->search->internalBreakIter;
-        int32_t next = ubrk_following(bi, startIdx);
+        UBreakIterator *breakiterator = getBreakIterator(strsrch, status);
+        if (breakiterator == NULL|| U_FAILURE(*status)) {
```
markusicu, line 4236: check status only, as above?

icu4c/source/i18n/usearch.cpp
```diff
@@ -602,13 +622,12 @@ void checkBreakBoundary(const UStringSearch *strsrch, int32_t * /*start*/,
 * @param end target text end offset
 */
 static
-UBool isBreakUnit(const UStringSearch *strsrch, int32_t start,
-                               int32_t    end)
+UBool isBreakUnit(const UStringSearch *strsrch, int32_t start, int32_t end, UErrorCode *status)
```
markusicu, line 625: reference please for the status, not a pointer

icu4c/source/i18n/usearch.cpp
```diff
@@ -3626,7 +3656,7 @@ U_NAMESPACE_END
  * has an external break iterator, use that. Otherwise use the internal character
  * break iterator.
  */
-static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
+static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex, UErrorCode *status) {
```
markusicu, line 3659: reference please for the status, not a pointer

icu4c/source/i18n/usearch.cpp
```diff
@@ -3688,7 +3713,7 @@ static int32_t nextBoundaryAfter(UStringSearch *strsrch, int32_t startIndex) {
  * has an external break iterator, test using that, otherwise test
  * using the internal character break iterator.
  */
-static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index) {
+static UBool isBreakBoundary(UStringSearch *strsrch, int32_t index, UErrorCode *status) {
```
markusicu, line 3716: reference please for the status, not a pointer

aheninger (approved):
A fair number of the changes are to code that is excluded by `#if BOYER_MOORE`

Maybe we should think about removing it entirely; it's presence makes the active code much harder to see and follow, and, as you noted, it's been dead for ages. That should probably be done in a separate PR, though.

markusicu:
> A fair number of the changes are to code that is excluded by `#if BOYER_MOORE`
> 
> Maybe we should think about removing it entirely; it's presence makes the active code much harder to see and follow, and, as you noted, it's been dead for ages. That should probably be done in a separate PR, though.

I am ok with doing that in a separate PR (before or after the lazy creation).

markusicu:
Also, Jeff, it would be nice if you could do the lazy creation in Java as well.

aheninger (dismissed):
Dang Github; I didn't intend to approve in my previous comment.

markusicu:
Please modify this PR to use https://unicode-org.atlassian.net/browse/ICU-21533
Getting this into ICU 70 would be lovely. Maybe even early?

jefgen:
Sorry for the long delay in getting back to this PR...

> > A fair number of the changes are to code that is excluded by `#if BOYER_MOORE`
> > Maybe we should think about removing it entirely; it's presence makes the active code much harder to see and follow, and, as you noted, it's been dead for ages. That should probably be done in a separate PR, though.
> 
> I am ok with doing that in a separate PR (before or after the lazy creation).

Thank you both. That sounds like a good idea. I'll file a separate ticket for doing that, and will make that change first. Once that is done, I'll rebase the changes in this PR which should hopefully make things much cleaner.

jefgen:
I forced-pushed an update to this PR after creating separate PRs for cleaning up the BOYER_MOORE code (#1800, #1830).
(I rebased on top of the latest changes from `main`, so there are no longer any changes for the dead BM code in this PR).
The updated PR also incorporates changes based on the feedback from Andy and Markus on the review.
Thank you both for the feedback!

icu4c/source/i18n/usearch.cpp
jefgen: I think I'll actually do a mass NULL -> nullptr change for this whole file, as it seems odd to leave it in a mixed state that is ~50/50% NULL and nullptr.

icu4c/source/i18n/usearch.cpp
jefgen: Changed to use `UErrorCode &status`.

icu4c/source/i18n/usearch.cpp
jefgen: Good point, thanks!

icu4c/source/i18n/usearch.cpp
jefgen: Good point.
I think it was just defensive coding, as it doesn't appear to be possible to have nullptr without the UErrorCode status also being an error at the same time.
I'll update to only check the error status (and not the pointer) in order to simplify things.
Thanks!

icu4c/source/i18n/usearch.cpp
jefgen: Thanks!

icu4c/source/i18n/usearch.cpp
```diff
@@ -562,6 +562,26 @@ inline void initialize(UStringSearch *strsrch, UErrorCode *status)
...
+    if (strsrch->search->breakIter != NULL) {
+        return strsrch->search->breakIter;
+    }
+
+    if (strsrch->search->internalBreakIter != NULL) {
+        return strsrch->search->internalBreakIter;
+    }
+
+    // Need to create the internal break iterator.
+    strsrch->search->internalBreakIter = ubrk_open(UBRK_CHARACTER,
```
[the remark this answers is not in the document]
jefgen: Thanks! I moved the whole function inside a `#if !UCONFIG_NO_BREAK_ITERATION` block.

jefgen:
Ah, I just realized that the commit has the wrong ICU ticket number on it. I'll force push an update to fix it.

jefgen:
I fixed the commit ticket number and rebased on top of the latest changes from `main`. PTAL. :)
Thank you!

icu4c/source/i18n/usearch.cpp
```diff
@@ -2313,7 +2387,7 @@ U_CAPI UBool U_EXPORT2 usearch_searchBackwards(UStringSearch  *strsrch,
 UBool usearch_handleNextExact(UStringSearch *strsrch, UErrorCode *status)
 {
     if (U_FAILURE(*status)) {
-        setMatchNotFound(strsrch);
+        setMatchNotFound(strsrch, *status);
```
aheninger, line 2390: `setMatchNotFound()` has been changed to take no action when there is an incoming error.
Does it need to reset the matching state to no match even when there is an error? As is guaranteed to be the case here.

icu4c/source/i18n/usearch.cpp
```diff
@@ -947,9 +977,12 @@ U_CAPI void U_EXPORT2 usearch_setCollator(      UStringSearch *strsrch,
             strsrch->strength    = ucol_getStrength(collator);
             strsrch->ceMask      = getMask(strsrch->strength);
 #if !UCONFIG_NO_BREAK_ITERATION
-            ubrk_close(strsrch->search->internalBreakIter);
-            strsrch->search->internalBreakIter = ubrk_open(UBRK_CHARACTER, ucol_getLocaleByType(collator, ULOC_VALID_LOCALE, status),
-                                                     strsrch->search->text, strsrch->search->textLength, status);
+            if (strsrch->search->internalBreakIter != nullptr) {
+                ubrk_close(strsrch->search->internalBreakIter);
+                strsrch->search->internalBreakIter = ubrk_open(UBRK_CHARACTER,
```
markusicu, line 982: Why not just nullptr here? Will this BreakIterator be used right away?

icu4c/source/i18n/usearch.cpp
```diff
@@ -1753,21 +1789,26 @@ U_CAPI UBool U_EXPORT2 usearch_search(UStringSearch  *strsrch,
...
-    if (strsrch->pattern.pces == NULL) {
+    if (strsrch->pattern.pces == nullptr) {
         initializePatternPCETable(strsrch, status);
     }
 
     ucol_setOffset(strsrch->textIter, startIdx, status);
     CEIBuffer ceb(strsrch, status);
 
+    // An out-of-memory (OOM) failure can occur in the initializePatternPCETable function
+    // or CEIBuffer constructor, above so we need to check the status.
```
markusicu, line 1805: nit / optional
```suggestion
    // or CEIBuffer constructor above, so we need to check the status.
```

icu4c/source/i18n/usearch.cpp
markusicu: We should probably just delete this line. If this function is called with an incoming error, then it should just return without doing anything, as usual.

markusicu (commented):
otherwise lgtm

markusicu:
@jefgen ping...

> Also, Jeff, it would be nice if you could do the lazy creation in Java as well.

icu4c/source/i18n/usearch.cpp
jefgen: Hm, that's a good point Andy -- thank you.

Before, the `setMatchNotFound` function would have *always* set the state to "no match", but now it will do nothing and simply exit if there is an incoming error. The API docs for `usearch_next`/`usearch_previous` don't specifically call out what happens if an error occurs, but I think we should probably retain the previous behavior, so that the state will be set to "no match" on error.

I was trying to think of ways to add a test case for this, but there doesn't really seem to be any good way to test this. The `setMatchNotFound` is called at the beginning of `usearch_handle*` if there is an error, but these methods are not exported, so the test code can't directly call them. So we'd need to have `usearch_next`/`usearch_previous` somehow fail in the middle of execution, since they also check for failure first before doing anything.
I don't know if we'd want to export the `usearch_handle*` functions just for testing though (?).

I'll leave this w/o testing for now, and update the `setMatchNotFound` to always set "no match" as before.

icu4c/source/i18n/usearch.cpp
jefgen: Oh good point! We can/should lazily create it here as well. Thanks!

icu4c/source/i18n/usearch.cpp
jefgen: thx

jefgen:
> @jefgen ping...
> 
> > Also, Jeff, it would be nice if you could do the lazy creation in Java as well.

Sigh. I don't think I'll be able to get to the lazy creation in Java by the ICU 70 code freeze... I'll file a follow-up ticket for ICU 71 for this.

icu4c/source/i18n/usearch.cpp
markusicu: I still prefer to handle errors consistently -- and bail out early if there is an error.
But I don't feel strongly enough to overrule either of you.
It just seems like every time we make an exception, we will surprise ourselves, and that adds to the maintenance burden.

markusicu:
Hi @aheninger, you "requested changes" last November because of changes in obsolete Boyer Moore code which Jeff has since removed. Is it ok to dismiss the "request changes"?

markusicu:
@jefgen should we try to quickly discuss the question about error behavior? Do you want to just go with what you & Andy decided? Do you want to punt this once more to 71, and add the Java port?

jefgen:
> @jefgen should we try to quickly discuss the question about error behavior? Do you want to just go with what you & Andy decided? Do you want to punt this once more to 71, and add the Java port?

FWIW, I'd sort of lean towards keeping the behavior as-is for the `usearch_*` APIs in this PR (setting "match not found" on error).
We could consider changing the behavior, but I'd prefer to do that as a separate ticket for ICU 71.

markusicu (commented):
lgtm pse squash

jefgen:
> > FWIW, I'd sort of lean towards keeping the behavior as-is for the `usearch_*` APIs in this PR (setting "match not found" on error).
> > We could consider changing the behavior, but I'd prefer to do that as a separate ticket for ICU 71.
> 
> Works for me. Please submit a separate ticket for this, and add Peter, Andy, and myself to the "watchers" list.

Sure thing. I filed a new ticket here: https://unicode-org.atlassian.net/browse/ICU-21761

icu4c/source/i18n/usearch.cpp
jefgen: Filed a follow-up ticket for ICU 71 here: https://unicode-org.atlassian.net/browse/ICU-21761

markusicu (approved):
lgtm tnx

FrankYFTang:
When I try ICU70-RC in chrome many blink unit test break because 
 usearch_next return -127 . I wonder would that be related to this PR.

The breaking code is in line 180 (the DCHECK_EQ code) of https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/editing/iterators/text_searcher_icu.cc;l=180;bpv=1 
```

bool TextSearcherICU::NextMatchResultInternal(MatchResultICU& result) {
  const int match_start = usearch_next(searcher_, &status);
  DCHECK_EQ(status, U_ZERO_ERROR);
```

FrankYFTang:
I got failure in
```
    // Need to create the internal break iterator.
    strsrch->search->internalBreakIter = ubrk_open(UBRK_CHARACTER,
        ucol_getLocaleByType(strsrch->collator, ULOC_VALID_LOCALE, &status),
        strsrch->search->text, strsrch->search->textLength, &status);
```

FrankYFTang:
It try to create a BreakIteratore character instance with locale "en_US@collation=search" and fail with a status = -127 in ICU70-rc.

jefgen:
> usearch_next return -127 

Hm, if I understand correctly, -127 is a *warning*, not a *failure*.

`U_USING_DEFAULT_WARNING   = -127, `

https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/utypes.h#L425

It looks like perhaps the chromium code is checking for exact equal with `U_ZERO_ERROR`.
(The line `DCHECK_EQ(status, U_ZERO_ERROR);`)

> ```
> bool TextSearcherICU::NextMatchResultInternal(MatchResultICU& result) {
>   UErrorCode status = U_ZERO_ERROR;
>   const int match_start = usearch_next(searcher_, &status);
>   DCHECK_EQ(status, U_ZERO_ERROR);
> ```

I think this might be too strict.

Looking at the call to `usearch_open`, it considers `U_USING_DEFAULT_WARNING` to be success:

```c++
  UStringSearch* searcher =
      usearch_open(&kNewlineCharacter, 1, &kNewlineCharacter, 1,
                   search_collator_name.Utf8().c_str(), nullptr, &status);
  DCHECK(status == U_ZERO_ERROR || status == U_USING_FALLBACK_WARNING ||
         status == U_USING_DEFAULT_WARNING)
      << status;
  return searcher;
```

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/editing/iterators/text_searcher_icu.cc;l=53;bpv=1

Since the break iterator gets created lazily now, the result of `U_USING_DEFAULT_WARNING` status is moved from the call to `usearch_open` to `usearch_next`.

FrankYFTang:
looks like icu69-1 also create that break iterator with the same locale with -127

FrankYFTang:
I see. Probably I should change it to   DCHECK(U_SUCCESS(status));

jefgen:
Probably something like this:

```diff
bool TextSearcherICU::NextMatchResultInternal(MatchResultICU& result) {
  UErrorCode status = U_ZERO_ERROR;
  const int match_start = usearch_next(searcher_, &status);
-  DCHECK_EQ(status, U_ZERO_ERROR);
+  DCHECK(status == U_ZERO_ERROR || status == U_USING_FALLBACK_WARNING ||
+         status == U_USING_DEFAULT_WARNING)
+     << status;
```

or maybe just check that U_SUCCESS() is true?

PR #1474, opened 2020-11-18
ICU-21357 Support IEC Binary Prefixes in MeasureUnit

hugovdm:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21357
- [X] Issue accepted
- [X] Tests included

Resolves: https://github.com/icu-units/icu/issues/114

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -41,12 +41,20 @@ namespace {
 // TODO: Propose a new error code for this?
 constexpr UErrorCode kUnitIdentifierSyntaxError = U_ILLEGAL_ARGUMENT_ERROR;
 
-// Trie value offset for SI Prefixes. This is big enough to ensure we only
+// Trie value offset for SI or binary prefixes. This is big enough to ensure we only
 // insert positive integers into the trie.
-constexpr int32_t kSIPrefixOffset = 64;
+constexpr int32_t kPrefixOffset = 64;
+static_assert(kPrefixOffset + UMEASURE_PREFIX_INTERNAL_MIN_BIN > 0,
+              "kPrefixOffset is too small for minimum UMeasurePrefix value");
+static_assert(kPrefixOffset + UMEASURE_PREFIX_INTERNAL_MIN_SI > 0,
+              "kPrefixOffset is too small for minimum UMeasurePrefix value");
 
 // Trie value offset for compound parts, e.g. "-per-", "-", "-and-".
 constexpr int32_t kCompoundPartOffset = 128;
+static_assert(kCompoundPartOffset > kPrefixOffset + UMEASURE_PREFIX_INTERNAL_MAX_BIN,
+              "Ambiguous token values: prefix tokens are overlapping with CompoundPart tokens");
+static_assert(kCompoundPartOffset > kPrefixOffset + UMEASURE_PREFIX_INTERNAL_MAX_SI,
+              "Ambiguous token values: prefix tokens are overlapping with CompoundPart tokens");
 
 enum CompoundPart {
     // Represents "-per-"
@@ -90,30 +98,40 @@ enum PowerPart {
 // "fluid-ounce-imperial".
 constexpr int32_t kSimpleUnitOffset = 512;
 
-const struct SIPrefixStrings {
+const struct UnitPrefixStrings {
     const char* const string;
-    UMeasureSIPrefix value;
-} gSIPrefixStrings[] = {
-    { "yotta", UMEASURE_SI_PREFIX_YOTTA },
-    { "zetta", UMEASURE_SI_PREFIX_ZETTA },
-    { "exa", UMEASURE_SI_PREFIX_EXA },
-    { "peta", UMEASURE_SI_PREFIX_PETA },
-    { "tera", UMEASURE_SI_PREFIX_TERA },
-    { "giga", UMEASURE_SI_PREFIX_GIGA },
-    { "mega", UMEASURE_SI_PREFIX_MEGA },
-    { "kilo", UMEASURE_SI_PREFIX_KILO },
-    { "hecto", UMEASURE_SI_PREFIX_HECTO },
-    { "deka", UMEASURE_SI_PREFIX_DEKA },
-    { "deci", UMEASURE_SI_PREFIX_DECI },
-    { "centi", UMEASURE_SI_PREFIX_CENTI },
-    { "milli", UMEASURE_SI_PREFIX_MILLI },
-    { "micro", UMEASURE_SI_PREFIX_MICRO },
-    { "nano", UMEASURE_SI_PREFIX_NANO },
-    { "pico", UMEASURE_SI_PREFIX_PICO },
-    { "femto", UMEASURE_SI_PREFIX_FEMTO },
-    { "atto", UMEASURE_SI_PREFIX_ATTO },
-    { "zepto", UMEASURE_SI_PREFIX_ZEPTO },
-    { "yocto", UMEASURE_SI_PREFIX_YOCTO },
+    UMeasurePrefix value;
+} gUnitPrefixStrings[] = {
+    // SI prefixes
+    { "yotta", UMEASURE_PREFIX_YOTTA },
+    { "zetta", UMEASURE_PREFIX_ZETTA },
+    { "exa", UMEASURE_PREFIX_EXA },
+    { "peta", UMEASURE_PREFIX_PETA },
+    { "tera", UMEASURE_PREFIX_TERA },
+    { "giga", UMEASURE_PREFIX_GIGA },
+    { "mega", UMEASURE_PREFIX_MEGA },
+    { "kilo", UMEASURE_PREFIX_KILO },
+    { "hecto", UMEASURE_PREFIX_HECTO },
+    { "deka", UMEASURE_PREFIX_DEKA },
+    { "deci", UMEASURE_PREFIX_DECI },
+    { "centi", UMEASURE_PREFIX_CENTI },
+    { "milli", UMEASURE_PREFIX_MILLI },
+    { "micro", UMEASURE_PREFIX_MICRO },
+    { "nano", UMEASURE_PREFIX_NANO },
+    { "pico", UMEASURE_PREFIX_PICO },
+    { "femto", UMEASURE_PREFIX_FEMTO },
+    { "atto", UMEASURE_PREFIX_ATTO },
+    { "zepto", UMEASURE_PREFIX_ZEPTO },
+    { "yocto", UMEASURE_PREFIX_YOCTO },
+    // Binary prefixes
+    { "yobi", UMEASURE_PREFIX_YOBI },
+    { "zebi", UMEASURE_PREFIX_ZEBI },
+    { "exbi", UMEASURE_PREFIX_EXBI },
+    { "pebi", UMEASURE_PREFIX_PEBI },
+    { "tebi", UMEASURE_PREFIX_TEBI },
+    { "gibi", UMEASURE_PREFIX_GIBI },
+    { "mebi", UMEASURE_PREFIX_MEBI },
+    { "kibi", UMEASURE_PREFIX_KIBI },
 };
 
 /**
@@ -221,9 +239,9 @@ void U_CALLCONV initUnitExtras(UErrorCode& status) {
     BytesTrieBuilder b(status);
     if (U_FAILURE(status)) { return; }
 
-    // Add SI prefixes
-    for (const auto& siPrefixInfo : gSIPrefixStrings) {
-        b.add(siPrefixInfo.string, siPrefixInfo.value + kSIPrefixOffset, status);
+    // Add SI and binary prefixes
+    for (const auto& unitPrefixInfo : gUnitPrefixStrings) {
+        b.add(unitPrefixInfo.string, unitPrefixInfo.value + kPrefixOffset, status);
     }
     if (U_FAILURE(status)) { return; }
 
@@ -295,7 +313,7 @@ public:
 
     enum Type {
         TYPE_UNDEFINED,
-        TYPE_SI_PREFIX,
+        TYPE_PREFIX,
         // Token type for "-per-", "-", and "-and-".
         TYPE_COMPOUND_PART,
         // Token type for "per-".
@@ -309,7 +327,7 @@ public:
     Type getType() const {
         U_ASSERT(fMatch > 0);
         if (fMatch < kCompoundPartOffset) {
-            return TYPE_SI_PREFIX;
+            return TYPE_PREFIX;
         }
         if (fMatch < kInitialCompoundPartOffset) {
             return TYPE_COMPOUND_PART;
@@ -323,9 +341,9 @@ public:
         return TYPE_SIMPLE_UNIT;
     }
 
-    UMeasureSIPrefix getSIPrefix() const {
-        U_ASSERT(getType() == TYPE_SI_PREFIX);
-        return static_cast<UMeasureSIPrefix>(fMatch - kSIPrefixOffset);
+    UMeasurePrefix getUnitPrefix() const {
+        U_ASSERT(getType() == TYPE_PREFIX);
+        return static_cast<UMeasurePrefix>(fMatch - kPrefixOffset);
     }
 
     // Valid only for tokens with type TYPE_COMPOUND_PART.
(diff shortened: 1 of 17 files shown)


icu4c/source/i18n/unicode/measunit.h
```diff
@@ -72,159 +72,270 @@ enum UMeasureUnitComplexity {
...
+ * @draft ICU 69
  */
-typedef enum UMeasureSIPrefix {
+typedef enum UMeasurePrefix {
+    /**
+     * The absence of an SI or binary prefix.
+     *
+     * @draft ICU 69
+     */
+    UMEASURE_PREFIX_ONE = 30 + 0, // An arbitrary value
```
hugovdm, line 88: The "+ 0" here intended to remind that this is 10^0.

younies, line 88: Instead of using "30" as an offset for the SI prefixes. I prefer using a constant called for example `SI_PREFIX_OFFSET` which equals 30 or whatever value.

Then use this constant as an offset for all the other SI prefixes.

icu4c/source/i18n/unicode/measunit.h
```diff
@@ -72,159 +72,270 @@ enum UMeasureUnitComplexity {
...
 
     /**
      * SI prefix: yotta, 10^24.
      *
-     * @draft ICU 67
+     * @draft ICU 69
      */
-    UMEASURE_SI_PREFIX_YOTTA = 24,
+    UMEASURE_PREFIX_YOTTA = UMEASURE_PREFIX_ONE + 24,
```
younies, line 95: as suggested use `SI_PREFIX_OFFSET` instead of `UMEASURE_PREFIX_ONE`

icu4c/source/i18n/unicode/measunit.h
```diff
@@ -72,159 +72,270 @@ enum UMeasureUnitComplexity {
...
+    UMEASURE_PREFIX_INTERNAL_MAX_SI = UMEASURE_PREFIX_YOTTA,
 
     /**
      * SI prefix: zetta, 10^21.
      *
-     * @draft ICU 67
+     * @draft ICU 69
      */
-    UMEASURE_SI_PREFIX_ZETTA = 21,
+    UMEASURE_PREFIX_ZETTA = UMEASURE_PREFIX_ONE + 21,
```
younies, line 109: ditto

and for all the other SI prefixes

icu4c/source/i18n/unicode/measunit.h
```diff
@@ -72,159 +72,270 @@ enum UMeasureUnitComplexity {
...
+     * @internal
+     */
+    UMEASURE_PREFIX_INTERNAL_MIN_SI = UMEASURE_PREFIX_YOCTO,
+
+    /**
+     * ICU use only.
+     * Sets the arbitrary offset of the base-1024 binary prefixes' enum values.
+     * @internal
+     */
+    UMEASURE_PREFIX_INTERNAL_ONE_BIN = -60,
```
younies, line 249: Rename to be `BINARY_PREFIX_OFFSET`

OR any other name that means it is a constant and an offset.

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -192,6 +192,14 @@ void UnitsTest::testConverter() {
...
         {"megawatt", "watt", 1.0, 1000000},
         {"megawatt", "kilowatt", 1.0, 1000},
+        // Binary Prefixes
+        {"kilobyte", "byte", 1, 1000},
+        {"kibibyte", "byte", 1, 1024},
+        {"mebibyte", "byte", 1, 1048576},
+        {"gibibyte", "kibibyte", 1, 1048576},
+        {"pebibyte", "tebibyte", 4, 4096},
+        {"zebibyte", "pebibyte", 1.0/16, 65536.0},
+        {"yobibyte", "exbibyte", 1, 1048576},
```
younies, line 202: +1

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
```diff
@@ -117,185 +116,251 @@
...
+    public enum MeasurePrefix {
 
         /**
          * SI prefix: yotta, 10^24.
          *
          * @draft ICU 68
          * @provisional This API might change or be removed in a future release.
          */
-        YOTTA(24, "yotta"),
+        YOTTA(24, "yotta", 10),
```
younies, line 132: I prefer instead of using the number. use a constant. 
Using constant is more illustrative and error-prone. it is easy to miss type a number.

for example: `MeasurePrefix.SI_BASE`

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
```diff
@@ -117,185 +116,251 @@
...
 
         /**
          * SI prefix: zetta, 10^21.
          *
          * @draft ICU 68
          * @provisional This API might change or be removed in a future release.
          */
-        ZETTA(21, "zetta"),
+        ZETTA(21, "zetta", 10),
```
younies, line 140: ditto 

and all other SI prefixes

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
```diff
@@ -117,185 +116,251 @@
...
-        YOCTO(-24, "yocto");
+        YOCTO(-24, "yocto", 10),
+
+        /**
+         * IEC binary prefix: kibi, 1024^1.
+         *
+         * @draft ICU 69
+         * @provisional This API might change or be removed in a future release.
+         */
+        KIBI(1, "kibi", 1024),
```
younies, line 300: I prefer instead of using the number. use a constant. 
Using constant is more illustrative and error-prone. it is easy to miss type a number.

for example: `MeasurePrefix.BINARY_BASE`

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/MeasureUnitTest.java
```diff
@@ -3481,27 +3481,61 @@ public void TestIdentifiers() {
...
+            new TestCase("yottafoot", "yottafoot"),
+            new TestCase("kibibyte", "kibibyte"),
+            new TestCase("mebibyte", "mebibyte"),
+            new TestCase("gibibyte", "gibibyte"),
+            new TestCase("tebibyte", "tebibyte"),
+            new TestCase("pebibyte", "pebibyte"),
+            new TestCase("exbibyte", "exbibyte"),
+            new TestCase("zebibyte", "zebibyte"),
+            new TestCase("yobibyte", "yobibyte"),
+
```
younies, line 3531: +1

younies (dismissed):

icu4c/source/i18n/unicode/measunit.h
hugovdm: This was the name that was suggested in the meeting, and approved. I believe I must not change it without another ICU-TC review.

icu4c/source/i18n/unicode/measunit.h
hugovdm: I like limiting how much cruft is present in a public header file. Another constant means another exported symbol accessible to clients. This was discussed in the ICU-TC meeting, so reviewed and approved by the committee. I believe further changes  (to the public header file) would need another review.

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
hugovdm: This is also a public class - though, Java at least has private members. I hear what you're saying, but again: committee review. (At least I think it's harder to "mistype" numbers like 10 and 1024, it's tested in the unit tests, and client code doesn't need to be typing these numbers...)

icu4c/source/i18n/unicode/measunit.h
hugovdm: UMEASURE_PREFIX_ is the symbol prefix that indicates this is an enum value of the UMeasurePrefix type. This is the style of the public ICU4C APIs - partially also because C does not have namespaces, so public symbols need to be very clear. Again: this is the public API that was approved by the committee.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -631,7 +650,27 @@ compareSingleUnits(const void* /*context*/, const void* left, const void* right)
 
 } // namespace
 
+int32_t umeas_getPrefixPower(UMeasurePrefix unitPrefix) {
```
sffc, line 653: Please include the C API boilerplate.

```suggestion
U_CAPI int32_t U_EXPORT2
umeas_getPrefixPower(UMeasurePrefix unitPrefix) {
```

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -631,7 +650,27 @@ compareSingleUnits(const void* /*context*/, const void* left, const void* right)
...
+    if (unitPrefix >= UMEASURE_PREFIX_INTERNAL_MIN_BIN &&
+        unitPrefix <= UMEASURE_PREFIX_INTERNAL_MAX_BIN) {
+        return unitPrefix - UMEASURE_PREFIX_INTERNAL_ONE_BIN;
+    }
+    U_ASSERT(unitPrefix >= UMEASURE_PREFIX_INTERNAL_MIN_SI &&
+             unitPrefix <= UMEASURE_PREFIX_INTERNAL_MAX_SI);
+    return unitPrefix - UMEASURE_PREFIX_ONE;
+}
 
+int32_t umeas_getPrefixBase(UMeasurePrefix unitPrefix) {
```
sffc, line 663:
```suggestion
U_CAPI int32_t U_EXPORT2
umeas_getPrefixBase(UMeasurePrefix unitPrefix) {
```

icu4c/source/i18n/unicode/measunit.h
```diff
@@ -72,159 +72,270 @@ enum UMeasureUnitComplexity {
 };
 
 /**
- * Enumeration for SI prefixes, such as "kilo".
+ * Enumeration for SI and binary prefixes, e.g. "kilo-", "nano-", "mebi".
```
sffc, line 75: You have a `-` after kilo and nano but not mebi

icu4c/source/i18n/unicode/measunit.h
sffc: For public use we want people to use the new C functions umeas_getPrefixPower and umeas_getPrefixBase.  I don't feel strongly about whether to introduce `UPRV_SI_PREFIX_OFFSET` as an internal thing.  If we did, we could easily label it `@internal` to avoid an additional API review.

icu4c/source/i18n/unicode/measunit.h
sffc: See above for comment about `UPRV_SI_PREFIX_OFFSET`

icu4c/source/i18n/unicode/measunit.h
sffc: See above for comment about `UPRV_SI_PREFIX_OFFSET`

icu4c/source/i18n/units_converter.cpp
```diff
@@ -293,13 +300,16 @@ void loadConversionRate(ConversionRate &conversionRate, const MeasureUnitImpl &s
...
+    // This code corresponds to ICU4J's ConversionRates.getOffset().
     // In case of simple units (such as: celsius or fahrenheit), offsets are considered.
     if (checkSimpleUnit(source, status) && checkSimpleUnit(target, status)) {
         conversionRate.sourceOffset =
             sourceToBase.offset * sourceToBase.factorDen / sourceToBase.factorNum;
         conversionRate.targetOffset =
             targetToBase.offset * targetToBase.factorDen / targetToBase.factorNum;
     }
+    // TODO: should we consider failure if there's an offset for a
+    // not-simple-unit? What about kilokelvin / kilocelsius?
```
sffc, line 312: kilokelvin is fine; it could be used to describe temperatures at the surface of the sun, for example.  I don't think I've ever seen a kilocelsius or even have a guess about what that could mean.

We should, at least, have a test case that demonstrates what our behavior is in edge cases like this.  Follow-up issue?

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
sffc: I'm in favor of just writing 10 here on the Java side.  This code will never need to be changed, and we are doing a review right now to make sure it is correct.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
```diff
@@ -211,20 +211,24 @@ private void multiply(BigDecimal value, int power) {
...
+        /** Apply SI or binary prefix to the Factor. */
+        public Factor applyPrefix(MeasureUnit.MeasurePrefix unitPrefix) {
             Factor result = this.copy();
-            if (siPrefix == MeasureUnit.SIPrefix.ONE) {
+            if (unitPrefix == MeasureUnit.MeasurePrefix.ONE) {
                 return result;
             }
 
-            BigDecimal siApplied = BigDecimal.valueOf(Math.pow(10.0, Math.abs(siPrefix.getPower())));
+            Integer base = unitPrefix.getBase();
```
sffc, line 221: Why capital-I `Integer` rather than `int`?

sffc, line 221: Why capital-I `Integer` rather than `int`?

sffc (commented):
LGTM from me with some optional small comments

icu4j/main/classes/core/src/com/ibm/icu/util/MeasureUnit.java
hugovdm: thanks. I'm leaving this as just "10" and "1024", trusting conversion unit tests, prefix unit tests, and code review to ensure they're right.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitConverter.java
hugovdm: Heh, "I have no idea": I can't remember what inspired this.

(I find one for-loop using "Integer" in this file - should I change that to "int" too? I don't think that one was my inspiration though!)

hugovdm: Dupe.

icu4c/source/i18n/units_converter.cpp
hugovdm: I added a ticket for tracking handling of "undefined behaviour" - https://github.com/icu-units/icu/issues/127 - additional items can be added to it in the future.

icu4c/source/i18n/unicode/measunit.h
hugovdm: For now, I'm leaving this as just UMEASURE_PREFIX_ONE - there's only one copy of "30", so it's really just a question of whether this constant needs an alias. An alias would make it clearer where the arbitrary value is used (here and umeas_getPrefixPower), but I think the number of references to UMEASURE_PREFIX_ONE is not too many.

I did look at how vscode handles documentation for the symbol, and decided thereby that the "arbitrary value" comment should rather be combined with the main documentation for this symbol: I went adhead and did this.

icu4c/source/i18n/unicode/measunit.h
hugovdm: Done.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Done though I do suspect the C APIs won't be very useful, given that there's no C API that's actually making use of this enum? Leaving unresolved...

sffc (dismissed):
LGTM please squash (and check if younies has any more comments)

younies (approved):

hugovdm:
FYI: github-suggestions are squashbot safe, because the author is still the PR author. Example: fbb1ad2a96523a78522ce86811a25d0325474166
Shows also how co-authors can be indicated:
Co-authored-by: Shane F. Carr <shane@unicode.org>

I'll drop a note of this on the squashbot issue tracker.

younies (approved):

PR #1475, opened 2020-11-18
ICU-21402 replace sd and rg by subdivisionAlias

FrankYFTang:
Only replace if the replacement value fit alphanum{3,8} for now.
replacement value length of 2 are pending on resolution of
https://unicode-org.atlassian.net/browse/CLDR-14312

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21402
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -627,6 +627,17 @@ private:
                           LocalMemory<const char*>& types,
                           LocalMemory<int32_t>& replacementIndexes,
                           int32_t &length, UErrorCode &status);
+
+    // Read the subdivisionAlias data from alias to
+    // strings+types+replacementIndexes
+    // Allocate length items for types, to store the type field.
+    // Allocate length items for replacementIndexes,
+    // to store the index in the strings for the replacement variant.
+    void readSubdivisionAlias(UResourceBundle* alias,
+                          UniqueCharStrings* strings,
+                          LocalMemory<const char*>& types,
+                          LocalMemory<int32_t>& replacementIndexes,
+                          int32_t &length, UErrorCode &status);
 };
 
 /**
@@ -647,6 +658,7 @@ public:
     const CharStringMap& scriptMap() const { return script; }
     const CharStringMap& territoryMap() const { return territory; }
     const CharStringMap& variantMap() const { return variant; }
+    const CharStringMap& subdivisionMap() const { return subdivision; }
 
     static void U_CALLCONV loadData(UErrorCode &status);
     static UBool U_CALLCONV cleanup();
@@ -658,11 +670,13 @@ private:
               CharStringMap scriptMap,
               CharStringMap territoryMap,
               CharStringMap variantMap,
+              CharStringMap subdivisionMap,
               CharString* strings)
         : language(std::move(languageMap)),
           script(std::move(scriptMap)),
           territory(std::move(territoryMap)),
           variant(std::move(variantMap)),
+          subdivision(std::move(subdivisionMap)),
           strings(strings) {
     }
 
@@ -676,6 +690,7 @@ private:
     CharStringMap script;
     CharStringMap territory;
     CharStringMap variant;
+    CharStringMap subdivision;
     CharString* strings;
 
     friend class AliasDataBuilder;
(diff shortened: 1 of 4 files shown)


icu4c/source/common/locid.cpp
```diff
@@ -1559,11 +1646,40 @@ AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
...
+        changed = 0;
+        Locale temp(locale);
+        LocalPointer<icu::StringEnumeration> iter(locale.createKeywords(status));
+        if (U_SUCCESS(status) && !iter.isNull()) {
+            const char* key;
+            while ((key = iter->next(nullptr, status)) != nullptr) {
+                status = U_ZERO_ERROR;
+                CharString value;
+                CharStringByteSink valueSink(&value);
+                locale.getKeywordValue(key, valueSink, status);
```
markusicu, line 1667: I propose that you get the keyword value only if the key is one of the two where we might replace the value.

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1268,12 +1268,31 @@ public String replace() {
...
                     ((variants == null) ? "" : Utility.joinStrings("_", variants)));
-                if (extensions != null) {
-                    result += extensions;
+            if (extensions != null) {
+                boolean keywordChanged = false;
+                ULocale temp = new ULocale(result + extensions);
+                Iterator<String> keywords = temp.getKeywords();
+                while (keywords != null && keywords.hasNext()) {
+                    String key = keywords.next();
+                    String value = temp.getKeywordValue(key);
```
markusicu, line 1282: As in C++, I propose fetching the value only if the key is one of the two relevant ones.

FrankYFTang:
PTAL

icu4c/source/common/locid.cpp
```diff
@@ -1559,11 +1646,40 @@ AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
...
+                    CharStringByteSink valueSink(&value);
+                    locale.getKeywordValue(key, valueSink, status);
+                    if (U_FAILURE(status)) {
+                        status = U_ZERO_ERROR;
+                        continue;
+                    }
+                    if (replaceSubdivision(value, status)) {
+                        changed++;
+                    }
+                    temp.setKeywordValue(key, value.data(), status);
```
markusicu, line 1676: Sorry, found a problem: if U_FAILURE after this line, then the loop just continues. Worst case, you get to the second relevant key and clobber the status back to a zero-error.

Please bail out of the loop after this line if there is a failure code. I am not looking at enough context to see if you just need to "break" or if you should return immediately with something.

FrankYFTang:
PTAL

icu4c/source/common/locid.cpp
```diff
@@ -1559,11 +1646,43 @@ AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
...
+    }
     if (extensionsStr != nullptr) {
+        changed = 0;
+        Locale temp(locale);
+        LocalPointer<icu::StringEnumeration> iter(locale.createKeywords(status));
+        if (U_SUCCESS(status) && !iter.isNull()) {
+            const char* key;
+            while ((key = iter->next(nullptr, status)) != nullptr) {
+                if (uprv_strcmp("sd", key) == 0 || uprv_strcmp("rg", key) == 0) {
+                    status = U_ZERO_ERROR;
```
markusicu, line 1665: Is this actually necessary? I would remove this line.

icu4c/source/common/locid.cpp
```diff
@@ -1559,11 +1646,43 @@ AliasReplacer::replace(const Locale& locale, CharString& out, UErrorCode status)
...
+                    if (U_FAILURE(status)) {
+                        status = U_ZERO_ERROR;
+                        continue;
+                    }
+                    if (replaceSubdivision(value, status)) {
+                        changed++;
+                    }
+                    temp.setKeywordValue(key, value.data(), status);
+                    if (U_FAILURE(status)) {
+                      return false;
```
markusicu, line 1678:
```suggestion
                        return false;
```

markusicu (commented):
1 or 2 minor fixes, then please squash

FrankYFTang:
PTAL

markusicu (approved):
lgtm

PR #1477, opened 2020-11-19
ICU-21381 AppVeyor CI Builds: Speed up the Cygwin setup download.

jefgen:
While working on other things, I noticed that the AppVeyor CI setup uses the `Invoke-WebRequest` method to download the Cygwin installer setup program.

Based on experimentation and reading other reports online, it turns out that this is _really_ very slow way to download things. It takes several minutes+ to download files that are only a few megabytes in size. If we use `System.Net.WebClient.DownloadFile` method instead then the download only takes _seconds_.

Changing this should help a bit to speed up the AppVeyor builds.
They have been timing out and failing recently as they are now regularly running up against the 60 minute hard limit on AppVeyor.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21381
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,7 +4,7 @@ image: Visual Studio 2017
 platform: x64
 
 # Don't clone the entire repo.
-clone_depth: 10
+clone_depth: 5
 
 # Cache things between builds to speed things up and save bandwidth.
 cache:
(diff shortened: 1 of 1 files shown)


axelandrejs (approved):

.appveyor.yml
```diff
@@ -4,7 +4,7 @@ image: Visual Studio 2017
 platform: x64
 
 # Don't clone the entire repo.
-clone_depth: 10
+clone_depth: 5
```
erik0686, line 7: Out of curiosity... why reduce the depth?

erik0686 (approved):

.appveyor.yml
jefgen: Good question. I was trying to think of any other ways to help speed up the AppVeyor builds -- since they are now failing due to timeouts on a semi-regular basis. :(
By slight reducing the clone depth we can save a bit more time as the build doesn't need to clone down as much data.

erik0686: And does the appveyor checks need a certain depth? Like, why stop at 5?

jefgen: The number is actually somewhat arbitrary. Most of the time we only need the top most commit (ie: a depth of 1). The only time that we need to sync down more commits is if somebody wants to do a "re-build" on older commits. For example, if a build times out or fails, and somebody wants to kick off a rebuild -- then we'd need to sync down enough history to "reach" the older commit to rebuild.

erik0686: Ahhh I see, that's good to know 👍  Thank you Jeff :D

daniel-ju (approved):

.appveyor.yml
jefgen: No problem! :)

PR #1483, opened 2020-11-24
ICU-21306 Pull CLDR 38's unit prefixes into ICU

hugovdm:
"ant build-icu-data" was run with cldr and cldr-staging repositories at tag `release-38`.

* This PR has only prefixes.

* Powers need inflections: we don't have any examples of gender and case yet, so this change needs some thought and care. (Are there examples of other features / XML attributes only present in some languages?)

* Versions:
  * With my first version, I added /prefix/ to the path: 0783db5d9ae04c7e54f18ce70347bd4880835b27
  * With my second version, I dropping the prefix/, such that all sub-entities of compoundUnit entities would be siblings in the ICU resource file: 2de968ecb265ec90486f00c7460ba49cd9565dd9 - trusting that each compoundUnit entry in CLDR XML will have only one of compoundUnitPattern, unitPrefixPattern, or compoundUnitPattern1.
  * With my third version, I dropped the powers, keeping only the prefixes (83e68274a9).

* Find examples of compoundUnit entries before and after line 6045: https://github.com/unicode-org/cldr/blob/4b1a7f5fc7c6a15840eb262b64845081e2918e8f/common/main/en.xml#L6045

* Spec: https://unicode-org.github.io/cldr/ldml/tr35-general.html#Unit_Elements

Tracking issue: icu-units#48

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21306
- [x] Issue accepted
- [ ] Tests included


sffc (commented):
LGTM please squash

sffc (dismissed):

younies (approved):

hugovdm:
Not supporting "case" for compoundUnitPattern1 yet. Why this feels wrong:
the format might change, so code developed with the current data might need
to change.

I think I should either entirely remove the powers from this PR (keeping
only prefixes, which is what I need first), so I can show Younies how this
is done later, or I should simply include it within this PR already - which
would then probably need a bit more discussion.

Including it in this PR is probably the fastest / most efficient solution -
if we go for that: Younies, I'll be sure to involve you in the next
ldml2icu change?

younies (approved):

hugovdm:
I'm merging this one: it has only prefixes now. In a follow-up for adding compound patterns with inflections, we can also revisit the existing rules.

PR #1486, opened 2020-11-30
ICU-21349 Make appendSingleUnit behaviour in C++ comply with Java

younies:
##### Checklist

- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -380,7 +380,53 @@ public:
 
     MeasureUnitImpl parse(UErrorCode& status) {
         MeasureUnitImpl result;
-        parseImpl(result, status);
+
+        if (U_FAILURE(status)) {
+            return result;
+        }
+        if (fSource.empty()) {
+            // The dimenionless unit: nothing to parse. leave result as is.
+            return result;
+        }
+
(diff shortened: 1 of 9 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -821,14 +791,51 @@ MeasureUnitImpl MeasureUnitImpl::forMeasureUnitMaybeCopy(
...
+
+    if (oldUnit) {
+        // Both dimensionalities will be positive, or both will be negative, by
+        // virtue of isCompatibleWith().
+        oldUnit->dimensionality += singleUnit.dimensionality;
+
+        return false;
+    }
+
+    // TODO: question, shall we add a copy of singleUnit. I think we should, because the added singleUnit
```
sffc, line 824: SingleUnit is cheap to copy (no memory allocation), so yes, we should copy it.

sffc (commented):
LGTM pending Hugo's review

icu4c/source/i18n/measunit_extra.cpp
younies: done. Actually, it is copied already and we do not need the copy() function as in Java.

sffc (approved):

PR #1488, opened 2020-11-30
ICU-21243 Migrates preparseucd.py script to Python 3. Python 3 changes

gnrunge:
the order of elements in an iterator from Python 2 with the result
that the generated data in ppucd.txt changes with respect to the selection
of a property value used to compact the output when there is a
property with equal count of the two most frequent values. This
change doesn't change the validity of the generated ppucd.txt file.

While at it, also migrated script parsescriptmetadata.py to Python 3.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21243
- [ ] Issue accepted
- [ ] Tests included


markusicu (approved):
lgtm tnx!

The ppucd.txt changes look annoying but ok.
I didn't try it out for myself. I will see how it goes when I do the next real Unicode update.

PR #1489, opened 2020-11-30
ICU-21118 check that dst and src are not null in uprv_memcpy

erik0686:
It was pointed out that uprv_memcpy() and uprv_memmove() internally use memcpy and memmove respectively, and these do not allow dst or src to be null. In this PR, I am adding a check for Debug mode for these two functions to make sure that doesn't happen.

After building ICU on Linux there was a case where adding this check could bring errors on the tests, so I am adding a check to uloc_tag.cpp as well.

After this, I built again on Linux and Windows and found no more issues. And all tests pass successfully :D 

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21118
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -31,14 +31,25 @@
 #include <stddef.h>
 #include <string.h>
 #include "unicode/localpointer.h"
+#include "uassert.h"
 
 #if U_DEBUG && defined(UPRV_MALLOC_COUNT)
 #include <stdio.h>
 #endif
 
 
-#define uprv_memcpy(dst, src, size) U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size)
-#define uprv_memmove(dst, src, size) U_STANDARD_CPP_NAMESPACE memmove(dst, src, size)
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size); \
+} UPRV_BLOCK_MACRO_END
+
+#define uprv_memmove(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    U_STANDARD_CPP_NAMESPACE memmove(dst, src, size); \
+} UPRV_BLOCK_MACRO_END
+
 
 /**
  * \def UPRV_LENGTHOF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -2049,7 +2049,10 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
         *status = U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
-    uprv_memcpy(tagBuf, tag, tagLen);
+    
+    if (tagLen > 0) {
+        uprv_memcpy(tagBuf, tag, tagLen);
+    }
     *(tagBuf + tagLen) = 0;
 
     /* create a ULanguageTag */


icu4c/source/common/uloc_tag.cpp
```diff
@@ -2049,7 +2049,9 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
         *status = U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
-    uprv_memcpy(tagBuf, tag, tagLen);
+    if (tagLen > 0 && tag != NULL) {
```
jefgen, line 2052: nit: blank line between successive if statements.

jefgen, line 2052: Since this file is .cpp, let's use `nullptr` here instead of `NULL`.

jefgen (changes requested):
Thanks Erik! A few minor nits... :)

icu4c/source/common/cmemory.h
```diff
@@ -31,14 +31,25 @@
...
 
 #if U_DEBUG && defined(UPRV_MALLOC_COUNT)
 #include <stdio.h>
 #endif
 
 
-#define uprv_memcpy(dst, src, size) U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size)
-#define uprv_memmove(dst, src, size) U_STANDARD_CPP_NAMESPACE memmove(dst, src, size)
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    U_ASSERT(dst != NULL); \
```
jefgen, line 42: Just as comment/thought: since this header is used by plain .c files, we can't use `nullptr` here unless we wanted to fork the code with something like `#ifdef __cplusplus`. But it's probably cleaner to just use `NULL` here rather than adding an ifdef, but I don't really have a strong opinion on it either way.

jefgen (approved):
lgtm, but please wait for Markus to review as well. Thanks!

icu4c/source/common/cmemory.h
markusicu: I am not sure how much C code still uses the internal header files... probably cintltst. And maybe some tools code, although we could (at some point) switch those all to C++ and to using the standard library.

But ok for here & now.

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2049,7 +2049,10 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
         *status = U_MEMORY_ALLOCATION_ERROR;
         return NULL;
     }
-    uprv_memcpy(tagBuf, tag, tagLen);
+    
+    if (tagLen > 0 && tag != nullptr) {
```
markusicu, line 2053:
```suggestion
    if (tagLen > 0) {
```
This should do. If we still get a nullptr, we want to know.

markusicu (commented):
mostly lgtm

markusicu (approved):
lgtm tnx

jefgen (approved):
lgtm, thanks @erik0686!

tedyu:
I checked release 68.2 but I didn't find this fix among the JIRAs.

which release would contain this fix ? If none, when would the next release be made ?

thanks

jefgen:
> I checked release 68.2 but I didn't find this fix among the JIRAs.
> which release would contain this fix ? If none, when would the next release be made ?

This PR was merged after the 68.2 release, so this change wouldn't be included as part of it.
The most current release is 68.2, so that would mean that there actually isn't any releases out with this change.
The next release would be 69.1, which I think would be sometime i
n March/April.

tedyu:
Thanks for the response.

I will watch for the next release.

jefgen:
FWIW, you could possibly consider cherry-picking this change onto `maint/maint-68` if you need the change sooner.

The change in `icu4c/source/common/cmemory.h` is a no-op for release builds; the only real change is the 3 line change in `icu4c/source/common/uloc_tag.cpp`. :)

tedyu:
Assuming 68.3 release would come sooner than 69.1, it is nice to cherry-pick the fix.

We have been experiencing the issue (using 67.1) for quite a while.

Thanks

PR #1495, opened 2020-12-02
ICU-21405 BRS 68.2 Version update and regenerate configure for v68.2

erik0686:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21405
- [X] Issue accepted
- [ ] Tests included


icu4c/readme.html
```diff
@@ -3,7 +3,7 @@
 
 <html lang="en-US" xmlns="http://www.w3.org/1999/xhtml" xml:lang="en-US">
   <head>
-    <title>ReadMe for ICU 68.1</title>
```
jefgen, line 6: Thanks Erik!
Would you mind also updating the ICU4J readme file too?
https://github.com/unicode-org/icu/blob/master/icu4j/readme.html

echeran: Thanks, Erik! That covers the ICU4C Readme version update for me. I created a PR for the ICU4J Readme version update in #1497.

erik0686: Oh thank you very much, Elango!

daniel-ju (approved):

icu4c/readme.html
jefgen: Thanks Elango!

jefgen (approved):
Thanks!

echeran (approved):
LGTM.

The PR seems to be stalled on the PR check for the Jira issue being accepted, but now the Jira issue is accepted, but the PR check hasn't changed. Does that mean that we need to re-trigger the PR checks?  My PR for Readme version number update is stalled on the PR check for signing the license/CLA (even stranger), so I may have to replicate the same fix behavior required here.

I also wonder if this stalled PR check behavior is related to the Travis PR checks being stalled.

jefgen:
> The PR seems to be stalled on the PR check for the Jira issue being accepted, but now the Jira issue is accepted, but the PR check hasn't changed. Does that mean that we need to re-trigger the PR checks? My PR for Readme version number update is stalled on the PR check for signing the license/CLA (even stranger), so I may have to replicate the same fix behavior required here.
> 
> I also wonder if this stalled PR check behavior is related to the Travis PR checks being stalled.

I think the Jira issue checker was maybe just out-of-date and/or stuck somehow. However, you can get it to re-check the Jira ticket status by clicking on the "Details" link on the right and then using the "Update GitHub Status" button on the separate webpage.
I did this just now and it looks like it is now updated. 

However, for the Travis CI check, it seems like it might have been waiting in a "queued" state for some time and it just started actually running the builds now. It seems that Travis keeps getting more and more slow and/or throttled down. :(

PR #1496, opened 2020-12-02
ICU-21349 Adjust C++ `MeasureUnitImpl::serialize` to be as same as the Java one

younies:
Closes: https://github.com/icu-units/icu/issues/79

##### Checklist

- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/common/charstr.cpp
+++ b/icu4c/source/common/charstr.cpp
@@ -141,6 +141,38 @@ CharString &CharString::append(const char *s, int32_t sLength, UErrorCode &error
     return *this;
 }
 
+CharString &CharString::appendNumber(int32_t number, UErrorCode &status) {
+    if (number < 0) {
+        this->append('-', status);
+        if (U_FAILURE(status)) {
+            return *this;
+        }
+    }
+
+    if (number == 0) {
+        this->append('0', status);
+        return *this;
+    }
+
+    int32_t numLen = 0;
+    while (number != 0) {
+        int32_t residue = number % 10;
+        number /= 10;
+        this->append(std::abs(residue) + '0', status);
+        numLen++;
+        if (U_FAILURE(status)) {
+            return *this;
+        }
+    }
+
+    int32_t start = this->length() - numLen, end = this->length() - 1;
+    while(start < end) {
+        std::swap(this->data()[start++], this->data()[end--]);
+    }
+
+    return *this;
+}
+
 char *CharString::getAppendBuffer(int32_t minCapacity,
                                   int32_t desiredCapacityHint,
                                   int32_t &resultCapacity,
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -30,6 +30,7 @@
 #include "unicode/ures.h"
 #include "unicode/ustringtrie.h"
 #include "uresimp.h"
+#include "util.h"
 #include <cstdlib>
 
 U_NAMESPACE_BEGIN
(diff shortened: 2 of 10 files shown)


icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -758,6 +657,43 @@ const char *SingleUnitImpl::getSimpleUnitID() const {
...
+    if (absPower == 1) {
+        // no-op
+    } else if (absPower == 2) {
+        result.append(StringPiece("square-"), status);
+    } else if (absPower == 3) {
+        result.append(StringPiece("cubic-"), status);
+    } else if (absPower <= 15) {
+        result.append(StringPiece("pow"), status);
+        // TODO: could we use std::to_string
+        result.append(StringPiece(std::to_string(absPower)), status);
```
sffc, line 674: Don't use std::to_string.  Use ICU_Utility::appendNumber instead (feel free to add a version of that helper function that works on 8-bit strings), or carry over the `absPower + '0'` logic from before.

sffc (changes requested):

icu4c/source/i18n/measunit_extra.cpp
younies: Okay, Now I have added an `appendNumber` function for ICU_Utility and for `CharString`, I prefer the one in the CharString.

Choose one in order to finalise the PR :)

icu4c/source/common/charstr.cpp
```diff
@@ -141,6 +141,41 @@ CharString &CharString::append(const char *s, int32_t sLength, UErrorCode &error
...
+    while (number > 0) {
+        int32_t residue = number % 10;
+        number /= 10;
+        numberStr.append(residue + '0', status);
+        if (U_FAILURE(status)) {
+            return *this;
+        }
+    }
+
+    char *clonedNumber = numberStr.cloneData(status);
```
sffc, line 168: You should append directly to `this` rather than creating this completely unnecessary and unsafe temporary variable.

icu4c/source/common/util.cpp
```diff
@@ -65,6 +65,39 @@ UnicodeString& ICU_Utility::appendNumber(UnicodeString& result, int32_t n,
     return result;
 }
 
+void ICU_Utility::appendNumber(CharString &word, int32_t number, UErrorCode &status) {
```
sffc, line 68: Put this logic in one place or the other.  Don't duplicate it

sffc (changes requested):

icu4c/source/common/util.cpp
younies: As I mentioned in another comment, I made both of places for the discussion.

Removed!

icu4c/source/common/charstr.cpp
younies: Done, I flipped the tail of the word.

icu4c/source/common/charstr.cpp
```diff
@@ -141,6 +141,39 @@ CharString &CharString::append(const char *s, int32_t sLength, UErrorCode &error
     return *this;
 }
 
+CharString &CharString::appendNumber(int32_t number, UErrorCode &status) {
```
sffc, line 144: Can you add a unit tests for this function please?  The call site only ever formats numbers between 1 and 15.  I want to make sure the logic is correct for larger numbers.  You can add the test wherever CharString is tested.

sffc (changes requested):

icu4c/source/common/charstr.cpp
```diff
@@ -141,6 +141,38 @@ CharString &CharString::append(const char *s, int32_t sLength, UErrorCode &error
...
+        }
+    }
+
+    if (number == 0) {
+        this->append('0', status);
+        return *this;
+    }
+
+    int32_t numLen = 0;
+    while (number != 0) {
```
aheninger, line 158: If you change this to a do ... while() loop, so that it always runs at least once, you can get rid of the separate handling of zero, above.

Should number be an int64_t, just in case we want it some day?

Is itoa() safe to use from our code? It seems a shame to have to reinvent the wheel. It's not officially standard, so I don't know.

icu4c/source/common/charstr.cpp
```diff
@@ -141,6 +141,38 @@ CharString &CharString::append(const char *s, int32_t sLength, UErrorCode &error
     return *this;
 }
 
+CharString &CharString::appendNumber(int32_t number, UErrorCode &status) {
```
aheninger, line 144: It would be good to ask Markus to review this addition to CharString, if he hasn't already. He has opinions on how it should evolve.

sffc (approved):
LGTM, but you can wait for @markusicu's review as @aheninger suggested

sffc (approved):

hugovdm:
Assigned during the ICU-TC meeting, merged after the meeting: I changed assignment to sffc, since this is already merged with his approval.

icu4c/source/i18n/measunit_impl.h
```diff
@@ -42,6 +42,12 @@ struct U_I18N_API SingleUnitImpl : public UMemory {
      */
     const char *getSimpleUnitID() const;
 
+    /**
+     * Generates and append a neutral identifier string for a single unit which means we do not include
+     * the dimension signal.
```
hugovdm, line 47: What is a "dimension signal"? This sentence doesn't clarify much for me - I don't know what it's appending without reading the code, and I also need to notice that it's appended to the first parameter (not to *this).

In the code I see this:
int32_t absPower = std::abs(this->dimensionality);
I see dimensionality _is_ included in the output - it's only whether the dimensionality is positive or negative that's skipped - i.e. the sign / "signum" is skipped.

PR #1498, opened 2020-12-03
ICU-21425 build: updates to maven/publication

srl295:
ICU-21425

- use SPDX license rather than raw license file
- minor srl update
- upload a full pom.xml instead of a generated one

NOTE: this is on the head fork, because it tests workflow stuff.

### DIFF
--- a/icu4j/maven/icu4j-localespi/pom.xml
+++ b/icu4j/maven/icu4j-localespi/pom.xml
@@ -20,8 +20,9 @@
 
   <licenses>
     <license>
-      <name>Unicode/ICU License</name>
-      <url>https://raw.githubusercontent.com/unicode-org/icu/master/icu4c/LICENSE</url>
+      <name>Unicode-DFS-2016</name>
+      <url>https://www.unicode.org/copyright.html</url>
+      <comment>Unicode License Agreement - Data Files and Software (2016)</comment>
       <distribution>repo</distribution>
     </license>
   </licenses>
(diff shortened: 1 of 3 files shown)


srl295:
grr, -DpomFile=… did not seem to work in the maven workflow.

Probably the only way forward on that would be to somehow use these `icu4j/maven/icu4j/pom.xml` files to actually drive continuous icu4j build/distribution.  This is what ICU-12811 is about.

So, i'll suggest this as a minor update to improve the licensing structure.

icu4j/maven/icu4j-localespi/pom.xml
```diff
@@ -20,8 +20,9 @@
 
   <licenses>
     <license>
-      <name>Unicode/ICU License</name>
-      <url>https://raw.githubusercontent.com/unicode-org/icu/master/icu4c/LICENSE</url>
+      <name>Unicode-DFS-2016</name>
+      <url>https://www.unicode.org/copyright.html</url>
```
srl295, line 24: Hmm. This doesn't include the 3rd party licenses (thai dictionary etc). So perhaps this is not a good change.

For CLDR, I'm using a plugin to collect 3rd party licenses: https://github.com/unicode-org/cldr/pull/819/files#diff-47e6781b86a294ebbe230cf4c1399c49c7cbd4a8ca8a8cdf12a66e62fdcf4784R86-R91

srl295: ^ probably the 3rd party licenses would need to come into Maven format.

PR #1438, opened 2020-10-29
ICU-21363 Add support for IBM's XL compiler

datalogics-robb:
IBM's XL compiler can only support c++11 when the 'clang front end' is used.
This change modifies the configure step to use xlclang and  xlclang++

##### Checklist

- [*] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21363
- [*] Updated PR title and link in previous line to include Issue number
- [ ] Issue accepted
- [ ] Tests included (not applicable)
- [*] Documentation is changed or added

PR #1499, opened 2020-12-04
ICU-21363 Resurrect support for IBM's XL compiler

yumaoka:
Yoshito is making this change on behalf of @datalogics-robb.
IBM's XL compiler can only support c++11 when the 'clang front end' is used.
This change modifies the configure step to use xlclang and  xlclang++
Change readme.html reference to xlC to refer to xlclang++ instead.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21363
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/runConfigureICU
+++ b/icu4c/source/runConfigureICU
@@ -36,7 +36,7 @@ the runConfigureICU command:
 
 The following names can be supplied as the argument for platform:
 
-    AIX                 Use the IBM Visual Age xlc_r/xlC_r compilers on AIX
+    AIX                 Use the IBM XL xlclang/xlclang compilers on AIX
     AIX/GCC             Use the GNU gcc/g++ compilers on AIX
     Cygwin              Use the GNU gcc/g++ compilers on Cygwin
     Cygwin/MSVC         Use the Microsoft Visual C++ compiler on Cygwin
@@ -49,7 +49,7 @@ The following names can be supplied as the argument for platform:
     Linux/gcc           Use the GNU gcc/g++ compilers on Linux
     Linux/ECC           Use the Intel ECC compiler on Linux
     Linux/ICC           Use the Intel ICC compiler on Linux
-    Linux/VA            Use the IBM Visual Age compiler on Power PC Linux
+    Linux/VA            Use the IBM XL compiler on Power PC Linux
     MacOSX              Use the default compilers on MacOS X (Darwin)
     MacOSX/GCC          Use the GNU gcc/g++ compilers on MacOSX (Darwin)
     MinGW               Use the GNU gcc/g++ compilers on MinGW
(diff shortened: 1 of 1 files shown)


yumaoka:
@jefgen - I know you have no way to verify this (actually I also don't have AIX env... ouch). Because you provided advice on #1438, I assume you have some history. I created this PR on behalf of @datalogics-robb, so I can no longer approve this PR by myself. Can you approve this?

yumaoka:
@datalogics-robb - please check this PR. thanks. If you're fine with this one, I'll take this one, and close yours (#1438).

jefgen:
> @jefgen - I know you have no way to verify this... Because you provided advice on #1438, I assume you have some history. I created this PR on behalf of @datalogics-robb, so I can no longer approve this PR by myself. Can you approve this?

Sure, I can take a look to see if this matches the other PR.

Looking at the original PR, there were 3 iterations (or commits) that were made:
1. https://github.com/unicode-org/icu/commit/558089456974e8fed023c76c95aa016d7371aba0
2. https://github.com/unicode-org/icu/commit/fbc0337f561ac5c318b63026efd7d4e9e309b36c
3. https://github.com/unicode-org/icu/commit/02e5ffdbd98f3647bb867e9567fad799bd9249a4

The changes in this PR don't quite match the "final" version of the commit from the other PR when it was stamped with LGTM.
It looks like this PR is using the change from the second commit/iteration.

I'm not sure if that is intentional or not though. It looks like the diff between iterations 2 and 3 was changes to the optimization level:
https://github.com/unicode-org/icu/compare/unicode-org:fbc0337..unicode-org:02e5ffd

So perhaps it would be best to use the changes from iteration 3 maybe?

----

Also, FWIW, I wonder if you might possibly consider adding a "co-authored by" to the commit message? That way GitHub will also list Rob as the author of this commit.

I think you'd just need to add this to the commit message:
```
Co-authored-by: From: Rob Boehne <robb@datalogics.com>
```
There is a bit more info on the page here:
https://github.blog/2018-01-29-commit-together-with-co-authors/

datalogics-robb (approved):

datalogics-robb:
XLC (with the xlC front-end) has different flags to control optimization.  xlclang uses flags that are compatible with llvm clang, where -O3 is pretty standard.  Either way, it isn't likely to make much difference, so if someone is more comfortable with -O2 that should be fine too (and not likely to make an impact on performance).

jefgen:
Thanks @datalogics-robb!

jefgen (approved):

markusicu:
@yumaoka This is probably out of sync now with the Markdown version of the readme, right? @echeran copied the readmes into the User Guide, and we will work on moving pieces of them where they make more sense.

PR #1500, opened 2020-12-06
ICU-21427 Don't ignore already checked-in files under "tools/cldr/lib".

jefgen:
I recently ran into trouble trying to mirror/copy files from one ICU clone to another clone due to there being already checked-in files that were ignored by the `.gitignore` file under `tools/cldr`

The problem is that it excludes everything under the lib folder:
```
/lib/**
```

However, there are 3 checked in files in that folder which likely shouldn't be `gitignored`:
```
install-cldr-jars.txt
README.txt
pom.xml
```

The `README.txt` file is already _excluded_ by the `.gitignore` file:
```
!/lib/README.txt
```
We should likely add the other two files to the `.gitignore` file as well.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21427
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/cldr/.gitignore
+++ b/tools/cldr/.gitignore
@@ -1,6 +1,8 @@
-# Exclude the Maven local repository but keep the lib directory and the top-level readme.
+# Exclude the Maven local repository but keep the lib directory and the top-level readme, scripts and build config.
 /lib/**
 !/lib/README.txt
+!/lib/install-cldr-jars.sh
+!/lib/pom.xml
 
 # Ignore the default Maven target directory.
 /cldr-to-icu/target


daniel-ju (dismissed):

markusicu (approved):
lgtm

srl295 (approved):
#1533 has additional comments

jefgen:
> #1533 has additional comments

Good point, I'll add the comments from that, set this as co-authored by @gvictor, and then force push in a moment.

daniel-ju (approved):

PR #1501, opened 2020-12-07
ICU-20107 binary compat: APIs may move between libraries

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20107

User Guide binary compatibility: Clarify that there is no guarantee that the exports from common/i18n are stable in terms of not moving between the two libaries.

https://github.com/markusicu/icu/blob/clarify-binary-compatibility/docs/userguide/design.md

### DIFF
--- a/docs/userguide/design.md
+++ b/docs/userguide/design.md
@@ -776,6 +776,8 @@ For example, here is how an API might be tagged in various versions:
 
 ### ICU Binary Compatibility
 
+*Using ICU as an Operating System Level Library*
+
 ICU4C may be configured for use as a system library in an environment where
 applications that are built with one version of ICU must continue to run without
 change with later versions of the ICU shared library.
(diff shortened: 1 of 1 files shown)


docs/userguide/design.md
```diff
@@ -840,6 +843,22 @@ libraries that resulted from the `–-disable-renaming` ICU build
...
+More recently, the ListFormatter moved from the common library to i18n
+when its features grew beyond primitive patterns to also support
+FieldPosition and FormattedValue features.
+
+There is also a third, “io” library.
+It is possible that some of its functionality may be moved to the i18n or common
+libraries.
+(A likely candidate might be `operator<<(std::ostream& stream, const UnicodeString& s)`
+although there are no actual plans to do so at the time of this writing.)
+
```
jefgen, line 861: Thanks Markus.

I wonder if it would be possible to add the following (or something like it) in order to add some emphasis on why this can be rather problematic on some platforms:

> One can build a combined library which provides the exports from both the “common” and “i18n” libraries, in order to provide a single library for linking against.
>
> This may be needed for some platforms where there is a strong relationship between an API and the library that implements it. For example, on Windows platforms, attempting to find an API that has been moved with a `LoadLibrary`/`GetProcAddress` approach will fail, unless you are using a combined library.

markusicu: sgtm, amended

jefgen: thx

jefgen (approved):
lgtm, thanks!

PR #1506, opened 2020-12-10
ICU-21018 Fix typos across repo that start with letter A and B

erik0686:
There are many typos across the repo. This PR addresses those reported in the ticket ICU-21018. They kindly provided a list, and in this PR I am fixing the ones starting with **letters A and B**. There are a lot so I thought breaking the fixes into several PR's would be better. 
There are a couple of comments I have on some of the words that are in the list.

**acent** --- be careful when replacing, since words like "adjacent" or "acento" include this word. There are also occurrences of acent that are correct in other languages. I would suggest matching the whole word when searching in VS Code or your editor of preference and manually checking if it is a correct instance of the word.
**analyse** --- this word is correct in british english, maybe leave it like that?
**analysed** --- same as before
**availble** --- There is a test name ( testAvailbleLocales() ) that has this typo, can be changes just like that? Please let me know :) You can find those tests in plurults.cpp line 61 and in plurults.h
**behaviour** --- I think this is accepted in English in many parts of the world. 

Please check RBManager_* files, I don't know if these are self generated or if it's okay to modify them.
Please check Transliterator_Han_Latin_Definition, I have a feeling these files can't be modified just like that?

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21018
- [X] Issue accepted
- [ ] Tests included


icu4c/source/i18n/double-conversion-bignum-dtoa.cpp
```diff
@@ -384,7 +384,7 @@ static void BignumToFixed(int requested_digits, int* decimal_point,
 // Returns an estimation of k such that 10^(k-1) <= v < 10^k where
 // v = f * 2^exponent and 2^52 <= f < 2^53.
 // v is hence a normalized double with the given exponent. The output is an
-// approximation for the exponent of the decimal approimation .digits * 10^k.
```
jefgen, line 387: The various `double-conversion*` files in the ICU repo actually come from an external project. The upstream repo is on GitHub here: https://github.com/google/double-conversion

So unfortunately these fixes will get *overwritten* the next time the double-conversion library is updated.

It looks like the typos exist upstream in the current version of the library:
https://github.com/google/double-conversion/blob/master/double-conversion/bignum-dtoa.cc#L373
https://github.com/google/double-conversion/blob/master/double-conversion/bignum-dtoa.cc#L551

Would you mind maybe filing an issue and/or creating a PR for the upstream repo? 
[It would be a nice way to contribute to the upstream project. :) ]

icu4c/source/i18n/regexcmp.cpp
```diff
@@ -2485,7 +2485,7 @@ void        RegexCompile::compileInterval(int32_t InitOp,  int32_t LoopOp)
     fRXPat->fCompiledPat->setElementAt(fIntervalLow, topOfBlock+2);
     fRXPat->fCompiledPat->setElementAt(fIntervalUpper, topOfBlock+3);
 
-    // Apend the CTR_LOOP op.  The operand is the location of the CTR_INIT op.
+    // append the CTR_LOOP op.  The operand is the location of the CTR_INIT op.
```
jefgen, line 2488: This one should probably stay capitalized (beginning of sentence).

icu4c/source/i18n/name2uni.cpp
```diff
@@ -119,7 +119,7 @@ void NameUnicodeTransliterator::handleTransliterate(Replaceable& text, UTransPos
         return;
     }
 
-    // Accomodate the longest possible name
+    // accommodate the longest possible name
```
jefgen, line 122: This one should probably stay capitalized (beginning of sentence).

icu4c/source/i18n/uni2name.cpp
```diff
@@ -81,7 +81,7 @@ void UnicodeNameTransliterator::handleTransliterate(Replaceable& text, UTransPos
         return;
     }
 
-    // Accomodate the longest possible name plus padding
+    // accommodate the longest possible name plus padding
```
jefgen, line 84: This one should probably stay capitalized (beginning of sentence).

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
```diff
@@ -3,7 +3,7 @@
 # @fileLanguage  Finnish
 # @fileCountry   
 # @fileVariant   
-# @fileManager   Jussi Myllym�ki
```
jefgen, line 6: There are a large number of diffs in this file which don't show any actual change in the GitHub diff viewer.
Did the encoding of the file accidentally get changed? If so we probably want/need to change it back.

tools/unicodetools/com/ibm/rbm/resources/RBManager_sv.properties
```diff
@@ -325,7 +325,7 @@ dialog_preferences_locale_iso=ISO\ Standarder\:
 dialog_preferences_locale_machine=Kompatibel\ utrustning\:
 # @translated true @created 2000-11-10 @modified 2000-11-08 @creator Jared Jackson @modifier Stefan Edlund
 # @comment 
-dialog_preferences_locale_warning=OBS.\ Andringar\ till\ lokalen\ ar\ inte\ aktiva\ tills\ applikatoinen\ aterstartas
```
jefgen, line 328: Interesting.... do you know Swedish?

vendor/double-conversion/upstream/double-conversion/bignum-dtoa.cc
```diff
@@ -370,7 +370,7 @@ static void BignumToFixed(int requested_digits, int* decimal_point,
 // Returns an estimation of k such that 10^(k-1) <= v < 10^k where
 // v = f * 2^exponent and 2^52 <= f < 2^53.
 // v is hence a normalized double with the given exponent. The output is an
-// approximation for the exponent of the decimal approimation .digits * 10^k.
```
jefgen, line 373: Similar to the above comment, this comes from the upstream Double-Conversion repo, so ideally it should be fixed upstream.

jefgen (changes requested):

jefgen:
Thanks for tackling all this!

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
erik0686: Possibly :( for some reason the ä became some weird character in various places at some point, and then i tried to change it back by copying the character again. Thanks for this, let me change it back

icu4c/source/i18n/double-conversion-bignum-dtoa.cpp
erik0686: Sure, I can do that in the upstream repo :) Should I return these words in these files to what they were?

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
erik0686: I may be wrong but this is something I found out:

When I accidentally changed the a with the umlaut (ä), it changed from hex <E4> to <EF><BF><BD>, which I believe is the little question mark in a square.
As far as I understand after some research, <E4> is the hex for ä in UTF-16 LE, and since I saved the file as UTF-8, it may have broken that character. The ä in hex in UTF-8 is <C3> <A4>, which is why git shows the difference. 

Should I change it back to UTF-16? My first guess is that everything should be UTF-8, but since this is a name used in other language/country, maybe UTF-16 is the right thing to do. Please let me know if you thing I should use one specifically :)

tools/unicodetools/com/ibm/rbm/resources/RBManager_sv.properties
erik0686: not really hahaha I googled it :laughing:

icu4c/source/i18n/double-conversion-bignum-dtoa.cpp
erik0686: FYI I created a PR on the upstream repo :) I'll probably have to do several PR's there since there might be misspelled words I haven't got to yet, but I'll get to those as I'm going through the list in the ticket on ICU

erik0686: Update, it has been merged upstream :)

jefgen: Nice! -- Thank you!
(Btw, it might be a good idea to link to the upstream PR in the comment here, so that there's some cross-linking for the future in case we forget. :) )

jefgen: > Should I return these words in these files to what they were?

Since it's been fixed upstream I think it would be fine to fix this in the ICU repo now. When the double-conversion library is updated again then these will already be fixed and there won't be any diffs. :)

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
jefgen: Hmm, interesting... thanks for the explanation. :)

Doing some [searching](https://stackoverflow.com/a/46926020)... it looks like Java's `ResourceBundle` files are actually encoded using ISO-8859-1, and not UTF-8, nor UTF-16.  At least up until Java 9, where the default was changed to use UTF-8.

I think when you saved the file as UTF-8 it changed the encoding of non-ASCII characters from ISO-8859-1 to UTF-8 -- which is why the file looks like it has so many diffs.

I think we need to change it back to use ISO-8859-1, and _then_ make the changes to fix the typos in the file.

FWIW, if you're using VSCode, you can change the encoding for a file using the status bar at the bottom ([link](https://stackoverflow.com/questions/30082741/change-the-encoding-of-a-file-in-visual-studio-code#:~:text=In%20the%20bottom%20bar%20of%20VSCode%2C%20you%27ll%20see,that%20this%20only%20applies%20to%20newly%20created%20files.)).

For example, here what it looks like with ISO-8859-1 set:
image

tools/unicodetools/com/ibm/rbm/resources/RBManager_sv.properties
jefgen: :)

icu4c/source/i18n/double-conversion-bignum-dtoa.cpp
erik0686: of course :) the link is this:
https://github.com/google/double-conversion/pull/145

erik0686: Of course, here is the link! https://github.com/google/double-conversion/pull/145

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
erik0686: Thank you for this, Jeff! I have changed it back to the correct encoding.

jefgen:
> There are a couple of comments I have on some of the words that are in the list.
> ...
> **availble** --- There is a test name ( testAvailbleLocales() ) that has this typo, can be changes just like that? Please let me know :) You can find those tests in plurults.cpp line 61 and in plurults.h

FWIW, I think it would be totally fine to fix the typo in the test method name (`testAvailbleLocales`), as this incorrectly spelled name is only used in the test code so it wouldn't change anything in the library code.

jefgen (approved):
lgtm, thanks.

Though you may want to fix the typo for testAvailbleLocales and also change the title of this PR to mention that it is fixing ones with A and B. :)

tools/unicodetools/com/ibm/rbm/resources/RBManager_fi.properties
jefgen: Thanks!

jefgen (approved):
lgtm, thanks for fixing these!

PR #1508, opened 2020-12-11
ICU-21434 First version of a GitHub Action CI script. Contains select…

gnrunge:
…ed CI targets

migrated from Travis CI.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21434
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -1,0 +1,80 @@
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+#
+# GitHub Action configuration script for ICU continuous integration tasks.
+
+name: GHA CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: '*'
+
+jobs:
+
+  # ICU4J build and unit test
+  icu4j-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+        with:
+          lfs: true
+      - name: Checkout lfs objects
+        run: git lfs pull
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
+      - name: ICU4J
+        run: |
+          cd icu4j;
+          ant init;
+          ant check;
+
+  # clang release build with some options to enforce useful constraints.
+  # Includes dependency checker on an in-source, optimized build.
+  # Includes checking @draft etc. API tags vs. ifndef guards like
+  # U_HIDE_DRAFT_API and U_FORCE_HIDE_DRAFT_API.
+  # (FORCE guards make this tool pass but won't compile to working code.
+  # See the testtagsguards.sh script for details.)
+  clang-release-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with clang
+        run: |
+          sudo apt-get -y install doxygen;
+          cd icu4c/source;
+          ./runConfigureICU Linux --enable-layoutex=no;
+          make -j 2;
+          make -j 2 check;
+          ( cd test/depstest && python3 depstest.py ../../../source/ );
+          ( cd .. && source/test/hdrtst/testtagsguards.sh );
+          make dist
+        env:
+          CPPFLAGS: -DU_NO_DEFAULT_INCLUDE_UTF_HEADERS=1
+          CFLAGS: -Wimplicit-fallthrough
+          CXXFLAGS: -Wimplicit-fallthrough
+
+  # Copyright scan
+  copyright-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: perl tools/scripts/cpysearch/cpyscan.pl
+
+  # Check compilation of internal headers.
+  internal-header-compilation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: cd icu4c/source; test/hdrtst/testinternalheaders.sh
+
+  # Check source files for valid UTF-8 and for absence of BOM.
+  valid-UTF-8-and-no-BOM-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: tools/scripts/icu-file-utf8-check.py
--- a/icu4c/source/test/depstest/dependencies.txt
+++ b/icu4c/source/test/depstest/dependencies.txt
@@ -58,6 +58,8 @@ group: c_strings
     # We must not use tolower and toupper because they are system-locale-sensitive (Turkish i).
     strlen strchr strrchr strstr strcmp strncmp strcpy strncpy strcat strncat
     memchr memcmp memcpy memmove memset
+    # BSD version of memcmp().
+    bcmp
     # Additional symbols in an optimized build.
     __strcpy_chk __strncpy_chk __strcat_chk __strncat_chk
     __rawmemchr __memcpy_chk __memmove_chk __memset_chk


.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+#
+# GitHub Action configuration script for ICU continuous integration tasks.
+
+name: ICU Continuous Integration
```
jefgen, line 6: Since this name shows up for every single build line in the small status box, I wonder if maybe we could use a slightly shorter or abbreviated name, so that text doesn't overflow the space (?)

image

For example, for the Azure CI builds I just used "CI" since you can see the blue azure logo. Maybe even "CI" would work for this too, since we can see the GitHub logo?

jefgen (dismissed):
LGTM. Thank you for doing this! 👍

.github/workflows/icu_ci.yml
echeran: As another alternative, we can call it "GH" or "GHA" to help distinguish it from the Azure CI.

jefgen: Good idea, "GH" or "GHA", both sgtm.

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
...
+jobs:
+
+  # ICU4J build and unit test
+  icu4j-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+        with:
+          java-version: 1.11
```
echeran, line 23: What does this line do? I don't think `java-version` is a parameter to the `actions/checkout` action (or if it is, I can't easily find the doc for it), but it wouldn't make sense since conceptually, a `git checkout` and the version of Java you use are orthogonal.

The [`actions/java-setup`](https://github.com/actions/setup-java) command seems to be the way to configure your Java version, and because it is in https://github.com/actions, it means that it is maintained by GH itself.

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
...
+          lfs: true
+      - name: Checkout lfs objects
+        run: git lfs pull
+      - name: ICU4J
+        run: |
+          cd icu4j;
+          ant init;
+          ant check;
+          find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;
```
echeran, line 32: I see the successful negative testing to prove that the failure mode is handled as expected ([after introducing an error for negative testing](https://github.com/gnrunge/icu/runs/1539231211?check_suite_focus=true); [after reverting error for happy case testing](https://github.com/gnrunge/icu/runs/1539317756?check_suite_focus=true))

I'm still puzzled by impact of the `find ...\;` line. My interpretation is that this line should print out all files in the output dir that failed.  I see what you say that it [comes over from the Travis script](https://github.com/unicode-org/icu/blob/bd9c25ba1a6fe3169f9e2d67556419826250c231/.travis.yml#L39-L40), but from the Travis script's heading `after-failure` and examining the output log, I suspect that this line is not actually executing in GHA. I think GHA is exiting early on the failure from the `ant check` line.

Since that line is not doing anything, maybe we can remove that line and put a TODO?  And maybe later try out what would happen if you change
```
...
          ant check;
          find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;
```
into
```
...
          ant check ||
            (find out/junit-results -name "*.txt" -exec grep -l FAILED {} \; 
              && exit 1)
```

.github/workflows/icu_ci.yml
gnrunge: It sets environment:
env:
    JAVA_HOME_11.0.9_x64: /opt/hostedtoolcache/jdk/11.0.9/x64
    JAVA_HOME: /opt/hostedtoolcache/jdk/11.0.9/x64
    JAVA_HOME_11_0_9_X64: /opt/hostedtoolcache/jdk/11.0.9/x64

.github/workflows/icu_ci.yml
gnrunge: I removed the line, migrated over from Travis CI originally but confirmed now that 'ant check' fails (returns non-0) when a test fails. It is a no-op in GHA.

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with clang
+        run: |
+          sudo apt-get -y install doxygen;
+          cd icu4c/source;
+          ./runConfigureICU Linux --enable-layoutex=no;
+          make -j 2;
+          make -j 2 check;
+          # ( cd test/depstest && python3 depstest.py ../../../source/ ) ICU-21436
```
sffc, line 51: Please spend time to figure out why depstest is failing in the GitHub Actions environment.  It was passing in the Travis environment.  I have seen depstest fail in Clang where it doesn't fail in GCC; it may have something to do with the standard library environment.

sffc (dismissed):

.github/workflows/icu_ci.yml
gnrunge: I have an ICU ticket open (ICU-21436) and it can be assigned to me next Wednesday.
This is my last afternoon till Jan. 12 and I think merging this PR even with this ticket open is still beneficial for public ICU so migration away from Travis CI can continue.
Something is brittle, it fails in my Linux machine as well.

echeran (dismissed):

echeran:
LGTM

.github/workflows/icu_ci.yml
markusicu: Looks like bcmp() should be ok.
Would you mind trying to change the deps txt file?

After this line: https://github.com/unicode-org/icu/blob/master/icu4c/source/test/depstest/dependencies.txt#L60
add the following two:
````
    # BSD version of memcmp().
    bcmp
```

If there is no time now, I am ok with merging this PR and improving on it later.

gnrunge: Updated dependencies.txt, enabled depstest, squashed and pushed.

markusicu (approved):
lgtm

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
...
+  # ICU4J build and unit test
+  icu4j-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+        with:
+          lfs: true
+      - name: Checkout lfs objects
+        run: git lfs pull
```
srl295, line 25: did you find this necessary?

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
...
+  clang-release-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with clang
+        run: |
+          sudo apt-get -y install doxygen;
+          cd icu4c/source;
+          ./runConfigureICU Linux --enable-layoutex=no;
```
srl295, line 50: There's an existing issue for this off by default, ICU-13837

.github/workflows/icu_ci.yml
```diff
@@ -0,0 +1,80 @@
...
+          make -j 2;
+          make -j 2 check;
+          ( cd test/depstest && python3 depstest.py ../../../source/ );
+          ( cd .. && source/test/hdrtst/testtagsguards.sh );
+          make dist
```
srl295, line 55: bonus points: package up the dist file and make it available as an artifact…

srl295 (commented):
@gnrunge post merge LGTM with some comments

PR #1512, opened 2020-12-14
ICU-13837 c build: make --disable-layoutex the default

srl295:
ICU-13837

- also update configure.ac that was missed in #1265 cad3903cc25c0db7f392adbc024829aea6591a50 ICU-21051
- also, some FALSE/TRUE were missed in source/test/letest

-----
Notes:
- see https://github.com/harfbuzz/icu-le-hb/issues/13 about building icu-le-hb
-  I did not fix some FALSE/TRUE that may be in source/samples/layout as I do not have a way to test it currently.

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -90,7 +90,7 @@ jobs:
       lfs: true
       fetchDepth: 10
     - script: |
-        export CXXFLAGS="-std=c++14 -Winvalid-constexpr" && cd icu4c/source && ./runConfigureICU --enable-debug --disable-release Linux --disable-layout --disable-layoutex && make -j2 check
+        export CXXFLAGS="-std=c++14 -Winvalid-constexpr" && cd icu4c/source && ./runConfigureICU --enable-debug --disable-release Linux && make -j2 check
       displayName: 'Build and Test C++14'
       env:
         CC: clang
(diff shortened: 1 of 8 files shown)


srl295:
- [x] removed layoutex mention in the github action workflow

srl295:
@markusicu can you make sure this gets noted when  69.1 is released?

jefgen:
> * [x]  removed layoutex mention in the github action workflow

Thanks @srl295!

I think it could probably also be removed from the CI builds too:
https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-pipelines.yml#L93
https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-valgrind.yml#L50

I can open another PR with the same ticket number, or perhaps maybe you might not mind doing it here too? :)

srl295:
> > * [x]   removed layoutex mention in the github action workflow
> 
> Thanks @srl295!
> 
> I think it could probably also be removed from the CI builds too:
> https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-pipelines.yml#L93
> https://github.com/unicode-org/icu/blob/master/.ci-builds/.azure-valgrind.yml#L50
> 
> I can open another PR with the same ticket number, or perhaps maybe you might not mind doing it here too? :)

done

markusicu:
> @markusicu can you make sure this gets noted when 69.1 is released?

Done: http://site.icu-project.org/download/69#TOC-ICU4C-Specific-Changes

markusicu (approved):
lgtm tnx!

jefgen (approved):
lgtm, thanks @srl295 !

PR #1513, opened 2020-12-15
ICU-21346 Update user guide section on initialization.

aheninger:
This change is narrowly focused on the initialization and cleanup / termination
of ICU. The whole user-guide section (file) needs to be reviewed for out-of-date
information, but not as part of this change.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21346
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/docs/userguide/design.md
+++ b/docs/userguide/design.md
@@ -206,35 +206,92 @@ service objects, organized according to locale. Then, if a particular locale's
 formatter is in high demand, that formatter can be used, and then returned to
 the pool.
 
-#### ICU Memory Usage
+### ICU Memory Usage
 
 ICU4C APIs are designed to allow separate heaps for its libraries vs. the
 application. This is achieved by providing functions to allocate and release
 objects owned by ICU4C using only ICU4C library functions. For more details see
-the Memory Usage section in the [Coding Guidelines](dev/codingguidelines.md).
+the Memory Usage section in the [Coding Guidelines](dev/codingguidelines.md#memory-usage).
 
-#### ICU Initialization and Termination
+### ICU4C Initialization and Termination
 
 The ICU library does not normally require any explicit initialization prior to
 use. An application begins use simply by calling any ICU API in the usual way.
-(There is one exception to this, described below.)
+There are, however, a few functions affecting ICU's configuration, that, if used,
+must be called first, before other use of ICU in a process. These are outlined below.
 
-In C++ programs, ICU objects and APIs may safely be used during static
-initialization of other application-defined classes or objects. There are no
-order-of-initialization problems between ICU and static objects from other
-libraries because ICU does not rely on C++ static object initialization for its
-normal operation.
+1. `u_setMemoryFunctions()`. This function replaces the standard library heap
+allocation functions used by ICU with alternate versions, provided by the
+application. If it is needed, `u_setMemoryFunctions()` must be called first, before
+any other use of ICU. This functionality is not commonly used.
 
-When an application is terminating, it may optionally call the function
-`u_cleanup(void)`, which will free any heap storage that has been allocated and
-held by the ICU library. The main benefit of `u_cleanup()` occurs when using
-memory leak checking tools while debugging or testing an application. Without
-`u_cleanup()`, memory being held by the ICU library will be reported as leaks.
+2. ICU Data Locating Functions, `u_setCommonData()`, `u_setDataDirectory()`, and
+`u_setAppData()`. One or more of these functions will be required when ICU is
+configured to load its data directly from files rather than taking it from the
+default data DLL, and the files are not in the default location. Again, this is
+not common. See [ICU Data](icudata#icu-data-directory).
 
-(For some platforms, the configure option `--enable-auto-cleanup` (or
-defining the option `UCLN_NO_AUTO_CLEANUP` to 0) will add code which
-automatically cleans up ICU when its shared library is unloaded. See comments in
-`ucln_imp.h`)
+3. Sanity check that ICU is functioning and able to access data. This is
+important because configuration or installation problems that leave ICU unable
+to load its data do occur, and the resulting failures can be confusing.
+Since not all ICU APIs have UErrorCode parameters, in the absence of data they
+may sometimes silently return incorrect results.
+
+   The function `ulocdata_getCLDRVersion()` is suitable; it is small and light
+weight, requires data, and reports the error in the absence of data.
+
+
+When an application is terminating it should call the function `u_cleanup()`,
+which frees all heap storage and other system resources that are held internally
+by the ICU library. While the use of `u_cleanup()` is not strictly required,
+failure to call it will cause memory leak checking tools to report problems for
+resources being held by ICU library.
+
+Before calling `u_cleanup()`, all ICU objects that were created by the
+application must be deleted, and all ICU services (plain C APIs) must be closed.
+
+For some platforms the configure option `--enable-auto-cleanup`, or defining
+the option `UCLN_NO_AUTO_CLEANUP` to 0, will add code which automatically cleans
+up ICU when its shared library is unloaded. See comments in `ucln_imp.h`
+
+#### C++ Static Initialization and Destruction
+
+The ICU library itself does not rely on C++ static initializers, meaning that
+applications will not encounter order-of-initialization problems from the use of
+ICU.
+
+There are, however, some significant limitations for applications that make use
+of ICU at C++ static initialization time:
+
+1.  `u_setMemoryFunctions()` and the data locating functions, if needed, must
+still be called before any other use of ICU. Which includes any use during the
+construction of static objects.
+
+2.  `u_cleanup()` can only be called after all other ICU-using objects have been
+deleted. Finding a suitable time and place for the call to `u_cleanup()` may be
+difficult, however. Refer to the C++ literature on the order of static
+initialization and destruction.
+
+3.  Destruction of static objects that are scoped to a code block. These, by the
+conventions of C++, are lazily initialized when the code block is first entered,
+so there are no issues during static initialization. But object destruction
+happens when the program terminates, leaving the problem of where to call
+`u_cleanup()`, as discussed above.
+
+#### Dynamically Loading and Unloading ICU
+
+Applications may arrange to dynamically load the ICU library when it is needed,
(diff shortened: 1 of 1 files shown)


aheninger:
The revised user guide section is viewable here: https://aheninger.github.io/icu/userguide/design.html#icu4c-initialization-and-termination

docs/userguide/design.md
```diff
@@ -206,35 +206,84 @@ service objects, organized according to locale. Then, if a particular locale's
...
+application. If it is needed, `u_setMemoryFunctions()` must be called first, before
+any other use of ICU. This functionality is not commonly used.
+
+2. ICU Data Locating Functions, `u_setCommonData()`, `u_setDataDirectory()`, and
+`u_setAppData()`. One or more of these functions will be required when ICU is
+configured to load its data directly from files rather than taking it from the
+default data dll, and the files are not in the default location. Again, this is
+not common. See [ICU Data](icudata#icu-data-directory).
+
+3. `u_init()`. This function performs a sanity check on ICU's ability to access
```
markusicu, line 234: This is all true, but isn't u_init() down to loading a single data file (convrtrs.txt or something)?
As a "sanity check", should we suggest trying to call something that the program will call later anyway, and that will require data, such as formatting a number or a date?

docs/userguide/design.md
```diff
@@ -206,35 +206,84 @@ service objects, organized according to locale. Then, if a particular locale's
...
+the Memory Usage section in the [Coding Guidelines](dev/codingguidelines.md#memory-usage).
 
-#### ICU Initialization and Termination
+### ICU4C Initialization and Termination
 
 The ICU library does not normally require any explicit initialization prior to
 use. An application begins use simply by calling any ICU API in the usual way.
-(There is one exception to this, described below.)
+There are, however, a few functions affecting ICU's configuration, that, if used,
+must be called first, before other use of ICU in a process. These are outlined below.
```
markusicu, line 221: Possible addition: There are some ICU APIs that can be used after setting the memory functions (if that is desired, which as you say is rare) but before configuring the data loading -- because they don't need/load any data.

Trouble is, we can't guarantee a fixed, stable list of such APIs. Normalization has gone in & out of needing data, and currently it depends on the normalization mode.

docs/userguide/design.md
```diff
@@ -206,35 +206,84 @@ service objects, organized according to locale. Then, if a particular locale's
...
-`ucln_imp.h`)
+1.  `u_setMemoryFunctions()` and the data locating functions, if needed, must
+still be called before any other use of ICU. Which includes any use during the
+construction of static objects.
+
+2.  `u_cleanup()` can only be called after all ICU objects have been destructed
+(C++) or closed (plain C APIs). Meaning that `u_cleanup()` cannot be called
+directly just before the application's `main()` function returns. It must either
+be skipped, or called from some other static destructor that runs after all use
+of ICU is done. Skipping `u_cleanup()` is not desirable; doing so will cause leak
```
markusicu, line 269: This seems confusing: Above, u_cleanup() is option; now, skipping it is not desirable.

I think what you mean is, when you use a leak checker, you do want to call u_cleanup(), you just have to find a good time & place after all of your ICU-using static objects are cleaned up.

docs/userguide/design.md
```diff
@@ -206,35 +206,84 @@ service objects, organized according to locale. Then, if a particular locale's
...
+detection tools to falsely report problems with the ICU library.
+
+3.  Destruction of static objects that are scoped to a code block. These, by the
+conventions of C++, are lazily initialized when the code block is first entered,
+so there are no issues during static initialization. But object destruction
+happens when the program terminates, leaving the problem of where to call
+`u_cleanup()`, as discussed above.
+
+#### Dynamically Loading and Unloading ICU
```
markusicu, line 278: I had a comment suggesting such a section, and here it is :-)
@jefgen do you want to add something here?

docs/userguide/design.md
aheninger: I agree that it's inconsistent, and am changing the earlier section to recommend the use of u_cleanup(). And revising the text here.

I'm almost to the point of saying that application code that is not compatible with leak checkers should be considered fatally broken, meaning that u_cleanup() is mandatory. And that C++ static init is so broken that we should not claim to support ICU use at static init, for lack of any reliable way to call u_cleanup(). But maybe some environments have some way to do it that I don't know about. And it will work OK for small apps, with all ICU static objects defined in a single file, which gives a well defined ordering.

I don't think it would be appropriate to write a whole treatise on C++ static init problems and solutions for this section of the user guide.

docs/userguide/design.md
markusicu: wfm

docs/userguide/design.md
```diff
@@ -262,12 +257,10 @@ of ICU at C++ static initialization time:
...
 construction of static objects.
 
-2.  `u_cleanup()` can only be called after all ICU objects have been destructed
-(C++) or closed (plain C APIs). Meaning that `u_cleanup()` cannot be called
-directly just before the application's `main()` function returns. It must either
-be skipped, or called from some other static destructor that runs after all use
-of ICU is done. Skipping `u_cleanup()` is not desirable; doing so will cause leak
-detection tools to falsely report problems with the ICU library.
+2.  `u_cleanup()` can only be called after all other ICU-using objects have been
+deleted. Finding a suitable time and place for the call to `u_clueanup()` may be
```
markusicu, line 261:
```suggestion
deleted. Finding a suitable time and place for the call to `u_cleanup()` may be
```

markusicu (commented):
one more little fix, then please squash

@jefgen there was an earlier question for whether you have some Windows material

docs/userguide/design.md
aheninger: Done

markusicu (approved):
lgtm tnx

docs/userguide/design.md
jefgen: > The existing u_init() does a ucnv_io_countKnownConverters(). So what happens if ICU is configured with no converters?

FWIW, we actually ran into this issue with ICU when it was built using a data file instead of a data DLL. :)

IIRC, when there were no converters to load, the `u_init()` call effectively became a no-op, and so it never actually caused the ICU data file to be memory mapped and loaded at all, and would just return success. This mean that later on other APIs would fail with errors, as they would internally cause the data file to be loaded, and possibly hit out-of-memory, or otherwise fail to load the data file. Or worse, since not all ICU APIs have `UErrorCode` parameters it meant that sometimes APIs would randomly return garbage results... This was all despite initially calling `u_init()` and having it return *success*.

What we ended up doing instead was to call `ulocdata_getCLDRVersion()` first before calling other ICU APIs.

I chose this API instead for the following reasons:
- While some people may choose to omit the converters (due to size, or perhaps not needing them) when building a custom ICU data file, the CLDR version number takes up very little space in the data, and so is generally never removed or filtered out of the data.
- It's exported from the icui18n library, so being able to successfully call it means that both icui18n and icuuc have been loaded. Whereas `u_init` is exported from the icuuc library -- meaning that icui18n might not be loaded yet, and so you can possibly run into OOM issues later on when icui18n is loaded.
- The amount of work `ulocdata_getCLDRVersion` does is fairly trivial compared to other icui18n APIs.

FWIW, you can see an example of this in ChakraCore here:
https://github.com/microsoft/ChakraCore/blob/861a276c346746f3b87347bb094988563d7ee0a7/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp#L610
.NET Core also does the same here:
https://github.com/dotnet/runtime/blob/1d9e50cb4735df46d3de0cee5791e97295eaf588/src/libraries/Native/Unix/System.Globalization.Native/pal_icushim.c#L369

On one hand, I don't know how common it is to use a data file + omit converters. Though on the other hand, it might be nice to add a small note maybe, as this has bitten at least a few.

docs/userguide/design.md
jefgen: Thanks! -- Though I'm not sure if I really have very much to add though. (cc'ing @daniel-ju as well, in case I'm missing something whilst OOF. :) )

Perhaps the following points maybe (?):
- The loading and unloading of ICU should *never* be done inside [DllMain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain). (Loading one of the ICU libraries can cause other libraries or files to be loaded, leading to potential dead-lock.)
- When dynamically loading or unloading ICU, after closing and deleting all ICU objects and services, you should _always_ call `u_cleanup()` before unloading, otherwise you will leak OS Critical Sections. (Or perhaps maybe just say that you will leak things?).

Though the text already says to always call `u_cleanup()` -- so perhaps the second point is redundant.

- Sort of related to the comment here https://github.com/unicode-org/icu/pull/1513#discussion_r544952547 -- if you are using ICU with a data _file_ instead of a data _dll_ then you may want to call `ulocdata_getCLDRVersion()` after loading the ICU libraries before using other APIs to ensure the data file is actually loaded.

srl295: > you should always call u_cleanup() before unloading, otherwise you will leak OS Critical Sections. (Or perhaps maybe just say that you will leak things?).

its worth adding that not just dynamic memory might be leaked.

srl295 (approved):

aheninger:
I have another revision underway that picks up some more of the comments, so am holding off on merging for the moment.

aheninger:
I've added jefgen's review suggestions, and think it's now ready (except for needing squashing).
Rendered preview still at https://aheninger.github.io/icu/userguide/design.html#icu4c-initialization-and-termination

jefgen (approved):
lgtm, thank you. :)

jefgen (dismissed):
lgtm, please squash.

Sorry for the delay in reviewing. -- Thanks again for making these changes.

aheninger:
> Notice: the branch changed across the force-push!
> 
> * docs/userguide/design.md is different
> 
> [View Diff Across Force-Push](https://github.com/unicode-org/icu/compare/unicode-org:1cdba27..unicode-org:b92181c)
> 
> ~ Your Friendly Jira-GitHub PR Checker Bot

Note - the massive diffs across the force-push aren't real - they're from rebasing the branch, which wasn't really necessary, but I did in a moment of confusion. Should be harmless.

markusicu (approved):
lgtm

PR #1517, opened 2020-12-18
ICU-13705 Move BasicTimeZone::getOffsetFromLocal() to its parent Timezone

jungshik:
- Hoist getOffsetFromLocal() to Timezone class from BasicTimezone (Drop 'internal only' bit.)
 - Make the following enums public to control the behavior
   { kStandard, kDaylight, kFormer, kLatter}

 TimeZone implementation of the above method can only support kFormer
 and kLatter (moving what's in Calendar class to TimeZone class).
 Other child classes of BasicTimeZone can support all 4 options as they do
 now.

 This move will simplify some of callers (no need to dynamic_cast to
 BasicTimeZone and branch based on the result of dynamic_cast).

##### Checklist

- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -3161,47 +3161,10 @@ double Calendar::computeMillisInDay() {
 int32_t Calendar::computeZoneOffset(double millis, double millisInDay, UErrorCode &ec) {
     int32_t rawOffset, dstOffset;
     UDate wall = millis + millisInDay;
-    BasicTimeZone* btz = getBasicTimeZone();
-    if (btz) {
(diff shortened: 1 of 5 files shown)


jungshik:
C API is not yet included. Will add it.

jungshik:
> C API is not yet included. Will add it.

Well, there's no timezone C API as far as I can tell. So, there's no C API to add to this PR.

yumaoka:
We're going to make the functionality available though public API, but we plan to keep it in BasicTimeZone. We also plan to define C style enum options. Frank is working on finalizing the design. So this PR might will be likely obsoleted by Frank's coming changes.

yumaoka (commented):
We decided to keep getOffsetFromLocal in BasicTimeZone, but make it public - see #1556. We're not going to merge this one.

PR #1518, opened 2020-12-21
ICU-21279 decompose (NFD/NFKD) UTF-8 with Edits

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21279

This replaces the Normalizer2::normalizeUTF8() base class default implementation for NFD/NFKD. The base class converts to UTF-16, normalizes, and converts back; and it does not support Edits.

This PR adds a dedicated Normalizer2Impl::decomposeUTF8() that only converts substrings that do need normalization changes. As usual, the same code is also written to perform the isNormalizedUTF8() and spanQuickCheckYes() operations, but note that there is no public API for the UTF-8 version of spanQuickCheckYes().

There are some changes in the UTF-8 version of decomposeShort() which is now shared with composeUTF8().

There are some changes for wiring this into the Normalizer2 implementation class hierarchy.

The new code is called from the normalization conformance test, and there is a specific test function that covers various edge cases.

## Code review

This is tricky code. I would be happy to walk you through it.

It does logically the same as the UTF-16 `decompose(const UChar *s, ...)` function, especially with a fast path loop that skips over a sequence of characters with NFD_Quick_Check=Yes and ccc=0.

However, the UTF-8 version is significantly different:
In UTF-16, we append to the ReorderingBuffer which wraps the output string object and also takes care of canonical reordering; it can rewrite the last few characters of the output string directly.
By contrast, for UTF-8 we append to a ByteSink, and we cannot rewrite anything that has been committed to there.
We decompose a minimal substring into the UTF-16 ReorderingBuffer and then convert that to UTF-8 on the fly together with writing to the sink.

Also, with Edits support, we want to be precise about which substrings are copied unchanged vs. which are changed.

In these ways, the new decomposeUTF8() is more similar to the existing composeUTF8().

For the nitty-gritty details about the data structure, and especially about how we encode many normalization properties into the 16-bit `norm16` values, see the table in the design doc:
http://site.icu-project.org/design/normalization/custom#TOC-Per-character-lookup-values-.nrm-formatVersion-3

### DIFF
--- a/icu4c/source/common/norm2allmodes.h
+++ b/icu4c/source/common/norm2allmodes.h
@@ -38,7 +38,7 @@ public:
     virtual UnicodeString &
     normalize(const UnicodeString &src,
               UnicodeString &dest,
-              UErrorCode &errorCode) const {
+              UErrorCode &errorCode) const U_OVERRIDE {
         if(U_FAILURE(errorCode)) {
             dest.setToBogus();
             return dest;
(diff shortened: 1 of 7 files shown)


macchiati (approved):
All makes sense (but tricky stuff!)

markusicu:
Thanks, squashed, please re-approve.

macchiati (approved):

PR #1519, opened 2020-12-22
ICU-13702 allow empty string in UnicodeSet

markusicu:
... and ignore in matching like span().

https://unicode-org.atlassian.net/browse/ICU-13702 “Allow UnicodeSets to contain empty string”
and also
https://unicode-org.atlassian.net/browse/ICU-13805 “UnicodeSet.contains("") should not throw exception”

### DIFF
--- a/icu4c/source/test/intltest/usettest.cpp
+++ b/icu4c/source/test/intltest/usettest.cpp
@@ -98,6 +98,7 @@ UnicodeSetTest::runIndexedTest(int32_t index, UBool exec,
     TESTCASE_AUTO(TestIntOverflow);
     TESTCASE_AUTO(TestUnusedCcc);
     TESTCASE_AUTO(TestDeepPattern);
+    TESTCASE_AUTO(TestEmptyString);
     TESTCASE_AUTO_END;
 }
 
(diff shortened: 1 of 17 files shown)


icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/StringTokenizerTest.java
```diff
@@ -562,12 +562,12 @@ public void Test_GeneratePattern(){
             us._generatePattern(sb.append(1.0), true);
             us._generatePattern(sb.reverse(), true);
         } catch(Exception e){
-            errln("UnicodeSet._generatePattern is not suppose to return an exception.");
+            errln("UnicodeSet._generatePattern is not supposed to return an exception.");
```
markusicu, line 565: Couldn't help myself doing a replace-all on this misspelling...
Only one function has a real change. See around line 768.

markusicu, line 565: FYI  @poulsbo @macchiati A whole section in this test file looks like code coverage tests for class UnicodeSet. Why are these not in UnicodeSetTest?

poulsbo (dismissed):

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/StringTokenizerTest.java
poulsbo: @markusicu I'm guessing the answer is "no good reason."

macchiati (approved):

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -3984,3 +3985,31 @@ void UnicodeSetTest::TestDeepPattern() {
...
+    assertFalse("remove empty string", set.remove(u"").contains(u""));
+    assertTrue("add empty string", set.add(u"").contains(u""));
+    // missing API -- assertTrue("retain empty string", set.retain(u"").contains(u""));
+    assertFalse("complement-remove empty string", set.complement(u"").contains(u""));
+    assertTrue("complement-add empty string", set.complement(u"").contains(u""));
+    // no C++ equivalent for Java strings() -- assertTrue("strings()", set.strings().contains(u""));
+    UnicodeSetIterator sit(set);
+    assertTrue("set iterator.next()", sit.next());
+    assertTrue("set iterator has empty string", sit.isString() && sit.getString().isEmpty());
+    // The empty string is ignored in matching.
```
sffc, line 4004: I'm confused by this test because you add and remove the empty string too much. It's hard to convince myself when reading this test that the set actually contains the empty string at this point. It would be more clear if you had a line explicitly resetting the set before each chunk of tests.

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -3984,3 +3985,31 @@ void UnicodeSetTest::TestDeepPattern() {
...
     errorCode.reset();
 }
+
+void UnicodeSetTest::TestEmptyString() {
+    IcuTestErrorCode errorCode(*this, "TestEmptyString");
+    // Starting with ICU 69, the empty string is allowed in UnicodeSet. ICU-13702
+    UnicodeSet set(u"[{}]", errorCode);
+    if (!assertSuccess("set from pattern with {}", errorCode)) { return; }
+    assertTrue("set from pattern with {}", set.contains(u""));
+    assertFalse("remove empty string", set.remove(u"").contains(u""));
```
sffc, line 3995: `.remove()` mutates the set and returns a self-reference? I know that's how it works, but it's confusing that your assertion statement not only asserts a truth; it also does work.

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -3984,3 +3985,31 @@ void UnicodeSetTest::TestDeepPattern() {
...
+    assertTrue("add empty string", set.add(u"").contains(u""));
```
sffc, line 3996: Would also be good to test .isEmpty/.length of the set when the string is or is not in the set.

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -3984,3 +3985,31 @@ void UnicodeSetTest::TestDeepPattern() {
...
+    // missing API -- assertTrue("retain empty string", set.retain(u"").contains(u""));
```
sffc, line 3997: Is there a .clear/.reset/.removeAll function that you can test?

sffc (changes requested):

icu4c/source/test/intltest/usettest.cpp
markusicu: I rather like it that way: It shows both the function that changes the contents and immediately tests its outcome.
And it's nicely concise.

icu4c/source/test/intltest/usettest.cpp
markusicu: done in a couple of places

icu4c/source/test/intltest/usettest.cpp
markusicu: there are two: clear() and removeAllStrings(); now tested

icu4c/source/test/intltest/usettest.cpp
markusicu: added comments and empty lines

sffc (approved):

markusicu:
thanks, squashed, PTAL

echeran (approved):
RSLGTM

PR #1521, opened 2020-12-29
ICU-20698 UHashtable allow integer value zero

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20698

Make UHashtable usable with integer value zero, so that we need not add/subtract one (or do other tricks) for storing array indexes etc.

I am following the suggestion from the 2019 team ticket discussion: “Steven suggests instead adding a new variant of puti() that allows actually setting a 0 value, and a new variant of geti() that returns both the value and the “found” status.”

This turned out to be a fairly simple addition to the existing machinery.

New internal API:
- uhash_putiAllowZero()
- uhash_iputiAllowZero()
- uhash_getiAndFound()
- uhash_igetiAndFound()
- uhash_containsKey()
- uhash_icontainsKey()

### DIFF
--- a/icu4c/source/common/localematcher.cpp
+++ b/icu4c/source/common/localematcher.cpp
@@ -345,9 +345,8 @@ UBool compareLSRs(const UHashTok t1, const UHashTok t2) {
 int32_t LocaleMatcher::putIfAbsent(const LSR &lsr, int32_t i, int32_t suppLength,
                                    UErrorCode &errorCode) {
     if (U_FAILURE(errorCode)) { return suppLength; }
-    int32_t index = uhash_geti(supportedLsrToIndex, &lsr);
-    if (index == 0) {
-        uhash_puti(supportedLsrToIndex, const_cast<LSR *>(&lsr), i + 1, &errorCode);
+    if (!uhash_containsKey(supportedLsrToIndex, &lsr)) {
+        uhash_putiAllowZero(supportedLsrToIndex, const_cast<LSR *>(&lsr), i, &errorCode);
         if (U_SUCCESS(errorCode)) {
             supportedLSRs[suppLength] = &lsr;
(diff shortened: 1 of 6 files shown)


markusicu:
@poulsbo FYI you have your name on the UHashtable code... :-)

icu4c/source/common/uhash.cpp
```diff
@@ -133,8 +133,10 @@ static const float RESIZE_POLICY_RATIO_TABLE[6] = {
  * or a pointer.  If a hint bit is zero, then the associated
  * token is assumed to be an integer.
  */
+#define HINT_BOTH_INTEGERS (0)
```
srl295, line 136: This is a great addition for clarity

icu4c/source/common/localematcher.cpp
```diff
@@ -345,9 +345,8 @@ UBool compareLSRs(const UHashTok t1, const UHashTok t2) {
 int32_t LocaleMatcher::putIfAbsent(const LSR &lsr, int32_t i, int32_t suppLength,
                                    UErrorCode &errorCode) {
     if (U_FAILURE(errorCode)) { return suppLength; }
-    int32_t index = uhash_geti(supportedLsrToIndex, &lsr);
-    if (index == 0) {
-        uhash_puti(supportedLsrToIndex, const_cast<LSR *>(&lsr), i + 1, &errorCode);
+    if (!uhash_containsKey(supportedLsrToIndex, &lsr)) {
+        uhash_putiAllowZero(supportedLsrToIndex, const_cast<LSR *>(&lsr), i, &errorCode);
```
srl295, line 349: Love being able to get rid of these hacks! 💯

srl295 (approved):
Sleep deprived LGTM

PR #1523, opened 2020-12-29
ICU-20698 C++ Hashtable allow integer value zero

markusicu:
https://unicode-org.atlassian.net/browse/ICU-20698

Same additions for the C++ wrapper as in PR #1521 for the C implementation.

Also change two more callers to use the new functionality.

### DIFF
--- a/icu4c/source/common/hash.h
+++ b/icu4c/source/common/hash.h
@@ -85,15 +85,21 @@ public:
 
     inline int32_t puti(const UnicodeString& key, int32_t value, UErrorCode& status);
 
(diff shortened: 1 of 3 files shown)


FrankYFTang (approved):
I am surprised that we do not have tests under tests/intltest for the Hashtable C++ implementation. Should we add one later?

markusicu:
> I am surprised that we do not have tests under tests/intltest for the Hashtable C++ implementation.

Right, I looked but didn't see one. At least this C++ class is a very thin wrapper over the C API, but that's not fully tested either. I beefed it up a bit in the other PR.

> Should we add one later?

sgtm --> https://unicode-org.atlassian.net/browse/ICU-21447

PR #1524, opened 2020-12-30
ICU-21117 Use Bazel to automate generation of Unicode data files

echeran:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21117
- [X] Issue accepted
- [ ] Tests included

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,5 @@
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This file tells Bazel where the root directory of the project is.
+# "Target" and file paths in Bazel BUILD files are relative to this root directory.
--- a/icu4c/source/common/BUILD
+++ b/icu4c/source/common/BUILD
@@ -1,0 +1,356 @@
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This file defines Bazel targets for a subset of ICU4C "common" library header and source files.
+# The configuration of dependencies among targets is strongly assisted by the
+# file in depstest that maintains such information, at
+# icu4c/source/test/depstest/dependencies.txt .
+
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+# When compiling code in the `common` dir, the constant
+# `U_COMMON_IMPLEMENTATION` needs to be defined. See 
+# https://unicode-org.github.io/icu/userguide/howtouseicu#c-with-your-own-build-system .
+
+# If linker errors occur, then this may be a sign that the dependencies were 
+# not specified correctly. Use dependencies.txt in depstest for assistance. See
+# https://stackoverflow.com/q/66111709/2077918 .
+
+cc_library(
+    name = "headers",
+    hdrs = glob([
+        "unicode/*.h", # public
+        "*.h",         # internal
+        ],
+        # Instead of using this checked-in file, our Bazel build process
+        # regenerates this file and then uses the new version.
+        exclude = ["norm2_nfc_data.h"],
+    ),
+    # We need to add includes in order to preserve existing source files'
+    # include directives that use traditional paths, not paths relative to
+    # Bazel workspace:
+    # https://stackoverflow.com/a/65635893/2077918
+    includes = ["."],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "platform",
+    srcs = [
+        "cmemory.cpp", 
+        "uobject.cpp",
+        "cstring.cpp",
+        "cwchar.cpp",
+        "uinvchar.cpp",
+        "charstr.cpp",
+        "unistr.cpp",
+        "appendable.cpp",
+        "stringpiece.cpp",
+        "ustrtrns.cpp",
+        "ustring.cpp",  
+        "ustrfmt.cpp",  
+        "utf_impl.cpp",
+        "putil.cpp",
+        "ucln_cmn.cpp",  
+        "udataswp.cpp",  
+        "umath.cpp",
+        "umutex.cpp",
+        "sharedobject.cpp",
+        "utrace.cpp",
+    ],
+    deps = [
+        ":headers",
+        # omit other deps b/c they are sys symbols
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+    linkopts = ["-ldl"],
+)
+
+cc_library(
+    name = "utrie",
+    srcs = ["utrie.cpp"],
+    deps = [":platform"],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "utrie2",
+    srcs = ["utrie2.cpp"],
+    deps = [":platform"],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "utrie2_builder",
+    srcs = ["utrie2_builder.cpp"],
+    deps = [
+        ":utrie",
+        ":utrie2",
+        ":platform",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "ucptrie",
+    srcs = ["ucptrie.cpp"],
+    deps = [":platform"],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "umutablecptrie",
+    srcs = ["umutablecptrie.cpp"],
+    deps = [":ucptrie"],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "bytestrie",
+    srcs = ["bytestrie.cpp"],
+    deps = [":platform"],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "bytestriebuilder",
+    srcs = ["bytestriebuilder.cpp"],
+    deps = [
+        ":bytestrie",
+        ":stringtriebuilder",
+        ":sort",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "stringtriebuilder",
+    srcs = ["stringtriebuilder.cpp"],
+    deps = [
+        ":uhash",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "uhash",
+    hdrs = [
+        "uhash.h",
+    ],
+    srcs = [
+        "uhash.cpp",
+    ],
+    deps = [
+        ":headers",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "errorcode",
+    hdrs = [
+    ],
+    srcs = [
+        "errorcode.cpp",
+    ],
+    includes = ["."],
+    deps = [
+        ":platform",
+        ":utypes",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "utypes",
+    srcs = [
+        "utypes.cpp",
+    ],
+    includes = ["."],
+    deps = [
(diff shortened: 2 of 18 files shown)


icu4c/source/common/BUILD
```diff
@@ -92,6 +92,170 @@ cc_library(
...
+    deps = [
+        ":headers",
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "uniset",
```
markusicu, line 110: In toolutil, I expect that most cc_library targets want to have a single .cpp file each.

In common and i18n, this approach should also work, but may not be necessary. My hunch is that making one cc_library target for each dependencies.txt group should be sufficiently granular.

E.g.
```
group: uniset_core
    unifilt.o unifunct.o
    uniset.o bmpset.o unisetspan.o
  deps
    patternprops
    icu_utility
    uvector
```

echeran: Done.

echeran:
@markusicu I updated the file copyright and description headers, and I added scripts to create and delete generated files. Can you verify that `common/norm2_nfc_data.h` is still being created correctly, and tell me if the scripts behave as intended?

WORKSPACE
```diff
@@ -0,0 +1,6 @@
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# A Bazel project is associated with directory the files and subdirectories
+# contained within. Each Bazel project requires a WORKSPACE file, which must
+# exist in the root directory of the project.
```
markusicu, line 6: Trouble parsing this... How about
```suggestion
# This file tells Bazel where the root directory of the project is.
# "Target" and file paths in Bazel BUILD files are relative to this root directory.
```

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This file defines Bazel targets for ICU4C header and source files that are
+# used across i18n algorithm implementations and related tools.
+# The configuration of dependencies among targets is strongly assisted by the
+# file in deptest that maintains such information, at
+# icu4c/source/test/depstest/dependencies.txt .
```
markusicu, line 8:
```suggestion
# This file defines Bazel targets for a subset of ICU4C "common" library header and source files.
# The configuration of dependencies among targets is strongly assisted by the
# file in depstest that maintains such information, at
# icu4c/source/test/depstest/dependencies.txt .
```

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+# not specified correctly. Use dependencies.txt in depstest for assistance. See
+# https://stackoverflow.com/q/66111709/2077918 .
+
+cc_library(
+    name = "headers",
+    hdrs = glob([
+        "unicode/*.h", # public
+        "*.h",         # internal
+        ],
+        exclude = ["norm2_nfc_data.h"],
```
markusicu, line 30: Please add a comment like
```
# Instead of this checked-in file, our Bazel build process regenerates this file and then uses the new version.
```

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+        "sharedobject.cpp",
+        "utrace.cpp",
+    ],
+    deps = [
+        ":headers",
+        # omit other deps b/c they are sys symbols
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+        "U_ATTRIBUTE_DEPRECATED=",
```
markusicu, line 72: Please add a comment like
```
# Linux platform defines.
```

markusicu, line 72: Why do we need to define U_ATTRIBUTE_DEPRECATED as empty? umachine.h defines it for gcc and MSVC and otherwise leaves it empty anyway. And it is indirectly used in definitions across the libraries, so defining it only for "platform" is inconsistent.

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+    ],
+)
+
+cc_library(
+    name = "uhash",
+    hdrs = [
+        # "hash.h",
```
markusicu, line 166: please remove

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+cc_library(
+    name = "normalizer2",
+    srcs = [
+        "normalizer2.cpp",
+        "normalizer2impl.cpp",
+    ],
+    includes = ["."],
+    hdrs = [
+        "normalizer2impl.h",
+        "norm2_nfc_data.h",
```
markusicu, line 337:
```suggestion
        "norm2_nfc_data.h",  # generated by gennorm2
```

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",  
+    ],
+)
+
+
+
+# Note: Bazel has chosen specifically to disallow any official support to break
+# through these constraints that would copying to or generating to source
+# directories: https://stackoverflow.com/questions/51283183/how-to-generate-files-in-source-folder-using-bazel
```
markusicu, line 351:
```suggestion
# Note: Bazel has chosen specifically to disallow any official support for breaking
# through constraints that would allow copying to or generating to source
# directories: https://stackoverflow.com/questions/51283183/how-to-generate-files-in-source-folder-using-bazel
```

icu4c/source/data/unidata/delete_build_headers.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This script deletes newly created header files produced by create_build_headers.sh ,
```
markusicu, line 6: This is not quite what I imagined.

First, I assumed that we would be accumulating a list of generated files. "bazelisk clean" is much better!

Second, I suggested deleting the to-be-generated files before "bazelisk build" to ensure that the build process cannot rely on the to-be-generated files while generating. Therefore, we don't want to "git restore" them either. "git restore" seems only useful when something went wrong and we want to get back to a source tree that works with the "normal" build system.

I suggest you start a TODO section near the top of unidata/changes.txt and put the "bazelisk clean" line there with a one-liner about deleting the to-be-generated files. We can clean that up later.

As for "git restore", can you just `git restore .` (or something) at top level and ? If so, we don't need this .sh file here at all.

Or do we need a script with a hardcoded list? In that case, I would rename this .sh to restore_output_files.sh or simply restore.sh.

icu4c/source/data/unidata/create_build_headers.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This script runs Bazel to create (generate) header files from data that are
```
markusicu, line 6: Not just header files. I suggest renaming this .sh to generate_output_files.sh or simply generate.sh.

icu4c/source/data/unidata/norm2/BUILD
```diff
@@ -0,0 +1,14 @@
...
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This Bazel build file is needed to declare targets for the files used as
+# inputs to binary executables that are a part of other Bazel genrule targets.
+
+package(
+    default_visibility = ["//visibility:public"],
+)
+
+# automatically export plain files as targets in the package
```
markusicu, line 11: plain?
plain text?

tools/unicode/c/genprops/BUILD
```diff
@@ -0,0 +1,50 @@
...
+        "//icu4c/source/common:headers",
+        
+        "//icu4c/source/tools/toolutil:ppucd",
+        "//icu4c/source/tools/toolutil:unewdata",
+        "//icu4c/source/tools/toolutil:writesrc",
+        "//icu4c/source/tools/toolutil:uoptions",
+        "//icu4c/source/tools/toolutil:uparse",
+        "//icu4c/source/tools/toolutil:toolutil",
+        "//icu4c/source/tools/toolutil:denseranges",
+
```
markusicu, line 48: In several targets you have empty lines just before the closing bracket. Why?

icu4c/source/data/unidata/create_build_headers.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Required environment variables:
+#   - ICU_SRC - the root directory of ICU source. This directory contains the
+#               `icu4c` directory.
+
+
+bazelisk build //icu4c/source/common:normalizer2
+cp $ICU_SRC/bazel-bin/icu4c/source/common/norm2_nfc_data.h $ICU_SRC/icu4c/source/common/norm2_nfc_data.h
+
+bazelisk run //tools/unicode/c/genprops $ICU_SRC/icu4c/
```
markusicu, line 17: Please add commands to copy the outputs from genprops back into $ICU_SRC.

markusicu: Actually, for now, genprops takes one folder for both input and output. It will write into the source tree, despite what Bazel likes. Might work for now, at least until we start also building and invoking genuca. Eventually genprops might need another parameter for the output folder.

icu4c/source/data/unidata/delete_build_headers.sh
markusicu: Reviewed together; I had still misunderstood. "bazelisk clean" deletes the *output* files, not the files in the source tree. We want to delete the latter, to guarantee that the bazel build does not pick up old versions of files that it generates.

Agreed on
- rename to clean.sh
- no "bazelisk clean" here but might mention it in unidata/changes.txt
- change "git restore" to "rm"

icu4c/source/common/BUILD
```diff
@@ -0,0 +1,361 @@
...
+        ":headers",
+        # omit other deps b/c they are sys symbols
+    ],
+    local_defines = [
+        "U_COMMON_IMPLEMENTATION",
+        "U_ATTRIBUTE_DEPRECATED=",
+        "U_HAVE_XLOCALE_H=0",
+        "U_HAVE_STRTOD_L=1",
+        "U_HAVE_ELF_H=1",
+        "_REENTRANT",
```
markusicu, line 76: I checked on these four. They are either not used anywhere, or not used in the common library, or documented as being obsolete. Let's remove them.

Details: https://unicode-org.atlassian.net/browse/ICU-21553

icu4c/source/common/BUILD
echeran: Done.

echeran (commented):
I believe I have applied all the changes as indicated in the PR and/or as discussed together. The only divergence is a minor one -- I put the comment `# Linux platform headers` in the first target, where other similar comments exist, not in the second target.

icu4c/source/common/BUILD
```diff
@@ -27,13 +26,16 @@ cc_library(
...
         exclude = ["norm2_nfc_data.h"],
     ),
-    # need to add includes in order to preserve existing source files'
+    # We need to add includes in order to preserve existing source files'
     # include directives that use traditional paths, not paths relative to
-    # Bazel workspace
+    # Bazel workspace:
     # https://stackoverflow.com/a/65635893/2077918
     includes = ["."],
+    # Linux platform defines.
```
markusicu, line 38: ... except the one that's left here is not a Linux platform define. Please remove.

icu4c/source/data/unidata/changes.txt
```diff
@@ -49,6 +49,12 @@ For new script codes see http://www.unicode.org/iso15924/codechanges.html
 
 ---------------------------------------------------------------------------- ***
 
+Unicode 14.0 update for ICU 69
```
markusicu, line 52:
```suggestion
Unicode 14.0 update for ICU 70
```

icu4c/source/data/unidata/clean.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+# This script deletes the header files that generate.sh tries to regenerate.
```
markusicu, line 6:
```suggestion
# This script deletes files that generate.sh tries to regenerate.
```

markusicu:
... and once you have made those three fixes feel free to squash right away.
How about something like this for a commit comment? --> "ICU-21117 Use Bazel to automate generation of Unicode data files"

echeran:
Those changes sounds good, and I applied them with the suggested commit message for the squash+push.

markusicu (approved):
lgtm tnx

PR #1526, opened 2021-01-06
ICU-21448 Update ICU tzdata to 2020f

yumaoka:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21448
- [x] Issue accepted
- [x] Tests included


icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -2575,51 +2582,47 @@ SimpleDateFormat::parse(const UnicodeString& text, Calendar& cal, ParsePosition&
             } else { // tztype == TZTYPE_DST
                 if (dst == 0) {
                     if (btz != NULL) {
```
yumaoka, line 2584: TZ database 2020e added daylight saving rules with 30 minutes offset around 1947 to 1968 in America/Belize.  This change revealed a design issue of "golden zone", which is used for mapping metazone (e.g. America_Central) back to timezone (e.g. America/Chicago for US, America/Belize for Belize).

America/Chicago is global golden zone for metazone America_Central. America/Belize is regional golden zone for metazone America_Central. When a date string "1970-07-01 12:00:00 Central Daylight Time" is parsed by ICU date format initialized with en_US locale, "Central Daylight Time" is mapped to metazone America_Central, then mapped to time zone America/Chicago. Chicago observed daylight saving time on July 1st, 1970, there are no problems parsing the strings into date value. However, if ICU date format is initialized with en_BZ locale, the metazone is mapped to time zone America/Belize. Belize did not observe daylight saving time on July 1st, 1970, so this implementation tries to adjust offset (by parsing "Central Daylight Time", DST is detected, but time zone offset of America/Belize has no daylight offset on the day).

Previous logic for adjusting daylight offset was based on nearest daylight saving time before and after the given time. tzdata up to 2020d did not have daylight saving rule up to 1968 with +30 minutes offset. Belize used +1 hour daylight offset in 1973, so this value was used with 2020d. As the result, it's still consistent with all other time zones mapped to America_Central.

tzdata 2020e added the irregular daylight saving time (+30 minutes offset) in 1968, so the old implementation evaluates daylight offset in Feb 1968 (+30 minutes) and Dec 1973 (+ 1 hour). In this case, Feb 1968 is closer, and the offset at that time is picked.

The new implementation changed precedence. When daylight saving time is not observed by the resolved time zone, it looks for future daylight saving time. It only looks for past daylight saving time only when daylight saving time was never used after the parsed time. This implementation works better for a few reasons.

- CLDR metazone based display names are never used for date before 1970.
- If Belize uses odd daylight saving offset after 1970, ICU does not use the metazone America_Central for the period, so this is not an issue.
- In general, there are fewer zones using daylight offset other than +1 hour in modern dates.

Strictly speaking, this is a limitation of the current golden zone concept. If we need to support parsing date string with zone name across regions, we need some extra data/mechanism. I'm not really proposing to improve the design, because we need to pay extra overhead (more data, more complicated code) just for resolving some edge case issues.

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -2575,51 +2582,47 @@ SimpleDateFormat::parse(const UnicodeString& text, Calendar& cal, ParsePosition&
...
+                        // This implementation resolves daylight saving time offset
+                        // closest rule after the given time.
+                        UDate baseTime = localMillis + raw;
+                        UDate time = baseTime;
+                        UDate limit = baseTime + MAX_DAYLIGHT_DETECTION_RANGE;
+                        TimeZoneTransition trs;
+                        UBool trsAvail;
+
+                        // Search for DST rule after the given time
+                        while (time < limit) {
```
yumaoka, line 2594: Small improvement. If there is a zone alter multiple offsets infinitely in future, and neither of them are not daylight saving time, while(TRUE) loop won't terminate. Updated implementation uses ~30 year window, and does not look for further. I'm sure there no such rules available at this moment although.

icu4c/source/test/intltest/tzfmttst.cpp
```diff
@@ -1340,4 +1341,52 @@ TimeZoneFormatTest::TestFormatTZDBNamesAllZoneCoverage(void) {
         }
     }
 }
+
```
yumaoka, line 1344: I added a new test case for validating the updated implementation with the problematic case. The same issue can be detected by running TestTimeRoundtrip() with full coverage mode, which may take several hours to complete.

icu4c/source/test/intltest/tztest.cpp
```diff
@@ -2074,6 +2074,8 @@ void TimeZoneTest::TestCanonicalID() {
         {"Asia/Vientiane", "Asia/Bangkok"},
         {"Atlantic/Jan_Mayen", "Europe/Oslo"},
         {"Atlantic/St_Helena", "Africa/Abidjan"},
+        {"Australia/Currie", "Australia/Hobart"},
```
yumaoka, line 2077: This test case is evaluating a set of equivalent zones. The test case was written when canonical ID definition depended on tz database's "Link" rules. The current ICU implementation uses canonical ID mapping coming from CLDR. This old test code is somewhat still useful, but we need to maintain these exceptions. The list of exception mapping has been growing, because tz database tends to use "Link" rule more and more recently (cutting off unreliable differences before 1970).

pedberg-icu (dismissed):

yumaoka:
@pedberg-icu I made a minor change in the ICU4C test case (adding status check after parse() call for just in case) and squashed. Could you please re-approve this PR?

icu4c/source/test/intltest/tztest.cpp
markusicu: This sounds like it would be useful as a test code comment. Follow-up PR?

markusicu (approved):
mostly rslgtm

PR #1527, opened 2021-01-06
ICU-21284 Correctly normalize Unit Identifiers

hugovdm:
Progress tracker:

- [X] Preserve order of unitQuantities, make everything work with the change.
- [x] Add correct sorting of single units within compound units.
- [x] Consider filing a more significant ticket than ICU-21284? Or better, expand the scope of ICU-21284?
- [x] Port changes to Java.
- [x] Ensure thorough unit test coverage as per ICU-21284.

Closes https://github.com/icu-units/icu/issues/70

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21284
- [X] Issue accepted
- [X] Tests included


icu4c/source/data/misc/units.txt
```diff
@@ -1898,50 +1898,137 @@ units:table(nofallback){
         }
     }
     unitQuantities{
```
younies, line 1900: I do not see a reason for just making new level without combining two or three entries together.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -194,6 +194,65 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
...
+        if (outIndex + array.getSize() > outSize) {
+            status = U_INDEX_OUTOFBOUNDS_ERROR;
+            return;
+        }
+
+        for (int32_t i = 0; array.getValue(i, value); ++i) {
+            U_ASSERT(i < array.getSize());
+            U_ASSERT(outIndex < outSize);
+            ResourceTable table = value.getTable(status);
+            U_ASSERT(table.getSize() == 1);
```
younies, line 229: It is obvious that the table size always equal to 1. so, I do not know why do we need it

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -194,6 +194,65 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
...
+            const char *key;
+            table.getKeyAndValue(0, key, value);
+            int32_t uTmpLen;
+            // TODO(review): how about we make gQuantities a UChar** instead,
```
younies, line 233: I do not like having a lot of TODOs in the code, you can have a small issue instead. Or you can review it right now.

hugovdm: This TODO is a question to the review - i.e. the intent is indeed to discuss/decide, and fix this, as part of the review.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Clear how, where? By looking at units.txt? Data might change, so code assumptions might break, and this is also intended as documentation to the reader of the code: we're only calling getKeyAndValue with index 0. "Why not other indices?" the reader might wonder, "ah, because there is only one, as is clearly asserted". Does that make more sense?

icu4c/source/data/misc/units.txt
hugovdm: Purpose: we need the unitQuantity elements in their original order.

Implementation: adding <FIFO> to the ldml2icu rules is the way to do this. All examples of FIFO that I found have /<FIFO>/ in the rule. If necessary, we could check with David if there's another way to do this, but I think this is normal.

Consequences: units.txt comes out like this. (This file is generated by the ldml2icu tool.)

sffc: Yeah.  In ICU restext, `{}` is used for both an array and a dictionary!  This is a key difference from JSON, where `[]` is used for arrays instead.  What Hugo has written here is the correct way to make an array of dictionaries.

icu4c/source/i18n/measunit_extra.cpp
sffc: +1 to what Hugo said.  Technically setting an error code is better than an assertion though, since ICU should be at least *somewhat* fault-tolerant to bad data files.  I don't remember the error we normally use in this situation; search the code base for examples.

icu4c/source/i18n/measunit_extra.cpp
sffc: In response to the review question: I strongly prefer if gQuantities is a `UChar**`, such that it contains pointers into the data bundle rather than owning a bunch of strings that might never be used.  I might also look into whether you can change `prefs.getPreferencesFor` to accept a `UChar*` as well; try to move the conversion as close as possible to where it is actually needed.  When you do this, you sometimes discover that you don't actually need to do any conversion at all.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -216,6 +293,35 @@ UBool U_CALLCONV cleanupUnitExtras() {
 
 void U_CALLCONV initUnitExtras(UErrorCode& status) {
     ucln_i18n_registerCleanup(UCLN_I18N_UNIT_EXTRAS, cleanupUnitExtras);
+    LocalUResourceBundlePointer unitsBundle(ures_openDirect(nullptr, "units", &status));
+
+    // Collect unitQuantities information into gSerializedUnitQuantitiesTrie and gQuantities.
```
sffc, line 298: Each simple unit is backed by a numeric ID into `gSimpleUnits`.  Instead of making a BytesTrie for lookup, can't you just make an array with the same length as `gSimpleUnits`?

sffc (commented):
Mostly LGTM!

icu4c/source/i18n/measunit_extra.cpp
hugovdm: I've done something a bit like this now, adding a map from "simple unit ID index" to "unitQuantities index". However we still need to look up strings for other use cases. I'm thus keeping the BytesTrie, using it to construct this new int->int map, as well as for the other use case (`U_I18N_API const char *getUnitQuantity(StringPiece baseUnitIdentifier, UErrorCode &status);`).

(We need to be able to look up derived units, like kilogram-meter-per-square-second [newton], so unitQuantities is about more than just simple units.)
- We need to map from the index of "newton" in gSimpleUnits to the

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Agreed, I've been coming even more to that conclusion with the latest addition (thereby unnecessarily having to repeat the UChar -> char* conversion).

**Not yet done. I'll request another review once I'm done fixing that.**

icu4c/source/i18n/measunit_extra.cpp
hugovdm: Going with U_INVALID_FORMAT_ERROR, looks reasonable enough.

icu4c/source/i18n/measunit_extra.cpp
hugovdm: As of 2741a2d I've now got "getUnitQuantity" doing the conversion from UChar* to char*. I also looked at taking it further, having getPreferencesFor accept a UChar*, however it turns out, the "category" comes from a key in the resources, not a value, so is itself a char*. Some conversion thus needs to happen - either I convert all "categories" to UChar at loading time, or I convert all unitQuantities from UChar* to char* at loading time (what I used to have, before 2741a2d), or some conversion happens at lookup time when constructing a NumberFormatter. For now, I'm leaving it with that - conversion happening at getUnitQuantity time.

sffc (commented):
I don't see anything else where I'd suggest changes, so LGTM

hugovdm:
@sffc pardon the many changes on the C++ side: they're in separate commits in case that helps, and they're mostly just renaming functions to be in sync with ICU4J. (There was already more Categories support in ICU4J than in ICU4C, so I'm going by its naming policy: "category" instead of the CLDR XML's "quantity".)

hugovdm:
@sffc For incremental reviews: I find diffs are also less confusing when comparing commits up to a merge, and commits from a merge - i.e. not including merge commits in the set of commits being reviewed. YMMV.

hugovdm:
GitHub has been giving some 500s today. I've pushed commit 9159beda6f, but this PR doesn't currently see it (I might need to nudge it somehow).

That latest commit tries to make MSVC happy by using an unsigned size_t to stare strlen results instead of an int32_t - but it's still passed to a function that expects an int32_t (CharString::append). ICU might need a going-over to change variables and internal parameters that ought to be unsigned / size_t?

sffc (commented):
LGTM please squash

sffc (approved):

hugovdm:
@pedberg-icu (2 of 2) Do you mind if I go ahead and merge this now? It will impact PR 1551.

hugovdm:
With timezones and deadlines, I've decided to go ahead and merge PRs now, despite impact on 1551. Hope it doesn't cause too much hassle/delay for integration!

pedberg-icu:
> With timezones and deadlines, I've decided to go ahead and merge PRs now, despite impact on 1551. Hope it doesn't cause too much hassle/delay for integration!

No problem, I will rebase 1551.

PR #1528, opened 2021-01-12
ICU-21116 Fix ListFormatterTest on Android

gvictor:
Android ignores assert statement by default.
The test also failed when running with ant and on OpenJDK runtime
on release-67-1. But no longer fails on master branch.


sffc (approved):
Thank you!

gvictor:
I don't have right to merge the PR. Can someone help?

sffc:
@jefgen @gnrunge @echeran @markusicu Can we add a CI test that runs *without* assertions enabled to catch things like this?

FrankYFTang (approved):

PR #1531, opened 2021-01-16
ICU-21450 Add asan check, tsan check, gcc debug build and test, gcc 8…

gnrunge:
… build and

test, and ICU4C release tools buildability check to GitHub Action Continuous Integration.

##### Checklist

- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -32,6 +32,33 @@ jobs:
           ant init;
           ant check;
 
+  # gcc debug build.
+  # Includes dependency checker.
+  # Note - the dependency checker needs to be run on both a debug and an optimized build.
+  # This one (gcc) for debug, and linux clang (see job below) for optimized.
+  #
+  # Test both out-of-source and in-source builds. This one (gcc) for out-of-source,
+  # and linux clang (below) for in-source.
+  #
+  # Invokes test/hdrtst to check public headers compliance.
+  gcc-debug-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with gcc
+        run: |
+          mkdir build;
+          cd build;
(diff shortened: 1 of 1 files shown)


.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,31 @@ jobs:
...
+  gcc-debug-build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with gcc
+        run: |
+          mkdir build;
+          cd build;
+          ../icu4c/source/runConfigureICU --enable-debug --disable-release Linux/gcc --prefix=/tmp/icu-prefix --enable-tracing;
```
echeran, line 53: optional: In the .travis.yml file, and in the job that you added right below this, you made `PREFIX` an env variable and reference it using `${PREFIX}`. You could do that here to for consistency's sake.

.github/workflows/icu_ci.yml
```diff
@@ -58,6 +83,62 @@ jobs:
           CFLAGS: -Wimplicit-fallthrough
           CXXFLAGS: -Wimplicit-fallthrough
 
+  # Out of source build with gcc 8 and c++14; executes icuinfo.
+  gcc-8-stdlib_14:
```
echeran, line 87: nit: It seems strange to see a YAML key with an underscore. On the other hand, down below in this PR, `INTLTEST_OPTS` is a key that must match the shell env variable to be created.

echeran (approved):
LGTM

.github/workflows/icu_ci.yml
gnrunge: Done.

.github/workflows/icu_ci.yml
gnrunge: Changed to 'stdlib14'.

PR #1535, opened 2021-01-19
ICU-21462 ICULocaleService.java has a race condition

grhoten:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21462
- [x] Issue accepted
- [ ] Tests included

I'm reordering the assignment that gets into this double check synchronized block to be the last step within the block. We want to avoid a scenario where loc is assigned to fallbackLocale before returning the fallbackLocaleName from this function.

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/ICULocaleService.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/ICULocaleService.java
@@ -612,9 +612,9 @@ public class ICULocaleService extends ICUService {
         if (loc != fallbackLocale) {
             synchronized (this) {
                 if (loc != fallbackLocale) {
-                    fallbackLocale = loc;
                     fallbackLocaleName = loc.getBaseName();
                     clearServiceCache();
+                    fallbackLocale = loc;
                 }
             }
         }


grhoten:
Good point. Apparently my default email address was used when checking this out. I'll have to fix that.

jefgen:
Looks like the CLA checker is happier now. :)

markusicu:
@yumaoka looks ok? Making it work with the before-synchronized `if` in the double-checked locking?

Besides that, does it make sense to use `==` for the locales? Should we not use Locale.equals or Objects.equals?

yumaoka:
>Besides that, does it make sense to use == for the locales? Should we not use Locale.equals or Objects.equals?

ULocale instance is immutable and ULocale#getDefault() returns a singleton. ULocale#getDefault() may return a new ULocale instance when ULocale#setDefault(ULocale) is called. If fallbackLocale is always initialized with ULocale#getDefault(), then `==` makes sense with the current implementation. However, ULocale#equals(Object) is supposed to check object's equality, ULocale#equals(Object) should not have any overhead. I don't think ULocale#getDefault() returning a singleton object is an API contract, so I prefer equals() method over `==` (in this code `!=`).

This code uses synchronize(this), so fallbackLocale / fallbackLocaleName are guarded by the synchronized block. However, when ULocale.setDefault(ULocale) is called between `fallbackLocaleName = loc.getBaseName()` and `fallbackLocale = loc`, then I think it will introduce inconsistent state. So I'm not sure if the fix really resolves the problem. I'll carefully look at this again tomorrow.

yumaoka:
OK - I now understand the problematic case.

Thread 1 enters synchronized block, updating fallbackLocale.
Thread 2 evaluate `if (loc != fallbackLocale)` outside of the synchronized block, then found `loc == fallbackLocale` .
Thread 2 returns fallbackLocaleName before Thread 1 is updating fallbackLocaleName in the synchronized block.

yumaoka (approved):
LGTM.

markusicu (approved):
lgtm

PR #1536, opened 2021-01-20
ICU-21450 Add builds for MacOS and for g++ with extra warnings to GHA CI.

gnrunge:
ICU-21450 Fix indent.

##### Checklist

- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -47,6 +47,8 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: ICU4C with gcc
+        env:
+          PREFIX: /tmp/icu-prefix
         run: |
           mkdir build;
           cd build;
@@ -56,8 +58,6 @@ jobs:
           ( cd ../icu4c/source/test/depstest && ./depstest.py ../../../../build/ );
           make install;
           PATH=$PREFIX/bin:$PATH make -C test/hdrtst check
-        env:
-          PREFIX: /tmp/icu-prefix
 
   # clang release build with some options to enforce useful constraints.
   # Includes dependency checker on an in-source, optimized build.
@@ -71,6 +71,10 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: ICU4C with clang
+        env:
+          CPPFLAGS: -DU_NO_DEFAULT_INCLUDE_UTF_HEADERS=1
+          CFLAGS: -Wimplicit-fallthrough
+          CXXFLAGS: -Wimplicit-fallthrough
         run: |
           sudo apt-get -y install doxygen;
           cd icu4c/source;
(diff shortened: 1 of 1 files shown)


.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,27 @@ jobs:
           ant init;
           ant check;
 
+  # Out of source build with gcc 8 and Wextra.
```
markusicu, line 35: Can we use a newer gcc version here? I suspect that "gcc 8" really wants to be the "latest available version of gcc".
https://gcc.gnu.org/releases.html

.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,27 @@ jobs:
...
+  gcc-8-Wextra:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with gcc 8 and extra warnings.
+        run: |
```
markusicu, line 42: Just curious: What does the pipe symbol do?

.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,27 @@ jobs:
...
+          mkdir build;
+          cd build;
+          ../icu4c/source/runConfigureICU Linux --disable-debug --enable-release --disable-layoutex --prefix=$PREFIX;
+          make -j2;
+          make -j2 check;
+          make install;
+          PATH=$PREFIX/bin:$PATH make -C test/hdrtst check
+        env:
+          PREFIX: /tmp/icu-prefix
```
markusicu, line 51: Optional: I assume that the order of "run" vs. "env" doesn't matter, but for human understanding I would prefer "env" before "run".

.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,27 @@ jobs:
           ant init;
           ant check;
 
+  # Out of source build with gcc 8 and Wextra.
+  gcc-8-Wextra:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C with gcc 8 and extra warnings.
```
markusicu, line 41: ... and hdrtst

.github/workflows/icu_ci.yml
```diff
@@ -32,6 +32,27 @@ jobs:
           ant init;
           ant check;
 
+  # Out of source build with gcc 8 and Wextra.
+  gcc-8-Wextra:
```
markusicu, line 36: Could we reasonably add the Wextra and hrdtst into another existing job, maybe `gcc-8-stdlib14`?
(And maybe upgrade the gcc version there if a newer one is available for GHA?)

.github/workflows/icu_ci.yml
sffc: It starts a multiline string in Yaml.

.github/workflows/icu_ci.yml
gnrunge: Switched to version 10, in what is now gcc-10-stdlib, this particular job is removed in accordance with comment below.

.github/workflows/icu_ci.yml
gnrunge: Rearranged.

.github/workflows/icu_ci.yml
gnrunge: Job has been removed in accordance to comment below.

.github/workflows/icu_ci.yml
gnrunge: -Wextra added to job gcc-10-stdlib14.
hdrtst is also executed by job gcc-debug-build-and-test.

markusicu (approved):
lgtm tnx

PR #1537, opened 2021-01-22
ICU-21466 Remove size limit of intvector

FrankYFTang:
##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21466
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/tools/genrb/reslist.cpp
+++ b/icu4c/source/tools/genrb/reslist.cpp
@@ -340,7 +340,8 @@ IntResource::~IntResource() {}
 IntVectorResource::IntVectorResource(SRBRoot *bundle, const char *tag,
                   const UString* comment, UErrorCode &errorCode)
         : SResource(bundle, tag, URES_INT_VECTOR, comment, errorCode),
-          fCount(0), fArray(new uint32_t[RESLIST_MAX_INT_VECTOR]) {
+          fCount(0), fSize(RESLIST_INT_VECTOR_INIT_SIZE),
+          fArray(new uint32_t[fSize]) {
     if (fArray == NULL) {
         errorCode = U_MEMORY_ALLOCATION_ERROR;
         return;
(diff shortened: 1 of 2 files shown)


icu4c/source/tools/genrb/reslist.cpp
```diff
@@ -352,6 +353,17 @@ IntVectorResource::~IntVectorResource() {
 }
 
 void IntVectorResource::add(int32_t value, UErrorCode &errorCode) {
+    if (fCount == fSize) {
+        uint32_t* tmp = new uint32_t[2 * fSize];
```
sffc, line 357: We can't `new[]` in library code, but this is genrb, and we are already `new[]`ing in this file, so it's fine.

icu4c/source/tools/genrb/reslist.cpp
```diff
@@ -352,6 +353,17 @@ IntVectorResource::~IntVectorResource() {
...
+        if (tmp == nullptr) {
+            errorCode = U_MEMORY_ALLOCATION_ERROR;
+            return;
+        }
+        uprv_memcpy(tmp, fArray, fSize * sizeof(uint32_t));
+        delete fArray;
```
sffc, line 363: `delete[]`

icu4c/source/tools/genrb/reslist.h
```diff
@@ -406,6 +406,7 @@ class IntVectorResource : public SResource {
 
     // TODO: UVector32
     uint32_t fCount;
+    uint32_t fSize;
```
sffc, line 409: Since we're in genrb (not library code) then maybe we should make this size_t

sffc (dismissed):

sffc (commented):
LGTM please squash

sffc (approved):

PR #1538, opened 2021-01-23
ICU-21450 Migrates ICU4C document build with doxygen from Travis CI t…

gnrunge:
…o GHA CI.

##### Checklist

- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -13,6 +13,21 @@ on:
 
 jobs:
 
+  # ICU4C docs build using doxygen..
+  icu4c-docs-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: ICU4C doc
+        run: |
+          sudo apt-get -y install doxygen;
+          cd icu4c/source;
+          ./runConfigureICU Linux --disable-renaming;
+           # Fail if 'warning:' appears in doxygen's output, but ignore warnings from file Doxyfile.
+           # Regex note: (?! ... ) is a negative lookahead. Succeed if the pattern is not present. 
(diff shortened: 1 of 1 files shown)


.github/workflows/icu_ci.yml
```diff
@@ -13,6 +13,21 @@ on:
 
 jobs:
 
+  # ICU4C docs build using doxygen..
+  icu4c-docs-build:
```
echeran, line 17: optional: rename as `icu4c-docs` ?

echeran (approved):
LGTM.

Btw, did there end up needing to be codebase updates to get this step to work with the latest version of doxygen?  Or am I misremembering a different tool in the Travis->GH Action conversion work whose latest version was causing warnings/errors?  Either way, I can see that this ported GH Action job succeeds.

gnrunge:
> LGTM.
> 
> Btw, did there end up needing to be codebase updates to get this step to work with the latest version of doxygen? Or am I misremembering a different tool in the Travis->GH Action conversion work whose latest version was causing warnings/errors? Either way, I can see that this ported GH Action job succeeds.

What happened is that I got the doxygen warnings when I executed the command in Travis CI on my Google Linux machine, which has doxygen 1.8.20. Created ticket#21464 (https://unicode-org.atlassian.net/browse/ICU-21464). A comment in .travis.yml says that Travis uses a down-rev doxygen version. It was then only later, when I executed the command on GHA,
that I found that GHA also uses an older doxygen version (1.8.13) which does not produce these warnings. So the ICU4C docs build fine on GHA, at least for the time being.
I have already fixed most of the warnings in ticket#21464, and will send a PR as soon as I am done with the five remaining warnings to make sure the doc build target keeps passing.

PR #1539, opened 2021-01-26
ICU-21465 Fix Windows Time Zone offset when Automatic DST setting is Off

jefgen:
This PR fixes a bug in the changes for ICU-13845, for handling when the "Automatic DST" setting is OFF in the Control Panel.

It turns out that the offset value for the calculated time zone when the "Automatic DST" setting is turn off is wrong. The sign for the time zone offset is flipped. (It's positive when it should be negative, and vice-versa).

I think this was perhaps due to confusion about the value for the UTC offset, how POSIX style offsets are handled, and what is reported by the Win32 API `GetDynamicTimeZoneInformation`.
It seems that this issue was actually present in the original PR for ICU-13845 here: https://github.com/unicode-org/icu/pull/1297 , and it was missed in the reviews on the other PR for the ticket here: https://github.com/unicode-org/icu/pull/1362 .

I also missed this when I was manually testing out the changes, as I was expecting the wrong values for the offset to be reported. (For example, `Etc/GMT-8` instead of `Etc/GMT+8`). 

Note: The actual change here is only one character. The rest are comments that I added in order to help explain things for others in the future.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21465
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/wintz.cpp
+++ b/icu4c/source/common/wintz.cpp
@@ -124,10 +124,26 @@ uprv_detectWindowsTimeZone()
         // No way to support when DST is turned off and the offset in minutes is not a multiple of 60.
         if (utcOffsetMins % 60 == 0) {
             char gmtOffsetTz[11] = {}; // "Etc/GMT+dd" is 11-char long with a terminal null.
-            // Note '-' before 'utcOffsetMin'. The timezone ID's sign convention
-            // is that a timezone ahead of UTC is Etc/GMT-<offset> and a timezone
-            // behind UTC is Etc/GMT+<offset>.
-            int ret = snprintf(gmtOffsetTz, UPRV_LENGTHOF(gmtOffsetTz), "Etc/GMT%+ld", -utcOffsetMins / 60);
+            // Important note on the sign convention for zones:
+            //
+            // From https://en.wikipedia.org/wiki/Tz_database#Area
+            //   "In order to conform with the POSIX style, those zone names beginning with "Etc/GMT" have their sign reversed
+            //   from the standard ISO 8601 convention. In the "Etc" area, zones west of GMT have a positive sign and those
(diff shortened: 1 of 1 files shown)


FrankYFTang:
Could you list a step-by-step (manual is ok) test set up procedure that we can follow to reproduce the problem before the fix, and verify your PR fix the problem after we apply the PR?

FrankYFTang:
Also, for the issue fixed in this PR, could you also scope under which timezone, what condition (control panel setting, os) and when (say between which months) would the issue be reproducible and then fixed by this PR. Is it reproducible in any timezone which observe daylight saving time?

axelandrejs (approved):

jefgen:
> Could you list a step-by-step (manual is ok) test set up procedure that we can follow to reproduce the problem before the fix, and verify your PR fix the problem after we apply the PR?

Yes, I've created a document that has different manual test cases that exercise various scenarios for most of the code paths for the Windows TZ mapping in ICU. (I also added notes on what the various test cases are exercising in the code.)
Link: https://docs.google.com/document/d/17Y_Ns4EBV8wnHioJNFc-dJCRRpJ3hpJZDYmTq6ekIsM

Test case 7 in that document reproduces the issue; with the fix the correct result of `Etc/GMT-10` should be reported back from the `icuinfo.exe` tool.

> Also, for the issue fixed in this PR, could you also scope under which timezone, what condition (control panel setting, os) and when (say between which months) would the issue be reproducible and then fixed by this PR. Is it reproducible in any timezone which observe daylight saving time?

This issue would effect anyone that has set the "Automatic DST" setting to be OFF in the Control Panel -- and potentially anyone with the random corruption that was seen on the Chrome bug report. It could occur for any time zone that has a non-zero offset, and wouldn't matter if it was during daylight saving time or not.

daniel-ju (approved):

FrankYFTang (approved):

PR #1543, opened 2021-01-26
ICU-21465 Cherry-pick: Fix Windows Time Zone mapping when the Automatic DST setting is OFF to ICU 68.

jefgen:
This PR is cherry-picking the change from PR #1539 into the `maint/maint-68` branch. (Cherry-picked from commit a1a1fafb7d1a0e1c9beee626b6b57e63c27b901b).

#### From the other PR description:

This PR fixes a bug in the changes for ICU-13845, for handling when the "Automatic DST" setting is OFF in the Control Panel.

It turns out that the offset value for the calculated time zone when the "Automatic DST" setting is turn off is wrong. The sign for the time zone offset is flipped. (It's positive when it should be negative, and vice-versa).

I think this was perhaps due to confusion about the value for the UTC offset, how POSIX style offsets are handled, and what is reported by the Win32 API `GetDynamicTimeZoneInformation`.
It seems that this issue was actually present in the original PR for ICU-13845 here: https://github.com/unicode-org/icu/pull/1297 , and it was missed in the reviews on the other PR for the ticket here: https://github.com/unicode-org/icu/pull/1362 .

I also missed this when I was manually testing out the changes, as I was expecting the wrong values for the offset to be reported. (For example, `Etc/GMT-8` instead of `Etc/GMT+8`). 

Note: The actual change here is only one character. The rest are comments that I added in order to help explain things for others in the future.

#### Testing:
I created a document that has different manual test cases that exercise various scenarios for most of the code paths for the Windows TZ mapping in ICU. (I also added notes on what the various test cases are exercising in the code.)
Link: https://docs.google.com/document/d/17Y_Ns4EBV8wnHioJNFc-dJCRRpJ3hpJZDYmTq6ekIsM
(The test case 7 in the document reproduces the issue that this PR fixes).

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21465
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/wintz.cpp
+++ b/icu4c/source/common/wintz.cpp
@@ -124,10 +124,26 @@ uprv_detectWindowsTimeZone()
         // No way to support when DST is turned off and the offset in minutes is not a multiple of 60.
         if (utcOffsetMins % 60 == 0) {
             char gmtOffsetTz[11] = {}; // "Etc/GMT+dd" is 11-char long with a terminal null.
-            // Note '-' before 'utcOffsetMin'. The timezone ID's sign convention
-            // is that a timezone ahead of UTC is Etc/GMT-<offset> and a timezone
-            // behind UTC is Etc/GMT+<offset>.
-            int ret = snprintf(gmtOffsetTz, UPRV_LENGTHOF(gmtOffsetTz), "Etc/GMT%+ld", -utcOffsetMins / 60);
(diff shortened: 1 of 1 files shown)


daniel-ju (approved):

ViniciusFXavier (commented):
👍

PR #1555, opened 2021-02-02
ICU-21461 uprops.h: remove unused gc macros

markusicu:
and some underused character constants

https://unicode-org.atlassian.net/browse/ICU-21461

It turned out that the gc macros were not used. (They probably were used some time ago.)

There were also some adjacent character constants (as enum constants) that are less useful now that we can use Unicode character literals. I removed some that were used in only one or two places.


echeran (approved):
LGTM

jefgen (approved):
lgtm, thanks for cleaning this up! 😄

PR #1557, opened 2021-02-02
ICU-21349 Change UnitConverter name to UnitsConverter

younies:
##### Checklist

- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsConverter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsConverter.java
@@ -11,12 +11,12 @@ import java.util.regex.Pattern;
 
 import com.ibm.icu.util.MeasureUnit;
 
-public class UnitConverter {
+public class UnitsConverter {
     private BigDecimal conversionRate;
     private BigDecimal offset;
 
     /**
-     * Constructor of `UnitConverter`.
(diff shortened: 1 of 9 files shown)


hugovdm (changes requested):
- UnitsConverter.java is currently looking like an addition and not a rename
- Java unit test failure seems to be about UnitConverter vs UnitsConverter

Otherwise, looks good. (I was unsure whether singular or plural makes most sense in the past, given that it usually converts a single input unit. Thus I mostly sat on the fence with the PRs that added it.)

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsConverter.java
```diff
@@ -11,12 +11,12 @@
...
+public class UnitsConverter {
     private BigDecimal conversionRate;
     private BigDecimal offset;
 
     /**
-     * Constructor of `UnitConverter`.
+     * Constructor of <code>UnitsConverter</code>.
      * NOTE:
      * - source and target must be under the same category
      * - e.g. meter to mile --> both of them are length units.
```
hugovdm, line 22: Doesn't look like this javadoc gets rendered to html, so this probably doesn't matter / is entirely optional:

A list in javadoc probably needs to be done via html <ul> / <li> rather than Markdown syntax. AFAIK.

hugovdm (approved):

PR #1551, opened 2021-02-01
ICU-21480 integrate CLDR release-39-alpha0 to ICU trunk for ICU 69rc

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21480
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [x] Tests included
- [x] Documentation is changed or added

This integrates the CLDR release-39-alpha0 to ICU trunk for ICU 69rc. There are several changes that are either significant or affect a lot of files; in some cases I have noted people that I think should look at them in more detail.
- The CLDR tools build process has changed to use maven, which affects portions of the CLDR-to-ICU integration and ICU data build process that use or refer to CLDR tools. The changes here are only a partial step towards what eventually needs to be done; @srl295 has done some further work on this and should look at the changes here.
- [CLDR-2698](https://unicode-org.atlassian.net/browse/CLDR-2698) effectively swapped the behavior of locales for _no_ and _nb_ to match the way other macrolanguages are handled (like zh and cmn) so that all the content formerly in _nb_ has been moved to _no_, and the _nb_ is now just a stub that aliases to _no_. However, there is an additional wrinkle: _no_ is also treated as the parent language for _nn_, so that things like formats and collation can be inherited by _nn_ from _no_. Also, we retain a localeDisplayName for _nb_ that is distinct from the name for _no_. This may have significant implications for some clients. @FrankYFTang  should look at this (perhaps especially at the changes to localeCanonicalization.txt) ; @jefgen should also look at this, and particularly the change to Windows LCID mapping in icu4c/source/common/locmap.cpp. On MSVC this leads to the following ICU4C cintltst failure:
    ```
	cldrtest  {
	 TestLocaleStructure   {
	  (Known issue 20181) ICU-20181 Fix LCID mapping for ckb vs ku
	   ERROR:   no_NO has 0x0414 and the number resolves wrongfully to nb_NO
	 } ---[1 ERRORS in /tsutil/cldrtest/TestLocaleStructure]  (140ms)  
	 ...
    ```
- [CLDR-14355](https://unicode-org.atlassian.net/browse/CLDR-14355) and [CLDR-13566](https://unicode-org.atlassian.net/browse/CLDR-13566) made changes to locale matcher behavior; in particular the former eliminated fallbacks between Hans and Hant, which affected a number of tests. @markusicu should check these.
- Some other CLDR changes affected many files:
[...]

PR #1566, opened 2021-02-07
ICU-21306 ldml2icu: add rules for Inflections-related CLDR data

hugovdm:
@pedberg-icu since you seem to be the primary maintainer of structLocale.txt - how does this file work? Is it manually maintained?
For myself, I also don't mind whether I contribute this before or after #1551 - whichever seems easier for the integration PRs?

I have added all the grammaticalFeatures data, though I'm not yet sure whether we'll use all of it in ICU.

Closes icu-units#48

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21306
- [X] Issue accepted
- [x] Tests included


icu4c/source/data/unit/de.txt
```diff
@@ -10,12 +10,42 @@ de{
     units{
         acceleration{
             g-force{
```
hugovdm, line 12: Using unit/de.txt as an exemplar for what I'm doing with the ldml2icu rules:

In this PR's current state, with its suggested ldml2icu rules, I'm adding "case" and "gender" right next to the plural forms.

This means they must be handled by the PluralTableSink: I'm skipping "case" at this time. When case is specified, its forms should be used instead of the normal (nominative) plural forms. I'm keeping "gender", as with "dnam" and "per".

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -236,6 +236,7 @@ make check 2>&1 | tee /tmp/icu4c-newData-makeCheck.txt
 # Note that if the new data has any differences in structure, you will have to
 # update test/testdata/structLocale.txt or /tsutil/cldrtest/TestLocaleStructure
```
hugovdm, line 237: Instructions don't say how to update structLocale.txt - as per FIXME added below, is it manually maintained now?

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -39,19 +39,36 @@ constexpr int32_t DNAM_INDEX = StandardPlural::Form::COUNT;
...
 constexpr int32_t PER_INDEX = StandardPlural::Form::COUNT + 1;
+/**
+ * Gender of the word, in languages with grammatical gender.
+ */
+constexpr int32_t GENDER_INDEX = StandardPlural::Form::COUNT + 2;
 // Number of keys in the array populated by PluralTableSink.
-constexpr int32_t ARRAY_LENGTH = StandardPlural::Form::COUNT + 2;
+constexpr int32_t ARRAY_LENGTH = StandardPlural::Form::COUNT + 3;
 
 static int32_t getIndex(const char* pluralKeyword, UErrorCode& status) {
```
hugovdm, line 49: Following examples from standardplural.cpp
We do have only three things to check here, whereas for both examples in standardplural.cpp it's six. Potentially not worth having a switch if we're only checking for 3 values? Shall I revert the commit in which I did this?

icu4c/source/test/testdata/structLocale.txt
```diff
@@ -11,6 +11,7 @@
 // Since ICU doesn't follow the traditional fallback model where all keys fallback to root,
 // root can't be used to validate the structure.
 // This file was generated from en.txt, and all ".*" were replaced with "".
```
hugovdm, line 13: As per FIXME below, I'm not trusting this line is up to date? How does one "generate" this file?

pedberg-icu: Responding to the above and to
> @pedberg-icu since you seem to be the primary maintainer of structLocale.txt - how does this file work? Is it manually maintained?

Yes, structLocale.txt is manually maintained these days. Basically it needs an empty entry corresponding to any entry that might occur in any locale. So for example if some locale has a gregorian-calendar availableFormats entry for skeleton MMMMEEEEd, structLocale.txt needs such an entry (with empty string for value), even though en does not have such an entry.

icu4c/source/data/unit/de.txt
```diff
@@ -10,12 +10,42 @@ de{
...
+                case{
+                    accusative{
+                        one{"{0} g-Kraft"}
+                        other{"{0} g-Kraft"}
+                    }
+                    dative{
```
pedberg-icu, line 18: This is really a comment for the CLDR structure & data, but seems like for situations like this where there are no distinctions in value for some or al of the cases, to reduce data size we might want to have a case "other" or "fallback" that can provide the default value if the specific case is not listed.

hugovdm: The current "other" or "fallback" is the _parent_ of "case" - that's the "nominative" case.

My current proposal being this, keeping existing structure as much the same as possible:

            g-force{
                case{
                    accusative{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                    dative{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                    genitive{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                }
                dnam{"g-Kraft"}
                gender{"feminine"}
                one{"{0} g-Kraft"}
                other{"{0} g-Kraft"}
            }

A cleaner suggestion would be this, when I call it **other** instead of **nominative** - since I'm also not sure whether it's linguistically "correct" to call it nominative for languages that don't have case? ;-) :

            g-force{
                case{
                    other{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                    accusative{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                    dative{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                    genitive{
                        one{"{0} g-Kraft"}
                        other{"{0} g-Kraft"}
                    }
                }
                dnam{"g-Kraft"}
                gender{"feminine"}
            }

hugovdm: Spec formaly falls back to "nominative", then to "no case":
https://www.unicode.org/reports/tr35/tr35.html#Lateral_Inheritance

I suggest that's very much an argument in favour of calling it "nominative", not "other", if I go with moving nominative into the "case" table.

hugovdm:
AppVeyor failure seems unrelated to this PR:

```
make[1]: Making `all' in `genbrk'
make[2]: Entering directory '/cygdrive/c/projects/icu/icu4c/source/tools/genbrk'
generating dependency information for genbrk.cpp
g++ -D_REENTRANT  -DU_HAVE_ELF_H=1 -DU_HAVE_STRTOD_L=1 -DU_HAVE_XLOCALE_H=0  -DWINVER=0x0601 -D_WIN32_WINNT=0x0601 -I../../common -I./../toolutil -DU_ATTRIBUTE_DEPRECATED= -O3  -W -Wall -pedantic -Wpointer-arith -Wwrite-strings -Wno-long-long -std=c++11   -c   -o genbrk.o genbrk.cpp
make[2]: *** [../../config/mh-cygwin64:76: genbrk.o] Segmentation fault
make[2]: Leaving directory '/cygdrive/c/projects/icu/icu4c/source/tools/genbrk'
```

genbrk: "Tool for generating RuleBasedBreakIterator data files (.brk files)."

icu4c/source/test/testdata/structLocale.txt
hugovdm: I have changed instructions at the top of the file: I removed this line and added mention of "manually maintained".

icu4c/source/i18n/number_longnames.cpp
sffc: :man_shrugging:

sffc (dismissed):

hugovdm:
Looks like I do need to get appveyor passing - I'll retry.

sffc:
There are lots of warnings, which you can ignore, but the error that caused cygwin to fail is:

> make[2]: *** [../../config/mh-cygwin64:76: genbrk.o] Segmentation fault

You didn't change anything having to do with genbrk, so I'm inclined to treat that as a flake.  After you squash, the tests will re-run and you can see if the failure is reproducible.

hugovdm:
Squashed: this triggers rebuild anyway. (And here I was, wondering how to retry AppVeyor... https://github.com/unicode-org/icu/pull/1566/checks lets one retry GHA CI and Azure pipelines only.)

hugovdm:
@pedberg-icu (1 of 2) Do you mind if I go ahead and merge this now? It will impact PR 1551.

hugovdm:
I've gone ahead and merged PR #1527, and I'll merge this one when approved again. I didn't see the unit tests rerun with the change in master... I'll hit the "retry" button on /checks - which will at least rerun most of them.

younies (approved):

sffc (approved):

hugovdm:
I went ahead and merged, because on PR 1527 @pedberg-icu replied with "no problem" to "I've decided to go ahead and merge PRs now". Thanks!

PR #1573, opened 2021-02-10
ICU-21492 Add test to show regexp_open bug

FrankYFTang:
##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21492
- [ ] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/test/cintltst/reapits.c
+++ b/icu4c/source/test/cintltst/reapits.c
@@ -171,6 +171,7 @@ static void TestUTextAPI(void);
 static void TestRefreshInput(void);
 static void TestBug8421(void);
 static void TestBug10815(void);
+static void TestURegexOpenCrash(void);
 
 void addURegexTest(TestNode** root);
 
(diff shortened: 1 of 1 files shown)


FrankYFTang:
Andy- this is just a PR to show the bug

aheninger (changes requested):
It would be better to add this test case to the data driven test file test/testdata/regextst.txt.
There's no need to write new code.

I recommend placing the new test case at the end of the existing bug regression tests, here:
https://github.com/unicode-org/icu/blob/3333fd557c398c98805f1ffdc63df796dc2bdcbd/icu4c/source/test/testdata/regextst.txt#L1499

The problem is in a complicated part of the code. Understanding exactly what is going on, and fixing it without accidentally creating other problems, might take some thought. I can take a look.

Maybe it would be better for moment to forgo this PR, verify that the bug can be reproduced with one line in the test data file, and document that line in a comment to the bug.

FrankYFTang:
abandon

PR #1577, opened 2021-02-13
ICU-21492 Fix regex compile assertion failure.

aheninger:
A regex pattern containing nested look-behind blocks could trigger an assertion
failure during pattern compilation. The problem was caused by an off-by-one
error in the code that computes an upper bound on the match length, needed
because look-behind expressions are constrained to not have unbounded match
length.

Nested look-behind blocks come into play because, when computing the maximum
match length of an outer block, any inner look-behind blocks are skipped over -
they do not directly contribute to the length matched by the outer block. The
problem was in the code that skips over these nested look-behind blocks.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21492
- [x] Issue accepted
- [x] Tests included

### DIFF
--- a/icu4c/source/i18n/regexcmp.cpp
+++ b/icu4c/source/i18n/regexcmp.cpp
@@ -3475,6 +3475,9 @@ int32_t   RegexCompile::minMatchLength(int32_t start, int32_t end) {
 //                     value may be longer than the actual maximum; it must
 //                     never be shorter.
 //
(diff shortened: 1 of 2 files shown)


markusicu (approved):
rslgtm tnx!!

I am not familiar with the code, but I trust you as the original author and long-time maintainer.
Thanks for adding the comment.

FrankYFTang (approved):

PR #1581, opened 2021-02-16
ICU-21480 integrate CLDR release 39 alpha1 to ICU trunk

pedberg-icu:
##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21480
- [x] Issue accepted
- [x] Tests included

This integrates CLDR release-39-alpha1 into ICU. Compared to the previous integration of CLDR release-39-alpha0, this includes the CLDR changes to make locale "nb" a _child_ of "no" instead of an _alias_ to it per [CLDR PR-1031 ](https://github.com/unicode-org/cldr/pull/1031), which also included a fix for CLDR LocaleIDParser to properly handle the parent relationship for nb/nn to no.

Some things to note:
- Probably because of the LocaleIDParser fix, this integration _does_ remove a lot of data from "nn" that is now inherited from "no" (which has content for nb_NO).
- Many of the data and code changes to "nb" vs "no" are just reverting changes that were made in the previous integration [PR-1551](https://github.com/unicode-org/icu); in that PR, "nb" was deprecated and aliased to "no" (which became undeprecated), and this PR undoes the relevant parts of that (the deprecation of "nb").
- There were some changes to the ICU4C resource bundle lookup code to handle the new situation that a locale ID without '_' can have a non-root parent. Currently these changes specifically address the "no"/"nb"/"nn" situation, we may want to work out a more general approach.

There were some tooling issues to note (in addition to fixing missing "break" statements in recently added "case" statements, etc.):
- For some reason, the version of data/rbnf/nb.txt generated by the tooling is mangled, with some partially inherited content instead of just being an empty stub with a %%Parent entry pointing to "no" (the CLDR xml file is empty). I had to fix this manually, and filed [ICU-21496](https://unicode-org.atlassian.net/browse/ICU-21496) about it.
- I was also not able to figure out how to get the tooling to generate the no_NO.txt data files as empty stubs without generating a whole suite of files no_Latn.txt, no_Latn_NO.txt, etc. The only way I could avoid the additional files was to add a redundant global alias from "no_NO" to "no" in build-icu-data.xml, which results in the right set of files but with some additional redundant %%ALIAS entries. I will file a JIRA ticket about that too.

(I would like to get these changes integrated so would prefer to fix the tooling as a separate later PR rather than in this PR)


tools/cldr/cldr-to-icu/build-icu-data.xml
```diff
@@ -323,7 +323,7 @@
                     // F-P
                     fa_AF, fa, ff, fil, fi, fo, fr_BE, fr_CH, fr, ga, he, hi, hr,
                     hu, hy, id, in, is, it, iw, ja, ka, kl, km, ko, ky, lb,
-                    lo, lrc, lt, lv, mk, ms, mt, my, nb, nl, nn, no, pl, pt_PT, pt,
+                    lo, lrc, lt, lv, mk, ms, mt, my, nb, ne, nl, nn, no, pl, pt_PT, pt,
```
jefgen, line 326: Just curious, where does the "ne" locale come from? (And why only for RBNF?)

pedberg-icu: It is not just for RBNF, it is in the regular localeId> list in this file, and has been a locale in ICU for some time now. The addition of RBNF data was per [CLDR-14177](https://unicode-org.atlassian.net/browse/CLDR-14177)

macchiati (approved):
There was one additional fix, removal of a "<". Clearly needed, just noting here.

icu4c/source/test/testdata/numberpermutationtest.txt
```diff
@@ -3504,7 +3504,7 @@ measure-unit/length-furlong unit-width-full-name rounding-mode-floor
 percent unit-width-narrow integer-width/##00
   es-MX
     00 %
-    1827.3645 %
+    1,827.3645 %
```
pedberg-icu, line 3507: All of the changes in this file are related to CLDR changing the minimumGrouping for es_419 locales from 2 to 1.

pedberg-icu:
> There was one additional fix, removal of a "<". Clearly needed, just noting here.

Do you mean in the Korean names of Japanese eras? That was a fix in the CLDR data, just here by virtue of integrating the CLDR data. There were some other fixes like that such as eliminating the literal &lt;LRM&gt; in Persian units data.

tools/cldr/cldr-to-icu/build-icu-data.xml
jefgen: Got it, thanks Peter!

hugovdm (commented):
Thanks for fixing another of my oversights! (The missed `break;`. :disappointed: Noted!)

PR #1584, opened 2021-02-17
ICU-21499 Suppress warning about non-NULL addresses in uprv_memcpy and uprv_memmove

jefgen:
Both Clang and GCC will output warnings when they know a variable cannot be `NULL`/`nullptr`.
This causes the two `U_ASSERT` macros added to `uprv_memcpy` and `uprv_memmove` in ICU-21118 to generate a number of noisy warnings.

We can suppress these warnings for Clang and GCC. Though doing so is a bit tricky, as we can't use `#ifdef` or `#pragma` inside of a macro `#define` token-string. The resulting output from the #define isn't processed by the preprocessor again, so if we have anything like `#pragma` in the token-string, it effectively gets ignored.
However, it turns out that we can use the C99 `_Pragma` [feature ](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html)for this (thanks to @erik0686 for this tip!).

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21499
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/common/cmemory.h
+++ b/icu4c/source/common/cmemory.h
@@ -37,19 +37,57 @@
 #include <stdio.h>
 #endif
 
-
+// uprv_memcpy and uprv_memmove
+#if defined(__clang__)
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    /* Suppress warnings about addresses that will never be NULL */ \
+    _Pragma("clang diagnostic push") \
+    _Pragma("clang diagnostic ignored \"-Waddress\"") \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    _Pragma("clang diagnostic pop") \
+    U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size); \
+} UPRV_BLOCK_MACRO_END
+#define uprv_memmove(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    /* Suppress warnings about addresses that will never be NULL */ \
+    _Pragma("clang diagnostic push") \
+    _Pragma("clang diagnostic ignored \"-Waddress\"") \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    _Pragma("clang diagnostic pop") \
+    U_STANDARD_CPP_NAMESPACE memmove(dst, src, size); \
+} UPRV_BLOCK_MACRO_END
+#elif defined(__GNUC__)
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    /* Suppress warnings about addresses that will never be NULL */ \
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Waddress\"") \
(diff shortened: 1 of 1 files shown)


icu4c/source/common/cmemory.h
```diff
@@ -37,19 +37,63 @@
...
+// uprv_memcpy
+#if defined(__clang__)
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    /* Suppress warnings about addresses that will never be NULL */ \
+    _Pragma("clang diagnostic push") \
+    _Pragma("clang diagnostic ignored \"-Waddress\"") \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size); \
+    _Pragma("clang diagnostic pop") \
```
markusicu, line 49: maybe pop before calling memcpy?

icu4c/source/common/cmemory.h
```diff
@@ -37,19 +37,63 @@
...
+} UPRV_BLOCK_MACRO_END
+#elif defined(__GNUC__)
 #define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    /* Suppress warnings about addresses that will never be NULL */ \
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Waddress\"") \
     U_ASSERT(dst != NULL); \
     U_ASSERT(src != NULL); \
     U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size); \
+    _Pragma("GCC diagnostic pop") \
```
markusicu, line 59: pop before calling memcpy?

if you change these, please also change memmove.

icu4c/source/common/cmemory.h
```diff
@@ -37,19 +37,63 @@
...
+#else
+#define uprv_memcpy(dst, src, size) UPRV_BLOCK_MACRO_BEGIN { \
+    U_ASSERT(dst != NULL); \
+    U_ASSERT(src != NULL); \
+    U_STANDARD_CPP_NAMESPACE memcpy(dst, src, size); \
+} UPRV_BLOCK_MACRO_END
+#endif
 
+// uprv_memmove
+#if defined(__clang__)
```
markusicu, line 70: optional: maybe use one single if-elif chain, with both memcpy and memmove next to each other in each branch?

markusicu (commented):
thanks!
looks good but some ideas/suggestions --

icu4c/source/common/cmemory.h
jefgen: ah, yes good point -- thanks!

icu4c/source/common/cmemory.h
jefgen: done

icu4c/source/common/cmemory.h
jefgen: good idea, I think that would help to clean it up. :)

markusicu (commented):
lgtm pse squash

erik0686 (approved):

markusicu (approved):
lgtm tnx

PR #1589, opened 2021-02-19
ICU-21323 Automates the uconfig variation BRS tasks. The test

gnrunge:
subsequently sets each of the UCONFIG_NO_XXX flags to '1' (exceptions apply)
and runs the ICU4C unit tests and the header test. Afterwards all
UCONFIG_NO_XXX flags are set to '1' and unit tests and header test
are executed.

To allow concurrent execution and reduce total run time, the script provides
the option to execute only the unit tests or only the header test.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21323
- [x] Issue accepted
- [ ] Tests included

### DIFF
--- a/tools/scripts/uconfig_vars_test.py
+++ b/tools/scripts/uconfig_vars_test.py
@@ -1,0 +1,219 @@
+# © 2021 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+"""Executes uconfig variations check.
+
+See
+http://site.icu-project.org/processes/release/tasks/healthy-code#TOC-Test-uconfig.h-variations
+for more information.
+"""
+
+import getopt
+import os
+import re
+import subprocess
+import sys
+
+excluded_unit_test_flags = ['UCONFIG_NO_CONVERSION', 'UCONFIG_NO_FILE_IO'];
+
+def ReadFile(filename):
+    """Reads a file and returns the content of the file
+
+    Args:
+      command: string with the filename.
+
+    Returns:
+      Content of file.
+    """
+
+    with open(filename, 'r') as file_handle:
+      return file_handle.read()
+
+def RunCmd(command):
+    """Executes the command, returns output and exit code, writes output to log
+
+    Args:
+      command: string with the command.
+
+    Returns:
+      stdout and exit code of command execution.
+    """
+
+    command += ' >> uconfig_test.log 2>&1'
+    print(command)
+    p = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE,
+                         stderr=subprocess.STDOUT, close_fds=True)
+    stdout, _ = p.communicate()
+    return stdout, p.returncode
+
+def ExtractUConfigNoXXX(uconfig_file):
+    """Parses uconfig.h and returns a list of UCONFIG_NO_XXX labels.
+    Initializes test result structure.
+
+    Args:
+      uconfig_file: common/unicode/uconfig.h as string.
+
+    Returns:
+      List of all UCONFIG_NO_XXX flags found in uconfig.h, initialized test
+      result structure.
+    """
+
+    uconfig_no_flags_all = []
+    test_results = {}
+    uconfig_no_regex = r'UCONFIG_NO_[A-Z_]*'
+
+    # Collect all distinct occurences of UCONFIG_NO_XXX matches in uconfig.h.
+    for uconfig_no_flag in re.finditer(uconfig_no_regex, uconfig_file):
+        if uconfig_no_flag.group(0) not in uconfig_no_flags_all:
+            uconfig_no_flags_all.append(uconfig_no_flag.group(0))
+
+    # All UCONFIG_NO_XXX flags found in uconfig.h come in form of a guarded
+    # definition. Verify the existence, report error if not found.
+    for uconfig_no_flag in uconfig_no_flags_all:
+        uconfig_no_def_regex = r'(?m)#ifndef %s\n#\s+define %s\s+0\n#endif$' % (
+        uconfig_no_flag, uconfig_no_flag)
+        uconfig_no_def_match = re.search(uconfig_no_def_regex, uconfig_file)
+        if not uconfig_no_def_match:
+            print('No definition for flag %s found!\n' % uconfig_no_flag)
+            sys.exit(1)
+
+    test_results = {f: {'unit_test': False, 'hdr_test': False} for f in uconfig_no_flags_all}
+    test_results['all_flags'] = {'unit_test': False, 'hdr_test' : False}
+
+    return uconfig_no_flags_all, test_results
+
+def BuildAllFlags(uconfig_no_list):
+    """Builds sequence of -Dflag=1 with each flag from the list."""
+
+    flag_list = ['-D' + uconfig_no + '=1' for uconfig_no in uconfig_no_list]
+
+    return ' '.join(flag_list)
+
+def RunUnitTests(uconfig_no_list, test_results):
+    """Iterates over all flags, sets each individually during ICU configuration
+       and executes the ICU4C unit tests.
+
+    Args:
+      uconfig_no_list: list of all UCONFIG_NO_XXX flags to test with.
+      test_results: dictionary to record test run results.
+
+    Returns:
+      test_results: updated test result entries.
+    """
+
+    for uconfig_no in uconfig_no_list:
+        _, exit_code = RunCmd(
+                './runConfigureICU Linux CPPFLAGS=\"-D%s=1"' % uconfig_no)
+        if exit_code != 0:
+            print('ICU4C configuration for flag %s failed' % uconfig_no)
+            sys.exit(1)
+        print('Running unit tests with %s set to 1.' % uconfig_no)
+        _, exit_code = RunCmd('make -j2 check')
+        test_results[uconfig_no]['unit_test'] = (exit_code == 0)
+        RunCmd('make clean')
+
+    # Configure ICU with all UCONFIG_NO_XXX flags set to 1 and execute
+    # the ICU4C unit tests.
+    all_unit_test_config_no = BuildAllFlags(uconfig_no_list)
+    _, exit_code = RunCmd(
+        'CPPFLAGS=\"%s\" ./runConfigureICU Linux' % all_unit_test_config_no)
+    if exit_code != 0:
+        print('ICU configuration with all flags set failed')
+        sys.exit(1)
+    print('Running unit tests with all flags set to 1.')
+    _, exit_code = RunCmd('make -j2 check')
+    test_results['all_flags']['unit_test'] = (exit_code == 0)
+    RunCmd('make clean')
+
+    return test_results
+
+def RunHeaderTests(uconfig_no_list, test_results):
+    """Iterates over all flags and executes the header test.
+
+    Args:
+      uconfig_no_list: list of all UCONFIG_NO_XXX flags to test with.
+      test_results: dictionary to record test run results.
+
+    Returns:
+      test_results: updated test result entries.
+    """
+
+    # Header tests needs different setup.
+    RunCmd('mkdir /tmp/icu_cnfg')
+    out, exit_code = RunCmd('./runConfigureICU Linux --prefix=/tmp/icu_cnfg')
+    if exit_code != 0:
+        print('ICU4C configuration for header test failed!')
+        print(out)
+        sys.exit(1)
+
+    _, exit_code = RunCmd('make -j2 install')
+    if exit_code != 0:
+        print('make install failed!')
+        sys.exit(1)
+
+    for uconfig_no in uconfig_no_list:
+        print('Running header tests with %s set to 1.' % uconfig_no)
+        _, exit_code = RunCmd(
+            'PATH=/tmp/icu_cnfg/bin:$PATH make -C test/hdrtst UCONFIG_NO=\"-D%s=1\" check' % uconfig_no)
+        test_results[uconfig_no]['hdr_test'] = (exit_code == 0)
+
+    all_hdr_test_flags = BuildAllFlags(uconfig_no_list)
+    print('Running header tests with all flags set to 1.')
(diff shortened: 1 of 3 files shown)


jefgen:
>Total execution time on GHA around two hours.

Hmm... I wonder, should this really be run on *every single PR* if the testing takes that long?

For comparison, the exhaustive tests take about ~1.5 hours, and IIRC that was considered too long for PRs -- so they were made to only run on the `master` branch *after* PRs get merged.

I wonder if perhaps we could try to do the same with the `UCONFIG_*` tests maybe?

gnrunge:
> > Total execution time on GHA around two hours.
> 
> Hmm... I wonder, should this really be run on _every single PR_ if the testing takes that long?
> 
> For comparison, the exhaustive tests take about ~1.5 hours, and IIRC that was considered too long for PRs -- so they were made to only run on the `master` branch _after_ PRs get merged.
> 
> I wonder if perhaps we could try to do the same with the `UCONFIG_*` tests maybe?

The uconfig variation BRS task has a high catch rate, i.e. it shows issues at almost every release (not for 69, for once). As such if it shows an issue as part of PR CI that would be beneficial for the PR author right away.

Possible option: one of the test scenarios is to set all UCONFIG_NO_* to 1 and execute unit test and header test. That is less than 10-15 minutes and has already a chance to catch an issue with a UCONFIG_NO_* flag. There may be constellations in which these flags protect each other and the tests run through, though, I don't know exactly. Run the 'all set' for every PR and the full test, individual set flags to 1, after the PR merge.

gnrunge:
Friendly ping!

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+        if uconfig_no_flag.group(0) not in uconfig_no_flags_all:
+            uconfig_no_flags_all.append(uconfig_no_flag.group(0))
+
+    test_results = {f: {'unit_test': False, 'hdr_test': False} for f in uconfig_no_flags_all}
+    test_results['all_flags'] = {'unit_test': False, 'hdr_test' : False}
+
+    return uconfig_no_flags_all, test_results
+
+def EnableUConfigNo(uconfig_file, uconfig_no_flag):
+    """Sets the uconfig_no_flag in common/unicode/uconfig.h file.
```
markusicu, line 93: It should not be necessary to rewrite uconfig.h. The config switches are all guarded by "if not defined, then define turnoff=0". So instead it would be simpler and cleaner (and require much less code) to just specify the relevant switch as a `-Dflag=1` option on the configure command line.

The symlink hack should also not be necessary.

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+        sys.exit(-1)
+    uconfig_no_def = uconfig_no_def_match.group(0)
+    set_uconfig_no_def = uconfig_no_def.replace(' 0\n', ' 1\n')
+    set_in_uconfig_file = uconfig_file.replace(uconfig_no_def, set_uconfig_no_def)
+
+    return set_in_uconfig_file
+
+def main():
+    # Read the options and determine what to run.
+    run_hdr = False
```
markusicu, line 117: optional: If we wanted to reduce the runtime further, we could have more actions that each test fewer combinations. This might mean hardcoding the list of options in the form of one GHA each (and some test to make sure we don't miss new options). So... probably not now. :-)

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # Iterate over all flags, set each individually one in uconfig.h and
+    # execute ICU4C unit tests or/and header test.
+    for uconfig_no in all_uconfig_no_flags:
+        with open('common/unicode/uconfig.h', 'w') as file_handle:
+            file_handle.write(EnableUConfigNo(orig_uconfig_file, uconfig_no))
+        RunCmd('make clean')
+
+        # Run ICU4C unit tests if requested, except for the excluded flags.
+        if run_unit and uconfig_no not in excluded_unit_test_flags:
```
markusicu, line 143:
```suggestion
        if run_unit and (uconfig_no not in excluded_unit_test_flags):
```

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+            print('Running unit tests with %s set to 1.' % uconfig_no)
+            _, exit_code = RunCmd('make -j2 check')
+            if exit_code == 0:
+                test_results[uconfig_no]['unit_test'] = True
+            else:
+                test_results[uconfig_no]['unit_test'] = False
```
markusicu, line 149: You should be able to replace this with
```
test_results[uconfig_no]['unit_test'] = (exit_code == 0)
```
with or without the parentheses to your liking

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+
+        # If requested, run header tests.
+        if run_hdr:
+            print('Running header tests with %s set to 1.' % uconfig_no)
+            _, exit_code = RunCmd(
+                'PATH=/tmp/icu_cnfg/bin:$PATH make -C test/hdrtst check')
+            if exit_code == 0:
+                test_results[uconfig_no]['hdr_test'] = True
+            else:
+                test_results[uconfig_no]['hdr_test'] = False
```
markusicu, line 159: ditto

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+        with open('common/unicode/uconfig.h', 'w') as file_handle:
+            file_handle.write(uconfig_no_all)
+        RunCmd('make clean')
+
+        print('Running unit tests with all flags set to 1.')
+        _, exit_code = RunCmd('make -j2 check')
+        if exit_code == 0:
+            test_results['all_flags']['unit_test'] = True
+        else:
+            test_results['all_flags']['unit_test'] = False
```
markusicu, line 177: ditto

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+            uconfig_no_all = EnableUConfigNo(uconfig_no_all, uconfig_no)
+        with open('common/unicode/uconfig.h', 'w') as file_handle:
+            file_handle.write(uconfig_no_all)
+        print('Running header tests with all flags set to 1.')
+        hdr_test_result_all, exit_code = RunCmd(
+            'PATH=/tmp/icu_cnfg/bin:$PATH make -C test/hdrtst check')
+        if exit_code == 0:
+            test_results['all_flags']['hdr_test'] = True
+        else:
+            test_results['all_flags']['hdr_test'] = False
```
markusicu, line 192: last one

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # Review test results and report any failures.
+    outcome = 0
+    print('Summary:\n')
+    for uconfig_no in all_uconfig_no_flags:
+        if run_unit and uconfig_no not in excluded_unit_test_flags:
+            if test_results[uconfig_no]['unit_test'] == False:
```
markusicu, line 199: Please change all `condition == False` to `not condition`.
And please add parentheses around that if it's after an `other_condition and ...`

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,217 @@
...
+        print('Running header tests with all flags set to 1.')
+        hdr_test_result_all, exit_code = RunCmd(
+            'PATH=/tmp/icu_cnfg/bin:$PATH make -C test/hdrtst check')
+        if exit_code == 0:
+            test_results['all_flags']['hdr_test'] = True
+        else:
+            test_results['all_flags']['hdr_test'] = False
+
+    # Review test results and report any failures.
+    outcome = 0
```
markusicu, line 195: Why is this a number?
I suggest `ok = True` and later setting `ok = False` and testing `if ok:`

markusicu:
> Friendly ping!

I was hoping Jeff would continue ;-)
Happy to take over.

tools/scripts/uconfig_vars_test.py
gnrunge: Underlying reason for doing it like this is runtime savings. Currently runConfigureICU and 'make install' are executed once only during setup.

markusicu: Hm. I didn't appreciate/remember that the hdrtst runs after `make install`.

Looking at the various Makefile.in files, each of the three libraries has an "install-headers" target. So for running hdrtst, it should suffice to configure, and then, for each of common|i18n|io you should be able to cd there and `make install-headers`. Then run the hdrtst. No need to actually build and install the whole libraries, which should also be much faster. And if that all is fast enough, you should be able to configure for each uconfig_no_something flag, without rewriting uconfig.h.

And yes, for running unit tests, you should not need to `make install`.

I don't remember if you already `make clean` between flags -- please make sure you do.

tools/scripts/uconfig_vars_test.py
markusicu: Oh, you don't just check the outcome, you also return it in `sys.exit(outcome)`. That works, but was not obvious to me, sorry.

Apparently, the value you give to sys.exit() should be non-negative: https://docs.python.org/3/library/sys.html#sys.exit

I suggest you either
- add a comment at the definition of `outcome` that this value will be returned via sys.exit() and that 0 means "success", and use value 1 for errors; or
- change it to boolean `ok`, and later use "if ok then exit(0) else print and exit(1)"

tools/scripts/uconfig_vars_test.py
gnrunge: The 'make install' is in the instructions (https://sites.google.com/site/icusite/processes/release/tasks/healthy-code?pli=1#TOC-Test-ICU4C-headers). I once tried to do without, had problems, and just continued as instructed.

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,219 @@
...
+
+    return uconfig_no_flags_all, test_results
+
+def BuildAllFlags(uconfig_no_list):
+    """Builds sequence of -Dflag=1 with each flag from the list."""
+
+    flag_list = ['-D' + uconfig_no for uconfig_no in uconfig_no_list]
+    flag_value_list = [flag + '=1' for flag in flag_list]
+
+    return ' '.join(flag_value_list)
```
markusicu, line 91:
```suggestion
    flag_list = ['-D' + uconfig_no + '=1' for uconfig_no in uconfig_no_list]
    return ' '.join(flag_list)
```

tools/scripts/uconfig_vars_test.py
gnrunge: Changed, the UCONFIG_NO_XXX flag is no passed in via runConfigureICU (unit tests) or test/hdrtst/Makefile; test/hdrtst/Makefile.in adjusted accordingly.

tools/scripts/uconfig_vars_test.py
gnrunge: Done.

tools/scripts/uconfig_vars_test.py
gnrunge: Done throughout.

tools/scripts/uconfig_vars_test.py
gnrunge: Done.

tools/scripts/uconfig_vars_test.py
gnrunge: Done.

tools/scripts/uconfig_vars_test.py
gnrunge: Done throughout.

tools/scripts/uconfig_vars_test.py
gnrunge: Done, also changed '-1' --> '1'.

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,219 @@
...
+    all_unit_test_config_no = BuildAllFlags(uconfig_no_list)
+    _, exit_code = RunCmd(
+        'CPPFLAGS=\"%s\" ./runConfigureICU Linux' % all_unit_test_config_no)
+    if exit_code != 0:
+        print('ICU configuration with all flags set failed')
+        sys.exit(1)
+    print('Running unit tests with all flags set to 1.')
+    _, exit_code = RunCmd('make -j2 check')
+    test_results['all_flags']['unit_test'] = (exit_code == 0)
+
```
markusicu, line 127: do another `make clean` here?

tools/scripts/uconfig_vars_test.py
```diff
@@ -0,0 +1,219 @@
...
+    Args:
+      uconfig_no_list: list of all UCONFIG_NO_XXX flags to test with.
+      test_results: dictionary to record test run results.
+
+    Returns:
+      test_results: updated test result entries.
+    """
+
+    # Header tests needs different setup/
```
markusicu, line 141:
```suggestion
    # Header tests needs different setup.
```

markusicu (commented):
nice changes!
minor comments below

tools/scripts/uconfig_vars_test.py
gnrunge: Done.

tools/scripts/uconfig_vars_test.py
gnrunge: Right, done.

tools/scripts/uconfig_vars_test.py
gnrunge: Added another make clean.
Not really needed I think but sets the client back to a defined state.

markusicu (approved):
lgtm tnx

tools/scripts/uconfig_vars_test.py
gnrunge: Done.

PR #1593, opened 2021-02-23
ICU-21503 Adds status checks to prevent segmentation fault when test …

gnrunge:
…runs

with stubdata only.

##### Checklist

- [ ] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21503
- [ ] Issue accepted
- [ ] Tests included

### DIFF
--- a/icu4c/source/test/intltest/normconf.cpp
+++ b/icu4c/source/test/intltest/normconf.cpp
@@ -282,9 +282,12 @@ void NormalizerConformanceTest::TestConformance(FileStream *input, int32_t optio
 
 namespace {
 
-UBool isNormalizedUTF8(const Normalizer2 &norm2, const UnicodeString &s, UErrorCode &errorCode) {
+UBool isNormalizedUTF8(const Normalizer2 *norm2, const UnicodeString &s, UErrorCode &errorCode) {
+    if (norm2 == nullptr) {
+        return true;
+    }
     std::string s8;
-    return norm2.isNormalizedUTF8(s.toUTF8String(s8), errorCode);
+    return norm2->isNormalizedUTF8(s.toUTF8String(s8), errorCode);
 }
 
 }  // namespace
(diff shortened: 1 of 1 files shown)


icu4c/source/test/intltest/normconf.cpp
```diff
@@ -366,6 +366,9 @@ UBool NormalizerConformanceTest::checkConformance(const UnicodeString* field,
         }
     }
     if(options==0 && !isNormalizedUTF8(*nfd, field[2], status)) {
+        if(U_FAILURE(status)) {
```
markusicu, line 369: 1. The NFC/NFD data is hardcoded in ICU4C, so this should never fail for lack of data.
2. Unlike other places where we need to check the UErrorCode because the returned string/bundle/object/... is a nullptr, no object is returned here so we can't crash for that kind of reason. On which line does the crash occur when there is no data?

gnrunge: Core was generated by `./intltest -w'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  (anonymous namespace)::isNormalizedUTF8 (norm2=..., s=..., errorCode=@0x7ffdbe221024: U_FILE_ACCESS_ERROR) at normconf.cpp:287
287	    return norm2.isNormalizedUTF8(s.toUTF8String(s8), errorCode);

The fatal call to isNormalizedUTF8(...) comes from
    if(options==0 && !isNormalizedUTF8(*nfkd, field[4], status)) {
 on line 394. When I take out the first 'return FALSE;' in line 370 execution continues till 394.

markusicu: Thanks, got it.

The class members nfc, nfd, nfkc, nfkd are pointers which are initialized in the constructor. When data is missing, then one or more of these is a nullptr and the constructor outputs a failure. It does not retain the UErrorCode for the test functions that are called later, but we can check for nullptr, more or less like the checkNorm() function already does on line 482.

So first please revert your current changes.

I think the simplest fix is to modify isNormalizedUTF8() (lines 285..288) to look like this:
```
UBool isNormalizedUTF8(const Normalizer2 *norm2, const UnicodeString &s, UErrorCode &errorCode) {
    if (norm2 == nullptr) {
        return true;
    }
    std::string s8;
    return norm2->isNormalizedUTF8(s.toUTF8String(s8), errorCode);
}
```
and, because this changes the function signature, then search for any `isNormalizedUTF8(*nf` and replace with `isNormalizedUTF8(*nf` (removing the `*`).

markusicu (approved):
lgtm tnx

PR #1597, opened 2021-02-24
ICU-20941 Port Arbitrary Units support from ICU4C to ICU4J

hugovdm:
Closes: icu-units#146

Makes some simultaneous tweaks/fixes to ICU4C.

Commit history, pre-squash: https://github.com/hugovdm/icu/tree/pr1597/presquash
* I have commits that have copy/pasted C++ code (that doesn't compile) followed by commits that convert that to Java, such that the finer-grained git commit history _might_ be easier to review than comparing it side-by-side with ICU4C.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-20941
- [X] Issue accepted
- [X] Tests included

### DIFF
--- a/icu4c/source/i18n/number_longnames.cpp
+++ b/icu4c/source/i18n/number_longnames.cpp
@@ -143,35 +143,32 @@ enum PlaceholderPosition { PH_EMPTY, PH_NONE, PH_BEGINNING, PH_MIDDLE, PH_END };
  *   found.
  * @param joinerChar Iff the placeholder was at the beginning or end, joinerChar
  *   contains the space character (if any) that separated the placeholder from
- *   the rest of the pattern. Otherwise, joinerChar is set to NUL.
+ *   the rest of the pattern. Otherwise, joinerChar is set to NUL. Only one
+ *   space character is considered.
  */
 void extractCorePattern(const UnicodeString &pattern,
                         UnicodeString &coreUnit,
                         PlaceholderPosition &placeholderPosition,
                         UChar &joinerChar) {
     joinerChar = 0;
+    int32_t len = pattern.length();
     if (pattern.startsWith(u"{0}", 3)) {
         placeholderPosition = PH_BEGINNING;
         if (u_isJavaSpaceChar(pattern[3])) {
             joinerChar = pattern[3];
-            coreUnit.setTo(pattern, 4, pattern.length() - 4);
-            // Expecting no double spaces
-            U_ASSERT(!u_isJavaSpaceChar(pattern[4]));
+            coreUnit.setTo(pattern, 4, len - 4);
         } else {
-            coreUnit.setTo(pattern, 3, pattern.length() - 3);
+            coreUnit.setTo(pattern, 3, len - 3);
         }
     } else if (pattern.endsWith(u"{0}", 3)) {
         placeholderPosition = PH_END;
-        int32_t len = pattern.length();
         if (u_isJavaSpaceChar(pattern[len - 4])) {
-            coreUnit.setTo(pattern, 0, pattern.length() - 4);
+            coreUnit.setTo(pattern, 0, len - 4);
             joinerChar = pattern[len - 4];
-            // Expecting no double spaces
-            U_ASSERT(!u_isJavaSpaceChar(pattern[len - 5]));
         } else {
-            coreUnit.setTo(pattern, 0, pattern.length() - 3);
+            coreUnit.setTo(pattern, 0, len - 3);
         }
-    } else if (pattern.indexOf(u"{0}", 0, 1, pattern.length() - 2) == -1) {
+    } else if (pattern.indexOf(u"{0}", 3, 1, len - 2) == -1) {
         placeholderPosition = PH_NONE;
         coreUnit = pattern;
     } else {
@@ -361,7 +358,7 @@ void getInflectedMeasureData(StringPiece subKey,
     key.append(subKey, status);
 
     UErrorCode localStatus = status;
-    ures_getAllItemsWithFallback(unitsBundle.getAlias(), key.data(), sink, status);
+    ures_getAllItemsWithFallback(unitsBundle.getAlias(), key.data(), sink, localStatus);
     if (width == UNUM_UNIT_WIDTH_SHORT) {
         status = localStatus;
         return;
@@ -777,11 +774,6 @@ void LongNameHandler::forMeasureUnit(const Locale &loc,
     //
     // 1. If the unitId is empty or invalid, fail
     // 2. Put the unitId into normalized order
-    //
-    // We just need to check if it is a MeasureUnit this constructor handles:
-    // this constructor does not handle mixed units
-    U_ASSERT(uprv_strcmp(unitRef.getType(), "") != 0 ||
-             unitRef.getComplexity(status) != UMEASURE_UNIT_MIXED);
     U_ASSERT(fillIn != nullptr);
 
     if (uprv_strcmp(unitRef.getType(), "") != 0) {
@@ -812,6 +804,9 @@ void LongNameHandler::forMeasureUnit(const Locale &loc,
         // fillIn->parent = parent;
         // return;
     } else {
+        // Check if it is a MeasureUnit this constructor handles: this
+        // constructor does not handle mixed units
+        U_ASSERT(unitRef.getComplexity(status) != UMEASURE_UNIT_MIXED);
         forArbitraryUnit(loc, unitRef, width, unitDisplayCase, fillIn, status);
         fillIn->rules = rules;
         fillIn->parent = parent;
@@ -1070,7 +1065,7 @@ void LongNameHandler::processPatternTimes(MeasureUnitImpl &&productUnit,
             getInflectedMeasureData(dimensionalityKey.toStringPiece(), loc, width, gender,
                                     singleCaseVariant, dimensionalityPrefixPatterns, status);
             if (U_FAILURE(status)) {
-                // At the time of writing, only power2 and power3 are supported.
+                // At the time of writing, only pow2 and pow3 are supported.
                 // Attempting to format other powers results in a
                 // U_RESOURCE_TYPE_MISMATCH. We convert the error if we
                 // understand it:
@@ -1410,6 +1405,7 @@ void MixedUnitLongNameHandler::forMeasureUnit(const Locale &loc,
                                               const MicroPropsGenerator *parent,
                                               MixedUnitLongNameHandler *fillIn,
                                               UErrorCode &status) {
+    U_ASSERT(mixedUnit.getComplexity(status) == UMEASURE_UNIT_MIXED);
     U_ASSERT(fillIn != nullptr);
     if (U_FAILURE(status)) {
         return;
@@ -1417,6 +1413,7 @@ void MixedUnitLongNameHandler::forMeasureUnit(const Locale &loc,
 
     MeasureUnitImpl temp;
     const MeasureUnitImpl &impl = MeasureUnitImpl::forMeasureUnit(mixedUnit, temp, status);
+    // Defensive, for production code:
     if (impl.complexity != UMEASURE_UNIT_MIXED) {
         // Should be using the normal LongNameHandler
         status = U_UNSUPPORTED_ERROR;
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
@@ -120,6 +120,29 @@ public final class PatternProps {
     }
 
     /**
+     * @return s except with leading and trailing SpaceChar characters removed.
+     */
+    public static String trimSpaceChar(String s) {
+        if (s.length() == 0 ||
+            (!Character.isSpaceChar(s.charAt(0)) && !Character.isSpaceChar(s.charAt(s.length() - 1)))) {
+            return s;
+        }
+        int start = 0;
+        int limit = s.length();
+        while (start < limit && Character.isSpaceChar(s.charAt(start))) {
+            ++start;
+        }
+        if (start < limit) {
+            // There is non-SpaceChar at start; we will not move limit below that,
+            // so we need not test start<limit in the loop.
+            while (isWhiteSpace(s.charAt(limit - 1))) {
+                --limit;
+            }
+        }
+        return s.substring(start, limit);
+    }
+
+    /**
      * Tests whether the CharSequence contains a "pattern identifier", that is,
      * whether it contains only non-Pattern_White_Space, non-Pattern_Syntax characters.
      * @return true if there are no Pattern_White_Space or Pattern_Syntax characters in s.
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
@@ -2,6 +2,7 @@
 // License & terms of use: http://www.unicode.org/copyright.html
 package com.ibm.icu.impl.number;
 
+import java.util.ArrayList;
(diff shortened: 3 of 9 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
```diff
@@ -119,6 +119,29 @@ public static String trimWhiteSpace(String s) {
         return s.substring(start, limit);
     }
```
hugovdm, line 121: trimSpaceChar is a tweaked copy of trimWhiteSpace - formatted with clang-format rather than keeping consistent formatting.

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -185,15 +433,108 @@ private static String getPerUnitFormat(ULocale locale, UnitWidth width) {
...
+                } else {
+                    this.value1 = value;
+                }
+            } catch (Exception e) {
+                // Fall back to uninflected.
+            }
+        }
+
+        // FIXME(review): is "getValue0" more idiomatic?
+        String value0(String compoundValue) {
```
hugovdm, line 513: Would getValue0() and getValue1() be more idiomatic? (For both ICU4C and ICU4J?)

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -361,7 +358,7 @@ void getInflectedMeasureData(StringPiece subKey,
     key.append(subKey, status);
 
     UErrorCode localStatus = status;
-    ures_getAllItemsWithFallback(unitsBundle.getAlias(), key.data(), sink, status);
+    ures_getAllItemsWithFallback(unitsBundle.getAlias(), key.data(), sink, localStatus);
```
hugovdm, line 361: Found a bug: failure here would break proper fallback.

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -1410,13 +1405,15 @@ void MixedUnitLongNameHandler::forMeasureUnit(const Locale &loc,
...
+    U_ASSERT(mixedUnit.getComplexity(status) == UMEASURE_UNIT_MIXED);
     U_ASSERT(fillIn != nullptr);
     if (U_FAILURE(status)) {
         return;
     }
 
     MeasureUnitImpl temp;
     const MeasureUnitImpl &impl = MeasureUnitImpl::forMeasureUnit(mixedUnit, temp, status);
+    // Defensive, for production code:
     if (impl.complexity != UMEASURE_UNIT_MIXED) {
```
hugovdm, line 1417: Since we don't get read-only access to MeasureUnitImpl in ICU4J, we can't do this quick/cheap check for complexity in production code.

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -1410,13 +1405,15 @@ void MixedUnitLongNameHandler::forMeasureUnit(const Locale &loc,
                                               const MicroPropsGenerator *parent,
                                               MixedUnitLongNameHandler *fillIn,
                                               UErrorCode &status) {
+    U_ASSERT(mixedUnit.getComplexity(status) == UMEASURE_UNIT_MIXED);
```
hugovdm, line 1408: Since we don't get read-only access to MeasureUnitImpl in ICU4J, we can't do the quick/cheap check for impl.complexity there that we have below (line 1417). Instead, for ICU4J, I just trust the invariant, checked in debug mode via an assert.

Here in ICU4C, I've added the same assert, for some more consistency. Assert failures are anyway also much easier to debug in ICU4C than status-based failures.

icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
```diff
@@ -119,6 +119,29 @@ public static String trimWhiteSpace(String s) {
...
 
+    /**
+     * @return s except with leading and trailing SpaceChar characters removed.
+     */
+    public static String trimSpaceChar(String s) {
```
hagbard, line 125: What sore of semantics do you want here?
This *looks* like it's unicode friendly except it's calling "isSpaceChar(char)" and not "isSpaceChar(int)":

https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isSpaceChar(int)

If you're only looking for ASCII this can be done far neater/smaller using "indexOf()"/"lastIndexOf()" and if you're trying to be Unicode friendly, use "isSpaceChar(int)".

Also, if performance isn't an issue in the case where

icu4j/main/classes/core/src/com/ibm/icu/impl/number/MixedUnitLongNameHandler.java
```diff
@@ -61,6 +61,9 @@ public static MixedUnitLongNameHandler forMeasureUnit(ULocale locale,
                                                           String unitDisplayName,
                                                           PluralRules rules,
                                                           MicroPropsGenerator parent) {
+        // In Java we rely on the assert. In ICU4C, where we can efficiently
```
hagbard, line 64: What does "we rely on the assert" mean? assert is mostly ignored at runtime. 

From https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html:
""By default, assertions are disabled at runtime.""

Since this is a public method anyone can call, assert is simply the wrong thing to use here. An assert means "internal error in my code", not "someone called a method with unexpected arguments".

icu4j/main/classes/core/src/com/ibm/icu/number/NumberFormatterImpl.java
```diff
@@ -412,6 +412,14 @@ private static MicroPropsGenerator macrosToMicroGenerator(MacroProps macros, Mic
                 MeasureUnit unit = macros.unit;
                 if (macros.perUnit != null) {
                     unit = unit.product(macros.perUnit.reciprocal());
+                    // This isn't strictly necessary, but was what we specced
+                    // out when perUnit became a backward-compatibility thing:
+                    // unit/perUnit use case is only valid if both units are
+                    // built-ins, or the product is a built-in.
+                    if (unit.getType() == null && (macros.unit.getType() == null || macros.perUnit.getType() == null)) {
+                        throw new UnsupportedOperationException(
```
hagbard, line 420: UnsupportedOperationException might not be the best exception here.
IllegalStateException is more for when data expectations are violated.
UnsupportedOperationException is more for "this class doesn't support a specific feature" (e.g. writing values to an immutable array).

hagbard (commented):
I know you want this in urgently, but there are a couple of things worth tweaking.

icu4c/source/i18n/number_longnames.cpp
younies: +1

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -61,14 +67,241 @@ static String getWithPlural(String[] strings, StandardPlural plural) {
...
+    }
+
+    /**
+     * Returns three outputs extracted from pattern.
+     *
+     * @param coreUnit is extracted as per Extract(...) in the spec:
+     *   https://unicode.org/reports/tr35/tr35-general.html#compound-units
+     * @param PlaceholderPosition indicates where in the string the placeholder
+     *   was found.
+     * @param joinerChar Iff the placeholder was at the beginning or end,
```
younies, line 85: do you mean `If` or `If and only if`

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -2095,39 +2243,57 @@ public TestCase(String locale, String unitIdentifier, String expectedGender) {
...
         FormattedNumber fn;
         for (TestCase t : cases) {
             // TODO(icu-units#140): make this work for more than just UnitWidth.FULL_NAME
+        //     formatter = NumberFormatter.with()
+        //                     .unit(MeasureUnit.forIdentifier(t.unitIdentifier))
+        //                     .locale(new ULocale(t.locale));
+        //     fn = formatter.format(1.1);
+        //     assertEquals("Testing gender with default width, unit: " + t.unitIdentifier +
+        //                      ", locale: " + t.locale,
+        //                  t.expectedGender, fn.getGender());
```
younies, line 2287: indentation

icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
hugovdm: Last sentence isn't complete?

Ah, surrogate pairs not supported, right?

* indexOf() / lastIndexOf() looks for one specific character, not sufficient.
* My very specific need is to trim space characters including the "NBSP": 
https://www.fileformat.info/info/unicode/char/00a0/index.htm - which is a SpaceChar but not Whitespace. isSpaceChar(char) seems adequate for my needs, but this should ideally behave consistently.

C++ code calls this:

```
U_CAPI UBool U_EXPORT2
u_isJavaSpaceChar(UChar32 c) {
    uint32_t props;
    GET_PROPS(c, props);
    return (UBool)((CAT_MASK(props)&U_GC_Z_MASK)!=0);
}
```

However the calling code is also just looking at one UChar at a time (16-bit). Here's my speculation now: maybe all space characters are within the Basic Multilingual Plane. So I'm tentatively asserting this code is Unicode friendly in the sense that it trims all Space Characters that are within the BMP?

In which case this change is unnecessary, but is the kind of code I think you're suggesting:

```
        if (s.length() == 0 || (!Character.isSpaceChar(s.codePointAt(0)) &&
                                !Character.isSpaceChar(s.codePointBefore(s.length())))) {
            return s;
        }
```

hugovdm: Do you think it necessary to document that it trims only BMP space characters?

icu4j/main/classes/core/src/com/ibm/icu/impl/number/MixedUnitLongNameHandler.java
hugovdm: ICU4J has public methods that are for internal use only. Additionally, classes that fall within the "impl" directories are also considered internal-use-only. (In public API classes, we mark internal-use-only methods with @Deprecated.)

I'm using an assert because I do want it ignored for production code - getComplexity() can be expensive, if it is a built-in unit, the unit's identifier has to be parsed.
 
 "assert false" does fail for "ant check", so works for my purposes.

hugovdm: As in: calling MixedUnitLongNameHandler with anything other than a mixed unit is indeed considered an internal error in ICU4J code.

It would have been very nice if MeasureUnit was a class hierarchy. Unfortunately it isn't. :-/

icu4j/main/classes/core/src/com/ibm/icu/number/NumberFormatterImpl.java
hugovdm: In NumberFormatterApiTest.java, the oldest reference I found to UnsupportedOperationException was this:

```
    static void assertUndefinedSkeleton(UnlocalizedNumberFormatter f) {
        try {
            String skeleton = f.toSkeleton();
            fail("Expected toSkeleton to fail, but it passed, producing: " + skeleton);
        } catch (UnsupportedOperationException expected) {}
    }
```
This is from three years ago (March 2018):
https://github.com/unicode-org/icu/commit/c2fa8cacad7fdf0c929b5686306c68cd88360837

I suspect consistency with this behaviour makes sense, and changing the behaviour would impact client code.

To illustrate by usage examples, these errors are the kind of thing that client code *does* have access to. These lines are fine - capitals for variables I'm not defining here:
```
nf = NumberFormatter.forSkeleton("unit/hectometer");
nf = NumberFormatter.forSkeleton("unit/hectometer-per-second");
nf = NumberFormatter.unit(HECTOMETER);
nf = NumberFormatter.unit(HECTOMETER_PER_SECOND);
```

These are unsupported, and throw an UnsupportedOperationException:
```
nf = NumberFormatter.forSkeleton("unit/hectometer per-unit/duration-second");
nf = NumberFormatter.unit(HECTOMETER).perUnit(SECOND);
```

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
hugovdm: I mean "if and only if".

hugovdm: (This doesn't end up in [ICU4J API documentation](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/).)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
hugovdm: Done.

(VSCode's doing that one wrong, so I did it manually. I have no idea how to get VSCode to format better by default - e.g. how to make it follow clang-format style...)

icu4j/main/classes/core/src/com/ibm/icu/impl/number/MixedUnitLongNameHandler.java
hugovdm: I've tried to improve this error message. Does it look better now?

younies (approved):

hugovdm:
Squashing optimistically...

younies (approved):

icu4j/main/classes/core/src/com/ibm/icu/impl/PatternProps.java
sffc: Whitespace is all in the BMP.  If that ever changes, it will break a lot of assumptions throughout ICU and elsewhere.

hugovdm: 17 current "space separators" all in the BMP:
https://www.compart.com/en/unicode/category/Zs

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -61,14 +67,241 @@ static String getWithPlural(String[] strings, StandardPlural plural) {
...
+        } else {
+            key.append(builtinUnit.getSubtype());
+        }
+        key.append("/gender");
+
+        try {
+            ICUResourceBundle stackBundle =
+                (ICUResourceBundle)unitsBundle.getWithFallback(key.toString());
+            return stackBundle.getString();
+        } catch (Exception e) {
```
sffc, line 147: Optional: it is more idiomatic in Java to catch specific types of exceptions

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
sffc: It doesn't really matter for an internal class.  What you wrote is fine.

sffc (approved):

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
hugovdm: Discussed, and briefly surveyed other "catch (Exception)" cases in ICU4J (more than 50, while there are less than 90 "catch (MissingResourceError)" cases, for a little bit of an apples&oranges contrast. ("Fruitbowl contrast"? :-P)

PR #1601, opened 2021-02-24
ICU-21480 Update French unit-times pattern in unit tests

hugovdm:
This PR touches only ICU4C, because the equivalent ICU4J PR is nr 1597, which is not finished yet.

##### Checklist

- [X] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21480
- [X] Issue accepted
- [X] Tests included


jefgen (approved):
thanks!

pedberg-icu (approved):

PR #1608, opened 2021-02-28
ICU-21513 check if TARGET_OS_SIMULATOR has been defined

mojca:
clang 11 throws an error when using undefined variables starting with
TARGET_OS_, and TARGET_OS_SIMULATOR is missing in earlier versions
of macOS, so the code would not compile without an additional guard.

See also ICU-13440.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21513

### DIFF
--- a/icu4c/source/common/putil.cpp
+++ b/icu4c/source/common/putil.cpp
@@ -1361,7 +1361,7 @@ uprv_pathIsAbsolute(const char *path)
 
 /* Backup setting of ICU_DATA_DIR_PREFIX_ENV_VAR
    (needed for some Darwin ICU build environments) */
-#if U_PLATFORM_IS_DARWIN_BASED && TARGET_OS_SIMULATOR
+#if U_PLATFORM_IS_DARWIN_BASED && defined(TARGET_OS_SIMULATOR) && TARGET_OS_SIMULATOR
 # if !defined(ICU_DATA_DIR_PREFIX_ENV_VAR)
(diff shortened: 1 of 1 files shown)


icu4c/source/common/putil.cpp
```diff
@@ -1361,7 +1361,7 @@ uprv_pathIsAbsolute(const char *path)
 
 /* Backup setting of ICU_DATA_DIR_PREFIX_ENV_VAR
    (needed for some Darwin ICU build environments) */
-#if U_PLATFORM_IS_DARWIN_BASED && TARGET_OS_SIMULATOR
+#if U_PLATFORM_IS_DARWIN_BASED && defined(TARGET_OS_SIMULATOR) && TARGET_OS_SIMULATOR
```
mojca, line 1364: In theory this could have been
```
#if U_PLATFORM_IS_DARWIN_BASED && (defined(TARGET_OS_SIMULATOR) && TARGET_OS_SIMULATOR) || (defined(TARGET_IPHONE_SIMULATOR) && TARGET_IPHONE_SIMULATOR)
```
but I doubt that it makes much sense to compile the latest ICU for iPhone on those ancient OSes.

(Not that I know what this code is good for or that I bothered checking when it became deprecated; the above comment is pure speculation.)

richgillam (approved):
It's possible @pedberg-icu will have thoughts about whether there's a better way to do this, but as far as I can tell, this change appears to be both reasonable and harmless, so even if there is a better way, I don't think we need to keep this change from being integrated.

pedberg-icu (approved):

PR #1618, opened 2021-03-04
ICU-21521 Fix cast of uprv_strcmp

FrankYFTang:
uprv_strcmp is strcmp which return int by C++ standard
On arm64-v8a or other implementation strcmp may return positive int > 127 and cause the cast to int8_t as negative number.
Return only 0, 1, -1 to solve the issue.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21521

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -1577,8 +1577,12 @@ AliasReplacer::replaceTransformedExtensions(
             tkey = nextTKey;
         } while (tkey != nullptr);
         tfields.sort([](UElement e1, UElement e2) -> int8_t {
-            return uprv_strcmp(
+            // uprv_strcmp return int and in some platform, such as arm64-v8a,
+            // it may return positive values > 127 which cause the casted value
+            // of int8_t negative.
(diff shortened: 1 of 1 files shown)


sffc:
It seems like it would be better to change `uprv_strcmp` to "fix" the return value on different platforms to make it have the same semantics on all platforms, no?

markusicu (dismissed):
Change looks good.

optional: I doubt that this is an issue only on one platform. I would remove the reference to that platform but leave the rest of the int vs. int8_t comments.

markusicu:
For later, I suggest you submit another ticket to try to make the UElementComparator return an int32_t and remove code that artificially narrows values into the int8_t range.

FrankYFTang:
> submit another ticket to try to make the UElementComparator return an int32_t and remove code that artificially narrows values into the int8_t range.

https://unicode-org.atlassian.net/browse/ICU-21526

FrankYFTang:
PTAL

markusicu (approved):
lgtm

PR #1620, opened 2021-03-05
ICU-21123 Calculate built-in units' gender when CLDR lacks the data

hugovdm:
Closes icu-units#141

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21123

### DIFF
--- a/icu4c/source/common/uresimp.h
+++ b/icu4c/source/common/uresimp.h
@@ -270,11 +270,13 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
  * function can perform fallback on the sub-resources of the table.
  * @param resB              a resource
  * @param inKey             a key associated with the requested resource
+ * @param len               if not NULL, used to return the length of the string
  * @param status: fills in the outgoing error code
  *                could be <TT>U_MISSING_RESOURCE_ERROR</TT> if the key is not found
  *                could be a non-failing error 
  *                e.g.: <TT>U_USING_FALLBACK_WARNING</TT>,<TT>U_USING_DEFAULT_WARNING </TT>
- * @return                  a pointer to a UResourceBundle struct. If fill in param was NULL, caller must delete it
+ * @return returns a pointer to a zero-terminated UChar array which lives in a
+ *         memory mapped/DLL file.
  */
 U_CAPI const UChar* U_EXPORT2 
 ures_getStringByKeyWithFallback(const UResourceBundle *resB, 
--- a/icu4c/source/i18n/number_longnames.cpp
+++ b/icu4c/source/i18n/number_longnames.cpp
@@ -183,9 +183,10 @@ void extractCorePattern(const UnicodeString &pattern,
 
 // Gets the gender of a built-in unit: unit must be a built-in. Returns an empty
 // string both in case of unknown gender and in case of unknown unit.
-const char *getGenderForBuiltin(const Locale &locale, MeasureUnit builtinUnit, UErrorCode &status) {
+UnicodeString
+getGenderForBuiltin(const Locale &locale, const MeasureUnit &builtinUnit, UErrorCode &status) {
     LocalUResourceBundlePointer unitsBundle(ures_open(U_ICUDATA_UNIT, locale.getName(), &status));
-    if (U_FAILURE(status)) { return ""; }
+    if (U_FAILURE(status)) { return {}; }
 
     // Map duration-year-person, duration-week-person, etc. to duration-year, duration-week, ...
     // TODO(ICU-20400): Get duration-*-person data properly with aliases.
@@ -205,18 +206,18 @@ const char *getGenderForBuiltin(const Locale &locale, MeasureUnit builtinUnit, U
     key.append("/gender", status);
 
     UErrorCode localStatus = status;
-    StackUResourceBundle fillIn;
-    ures_getByKeyWithFallback(unitsBundle.getAlias(), key.data(), fillIn.getAlias(), &localStatus);
+    int32_t resultLen = 0;
+    const UChar *result =
+        ures_getStringByKeyWithFallback(unitsBundle.getAlias(), key.data(), &resultLen, &localStatus);
     if (U_SUCCESS(localStatus)) {
         status = localStatus;
-        UnicodeString directString = ures_getUnicodeString(fillIn.getAlias(), &status);
-        return getGenderString(directString, status);
+        return UnicodeString(true, result, resultLen);
     } else {
         // TODO(icu-units#28): "$unitRes/gender" does not exist. Do we want to
         // check whether the parent "$unitRes" exists? Then we could return
         // U_MISSING_RESOURCE_ERROR for incorrect usage (e.g. builtinUnit not
         // being a builtin).
-        return "";
+        return {};
     }
 }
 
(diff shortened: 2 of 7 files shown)


icu4c/source/common/uresimp.h
```diff
@@ -270,11 +270,13 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
  * function can perform fallback on the sub-resources of the table.
  * @param resB              a resource
  * @param inKey             a key associated with the requested resource
+ * @param len               if not NULL, used to return the length of the string
```
younies, line 273: NULL/``null_ptr``, 

Also, I prefer  the following comment:
`@ param len            the length of the string`

and if there is no string, return 0

icu4c/source/common/uresimp.h
```diff
@@ -270,11 +270,13 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
...
  * @param status: fills in the outgoing error code
  *                could be <TT>U_MISSING_RESOURCE_ERROR</TT> if the key is not found
  *                could be a non-failing error 
  *                e.g.: <TT>U_USING_FALLBACK_WARNING</TT>,<TT>U_USING_DEFAULT_WARNING </TT>
- * @return                  a pointer to a UResourceBundle struct. If fill in param was NULL, caller must delete it
+ * @return if not NULL, a pointer to a zero-terminated UChar array which lives
```
younies, line 278: optional.: ``null_ptr``

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -609,6 +609,102 @@ private static String getDeriveCompoundRule(ULocale locale, String feature, Stri
...
+        // a compound unit, even if unneeded. This algorithm does a lazier evaluation.
+
+        MeasureUnitImpl mui = unit.getCopyOfMeasureUnitImpl();
+        ArrayList<SingleUnitImpl> singleUnits = mui.getSingleUnits();
+        int singleUnitIndex = 0;
+        if (mui.getComplexity() == MeasureUnit.Complexity.COMPOUND) {
+            int startSlice = 0;
+            // inclusive
+            int endSlice = singleUnits.size() - 1;
+            assert endSlice > 0 : "COMPOUND units have more than one single unit";
```
younies, line 623: this assert could be moved to the MeasureUnit to ensure the compound units have more than one unit. I think it is too much to have this assert here

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -609,6 +609,102 @@ private static String getDeriveCompoundRule(ULocale locale, String feature, Stri
...
+        }
+        // Check for any SI and binary prefix gender override:
+        if (Math.abs(singleUnit.getDimensionality()) != 1) {
+            String prefixRule = getDeriveCompoundRule(locale, "gender", "prefix");
+            if (prefixRule.length() != 1) {
+                // Fixed gender for -powN- units
+                return prefixRule;
+            }
+            // prefixRule[0] == u'0'; u'1' not currently in spec.
+        }
```
younies, line 691: the function is too long already, I think moving this part of the code to a different function.

younies: Also, this part can be used later in another places.

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -778,6 +778,107 @@ const UChar *trimSpaceChars(const UChar *s, int32_t &length) {
...
+    // Check for any SI and binary prefix gender override:
+    if (std::abs(singleUnit->dimensionality) != 1) {
+        UnicodeString prefixRule = getDeriveCompoundRule(locale, "gender", "prefix", status);
+        if (prefixRule.length() != 1) {
+            // Fixed gender for -powN- units
+            return prefixRule;
+        }
+        // prefixRule[0] == u'0'; u'1' not currently in spec.
+    }
+    // Now we've boiled it down to the gender of one simple unit identifier:
```
younies, line 862: I think moving this part of code outside the function will make the function shorter and more readable. Also, this part can be used later in another function.

younies (dismissed):

icu4c/source/common/uresimp.h
hugovdm: I'd usually prefer "nullptr", but this is an old file and has lots of references to NULL - so I'll leave this as "NULL" for consistency.

Your suggestion is ambiguous: sounds like it's an input parameter rather than output.

I could say "returns the length of the string. Pass NULL if this is not needed" - but this feels unwieldy. I want to mention in the docs that len need not be provided, which is why I went with the "if not NULL" to start with. It sounds like my comment wasn't entirely clear to you though. I'm open to other suggestions as to how to make it clearer. :thinking:

icu4c/source/common/uresimp.h
hugovdm: Looking at examples from other documentation, and since people should check status for errors, not checking if the returned pointer is NULL, I've removed mention of potentially returning "NULL". Thus this is now ures_getString the same "@return" documentation as the ures_getString() this function calls.

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -183,9 +183,9 @@ void extractCorePattern(const UnicodeString &pattern,
 
 // Gets the gender of a built-in unit: unit must be a built-in. Returns an empty
 // string both in case of unknown gender and in case of unknown unit.
-const char *getGenderForBuiltin(const Locale &locale, MeasureUnit builtinUnit, UErrorCode &status) {
+UnicodeString getGenderForBuiltin(const Locale &locale, MeasureUnit builtinUnit, UErrorCode &status) {
```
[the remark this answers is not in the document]
hugovdm: Good catch, thanks!

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -609,6 +609,102 @@ private static String getDeriveCompoundRule(ULocale locale, String feature, Stri
     /// END DATA LOADING ///
     ////////////////////////
 
+    private static String calculateGenderForUnit(ULocale locale, MeasureUnit unit) {
```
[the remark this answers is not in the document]
hugovdm: Done!

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
hugovdm: The point of this assert is to locally document the invariants. I feel as far as this PR is concerned, MeasureUnit does what it does, we trust it or we don't trust it - so I could delete the assert, or keep the assert, no need to move it to MeasureUnit.

What this bit of documentation does, is alleviate the reader from wondering whether the value produced by the previous line, `int endSlice = singleUnits.size() - 1;`, is valid. What if size() is zero, and endSlice is thus -1? And then I call `singleUnits.get(endSlice)` in the next line? What does Java do when you call get(index) with index being out of bounds for an array? I don't know - ideally I would know, but this assert makes it clear that for the purpose of this code, I don't need to know.

Do you still think this assert is too much? Did you wonder about the value of endSlice when you reviewed this code?

icu4j/main/classes/core/src/com/ibm/icu/impl/number/LongNameHandler.java
hugovdm: This function is 82 lines. It encapsulates the whole algorithm for calculating gender.

You're suggesting factoring out something similar to `calculateGenderForSingleUnit()` would be useful to do - I'm not convinced yet. My thoughts:
* I can't imagine this new function being used anywhere else - so all it does is to give a "name" to the blank line at line 671 (and add another not-very-useful method to the list of LongNameHandler methods).
* Would giving it a function name, and calling that function, be more readable than the comment I went with instead: `// Now we know which singleUnit's gender we want`?

icu4c/source/i18n/number_longnames.cpp
hugovdm: Same as with LongNameHandler.java - I don't see how splitting it into two parts would make it more readable. Reading only half of an algorithm might feel more readable, but when an "easier to understand" feeling is achieved by simply "not understanding the whole algorithm", it's not really a benefit. (And as mentioned, the second half the algorithm doesn't seem useful to me in the absence of the first half.)

younies (dismissed):

icu4c/source/i18n/number_longnames.cpp
```diff
@@ -778,6 +779,107 @@ const UChar *trimSpaceChars(const UChar *s, int32_t &length) {
     return s + start;
 }
 
+UnicodeString calculateGenderForUnit(const Locale &locale, const MeasureUnit &unit, UErrorCode &status) {
```
sffc, line 782: Move the comment discussing the purpose of this function, and how it relates to forArbitraryUnit/processPatternTimes, to a doc comment, and elaborate a bit more.

sffc (approved):

PR #1621, opened 2021-03-05
ICU-21526 Change return of UElementComparator to int32_t

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21526

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -1576,13 +1576,8 @@ AliasReplacer::replaceTransformedExtensions(
             }
             tkey = nextTKey;
         } while (tkey != nullptr);
-        tfields.sort([](UElement e1, UElement e2) -> int8_t {
-            // uprv_strcmp return int and in some platform, such as arm64-v8a,
-            // it may return positive values > 127 which cause the casted value
-            // of int8_t negative.
-            int res = uprv_strcmp(
-                (const char*)e1.pointer, (const char*)e2.pointer);
-            return (res == 0) ? 0 : ((res > 0) ? 1 : -1);
+        tfields.sort([](UElement e1, UElement e2) -> int32_t {
+            return uprv_strcmp((const char*)e1.pointer, (const char*)e2.pointer);
         }, status);
         for (int32_t i = 0; i < tfields.size(); i++) {
              if (output.length() > 0) {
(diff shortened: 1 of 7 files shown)


FrankYFTang:
not sure we should land this for 69.1. maybe we should wait till the tree open of 70.1

markusicu:
> not sure we should land this for 69.1. maybe we should wait till the tree open of 70.1

Yes, this would be an enhancement (making code less brittle), not a bug fix.
And once we get there I will have comments :-)

icu4c/source/common/locid.cpp
```diff
@@ -1576,10 +1576,8 @@ AliasReplacer::replaceTransformedExtensions(
...
         } while (tkey != nullptr);
-        tfields.sort([](UElement e1, UElement e2) -> int8_t {
-            // uprv_strcmp return int and in some platform, such as arm64-v8a,
-            // it may return positive values > 127 which cause the casted value
-            // of int8_t negative.
+        tfields.sort([](UElement e1, UElement e2) -> int32_t {
+            // uprv_strcmp may return value other than {-1, 0, 1}.
             int res = uprv_strcmp(
                 (const char*)e1.pointer, (const char*)e2.pointer);
             return (res == 0) ? 0 : ((res > 0) ? 1 : -1);
```
markusicu, line 1583:
```suggestion
            return uprv_strcmp((const char*)e1.pointer, (const char*)e2.pointer);
```
I am pretty sure that strcmp() at worst returns the difference between unequal bytes, which means that it needs 9 bits, which easily fit into an int32_t. And usually an int is the same as an int32_t.

markusicu (commented):
Now that ICU 69 is off on the maintenance branch, we can proceed here targeting master/main.

Please change the PR title to "ICU-21526 Change return of UElementComparator to int32_t", and once you squash please use this for the commit message, too.

FrankYFTang:
got error on icu4c-release-tools build 
```
Error:  Failed to execute goal on project icu4c-apireport: Could not resolve dependencies for project org.unicode.icu:icu4c-apireport:jar:1.0-SNAPSHOT: Could not transfer artifact xalan:xalan:jar:2.7.2 from/to central (https://repo.maven.apache.org/maven2): Transfer failed for https://repo.maven.apache.org/maven2/xalan/xalan/2.7.2/xalan-2.7.2.jar 502 Gateway Error -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
Error: Process completed with exit code 1.
```
Not sure how to fix this

icu4c/source/common/locid.cpp
FrankYFTang: but that does not mean it will return 0, -1 and 1 which is still mandated by UElementComparator

markusicu: Yes, see my [ticket comment from March 5](https://unicode-org.atlassian.net/browse/ICU-21526?focusedCommentId=160438). Please take a look at the UElementComparator call sites, I bet that few or none rely on these three values.

FrankYFTang: I understand but icu4c/source/common/uelement.h still state

```
 * @return 0 if the two elements are equal, -1 if e1 is < e2, or +1 if e1 is > e2.
```

do you want me to change it to
```
 * @return 0 if the two elements are equal, a negative value  if e1 is < e2, or a positive value if e1 is > e2.
```
? If not changing there, then we cannot drop this change here, right?

markusicu: Please look at the call sites. If they don't actually need the return value to be one of these three values, or if they are easy to fix, then please change the "API" docs.

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm

PR #1630, opened 2021-03-08
ICU-20918 ucal_openTimeZones() to call ucal_openTimeZoneIDEnumeration()

yumaoka:
ucal_openTimeZones(UErrorCode*) used to call TimeZone::createEnumeration(), which does not return UErrorCode. TimeZone::createEnumeration() internally calls TZEnumeration::create(USystemTimeZoneType, const char*, const int32_t*, UErrorCode&), but it internally defines and eats UErrorCode. To set error status properly in ucal_openTimeZones(UErrorCode*), we should use TimeZone::createTimeZoneIDEnumeration(USystemTimeZoneType, const char*, const int32_t*, UErrorCode&), which is wrapped by ucal_openTimeZoneIDEnumeration(USystemTimeZoneType, const char*, const int32_t*, UErrorCode*), with UCAL_ZONE_TYPE_ANY, null region/offset.

Same thing applied to ucal_openCountryTimeZones(const char*, UErrorCode*).

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20918

### DIFF
--- a/icu4c/source/i18n/ucal.cpp
+++ b/icu4c/source/i18n/ucal.cpp
@@ -33,8 +33,8 @@ U_NAMESPACE_USE
 
 static TimeZone*
 _createTimeZone(const UChar* zoneID, int32_t len, UErrorCode* ec) {
-    TimeZone* zone = NULL;
-    if (ec != NULL && U_SUCCESS(*ec)) {
+    TimeZone* zone = nullptr;
+    if (ec != nullptr && U_SUCCESS(*ec)) {
         // Note that if zoneID is invalid, we get back GMT. This odd
         // behavior is by design and goes back to the JDK. The only
         // failure we will see is a memory allocation failure.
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/ucal.cpp
```diff
@@ -58,12 +58,12 @@ ucal_openTimeZoneIDEnumeration(USystemTimeZoneType zoneType, const char* region,
 
 U_CAPI UEnumeration* U_EXPORT2
 ucal_openTimeZones(UErrorCode* ec) {
-    return uenum_openFromStringEnumeration(TimeZone::createEnumeration(), ec);
+    return ucal_openTimeZoneIDEnumeration(UCAL_ZONE_TYPE_ANY, NULL, NULL, ec);
```
markusicu, line 61: Would you mind amending your commit to using nullptr instead of NULL? (Both functions you are changing.)

yumaoka: ok - will do.

markusicu (approved):
lgtm tnx

PR #1632, opened 2021-03-09
ICU-21330 Use =0 and =1 plural forms in compact notation

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21330

### DIFF
--- a/icu4c/source/i18n/number_compact.cpp
+++ b/icu4c/source/i18n/number_compact.cpp
@@ -55,7 +55,7 @@ int32_t countZeros(const UChar *patternString, int32_t patternLength) {
 } // namespace
 
 // NOTE: patterns and multipliers both get zero-initialized.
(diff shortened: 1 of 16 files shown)


echeran (approved):
LGTM.

Are there other instances of plural rules that use the `=0` and `=1` cases outside of compact numbers? In theory, it would be interesting to have a test that shows the difference between `ONE` and `=1`, or `ZERO` and `=0` for a non-compact number. But rules may not exist with such a pair of plural categories present. And if the original bug came up recently because this is the first known instance of a rule using `=1` / `=0`, then the question is moot.

pedberg-icu (approved):
Looks good, thanks

sffc:
The model of "give me the plural form of this number" doesn't accommodate itself well to =0, =1, so other components needing this logic will need to have a bit of custom code as I did here with compact notation unfortunately.

echeran (approved):
Rubber stamping after the previous approval and squash + force push for single commit.

PR #1634, opened 2021-03-10
ICU-21480 brs 69rc, add tests to verify that rbnf spellout is the same for no,nb

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21480

Per suggestion from George Rhoten, add tests to verify that RBNF spellout behaves the same way for "no" and "nb" after the parent change in earlier CLDR data integrations. Apparently aliasing does not work normally for RBNF, but parent inheritance does seem to, in this case at  least.

### DIFF
--- a/icu4c/source/test/intltest/itrbnf.cpp
+++ b/icu4c/source/test/intltest/itrbnf.cpp
@@ -77,6 +77,7 @@ void IntlTestRBNF::runIndexedTest(int32_t index, UBool exec, const char* &name, 
         TESTCASE(25, TestCompactDecimalFormatStyle);
         TESTCASE(26, TestParseFailure);
         TESTCASE(27, TestMinMaxIntegerDigitsIgnored);
+        TESTCASE(28, TestNorwegianSpellout);
 #else
         TESTCASE(0, TestRBNFDisabled);
 #endif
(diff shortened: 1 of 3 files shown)


grhoten (approved):
The important check is 1, which is different between nn and nb right now.

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -77,6 +77,7 @@ void IntlTestRBNF::runIndexedTest(int32_t index, UBool exec, const char* &name,
         TESTCASE(25, TestCompactDecimalFormatStyle);
         TESTCASE(26, TestParseFailure);
         TESTCASE(27, TestMinMaxIntegerDigitsIgnored);
+        TESTCASE(28, TestNorwegianSpellout);
```
markusicu, line 80: At some point it would be nice to switch this to TESTCASE_AUTO...

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -1604,6 +1605,38 @@ IntlTestRBNF::TestThaiSpellout()
...
+            { "1,001", "tusen og \\u00E9n" },
+            { "1,100", "tusen hundre" },
+            { "6.789", "seks komma sju \\u00E5tte ni" },
+            { "-5.678", "minus fem komma seks sju \\u00E5tte" },
+            { NULL, NULL }
+        };
+        doTest(noFormatter, testDataDefault, TRUE);
+        doTest(nbFormatter, testDataDefault, TRUE);
+    }
+    delete nbFormatter;
```
markusicu, line 1636: ok, but LocalPointer would be nicer

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -820,6 +820,35 @@ public void TestFractionalRuleSet() {
...
+    {
+        Locale noLocale = new Locale("no", "", "");
+        Locale nbLocale = new Locale("nb", "", "");
+        RuleBasedNumberFormat noFormatter = new RuleBasedNumberFormat(noLocale,
+                RuleBasedNumberFormat.SPELLOUT);
+        RuleBasedNumberFormat nbFormatter = new RuleBasedNumberFormat(nbLocale,
+                RuleBasedNumberFormat.SPELLOUT);
+
+        String[][] testDataDefault = {
+                { "1", "\u00E9n" },
```
markusicu, line 834: in java, why escapes rather than real characters?

in c++, i see that the shared test function takes `const char *` which may not be unicode, but in java we have no such constraint.

pedberg-icu: I was following the existing precedent. I imagine the reason is to make it easier to copy data from the C test to Java, or vice versa (since these same tests exist in both.

icu4c/source/test/intltest/itrbnf.cpp
pedberg-icu: Yes, but I was following existing precedent in the file to make it easier to compare the various tests for different locales. Probably what we need is a ticket to make a global change (or at least changes for entire files at one time rather than piecemeal).

icu4c/source/test/intltest/itrbnf.cpp
pedberg-icu: Yes, but not worth doing for small test addition like this. Better to do this for test across the board.

PR #1635, opened 2021-03-10
ICU-21358 Use sign position to format approximate numbers

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21358

### DIFF
--- a/icu4c/source/i18n/numparse_affixes.cpp
+++ b/icu4c/source/i18n/numparse_affixes.cpp
@@ -294,18 +294,20 @@ void AffixMatcherWarehouse::createAffixMatchers(const AffixPatternProvider& patt
         }
 
         // Generate Prefix
+        // TODO: Handle approximately sign?
         bool hasPrefix = false;
         PatternStringUtils::patternInfoToStringBuilder(
-                patternInfo, true, type, StandardPlural::OTHER, false, sb);
+                patternInfo, true, type, false, StandardPlural::OTHER, false, sb);
         fAffixPatternMatchers[numAffixPatternMatchers] = AffixPatternMatcher::fromAffixPattern(
                 sb, *fTokenWarehouse, parseFlags, &hasPrefix, status);
         AffixPatternMatcher* prefix = hasPrefix ? &fAffixPatternMatchers[numAffixPatternMatchers++]
                                                 : nullptr;
 
         // Generate Suffix
+        // TODO: Handle approximately sign?
         bool hasSuffix = false;
         PatternStringUtils::patternInfoToStringBuilder(
-                patternInfo, false, type, StandardPlural::OTHER, false, sb);
+                patternInfo, false, type, false, StandardPlural::OTHER, false, sb);
         fAffixPatternMatchers[numAffixPatternMatchers] = AffixPatternMatcher::fromAffixPattern(
                 sb, *fTokenWarehouse, parseFlags, &hasSuffix, status);
         AffixPatternMatcher* suffix = hasSuffix ? &fAffixPatternMatchers[numAffixPatternMatchers++]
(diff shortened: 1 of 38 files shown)


sffc:
@macchiati I assigned this one to you for review so you can see how the new approximatelySign affects output.  You can just look at the test files.  If it looks good, I will port to J and add another reviewer.

icu4c/source/i18n/numrange_impl.cpp
```diff
@@ -128,7 +129,16 @@ NumberRangeFormatterImpl::NumberRangeFormatterImpl(const RangeMacroProps& macros
     getNumberRangeData(macros.locale.getName(), nsName, data, status);
     if (U_FAILURE(status)) { return; }
     fRangeFormatter = data.rangePattern;
-    fApproximatelyModifier = {data.approximatelyPattern, kUndefinedField, false};
+
```
sffc, line 132: I should stop loading `data.approximatelyPattern` to save some time in the resource bundle.

icu4c/source/i18n/numparse_affixes.cpp
```diff
@@ -294,18 +294,20 @@ void AffixMatcherWarehouse::createAffixMatchers(const AffixPatternProvider& patt
         }
 
         // Generate Prefix
+        // TODO: Handle approximately sign?
         bool hasPrefix = false;
         PatternStringUtils::patternInfoToStringBuilder(
-                patternInfo, true, type, StandardPlural::OTHER, false, sb);
+                patternInfo, true, type, false, StandardPlural::OTHER, false, sb);
```
macchiati, line 300: In general I prefer enums (with semantics) over naked booleans. Not a blocker for this PR, since that is somewhat orthongonal to this PR since the original patternInfoToStringBuilder already uses those.

macchiati (approved):

icu4c/source/i18n/numparse_affixes.cpp
sffc: Agreed, but this is a deeply internal function, which, as you say, already uses booleans.  I would not use booleans if this were near the public API surface.

echeran (commented):
Since it appeared that the implementation needed to do a reset of the exponent value in order to do re-formatting (ex: when formatting settings change), should there also be a test that includes numbers with exponents? If there were, I may have missed them.

Otherwise, looks good.

icu4c/source/test/intltest/numbertest_affixutils.cpp
```diff
@@ -25,7 +25,7 @@ class DefaultSymbolProvider : public SymbolProvider {
             case TYPE_PLUS_SIGN:
                 return fSymbols.getConstSymbol(DecimalFormatSymbols::ENumberFormatSymbol::kPlusSignSymbol);
             case TYPE_APPROXIMATELY_SIGN:
-                return fSymbols.getConstSymbol(DecimalFormatSymbols::ENumberFormatSymbol::kApproximatelySignSymbol);
+                return u"≃";
```
echeran, line 28: I always thought that the approximately equals sign is ≈. And Latex and the [interwebs/Wikipedia](https://en.wikipedia.org/wiki/Approximation#Unicode) concurs.

sffc: Acknowledged; the intent of this test is to return some "interesting" symbol to make sure that the symbol gets passed through the stack.  The actual locale data is determined by translators.

sffc:
All tests are passing!  Thanks @echeran for your prompt review.  Unfortunately, since the change touches DecimalFormatSymbols public header files, ICU-TC decided to hold this back until ICU 70.

sffc:
Friendly ping for a re-review.

echeran (approved):
LGTM

richgillam (approved):
Haven't looked at this myself, but Shane needs a quick ship-it and others have already approved...

PR #1648, opened 2021-03-14
ICU-21480 brs 69rc, adjust nonstable API macros

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21480

Adjusted nonstable API guard conditionals.
- Included more items in #ifndef U_HIDE_INTERNAL_API
- Broke some #ifndef U_HIDE_DRAFT_API into multiple sections for (likely) different version tags; removed redundant #ifndef U_HIDE_DRAFT_API
- Gave explicit numeric values to UNumberUnitWidth enum values to prevent problems when new enum values are added or when compiling with U_HIDE_DRAFT_API defined.
- Removed no-longer-applicable #ifndef U_HIDE_DRAFT_API around inline impl of a @stable method

### DIFF
--- a/icu4c/source/i18n/unicode/basictz.h
+++ b/icu4c/source/i18n/unicode/basictz.h
@@ -175,7 +175,6 @@ public:
         kFormer = 0x04, /* UCAL_TZ_LOCAL_FORMER */
         kLatter = 0x0C  /* UCAL_TZ_LOCAL_LATTER */
     };
-#endif  /* U_HIDE_INTERNAL_API */
 
     /**
      * Get time zone offsets from local wall time.
@@ -183,6 +182,7 @@ public:
      */
     void getOffsetFromLocal(UDate date, int32_t nonExistingTimeOpt, int32_t duplicatedTimeOpt,
         int32_t& rawOffset, int32_t& dstOffset, UErrorCode& status) const;
+#endif  /* U_HIDE_INTERNAL_API */
 
 protected:
 
--- a/icu4c/source/i18n/unicode/numberformatter.h
+++ b/icu4c/source/i18n/unicode/numberformatter.h
@@ -346,15 +346,15 @@ class U_I18N_API Notation : public UMemory {
 
     union NotationUnion {
         // For NTN_SCIENTIFIC
-        /** @internal */
+        /** @internal (private) */
         struct ScientificSettings {
-            /** @internal */
+            /** @internal (private) */
             int8_t fEngineeringInterval;
-            /** @internal */
+            /** @internal (private) */
             bool fRequireMinInt;
-            /** @internal */
+            /** @internal (private) */
             impl::digits_t fMinExponentDigits;
-            /** @internal */
+            /** @internal (private) */
             UNumberSignDisplay fExponentSignDisplay;
         } scientific;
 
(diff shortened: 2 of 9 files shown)


icu4c/source/i18n/unicode/basictz.h
```diff
@@ -175,14 +175,14 @@ class U_I18N_API BasicTimeZone: public TimeZone {
...
     };
-#endif  /* U_HIDE_INTERNAL_API */
 
     /**
      * Get time zone offsets from local wall time.
      * @internal
      */
     void getOffsetFromLocal(UDate date, int32_t nonExistingTimeOpt, int32_t duplicatedTimeOpt,
         int32_t& rawOffset, int32_t& dstOffset, UErrorCode& status) const;
+#endif  /* U_HIDE_INTERNAL_API */
```
pedberg-icu, line 186: Move doen to include the method above

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -2217,7 +2217,10 @@ class U_I18N_API NumberFormatterSettings {
      * @draft ICU 68
      */
     Derived usage(StringPiece usage) &&;
+#endif // U_HIDE_DRAFT_API
 
+#ifndef U_HIDE_DRAFT_API
+#ifndef U_HIDE_INTERNAL_API
     /**
```
pedberg-icu, line 2224: The methods following this were @internal ICU 69 tech preview. In addition to adding the #ifndef U_HIDE_INTERNAL_API, I split them into a separate ifndef U_HIDE_DRAFT_API block since they will likely end up with different dtaf versions than the regular @draft ICU 68 methods before them.

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -2777,7 +2781,9 @@ class U_I18N_API FormattedNumber : public UMemory, public FormattedValue {
 };
 
 #ifndef U_HIDE_DRAFT_API
-// Note: This is draft ICU 65
+/**
+ * @draft ICU 65
+ */
 template<typename StringClass>
```
pedberg-icu, line 2787: Draft status was not marked properly. There were some other places like this that I also fixed.

icu4c/source/i18n/unicode/unumberformatter.h
```diff
@@ -235,14 +235,16 @@ typedef enum UNumberUnitWidth {
      *
      * @stable ICU 60
      */
-            UNUM_UNIT_WIDTH_HIDDEN,
+            UNUM_UNIT_WIDTH_HIDDEN = 6,
```
pedberg-icu, line 239: Needed explicit numeric values because this is preceded by enum values conditionalized on #ifndef U_HIDE_DRAFT_API.

icu4c/source/i18n/unicode/unumberrangeformatter.h
```diff
@@ -354,7 +354,6 @@ unumrf_resultGetIdentityResult(
     UErrorCode* ec);
 
 
-#ifndef U_HIDE_DRAFT_API
```
pedberg-icu, line 357: This conditional block was redundant, it was enclosing two @draft ICU 68 functions that were part of a larger block of @draft ICU 68 functions already enclosed in #ifndef U_HIDE_DRAFT_API

icu4c/source/i18n/unicode/numberformatter.h
sffc: Actually, these are the inline impls of methods that are now `@stable`.  They aren't docstrings.  We should just remove the conditional.

sffc (changes requested):
Mostly looks good, but one comment that applies to several of the changes

icu4c/source/i18n/unicode/numberformatter.h
pedberg-icu: Thanks, good catch. But this one was the only one that was now stable. I removed the conditionals for this one and changed the other two places so the comment clarifies that it is the inline impl of a @draft ICU 68 method.

pedberg-icu:
> Mostly looks good, but one comment that applies to several of the changes
Thanks. I think I addressed that.

sffc (approved):

PR #1655, opened 2021-03-16
ICU-21520 Housekeeping: Fixes typos in name of test data file; removes a regex that

gnrunge:
evaluates to the empty string and the result files are missing the
perf test name as a result. E.g. '_Mar_16_155820.html' instead of
'unisetperf_Mar_16_155820.html'.
Not obvious why the regex was even created since it doesn't seem
necessary because the name of the caller doesn't require additional
parsing. The particular code is from 2002, maybe changes in Perl
account for the issue(?)

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21520


markusicu (approved):
rslgtm

PR #1656, opened 2021-03-17
ICU-21537 Fix invalid free by long locale name

FrankYFTang:
Free baseName and set to null when we reallocate fullName
so we will call initBaseName properly later again.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21537


FrankYFTang:
ok, I think I find the problem. it is not a one off issue. The problem is there are a baseName and fullName in the locale, and they could be the same if there are no extension. When there are extension, it need to be two different string. The logic to free the baseName was incorrect and assume if it is different than the fullName then it is an extra allocated member but not the one preallocate in the object (fullNameBuffer). but if the first extension is exceeding the fullNameBuffer size, then we allocate fullName and then leave the baseName still pointing to fullNameBuffer. For that case, we should not free the baseName

markusicu (approved):
lgtm tnx!

PR #1662, opened 2021-03-22
ICU-21550 Add zzzz to subdivision if len==2

FrankYFTang:
Use subdivisionAlias when len is 2 by appending "zzzz" per changes to
UTS35 in https://github.com/unicode-org/cldr/pull/1132

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21550

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -1519,9 +1519,12 @@ AliasReplacer::replaceSubdivision(
         // Found replacement data for this subdivision.
         size_t len = (firstSpace != nullptr) ?
             (firstSpace - replacement) : uprv_strlen(replacement);
-        // Ignore len == 2, see CLDR-14312
-        if (3 <= len && len <= 8) {
+        if (2 <= len && len <= 8) {
             output.append(replacement, (int32_t)len, status);
+            if (2 == len) {
+                // Add 'zzzz' based on changes to UTS #35 for CLDR-14312.
+                output.append("zzzz", 4, status);
+            }
         }
         return true;
     }
--- a/icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
@@ -1402,11 +1402,12 @@ public final class ULocale implements Serializable, Comparable<ULocale> {
                     throw new IllegalArgumentException(
                         "Incorrect key [" + aliasFrom + "] in alias:territory.");
                 }
-                if (aliasTo.length() < 3 || aliasTo.length() > 8) {
-                    // Ignore replacement < 3 for now. see CLDR-14312
-                    // throw new IllegalArgumentException(
-                    //    "Incorrect value [" + aliasTo + "] in alias:subdivision.");
-                    continue;
+                if (aliasTo.length() == 2) {
+                    // Add 'zzzz' based on changes to UTS #35 for CLDR-14312.
+                    aliasTo += "zzzz";
+                } else if (aliasTo.length() < 2 || aliasTo.length() > 8) {
+                    throw new IllegalArgumentException(
+                        "Incorrect value [" + aliasTo + "] in alias:territory.");
                 }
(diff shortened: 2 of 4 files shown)


icu4c/source/common/locid.cpp
```diff
@@ -1519,9 +1519,13 @@ AliasReplacer::replaceSubdivision(
         // Found replacement data for this subdivision.
         size_t len = (firstSpace != nullptr) ?
             (firstSpace - replacement) : uprv_strlen(replacement);
-        // Ignore len == 2, see CLDR-14312
         if (3 <= len && len <= 8) {
             output.append(replacement, (int32_t)len, status);
+        } else if (2 == len) {
+          // Add 'zzzz' if len == 2 based on changes to UTS35 in
+          // https://github.com/unicode-org/cldr/pull/1132
+          output.append(replacement, (int32_t)len, status);
```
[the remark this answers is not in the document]
FrankYFTang: This code is inside AliasReplacer::replaceSubdivision()
and the only place call it is in AliasReplacer::replace() with the following

```
        Locale temp(locale);
...
                    CharString replacement;
                    if (uprv_strlen(key) == 2) {
                        if (replaceSubdivision(value.toStringPiece(), replacement, status)) {
                            changed++;
                            temp.setKeywordValue(key, replacement.data(), status);
                        }
                    } else {
```
so the maybe-uppercase "AXzzzz" may be passed to Locale::setKeywordValue() right after and setKeyword will take care the normalization as usual.

icu4c/source/common/locid.cpp
```diff
@@ -1519,9 +1519,13 @@ AliasReplacer::replaceSubdivision(
         // Found replacement data for this subdivision.
         size_t len = (firstSpace != nullptr) ?
             (firstSpace - replacement) : uprv_strlen(replacement);
-        // Ignore len == 2, see CLDR-14312
         if (3 <= len && len <= 8) {
             output.append(replacement, (int32_t)len, status);
+        } else if (2 == len) {
+          // Add 'zzzz' if len == 2 based on changes to UTS35 in
```
markusicu, line 1525: Optional: You could combine this with the 3..8 case:
```
        if (2 <= len && len <= 8) {
            output.append(replacement, (int32_t)len, status);
            if (len == 2) {
                // Add 'zzzz' based on changes to UTS #35 for CLDR-14312.
                output.append("zzzz", 4, status);
            }
```

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1403,10 +1403,15 @@ private static synchronized void loadAliasData() {
                         "Incorrect key [" + aliasFrom + "] in alias:territory.");
                 }
                 if (aliasTo.length() < 3 || aliasTo.length() > 8) {
-                    // Ignore replacement < 3 for now. see CLDR-14312
-                    // throw new IllegalArgumentException(
-                    //    "Incorrect value [" + aliasTo + "] in alias:subdivision.");
-                    continue;
+                    if (aliasTo.length() == 2) {
```
markusicu, line 1406: It looks like in C++ you keep the alias data as is and add the zzzz while canonicalizing -- but here in Java you modify the alias data.
Why the difference?
Is it possible that subdivision aliases may be used in some place where the added zzzz would not be desirable? In that case, the C++ approach seems better.

markusicu, line 1406: ... and if you do want to change the alias data, I would pull the length==2 case out of the error handling for clarity.

FrankYFTang: In C++, the alias data is directly referencing to the memory in the memory mapped resources, so we cannot grow the memory for 4 other char. In Java, it is separately allocated.

FrankYFTang:
PTAL

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm tnx

PR #1666, opened 2021-03-24
ICU-21554 - fix path seperator in toolutil for mingw

autoantwort:
Issue filed: https://unicode-org.atlassian.net/browse/ICU-21554


jefgen:
Thank you very much for filing a JIRA ticket, and for creating this PR! 

Regarding this:
> This PR is not enough to fix [ICU-21554](https://unicode-org.atlassian.net/browse/ICU-21554), but it a step to reach this goal.

FWIW, I think the changes in this PR make sense, in that we'd likely want to use the same path separator for mingw as cygwin.
However, I'm not very familiar with building ICU under mingw though. But I think this change could likely be merged even if it doesn't fully address the issue.

longnguyen2004:
https://github.com/unicode-org/icu/blob/fcd8c34f5eb1c8bf42e62381d93b633747f7c52e/icu4c/source/tools/pkgdata/pkgdata.cpp#L536-L540

Should we treat mingw to be the same as cygwin here? Maybe we should execute the command directly, instead of going through `bash`. This would be another way to solve the problem.

autoantwort:
> Should we treat mingw to be the same as cygwin here? Maybe we should execute the command directly, instead of going through `bash`. This would be another way to solve the problem.

I tried it, but then I get:
```
      0 [main] ar (17604) C:\git-sdk-64\usr\bin\ar.exe: *** fatal error - cygheap base mismatch detected - 0x180347408/0x180317408.
This problem is probably due to using incompatible versions of the cygwin DLL.
Search for cygwin1.dll using the Windows Start->Find/Search facility
and delete all but the most recent version.  The most recent version *should*
reside in x:\cygwin\bin, where 'x' is the drive on which you have
installed the cygwin distribution.  Rebooting is also suggested if you
are unable to find another cygwin DLL.
make[1]: *** [Makefile:131: libsicudt.a] Error 127
make: *** [Makefile:153: all-recursive] Error 2

```

longnguyen2004:
@autoantwort Maybe don't use git bash, but an actual MSYS2?

autoantwort:
Afaik is is executed in MSYS2, but I also see that it executes ar.exe from C:\git-sdk-64\usr\bin\

autoantwort:
> @autoantwort Maybe don't use git bash, but an actual MSYS2?

Yeah I was at a wrong branch internally. Works now and fixes the issue. Now only the lib names are wrong:
```
ar: creating libsicutestd.a
C:/Qt/Tools/mingw810_64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -licutud
C:/Qt/Tools/mingw810_64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -licuind
C:/Qt/Tools/mingw810_64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -licuucd
C:/Qt/Tools/mingw810_64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -licudtd
```
There is a `s` missing in the lib names

longnguyen2004:
Alright then, I think we should proceed with my approach. I just need feedback from others (specifically the people who handle the mingw build system).

autoantwort:
Ok, I have created #1682

PR #1669, opened 2021-03-26
ICU-21559 more guards for self-assignment

markusicu:
https://unicode-org.atlassian.net/browse/ICU-21559

Implementation:
- I looked at all implementations of assignment operators in ICU4C.
- (Not move operators, those have different rules.)
- Many already guard against self-assignment.
- Many don't, but it's harmless (e.g., no pointers/memory management).
- StringSearch was testing for equality not identity.
- I added guards where I could not convince myself that it was not necessary.
- I added a guard in class Edits just to avoid a memcpy().

Tests:
- I added a few tests with explicit syntax to try to avoid the compiler optimizing the self-assignment away.
- Only the DateFormatSymbols test actually failed with "AddressSanitizer: heap-use-after-free" before the fix.
- Many of these operators are internal, protected (Transliterator), or private (CompoundTransliterator).

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21559


icu4c/source/test/intltest/numbertest_api.cpp
```diff
@@ -1209,6 +1209,7 @@ void NumberFormatterApiTest::unitArbitraryMeasureUnits() {
               .unit(MeasureUnit::forIdentifier("pow4-mile", status))
               .unitWidth(UNUM_UNIT_WIDTH_FULL_NAME)
               .locale("en-ZA");
+    lnf.operator=(lnf);  // self-assignment should be a no-op
```
sffc, line 1212: (on phone)

Seems like an odd place for this test. A better place would be in the move operator test further down.

sffc (commented):
OK, but:

1. Why now? Most of these APIs are oldish and stable.
2. This PR is a bandaid that doesn't fix this problem when it will inevitably happen again. A better fix would be one centered around a lint of some sort.

markusicu:
> 1. Why now?

Because someone found two of these and Victor pinged me about them. See the GoogleIssues in the ticket.
Seemed like a good "fix a bug per week" project.

> Most of these APIs are oldish and stable.

It's still a bug.

> 2. This PR is a bandaid that doesn't fix this problem when it will inevitably happen again. A better fix would be one centered around a lint of some sort.

This is true, but since we are not set up for such linting, and I am not sure if a linter can even catch these reliably, the next best thing is to inspect the implementations and fix the ones that need fixing.

icu4c/source/test/intltest/numbertest_api.cpp
markusicu: I just looked for an object with interesting settings that look like they could cause problems in the lnfCopyHelper() and added a copy assignment. If you have a better place, or a more specific test case, please provide a specific location or sample code.

jefgen (approved):
fwiw, lgtm.
Thanks for going through all of the `operator=` implementations and checking...

PR #1670, opened 2021-03-30
ICU-21562 Add include to fix std:abs build problem

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21562

### DIFF
--- a/icu4c/source/common/charstr.cpp
+++ b/icu4c/source/common/charstr.cpp
@@ -14,6 +14,8 @@
 *   created by: Markus W. Scherer
 */
 
+#include <cstdlib>
+
 #include "unicode/utypes.h"
 #include "unicode/putil.h"
 #include "charstr.h"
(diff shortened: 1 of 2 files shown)


markusicu (approved):
lgtm

hugovdm (commented):
Thanks!

I'll experiment with this when I poke at "include-what-you-use" (https://unicode-org.atlassian.net/browse/ICU-8530) - e.g. removing the include to see what happens.

FrankYFTang:
The problem is "libc++ unexpectedly defines std::abs(float) in cmath and std::abs(int) in cstdlib."

see also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89855
My understanding is different platforms ( or compiler + library) somehow define them in different header file ...

FrankYFTang:
Therefore, you need to try it on all platform OR you check the C++ documentation
https://en.cppreference.com/w/cpp/numeric/math/abs
said


Defined in header <cstdlib> |   |  
-- | -- | --
Defined in header <cmath>

PR #1671, opened 2021-03-31
ICU-21563 Fix delete-incomplete on windows

FrankYFTang:
Moving template after declaration of the MeasureUnitImpl

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21563

### DIFF
--- a/icu4c/source/i18n/measunit_impl.h
+++ b/icu4c/source/i18n/measunit_impl.h
@@ -20,22 +20,6 @@ class LongNameHandler;
 }
 } // namespace number
 
-// Export an explicit template instantiation of the LocalPointer that is used as a
-// data member of MeasureUnitImpl.
-// (When building DLLs for Windows this is required.)
-#if U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN
-#if defined(_MSC_VER)
-// Ignore warning 4661 as LocalPointerBase does not use operator== or operator!=
-#pragma warning(push)
-#pragma warning(disable : 4661)
-#endif
-template class U_I18N_API LocalPointerBase<MeasureUnitImpl>;
-template class U_I18N_API LocalPointer<MeasureUnitImpl>;
-#if defined(_MSC_VER)
-#pragma warning(pop)
-#endif
-#endif
-
 static const char16_t kDefaultCurrency[] = u"XXX";
 static const char kDefaultCurrency8[] = "XXX";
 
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/measunit_impl.h
```diff
@@ -330,6 +302,34 @@ struct U_I18N_API MeasureUnitImplWithIndex : public UMemory {
...
+// MaybeStackVector. This is required when building DLLs for Windows. (See
+// datefmt.h, collationiterator.h, erarules.h and others for similar examples.)
+#if U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN
+template class U_I18N_API MaybeStackArray<SingleUnitImpl *, 8>;
+template class U_I18N_API MemoryPool<SingleUnitImpl, 8>;
+template class U_I18N_API MaybeStackVector<SingleUnitImpl, 8>;
+template class U_I18N_API MaybeStackArray<MeasureUnitImplWithIndex *, 8>;
+template class U_I18N_API MemoryPool<MeasureUnitImplWithIndex, 8>;
+template class U_I18N_API MaybeStackVector<MeasureUnitImplWithIndex, 8>;
+#endif
```
jefgen, line 315: Can these be combined into a single ifdef?

FrankYFTang: somehow the windows build is not happy about the PR. 
so on chromium, the template have to be define AFTER the whole class to build
but on here, the template have to be define BEFORE it got used to build
MaybeStackVector<MeasureUnitImplWithIndex, 8> is USED before it got defined so I don't know how to fix that part yet.

FrankYFTang: Let's worry about about the combining the #ifdef after we can figure out how to make it build on BOTH places first. Because until then, I may need to move them around and it is much easier to keep them under different #ifdef to move around until where it will build first.

FrankYFTang: if that is just too hard, I will just patch v8 locally and abandon this PR instead....

FrankYFTang: DONE

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm

icu4c/source/i18n/measunit_impl.h
jefgen: I see you had to move the `MeasureUnitImplWithIndex` as well. Thanks Frank!

hugovdm (commented):
Thanks! Does this show up in CI yet?

jefgen:
> Thanks! Does this show up in CI yet?

Do you mean that the error would show up in one of the CI builds?
If so, then I think the answer is no, as we don't yet have a clang-cl build bot setup for Windows.

PR #1683, opened 2021-04-02
ICU-21539 Fix getUnicodeKeywords to skip non keys

FrankYFTang:
Also add test in Java

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21539

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -2527,16 +2527,16 @@ public:
 
     virtual const char* next(int32_t* resultLength, UErrorCode& status) {
         const char* legacy_key = KeywordEnumeration::next(nullptr, status);
-        if (U_SUCCESS(status) && legacy_key != nullptr) {
+        while (U_SUCCESS(status) && legacy_key != nullptr) {
             const char* key = uloc_toUnicodeLocaleKey(legacy_key);
-            if (key == nullptr) {
-                status = U_ILLEGAL_ARGUMENT_ERROR;
-            } else {
+            if (key != nullptr) {
                 if (resultLength != nullptr) {
                     *resultLength = static_cast<int32_t>(uprv_strlen(key));
                 }
                 return key;
             }
+            // Not a Unicode keyword, could be a t, x or other, continue to look at the next one.
+            legacy_key = KeywordEnumeration::next(nullptr, status);
         }
         if (resultLength != nullptr) *resultLength = 0;
         return nullptr;
--- a/icu4c/source/test/intltest/loctest.cpp
+++ b/icu4c/source/test/intltest/loctest.cpp
@@ -239,8 +239,10 @@ void LocaleTest::runIndexedTest( int32_t index, UBool exec, const char* &name, c
     TESTCASE_AUTO(TestKeywordVariantParsing);
     TESTCASE_AUTO(TestCreateKeywordSet);
     TESTCASE_AUTO(TestCreateKeywordSetEmpty);
+    TESTCASE_AUTO(TestCreateKeywordSetWithPrivateUse);
     TESTCASE_AUTO(TestCreateUnicodeKeywordSet);
     TESTCASE_AUTO(TestCreateUnicodeKeywordSetEmpty);
+    TESTCASE_AUTO(TestCreateUnicodeKeywordSetWithPrivateUse);
     TESTCASE_AUTO(TestGetKeywordValueStdString);
     TESTCASE_AUTO(TestGetUnicodeKeywordValueStdString);
     TESTCASE_AUTO(TestSetKeywordValue);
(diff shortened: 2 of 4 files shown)


icu4c/source/test/intltest/loctest.cpp
```diff
@@ -4101,6 +4101,34 @@ LocaleTest::TestCreateUnicodeKeywordSet(void) {
                result.find("ca") != result.end());
     assertTrue("set::find(\"co\")",
                result.find("co") != result.end());
+
+    static const char tag[] = "en-US-u-ca-gregory-x-foo";
+    Locale l2 = Locale::forLanguageTag(tag, status);
+    result.clear();
+    l2.getKeywords<std::string>(
```
roubert, line 4108: The way these test cases are structured, it would be more logical if this additional testing of `getKeywords()` was moved into the `TestCreateKeywordSet` test case above, so that it's kept together with the already existing testing of `getKeywords()` there.

Even more logical would be to add new separate test cases, instead of adding new code to the existing ones, so that each one of these test cases continues to simply test one single thing each (that'll make the code simpler, no calls to `clear()`, no variables named `l2`, etc., and a test failure would then point directly to the problem).

Then you'd end up with three test cases for old style keywords:

`TestCreateKeywordSet`
`TestCreateKeywordSetEmpty`
`TestCreateKeywordSetWithPrivateUse`

And three test cases for Unicode keywords:

`TestCreateUnicodeKeywordSet`
`TestCreateUnicodeKeywordSetEmpty`
`TestCreateUnicodeKeywordSetWithPrivateUse`

icu4c/source/test/intltest/loctest.cpp
```diff
@@ -4101,6 +4101,34 @@ LocaleTest::TestCreateUnicodeKeywordSet(void) {
...
+            std::insert_iterator<decltype(result)>(result, result.begin()),
+            status);
+    status.errIfFailureAndReset("getKeywords \"%s\"", l2.getName());
+    assertTrue("getKeywords set::find(\"calendar\")",
+               result.find("calendar") != result.end());
+    assertTrue("getKeywords set::find(\"ca\")",
+               result.find("ca") == result.end());
+    assertTrue("getKeywords set::find(\"x\")",
+               result.find("x") != result.end());
+    assertTrue("getKeywords set::find(\"co\")",
```
roubert, line 4118: The test shouldn't be looking for `"co"` here, there's no such thing in the input language tag so it couldn't possibly show up, even if there were bugs. What the test should be looking for here is `"foo"`, which could possibly show up if some bug got values and keys mixed up.

icu4c/source/test/intltest/loctest.cpp
```diff
@@ -4101,6 +4101,34 @@ LocaleTest::TestCreateUnicodeKeywordSet(void) {
...
+    result.clear();
+    l2.getUnicodeKeywords<std::string>(
+            std::insert_iterator<decltype(result)>(result, result.begin()),
+            status);
+    status.errIfFailureAndReset("getUnicodeKeywords \"%s\"", l2.getName());
+    assertTrue("getUnicodeKeywords set::find(\"ca\")",
+               result.find("ca") != result.end());
+    assertTrue("getUnicodeKeywords set::find(\"x\")",
+               result.find("x") == result.end());
+    assertTrue("getUnicodeKeywords set::find(\"co\")",
```
roubert, line 4130: See previous comment, `"foo"` instead of `"co"`, but just as the existing test code for `getUnicodeKeywords()` skips this entirely because it has a test for `size()` instead, I think that's what you'd better do here too.

icu4c/source/common/locid.cpp
```diff
@@ -2527,16 +2527,16 @@ class UnicodeKeywordEnumeration : public KeywordEnumeration {
...
-            if (key == nullptr) {
-                status = U_ILLEGAL_ARGUMENT_ERROR;
-            } else {
+            if (key != nullptr) {
                 if (resultLength != nullptr) {
                     *resultLength = static_cast<int32_t>(uprv_strlen(key));
                 }
                 return key;
             }
+            // Not an unicode keyword, could be a t, x or other, continue to look at the next one.
```
roubert, line 2538: an unicode → a Unicode

roubert (changes requested):

icu4c/source/test/intltest/loctest.cpp
FrankYFTang: fixed

icu4c/source/test/intltest/loctest.cpp
FrankYFTang: fixed

icu4c/source/common/locid.cpp
FrankYFTang: fixed

FrankYFTang:
PTAL

roubert (approved):
LGTM

PR #1687, opened 2021-04-05
ICU-21459 properly guard BytesTrie.Result.getValue()

markusicu:
Result.matches() is an insufficient guard for Result.getValue().
I simplified the match loop so that there is only one place for checking hasValue() and recording a match.

It did not fail in C++ because the assertion is commented out there because it's in a public header file -- but the C++ code was bug-compatible with Java.

Other bug fixes: Neither implementation was thread-safe for copies/clones used in parallel threads.
- In Java, changed clone() to clone the fields (which are all stateful).
- In C++, made the worker function use stack-local copies of the trie iterator objects so that the shared ones remain const, and changed the data refcount to an atomic int.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21459

### DIFF
--- a/icu4c/source/common/filteredbrk.cpp
+++ b/icu4c/source/common/filteredbrk.cpp
@@ -20,6 +20,7 @@
 #include "ubrkimpl.h" // U_ICUDATA_BRKITR
 #include "uvector.h"
 #include "cmemory.h"
+#include "umutex.h"
 
 U_NAMESPACE_BEGIN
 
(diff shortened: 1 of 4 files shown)


aheninger (commented):
Your change looks good to me.

But, the UCharsTrie instances held in fData look suspect. UCharsTrie are non-const things, but fData (class SimpleFilteredSentenceBreakData) look like it's intended to be shared.
Look at the copy constructor for SimpleFilteredSentenceBreakIterator.

markusicu:
> But, the UCharsTrie instances held in fData look suspect. UCharsTrie are non-const things, but fData (class SimpleFilteredSentenceBreakData) look like it's intended to be shared.

You are right, it's because the fData tries own their data arrays. They must not be modified.
I changed the code to use shallow-copy iterators.
I also changed the fData refcount into an atomic int.

PTAL

icu4c/source/common/filteredbrk.cpp
```diff
@@ -261,59 +285,62 @@ SimpleFilteredSentenceBreakIterator::breakExceptionAt(int32_t n) {
...
     utext_setNativeIndex(fText.getAlias(), n); // from n..
-    fData->fBackwardsTrie->reset();
-    UChar32 uch;
 
     //if(debug2) u_printf(" n@ %d\n", n);
     // Assume a space is following the '.'  (so we handle the case:  "Mr. /Brown")
-    if((uch=utext_previous32(fText.getAlias()))==(UChar32)0x0020) {  // TODO: skip a class of chars here??
+    if(utext_previous32(fText.getAlias())==u' ') {  // TODO: skip a class of chars here??
       // TODO only do this the 1st time?
       //if(debug2) u_printf("skipping prev: |%C| \n", (UChar)uch);
```
aheninger, line 293: Removing uch has broken the debug printfs. I don't think it matters.

icu4j/main/classes/core/src/com/ibm/icu/impl/SimpleFilteredSentenceBreakIterator.java
```diff
@@ -72,28 +72,22 @@ private final boolean breakExceptionAt(int n) {
...
         int uch;
 
-
-
         // Assume a space is following the '.' (so we handle the case: "Mr. /Brown")
         if ((uch = text.previousCodePoint()) == ' ') { // TODO: skip a class of chars here??
             // TODO only do this the 1st time?
         } else {
             uch = text.nextCodePoint();
         }
```
aheninger, line 81: Up in the constructor, lines 46 and 47, do we need to clone the CharTries? Or maybe it could be done in the clone() function. Or, given the timing, defer the problem? It's a bug, but not a regression.

icu4c/source/common/filteredbrk.cpp
```diff
@@ -261,59 +285,62 @@ SimpleFilteredSentenceBreakIterator::breakExceptionAt(int32_t n) {
     int32_t bestValue = -1;
     // loops while 'n' points to an exception.
     utext_setNativeIndex(fText.getAlias(), n); // from n..
-    fData->fBackwardsTrie->reset();
-    UChar32 uch;
 
     //if(debug2) u_printf(" n@ %d\n", n);
     // Assume a space is following the '.'  (so we handle the case:  "Mr. /Brown")
-    if((uch=utext_previous32(fText.getAlias()))==(UChar32)0x0020) {  // TODO: skip a class of chars here??
+    if(utext_previous32(fText.getAlias())==u' ') {  // TODO: skip a class of chars here??
```
srl295, line 291: 👍

srl295 (dismissed):
lgtm

icu4c/source/common/filteredbrk.cpp
srl295: Agreed.

icu4j/main/classes/core/src/com/ibm/icu/impl/SimpleFilteredSentenceBreakIterator.java
srl295: looks like it should be cloned

markusicu: Multiple instances must be independent of each other.

Either clone() needs to clone all of its fields (which I have done now), or else the worker function needs to clone the tries (more memory allocations), or the worker functions needs to create new trie iterator objects for the raw data (also more memory allocations).

I left TODOs for equals() and hashCode() not having meaningful implementations.

srl295 (approved):
changes lgtm, thanks

PR #1688, opened 2021-04-06
ICU-21569 LSTM Part 1- data file and build tool

FrankYFTang:
NOT including the building of LSTM files by default
Allow developer to use filter to include LSTM

##### Checklist


FrankYFTang:
The lstm files are generated from 
https://github.com/unicode-org/lstm_word_segmentation/blob/master/convert_lstm_model.py

icu4c/source/data/BUILDRULES.py
```diff
@@ -457,6 +458,35 @@ def generate_translit(config, io, common_vars):
     ]
 
 
+def generate_brkitr_lstm(config, io, common_vars):
+    input_files = [InFile(filename) for filename in io.glob("brkitr/lstm/*.txt")]
+    dep_files = set(InFile(filename) for filename in io.glob("brkitr/lstm/*.txt"))
+    dep_files -= set(input_files)
+    dep_files = list(sorted(dep_files))
```
sffc, line 465: Suggestion (optional): I think `dep_files` ends up being empty.  You could delete it and pass an empty array for `dep_targets` down below in the RepeatedOrSingleExecutionRequest.

sffc (approved):

icu4c/source/data/BUILDRULES.py
FrankYFTang: DONE

FrankYFTang:
PTAL

sffc (approved):
LGTM from me, but I'll flag @markusicu and @srl295 as an FYI that this PR adds quite a bit of size to the ICU repository (not the release, just the repo, and probably icu-data.zip as well).

srl295:
Thought it's not in by default. What's the size delta ?

FrankYFTang:
> Thought it's not in by default. What's the size delta ?

What size are you talking about?

FrankYFTang:
The reason the size delta of .dat should be 0 (or maybe +30 bytes) is because all the lstm model by default won't be build. It will only be build if the filter file explicitly add them.

srl295:
Right, mis read. I mean what is the delta repo size?

FrankYFTang:
I think I already answer you the rough number above
> If you are talking about the size impact to the source data files in the source code- +5.7 M
> Just about ~30 lines changes of the pre-exist file and one .h/.cpp file which total about 900 lines.

srl295:
Sounds fine

PR #1689, opened 2021-04-06
ICU-21569 LSTM Part 2- include C++ implementation

FrankYFTang:
##### Checklist


aheninger (commented):
(Only a partial review)

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,751 @@
...
+    UVector32 &offsets, UVector32 &indices, UErrorCode &status) const
+{
+    if (offsets.ensureCapacity(endPos - startPos, status) &&
+            indices.ensureCapacity(endPos - startPos, status)) {
+        utext_setNativeIndex(text, startPos);
+        int32_t current;
+        UChar str[2] = {0, 0};
+        while (U_SUCCESS(status) &&
+               (current = (int32_t)utext_getNativeIndex(text)) < endPos) {
+            str[0] = (UChar) utext_next32(text);
```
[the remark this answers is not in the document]
FrankYFTang: Added comment and assert

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,751 @@
...
+const LanguageBreakEngine*
+CreateLSTMBreakEngine(UScriptCode script, UErrorCode& status)
+{
+    UnicodeString name = defaultLSTM(script, status);
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+    const LSTMData* data = new LSTMResourceData(name, status);
+    if (data == nullptr) {
+        return nullptr;
```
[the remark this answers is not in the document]
FrankYFTang: I rewrite the logic. so now if it is null I won't construct the engine here.

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,751 @@
...
+
+#define MAX_GRAPHEME_CLSTER_LENTH 10
+void GraphemeClusterVectorizer::vectorize(
+    UText *text, int32_t startPos, int32_t endPos,
+    UVector32 &offsets, UVector32 &indices, UErrorCode &status) const
+{
+    if (offsets.ensureCapacity(endPos - startPos, status) &&
+            indices.ensureCapacity(endPos - startPos, status)) {
+        LocalPointer<BreakIterator> graphemeIter(
+            BreakIterator::createCharacterInstance(Locale(), status));
```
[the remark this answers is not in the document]
FrankYFTang: But there are no locale of the break iterator because this LanguageBreakEngine will be shared across all BreakIterator and we have only character break iterator in the system anyway. passing in a locale to make this LanguageBreakEngine not shareable across different locale will be overdesign for a non-existing issue.

aheninger:
The builds are producing a message that looks problematic, although it doesn't register as an error. Happens both in this PR, and in main

```
LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k root.txt
LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k ru.txt
./brkitr/root.txt:26: The dependency file ./out/build/icudt69l/brkitr/Thai_graphclust_model4_heavy.res does not exist. Please make sure it exists.

./brkitr/root.txt:27: The dependency file ./out/build/icudt69l/brkitr/Burmese_graphclust_model5_heavy.res does not exist. Please make sure it exists.

LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k zh.txt
```

FrankYFTang:
> The builds are producing a message that looks problematic, although it doesn't register as an error. Happens both in this PR, and in main
> 
> ```
> LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k root.txt
> LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k ru.txt
> ./brkitr/root.txt:26: The dependency file ./out/build/icudt69l/brkitr/Thai_graphclust_model4_heavy.res does not exist. Please make sure it exists.
> 
> ./brkitr/root.txt:27: The dependency file ./out/build/icudt69l/brkitr/Burmese_graphclust_model5_heavy.res does not exist. Please make sure it exists.
> 
> LD_LIBRARY_PATH=../lib:../stubdata:../tools/ctestfw:$LD_LIBRARY_PATH  ../bin/genrb -s ./brkitr -d ./out/build/icudt69l/brkitr/ -i ./out/build/icudt69l  -k zh.txt
> ```

This is because by default these lstm are not included in the build w/o a filter file explicit including them.

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,792 @@
...
+    ConstArray1D() : data_(nullptr), d1_(0) {}
+
+    ConstArray1D(const float* data, int32_t d1) : data_(data), d1_(d1) {}
+
+    virtual ~ConstArray1D();
+
+    // Init the object, the object does not own the data nor copy.
+    // It is designed to directly use data from memory mapped resources.
+    void init(const int32_t* data, int32_t d1) {
+        data_ = reinterpret_cast<const float*>(data);
```
sffc, line 74: Please assert that the system uses IEEE doubles so that this cast is safe (and error if not).

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+    ConstArray2D(const float* data, int32_t d1, int32_t d2)
+        : data_(data), d1_(d1), d2_(d2) {}
+
+    virtual ~ConstArray2D();
+
+    // Init the object, the object does not own the data nor copy.
+    // It is designed to directly use data from memory mapped resources.
+    void init(const int32_t* data, int32_t d1, int32_t d2) {
+        data_ = reinterpret_cast<const float*>(data);
```
sffc, line 110: Another cast to check.

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,792 @@
...
+
+LSTMData::LSTMData()
+{
+}
+
+LSTMData::~LSTMData()
+{
+}
+
+class LSTMResourceData : public LSTMData {
```
sffc, line 350: Suggestion (optional): measure the performance difference of using a bare struct with fields, rather than a class with a vtable.  You only have 1 impl of the LSTMData interface.

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,792 @@
...
+        logp.dotProduct(fbRow, fData->outputW()).add(fData->outputB());
+
+        // current = argmax(logp)
+        LSTMClass current = (LSTMClass)logp.maxIndex();
+        // BIES logic.
+        if (current == BEGIN || current == SINGLE) {
+            if (i != 0) {
+                foundBreaks.addElement(offsetsBuf[i], status);
+            }
+        }
```
sffc, line 685: Thought: This is where we could include a BIES normalizer algorithm, because this could result in a break in the middle of a grapheme cluster if code point mode is used (this is fairly rare but can happen, especially on large texts).

icu4c/source/test/testdata/BUILDRULES.py
```diff
@@ -10,6 +10,7 @@ def generate(config, io, common_vars):
     requests += generate_rb(config, io, common_vars)
     requests += generate_sprep(config, io, common_vars)
     requests += generate_conv(config, io, common_vars)
+    requests += generate_lstm(config, io, common_vars)
```
sffc, line 13: Please merge main into your branch so that these diffs go away

icu4c/source/common/lstmbe.cpp
FrankYFTang: currently, the manual_predict, I try to mimic from the python code does not call the bies normalization and I intend to keep a bug-to-bug compatibility during the porting.

icu4c/source/test/testdata/BUILDRULES.py
FrankYFTang: Thie is the BUILDRULES.py under test/testdata not the one under data/

icu4c/source/common/lstmbe.cpp
FrankYFTang: change to struct

icu4c/source/common/lstmbe.cpp
FrankYFTang: Add assert and put if and return error in outter code.

FrankYFTang:
PTAL

icu4c/source/test/testdata/BUILDRULES.py
```diff
@@ -169,6 +170,31 @@ def generate_conv(config, io, common_vars):
         )
     ]
 
+def generate_lstm(config, io, common_vars):
+    basenames = [
+        "Thai_graphclust_model4_heavy",
+        "Thai_codepoints_exclusive_model5_heavy",
+        "Burmese_graphclust_model5_heavy"
```
sffc, line 177: Since all you're doing is converting some txt files to rb files, just add these new files to the list up above in generate_rb

FrankYFTang: done

FrankYFTang:
somehow the testing results are different in windows . Wonder what could the bug.

FrankYFTang:
Since I have hard time to debug on window I add debugging info and turn it on for now. That will help me to narrow down the issues on Windows.

FrankYFTang:
ok, I think I solve the window problem. Somehow the genrb on Windows treat the resources file in test/testdata as in Windows cp 1252 instead of UTF8. I have to use -eUTF-8 in the py to force it to UTF8 to build the test resources correctly. PTAL

sffc:
> ok, I think I solve the window problem. Somehow the genrb on Windows treat the resources file in test/testdata as in Windows cp 1252 instead of UTF8. I have to use -eUTF-8 in the py to force it to UTF8 to build the test resources correctly. PTAL

@jefgen do you have any idea about this?  We have plenty of resource bundle files that contain UTF-8.  How come these three new files need to be special-cased?

sffc (approved):

sffc (approved):
LGTM (but feel free to wait for the other reviewers as well)

FrankYFTang:
> > ok, I think I solve the window problem. Somehow the genrb on Windows treat the resources file in test/testdata as in Windows cp 1252 instead of UTF8. I have to use -eUTF-8 in the py to force it to UTF8 to build the test resources correctly. PTAL
> 
> @jefgen do you have any idea about this? We have plenty of resource bundle files that contain UTF-8. How come these three new files need to be special-cased?

Here is my guess. The resources .txt file under data/ are all include the BOM in UTF8 (ef bb bf)  of the first three bytes and genrb detect it and therefore use UTF-8. The .txt file I newly added are not generated with the BOM in UTF8 and therefore somehow Linux and Mac use UTF-8 as the default but Windows use cp1252 as the default.

FrankYFTang:
I probably should fix this by just adding BOM in UTF8 to the beginning of the files.

jefgen:
> > > ok, I think I solve the window problem. Somehow the genrb on Windows treat the resources file in test/testdata as in Windows cp 1252 instead of UTF8. I have to use -eUTF-8 in the py to force it to UTF8 to build the test resources correctly. PTAL
> > 
> > 
> > @jefgen do you have any idea about this? We have plenty of resource bundle files that contain UTF-8. How come these three new files need to be special-cased?
> 
> Here is my guess. The resources .txt file under data/ are all include the BOM in UTF8 (ef bb bf) of the first three bytes and genrb detect it and therefore use UTF-8. The .txt file I newly added are not generated with the BOM in UTF8 and therefore somehow Linux and Mac use UTF-8 as the default but Windows use cp1252 as the default.

(Sorry for the delayed reply; I was out for the last two days).

Yes, I believe Frank's guess here is correct and the issue is/was that the data files didn't have the UTF-8 BOM at the beginning of the file.

Looking at the `genrb` code, it calls `T_FileStream_open` to open/read the data files. This is mostly just a wrapper that uses  `fopen` to actually open/read the data files. https://github.com/unicode-org/icu/blob/main/icu4c/source/tools/toolutil/filestrm.cpp#L33

The thing to note is that it passes the file name to `fopen` using a `char*`, so not a `wchar_t*`, (which is sometimes called a "narrow" string). This means that the encoding of the file is assumed to be in the system default ANSI codepage, unless there is a BOM or the `mode` argument has an encoding setting set.

From the MSDN docs: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen
> The fopen function opens the file that is specified by filename. By default, a narrow filename string is interpreted using the ANSI codepage (CP_ACP).
> ...
> If the file already exists and is opened for reading or appending, the Byte Order Mark (BOM), if it present in the file, determines the encoding. The BOM encoding takes precedence over the encoding that is specified by the ccs flag.

So for the Windows build bots, the default ANSI codepage is 1252, whereas on Linux and Mac the default system code page is (almost always) going to be UTF-8.

[cc/FYI: @ShawnSteele, as this is perhaps another example of how UTF-8 by default on Windows might help].

icu4c/source/common/lstmbe.cpp
```diff
@@ -0,0 +1,815 @@
...
+
+/**
+ * A class to allocate data as a writable 1D array.
+ * This is the main class implement matrix operation.
+ */
+class Array1D : public ReadArray1D {
+public:
+    Array1D() : memory_(nullptr), data_(nullptr), d1_(0) {}
+    Array1D(int32_t d1)
+        : memory_(uprv_malloc(d1 * sizeof(float))),
```
jefgen, line 166: I'm just wondering, how does this new code handle memory allocation failure? 
(For example, in this constructor, the call to `uprv_malloc` can fail and it might return `nullptr`).

So in this code:
```c++
 .add(Array1D(b.d1()).dotProduct(h, U))
```
If OOM happens when `Array1D` is created, then it's `data_` member will point at `memory_` which is `nullptr`. 
The call to `add` will then attempt to use `data_` without checking if it's `nullptr` first:

```c++
    inline Array1D& add(const ReadArray1D& a) {
        U_ASSERT(a.d1() == d1());
        for (int32_t i = 0; i < d1(); i++) {
            data_[i] += a.get(i);                //  This line will crash/segfault
        }
        return *this;
    }
```

It looks like it will maybe crash/segfault if OOM occurs.

FrankYFTang: Good catch, I will add a new PR to address that.

jefgen: Thanks! :)

FrankYFTang: I intend to address all the memory issues in https://github.com/unicode-org/icu/pull/1707 . Please review. Thanks

jefgen: Thanks Frank!

PR #1384, opened 2020-09-30
ICU-21249 Change AppVeyor to not use parallel build due to Cygwin stability issues

jefgen:
(No code changes in this PR).
The Cygwin parallel build often fails (it is flaky) due to race-condition(s) and issues with `ln -f` failing.
This PR changes the AppVeyor build to be single-threaded in order to have more stability for the CI builds.

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-21249
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [ ] Tests included
- [ ] Documentation is changed or added

PR #1694, opened 2021-04-07
ICU-21535 Port Cygwin CI build from AppVeyor to Azure Pipelines

daniel-ju:
There are no code changes in this PR. 

Now that there's caching functionality in Azure Pipelines, we can port the Cygwin build over and remove the somewhat flaky AppVeyor CI. The cache in Azure Pipeline for a specific key is immutable though, so in order to prevent always using an old/stale cache, the cache will be keyed off of a hash of the contents of the `.azure-pipelines.yml` file itself in combination with a Cygwin version.

The build steps are mostly the same between the two, although I added a step to ensure Git does a checkout with the expected line endings to run `runConfigureICU`.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21535


jefgen:
Thanks @daniel-ju for doing this! 👍

jefgen:
> updated the make command to use 2 cores to speed up the build (from ~60min to ~30min).

IIRC, we had to turn off the parallel build for Cygwin with AppVeyor as it kept having random flaky crashes when using more than one core. :(  
(See https://github.com/unicode-org/icu/pull/1384).

I wonder if Azure CI might be more stable though... I guess we can try using a parallel build again, and if it's still flaky then we can go back to a single-threaded build.

.ci-builds/.azure-pipelines.yml
```diff
@@ -394,3 +394,59 @@ jobs:
       env:
         CC: clang
         CXX: clang++
+#-------------------------------------------------------------------------
+- job: ICU4C_Cygwin_GCC_x86_64_Release
+  displayName: 'C: Cygwin GCC x86_64 Release'
+  timeoutInMinutes: 60
+  pool:
+    vmImage: 'vs2017-win2016'
```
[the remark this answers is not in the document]
daniel-ju: That makes sense, I'll update the image here. Thanks! :)

daniel-ju:
> IIRC, we had to turn off the parallel build for Cygwin with AppVeyor as it kept having random flaky crashes when using more than one core. :(
> (See #1384).

@jefgen Ah, I completely forgot about that change, thanks for bringing it up!

Hmm, considering that the CI build/tests seems to be hanging now but not when I tested this, I think it might be better to switch back to a single threaded build again. I'll investigate a little more.

jefgen:
> Hmm, considering that the CI build/tests seems to be hanging now but not when I tested this, I think it might be better to switch back to a single threaded build again. I'll investigate a little more.

Hmm... It looks like the build might have simply timed-out, rather than failed. 

```
##[error]The job running on agent Azure Pipelines 8 ran longer than the maximum time of 60 minutes. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134
```

I think we might want to increase the value for `timeoutInMinutes` to be more than 60 minutes -- perhaps maybe something like 90 minutes?

.appveyor.yml
```diff
@@ -1,86 +0,0 @@
-# AppVeyor configuration for CI Builds of ICU4C.
```
jefgen, line 1: Oh, one other thought:
I guess we can delete the `.appveyor.xml` lines from some of the build configuration files too. :)

https://github.com/daniel-ju/icu/blob/daju/port-cygwin-build-to-azure-ci/.ci-builds/.azure-valgrind.yml#L26 
https://github.com/daniel-ju/icu/blob/daju/port-cygwin-build-to-azure-ci/.ci-builds/.azure-exhaustive-tests.yml#L25

.ci-builds/.azure-pipelines.yml
```diff
@@ -394,3 +394,59 @@ jobs:
...
+    ICU_CI_CACHE: c:\icu-ci-cache
+    CYG_URL: https://cygwin.com/setup-x86_64.exe
+    CYG_MIRROR: http://mirrors.kernel.org/sourceware/cygwin/
+    CYG_PACKAGES: automake,gcc-core,gcc-g++,make,pkg-config,perl,python3
+    CYG_ROOT: c:\cygwin-root
+    CYG_CACHE: '$(ICU_CI_CACHE)\cygwin64-v3'
+    CYG_CACHED_SETUP: '$(CYG_CACHE)\setup.exe'
+  steps:
+    - script: |
+        git config --global core.autocrlf input
```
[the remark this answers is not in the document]
daniel-ju: I tried using `autocrlf false` but still end up getting errors for `\r` characters. I think there may be files that are already  checked-in using CRLF.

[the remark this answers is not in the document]
jefgen: Oh, good point -- I think you are right.
I think the `*.bat` files are checked-in using CRLF, for example: https://raw.githubusercontent.com/unicode-org/icu/main/icu4c/source/allinone/icucheck.bat has CRLF.

So perhaps this needs to stay as `input` then. I wonder if it might be worth adding a comment though...

.appveyor.yml
daniel-ju: Good point, done! :)

.ci-builds/.azure-pipelines.yml
```diff
@@ -394,3 +394,60 @@ jobs:
...
+      displayName: 'Configure Git to checkout with Unix line endings (LF)'
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
+    # Cache expires after 7 days of no activity.
+    - task: Cache@2
+      displayName: 'Restore Cygwin cache'
+      inputs:
+        key: cygwin
```
jefgen, line 423: Hmm... I was reading a bit more about how the Azure CI caching works and it seems that it doesn't quite work like the AppVeyor caching.

https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#use-cache-task
https://www.appveyor.com/docs/build-cache/

The Azure caches are immutable, meaning that once they are created you can't update/change them at all.
Whereas the AppVeyor caches are more like a "folder" that simply gets saved/restored each time the build is run.

The content of the Azure cache is tied to the "key". Since we use the same "key" to identify the cache each time, I think that this means we will forever be restoring the same (possibly old) cached data repeatedly. It also won't age out either, since there will likely be a build occurring within 7 days, causing a cache hit.

It looks like the general usage pattern for the Azure caches is to have the "key" value generated from the hashed contents of a ".lock" file (which lists out various packages and version numbers). However, I don't think there is really an equivalent for the Cygwin setup off-hand.

I'm trying to think of other ways that we could generate and/or use a new "key" value so that we won't end up with the same stale cache being used over and over again.

However, the other tricky part is that the file needs to exist when the `Cache@2` task first runs -- meaning that it effectively needs to be checked in to the git repo, as the Cygwin setup won't exist at all during the very first initial run...

I wonder if we could use the `.azure-pipelines.yml` file itself? The key would be the hashed content of the file, meaning that if anyone touches the file, then the cache would be updated. This would at least be better than it *never* being updated. Additionally, if we ever changed the set of `CYG_PACKAGES` (automake,gcc-core,gcc-g++,make,pkg-config,perl,python3) then I think we'd want to update the cache as well.

I think we could do this by changing the key to this:

```yml
    # Use the contents of the file ".azure-pipelines.yml" as part of the key, as that contains the list of CYG_PACKAGES.
    - task: Cache@2
      displayName: 'Restore Cygwin cache'
      inputs:
        key: 'cygwin | .ci-builds/.azure-pipelines.yml'
```

What do you think? 😊

daniel-ju: Thanks Jeff, you bring up a great point about the cache being immutable. I like your idea of using the `.azure-pipelines.yml` file itself so that we can update Cygwin if we changed `CYG_PACKAGES`. I was thinking in addition to that, we should include the Cygwin version as part of the key as well, so that when a new version is released we would be able to update the version here:

```yml
    - task: Cache@2
      displayName: 'Restore Cygwin cache'
      inputs:
        # Use the contents of the file ".azure-pipelines.yml" as part of the key, as that contains the list of CYG_PACKAGES.
        # Also include the Cygwin version as part of the key. If we want to use a newer version of Cygwin, we can update the key here.
        # Note: The version in the key may become out of sync with (older than) the version we actually use if we update this file
        #       without updating the key. Any updates to this file guarantees that we're use the latest version.
        key: 'cygwin-3.2.0 | .ci-builds/.azure-pipelines.yml'
```
This would effectively be the same since we need to update this file to change the version anyway, but provides a place to do so. How does this sound to you?

.ci-builds/.azure-pipelines.yml
```diff
@@ -394,3 +394,64 @@ jobs:
...
+      fetchDepth: 10
+    # Cache expires after 7 days of no activity.
+    - task: Cache@2
+      displayName: 'Restore Cygwin cache'
+      inputs:
+        # Use the contents of the file ".azure-pipelines.yml" as part of the key, as that contains the list of CYG_PACKAGES.
+        # Also include the Cygwin version as part of the key. If we want to use a newer version of Cygwin, we can update the key here.
+        # Note: The version in the key may become out of sync with (older than) the version we actually use if we update this file
+        #       without updating the version in the key. Any updates to this file guarantees that we're using the latest version.
+        key: '"cygwin-3.2" | .ci-builds/.azure-pipelines.yml'
```
jefgen, line 427: Just curious, does the cygwin-3.2 need to be enclosed with double quotes? (FWIW, I don't think it really matters, I was just curious if you maybe tried without and needed to add them for some reason perhaps...)

daniel-ju: Yeah, we need the double quotes here to tell Azure CI that it's a string and not a file path. Otherwise it would error with file not found. The Azure caching docs say 'Keep in mind that any key segment that "looks like a file path" will be treated like a file path', so I guess adding `.ci-builds/.azure-pipelines.yml` as part of the key made it think the unquoted string was a file path.

jefgen: Thanks Daniel.

Hm, in that case I think the Caching docs for Azure pipelines are wrong/misleading.

```
    key: 'yarn | "$(Agent.OS)" | yarn.lock'
```
> In this example, the cache key contains three parts: a static string ("yarn"), the OS the job is running on since this cache is unique per operating system, and the hash of the yarn.lock file that uniquely identifies the set of dependencies in the cache.

I'm going to file an issue on the page for it.

jefgen: Filed an issue here: https://github.com/MicrosoftDocs/azure-devops-docs/issues/10500

daniel-ju: Actually the docs call out this (I unintentionally excluded the 2nd sentence above):

> This file must exist at the time the task is run. Keep in mind that any key segment that "looks like a file path" will be treated like a file path. In particular, this includes segments containing a `.`.

So since `cygwin-3.2` includes a `.`, it treats it as a file path without the quotes.

jefgen: Hm, I wonder if it is maybe only the `.` that causes it to get messed up then?
(Though in any case, if someone had a file named "yarn" then I think the example would still get messed up.)
I like the idea of using a variable for the version! :)

jefgen (approved):
lgtm, thank you! :)

.ci-builds/.azure-pipelines.yml
daniel-ju: I just tested this with just `cygwin-3.2` as the key without quotes, and Azure CI thinks it's a file -- so it seems it's just the `.` that makes the difference. But good point, filenames without `.` at all would still get messed up. Thanks for filing the issue! :)

PR #1695, opened 2021-04-07
ICU-21578 Merge maint/maint-69 into main

sffc:
Unfortunately it wasn't a clean merge, because icudata.jar changed on both branches separately (CLDR 39 on the maint branch, and laodict on the main branch).  So I had to make an intermediate branch with the merge commit.  I re-generated the jar files using the instructions in icu4j-readme.txt.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21578

ALLOW_MANY_COMMITS=true
DISABLE_JIRA_ISSUE_MATCH=true

### DIFF
--- a/icu4c/source/common/charstr.cpp
+++ b/icu4c/source/common/charstr.cpp
@@ -14,6 +14,8 @@
 *   created by: Markus W. Scherer
 */
 
+#include <cstdlib>
+
 #include "unicode/utypes.h"
 #include "unicode/putil.h"
 #include "charstr.h"
--- a/icu4c/source/common/edits.cpp
+++ b/icu4c/source/common/edits.cpp
@@ -86,6 +86,7 @@ Edits &Edits::moveArray(Edits &src) U_NOEXCEPT {
 }
 
 Edits &Edits::operator=(const Edits &other) {
+    if (this == &other) { return *this; }  // self-assignment: no-op
     length = other.length;
     delta = other.delta;
     numChanges = other.numChanges;
--- a/icu4c/source/common/filteredbrk.cpp
+++ b/icu4c/source/common/filteredbrk.cpp
@@ -20,6 +20,7 @@
 #include "ubrkimpl.h" // U_ICUDATA_BRKITR
 #include "uvector.h"
 #include "cmemory.h"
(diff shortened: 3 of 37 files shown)


sffc:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

What does this mean?  @jefgen

jefgen:
> > Azure Pipelines could not run because the pipeline triggers exclude this branch/path.
> 
> What does this mean? @jefgen 

Hmm, I thought at first that maybe this might be due to the rename of `master` to `main`. Perhaps the YML file might still have `master`, which would explain why the branch `main` didn't work.
However, looking in the intermediate branch I do see `main` in the config file here:
https://github.com/sffc/icu/blob/ICU-21578-merge/.ci-builds/.azure-exhaustive-tests.yml#L13

I checked the manual config for the azp bot, and FWIW, it did still have `master` in a few places. I updated them just now and I'll try a build request again. But if that doesn't work then I'll kick off a manual build.

jefgen:
> Azure Pipelines could not run because the pipeline triggers exclude this branch/path.

Hm, I'm not sure off-hand... I'll need to take another look later.

I've kicked off a manual build here for now:
https://dev.azure.com/ms/icu/_build/results?buildId=154271&view=results
(though the status should still be reported as one of the "checks").

jefgen:
Update: The manual run for the Exhaustive tests all passed.

jefgen:
Note: For comparison I looked at the diffs here: https://github.com/unicode-org/icu/compare/main...maint/maint-69
However, they aren't quite the same as the diffs on GitHub seem to show changes that are already in *both* branches (like the 'zzzz' change in [locid.cpp](https://github.com/unicode-org/icu/compare/main...maint/maint-69#diff-57a1b687829da97c47474809de6546f01672a20a5902054e78b32bdb2cd6d62fR1524) for example).
Otherwise, going through the changes for each commit, the merge here looks good. 👍

jefgen (approved):
lgtm, thanks @sffc!

sffc:
Merged via merge commit.

PR #1699, opened 2021-04-18
ICU-21591 Release lock in SimpleDateFormat::tzFormat in case of failure

filipnavara:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21591

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -4313,11 +4313,9 @@ SimpleDateFormat::tzFormat(UErrorCode &status) const {
         {
             if (fTimeZoneFormat == NULL) {
                 TimeZoneFormat *tzfmt = TimeZoneFormat::createInstance(fLocale, status);
-                if (U_FAILURE(status)) {
-                    return NULL;
+                if (U_SUCCESS(status)) {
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -4313,11 +4313,9 @@ SimpleDateFormat::tzFormat(UErrorCode &status) const {
         {
```
aheninger, line 4313: The double-checked lock idiom used here is also unsafe. If could be fixed by making fTimeZoneFormat an atomic type, and using the appropriate atomic access functions, but doing so would be awkward and complicated because fTimeZoneFormat is declared in a public header. The most expedient fix is probably to ditch the double checked idiom, and just use the mutex.

aheninger:
Looking further into the uses of SimpleDateFormet::fTimeZoneFormat, it looks like there's another thread safety problem in the assignment operator.

If you like, I could spin up a new pull request to address all of these, including the original problem that you identified.

filipnavara:
> If you like, I could spin up a new pull request to address all of these, including the original problem that you identified.

I'd be happy if you do so. Otherwise I will get back to it on the weekend.

aheninger:
I opened #1701 with the additional related fixes.

PR #1701, opened 2021-04-25
ICU-21591 Release lock in SimpleDateFormat::tzFormat in case of failure

aheninger:
Also remove the use of the unsafe double-checked lock idiom in the same
function, SimpleDateFormat::tzFormat(). Synchronization now always uses a
mutex, which is slower, but in the context of format or parse operations,
shouldn't be significant.

Added synchronization to one more unsafe direct reference to a const
SimpleDateFormat::fTimeZoneFormat. In the assignment operator.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21591

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -64,6 +64,7 @@
 #include "uassert.h"
 #include "cmemory.h"
 #include "umutex.h"
+#include "mutex.h"
 #include <float.h>
 #include "smpdtfst.h"
 #include "sharednumberformat.h"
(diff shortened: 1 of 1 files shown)


aheninger:
Followup replacement for #1699. Hoping to avoid making extra work for filipnavara, who identified the original problem.

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -594,11 +595,20 @@ SimpleDateFormat& SimpleDateFormat::operator=(const SimpleDateFormat& other)
...
-    if (other.fTimeZoneFormat) {
-        fTimeZoneFormat = new TimeZoneFormat(*other.fTimeZoneFormat);
+    fTimeZoneFormat = nullptr;
+    TimeZoneFormat *otherTZFormat;
+    {
+        Mutex m(&LOCK);
+        otherTZFormat = other.fTimeZoneFormat;
+    }
+    if (otherTZFormat) {
+        fTimeZoneFormat = new TimeZoneFormat(*otherTZFormat);
```
markusicu, line 611: I had to think through your change and comment here...

I think the main point is that this pointer is lazy-initialized, so in normal, valid use there could be a race from nullptr to pointer to object, but not a race in transition *to* nullptr. That is, if you grab the pointer and it's not nullptr you need not remain inside the Mutex because the object is fully built and won't be deleted. Right?

Please beef up the comment...

markusicu (commented):
Thanks!
lgtm pse squash

markusicu (approved):
lgtm tnx

PR #1704, opened 2021-04-29
ICU-21569 Add GA to test LSTM configuration

FrankYFTang:
##### Checklist

##### Content of the PR

1. Add GA to test BreakIterator under LSTM configuration (remove Thai
and Burmese dictionary and include Thai and Burmese LSTM)
2. Add LSTMDataName for the purpose of testing.
3. Add file base test code to test BreakIterator match results from test
file generated by pythong code in
https://github.com/unicode-org/lstm_word_segmentation/blob/master/segment_text.py
4. Fix a LSTMBreakEngine::divideUpDictionaryRange bug when the return value
should only contains the number of words found when the passed in foundBreaks
already contains some data.
5. Change the cintltest TestSwapData from testing thaidict to laodict so
it will not break while we filter out thaidict under the LSTM
configuration.


icu4c/source/test/intltest/intltest.cpp
```diff
@@ -2428,6 +2429,27 @@ UChar *IntlTest::ReadAndConvertFile(const char *fileName, int &ulen, const char
     return retPtr;
 }
 
+#if !UCONFIG_NO_BREAK_ITERATION
+UBool LSTMDataIsBuilt() {
```
sffc, line 2433: Suggestion: Add these methods to a special subclass of `IntlTest` in `itrbbi.h`, similar to `IntlTestWithFieldPosition` in `itformat.h`.

sffc (commented):
I reviewed the PR, but only skimmed the test function; let me know if you'd like me to read the test function in more detail.

icu4c/source/test/intltest/intltest.cpp
FrankYFTang: The problem is we need to access these methods on any classes which internally use rbbi, not just the test on rbbi itself. For example, we need to call it in TransliteratorTest::TestThai inside transtst.cpp  .

sffc (approved):

PR #1706, opened 2021-04-30
ICU-21569 LSTM Part 3 Add Java implementation

FrankYFTang:
ICU-21569 Add Unit tests

##### Checklist


FrankYFTang:
@sffc - I need to remove the removal of dictionary from the filter- otherwise the jar won't be built because the "dependency" rule in the root.txt . so the filter file is change from  1) add th/my lstm, and 2) take out th/my dictionary and to only 1) but should be good enough for testing purpose.

aheninger:
Taking myself off as reviewer for this PR. I will have limited time in the near term.

FrankYFTang:
PTAL

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/LSTMBreakEngine.java
```diff
@@ -0,0 +1,444 @@
...
+            for (char c = fIter.current();
+                 c != CharacterIterator.DONE && fIter.getIndex() < rangeEnd;
+                 c = fIter.next()) {
+                offsets.add(fIter.getIndex());
+                indicies.add(getIndex(String.valueOf(c)));
+            }
+        }
+    }
+
+    class GraphemeClusterVectorizer extends Vectorizer {
```
sffc, line 178: Thought (not actionable): This can probably be optimized at some point to reduce memory allocations (e.g., of the substrings).

FrankYFTang: please articulate your idea. I am not sure sure what you are sugestting here.

sffc (approved):

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/LSTMBreakEngine.java
sffc: If you had a CharSequence input instead of CharacterIterator, you could make a subSequence of the CharSequence which is probably more lightweight than your StringBuilder approach.  But, as I said, this requires more code churn and is probably premature optimization.

FrankYFTang:
squashed, PTAL

sffc (approved):

PR #1711, opened 2021-05-04
ICU-21556 Support currency as decimal separator in patterns

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21556

### DIFF
--- a/icu4c/source/i18n/number_formatimpl.cpp
+++ b/icu4c/source/i18n/number_formatimpl.cpp
@@ -352,10 +352,11 @@ NumberFormatterImpl::macrosToMicroGenerator(const MacroProps& macros, bool safe,
         return nullptr;
     }
     fPatternModifier.adoptInstead(patternModifier);
-    patternModifier->setPatternInfo(
-            macros.affixProvider != nullptr ? macros.affixProvider
-                                            : static_cast<const AffixPatternProvider*>(fPatternInfo.getAlias()),
-            kUndefinedField);
+    const AffixPatternProvider* affixProvider =
+        macros.affixProvider != nullptr
+            ? macros.affixProvider
+            : static_cast<const AffixPatternProvider*>(fPatternInfo.getAlias());
+    patternModifier->setPatternInfo(affixProvider, kUndefinedField);
     patternModifier->setPatternAttributes(fMicros.sign, isPermille, macros.approximately);
     if (patternModifier->needsPlurals()) {
         patternModifier->setSymbols(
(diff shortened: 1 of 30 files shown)


icu4c/source/i18n/number_formatimpl.cpp
```diff
@@ -524,15 +528,30 @@ int32_t NumberFormatterImpl::writeNumber(const MicroProps& micros, DecimalQuanti
...
+            if (micros.currencyAsDecimal) {
+                length += string.insert(
                     length + index,
-                    micros.useCurrency ? micros.symbols->getSymbol(
-                            DecimalFormatSymbols::ENumberFormatSymbol::kMonetarySeparatorSymbol) : micros
-                            .symbols
-                            ->getSymbol(
-                                    DecimalFormatSymbols::ENumberFormatSymbol::kDecimalSeparatorSymbol),
+                    // Note: This unconditionally substitutes the standard short symbol.
+                    // TODO: Should we support narrow or other variants?
```
macchiati, line 535: Yes, we should support at least narrow and standard short. And given that, I think it is cleanest and most understandable to support all.

macchiati (changes requested):
Code looks good. I recommend one change.

icu4c/source/i18n/number_formatimpl.cpp
sffc: Okay. I wrote the code this way since the short symbol was easy to access at this point in the code. But I'll go ahead and rewire things so that I can get access to other symbol widths from here.

macchiati: Did you want to submit this first for now, and then follow this up under the same ticket to support all the symbols?

sffc: Yes, given the imminent CLDR deadline, the expedient thing would be to check this in, but leave the ticket open, and follow up to finish the ticket before ICU 70.  (This is a very busy week with my intern starting...)

younies:
@sffc , based on your discussion with @macchiati, this should be submitted by now. any updates?

sffc:
I assume CLDR pulled this change in for Survey Tool? Since it's now closer to the ICU 70 release, I'll revisit this ticket and address Mark's feedback before merging.

sffc:
^ this is rebasing against main.  I will follow up soon with the actual code changes to reflect Mark's feedback.

icu4c/source/i18n/number_formatimpl.cpp
sffc: Okay, I'm finally addressing this.  So, an issue right now is that we're parsing only a single currency placeholder ('¤') as the decimal separator.  But how do we know whether to use the narrow or formal symbol?

Eventually, once we have this pattern in CLDR, we can do this in NumberFormatter.  However, since patterns are only available in DecimalFormat, we don't currently have a way to select the negative symbol in the ICU API.

But, I will wire it up nonetheless.

macchiati (dismissed):

markusicu (approved):
rslgtm after squash

PR #1712, opened 2021-05-05
ICU-21569 LSTM Optimize memory usage.

FrankYFTang:
Refactor the array allocation out of compute method and out of the loop
to reduce repeated malloc/free. Move all the array allocation code to
the same place before the loop.
This should dramatically reduce the number of calls to malloc/free.

##### Checklist

### DIFF
--- a/icu4c/source/common/lstmbe.cpp
+++ b/icu4c/source/common/lstmbe.cpp
@@ -209,12 +209,11 @@ public:
         return Array1D(data_ + from, size);
     }
 
-    // Dot product of a 1D array and a 2D array into this one.
-    inline Array1D& dotProduct(const ReadArray1D& a, const ReadArray2D& b) {
+    // Add dot product of a 1D array and a 2D array into this one.
+    inline Array1D& addDotProduct(const ReadArray1D& a, const ReadArray2D& b) {
         U_ASSERT(a.d1() == b.d1());
         U_ASSERT(b.d2() == d1());
         for (int32_t i = 0; i < d1(); i++) {
-            data_[i] = 0;
             for (int32_t j = 0; j < a.d1(); j++) {
                 data_[i] += a.get(j) * b.get(j, i);
             }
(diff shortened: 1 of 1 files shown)


roubert:
Hm, this is all code I've never seen before. Is there maybe someone who has seen it before who could review this instead?

FrankYFTang:
> Hm, this is all code I've never seen before. Is there maybe someone who has seen it before who could review this instead?

I recently added this code. This PR is a simple mechanical refactoring only. Basically we used to have

for loop {
   call compute(arg1a, arg2a ... arg xa, status)
}
...
for loop {
   call compute(arg1b, arg2b ... arg xb, status)
}
compute () {
  Array1D hU
  // use hU as temp memory
....
  Array1D ic
  // use ic as temp memory
...
}

and I now change it to 

  Array1D hU
  Array1D ic
for loop {
   call compute(arg1a, arg2a ... arg xa, w1, w2)
}
...
for loop {
   call compute(arg1b, arg2b ... arg xb, w1, w2)
}
compute ( .... Array1D& hu, Array1D& ic) {
  // use hU as temp memory
....
  // use ic as temp memory
...
}

so hU and ic will only be called allocated on the stack once and won't need to be called inside the compute repeatedly

Notice, there are two version of Array1D constructor, the one which passed in a status will malloc and the other just take the memory which do not need to malloc . The one which take the pass in memory is cheap to allocate on the stack, the one which take a number and status which malloc internally is expensive. 

I change move all the calls which need to malloc inside compute out so we no longer need to pass in status to the compute and check after the return. All the status check of memory allocation now is done once before the first for loop.

There are really no logic change in this PR, just moving the code out of function and move some line up in the function. All the code before/after should be logical equivalent

icu4c/source/common/lstmbe.cpp
```diff
@@ -589,36 +589,29 @@ void GraphemeClusterVectorizer::vectorize(
...
-    Array1D ifco(b.d1(), status);
-    {
-        Array1D hU(b.d1(), status);
-        if (U_FAILURE(status)) return;
-        ifco.dotProduct(x, W)
-            .add(hU.dotProduct(h, U))
-            .add(b);
-    } // delocate hU
+    ifco.dotProduct(x, W)
+        .add(hU.dotProduct(h, U))
```
sffc, line 603: Suggestion (optional): Add a new specialized function `.addDotProduct` that avoids the need for allocating `hU`

FrankYFTang:
PTAL

sffc (dismissed):

FrankYFTang:
SQUASHED, please approve again.

sffc (approved):

PR #1716, opened 2021-05-11
ICU-21569 Remove extra LocalUResourceBundlePointer

FrankYFTang:
Remove LocalUResourceBundlePointer fDictRes which shadowed the real
UUResourceBundle*.

ICU-21569 fix assert

##### Checklist

### DIFF
--- a/icu4c/source/common/lstmbe.cpp
+++ b/icu4c/source/common/lstmbe.cpp
@@ -408,18 +408,19 @@ LSTMData::LSTMData(UResourceBundle* rb, UErrorCode &status)
     int32_t data_len = 0;
     const int32_t* data = ures_getIntVector(fDataRes, &data_len, &status);
     if (U_FAILURE(status)) return;
-    LocalUResourceBundlePointer fDictRes(
-        ures_getByKey(rb, "dict", nullptr, &status));
-    int32_t num_index = ures_getSize(fDictRes.getAlias());
+    fDictRes = ures_getByKey(rb, "dict", nullptr, &status);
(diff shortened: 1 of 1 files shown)


icu4c/source/common/lstmbe.cpp
```diff
@@ -408,18 +408,19 @@ LSTMData::LSTMData(UResourceBundle* rb, UErrorCode &status)
     int32_t data_len = 0;
     const int32_t* data = ures_getIntVector(fDataRes, &data_len, &status);
     if (U_FAILURE(status)) return;
-    LocalUResourceBundlePointer fDictRes(
-        ures_getByKey(rb, "dict", nullptr, &status));
-    int32_t num_index = ures_getSize(fDictRes.getAlias());
+    fDictRes = ures_getByKey(rb, "dict", nullptr, &status);
```
sffc, line 411: I think if you want to re-use the `rb`, you need to pass it as the third `fillIn` argument.  Otherwise you leak memory.

FrankYFTang: I cannot re-use rb since I need to keep fDictRes open till the destruction of this object because I am using the string inside the resources in the hash table until the destructor got called and I already call ures_close(fDictRes); in line 470 inside the destructor.

sffc: Oh I see now. You had a shadow variable named fDictRes and now you're getting rid of the shadow variable.

sffc (approved):

PR #1718, opened 2021-05-12
ICU-21569 LSTM fix resource issue.

FrankYFTang:
Remember the Bundle in LSTMData and close in the destructor
because fName is pointing to string data inside rb.

##### Checklist

### DIFF
--- a/icu4c/source/common/lstmbe.cpp
+++ b/icu4c/source/common/lstmbe.cpp
@@ -373,13 +373,12 @@ struct LSTMData : public UMemory {
     ConstArray1D fOutputB;
 
 private:
-    UResourceBundle* fDataRes;
-    UResourceBundle* fDictRes;
+    UResourceBundle* fBundle;
 };
 
 LSTMData::LSTMData(UResourceBundle* rb, UErrorCode &status)
     : fDict(nullptr), fType(UNKNOWN), fName(nullptr),
-      fDataRes(nullptr), fDictRes(nullptr)
+      fBundle(rb)
 {
     if (U_FAILURE(status)) {
         return;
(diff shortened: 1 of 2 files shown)


FrankYFTang:
Markus- could you take a look at how I keep the resource open and close it in the destructor? Is this the right way to do this? or should I only  keep rb open and do not need to remember fDataRes and fDictRes to close them in the destructor since rb is owning the memory anyway?

sokarax (dismissed):

markusicu (dismissed):
lgtm tnx

icu4c/source/common/lstmbe.cpp
```diff
@@ -403,24 +402,21 @@ LSTMData::LSTMData(UResourceBundle* rb, UErrorCode &status)
...
-    if (U_FAILURE(status)) return;
-    U_ASSERT(fDictRes != nullptr);
-    int32_t num_index = ures_getSize(fDictRes);
+    const int32_t* data = ures_getIntVector(dataRes.getAlias(), &data_len, &status);
+    int32_t num_index = ures_getSize(dictRes.getAlias());
     fDict = uhash_open(uhash_hashUChars, uhash_compareUChars, nullptr, &status);
     if (U_FAILURE(status)) return;
 
-    ures_resetIterator(fDictRes);
+    ures_resetIterator(dictRes.getAlias());
```
markusicu, line 414: optional: You should not need to reset this if you have not iterated over this resource before.

optional*2: You could look at ures_getValueWithFallback() (in https://github.com/unicode-org/icu/blob/main/icu4c/source/common/uresimp.h) for another, more modern, way of reading resource data via ResourceDataValue (sink API, see https://github.com/unicode-org/icu/blob/main/icu4c/source/common/resource.h). `getArray()` then `while (array.getValue(index, value)) { const UChar* str = value.getString(...); ... }`

markusicu (approved):
lgtm but some ideas

markusicu (commented):
tnx pse squash

markusicu (approved):
lgtm tnx

FrankYFTang:
rebase and force push approve again

sffc (approved):
Rubber Stamp

PR #1719, opened 2021-05-14
ICU-21569 Revert "ICU-21569 LSTM Add test to test long string."

FrankYFTang:
Reverts unicode-org/icu#1717

### DIFF
--- a/icu4c/source/test/intltest/lstmbetst.cpp
+++ b/icu4c/source/test/intltest/lstmbetst.cpp
@@ -28,8 +28,6 @@ void LSTMBETest::runIndexedTest( int32_t index, UBool exec, const char* &name, c
     TESTCASE_AUTO(TestThaiGraphclust);
     TESTCASE_AUTO(TestThaiCodepoints);
     TESTCASE_AUTO(TestBurmeseGraphclust);
-    TESTCASE_AUTO(TestThaiGraphclustWithLargeMemory);
-    TESTCASE_AUTO(TestThaiCodepointsWithLargeMemory);
 
     TESTCASE_AUTO_END;
 }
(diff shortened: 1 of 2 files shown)


FrankYFTang:
My compiler setup won't let me reproduce the problem (it tell me during runConfigure I don't have the compiler) so I would like to rollback to fix the trunk first and take more time to reproduce the issue and fix it before my next attempt.

jefgen (approved):
sgtm, thanks Frank!

FrankYFTang:
> sgtm, thanks Frank!

hum.... I am puzzle. The same test break in this revert PR. so even if I revert the PR the problem is still here. So that is NOT cause by this PR

FrankYFTang:
> > sgtm, thanks Frank!
> 
> hum.... I am puzzled now. The same test break in this revert PR. so even if I revert the PR the problem is still here. So that is NOT cause by this PR

ok, I think I am reverting the wrong PR. sorry about this.

sffc:
The error happens right at the top of intltest:

```
   utility {
      MultithreadTest {
         TestThreads {
Makefile:111: recipe for target 'check-local' failed
make[2]: Leaving directory '/home/vsts/work/1/s/icu4c/source/test/intltest'
make[2]: *** [check-local] Segmentation fault (core dumped)
```

That failure occurs on a CI that doesn't run the LSTM code.  So I wonder if this is a flaky bot.  @jefgen ?

jefgen:
> That failure occurs on a CI that doesn't run the LSTM code. So I wonder if this is a flaky bot. @jefgen ?

Yeah, that is a good point. It is somewhat odd though in that it seems to be failing somewhat consistently though. The last two commits on `main` failed, and the last two most recent PRs also failed in the same way as well.

Perhaps the build bot image was updated recently?

FrankYFTang:
wrong approach. The real bug is the new version of docker image.

jefgen:
The issue turned out to be in the Ubuntu 18.04 image recent update, and not related to any ICU code changes at all.

There is a work-around in the PR here: https://github.com/unicode-org/icu/pull/1723

PR #1720, opened 2021-05-14
Revert "ICU-21569 Remove extra LocalUResourceBundlePointer"

FrankYFTang:
Reverts unicode-org/icu#1716


FrankYFTang:
wrong approach, The bug is in the env

jefgen:
The issue turned out to be in the Ubuntu 18.04 image recent update, and not related to any ICU code changes at all.

There is a work-around in the PR here: https://github.com/unicode-org/icu/pull/1723

PR #1732, opened 2021-05-21
ICU-21627 - Fix Makefile bugs

pchemguy:
The two libraries are in the wrong order causing the static build to fail (at least on Windows). There are three more Makefile's (in tests and samples) with the same issue. However, those have no apparent effect on the build process (it completes successfully), so I did not fix them.

Static libraries should not use numeric suffixes on the MinGW/Windows platform. Additionally, the static "s" suffix should also be dropped.

https://unicode-org.atlassian.net/browse/ICU-21627


jefgen:
>  markusicu changed the base branch from maint/maint-68 to main

Ah, I think this might have caused git/GitHub to get confused, as now this PR shows 15 commits to be merged into `main` instead of just one.

I think the commit for this PR might need to be rebased onto `main` and then this PR could be forced pushed to be updated.

jefgen:
Thank you very much for creating this PR, signing the CLA, and also for filing a ticket! :)

FWIW, the changes in the commit here: https://github.com/unicode-org/icu/pull/1732/commits/956683a75ed39c470b0e388d11ffd8e55c641edf seem mostly fine to me. 

However, I'm not super familiar with the MinGW build environment. 
There is another recent PR #1733, which I think will conflict with these changes though.

For the static library changes, that PR removes `STATIC_PREFIX_WHEN_USED` -- which perhaps does a similar thing to setting it to empty string I think (?).

Would you be able to take a look at the other PR?
It would be great to have people that are more familiar with MinGW, in order to help to review MinGW related the changes.

Thank you!

pchemguy:
> Thank you very much for creating this PR, signing the CLA, and also for filing a ticket! :)
> 
> FWIW, the changes in the commit here: [956683a](https://github.com/unicode-org/icu/commit/956683a75ed39c470b0e388d11ffd8e55c641edf) seem mostly fine to me.
> 
> However, I'm not super familiar with the MinGW build environment.
> There is another recent PR #1733, which I think will conflict with these changes though.
> 
> For the static library changes, that PR removes `STATIC_PREFIX_WHEN_USED` -- which perhaps does a similar thing to setting it to empty string I think (?).
> 
> Would you be able to take a look at the other PR?
> It would be great to have people that are more familiar with MinGW, in order to help to review MinGW related the changes.
> 
> Thank you!

I can take my commit, replace source/config/mh-mingw with [mh-mingw](https://github.com/unicode-org/icu/blob/f7761e380bcd716f0a64a8b1f9717c64ac2e4d68/icu4c/source/config/mh-mingw) from "ICU-21628 MinGW/Windows incorrect build #1733" and it builds statically (Windows/MinGW).

srl295:
@pchemguy you do need to rebase this on `main` as mentioned

pchemguy:
> @pchemguy you do need to rebase this on `main` as mentioned
@srl295
I just did and created https://github.com/unicode-org/icu/pull/1735

autoantwort:
This PR can be closed in favor of #1735

pchemguy:
Closed in favor of #1735

PR #1735, opened 2021-05-27
ICU-21627 - Fix Makefile bugs

pchemguy:
The two libraries are in the wrong order causing the static build to fail (at least on Windows). There are three more Makefile's (in tests and samples) with the same issue. However, those have no apparent effect on the build process (it completes successfully), so I did not fix them.

Static libraries should not use numeric suffixes on the MinGW/Windows platform. Additionally, the static "s" suffix should also be dropped.

https://unicode-org.atlassian.net/browse/ICU-21627

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21627


longnguyen2004:
Thanks for this PR!
I've tried cross-compiling from Linux to MinGW, and the resulting data lib has the wrong name (`icudt.a` instead of `libicudt.a`) and is in the wrong directory (`bin` instead of `lib`). Can you look into why this is the case?

pchemguy:
> Thanks for this PR!
> I've tried cross-compiling from Linux to MinGW, and the resulting data lib has the wrong name (`icudt.a` instead of `libicudt.a`) and is in the wrong directory (`bin` instead of `lib`). Can you look into why this is the case?

pchemguy:
This is strange. I just compiled again (Windows/MSYS2?MinGW) a few times, and I noticed that I also have issues with *icudt.*:
In the build/lib I get "liblibicudt.dll.a" and "libliblibicudt.a", that is with 2x and 3x "lib". In the dist/bin, I also have *icudt.a (the rest are in the dist/bin, but the file is named "libliblibicudt.a" (3x "lib"), not "icudt.a".
The problem is, I am not very experienced with makefiles, and the ICU build system with all these recursive calls and multiple makefiles spread across various dirs with very few comments, it is really hard for me to pinpoint the lines where the relevant name of a given target is actually formed.

longnguyen2004:
You can try to run pkgdata with `--verbose` to see debug info about the library name, MinGW is a bit of a mess right now.
Also, pkgdata also need some rework to properly accommodate cross compilation. I've already created an issue for that.

utzcoz:
I also encountered similar problem when trying to build icu4c in Windows 10's msys2 + MinGW64 environment. These are steps that we can use to reproduce failures:

```
# Start MSYS2 MinGW64 Terminal Window with built-in MinGW64 environment
pacman -Syu
pacman -Sy base-devel mingw-w64-x86_64-toolchain
pacman -Sy git mingw-w64-x86_64-cmake mingw-w64-x86_64-ninja
git clone https://github.com/unicode-org/icu.git
cd icu/icu4c/source
./runConfigureICU MinGW --enable-static --disable-shared
make -j4
```

And we can get some undefined references not found failures. But if we remove `--disable-shared` from `runConfigureICU`, it can pass the building.

PR #1742, opened 2021-06-15
ICU-21581 integrate CLDR release-40-m1 to ICU trunk

pedberg-icu:
##### Checklist

This integrates into ICU the CLDR data in xml as of a few days after the start of data submission for CLDR 40; this is tagged CLDR "release-40-m1". This should have most or all of the new structure going into CLDR 40, but does not yet have any results of data submission and vetting via the CLDR Survey Tool.

Some notable items:
- This includes an ICU version of the new CLDR datetimeSkeletons for standard patterns added per [CLDR-13425](https://unicode-org.atlassian.net/browse/CLDR-13425). In ICU this is currently a new DateTimeSkeletons bundle whose entries parallel the first 8 entries in DateTimePatterns, but this can be changed. This affects most locale files.
- New measure units "concentr: item" and "force: kilowatt-hour-per-100-kilometer" per [CLDR-14504](https://unicode-org.atlassian.net/browse/CLDR-14504) and side fix under [CLDR-14824](https://unicode-org.atlassian.net/browse/CLDR-14824).
- Time cycles and time preferences are updated for Chinese regions and locales per [CLDR-10560](https://unicode-org.atlassian.net/browse/CLDR-10560): Time formats for CN (and zh) switch from ah to HH, time formats for TW (and zh_Hant) switch from ah to Bh, and skeleton character C no longer results in Bh for CN, HK, MO.
- Unicode 14 scripts are added per [CLDR-14801](https://unicode-org.atlassian.net/browse/CLDR-14801)
- New Yukon metazone per [CLDR-14709](https://unicode-org.atlassian.net/browse/CLDR-14709)
- And other changes: Adjustments to grammar data, more availableFormats, remove/add some narrow symbols for BYN/RUR, etc.

Note: in [ICU-21569](https://unicode-org.atlassian.net/browse/ICU-21569), `icu4c/source/data/brkitr/root.txt` was manually edited to add `lstm` entries without adding the corresponding source data in `icu4c/source/data/xml/brkitr/root.xml`, so the normal CLDR-ICU integration deletes them. In this PR those `lstm` entries in `icu4c/source/data/brkitr/root.txt` are manually restored, but for the next CLDR-ICU integration, `icu4c/source/data/xml/brkitr/root.xml` should be updated to generate them automatically.


sffc (approved):
Looks great!

icu4c/source/data/misc/supplementalData.txt
```diff
@@ -11357,6 +11362,7 @@ supplementalData:table(nofallback){
                 "energy-british-thermal-unit",
                 "force-pound-force",
                 "force-newton",
+                "force-kilowatt-hour-per-100-kilometer",
```
markusicu, line 11365: Huh? That's a measure of consumption, not of force.

icu4c/source/data/unit/en.txt
```diff
@@ -422,6 +427,11 @@ en{
             }
         }
         force{
+            kilowatt-hour-per-100-kilometer{
```
markusicu, line 430: again, consumption not force

markusicu (changes requested):

icu4c/source/data/misc/supplementalData.txt
sffc: By definition, energy over distance equals force.  So the "real" category should be force.  But the "fake" category could still be consuption if that makes more sense.

markusicu: kWh/100km should go together with
```
                "consumption-liter-per-100-kilometer",
                "consumption-liter-per-kilometer",
                "consumption-mile-per-gallon",
                "consumption-mile-per-gallon-imperial",
```

macchiati:
The prefix in the long cldr ID no longer has much meaning, and there is a
ticket to remove it (thus ending up just with the short cldrID.

It has been replaced at a lower level by the SI quantity.  And the quantity
for kWh/100km is force.

cc @Shane Carr ***@***.***>

macchiati:
In particular, the old prefixes are rubbish for conversion, since they
group unconvertable units together, and separate convertible units.

Mark

icu4c/source/data/misc/supplementalData.txt
markusicu: https://unicode-org.atlassian.net/browse/CLDR-14837

PR #1743, opened 2021-06-16
ICU-21641 Fix "emoji" and "eor" for Collator::getKeywordValues

FrankYFTang:
icu::Collator::getKeywordValues("collation", status) does not report "emoji" and "eor"

Use res_open intead of ures_openDirect  inside ures_getKeywordValues so it access the fallback resource in "root". This make it align to Java and will also count "emoji" and "eor" which is specified in "root" 
Also add test for both C++ and Java
Java behavior correctly without any fix.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21641

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -2954,7 +2954,7 @@ ures_getKeywordValues(const char *path, const char *keyword, UErrorCode *status)
         UResourceBundle   *bund = NULL;
         UResourceBundle   *subPtr = NULL;
         UErrorCode subStatus = U_ZERO_ERROR; /* don't fail if a bundle is unopenable */
-        bund = ures_openDirect(path, locale, &subStatus);
+        bund = ures_open(path, locale, &subStatus);
         
 #if defined(URES_TREE_DEBUG)
         if(!bund || U_FAILURE(subStatus)) {
--- a/icu4c/source/test/intltest/svccoll.cpp
+++ b/icu4c/source/test/intltest/svccoll.cpp
@@ -560,7 +560,9 @@ static const int32_t KW_COUNT = UPRV_LENGTHOF(KW);
 
 static const char* KWVAL[] = {
     "phonebook",
-    "stroke"
+    "stroke",
+    "emoji",
+    "eor"
 };
 static const int32_t KWVAL_COUNT = UPRV_LENGTHOF(KWVAL);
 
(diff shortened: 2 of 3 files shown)


icu4c/source/common/uresbund.cpp
```diff
@@ -2950,7 +2950,15 @@ ures_getKeywordValues(const char *path, const char *keyword, UErrorCode *status)
     valuesBuf[0]=0;
     valuesBuf[1]=0;
     
-    while((locale = uenum_next(locs, &locLen, status)) != 0) {
+    // res_index.txt does not include "root" but we should also include the
+    // keyword values which only coded in "root" to the result, such as "emoji"
+    // and "eor" for "coll".
+    bool addRoot = true;
```
roubert, line 2956: I think that a simpler way of doing this would be to instead rewrite the loop like this:

```C++
    static constexpr char ROOT[] = "root";
    do {
        locale = uenum_next(locs, &locLen, status);
        if (locale == NULL) locale = ROOT;

        // ...
        
    } while (locale != ROOT);
```

That avoids calling `uenum_next()` one superfluous time after it already has reached the end and I think it will also be more clear to the human reader what the loop does.

FrankYFTang: I found a better fix

FrankYFTang: sorry, after playing with ICU4J I realize my attempted fix was wrong. Instead, I should just use res_open intead of res_openDirect and the loop will look into the fallback resources inside "root" and do not need to look into one more resource after the loop. PTAL. Noitce the only calling ures_getKeywordValues is inside 
ucol_getKeywordValues

which is where we intend to fix and the only internal place which call that function is Collator::getKeywordValues. so there are no risk to change this code for other  API.

roubert (dismissed):
LGTM

FrankYFTang:
PTAL

roubert (approved):

icu4c/source/common/uresbund.cpp
markusicu: I suspect that it used to call ures_openDirect() as an optimization: ures_open() also opens all parent bundles down to root, but since the code here visits all locales, including the parent bundles, it will see their contents even with ures_openDirect(). Except if the "all locales" skips root, and if parent bundles for some reason are not listed as "available". So this change will make the operation slower, but also more reliably correct.

PR #1752, opened 2021-06-28
Unicode Transforms Selection

MasterInQuestion:
```Text
= Unicode Transforms Selection =

[
|*| Source: https://github.com/unicode-org/icu/pull/1752#issue-679367941
|*| Last update: CE 2021-06-28 20:20 UTC ]


Various selected (and amended) info about Unicode Transforms (Unicode-style RegEx?) from the ICU (International Components for Unicode) Documentation.


Off-Topic: Why not to update the corresponding ".md" files?.. Too reluctant to bother that clumsy format.


----


[ General Transforms # Rule-Based Transliterators # Revisiting (base: commit 5a714ae8f71f8b302ccc9604c4b192817b239b27 (CE 2020-08-19 18:49:35 UTC) ):
https://unicode-org.github.io/icu/userguide/transforms/general/#revisiting

==== Revisiting ====

If the resulting text contains a "|" (Vertical Line) (U+007C, 0x7C, &#124;): it indicates the following processing shall proceed from that very point; and the transform shall revisit part of the resulting text.


For example, given:
[
	'x' > 'y' | 'z';
	'za' > 'w';
]
, the input string "xa" should convert to "yw":

|1| First, "xa" should be converted to "yza".
|2| Then the processing should continue right after the character "y": pick up the "za", and convert it.

Had there not have the "|", the result would simply have been "yza".


Compare:
[
	'x' > 'yz'; # [1]
	^'yz' {'a'$ > 'w'; # [2]
#
# Which would output:
# |*| "yzw", with the very input string "xa"; or
# |*| "yzaa", with the slightly modified input string "xaa".
#
#
#
#
# == Notes ==
#
# [1]
# [
# Semantic equivalent of:
# [
# 	'x' > 'y' 'z';
# ]
# ,
# [
# 	x > y z;
# ]
# ,
# [
# 	x>
# 	yz
#
# 	;
# ]
# , and so on.
#
# For more details, check: # Rule-Based Transliterators. ]
#
#
# [2]
# [
# Pattern:
# [
# ^'yz' {'a'$
# ]
# ; approximately (ignoring the implementation dependent trailing line-break handling difference) the /(?<=^yz)a$/g in Perl-style RegEx. ("a", at the end of input; preceded by: "yz", at the start of input)
#
# For more details, check:
# |*| ## Conversion Rules, Regular Expression
# |*| Transform Rule Tutorial ## Repetition, Æther, Copying: https://unicode-org.github.io/icu/userguide/transforms/general/rules.html#repetition ]
#
] ]


----


[ General Transforms # Regular Expression (base: commit 5a714ae8f71f8b302ccc9604c4b192817b239b27 (CE 2020-08-19 18:49:35 UTC) ):
|*| https://unicode-org.github.io/icu/userguide/transforms/general/#regular-expression
|*| http://userguide.icu-project.org/transforms/general#TOC-Regular-Expression

=== Regular Expression ===

The rules are similar to regular expressions in offering:
|*| Variables,
|*| Property matches ("[:...:]", "\p{...}", "\P{...}"), [ Check UnicodeSet # UnicodeSet Patterns (https://unicode-org.github.io/icu/userguide/strings/unicodeset.html#unicodeset-patterns) for details. ]
|*| Contextual matches (a.k.a. "Lookaround", position assertion),
|*| Capturing and rearrangement ("(...)" and "$1", "$2"...), [ The syntax sugar (mostly) for explicit non-capturing group ("(?:...)") seems to be missing. ]
|*| Quantifiers ("*", "+", "?").

They are more powerful in offering:
|*| Ordered Rules, [ Explanation Needed ]
|*| Cursor Backup (Revisiting), [ Probably not when comparing to Perl (which allows the cursor position be directly manipulated (with some limitations) during RegEx operation). ]
|*| Buffered/Keyboard support.

They are less powerful in that they have:
|*| Inferior quantifier flexibility:
|**| Only greedy quantifiers, ("*?", "+?", "??" are not supported)
|**| Missing support for finer granularity quantifiers (e.g. "{1,2}", "{3,}", etc.);
|*| No backup (explicit "Alternation") (so no, for example:
[
	'aa' | 'abc' > 'X'; # Invalid
]
(intended Perl-style RegEx replace /aa|abc/X/g equivalent) [ Though the very semantic may be somewhat trivially reconstructed with:
[
	'aa' > 'X';
	'abc' > 'X';
]
; or for such simple cases, using some unusual property exclusive to the UnicodeSet:
[
	[{aa}{abc}] > 'X';
]
(check UnicodeSet # String Values in Sets (https://unicode-org.github.io/icu/userguide/strings/unicodeset.html#string-values-in-sets) for details) ]),
|*| No input-side back references.


Here is a simple example demonstrating the difference between a set of transform rules, and successively applying regular expression replacements:

''(To Be Done)'' [ Never? ]


Since the transform processes each of its rules at each point, it catches the "yx" before the "xy" in the second case.

Since each RegEx match/replace is supposed to evaluate on the full previously processed part: [ Or to give output compatible to such behavior. ] that very effect may not be so easily achieved. (would require more complicated processing like through the "/e" RegEx flag in Perl)

Simply using multiple regular expressions cannot account for the interaction and ordering of characters ''with the'' [ Original: and ] rules. [ One may, however, easily simulate the RegEx behavior with transform rules: by inserting a transform rule (e.g. `:: Null;`) to split the conversion rules into passes. ]


For more details on constructing rules, see the Transform Rule Tutorial (https://unicode-org.github.io/icu/userguide/transforms/general/rules.html). ]
```


richgillam:
We can't figure out what you're trying to do here.  It looks like you're proposing changes to the user guide.  If that's what you want to do, you should make changes to the actual user guide files and then post THAT as a PR, rather than putting the text and proposed changes into the PR description field.  (You also need to sign the CLA and file a JIRA ticket.)

PR #1762, opened 2021-07-08
ICU-21639 Changed several internal methods in ULocale to work correct…

richgillam:
…ly on locale IDs that are longer than ULOC_FULLNAME_CAPACITY.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21639

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -1181,13 +1181,13 @@ error:
     }
 }
 
-static UBool
+static int32_t
 do_canonicalize(const char*    localeID,
          char* buffer,
          int32_t bufferCapacity,
          UErrorCode* err)
 {
-    uloc_canonicalize(
+    int32_t canonicalizedSize = uloc_canonicalize(
         localeID,
         buffer,
         bufferCapacity,
@@ -1195,16 +1195,14 @@ do_canonicalize(const char*    localeID,
 
     if (*err == U_STRING_NOT_TERMINATED_WARNING ||
         *err == U_BUFFER_OVERFLOW_ERROR) {
-        *err = U_ILLEGAL_ARGUMENT_ERROR;
-
-        return FALSE;
+        return canonicalizedSize;
     }
     else if (U_FAILURE(*err)) {
 
-        return FALSE;
+        return -1;
     }
     else {
-        return TRUE;
+        return canonicalizedSize;
     }
 }
 
@@ -1241,12 +1239,17 @@ static UBool
 _ulocimp_addLikelySubtags(const char* localeID,
                           icu::ByteSink& sink,
                           UErrorCode* status) {
-    char localeBuffer[ULOC_FULLNAME_CAPACITY];
-
-    if (do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status)) {
-        return _uloc_addLikelySubtags(localeBuffer, sink, status);
+    PreflightingLocaleIDBuffer localeBuffer;
+    do {
+        localeBuffer.requestedCapacity = do_canonicalize(localeID, localeBuffer.getBuffer(),
+            localeBuffer.getCapacity(), status);
+    } while (localeBuffer.needToTryAgain(status));
+    
+    if (U_SUCCESS(*status)) {
+        return _uloc_addLikelySubtags(localeBuffer.getBuffer(), sink, status);
+    } else {
+        return FALSE;
     }
-    return FALSE;
 }
 
 U_CAPI void U_EXPORT2
@@ -1289,11 +1292,13 @@ U_CAPI void U_EXPORT2
 ulocimp_minimizeSubtags(const char* localeID,
                         icu::ByteSink& sink,
                         UErrorCode* status) {
-    char localeBuffer[ULOC_FULLNAME_CAPACITY];
-
-    if (do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status)) {
-        _uloc_minimizeSubtags(localeBuffer, sink, status);
-    }
+    PreflightingLocaleIDBuffer localeBuffer;
+    do {
+        localeBuffer.requestedCapacity = do_canonicalize(localeID, localeBuffer.getBuffer(),
+            localeBuffer.getCapacity(), status);
+    } while (localeBuffer.needToTryAgain(status));
+    
+    _uloc_minimizeSubtags(localeBuffer.getBuffer(), sink, status);
 }
 
 // Pairs of (language subtag, + or -) for finding out fast if common languages
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -478,15 +478,19 @@ static const CanonicalizationMap CANONICALIZE_MAP[] = {
 /* Test if the locale id has BCP47 u extension and does not have '@' */
 #define _hasBCP47Extension(id) (id && uprv_strstr(id, "@") == NULL && getShortestSubtagLength(localeID) == 1)
 /* Converts the BCP47 id to Unicode id. Does nothing to id if conversion fails */
-#define _ConvertBCP47(finalID, id, buffer, length,err) UPRV_BLOCK_MACRO_BEGIN { \
-    if (uloc_forLanguageTag(id, buffer, length, NULL, err) <= 0 || \
-            U_FAILURE(*err) || *err == U_STRING_NOT_TERMINATED_WARNING) { \
-        finalID=id; \
-        if (*err == U_STRING_NOT_TERMINATED_WARNING) { *err = U_BUFFER_OVERFLOW_ERROR; } \
-    } else { \
-        finalID=buffer; \
-    } \
-} UPRV_BLOCK_MACRO_END
+static int32_t _ConvertBCP47(
+            const char*& finalID, const char* id, char* buffer, int32_t length, UErrorCode* err) {
+    int32_t localeIDSize = uloc_forLanguageTag(id, buffer, length, NULL, err);
+    if (localeIDSize <= 0 || U_FAILURE(*err) || *err == U_STRING_NOT_TERMINATED_WARNING) {
+        finalID=id;
+        if (*err == U_STRING_NOT_TERMINATED_WARNING) {
+            *err = U_BUFFER_OVERFLOW_ERROR;
+        }
+    } else {
+        finalID=buffer;
+    }
+    return localeIDSize;
+}
 /* Gets the size of the shortest subtag in the given localeID. */
 static int32_t getShortestSubtagLength(const char *localeID) {
     int32_t localeIDLength = static_cast<int32_t>(uprv_strlen(localeID));
@@ -1474,7 +1478,7 @@ _canonicalize(const char* localeID,
               uint32_t options,
               UErrorCode* err) {
     int32_t j, fieldCount=0, scriptSize=0, variantSize=0;
-    char tempBuffer[ULOC_FULLNAME_CAPACITY];
+    PreflightingLocaleIDBuffer tempBuffer;
     const char* origLocaleID;
     const char* tmpLocaleID;
     const char* keywordAssign = NULL;
@@ -1485,7 +1489,10 @@ _canonicalize(const char* localeID,
     }
 
     if (_hasBCP47Extension(localeID)) {
-        _ConvertBCP47(tmpLocaleID, localeID, tempBuffer, sizeof(tempBuffer), err);
+        do {
+            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeID,
+                tempBuffer.getBuffer(), tempBuffer.getCapacity(), err);
+        } while (tempBuffer.needToTryAgain(err));
     } else {
         if (localeID==NULL) {
            localeID=uloc_getDefault();
--- a/icu4c/source/common/ulocimp.h
+++ b/icu4c/source/common/ulocimp.h
@@ -307,4 +307,72 @@ U_CAPI const char* const* ulocimp_getKnownCanonicalizedLocaleForTest(int32_t* le
 // Return true if the value is already canonicalized.
 U_CAPI bool ulocimp_isCanonicalizedLocaleForTest(const char* localeName);
 
+/**
+ * A utility class for handling locale IDs that may be longer than ULOC_FULLNAME_CAPACITY.
+ * This encompasses all of the logic to allocate a temporary locale ID buffer on the stack,
+ * and then, if it's not big enough, reallocate it on the heap and try again.
+ *
+ * You use it like this:
+ * UErrorCode err = U_ZERO_ERROR;
+ *
+ * PreflightingLocaleIDBuffer tempBuffer;
+ * do {
+ *     tempBuffer.requestedCapacity = uloc_doSomething(localeID, tempBuffer.getBuffer(), tempBuffer.getCapacity(), &err);
+ * } while (tempBuffer.needToTryAgain(&err));
+ * if (U_SUCCESS(err)) {
+ *     uloc_doSomethingWithTheResult(tempBuffer.getBuffer());
+ * }
+ */
+class PreflightingLocaleIDBuffer {
+private:
+    char stackBuffer[ULOC_FULLNAME_CAPACITY];
+    char* heapBuffer = nullptr;
+    int32_t capacity = ULOC_FULLNAME_CAPACITY;
+    
+public:
+    int32_t requestedCapacity = ULOC_FULLNAME_CAPACITY;
+
+    // No heap allocation. Use only on the stack.
+    static void* U_EXPORT2 operator new(size_t) U_NOEXCEPT = delete;
+    static void* U_EXPORT2 operator new[](size_t) U_NOEXCEPT = delete;
+#if U_HAVE_PLACEMENT_NEW
+    static void* U_EXPORT2 operator new(size_t, void*) U_NOEXCEPT = delete;
+#endif
+
+    PreflightingLocaleIDBuffer() {}
+    
+    ~PreflightingLocaleIDBuffer() { uprv_free(heapBuffer); }
+    
+    char* getBuffer() {
+        if (heapBuffer == nullptr) {
+            return stackBuffer;
+        } else {
+            return heapBuffer;
+        }
+    }
+    
+    int32_t getCapacity() {
+        return capacity;
+    }
+    
+    bool needToTryAgain(UErrorCode* err) {
+        if (heapBuffer != nullptr) {
+            return false;
+        }
+    
+        if (*err == U_BUFFER_OVERFLOW_ERROR || *err == U_STRING_NOT_TERMINATED_WARNING) {
+            int32_t newCapacity = requestedCapacity + 2;    // one for the terminating null, one just for paranoia
+            heapBuffer = static_cast<char*>(uprv_malloc(newCapacity));
+            if (heapBuffer == nullptr) {
+                *err = U_MEMORY_ALLOCATION_ERROR;
+            } else {
+                *err = U_ZERO_ERROR;
+                capacity = newCapacity;
+            }
+            return U_SUCCESS(*err);
+        }
+        return false;
+    }
+};
+
 #endif
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -58,6 +58,7 @@ static void TestBug20370(void);
 static void TestBug20321UnicodeLocaleKey(void);
 
 static void TestUsingDefaultWarning(void);
+static void TestExcessivelyLongIDs(void);
 
 void PrintDataTable();
 
@@ -281,6 +282,7 @@ void addLocaleTest(TestNode** root)
     TESTCASE(TestBug20321UnicodeLocaleKey);
     TESTCASE(TestUsingDefaultWarning);
     TESTCASE(TestBug21449InfiniteLoop);
+    TESTCASE(TestExcessivelyLongIDs);
 }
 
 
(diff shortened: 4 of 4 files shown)


icu4c/source/test/cintltst/cloctst.c
```diff
@@ -7009,3 +7011,35 @@ static void TestBug21449InfiniteLoop() {
...
+static void TestExcessivelyLongIDs(void) {
+    const char* reallyLongID = "de-u-cu-eur-em-default-hc-h23-ks-level1-lb-strict-lw-normal-ms-metric-nu-latn-rg-atzzzz-sd-atat1-ss-none-tz-atvie-va-posix";
+    char minimizedID[ULOC_FULLNAME_CAPACITY];
+    char maximizedID[ULOC_FULLNAME_CAPACITY];
+    int32_t actualMinimizedLength = 0;
+    int32_t actualMaximizedLength = 0;
+    UErrorCode err = U_ZERO_ERROR;
+    
+    actualMinimizedLength = uloc_minimizeSubtags(reallyLongID, minimizedID, ULOC_FULLNAME_CAPACITY, &err);
+    assertTrue("uloc_minimizeSubtags() with too-small buffer didn't fail as expected", U_FAILURE(err) && actualMinimizedLength > ULOC_FULLNAME_CAPACITY);
```
FrankYFTang, line 7025: could you wrap this long assert (and below) around 80 cols, or at least < 100 cols so we can view it easily.

icu4c/source/common/uloc.cpp
```diff
@@ -478,15 +478,18 @@ static const CanonicalizationMap CANONICALIZE_MAP[] = {
...
-#define _ConvertBCP47(finalID, id, buffer, length,err) UPRV_BLOCK_MACRO_BEGIN { \
-    if (uloc_forLanguageTag(id, buffer, length, NULL, err) <= 0 || \
-            U_FAILURE(*err) || *err == U_STRING_NOT_TERMINATED_WARNING) { \
-        finalID=id; \
-        if (*err == U_STRING_NOT_TERMINATED_WARNING) { *err = U_BUFFER_OVERFLOW_ERROR; } \
-    } else { \
-        finalID=buffer; \
-    } \
-} UPRV_BLOCK_MACRO_END
+static int32_t _ConvertBCP47(const char*& finalID, const char* id, char* buffer, int32_t length, UErrorCode* err) {
```
FrankYFTang, line 481: could you wrap line shorter? thanks

FrankYFTang (dismissed):
thanks for working on this. Please wrap some of the very long line shorter. I am not sure what is the sytle for C++ in ICU but I think some of your new line is just way too long. Maybe wrap to 80? (or 100?)

icu4c/source/common/loclikely.cpp
```diff
@@ -1243,8 +1241,16 @@ _ulocimp_addLikelySubtags(const char* localeID,
                           UErrorCode* status) {
     char localeBuffer[ULOC_FULLNAME_CAPACITY];
 
-    if (do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status)) {
-        return _uloc_addLikelySubtags(localeBuffer, sink, status);
+    int32_t canonicalizedSize = do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status);
+    if (*status == U_BUFFER_OVERFLOW_ERROR || *status == U_STRING_NOT_TERMINATED_WARNING) {
+        icu::LocalMemory<char> heapLocaleBuffer((char*)uprv_malloc(canonicalizedSize + 2));
```
jefgen, line 1246: Hmm... if `uprv_malloc` fails to allocate memory (due to OOM), then I'm not sure if `status` will be set to `U_MEMORY_ALLOCATION_ERROR`.
It doesn't look like either `do_canonicalize` or `uloc_canonicalize` check that the input buffer is non-NULL, so it might need to be checked here beforehand (with `heapLocaleBuffer.isNull()`) and the `status` set.
(Also, same with the lines 1302-1304 below as well).

icu4c/source/common/uloc.cpp
```diff
@@ -1485,7 +1490,12 @@ _canonicalize(const char* localeID,
     }
 
     if (_hasBCP47Extension(localeID)) {
-        _ConvertBCP47(tmpLocaleID, localeID, tempBuffer, sizeof(tempBuffer), err);
+        int32_t convertedSize = _ConvertBCP47(tmpLocaleID, localeID, tempBuffer, sizeof(tempBuffer), err);
+        if (*err == U_BUFFER_OVERFLOW_ERROR || *err == U_STRING_NOT_TERMINATED_WARNING) {
+            heapTempBuffer.adoptInstead((char*)uprv_malloc(convertedSize + 2));
```
jefgen, line 1495: I think we also need to check if `uprv_malloc` fails here, as `_ConvertBCP47` won't set `err` to `U_MEMORY_ALLOCATION_ERROR` if the buffer is NULL/nullptr.

icu4c/source/common/loclikely.cpp
jefgen: nit: maybe consider a C++ style cast since this file a .cpp? (Though to be fair, the rest of the file doesn't use them, so perhaps moot).

FrankYFTang: agree with jeff about checking the null return of malloc, we surely should check that and set to U_MEMORY_ALLOCATION_ERROR if the return of malloc is nullptr

icu4c/source/common/loclikely.cpp
```diff
@@ -1291,7 +1297,15 @@ ulocimp_minimizeSubtags(const char* localeID,
                         UErrorCode* status) {
     char localeBuffer[ULOC_FULLNAME_CAPACITY];
 
-    if (do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status)) {
+    int32_t canonicalizedSize = do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status);
+    if (*status == U_BUFFER_OVERFLOW_ERROR || *status == U_STRING_NOT_TERMINATED_WARNING) {
+        icu::LocalMemory<char> heapLocaleBuffer((char*)uprv_malloc(canonicalizedSize + 2));
```
FrankYFTang, line 1302: same as above, check the nullptr of malloc and set U_MEMORY_ALLOCATION_ERROR if nullptr

icu4c/source/common/uloc.cpp
FrankYFTang: agree with jeff

FrankYFTang (changes requested):

icu4c/source/common/loclikely.cpp
richgillam: Good catch(es).  Done.

It makes for pretty ugly code.  I thought about going to the extra trouble of factoring all this logic out into some kind of utility object, but it didn't seem worth the trouble for just these two (well, three) call sites.  I could do that if you guys insisted, but I can't help thinking it'd be better to wait until we see we need this kind of thing in more places.

icu4c/source/common/uloc.cpp
richgillam: Done.  See comments above.

icu4c/source/common/loclikely.cpp
richgillam: Yes, I caught this one too.

icu4c/source/common/loclikely.cpp
```diff
@@ -1243,11 +1243,16 @@ _ulocimp_addLikelySubtags(const char* localeID,
...
-        icu::LocalMemory<char> heapLocaleBuffer((char*)uprv_malloc(canonicalizedSize + 2));
-        *status = U_ZERO_ERROR;
-        do_canonicalize(localeID, heapLocaleBuffer.getAlias(), canonicalizedSize + 2, status);
-        if (U_SUCCESS(*status)) {
-            return _uloc_addLikelySubtags(heapLocaleBuffer.getAlias(), sink, status);
+        icu::LocalMemory<char> heapLocaleBuffer(static_cast<char*>(uprv_malloc(canonicalizedSize + 2)));
+        if (heapLocaleBuffer.isNull()) {
+            *status = U_MEMORY_ALLOCATION_ERROR;
+            return FALSE;
+        } else {
```
FrankYFTang, line 1250: remove the else { } since the last line in the if already return.

icu4c/source/common/loclikely.cpp
```diff
@@ -1299,11 +1304,16 @@ ulocimp_minimizeSubtags(const char* localeID,
...
-        icu::LocalMemory<char> heapLocaleBuffer((char*)uprv_malloc(canonicalizedSize + 2));
-        *status = U_ZERO_ERROR;
-        do_canonicalize(localeID, heapLocaleBuffer.getAlias(), canonicalizedSize + 2, status);
-        if (U_SUCCESS(*status)) {
-            _uloc_minimizeSubtags(heapLocaleBuffer.getAlias(), sink, status);
+        icu::LocalMemory<char> heapLocaleBuffer(static_cast<char*>(uprv_malloc(canonicalizedSize + 2)));
+        if (heapLocaleBuffer.isNull()) {
+            *status = U_MEMORY_ALLOCATION_ERROR;
+            return;
+        } else {
```
FrankYFTang, line 1311: ditto

icu4c/source/common/uloc.cpp
```diff
@@ -1493,8 +1493,13 @@ _canonicalize(const char* localeID,
         int32_t convertedSize = _ConvertBCP47(tmpLocaleID, localeID, tempBuffer, sizeof(tempBuffer), err);
         if (*err == U_BUFFER_OVERFLOW_ERROR || *err == U_STRING_NOT_TERMINATED_WARNING) {
             heapTempBuffer.adoptInstead((char*)uprv_malloc(convertedSize + 2));
-            *err = U_ZERO_ERROR;
-            _ConvertBCP47(tmpLocaleID, localeID, heapTempBuffer.getAlias(), convertedSize + 2, err);
+            if (heapTempBuffer.isNull()) {
+                *err = U_MEMORY_ALLOCATION_ERROR;
+                return;
+            } else{
```
FrankYFTang, line 1499: ditto

FrankYFTang (changes requested):

icu4c/source/common/loclikely.cpp
richgillam: I disagree.  It's a stylistic issue.  You're right that the `else` isn't strictly necessary, but I think it clarifies the intent for anybody reading this code.  **IF** we got back `nullptr` from the `uprv_malloc()` call, just return `U_MEMORY_ALLOCATION_ERROR`.  **ELSE** go ahead and call `doCanonicalize()` (or whatever).

I'll change this if you insist, but I really do feel it's a step backward in readability.

FrankYFTang: first of all, I agree w/ you this is a stylistic issue but look at the code here and how many level of the if we have. I actually prefer we rewrite this part of if and completely remove all else. For example, rewrite to the following so there are no nested if / else but only one level deep if block, I think the following code is much easier to understand and much better in readability since human brain have no need to hang on what is the IF and what is IF for the Else when looking at the 2nd IF nested inside. 
```
int32_t canonicalizedSize = do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status);
if (U_SUCCESS(*status)) {
       return _uloc_addLikelySubtags(localeBuffer, sink, status);
}
if (*status != U_BUFFER_OVERFLOW_ERROR && *status != U_STRING_NOT_TERMINATED_WARNING) {
      return FALSE;
}
// localeID is too long by using localeBuffer so we allocate longer buffer in heapLocaleBuffer and try it again.
icu::LocalMemory<char> heapLocaleBuffer(static_cast<char*>(uprv_malloc(canonicalizedSize + 2)));
 if (heapLocaleBuffer.isNull()) {
     return FALSE;
}
*status = U_ZERO_ERROR;
do_canonicalize(localeID, heapLocaleBuffer.getAlias(), canonicalizedSize + 2, status);
if (U_FAILURE(*status)) {
   return FALSE;
}
return _uloc_addLikelySubtags(heapLocaleBuffer.getAlias(), sink, status);
```

Can we change all the code around this to make sure there are no nested if/else?

FrankYFTang (changes requested):

icu4c/source/common/loclikely.cpp
FrankYFTang: This article explain about the Bubble style vs gateway style and show you why such "style" is consider harder to read . The example is in Javascript not C++ but I believe the principle is the same. https://wpshout.com/unconditionally-refactoring-nested-statements-cleaner-code/ In theory, for any if or else block which the last statememt is return (or throw) it can be refactor to the "gateway style" to make it easier for human to track because with the gateway style, the reader only need to remember ONE thing at a time and have no need for "back tracking", but in the Bubble style, human reader need to "back track" multiple condiction and while the function grow longer LATER, it is much harder to "back track"

FrankYFTang (changes requested):

icu4c/source/common/loclikely.cpp
jefgen: Thanks!

Hmm.. I don't know if it is worth factoring it out into a separate helper for now. This seems fine for the 2/3 call sites.

But thinking about it some more:
It makes it wonder if maybe we'd want something like `LocalPointer`'s [`adoptInsteadAndCheckErrorCode`](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/localpointer.h#L319), but for the `LocalMemory` class.

`LocalMemory` already has [`adoptInstead`](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/cmemory.h#L243)  -- so I think adding a `adoptInsteadAndCheckErrorCode` wouldn't be out of place for the class... (But that probably wants to be with a different ticket for discussion...)

icu4c/source/common/loclikely.cpp
jefgen: I'd agree that it's a stylistic issue as well.
I hadn't heard the terms "Bubble style" vs "Gateway style" before -- but I've seen it called "early return" pattern versus the "arrow" or "triangle" pattern of nested if statements.
FWIW, I'd lean towards having less nested ifs where it makes things clearer.
Though for a small function like this I wouldn't block the PR. I'd leave it up to you(s) as to the style. 🙂

However, it might be worth noting that there's already plenty of ICU code that uses the pattern of "early return" without nested ifs. For example, most functions with an error parameter start off with this:

```c++
if (U_FAILURE(*status)) {
    return FALSE;
}
```

However, one thing to note -- we can't write it exactly as-is in the example above though:
```c++
int32_t canonicalizedSize = do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status);
if (U_SUCCESS(*status)) {
    return _uloc_addLikelySubtags(localeBuffer, sink, status);
}
if (*status != U_BUFFER_OVERFLOW_ERROR && *status != U_STRING_NOT_TERMINATED_WARNING) {
    return FALSE;
}
```
Because `U_STRING_NOT_TERMINATED_WARNING` is a *warning*, not an *error*, so it will be considered as `U_SUCCESS == true`, and we won't return `FALSE`. The check for buffer overflow / not terminated needs to happen *before* the check for `U_SUCCESS`.

icu4c/source/common/uloc.cpp
jefgen: nit: You added a `static_cast<char*>` in the other file, but not here. ;)

icu4c/source/common/loclikely.cpp
richgillam: After thinking it over, I decided the best way to address both your and Frank's concerns _was_ to add a helper class.  I hope you agree.  I personally think It makes the code look a lot simpler.  I'm not sure ulocimp.h was the right place to put the helper class; if you all have a better idea, let me know.

icu4c/source/common/loclikely.cpp
```diff
@@ -1241,12 +1239,16 @@ static UBool
 _ulocimp_addLikelySubtags(const char* localeID,
                           icu::ByteSink& sink,
                           UErrorCode* status) {
-    char localeBuffer[ULOC_FULLNAME_CAPACITY];
-
-    if (do_canonicalize(localeID, localeBuffer, sizeof localeBuffer, status)) {
-        return _uloc_addLikelySubtags(localeBuffer, sink, status);
+    PreflightingLocaleIDBuffer localeBuffer;
+    do {
+        *(localeBuffer.getCapacityPtr()) = do_canonicalize(localeID, localeBuffer.getBuffer(), localeBuffer.getCapacity(), status);
```
jefgen, line 1244: Re: `*(localeBuffer.getCapacityPtr()) = `

FWIW, this seems a bit odd (at least to me) -- assigning the resulting size to the "capacity".

Ex: For the stack-allocated case where the data fits, it would mean that the `PreflightingLocaleIDBuffer` now thinks the capacity is *less* than the actual buffer length (`ULOC_FULLNAME_CAPACITY`).

It would also mean that the new capacity wouldn't include the null terminator, as the resulting size reported by `uloc_canonicalize` doesn't include the terminating null in the size. 
So for something like "en-US", it would think the capacity was only 5.

On the other hand, when the data *doesn't fit* then the capacity would be the same as the data size + 2.

Maybe I'm just getting confused by the term "capacity" though?
I wonder if perhaps calling it "size" instead of "capacity" would be less confusing?

icu4c/source/common/ulocimp.h
```diff
@@ -307,4 +307,62 @@ U_CAPI const char* const* ulocimp_getKnownCanonicalizedLocaleForTest(int32_t* le
...
+class PreflightingLocaleIDBuffer {
+private:
+    char stackBuffer[ULOC_FULLNAME_CAPACITY];
+    char* heapBuffer = nullptr;
+    int32_t capacity = ULOC_FULLNAME_CAPACITY;
+    
+public:
+    PreflightingLocaleIDBuffer() {}
+    
+    ~PreflightingLocaleIDBuffer() { delete [] heapBuffer; }
```
jefgen, line 335: Hm, I think this needs to also derive from the `UMemory` class in order to override the `new`/`delete` operators.
(Or you could use `uprv_malloc`/`uprv_free` directly.)

richgillam: Why do I need to override the new/delete operators?  What am I missing here?

jefgen: The ICU library code allows for overriding/changing the memory allocation functions, allowing the library code to possibly use a different heap than the application (or whatever is calling ICU).

There's a macro for controlling this called `U_OVERRIDE_CXX_ALLOCATION`, which has some comments about it:
https://github.com/unicode-org/icu/blob/2dce62892b6b00df02195a35ca09ebd8e082e80f/icu4c/source/common/unicode/uconfig.h#L114-L123

Also from the ICU user guide docs, there is this part here:
(From https://unicode-org.github.io/icu/userguide/dev/codingguidelines.html#c-class-layout-and-boilerplate )
> In general, all instantiable classes (i.e., all classes except for interface/mixin classes and ones with only static member functions) inherit the UMemory base class. UMemory provides new/delete operators, which allows to keep the ICU heap separate from the application heap, or to customize ICU’s memory allocation consistently.

icu4c/source/common/loclikely.cpp
richgillam: I decided to split it into two variables: `capacity`, which gives the current capacity of the buffer, and `requestedCapacity`, which is the capacity that the call we're wrapping says it needs.  I agree this is clearer, but it comes at the expense of using two variables instead of one, since the real capacity and the requested capacity are never used at the same time if you use this class the way it was intended to be used.  I could just change the name of that setter function to make it clearer what the caller is doing and leave things with a single internal variable (maybe under a new name), but the extra four (or eight) bytes aren't that big a deal and it _is_ clearer this way, I think...

By the way, I acknowledge that `*(buffer.getRequestedCapacityPtr()) = uloc_doSomething(...);` looks a little weird.  What I was trying to do was stick as close as I could to the way you'd normally use a function like this: `bytesNeeded = uloc_doSomething(...);`.  I was concerned that using a regular setter function would make it harder to see what the calling code was actually doing: `buffer.setRequestedCapacity(uloc_doSomething(...));` kind of buries the `uloc_doSomething(...)` part, and that's really where the reader's focus should be.  If you guys think what I'm doing actually hurts readability more than a normal setter function (or have some other better idea), let me have it.

icu4c/source/common/ulocimp.h
richgillam: (sigh)  Okay, I changed it to use `uprv_malloc()` and `uprv_free()`.  I also added code to prevent heap allocation of the new class.  I'm hoping one or the other of these changes will also take care of the couple of build failures I'm seeing right now.

jefgen: Thanks! :)

icu4c/source/common/loclikely.cpp
jefgen: Thanks for explaining about the `*(buffer.getRequestedCapacityPtr()) = uloc_doSomething(...);` style.
FWIW, I think that the approach with the `do... while` loop helps to make it more clear, as it puts the focus on the method being called, rather than the logic of allocating the buffer, etc..

Regarding splitting the variables, I think that helps to make it more clear, but maybe Frank/others might feel otherwise? I'm not sure what their thoughts might be.

Though one comment/issue: with them split like this now, I wonder if we might end up with odd cases in error situations...

In `needToTryAgain`, if the memory allocation fails, since we update the value of `capacity` first, we can have a case where  `capacity` gets set to a larger value, but the `heapBuffer` isn't allocated. I'm not sure how much it matters though, as the `needToTryAgain` will return FALSE, but it might be curious to have some `PreflightingLocaleIDBuffer` objects with both `requestedCapacity` and `capacity` larger than `ULOC_FULLNAME_CAPACITY`, yet no actual heap buffer.

I wonder if we might want to do something like this perhaps?
```c++
    newCapacity = requestedCapacity + 2;    // one for the terminating null, one just for paranoia
    heapBuffer = static_cast<char*>(uprv_malloc(newCapacity));
    if (heapBuffer == nullptr) {
        *err = U_MEMORY_ALLOCATION_ERROR;
    } else {
        *err = U_ZERO_ERROR;
        capacity = newCapacity;
    }
    return U_SUCCESS(*err);
```

richgillam: >I wonder if we might want to do something like this perhaps?

Sure...

icu4c/source/common/ulocimp.h
```diff
@@ -307,4 +307,75 @@ U_CAPI const char* const* ulocimp_getKnownCanonicalizedLocaleForTest(int32_t* le
...
+ * A utility class for handling locale IDs that may be longer than ULOC_FULLNAME_CAPACITY.
+ * This encompasses all of the logic to allocate a temporary locale ID buffer on the stack,
+ * and then, if it's not big enough, reallocate it on the heap and try again.
+ *
+ * You use it like this:
+ * UErrorCode err = U_ZERO_ERROR;
+ *
+ * PreflightingLocaleIDBuffer tempBuffer;
+ * do {
+ *     *(tempBuffer.getRequestedCapacityPtr()) = uloc_doSomething(localeID, tempBuffer.getBuffer(), tempBuffer.getCapacity(), &err);
```
FrankYFTang, line 320: why not just a void requestCapacity(int32_t newCapacity) ?

richgillam: Well, the original idea was to keep it looking as much like how you'd normally use one of these functions as possible:

```
newCapacity = uloc_doSomething(...);
```

My concern with using a setter...

```
tempBuffer.requestCapacity(uloc_doSomething(...));
```

...was that it obscured the important part of the line, the `uloc_doSomething()` call, making it harder to see what the calling function was doing and putting the emphasis on all the crap we're having to do to make sure we have an appropriately-sized character buffer (the complaint against my original fix for this problem).

That said, I'm not sure `*(tempBuffer.getRequestedCapacityPtr) =` is really doing that.  It's weird-looking and long, and both of those work against the goal of emphasizing the `uloc_doSomething()` part of the line.  `tempBuffer.requestedCapacity = ...` would be better, but I didn't like the idea of making an internal member public here.

The more I think about it, the more I think making `requestedCapacity` public and getting rid of that weird getter would be better, combined with some extra validity checking in `needToTryAgain()`.  What do you guys think?

jefgen: > making `requestedCapacity` public and getting rid of that weird getter would be better, ... What do you guys think?

FWIW, I'd be inclined to agree. I think that it might help to simplify things.

For example, instead of this:
`*(tempBuffer.getRequestedCapacityPtr()) = uloc_doSomething(localeID, tempBuffer.getBuffer(), tempBuffer.getCapacity(), &err);`
it might look something like this:
`tempBuffer.requestedCapacity = uloc_doSomething(localeID, tempBuffer.getBuffer(), tempBuffer.getCapacity(), &err);`
 
Since it's an internal helper class, making it public seems fine.

richgillam: Okay, I went ahead and made that change.  Thanks a lot!

If this looks good to you, I'll squash on your say-so while we're waiting for Frank.

jefgen: Ah, I hate to kept making nit comments on your PR.... but the example in the comment block in `ulocimp.h` still has the method `getRequestedCapacityPtr()`  :)

```c++
*(tempBuffer.getRequestedCapacityPtr()) = uloc_doSomething(localeID, tempBuffer.getBuffer(),
```

richgillam: Ugh.  Sorry about that.  Fixed now.

jefgen (approved):
Thanks @richgillam. FWIW, I think the changes seem fine with me, but it would be good to see if Frank has any more comments or thoughts.

FrankYFTang:
> Thanks @richgillam. FWIW, I think the changes seem fine with me, but it would be good to see if Frank has any more comments or thoughts.

I intended to propose a change to the comment to remove the Ptr but somehow just make it committed to the PR . sorry

FrankYFTang (approved):

richgillam:
>I intended to propose a change to the comment to remove the Ptr but somehow just make it committed to the PR . sorry

Good catch.  Thanks for fixing that.

FrankYFTang (approved):

PR #1763, opened 2021-07-08
ICU-21624 Fixed it so that a DecimalFormat no longer owns two separat…

richgillam:
…e DecimalFormatSymbols objects.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21624

### DIFF
--- a/icu4c/source/i18n/decimfmt.cpp
+++ b/icu4c/source/i18n/decimfmt.cpp
@@ -439,7 +439,7 @@ DecimalFormat::DecimalFormat(const DecimalFormat& source) : NumberFormat(source)
         return; // no way to report an error.
     }
     UErrorCode status = U_ZERO_ERROR;
-    fields->symbols.adoptInsteadAndCheckErrorCode(new DecimalFormatSymbols(*source.fields->symbols), status);
+    fields->symbols.adoptInsteadAndCheckErrorCode(new DecimalFormatSymbols(*source.getDecimalFormatSymbols()), status);
     // In order to simplify error handling logic in the various getters/setters/etc, we do not allow
     // any partially populated DecimalFormatFields object. We must have a fully complete fields object
     // or else we set it to nullptr.
@@ -463,7 +463,7 @@ DecimalFormat& DecimalFormat::operator=(const DecimalFormat& rhs) {
     fields->properties = rhs.fields->properties;
     fields->exportedProperties.clear();
     UErrorCode status = U_ZERO_ERROR;
-    LocalPointer<DecimalFormatSymbols> dfs(new DecimalFormatSymbols(*rhs.fields->symbols), status);
+    LocalPointer<DecimalFormatSymbols> dfs(new DecimalFormatSymbols(*rhs.getDecimalFormatSymbols()), status);
     if (U_FAILURE(status)) {
         // We failed to allocate DecimalFormatSymbols, release fields and its members.
         // We must have a fully complete fields object, we cannot have partially populated members.
@@ -507,7 +507,7 @@ UBool DecimalFormat::operator==(const Format& other) const {
     if (fields == nullptr || otherDF->fields == nullptr) {
         return false;
     }
-    return fields->properties == otherDF->fields->properties && *fields->symbols == *otherDF->fields->symbols;
+    return fields->properties == otherDF->fields->properties && *getDecimalFormatSymbols() == *otherDF->getDecimalFormatSymbols();
 }
 
 UnicodeString& DecimalFormat::format(double number, UnicodeString& appendTo, FieldPosition& pos) const {
@@ -794,7 +794,11 @@ const DecimalFormatSymbols* DecimalFormat::getDecimalFormatSymbols(void) const {
     if (fields == nullptr) {
         return nullptr;
     }
-    return fields->symbols.getAlias();
+    if (!fields->symbols.isNull()) {
+        return fields->symbols.getAlias();
+    } else {
+        return fields->formatter.getDecimalFormatSymbols();
+    }
 }
 
 void DecimalFormat::adoptDecimalFormatSymbols(DecimalFormatSymbols* symbolsToAdopt) {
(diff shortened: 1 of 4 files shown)


richgillam:
A couple notes:

1. This PR includes no new tests; it's a refactoring of existing code, and so I felt that verifying that the existing unit tests still pass with this change in place was sufficient (there's also no good way to test for this change, short of manually inspecting memory in a debugger).
2. The DecimalFormat class on the Java side does not appear to have been refactored to use the new NumberFormatter stuff as its implementation, so the problem only exists on the C++ side.

richgillam:
I was mistaken in my earlier statement that the DecimalFormat code in Java doesn't use the new NumberFormatter stuff; I was looking at the wrong version of DecimalFormat (the one in java.text-- I need to figure out how to use my new IDE).  However, now that I've inspected the _actual_ version of DecimalFormat in ICU4J, I still think it doesn't have this problem-- it looks like the DecimalFormat and its underlying LocalizedNumberFormatter object do indeed both have "symbols" variables, but if I'm reading the code correctly, they share the same instance instead of making a copy.

icu4c/source/i18n/decimfmt.cpp
```diff
@@ -1609,8 +1613,9 @@ void DecimalFormat::touch(UErrorCode& status) {
     }
 
     // In C++, fields->symbols is the source of truth for the locale.
```
jefgen, line 1615: Just curious, is the comment about `fields->symbols` being the source of truth still accurate?

richgillam: Not really.  I went back and tried to wordsmith this comment so that it's true again.

jefgen: Thanks!

icu4c/source/i18n/decimfmt.cpp
```diff
@@ -1619,9 +1625,10 @@ void DecimalFormat::touch(UErrorCode& status) {
     // Since memory has already been allocated for the formatter, we can move assign a stack-allocated object
     // and don't need to call new. (Which is slower and could possibly fail).
     fields->formatter = NumberPropertyMapper::create(
-        fields->properties, *fields->symbols, fields->warehouse, fields->exportedProperties, status
+        fields->properties, *symbols, fields->warehouse, fields->exportedProperties, status
```
sffc, line 1628: `*symbols` could refer to `fields->formatter.getDecimalFormatSymbols()` here. Is it safe to copy out of it while re-assigning `fields->formatter`?

richgillam: I see what you're saying.  I think you're right-- if the `symbols` object came from the underlying `LocalizedNumberFormatter`, it seems like it'd get deleted as part of the process of assigning over the `LocalizedNumberFormatter`.  Unless we know this situation can never happen, or the move assignment for `LocalizedNumberFormatter` is smart about avoiding this kind of thing.  What's surprising to me is that I'm not seeing any ASAN errors with this change in place when I run all the unit tests-- I'd think a problem like this would cause double-delete failures out the wazoo.

Do you have any thoughts on the best way to solve this, without adding extra allocations or copy operations?  Part of me thinks changing `LocalPointer::operator=()` to check for self-assignment would be one possibility.

sffc: How about if we change `NumberPropertyMapper::create` to take `LocalPointer<DecimalFormatSymbols>` instead of `const DecimalFormatSymbols&`, so that we take ownership rather than cloning?

richgillam: Okay, I think I've figured this out.

1) The code seems to be correct as written, at least when I follow it along in the debugger.  Here's what I see:

a) On line 1617, we call `DecimalFormat::getDecimalFormatSymbols()`.  If `fields->symbols` is null, this returns a pointer to the symbols object owned by `fields->formatter`.

b) On line 1627, we create a new formatter and move-assign it over the old one.  We first call `NumberPropertyMapper::create()`, passing it a reference to the symbols object we obtained in step a.  This calls `NumberPropertyMapper::oldToNew()`, which calls `SymbolsWrapper::setTo()`, which **clones** the symbols object from line 1617.

c) `NumberPropertyMapper::create()` eventually returns an `UnlocalizedNumberFormatter` that owns an entirely new `DecimalFormatSymbols` object that's a copy of the one from the old formatter.  `DecimalFormat::touch()` then calls `locale()` on that formatter, which creates a `LocalizedNumberFormatter`, which has been move-copied from the old `UnlocalizedNumberFormatter`, so it now has ownership of the `DecimalFormatSymbols` object created in step b above.

d) Then `DecimalFormat::touch()` assigns that object to `fields->formatter`, using the move-assignment operator on `LocalizedNumberFormat`.  This deletes the old `DecimalFormatSymbols` object from step a, replacing it with the copy we created in step b.

e) At this point, `symbols` from line 1617 of `DecimalFormat::touch()` is a dangling pointer, but this variable isn't referenced again in this function after line 1627.

So we're copying the object before we delete it and not referencing it again after.  There's no double deletion, no referencing a deleted object through a dangling pointer, and no memory leak.

2) I completely agree with you that it'd be better if we weren't copying anything and the `LocalizedNumberFormat` just took ownership of the existing `DecimalFormatSymbols` object.  But that's what it's been doing all along, as far as I can tell, and to be honest, I'm scared to try to make that change.  I tried this afternoon, and it rippled through a lot of different classes, and I quickly got lost trying to keep track of who owned it when and how I pass that along without violating any other assumptions this code may be making.  My first attempt blew up with exactly the kind of double-delete problem I was worried about.  Given that what I've got right now seems to work correctly without changing any of the assumptions built into the rest of this code, I'd really rather that somebody who understands this code better than I do (or maybe that should just be "somebody who understands this code", since I'm not sure I do) make that change.

sffc: Okay, as long as the move-assignment (step d) is guaranteed to happen after the clone (step b), we're fine.

sffc (commented):
LGTM please squash

icu4c/source/i18n/decimfmt.cpp
sffc: One suggestion before you squash: could you add a comment to the file explaining this, so that someone in the future doesn't accidentally do something like reference `symbols` after it becomes a dangling pointer?

richgillam: Oops.  I didn't see your comment before I squashed.  Do you still want me to add that comment and squash again?

sffc: Yes please :)

richgillam: Done!

sffc (approved):

PR #1766, opened 2021-07-09
ICU-21669 Replace UPRV_UNREACHABLE that can actually be reached with set error status

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21669

Change UPRV_UNREACHABLE that can actually be reached (possibly due to memory corruption or other problems) with setting of error status (and return, if code will not already immediately return). Not possible to create unit tests for this since it is an abnormal situation.

### DIFF
--- a/icu4c/source/common/uloc_keytype.cpp
+++ b/icu4c/source/common/uloc_keytype.cpp
@@ -168,12 +168,11 @@ initFromResourceBundle(UErrorCode& sts) {
         }
 
         // look up type map for the key, and walk through the mapping data
-        tmpSts = U_ZERO_ERROR;
-        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &tmpSts));
-        if (U_FAILURE(tmpSts)) {
-            // type map for each key must exist
-            UPRV_UNREACHABLE;
-        } else {
+        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &sts));
+        // We fail here if typeMap does not have an entry corresponding to every entry in keyMap (should
+        // not happen for valid keyTypeData), or if ures_getByKeyfails fails for some other reason.
+        // Either way, we should just return the error, not invoke UPRV_UNREACHABLE i.e. abort().
+        if (U_SUCCESS(sts)) {
             LocalUResourceBundlePointer typeMapEntry;
 
             while (ures_hasNext(typeMapResByKey.getAlias())) {
--- a/icu4c/source/i18n/rematch.cpp
+++ b/icu4c/source/i18n/rematch.cpp
@@ -879,7 +879,12 @@ UBool RegexMatcher::find(UErrorCode &status) {
         }
 
     default:
-        UPRV_UNREACHABLE;
+        // Unknown value in fPattern->fStartType, should be from StartOfMatch enum.
+        // Should never use UPRV_UNREACHABLE (unconditional abort) in code
+        // that may be reached in production (as this can do in some
+        // circumstances). Instead, set error status and return.
+        status = U_INTERNAL_PROGRAM_ERROR;
+        return FALSE;
     }
 
     UPRV_UNREACHABLE;
@@ -1135,7 +1140,12 @@ UBool RegexMatcher::findUsingChunk(UErrorCode &status) {
     }
 
     default:
-        UPRV_UNREACHABLE;
+        // Unknown value in fPattern->fStartType, should be from StartOfMatch enum.
+        // Should never use UPRV_UNREACHABLE (unconditional abort) in code
+        // that may be reached in production (as this can do in some
+        // circumstances). Instead, set error status and return.
+        status = U_INTERNAL_PROGRAM_ERROR;
+        return FALSE;
     }
 
     UPRV_UNREACHABLE;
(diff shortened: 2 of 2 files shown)


icu4c/source/common/uloc_keytype.cpp
```diff
@@ -168,12 +168,11 @@ initFromResourceBundle(UErrorCode& sts) {
...
         // look up type map for the key, and walk through the mapping data
-        tmpSts = U_ZERO_ERROR;
-        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &tmpSts));
-        if (U_FAILURE(tmpSts)) {
-            // type map for each key must exist
-            UPRV_UNREACHABLE;
-        } else {
+        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &sts));
+        // We fail here if typeMap does not have an entry corresponding to every entry in keyMap (should
+        // not happen for valid keyTypeData), or if ures_getByKeyfails fails for some other reason.
```
jefgen, line 173: FWIW, this can also occur if the data file can't be loaded (or when using stubdata), or if someone over aggressively removes data with filtering and omits things like timezoneTypes.res.

icu4c/source/i18n/rematch.cpp
```diff
@@ -5672,7 +5686,11 @@ void RegexMatcher::MatchChunkAt(int32_t startIdx, UBool toEnd, UErrorCode &statu
         default:
             // Trouble.  The compiled pattern contains an entry with an
             //           unrecognized type tag.
-            UPRV_UNREACHABLE;
+            // Unknown opcode type in opType = URX_TYPE(pat[fp->fPatIdx]).
+            // Should never use UPRV_UNREACHABLE (unconditional abort) in code
```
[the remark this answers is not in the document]
markusicu: I agree. If someone clobbered the stack or the program code, all bets are off, and crashing is fair, so we should keep UPRV_UNREACHABLE in such cases.

If we can write a unit test that gets us to an "unreachable" place, then it's not unreachable, and we need to handle it.

If we struggle to write a unit test, but by analyzing the code we can convince ourselves that there is a reasonable way to get to some place, we should handle it.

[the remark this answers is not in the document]
markusicu: In other words, for the changes here in rematch.cpp, do you have a test case that gets here, or have you found ways that you think we could get here?

icu4c/source/common/uloc_keytype.cpp
```diff
@@ -168,12 +168,11 @@ initFromResourceBundle(UErrorCode& sts) {
...
 
         // look up type map for the key, and walk through the mapping data
-        tmpSts = U_ZERO_ERROR;
-        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &tmpSts));
-        if (U_FAILURE(tmpSts)) {
-            // type map for each key must exist
-            UPRV_UNREACHABLE;
-        } else {
+        LocalUResourceBundlePointer typeMapResByKey(ures_getByKey(typeMapRes.getAlias(), legacyKeyId, NULL, &sts));
+        // We fail here if typeMap does not have an entry corresponding to every entry in keyMap (should
```
FrankYFTang, line 172: please assert this will always SUCCESS

icu4c/source/i18n/rematch.cpp
```diff
@@ -879,7 +879,12 @@ UBool RegexMatcher::find(UErrorCode &status) {
         }
 
     default:
-        UPRV_UNREACHABLE;
+        // Unknown value in fPattern->fStartType, should be from StartOfMatch enum.
```
FrankYFTang, line 882: could we assert this will not happen in debug build

icu4c/source/i18n/rematch.cpp
```diff
@@ -1135,7 +1140,12 @@ UBool RegexMatcher::findUsingChunk(UErrorCode &status) {
     }
 
     default:
-        UPRV_UNREACHABLE;
+        // Unknown value in fPattern->fStartType, should be from StartOfMatch enum.
```
FrankYFTang, line 1143: same here, please assert in debug build

icu4c/source/i18n/rematch.cpp
```diff
@@ -4234,7 +4244,11 @@ void RegexMatcher::MatchAt(int64_t startIdx, UBool toEnd, UErrorCode &status) {
         default:
             // Trouble.  The compiled pattern contains an entry with an
             //           unrecognized type tag.
-            UPRV_UNREACHABLE;
+            // Unknown opcode type in opType = URX_TYPE(pat[fp->fPatIdx]).
```
FrankYFTang, line 4247: please assert in debug build

markusicu:
Discussed in today's team meeting -- consensus:
- Rename UPRV_UNREACHABLE to UPRV_UNREACHABLE_EXIT
- Add new UPRV_UNREACHABLE_ASSERT which does not exist in release mode
- In places where production code *has failed*, change to UPRV_UNREACHABLE_ASSERT and return an error.

pedberg-icu:
I am closing this and replacing with a new PR https://github.com/unicode-org/icu/pull/1860

PR #1767, opened 2021-07-13
ICU-21676 Fix strcpy buffer override in ultag_parse

FrankYFTang:
Only happen with legacy locale tag

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21676

### DIFF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -2089,6 +2089,7 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             legacyLen = checkLegacyLen;  /* back up for output parsedLen */
             int32_t replacementLen = static_cast<int32_t>(uprv_strlen(LEGACY[i+1]));
             newTagLength = replacementLen + tagLen - checkLegacyLen;
+            int32_t oldTagLength = tagLen;
             if (tagLen < newTagLength) {
                 uprv_free(tagBuf);
                 tagBuf = (char*)uprv_malloc(newTagLength + 1);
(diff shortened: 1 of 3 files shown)


gnrunge (dismissed):

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2102,7 +2103,10 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             parsedLenDelta = checkLegacyLen - replacementLen;
             uprv_strcpy(t->buf, LEGACY[i + 1]);
             if (checkLegacyLen != tagLen) {
-                uprv_strcpy(t->buf + replacementLen, tag + checkLegacyLen);
+                uprv_strncpy(t->buf + replacementLen, tag + checkLegacyLen,
+                             oldTagLength - checkLegacyLen);
+                // null terminate after strncpy
```
markusicu, line 2108:
```suggestion
                uprv_memcpy(t->buf + replacementLen, tag + checkLegacyLen,
                            oldTagLength - checkLegacyLen);
                // NUL-terminate after memcpy().
```
When we *know* how many bytes to copy, and never want to even think about a NUL-terminated source string, then we should use memcpy().

markusicu (commented):
looks good but I think memcpy() would be better

FrankYFTang:
> looks good but I think memcpy() would be better

thanks. That is a good suggestion. I should think about that.

FrankYFTang:
PTAL

roubert (approved):
LGTM

markusicu (approved):
lgtm tnx

PR #1770, opened 2021-07-19
ICU-21202 Fixed error in alias resolution in uresbund.cpp.

richgillam:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21202

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -1146,6 +1146,17 @@ static UResourceBundle *init_resb_result(const ResourceData *rdata, Resource r,
                                     r = res_findResource(&(dataEntry->fData), r, &myPath, &temp);
                                     if(r != RES_BOGUS) { /* found a resource, but it might be an indirection */
                                         resB = init_resb_result(&(dataEntry->fData), r, temp, -1, dataEntry, result, noAlias+1, resB, status);
+                                        if (temp == NULL || uprv_strcmp(keyPath, temp) != 0) {
+                                            // the call to init_resb_result() above will set resB->fKeyPath to be
+                                            // the same as resB->fKey, throwing away any additional path elements if we
+                                            // had them-- if the key path wasn't just a single resource ID, clear out
+                                            // the bundle's key path and re-set it to be equal to keyPath
+                                            ures_freeResPath(resB);
+                                            ures_appendResPath(resB, keyPath, (int32_t)uprv_strlen(keyPath), status);
+                                            if(resB->fResPath[resB->fResPathLen-1] != RES_PATH_SEPARATOR) {
+                                                ures_appendResPath(resB, RES_PATH_SEPARATOR_S, 1, status);
+                                            }
+                                        }
                                         result = resB;
                                         if(result) {
                                             r = result->fRes; /* switch to a new resource, possibly a new tree */
--- a/icu4c/source/test/cintltst/udatpg_test.c
+++ b/icu4c/source/test/cintltst/udatpg_test.c
@@ -45,6 +45,7 @@ static void TestOptions(void);
 static void TestGetFieldDisplayNames(void);
 static void TestGetDefaultHourCycle(void);
 static void TestGetDefaultHourCycleOnEmptyInstance(void);
+static void TestEras(void);
 
 void addDateTimePatternGeneratorTest(TestNode** root) {
     TESTCASE(TestOpenClose);
@@ -54,6 +55,7 @@ void addDateTimePatternGeneratorTest(TestNode** root) {
     TESTCASE(TestGetFieldDisplayNames);
     TESTCASE(TestGetDefaultHourCycle);
     TESTCASE(TestGetDefaultHourCycleOnEmptyInstance);
+    TESTCASE(TestEras);
 }
 
 /*
@@ -578,6 +580,38 @@ static void TestGetDefaultHourCycleOnEmptyInstance() {
     }
 
     udatpg_close(dtpgen);
+}
+
+// Test for ICU-21202: Make sure DateTimePatternGenerator supplies an era field for year formats using the
+// Buddhist and Japanese calendars for all English-speaking locales.
+static void TestEras(void) {
+    const char* localeIDs[] = {
+        "en_US@calendar=japanese",
+        "en_GB@calendar=japanese",
+        "en_150@calendar=japanese",
+        "en_001@calendar=japanese",
+        "en@calendar=japanese",
+        "en_US@calendar=buddhist",
+        "en_GB@calendar=buddhist",
+        "en_150@calendar=buddhist",
+        "en_001@calendar=buddhist",
+        "en@calendar=buddhist",
+    };
+    
+    UErrorCode err = U_ZERO_ERROR;
+    for (int32_t i = 0; i < UPRV_LENGTHOF(localeIDs); i++) {
+        const char* locale = localeIDs[i];
+        UDateTimePatternGenerator* dtpg = udatpg_open(locale, &err);
+        if (U_SUCCESS(err)) {
+            UChar pattern[200];
+            udatpg_getBestPattern(dtpg, u"y", 1, pattern, 200, &err);
+            
+            if (u_strchr(pattern, u'G') == NULL) {
+                log_err("missing era field for locale %s\n", locale);
+            }
+        }
+        udatpg_close(dtpg);
+    }
 }
 
 #endif
--- a/icu4c/source/test/intltest/dtptngts.cpp
+++ b/icu4c/source/test/intltest/dtptngts.cpp
@@ -1588,6 +1588,11 @@ void IntlTestDateTimePatternGeneratorAPI::testBestPattern() {
         "en_US",      "yMMMMEEEEd",  "EEEE, MMMM d, y",
         "en_US",      "yMMMMccccd",  "EEEE, MMMM d, y",
         "en_US",      "yMMMMeeeed",  "EEEE, MMMM d, y",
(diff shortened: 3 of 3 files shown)


richgillam:
The actual fix for the horizontal inheritance problem itself is a one-line change in uresbund.cpp: changing `temp` to `keyPath` in the recursive call to `init_resb_result()` on line 1148 of `init_resb_result()`. I figure I'd better take some time to explain why that fixes the problem...

To go back, the original symptom that I was trying to address was that when you created a `DateTimePatternGenerator` for `en_GB@calendar=buddhist` and asked it for a pattern for the skeleton string `"y"`, you'd get back a pattern with no era field, even though the Buddhist calendar requires the era field.  `en_GB.txt` doesn't have a `calendar/buddhist` entry at all; neither does its parent `en_001.txt`. `en.txt` has `calendar/buddhist`, but not `calendar/buddhist/availableFormats`. So we fall back to `root.txt`, which aliases `calendar/buddhist/availableFormats` over to `calendar/generic/availableFormats`.
`calendar/generic/availableFormats` in `en_GB.txt` doesn't have a pattern for `"y"`. Neither does `en_001.txt`.  `en.txt` DOES, and it gives us back `"y G"`, which is what we're supposed to be returning, but the resource bundle lookup doesn't get here. As far as I could make out, following the alias messes things up.

Things go off the rails in `getAllItemsWithFallback()` at lines 1994-2007 of `uresbund.cpp`. This function correctly follows the alias from `calendar/buddhist/availableFormats` in `root.txt` to `calendar/generic/availableFormats` in `en_GB.txt`, but then fails trying to ascend `en_GB`'s parent chain. That's because the call to `ures_getByKetWithFallback()` on line 2002 fails with a `MISSING_RESOURCE_ERROR` instead of taking us to `calendar/generic/availableFormats` in `en_001.txt`. That's because the resource path we're looking for is wrong: `bundle->fResPath` is set to `"availableFormats"`, when what it should be set to is `"calendar/generic/availableFormats"`.

The `fResPath` member of `UResourceBundle` is set in line 1236 of `init_resb_result()`, and it's set to the value passed in for `key`. This is usually the full resource path, but not in the case where we follow an alias. The alias-following logic is also in `init_resb_result()`, which eventually calls itself recursively to fill in the data in the `UResourceBundle`. Appropriately, the alias-following logic is where the comment "this one is a bit trickier." is: We locate the resource the alias points to with the call to `res_findResource()` on line 1146. This function returns the actual name of the resource in a fill-in parameter (in this case, `&temp`), but that's just the actual resource name, not its full path. So we end up passing `"availableFormats"` to `init_resb_result()` on line 1303 instead of `"calendar/generic/availableFormats"`, and this is what causes the call to `ures_getByKeyWithFallback()` in `getAllItemsWithFallback()` to fail. But as it turns out, we already have the full path at our disposal-- it's in `keyPath`. So the solution is just to pass that instead of `temp`.

Clear as mud, huh?

The big question I had is whether using `keyPath` instead of `temp` here would have side effects: If we have to follow multiple aliases, does that code only want the bottom resource name and not the full path? All I can say is that we have resource-bundle tests that check following multiple aliases, and they're apparently still passing (as are all the others).

I was hoping this would fix a bunch of our other horizontal-inheritance problems.  It fixes https://unicode-org.atlassian.net/browse/ICU-21428, but not the other horizontal-inheritance problems I've looked at so far.  I hope to have more PRs covering the other issues soon.

markusicu:
Thank you very much for the forensics, and the comprehensive write-up!

This looks plausible to me. Please fix the CI failures and then squash.

I am also a bit nervous about side effects, but if it fixes two bugs and does not cause other unit test failures, that seems good.

For side effects, in particular, I thought that when we get a UResourceBundle and ask it "who are you?" it should pretend to report the path where we logically found it, even if alias resolution detoured elsewhere. But apparently the `temp` name was already in the alias target, just not a whole path, so either this value is not what it would report from such an API or it already didn't do what I thought. (Which might be a pre-existing bug or my mis-remembering how it wants to work.) Anyway, we have long struggled with "who are you?" APIs vs. aliases and fallbacks (hence the many `@draft` getLocale() APIs) so maybe that's not important.

srl295 (dismissed):
thanks for the work on this

richgillam:
Turns out my original fix was too good to be true. When I posted it here, I got a Valgrind error (still surprised I didn't also get an ASAN error) that turned out to be legit.

There's a comment in the code explaining the new code, but to supply a little more background, the recursive call to `init_resb_result()` on line 1148 of uresbund.cpp (inside`init_resb_result()`) was corrupting memory. The `key` parameter to this function, where we were originally passing `temp` and I changed to pass `keyPath`, is used to set two different fields on `resB`: `fKey` and `fResPath`. We really don't want those to be the same value-- `fKey` should just be the key of that particular resource, and `fResPath` should be the resource's entire path. More importantly, `fResPath` has its own storage in the `UResourceBundle` object, but `fKey` is just a simple pointer. It's supposed to point into the static memory-mapped resource data (which is what the call to `res_findResource()` above the line in question was doing), but I changed it so that it was pointing into the stack, which is obviously a Very Bad Thing.

So I changed that line back to its old value and instead just added code to manually fix up `fResPath` afterward. It's a little clunky, but it seems to work and makes the memory-corruption problem go away, and I couldn't think of a better alternative.

icu4c/source/common/uresbund.cpp
```diff
@@ -1146,6 +1146,17 @@ static UResourceBundle *init_resb_result(const ResourceData *rdata, Resource r,
                                     r = res_findResource(&(dataEntry->fData), r, &myPath, &temp);
                                     if(r != RES_BOGUS) { /* found a resource, but it might be an indirection */
                                         resB = init_resb_result(&(dataEntry->fData), r, temp, -1, dataEntry, result, noAlias+1, resB, status);
+                                        if (temp == NULL || uprv_strcmp(keyPath, temp) != 0) {
```
markusicu, line 1149: optional: nullptr not NULL

icu4c/source/common/uresbund.cpp
```diff
@@ -1146,6 +1146,17 @@ static UResourceBundle *init_resb_result(const ResourceData *rdata, Resource r,
...
+                                            // the call to init_resb_result() above will set resB->fKeyPath to be
```
markusicu, line 1150: question: is that behavior a bug? or do we need it for something?

markusicu (commented):
still looks plausible but have a question
if that's not a bug, or not something that we need to look into now, then please squash

icu4c/source/common/uresbund.cpp
richgillam: There are lots of places in the same function where the existing code is using "`NULL`", and I didn't want to change all of them, and so I followed the established precedent.  Did we all decide that one should change "`NULL`" to `"nullptr`" everywhere when making other changes to a file?  If so, should I do that just to the functions I changed, or to the whole file?

icu4c/source/common/uresbund.cpp
richgillam: You're asking whether setting both `fKey` and `fResPath` to the same thing is a bug?  I'm not completely sure, to tell you the truth, but I think it's generally desirable behavior other functions are generally relying on.  I think a lot of the time, in a lot of contexts, they actually _are_ the same, or it doesn't really matter.  But I did run into two spots (see ICU-20400 for the other one) where we really do want them to be different.  I thought about expanding `init_resb_result()` so that it just took a separate path parameter and set both things independently, but since it's an internal function that's only called in about five or six places in this file, I decided to leave it alone and just fix up the res path after the fact.  But maybe it really _would_ make sense to have `init_resb_result()` just take two parameters for these values; I can take a shot at that if you want.

One interesting this is that the two spots where we want these values to be different were doing two things: Here, we're setting both to the "key" value and throwing away the additional path elements.  In the similar spot in ICU-20400, we were setting both to the "path" value.  That didn't seem to be hurting anything, since nobody in that code path was really looking at the "key" value-- it mainly made debugging more difficult.  But it was also scary because the resource bundle doesn't own the storage for its `fKey` member-- sticking a path in there only worked because we happened to be setting it to point to another resource's `fResPath` member (which _does_ own its storage).  Of course, if the other resource happened to go away before this one did, we'd be in trouble, but that wasn't happening in practice (Thank God).

Let me know if you want to take a whack at just having `init_resb_result()` take separate parameters for the key and path; that seems doable.  If you don't think it's worth the trouble, I'll squash.

icu4c/source/common/uresbund.cpp
markusicu: I like to use nullptr when I write new code, but it creates a mix with the existing NULL. Maybe for 71 we should plan to change all NULL to nullptr in .cpp files?

Anyway, I marked it as "optional", so you can leave it as is.

icu4c/source/common/uresbund.cpp
markusicu: Not worth now I think, but sounds fishy. Please submit another ticket with something like what you wrote here. We should really take some time (pair programming) looking through the code and the API semantics and thinking about what makes sense.

As I said elsewhere, sometimes we get into trouble because we have "who am I" APIs and the answer sometimes depends on what you did or what you want to use the answer for. (Mark likes to add "who am I" APIs, and I have been resisting for some time now.)

richgillam: Okay, I've filed https://unicode-org.atlassian.net/browse/ICU-21715 for the additional work, and I'll go ahead and squash this.  Thank you!

markusicu (approved):
lgtm

PR #1772, opened 2021-07-22
ICU-21688 [documentation] escape angle brackets in non-HTML content

dungsaga:
`<name>` should be escape with backslash to prevent it from being render as a HTML tag


dungsaga:
Please ignore this PR. I opened a more correct PR at https://github.com/unicode-org/icu/pull/1774

PR #1774, opened 2021-07-22
ICU-21688 escape angle bracket in non-HTML content

dungsaga:
`<name>` should be escaped with backslash to prevent it from being rendered as a HTML tag

Jira issue: https://unicode-org.atlassian.net/browse/ICU-21688


sven-oly (approved):
Thanks for this update.

PR #1777, opened 2021-07-24
ICU-20992 Added new internal API for iterating child resources while following…

richgillam:
… aliases, and changed DateTimePatternGenerator

to use it.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20992

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -2015,6 +2015,7 @@ void getAllItemsWithFallback(
     // then it would remove the parent's item.
     // We would deserialize parent values even though they are overridden in a child bundle.
     value.setData(bundle->getResData());
+    value.setValidLocaleDataEntry(bundle->fValidLocaleDataEntry);
     UResourceDataEntry *parentEntry = bundle->fData->fParent;
     UBool hasParent = parentEntry != NULL && U_SUCCESS(parentEntry->fBogus);
     value.setResource(bundle->fRes, ResourceTracer(bundle));
@@ -2055,6 +2056,49 @@ void getAllItemsWithFallback(
     }
 }
 
+struct GetAllChildrenSink : public ResourceSink {
+    // Destination sink
+    ResourceSink& dest;
+
+    GetAllChildrenSink(ResourceSink& dest)
+        : dest(dest) {}
+    virtual ~GetAllChildrenSink() override;
+    virtual void put(const char *key, ResourceValue &value, UBool isRoot,
+           UErrorCode &errorCode) override {
+        ResourceTable itemsTable = value.getTable(errorCode);
+        if (U_FAILURE(errorCode)) { return; }
+        for (int32_t i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
+            if (value.getType() == URES_ALIAS) {
+                ResourceDataValue& rdv = static_cast<ResourceDataValue&>(value);
+                StackUResourceBundle stackTempBundle;
+                UResourceBundle* aliasRB = getAliasTargetAsResourceBundle(rdv.getData(), rdv.getResource(), nullptr, -1,
+                                                                          rdv.getValidLocaleDataEntry(), nullptr, 0,
+                                                                          stackTempBundle.getAlias(), &errorCode);
+                if (U_SUCCESS(errorCode)) {
+                    ResourceDataValue aliasedValue;
+                    aliasedValue.setData(aliasRB->getResData());
+                    aliasedValue.setValidLocaleDataEntry(aliasRB->fValidLocaleDataEntry);
+                    aliasedValue.setResource(aliasRB->fRes, ResourceTracer(aliasRB));
+                    dest.put(key, aliasedValue, isRoot, errorCode);
+                }
+            } else {
+                dest.put(key, value, isRoot, errorCode);
+            }
+            if (U_FAILURE(errorCode)) { return; }
+        }
+    }
+};
+
+// Virtual destructors must be defined out of line.
+GetAllChildrenSink::~GetAllChildrenSink() {}
+
+U_CAPI void U_EXPORT2
+ures_getAllChildrenWithFallback(const UResourceBundle *bundle, const char *path,
+                                icu::ResourceSink &sink, UErrorCode &errorCode) {
+    GetAllChildrenSink allChildrenSink(sink);
+    ures_getAllItemsWithFallback(bundle, path, allChildrenSink, errorCode);
+}
+
 }  // namespace
 
 // Requires a ResourceDataValue fill-in, so that we need not cast from a ResourceValue.
@@ -2086,6 +2130,7 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
         }
     }
     value.setData(rb->getResData());
+    value.setValidLocaleDataEntry(rb->fValidLocaleDataEntry);
     value.setResource(rb->fRes, ResourceTracer(rb));
 }
 
--- a/icu4c/source/common/uresdata.h
+++ b/icu4c/source/common/uresdata.h
@@ -399,6 +399,8 @@ typedef struct ResourceData {
     UBool useNativeStrcmp;
 } ResourceData;
 
+struct UResourceDataEntry;   // forward declared for ResoureDataValue below; actually defined in uresimp.h
+
 /*
  * Read a resource bundle from memory.
  */
@@ -512,12 +514,17 @@ class ResourceDataValue : public ResourceValue {
 public:
     ResourceDataValue() :
         pResData(nullptr),
+        validLocaleDataEntry(nullptr),
         res(static_cast<Resource>(URES_NONE)),
         fTraceInfo() {}
     virtual ~ResourceDataValue();
 
     void setData(const ResourceData &data) {
         pResData = &data;
+    }
+    
+    void setValidLocaleDataEntry(UResourceDataEntry *entry) {
+        validLocaleDataEntry = entry;
     }
 
     void setResource(Resource r, ResourceTracer&& traceInfo) {
@@ -526,6 +533,8 @@ public:
     }
 
     const ResourceData &getData() const { return *pResData; }
+    UResourceDataEntry *getValidLocaleDataEntry() const { return validLocaleDataEntry; }
+    Resource getResource() const { return res; }
     virtual UResType getType() const;
     virtual const UChar *getString(int32_t &length, UErrorCode &errorCode) const;
     virtual const UChar *getAliasString(int32_t &length, UErrorCode &errorCode) const;
@@ -544,6 +553,7 @@ public:
 
 private:
     const ResourceData *pResData;
+    UResourceDataEntry *validLocaleDataEntry;
     Resource res;
     ResourceTracer fTraceInfo;
 };
--- a/icu4c/source/common/uresimp.h
+++ b/icu4c/source/common/uresimp.h
@@ -294,9 +294,38 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
                           UResourceBundle *tempFillIn,
                           icu::ResourceDataValue &value, UErrorCode &errorCode);
 
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and calls the specified `sink`'s `put()` method with that
+ * resource.  Then walks the bundle's parent chain, calling `put()` on the sink for each item in the
+ * parent chain.
+ * @param bundle The bundle to search
+ * @param path The path of the desired resource
+ * @param sink A `ResourceSink` that gets called for each resource in the parent chain
+ * @param errorCode The error code
+ */
 U_CAPI void U_EXPORT2
 ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
                              icu::ResourceSink &sink, UErrorCode &errorCode);
+
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and, if the resource is a table resource, iterates over its
+ * immediate child resources (again, following any aliases to get the individual resource values), and calls the specified
+ * `sink`'s `put()` method for each child resource (passing it that resource's key and either its actual value or,
+ * if that value is an alias, the value you get by following the alias).  Then walks back over the bundle's
+ * parent chain, similarly iterating over each parent table resource's child resources.
+ * Does not descend beyond one level of table children.
+ * @param bundle The bundle to search
+ * @param path The path of the desired resource
+ * @param sink A `ResourceSink` that gets called for each child resource of the specified resource (and each child
+ * of the resources in its parent chain).
+ * @param errorCode The error code.  This will be U_RESOURCE_TYPE_MISMATCH if the resource the caller
+ * is asking for isn't a table resource.
+ */
+U_CAPI void U_EXPORT2
+ures_getAllChildrenWithFallback(const UResourceBundle *bundle, const char *path,
+                                icu::ResourceSink &sink, UErrorCode &errorCode);
 
 #endif  /* __cplusplus */
 
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
@@ -395,6 +395,13 @@ public  class ICUResourceBundle extends UResourceBundle {
         }
     }
 
+    /**
+     * Locates the resource specified by `path` in this resource bundle (performing any necessary fallback and
+     * following any aliases) and calls the specified `sink`'s `put()` method with that resource.  Then walks the
+     * bundle's parent chain, calling `put()` on the sink for each item in the parent chain.
+     * @param path The path of the desired resource
+     * @param sink A `UResource.Sink` that gets called for each resource in the parent chain
+     */
     public void getAllItemsWithFallback(String path, UResource.Sink sink)
             throws MissingResourceException {
         // Collect existing and parsed key objects into an array of keys,
@@ -419,6 +426,47 @@ public  class ICUResourceBundle extends UResourceBundle {
         UResource.Key key = new UResource.Key();
         ReaderValue readerValue = new ReaderValue();
         rb.getAllItemsWithFallback(key, readerValue, sink);
+    }
+
+    /**
+     * Locates the resource specified by `path` in this resource bundle (performing any necessary fallback and
+     * following any aliases) and, if the resource is a table resource, iterates over its immediate child resources (again,
+     * following any aliases to get the individual resource values), and calls the specified `sink`'s `put()` method
+     * for each child resource (passing it that resource's key and either its actual value or, if that value is an
+     * alias, the value you get by following the alias).  Then walks back over the bundle's parent chain,
+     * similarly iterating over each parent table resource's child resources.
+     * Does not descend beyond one level of table children.
+     * @param path The path of the desired resource
+     * @param sink A `UResource.Sink` that gets called for each child resource of the specified resource (and each child
+     * of the resources in its parent chain).
+     */
+    public void getAllChildrenWithFallback(final String path, final UResource.Sink sink)
+            throws MissingResourceException {
+        class AllChildrenSink extends UResource.Sink {
+            @Override
+            public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
+                UResource.Table itemsTable = value.getTable();
+                for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
+                    if (value.getType() == ALIAS) {
+                        // if the current entry in the table is an alias, re-fetch it using getAliasedResource():
+                        // this will follow the alias (and any aliases it points to) and bring back the real value
+                        String aliasPath = value.getAliasString();
+                        ICUResourceBundle aliasedResource = getAliasedResource(aliasPath, wholeBundle.loader,
+                                "", null, 0, null,
+                                null, ICUResourceBundle.this);
+                        ICUResourceBundleImpl aliasedResourceImpl = (ICUResourceBundleImpl)aliasedResource;
+                        ReaderValue aliasedValue = new ReaderValue();
+                        aliasedValue.reader = aliasedResourceImpl.wholeBundle.reader;
+                        aliasedValue.res = aliasedResourceImpl.getResource();
+                        sink.put(key, aliasedValue, noFallback);
+                    } else {
+                        sink.put(key, value, noFallback);
+                    }
+                }
+            }
+        }
+
+        getAllItemsWithFallback(path, new AllChildrenSink());
     }
 
     private void getAllItemsWithFallback(
@@ -1481,10 +1529,27 @@ public  class ICUResourceBundle extends UResourceBundle {
             UResourceBundle requested) {
         WholeBundle wholeBundle = base.wholeBundle;
         ClassLoader loaderToUse = wholeBundle.loader;
+        String rpath = wholeBundle.reader.getAlias(_resource);
+        String baseName = wholeBundle.baseName;
+
+        // TODO: We need not build the baseKeyPath array if the rpath includes a keyPath
+        // (except for an exception message string).
+        // Try to avoid unnecessary work+allocation.
+        int baseDepth = base.getResDepth();
+        String[] baseKeyPath = new String[baseDepth + 1];
+        base.getResPathKeys(baseKeyPath, baseDepth);
+        baseKeyPath[baseDepth] = key;
+        return getAliasedResource(rpath, loaderToUse, baseName, keys, depth, baseKeyPath, aliasesVisited, requested);
+    }
+
+    protected static ICUResourceBundle getAliasedResource(
+            String rpath, ClassLoader loaderToUse, String baseName,
+            String[] keys, int depth, String[] baseKeyPath,
+            HashMap<String, String> aliasesVisited,
+            UResourceBundle requested) {
         String locale;
         String keyPath = null;
         String bundleName;
-        String rpath = wholeBundle.reader.getAlias(_resource);
         if (aliasesVisited == null) {
             aliasesVisited = new HashMap<>();
         }
@@ -1523,12 +1588,12 @@ public  class ICUResourceBundle extends UResourceBundle {
             } else {
                 locale = rpath;
             }
-            bundleName = wholeBundle.baseName;
+            bundleName = baseName;
         }
         ICUResourceBundle bundle = null;
         ICUResourceBundle sub = null;
         if(bundleName.equals(LOCALE)){
-            bundleName = wholeBundle.baseName;
+            bundleName = baseName;
             keyPath = rpath.substring(LOCALE.length() + 2/* prepending and appending / */, rpath.length());
 
             // Get the top bundle of the requested bundle
@@ -1550,11 +1615,8 @@ public  class ICUResourceBundle extends UResourceBundle {
             } else if (keys != null) {
                 numKeys = depth;
             } else {
-                depth = base.getResDepth();
-                numKeys = depth + 1;
-                keys = new String[numKeys];
-                base.getResPathKeys(keys, depth);
-                keys[depth] = key;
+                keys = baseKeyPath;
+                numKeys = baseKeyPath.length;
             }
             if (numKeys > 0) {
                 sub = bundle;
@@ -1564,7 +1626,7 @@ public  class ICUResourceBundle extends UResourceBundle {
             }
         }
         if (sub == null) {
-            throw new MissingResourceException(wholeBundle.localeID, wholeBundle.baseName, key);
+            throw new MissingResourceException(locale, baseName, baseKeyPath[baseKeyPath.length - 1]);
         }
         // TODO: If we know that sub is not cached,
         // then we should set its container and key to the alias' location,
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
@@ -213,13 +213,10 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
     private class AppendItemFormatsSink extends UResource.Sink {
         @Override
         public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
-            UResource.Table itemsTable = value.getTable();
-            for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
-                int field = getAppendFormatNumber(key);
-                assert field != -1;
-                if (getAppendItemFormat(field) == null) {
-                    setAppendItemFormat(field, value.toString());
-                }
+            int field = getAppendFormatNumber(key);
+            if (field < 0) { return; }
+            if (getAppendItemFormat(field) == null) {
+                setAppendItemFormat(field, value.toString());
             }
         }
     }
@@ -227,26 +224,14 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
     private class AppendItemNamesSink extends UResource.Sink {
         @Override
         public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
-            UResource.Table itemsTable = value.getTable();
-            for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
-                if (value.getType() != UResourceBundle.TABLE) {
-                    // Typically get either UResourceBundle.TABLE = 2 or ICUResourceBundle.ALIAS = 3.
-                    // Currently fillInMissing() is being used instead of following the ALIAS, so
-                    // skip ALIAS entries which cause UResourceTypeMismatchException in the line
-                    // UResource.Table detailsTable = value.getTable()
-                    continue;
-                }
-                int fieldAndWidth = getCLDRFieldAndWidthNumber(key);
-                if (fieldAndWidth == -1) { continue; }
-                int field = fieldAndWidth / DisplayWidth.COUNT;
-                DisplayWidth width = CLDR_FIELD_WIDTH[fieldAndWidth % DisplayWidth.COUNT];
-                UResource.Table detailsTable = value.getTable();
-                for (int j = 0; detailsTable.getKeyAndValue(j, key, value); ++j) {
-                    if (!key.contentEquals("dn")) continue;
-                    if (getFieldDisplayName(field, width) == null) {
-                        setFieldDisplayName(field, width, value.toString());
-                    }
-                    break;
+            int fieldAndWidth = getCLDRFieldAndWidthNumber(key);
+            if (fieldAndWidth == -1) { return; }
+            int field = fieldAndWidth / DisplayWidth.COUNT;
+            DisplayWidth width = CLDR_FIELD_WIDTH[fieldAndWidth % DisplayWidth.COUNT];
+            UResource.Table detailsTable = value.getTable();
+            if (detailsTable.findValue("dn", value)) {
+                if (getFieldDisplayName(field, width) == null) {
+                    setFieldDisplayName(field, width, value.toString());
                 }
             }
         }
@@ -277,16 +262,13 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
 
         @Override
         public void put(UResource.Key key, UResource.Value value, boolean isRoot) {
-            UResource.Table formatsTable = value.getTable();
-            for (int i = 0; formatsTable.getKeyAndValue(i, key, value); ++i) {
-                String formatKey = key.toString();
-                if (!isAvailableFormatSet(formatKey)) {
-                    setAvailableFormat(formatKey);
-                    // Add pattern with its associated skeleton. Override any duplicate derived from std patterns,
-                    // but not a previous availableFormats entry:
-                    String formatValue = value.toString();
-                    addPatternWithSkeleton(formatValue, formatKey, !isRoot, returnInfo);
-                }
+            String formatKey = key.toString();
+            if (!isAvailableFormatSet(formatKey)) {
+                setAvailableFormat(formatKey);
+                // Add pattern with its associated skeleton. Override any duplicate derived from std patterns,
+                // but not a previous availableFormats entry:
+                String formatValue = value.toString();
+                addPatternWithSkeleton(formatValue, formatKey, !isRoot, returnInfo);
             }
         }
     }
@@ -305,7 +287,7 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
         // Load append item formats.
         AppendItemFormatsSink appendItemFormatsSink = new AppendItemFormatsSink();
         try {
-            rb.getAllItemsWithFallback(
+            rb.getAllChildrenWithFallback(
                     "calendar/" + calendarTypeToUse + "/appendItems",
                     appendItemFormatsSink);
         }catch(MissingResourceException e) {
@@ -314,7 +296,7 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
         // Load CLDR item names.
         AppendItemNamesSink appendItemNamesSink = new AppendItemNamesSink();
         try {
-            rb.getAllItemsWithFallback(
+            rb.getAllChildrenWithFallback(
                     "fields",
                     appendItemNamesSink);
         }catch(MissingResourceException e) {
@@ -323,7 +305,7 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
         // Load the available formats from CLDR.
         AvailableFormatsSink availableFormatsSink = new AvailableFormatsSink(returnInfo);
         try {
-            rb.getAllItemsWithFallback(
+            rb.getAllChildrenWithFallback(
                     "calendar/" + calendarTypeToUse + "/availableFormats",
                     availableFormatsSink);
         } catch (MissingResourceException e) {
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -934,15 +934,11 @@ struct DateTimePatternGenerator::AppendItemFormatsSink : public ResourceSink {
 
     virtual void put(const char *key, ResourceValue &value, UBool /*noFallback*/,
             UErrorCode &errorCode) {
-        ResourceTable itemsTable = value.getTable(errorCode);
-        if (U_FAILURE(errorCode)) { return; }
-        for (int32_t i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
-            UDateTimePatternField field = dtpg.getAppendFormatNumber(key);
-            if (field == UDATPG_FIELD_COUNT) { continue; }
-            const UnicodeString& valueStr = value.getUnicodeString(errorCode);
-            if (dtpg.getAppendItemFormat(field).isEmpty() && !valueStr.isEmpty()) {
-                dtpg.setAppendItemFormat(field, valueStr);
-            }
+        UDateTimePatternField field = dtpg.getAppendFormatNumber(key);
+        if (field == UDATPG_FIELD_COUNT) { return; }
+        const UnicodeString& valueStr = value.getUnicodeString(errorCode);
+        if (dtpg.getAppendItemFormat(field).isEmpty() && !valueStr.isEmpty()) {
+            dtpg.setAppendItemFormat(field, valueStr);
         }
     }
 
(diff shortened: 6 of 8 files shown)


jefgen:
FYI, it looks like the [Valgrind ](https://dev.azure.com/ms/icu/_build/results?buildId=204183&view=logs&j=222db123-c128-52c1-3783-5c13cc4f3c2a&t=c3201fd5-4420-5c6e-55a3-f2e9171b135d) and Clang-Asan builds are complaining about memory leaks.

From the [Clang-Asan logs](
https://github.com/unicode-org/icu/pull/1777/checks?check_run_id=3164528973#step:3:10670):
```
Direct leak of 2386200 byte(s) in 11931 object(s) allocated from:
    #0 0x514bed in malloc (/home/runner/work/icu/icu/icu4c/source/test/intltest/intltest+0x514bed)
    #1 0x7fc29f1e174a in uprv_malloc /home/runner/work/icu/icu/icu4c/source/common/cmemory.cpp:58:20
    #2 0x7fc29f592773 in init_resb_result(ResourceData const*, unsigned int, char const*, int, UResourceDataEntry*, UResourceBundle const*, int, UResourceBundle*, UErrorCode*) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:1193:35
    #3 0x7fc29f5969e0 in ures_getByKeyWithFallback /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:1935:26
    #4 0x7fc29f5a8ea6 in GetAllChildrenSink::put(char const*, icu::ResourceValue&, signed char, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2098:48
    #5 0x7fc29f5984cf in (anonymous namespace)::getAllItemsWithFallback(UResourceBundle const*, icu::ResourceDataValue&, icu::ResourceSink&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:1971:10
    #6 0x7fc29f598bdb in (anonymous namespace)::getAllItemsWithFallback(UResourceBundle const*, icu::ResourceDataValue&, icu::ResourceSink&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2006:13
    #7 0x7fc29f597d8c in ures_getAllItemsWithFallback /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2068:5
    #8 0x7fc29f598e96 in ures_getAllChildrenWithFallback /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2119:5
    #9 0x7fc29fbac455 in icu::DateTimePatternGenerator::addCLDRData(icu::Locale const&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:1066:5
    #10 0x7fc29fba7a84 in icu::DateTimePatternGenerator::initData(icu::Locale const&, UErrorCode&, signed char) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:511:5
    #11 0x7fc29fba78ab in icu::DateTimePatternGenerator::DateTimePatternGenerator(icu::Locale const&, UErrorCode&, signed char) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:363:9
    #12 0x7fc29fba5bd5 in icu::DateTimePatternGenerator::createInstance(icu::Locale const&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:310:17
    #13 0x7daad9 in IntlTestDateTimePatternGeneratorAPI::testJjMapping() /home/runner/work/icu/icu/icu4c/source/test/intltest/dtptngts.cpp:1363:53
    #14 0x7c3324 in IntlTestDateTimePatternGeneratorAPI::runIndexedTest(int, signed char, char const*&, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/dtptngts.cpp:43:9
    #15 0x7ef9bd in IntlTest::runTestLoop(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:791:19
    #16 0x7eea36 in IntlTest::runTest(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:712:14
    #17 0x7edfc9 in IntlTest::callTest(IntlTest&, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:613:35
    #18 0x81b5ae in IntlTestFormat::runIndexedTest(int, signed char, char const*&, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/itformat.cpp:145:9
    #19 0x7ef9bd in IntlTest::runTestLoop(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:791:19
    #20 0x7eea36 in IntlTest::runTest(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:712:14
    #21 0x7edfc9 in IntlTest::callTest(IntlTest&, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:613:35
    #22 0x82637e in MajorTestLevel::runIndexedTest(int, signed char, char const*&, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/itmajor.cpp:98:21
    #23 0x7ef9bd in IntlTest::runTestLoop(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:791:19
    #24 0x7eea36 in IntlTest::runTest(char*, char*, char*) /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:712:14
    #25 0x7f906e in main /home/runner/work/icu/icu/icu4c/source/test/intltest/intltest.cpp:1498:15
    #26 0x7fc29ce990b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
```

richgillam:
>FYI, it looks like the Valgrind and Clang-Asan builds are complaining about memory leaks.

Ugh.

I'm pretty sure this is either the same problem as is holding up ICU-21202 or fairly closely related (I'd copied the ICU-21202 fix into here for testing, with the expectation that that fix would land first).  Unfortunately, that's a complicated problem that's going to take me a while to untangle.  I'm going to try to fix the ICU-21202 problem first and then see if it also takes care of this one.  I'll have to sit on this change for a day or two until I can get that figured out (I also have to stop to do my "real job" for a day or two).

Apologies for the delay...

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -213,40 +213,25 @@ private void consumeShortTimePattern(String shortTimePattern, PatternInfo return
...
         public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
-            UResource.Table itemsTable = value.getTable();
-            for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
-                int field = getAppendFormatNumber(key);
-                assert field != -1;
-                if (getAppendItemFormat(field) == null) {
-                    setAppendItemFormat(field, value.toString());
-                }
+            int field = getAppendFormatNumber(key);
+            assert field != -1;
```
markusicu, line 217: Pre-existing issue, but still...

The C++ code skips/ignores unknown keys, while this Java code here asserts that the key is known. Which means that when assertions are disabled (which is the default), then we ram forward anyway; unclear if the next line compensates for that. And when assertions are enabled, then we crash and burn.

We probably want to skip like in C++ (and like in the `put()` below): `if (field < 0) { return; }`

(We could make it strict and fail, but I generally prefer data readers to be lenient with additional data, as a way of being somewhat forward compatible.)

markusicu (commented):
Nice moving a common loop into an internal API. Comments...

Caveat: I have not yet caught up with your various emails, ticket comments, and your next PR. Feel free to let me know where I am yet ignorant ;-)

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
     getAllItemsWithFallback(rb, value, sink, errorCode);
 }
 
+struct GetAllChildrenSink : public ResourceSink {
```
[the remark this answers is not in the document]
richgillam: Done.

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
...
+    
+    // Original containing resource bundle
+    const UResourceBundle* containingBundle;
+    
+    // Table resource's key path
+    const char* tableKeyPath;
+    
+    GetAllChildrenSink(ResourceSink& dest, const UResourceBundle* containingBundle, const char* tableKeyPath)
+        : dest(dest), containingBundle(containingBundle), tableKeyPath(tableKeyPath) {}
+    virtual ~GetAllChildrenSink();
```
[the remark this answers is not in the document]
richgillam: Done.  In my defense, I copied this class from the resource sinks in DateTimePatternGenerator, which did neither of these things (anonymous namespace and "override").  Since I'm messing with those sinks here, should I go back and make these changes to them too?

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
...
+
+    virtual void put(const char *key, ResourceValue &value, UBool isRoot,
```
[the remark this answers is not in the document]
richgillam: Done.

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
...
+                // if the current entry in the table is an alias, re-fetch it using ures_getByKeyWithFallback():
+                // this will follow the alias (and any aliases it points to) and bring back the real value
+                CharString path;
+                path.append(tableKeyPath, errorCode);
+                path.append('/', errorCode);
+                path.append(key, errorCode);
+                
+                UResourceBundle child;
+                ures_initStackObject(&child);
+                ures_getByKeyWithFallback(containingBundle, path.data(), &child, &errorCode);
```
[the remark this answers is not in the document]
richgillam: >This seems inefficient: It will look for, find, and read this value’s same alias resource, and then follow the alias path; but we have already gotten this far.

I agree, but I wonder how inefficient is it in practice?  I did it this way (and did some similar things in ICU-20400) because it was a lot easier, I knew it would do the right thing, and there didn't appear to be a ready-made API that just followed the alias, but I swear I also found several precedents in the code already that do the same kind of thing, leading me to believe it wasn't _that_ inefficient, or that this wouldn't happen often enough to cause meaningful performance problems.

Another thing that swayed me this direction was that the alias might _itself_ point to an alias, which would _also_ need to be followed, and the code to do that got hairy in a hurry.  But then there are your suggestions below, which would address that...

>Also, I wonder if we should hoist this part of the logic into its own internal-public helper function...

I thought about this.  It probably _is_ a good idea, but I didn't pursue it both out of a desire to get something done and out of concern that there was a "good reason" why we didn't already have something like that.

>Finally, your changes here raise the question of whether we should ever return a ResourceValue with type=URES_ALIAS. Should we always follow aliases in all ures_getAllXyz() and ResourceTable.get/find...Value()?

That's an interesting question.  It makes intuitive sense to me, but I wonder if there are any client frameworks that actually care about alias resources and are doing their own thing?  Even if so, having two sets of APIs (or an extra parameter) seems like a reasonable approach.

I think everything you're suggesting here is reasonable, but that I need to think about it and experiment a bit before committing one way or the other.  I'm going to address all of your smaller comments first and then come back to this.  I'll either come to you to discuss this more or I'll put up another PR once I've thought these issues through.

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
...
+                    ResourceDataValue rdValue;
+                    rdValue.setData(&child.fResData);
+                    rdValue.setResource(child.fRes, ResourceTracer(&child));
+                    dest.put(key, rdValue, isRoot, errorCode);
+                    ures_close(&child);
+                }
+                // if following the alias didn't work, don't send _anything_ down to the destination sink
+            } else {
+                dest.put(key, value, isRoot, errorCode);
+            }
```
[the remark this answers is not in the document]
richgillam: Hmmm...

If I do that, we dump out of the loop the first time we either fail to follow an alias or get an error back from the client sink.  Is that really what we want to do, or do we want to try to call the sink with as many table elements as we can?  Dumping out of the loop early led to the symptom that prompted me to make these changes.

Of course, if the right answer is to try every element, the code I have still isn't correct-- I'd have to reset the error code at the top of the loop, or subsequent iterations are just wasting time.

I'm inclined to say resetting the error code at the top of the loop (and thus visiting all the items in the table) is preferable to breaking out (and _not_ visiting all the items).  Do you disagree?

icu4c/source/common/uresimp.h
```diff
@@ -291,10 +291,38 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
                           UResourceBundle *tempFillIn,
                           icu::ResourceDataValue &value, UErrorCode &errorCode);
 
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and calls the specified sink's `put()` method with that
+ * bundle.  Then walks that resource's parent chain, calling `put()` on the sink for each item in the
+ * parent chain.
```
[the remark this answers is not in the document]
richgillam: Fixed.

icu4c/source/common/uresimp.h
```diff
@@ -291,10 +291,38 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
...
 ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
                              icu::ResourceSink &sink, UErrorCode &errorCode);
 
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and, if the resource is a table resource, iterates over its
+ * child resources (again, following any aliases to get the individual resource values), and calls the specified
+ * sink's `put()` method for each child resource (passing it that resource's key and either its actual value or,
+ * if that value is an alias, the value you get by following the alias).  Then walks back over the table resource's
+ * parent chain, similarly iterating over each parent resource's child resources.
```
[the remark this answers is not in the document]
richgillam: Fixed.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -395,6 +395,13 @@ public void getAllItemsWithFallbackNoFail(String path, UResource.Sink sink) {
         }
     }
 
+    /**
+     * Locates the resource specified by `path` in this resource bundle (performing any necessary fallback and
```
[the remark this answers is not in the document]
richgillam: Made the same changes to the doc comments in the Java code.

icu4c/source/common/uresbund.cpp
```diff
@@ -2068,6 +2079,62 @@ ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
...
+        for (int32_t i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
+            if (value.getType() == URES_ALIAS) {
+                // if the current entry in the table is an alias, re-fetch it using ures_getByKeyWithFallback():
+                // this will follow the alias (and any aliases it points to) and bring back the real value
+                CharString path;
+                path.append(tableKeyPath, errorCode);
+                path.append('/', errorCode);
+                path.append(key, errorCode);
+                
+                UResourceBundle child;
```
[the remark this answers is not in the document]
richgillam: As far as I can make out, `ures_getByKeyWithFallback()` _is_ the C equivalent of `getValueWithFallback()` in Java.

icu4j/main/classes/core/src/com/ibm/icu/text/DateTimePatternGenerator.java
richgillam: I agree.  I replaced the assert with `if (field < 0) { return; }`.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: I am surprised that the compiler lets you say both `virtual` and `override`... if it works, it's ok.

I might be good to use `override` in the sub-sinks as well, but for now I am happy to get it right in the "API".

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: We can punt this discussion to another ticket or another PR.

As for how inefficient, it's an obvious duplication of work for a single resource item, so if there are a lot of aliases then it could be noticeable. More generally, if there are obvious duplications, I try to get rid of them :-) (Same for unnecessary memory allocations+initializations, such as the UResourceBundle that we are throwing away in the end.)

For my first suggestion here, I was thinking only of removing the first layer, and calling some function that takes the alias URes value and follows it, not copying the whole code that may recurse.

richgillam:
>This seems inefficient: It will look for, find, and read this value’s same alias resource, and then follow the alias path; but we have already gotten this far.
>
>It should be better to read the value’s alias path string here and follow that more directly. You should be able to resuse/call code from inside ures_getByKeyWithFallback() where it does this.
>
>Also, I wonder if we should hoist this part of the logic into its own internal-public helper function, maybe even on the ResourceValue class itself, for example like ResourceValue getAliasValue(UErrorCode &errorCode) const or self-modifying, non-const (and more efficient in Java / not allocating a new Value) void resolveAlias(UErrorCode &errorCode)? Might need more parameters. Might return a copy of itself (or be a no-op in the self-modify version) if this is not a URES_ALIAS, so that we could call it unconditionally.

I've been fooling around with this for several days now, and I've put up my best attempt to address these issues.  It's been an absolute nightmare.  Maybe there's a better way to do this than what I have here, but I swear I've tried everything and have been thwarted at every turn.

One of the biggest problems is that `ResourceValue` is an abstract interface, with all the implementation in `ResourceDataValue`, but `ResourceDataValue` is an internal class that not a lot of modules have access to.  `ResourceTable` doesn't have access to the `ResourceDataValue` that owns it.  Neither `ResourceDataValue` nor `ResourceTable` have access to many of the values I need to follow an alias-- those values are on `UResourceBundle` but not propagated down into `ResourceDataValue` or `ResourceTable`.  In fact, neither class has access to the function I need to call to follow the alias.  As if all that wasn't bad enough, it turns out that `ResourceTable::getKeyAndValue()`'s `value` parameter is _not_ an output-only parameter; it's an _input/output_ parameter-- following the alias and changing the value in place to be the aliased value messes up iteration through the table.  This actually seems bad independent of what I'm doing, as the behavior of `getKeyAndValue()` depends both on what's passed into it and on the lifetime of that object.

Adding a method to `ResourceValue` to just return the aliased value (or modify the object in place with the aliased value) is the most straightforward thing to do from an API point of view, but `ResourceValue`/`ResourceDataValue` is missing most of the internal values it needs to make that work-- the caller to the "follow alias" function would have to supply them, if _it_ had access to them.  I didn't get very far going this way.

Instead, I ended up adding a method to `ResourceTable` that would follow the alias.  This meant adding a new data member to `ResourceTable`, a pointer to `ResourceData`, but since `ResourceData` is an internal class that's not visible from `resource.h`, where `ResourceTable` is declared, I ended up making it a heap-allocated object, with all the attendant problems that brings.  But it does let me turn the `value` parameter to `getKeyAndValue()` into a true output-only parameter.  I could fix the ugliness here either by making `ResourceTable` abstract and introducing a separate implementation class under the covers, but then I'm adding table dispatch in a spot where we didn't need it before, or I could fix it by just including `uresdata.h` from `resource.h`, but that felt like an encapsulation violation.

The actual alias-following logic couldn't be in `ResourceTable` because it required calling `init_resb_result()`, which is internal to `uresbund.cpp`, so I added a new bridging method called `ures_followAlias()`.  This might be useful in other spots, but maybe not.

I haven't yet made any changes on the Java side-- the approach I took on the C++ isn't looking terribly feasible in Java and I haven't yet come up with an alternative that'll work better.  The changes I made on the C++ side change the behavior from what I had before, although the change seemed better to me and might have no practical significance.  But it does mean that the C++ code and the Java code actually do different things right now, which is bad.

Maybe we need to talk...

>Finally, your changes here raise the question of whether we should ever return a ResourceValue with type=URES_ALIAS. Should we always follow aliases in all ures_getAllXyz() and ResourceTable.get/find...Value()?

Yeah, there are too many spots where client code is actually examining the alias path to get what they need, rather than fetching the aliased resource itself.  We don't want to mess with this.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: The UErrorCode thing is a poor-man's version of exception handling. Unless you really want to catch an exception and deal with it, you should bail out. We can defer this handling if there are more calls that take UErrorCode and become no-ops, but that does not look to be the case here.

So unless there is a reason to "catch the exception" (and maybe the Java code does so), we should pass it through. In C/C++ code that sometimes means adding such checks.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: Except one returns a UResourceBundle and the other a ResourceValue.
But it looks like you replaced this code in the meantime, so this comment looks obsolete.

markusicu:
> I've been fooling around with this for several days now, and I've put up my best attempt to address these issues. It's been an absolute nightmare. Maybe there's a better way to do this than what I have here, but I swear I've tried everything and have been thwarted at every turn.

I know this is hairy, but what you write here sounds worse than I thought. Let me take a look at your changes.
I am sure that in some places we need to assume that a ResourceValue is actually a ResourceDataValue, but I don't like going beyond that and adding more state etc.
I need to get my head back deeper into this stuff...

> Maybe we need to talk...

Yeah. Let me look at this more. What's a good way to schedule a time slot?

richgillam:
>I know this is hairy, but what you write here sounds worse than I thought. Let me take a look at your changes.

I'm grateful for your advice and feedback.  You know this code and the goals and assumptions behind it way better than I do.

>I am sure that in some places we need to assume that a ResourceValue is actually a ResourceDataValue, but I don't like going beyond that and adding more state etc.

Either I'm missing something obvious, or I didn't really understand what you were suggesting, or this stuff is chopped up too finely to allow for a simple solution.  I don't like adding more state either, but if there's a better way, I couldn't find it.

>Yeah. Let me look at this more. What's a good way to schedule a time slot?

Send me an email or a Google Calendar invite, I guess.  I'm free all afternoon tomorrow.

I let myself spin in circles far longer than I should have and now I'm starting to panic about this stuff, so I'm really hoping we can get to some kind of resolution within the next couple of days.

icu4c/source/common/uresdata.cpp
```diff
@@ -817,10 +817,41 @@ res_getResource(const ResourceData *pResData, const char *key) {
...
+    this->tableData = new ResourceData(*that.tableData);
+
+    return *this;
+}
+
 UBool icu::ResourceTable::getKeyAndValue(int32_t i,
                                          const char *&key, icu::ResourceValue &value) const {
     if(0 <= i && i < length) {
         icu::ResourceDataValue &rdValue = static_cast<icu::ResourceDataValue &>(value);
+        rdValue.setData(tableData); // `value` is an output-only parameter, but these macros expect it to point to `tableData`
```
markusicu, line 854: I think I remember how this works so far: While iterating through resources in a bundle the ResourceData is always the same. So it never changes, and that's how we haven't gotten into trouble with these loops over getKeyAndValue(). (`resData` could be const if we didn't need to be able to default-construct ResourceDataValue.)

The *resource* (`res`) changes (.setResource()), but that's harmless because before the loop-over-table-items started we copied the table's own resource into its ResourceTable and already used it to fetch the pointers into the key and item arrays.

The problem is that an alias can point to a resource in a different bundle/file, and thus needs a different ResourceData pointer when that happens.

It seems like the solution would be for getKeyAndRealValue() to *return a new* ResourceValue, so that it does not overwrite the input value's data pointer. With that we would not need to add this tableData to ResourceTable.

New problem: ResourceValue is a base interface, so we can't return it by value (and we forbid copying). We could take an output reference, a new one separate from the input value, so that we don't clobber that one's resData, but then callers need to know about ResourceDataValue. Maybe that's acceptable until we figure out a better structure? Or, might as well return a ResourceDataValue by value???

Also: If the alias lives in another bundle, then we may also need to think about the lifetime of that bundle. Either we rely on the resource bundle cache, and convince ourselves that that's ok, or we add an abstract bundle pointer into the ResourceDataValue (yuck), probably with some mini interface that lets us ures_close() the bundle (yuck) without knowing all about UResourceBundle at that level.

----
Crazy idea: We might be able to make aliases completely transparent in the ResourceValue API, while avoiding many of these problems. When we get a value, don't immediately follow the alias. Only when we getString(), getTable(), getBinary(), etc. -- then we follow an alias if there is one, before we really do fetch the string/table/binary/whatever. Those functions already take a UErrorCode, and most of them don't have to return a ResourceValue nor a UResourceBundle. And, by following the alias only when actually accessing the value contents, we avoid wasting work when the caller looks at the key and ignores the value.

Now, we still have the problem with getTable() and getArray() that you are trying to solve here: While their lookup works with just pointers into the key and item arrays, for returning their items we will need to either switch the output value's resData or switch to a new ResourceValue.

FYI, schlepping around the ResourceData would be much simpler if we could make https://unicode-org.atlassian.net/browse/ICU-20769 work; in particular, we could carry an opaque pointer in places that don't need to know the details.

We would also still have to think about the lifetime of an alias-hosting bundle, but again maybe the bundle cache solves that.

sffc:
Please also heed the CI test failures.  clang-asan reveals that you may have done something that breaks the data tracing mechanism (the one documented at https://unicode-org.github.io/icu/userguide/icu_data/tracing.html).  To debug, re-run `./runConfigureICU` with the `--enable-tracing` flag, which has the effect of setting `U_ENABLE_TRACING` to true.

```
=================================================================
==18803==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffd4afd5888 at pc 0x7ff7fb8ef3be bp 0x7ffd4afd50b0 sp 0x7ffd4afd50a8
READ of size 8 at 0x7ffd4afd5888 thread T0
    #0 0x7ff7fb8ef3bd in icu::ResourceTracer::getFilePath(icu::CharString&, UErrorCode&) const /home/runner/work/icu/icu/icu4c/source/common/restrace.cpp:60:37
    #1 0x7ff7fb8ef70d in icu::ResourceTracer::getFilePath(icu::CharString&, UErrorCode&) const /home/runner/work/icu/icu/icu4c/source/common/restrace.cpp:65:18
    #2 0x7ff7fb8eec8a in icu::ResourceTracer::trace(char const*) const /home/runner/work/icu/icu/icu4c/source/common/restrace.cpp:26:5
    #3 0x7ff7fbb76816 in icu::res_getString(icu::ResourceTracer const&, ResourceData const*, unsigned int, int*) /home/runner/work/icu/icu/icu4c/source/common/./uresdata.h:485:15
    #4 0x7ff7fbb798a4 in icu::ResourceDataValue::getString(int&, UErrorCode&) const /home/runner/work/icu/icu/icu4c/source/common/uresdata.cpp:512:22
    #5 0x7ff7fc09c4ad in icu::ResourceValue::getUnicodeString(UErrorCode&) const /home/runner/work/icu/icu/icu4c/source/i18n/../common/resource.h:161:26
    #6 0x7ff7fc19657c in icu::DateTimePatternGenerator::AppendItemNamesSink::put(char const*, icu::ResourceValue&, signed char, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:972:47
    #7 0x7ff7fbb6f57e in (anonymous namespace)::GetAllChildrenSink::put(char const*, icu::ResourceValue&, signed char, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2126:18
    #8 0x7ff7fbb6621f in (anonymous namespace)::getAllItemsWithFallback(UResourceBundle const*, icu::ResourceDataValue&, icu::ResourceSink&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:1982:10
    #9 0x7ff7fbb6692b in (anonymous namespace)::getAllItemsWithFallback(UResourceBundle const*, icu::ResourceDataValue&, icu::ResourceSink&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2017:13
    #10 0x7ff7fbb65adc in ures_getAllItemsWithFallback /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2098:5
    #11 0x7ff7fbb66b95 in ures_getAllChildrenWithFallback /home/runner/work/icu/icu/icu4c/source/common/uresbund.cpp:2140:5
    #12 0x7ff7fc17b8c5 in icu::DateTimePatternGenerator::addCLDRData(icu::Locale const&, UErrorCode&) /home/runner/work/icu/icu/icu4c/source/i18n/dtptngen.cpp:1066:5
...
```

sffc:
Note: I see that the commit you had before force-pushing (085f2241e19c43d93b9618be7c801e5c8dcd7a81) passed the data tracing test.  So something you changed between that commit and now has broken the data tracing.

icu4c/source/common/uresdata.h
```diff
@@ -512,20 +514,27 @@ class ResourceDataValue : public ResourceValue {
...
+    }
 
     void setResource(Resource r, ResourceTracer&& traceInfo) {
         res = r;
         fTraceInfo = traceInfo;
     }
 
     const ResourceData &getData() const { return *pResData; }
+    UResourceDataEntry *getValidLocaleDataEntry() const { return validLocaleDataEntry; }
+    Resource getRes() const { return res; }
```
markusicu, line 537:
```suggestion
    Resource getResource() const { return res; }
```

Let's make this a proper pair with setResource().

icu4c/source/common/uresdata.h
```diff
@@ -512,20 +514,27 @@ class ResourceDataValue : public ResourceValue {
...
+        validLocaleDataEntry = entry;
+    }
 
     void setResource(Resource r, ResourceTracer&& traceInfo) {
         res = r;
         fTraceInfo = traceInfo;
     }
 
     const ResourceData &getData() const { return *pResData; }
+    UResourceDataEntry *getValidLocaleDataEntry() const { return validLocaleDataEntry; }
```
markusicu, line 536: Question: Should this be const? When we return a non-const pointer, there is a logical conflict with making the getter const, but maybe it needs to be const so that call sites work?

(And the data entry pointer must be non-const so that init_resb_result() can entryIncrease() on it.)

icu4c/source/common/uresbund.cpp
```diff
@@ -2055,6 +2056,47 @@ void getAllItemsWithFallback(
...
+    virtual void put(const char *key, ResourceValue &value, UBool isRoot,
+           UErrorCode &errorCode) override {
+        ResourceTable itemsTable = value.getTable(errorCode);
+        if (U_FAILURE(errorCode)) { return; }
+        for (int32_t i = 0; i < itemsTable.getSize(); ++i) {
+            itemsTable.getKeyAndValue(i, key, value);
+            if (value.getType() == URES_ALIAS) {
+                ResourceDataValue& rdv = static_cast<ResourceDataValue&>(value);
+                StackUResourceBundle stackTempBundle;
+                UResourceBundle* aliasRB = getAliasTargetAsResourceBundle(rdv.getData(), rdv.getRes(), key, -1, rdv.getValidLocaleDataEntry(), nullptr, 0, stackTempBundle.getAlias(), &errorCode);
```
markusicu, line 2075:
```suggestion
                UResourceBundle* aliasRB = getAliasTargetAsResourceBundle(
                    rdv.getData(), rdv.getResource(), nullptr, -1, rdv.getValidLocaleDataEntry(), nullptr,
                    0, stackTempBundle.getAlias(), &errorCode);
```

1. line breaks
2. let's pass in key=nullptr since it does not make sense to give the real key when we don't have a containerResPath

icu4c/source/common/uresbund.cpp
```diff
@@ -2055,6 +2056,47 @@ void getAllItemsWithFallback(
...
+                if (U_SUCCESS(errorCode)) {
+                    ResourceDataValue aliasedValue;
+                    aliasedValue.setData(aliasRB->getResData());
+                    aliasedValue.setValidLocaleDataEntry(aliasRB->fValidLocaleDataEntry);
+                    aliasedValue.setResource(aliasRB->fRes, ResourceTracer(aliasRB));
+                    dest.put(key, aliasedValue, isRoot, errorCode);
+                }
+            } else {
+                dest.put(key, value, isRoot, errorCode);
+            }
```
markusicu, line 2085: after this line please
```
if (U_FAILURE(errorCode)) { return; }
```

icu4c/source/common/uresimp.h
```diff
@@ -294,10 +294,39 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
                           UResourceBundle *tempFillIn,
                           icu::ResourceDataValue &value, UErrorCode &errorCode);
 
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and calls the specified `sink``'s `put()` method with that
+ * resource.  Then walks the bundle's parent chain, calling `put()` on the sink for each item in the
+ * parent chain.
```
markusicu, line 301:
```suggestion
 * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
 * necessary fallback and following any aliases) and calls the specified `sink`'s `put()` method with that
 * resource.  Then walks the bundle's parent chain, calling `put()` on the sink for each item in the
 * parent chain.
```

icu4c/source/common/uresimp.h
```diff
@@ -294,10 +294,39 @@ ures_getValueWithFallback(const UResourceBundle *bundle, const char *path,
...
+ */
 U_CAPI void U_EXPORT2
 ures_getAllItemsWithFallback(const UResourceBundle *bundle, const char *path,
                              icu::ResourceSink &sink, UErrorCode &errorCode);
 
+/**
+ * Locates the resource specified by `path` in the resource bundle specified by `bundle` (performing any
+ * necessary fallback and following any aliases) and, if the resource is a table resource, iterates over its
+ * immediate child resources (again, following any aliases to get the individual resource values), and calls the specified
+ * `sink``'s `put()` method for each child resource (passing it that resource's key and either its actual value or,
```
markusicu, line 315:
```suggestion
 * `sink`'s `put()` method for each child resource (passing it that resource's key and either its actual value or,
```

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -395,6 +395,13 @@ public void getAllItemsWithFallbackNoFail(String path, UResource.Sink sink) {
...
+     * following any aliases) and calls the specified `sink`'s `put()` method with that resource.  Then walks the
+     * bundle's parent chain, calling `put()` on the sink for each item in the parent chain.
+     * @param path The path of the desired resource
+     * @param sink A `ResourceSink` that gets called for each resource in the parent chain
```
markusicu, line 403:
```suggestion
     * @param sink A `UResource.Sink` that gets called for each resource in the parent chain
```

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -421,6 +428,45 @@ public void getAllItemsWithFallback(String path, UResource.Sink sink)
...
+    /**
+     * Locates the resource specified by `path` in this resource bundle (performing any necessary fallback and
+     * following any aliases) and, if the resource is a table resource, iterates over its immediate child resources (again,
+     * following any aliases to get the individual resource values), and calls the specified `sink`'s `put()` method
+     * for each child resource (passing it that resource's key and either its actual value or, if that value is an
+     * alias, the value you get by following the alias).  Then walks back over the bundle's parent chain,
+     * similarly iterating over each parent table resource's child resources.
+     * Does not descend beyond one level of table children.
+     * @param path The path of the desired resource
+     * @param sink A `ResourceSink` that gets called for each child resource of the specified resource (and each child
```
markusicu, line 440:
```suggestion
     * @param sink A `UResource.Sink` that gets called for each child resource of the specified resource (and each child
```

icu4c/source/common/uresbund.cpp
```diff
@@ -2055,6 +2056,47 @@ void getAllItemsWithFallback(
...
+
+    GetAllChildrenSink(ResourceSink& dest)
+        : dest(dest) {}
+    virtual ~GetAllChildrenSink() override;
+    virtual void put(const char *key, ResourceValue &value, UBool isRoot,
+           UErrorCode &errorCode) override {
+        ResourceTable itemsTable = value.getTable(errorCode);
+        if (U_FAILURE(errorCode)) { return; }
+        for (int32_t i = 0; i < itemsTable.getSize(); ++i) {
+            itemsTable.getKeyAndValue(i, key, value);
```
markusicu, line 2071:
```suggestion
        for (int32_t i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
```

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -421,6 +428,45 @@ public void getAllItemsWithFallback(String path, UResource.Sink sink)
...
+        class AllChildrenSink extends UResource.Sink {
+            @Override
+            public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
+                UResource.Table itemsTable = value.getTable();
+                for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
+                    if (value.getType() == ALIAS) {
+                        // if the current entry in the table is an alias, re-fetch it using getValueWithFallback():
+                        // this will follow the alias (and any aliases it points to) and bring back the real value
+                        String aliasPath = value.getAliasString();
+                        ICUResourceBundle aliasedResource = getAliasedResource(aliasPath, ICUResourceBundle.this.wholeBundle.loader, "", null, 0, new String[] { key.toString() }, null, ICUResourceBundle.this);
```
markusicu, line 454: - Long line; please keep it to roughly 100 chars.
- Again, we don't have a container key path. Does it work if you pass in `null` instead of the `new String[] ...` ?
- Over-qualified member access
  - `ICUResourceBundle.this.wholeBundle.loader` --> `wholeBundle.loader`
  - `ICUResourceBundle.this` --> `this`

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -421,6 +428,45 @@ public void getAllItemsWithFallback(String path, UResource.Sink sink)
...
+     */
+    public void getAllChildrenWithFallback(final String path, final UResource.Sink sink)
+            throws MissingResourceException {
+        class AllChildrenSink extends UResource.Sink {
+            @Override
+            public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
+                UResource.Table itemsTable = value.getTable();
+                for (int i = 0; itemsTable.getKeyAndValue(i, key, value); ++i) {
+                    if (value.getType() == ALIAS) {
+                        // if the current entry in the table is an alias, re-fetch it using getValueWithFallback():
```
markusicu, line 451:
```suggestion
                        // if the current entry in the table is an alias, re-fetch it using getAliasedResource():
```

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1481,10 +1527,24 @@ protected static ICUResourceBundle getAliasedResource(
             UResourceBundle requested) {
         WholeBundle wholeBundle = base.wholeBundle;
         ClassLoader loaderToUse = wholeBundle.loader;
+        String rpath = wholeBundle.reader.getAlias(_resource);
+        String baseName = wholeBundle.baseName;
+
+        int baseDepth = base.getResDepth();
+        String[] baseKeyPath = new String[baseDepth + 1];
+        base.getResPathKeys(baseKeyPath, baseDepth);
```
markusicu, line 1535: Question: Is it ok to ask for one more path key than the base has?

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1481,10 +1527,24 @@ protected static ICUResourceBundle getAliasedResource(
             UResourceBundle requested) {
         WholeBundle wholeBundle = base.wholeBundle;
         ClassLoader loaderToUse = wholeBundle.loader;
+        String rpath = wholeBundle.reader.getAlias(_resource);
+        String baseName = wholeBundle.baseName;
+
+        int baseDepth = base.getResDepth();
```
markusicu, line 1533: Please insert before this line:
```
// TODO: We need not build the baseKeyPath array if the rpath includes a keyPath
// (except for an exception message string).
// Try to avoid unnecessary work+allocation.
```

icu4c/source/common/uresdata.h
richgillam: I started with it const and changed it because `getAliasTargetAsResourceBundle()` wanted it as a non-const and casting away the const didn't seem like a good idea.  Changing the parameter to `getAliasTargetAsResourceBundle()` to be const too just pushes the problem back to `init_resb_result()`, which is also expecting a non-const pointer.

>(And the data entry pointer must be non-const so that init_resb_result() can entryIncrease() on it.)

Not sure what you're saying here.  As I read this, that _is_ the reason why I didn't declare this member variable to be const.  Is there a better solution?

icu4c/source/common/uresbund.cpp
richgillam: I'm making this change, but I have reservations.  I'm not sure the right answer to an error following an alias (or passing a value to the resource sink) should be to abort the whole iteration.

If the sink fails to accept the value, maybe that's a good reason to quit passing it stuff (the failure check wasn't in there before, but because we also don't clear the error code, we were getting that effect much more expensively, so maybe that's grounds for adding the failure check).

On the other hand, if we fail to follow the alias, I'd argue it's better to throw out that one value and keep iterating.  I think your argument was that we're getting the "wrong answer" in their case, and so it'd be better to be spectacularly wrong so that we have more of a chance of detecting that there's an error at all, since the error would be in our own resource bundles.  I understand the rationale here, but in real application code, it seems like the "do the best you can" approach would be less destructive.

As I said, I'm making the change anyway, but I'm not crazy about it.

icu4c/source/common/uresbund.cpp
richgillam: I dislike this style, but okay...

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: Shortening `ICUResourceBundle.this` to just `this` didn't compile, so I left that alone.  Shortening `ICUResourceBundle.this.wholeBundle.loader` to just `wholeBundle.loader` _did_ work, so I did change that one.

Passing `null` for `baseKeyPath` does seem to work and is parallel with the change I just made on the C++ side, so I'm making the change, although it worries me a little, maybe just because I don't understand this code well enough...

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: You mean pass `baseDepth + 1` to `getResPathKeys()`? No, that'll produce an `ArrayIndexOutOfBoundsException`.  I think the original code was appending a new item onto the end of the path in much the same way as I'm doing now.

icu4c/source/common/uresdata.h
markusicu: Sorry I wasn't clear. The returned pointer must be non-const. I meant to ask *whether the **function** can be non-const*. As in
```
UResourceDataEntry *getValidLocaleDataEntry() { return validLocaleDataEntry; }
```
(`const` removed *in the middle* of the line)

If it's inconvenient for this function to be non-const (e.g., you would then need a const_cast) then leave it.
This is actually not really important, so you could leave this one anyway.

icu4c/source/common/uresbund.cpp
markusicu: See my earlier reply (https://github.com/unicode-org/icu/pull/1777#discussion_r695280415) about UErrorCode emulating exceptions. If there was a good case for why we would catch and ignore an exception, then we would have a case for ignoring and resetting a failure code.

In this case it looks like you are right too, and the current version of the code actually merely becomes an *inefficient* no-op without the additional line.

icu4c/source/common/uresbund.cpp
markusicu: That's how I designed this API :-)
tnx

markusicu (approved):
I am ok with the changes now, open comments notwithstanding.
However, there was a GHA test failure; it didn't make sense to me so I restarted the GHA bots.
Please watch for the results.

markusicu:
Oh, and if the tests look ok please squash...!

markusicu:
Tests passed, please squash.

markusicu (approved):
lgtm tnx

PR #1780, opened 2021-07-28
ICU-21691: add semantic versioning support for OSGI

Buuhuu:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21691


yumaoka:
@Buuhuu Thank you for this PR. I read through your explanation and I understand the goals. This enhancement sounds good in general. I have one minor concern about package versioning. You suggested to start with 1.0.0 using the current ICU code as the baseline, and increment each package version when we add/remove/change APIs. Is my understanding correct? Practically, what is the benefit of managing API level at package level? In other words, if package export version uses the ICU4J API version (70 for next release) for all packages, I understand it might create redundant version numbers, but are there any issues with this. My concern is mostly extra task for release, and confusion which might come from these version numbers.

Buuhuu:
Thanks @yumaoka for your feedback and please apologise my late reply (I have been on vacation last week).

Your understanding is correct. For a detailed explanation about semantic versioning in OSGI please have a look at [1]. 

The biggest advantage of semantic versioning comes for API consumers, which are implementations that use but do not implement the APIs from the icu4j bundle. As long as the (stable) APIs - lets say in `com.ibm.icu.text` - do not change in an incompatible way, the major version of the exported package stays the same and matches the consumers import range. That means the consumer can use an updated bundle version with minor changes to the exported package at runtime and does not need to be rebuild against this version.

Versioning all exported packages individually makes sense to reduce the probability of an incompatible change that would require a major package version update. That means that if `com.ibm.icu.text` gets a major version update due to an incompatible change, consumers of `com.ibm.icu.math` are not affected by that and don't need to be rebuild.

I am not an expert in the development process of icu4j but I think you don't have to expect any issues/additional steps with the release. As said in the initial description, the baseline check guarantees the build passes only when the exported packages are exported in semantical correct versions which requires to update the versions already with a PR. Additionally the baseline check makes a recommendation which version to choose. However, I understand that drafting APIs in icu4j is an established process which may interfere with semantic versioning and so should be ignored. Managing the @BaselineIgnore annotations is indeed additional effort, but again not related to the release process as far as I understand it. 

I will add an update to the documentation to the PR as I just realised it is part of the same repo. 

1) https://docs.osgi.org/whitepaper/semantic-versioning/Semantic-Versioning-20190110.pdf

Buuhuu:
I updated the documentation.

I also annotated all internal symbols (excl. classes in `impl`) with `@BaselineIgnore("999.99.9")` and all draft symbols (again excl. `impl`) with `@BaselineIgnore("9.9.9")`. The both versions mean far far far in the future (999.99.9) and far in the future (9.9.9). They have to be set because `@BaselineIngore` ignores the symbol only as long as it is compared to a version smaller then the specified one (it is actually meant to be used to ignore API changes temporarily). 

While writing the docs I realised that changes to the package versions will happen rarely, meaning only when a draft API becomes stable and its `@BaselineIgnore` annotation is removed. However it looks like the evolution of different APIs is not linear. There are some that are still in draft since quite a while already while others got stable after a minimal period of time. That's the reason why we cannot, unfortunately just take next minor package version for `@BaselineIngore`.

PR #1786, opened 2021-08-04
ICU-20973 Update equality operators to support C++20

roubert:
https://unicode-org.atlassian.net/browse/ICU-20973

- Change all equality operator return types from `UBool` to `bool`.
- Manually resolve C++20 reversed argument order ambiguity.
- Use the Clang `-Wno-ambiguous-reversed-operator` flag.
- Update configure files from `configure.ac` using `autoreconf`.
- Update the Coding Guidelines with bool equality operators.

Manually tested with Clang 13 and GCC 11 with `-std=c++20`.

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -325,7 +325,7 @@ public:
    * false otherwise.
    * @stable ICU 2.0
    */
-  inline UBool operator== (const UnicodeString& text) const;
+  inline bool operator== (const UnicodeString& text) const;
 
   /**
    * Inequality operator. Performs only bitwise comparison.
@@ -334,7 +334,7 @@ public:
    * true otherwise.
    * @stable ICU 2.0
    */
-  inline UBool operator!= (const UnicodeString& text) const;
+  inline bool operator!= (const UnicodeString& text) const;
 
   /**
    * Greater than operator. Performs only bitwise comparison.
@@ -3946,7 +3946,7 @@ UnicodeString::doCompare(int32_t start,
   }
 }
 
-inline UBool
+inline bool
 UnicodeString::operator== (const UnicodeString& text) const
 {
   if(isBogus()) {
@@ -3957,7 +3957,7 @@ UnicodeString::operator== (const UnicodeString& text) const
   }
 }
 
-inline UBool
+inline bool
 UnicodeString::operator!= (const UnicodeString& text) const
 { return (! operator==(text)); }
 
--- a/icu4c/source/i18n/unicode/tmutfmt.h
+++ b/icu4c/source/i18n/unicode/tmutfmt.h
@@ -141,7 +141,7 @@ public:
      * @return      true if the given Format objects are not semantically equal.
      * @deprecated ICU 53
      */
-    UBool operator!=(const Format& other) const;
+    bool operator!=(const Format& other) const;
 
     /**
      * Set the locale used for formatting or parsing.
@@ -236,7 +236,7 @@ private:
     friend struct TimeUnitFormatReadSink;
 };
 
-inline UBool
+inline bool
 TimeUnitFormat::operator!=(const Format& other) const  {
     return !operator==(other);
 }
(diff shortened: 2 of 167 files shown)


icu4c/source/i18n/unicode/tmutfmt.h
```diff
@@ -141,7 +141,15 @@ class U_I18N_API TimeUnitFormat: public MeasureFormat {
...
-    UBool operator!=(const Format& other) const;
+    bool operator!=(const Format& other) const;
+
+#if defined(__cpp_impl_three_way_comparison) && \
+       __cpp_impl_three_way_comparison >= 201711
+    // Manually resolve C++20 reversed argument order ambiguity.
+    inline bool operator!=(const TimeUnitFormat& other) const {
+        return operator!=(static_cast<const Format&>(other));
+    }
+#endif
```
markusicu, line 152: Can you please explain the problem? Why is there ambiguity here but not for other `operator!=()` without this extra code?

Could we avoid it by conditionally (>= C++20) removing the `operator!=()` altogether?

If we need a new public API function like this, then we should add API comments, so that it does not show up in the docs as untagged / with unknown API status.

roubert: The compiler error message with C++20 without this workaround is:

```
icu4c/source/test/intltest/tufmtts.cpp:261:59: error: use of overloaded operator '!=' is ambiguous (with operand types 'icu_70::TimeUnitFormat' and 'icu_70::TimeUnitFormat')
    assertTrue("TimeUnitFormat: en and fr diff", (*tmf_en != tmf_fr));
                                                  ~~~~~~~ ^  ~~~~~~
icu4c/source/i18n/unicode/tmutfmt.h:240:17: note: candidate function
TimeUnitFormat::operator!=(const Format& other) const  {
                ^
icu4c/source/i18n/unicode/measfmt.h:151:18: note: candidate function
    virtual bool operator==(const Format &other) const;
                 ^
icu4c/source/i18n/unicode/measfmt.h:151:18: note: candidate function (with reversed parameter order)
```

According to C++20 resolution rules, the code is ambiguous becase `==` and `!=` have different left hand side data types (`MeasureFormat` vs. `TimeUnitFormat`).

Deleting the unnecessary and deprecated `TimeUnitFormat::operator!=()` would resolve the ambiguity and I hope to be able to do that before the RC, but that'd definitely be a public API change by anyone's definition, which is why I propose to start out with this workaround as a first step just to get this code to compile at all and so be able to set up C++20 continuous builds while working on getting rid of the workarounds.

icu4c/source/common/unicode/unistr.h
```diff
@@ -343,7 +343,7 @@ class U_COMMON_API UnicodeString : public Replaceable
    * greater than the characters in `text`, false otherwise
    * @stable ICU 2.0
    */
-  inline UBool operator> (const UnicodeString& text) const;
+  inline bool operator> (const UnicodeString& text) const;
```
markusicu, line 346: We have discussed only changing `==` and `!=`. I don't remember discussing changing any other APIs/operators. Do we need to change these? (Looks like other than this unistr.h file everything else is just equality.)

roubert: In order to just get the currently existing code to build as C++20, it's not necessary to change the return type of the relational operators now (the only reason I did it was that it seemed like a natural thing to do, making the code more consistent and paving the way for future C++20 work, such as the _spaceship_ operator `<=>`).

But it'd be trivial to revert if you think that'd be better.

markusicu: Please only change `==` and `!=` for now, matching the ticket and the discussion.

markusicu: I just read about the “spaceship operator” -- https://devblogs.microsoft.com/cppblog/simplify-your-code-with-rocket-science-c20s-spaceship-operator/

If you think that that will require us changing non-equality operators to return bool, then please submit a new ticket.

roubert: > Please only change `==` and `!=` for now, matching the ticket and the discussion.

Done.

markusicu (approved):
lgtm tnx!

PR #1787, opened 2021-08-05
ICU-21064 Fix declspec macro conflict on clang.

Lord-Kamina:
Finally got around to making this; should fix weird behavior with these macros under Clang, and possible clashes with some other libraries like C++RESTSDK

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21064

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -814,8 +814,8 @@ namespace std {
     /* Use the predefined value. */
 #elif defined(U_STATIC_IMPLEMENTATION)
 #   define U_EXPORT
-#elif defined(_MSC_VER) || (UPRV_HAS_DECLSPEC_ATTRIBUTE(dllexport) && \
-                            UPRV_HAS_DECLSPEC_ATTRIBUTE(dllimport))
+#elif defined(_MSC_VER) || (UPRV_HAS_DECLSPEC_ATTRIBUTE(__dllexport__) && \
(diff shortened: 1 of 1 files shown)


icu4c/source/common/unicode/platform.h
```diff
@@ -814,8 +814,8 @@ namespace std {
     /* Use the predefined value. */
 #elif defined(U_STATIC_IMPLEMENTATION)
 #   define U_EXPORT
-#elif defined(_MSC_VER) || (UPRV_HAS_DECLSPEC_ATTRIBUTE(dllexport) && \
-                            UPRV_HAS_DECLSPEC_ATTRIBUTE(dllimport))
+#elif defined(_MSC_VER) || (UPRV_HAS_DECLSPEC_ATTRIBUTE(__dllexport__) && \
```
jefgen, line 817: FYI/Note for others:
`UPRV_HAS_DECLSPEC_ATTRIBUTE` expands to `__has_declspec_attribute` (if defined), and `__has_declspec_attribute` is a clang language extension. Clang also supports using double-underscores for feature checking macros, so using `__dllexport__` should mean the same as using `dllexport`.

From https://clang.llvm.org/docs/LanguageExtensions.html#has-declspec-attribute :
> The attribute name can also be specified with a preceding and following `__` (double underscore) to avoid interference from a macro with the same name. For instance, `__dllexport__` can be used instead of `dllexport`.

jefgen (approved):
lgtm. Thank you for fixing this! :)

srl295 (approved):

PR #1789, opened 2021-08-06
ICU-13353 Fixed several problems preventing horizontal resource inher…

richgillam:
…itance from working as intended.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-13353

[I've edited the original description here to describe the current state of this PR...]

This PR fixes a number of problems with horizontal resource inheritance in ICU4C.  As a rule (with one exception I've found so far), horizontal inheritance works correctly in ICU4J, so this is about getting ICU4C to work the same as ICU4J.  The basic problem is that the internal implementation of resource bundles in ICU4C is completely different from ICU4J, making it difficult-to-impossible to simply port the code or algorithm from Java to C++.  Fortunately, porting the entire algorithm over doesn't appear to be necessary-- the problems I've found so far seem to be due to clear bugs in the ICU4C code.  I suspect there still are some bugs in here, but this is what I've found so far:

- Corrected an error in `ures_getByKeyWithFallback()` that kept us from properly following an alias when you start with the root resource bundle and ask for a path that doesn't exist but has a containing path that's an alias (e.g., you ask for `"unitsNarrow/mass/pound"` when all "root" has is an alias from `unitsNarrow` to `unitsShort`).
- Corrected an error in `ures_getByKeyWithFallback()` that was causing us to calculate the key path wrong when we followed an alias (we were getting things like `unitsShort/unitsNarrow/mass/pound` when we just wanted `unitsShort/mass/pound`).
- Removed some now-obsolete code in `DateIntervalInfo::initializeData()`.  This code was rendered obsolete by PR #1770.
- Removed workarounds for these issues in `getInflectedMeasureData()` and `getMeasureData()` in `number_longnames.cpp`.
- Changed the expected error code in one test in `number test_api.cpp` to match what the new implementation is giving us (not sure this is the _right_ error code, but I don't actually think the previous error code was right either).
- Restored a few unit tests in `NewResourceBundleTest` that had been hanging with endless loops previously.
- Added a Java version of a unit test I'd previously added only on the C++ side.

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -1910,10 +1910,83 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
   return resource;
 }
 
-U_CAPI UResourceBundle* U_EXPORT2 
-ures_getByKeyWithFallback(const UResourceBundle *resB, 
-                          const char* inKey, 
-                          UResourceBundle *fillIn, 
+static void createPath(const char* origResPath,
+                       int32_t     origResPathLen,
+                       const char* resPath,
+                       int32_t     resPathLen,
+                       const char* inKey,
+                       CharString& path,
+                       UErrorCode* status) {
+    // This is a utility function used by ures_getByKeyWithFallback() below.  This function builds a path from
+    // resPath and inKey, returning the result in `path`.  Originally, this function just cleared `path` and
+    // appended resPath and inKey to it, but that caused problems for horizontal inheritance.
+    //
+    // In normal cases, resPath is the same as origResPath, but if ures_getByKeyWithFallback() has followed an
+    // alias, resPath may be different from origResPath.  Not only may the existing path elements be different,
+    // but resPath may also have MORE path elements than origResPath did.  If it does, those additional path
+    // elements SUPERSEDE the corresponding elements of inKey.  So this code counts the number of elements in
+    // resPath and origResPath and, for each path element in resPath that doesn't have a counterpart in origResPath,
+    // deletes a path element from the beginning of inKey.  The remainder of inKey is then appended to
+    // resPath to form the result.  (We're not using uprv_strchr() here because resPath and origResPath may
+    // not be zero-terminated.)
+    path.clear();
+    const char* key = inKey;
+    if (resPathLen > 0) {
+        path.append(resPath, resPathLen, *status);
+        if (U_SUCCESS(*status)) {
+            const char* resPathLimit = resPath + resPathLen;
+            const char* origResPathLimit = origResPath + origResPathLen;
+            const char* resPathPtr = resPath;
+            const char* origResPathPtr = origResPath;
+            
+            // Remove from the beginning of resPath the number of segments that are contained in origResPath.
+            // If origResPath has MORE segments than resPath, this will leave resPath as the empty string.
+            while (origResPathPtr < origResPathLimit && resPathPtr < resPathLimit) {
+                while (origResPathPtr < origResPathLimit && *origResPathPtr != RES_PATH_SEPARATOR) {
+                    ++origResPathPtr;
+                }
+                if (origResPathPtr < origResPathLimit && *origResPathPtr == RES_PATH_SEPARATOR) {
+                    ++origResPathPtr;
+                }
+                while (resPathPtr < resPathLimit && *resPathPtr != RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                if (resPathPtr < resPathLimit && *resPathPtr == RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+            }
+            
+            // New remove from the beginning of `key` the number of segments remaining in resPath.
+            // If resPath has more segments than `key` does, `key` will end up empty.
+            while (resPathPtr < resPathLimit && *key != '\0') {
+                while (resPathPtr < resPathLimit && *resPathPtr != RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                if (resPathPtr < resPathLimit && *resPathPtr == RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                while (*key != '\0' && *key != RES_PATH_SEPARATOR) {
+                    ++key;
+                }
+                if (*key == RES_PATH_SEPARATOR) {
+                    ++key;
+                }
+            }
+        }
+        // Finally, append what's left of `key` to `path`.  What you end up with here is `resPath`, plus
+        // any pieces of `key` that aren't superseded by `resPath`.
+        // Or, to put it another way, calculate <#-segments-in-key> - (<#-segments-in-resPath> - <#-segments-in-origResPath>),
+        // and append that many segments from the end of `key` to `resPath` to produce the result.
+        path.append(key, *status);
+    } else {
+        path.append(inKey, *status);
+    }
+}
+
+U_CAPI UResourceBundle* U_EXPORT2
+ures_getByKeyWithFallback(const UResourceBundle *resB,
+                          const char* inKey,
+                          UResourceBundle *fillIn,
                           UErrorCode *status) {
     Resource res = RES_BOGUS, rootRes = RES_BOGUS;
     UResourceBundle *helper = NULL;
@@ -1928,24 +2001,32 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
 
     int32_t type = RES_GET_TYPE(resB->fRes);
     if(URES_IS_TABLE(type)) {
+        const char* origResPath = resB->fResPath;
+        int32_t origResPathLen = resB->fResPathLen;
         res = getTableItemByKeyPath(&resB->getResData(), resB->fRes, inKey);
         const char* key = inKey;
+        bool didRootOnce = false;
         if(res == RES_BOGUS) {
             UResourceDataEntry *dataEntry = resB->fData;
             CharString path;
             char *myPath = NULL;
             const char* resPath = resB->fResPath;
             int32_t len = resB->fResPathLen;
-            while(res == RES_BOGUS && dataEntry->fParent != NULL) { /* Otherwise, we'll look in parents */
-                dataEntry = dataEntry->fParent;
+            while(res == RES_BOGUS && (dataEntry->fParent != NULL || !didRootOnce)) { /* Otherwise, we'll look in parents */
+                if (dataEntry->fParent != NULL) {
+                    dataEntry = dataEntry->fParent;
+                } else {
+                    // We can't just stop when we get to a bundle whose fParent is NULL.  That'll work most of the time,
+                    // but if the bundle that the caller passed to us was "root" (which happens in getAllItemsWithFallback(),
+                    // this function will drop right out without doing anything if "root" doesn't contain the exact key path
+                    // specified.  In that case, we need one extra time through this loop to make sure we follow any
+                    // applicable aliases at the root level.
+                    didRootOnce = true;
+                }
                 rootRes = dataEntry->fData.rootRes;
 
                 if(dataEntry->fBogus == U_ZERO_ERROR) {
-                    path.clear();
-                    if (len > 0) {
-                        path.append(resPath, len, *status);
-                    }
-                    path.append(inKey, *status);
+                    createPath(origResPath, origResPathLen, resPath, len, inKey, path, status);
                     if (U_FAILURE(*status)) {
                         ures_close(helper);
                         return fillIn;
@@ -1956,7 +2037,7 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
                         res = res_findResource(&(dataEntry->fData), rootRes, &myPath, &key);
                         if (RES_GET_TYPE(res) == URES_ALIAS && *myPath) {
                             /* We hit an alias, but we didn't finish following the path. */
-                            helper = init_resb_result(dataEntry, res, NULL, -1, resB, helper, status); 
+                            helper = init_resb_result(dataEntry, res, NULL, -1, resB, helper, status);
                             /*helper = init_resb_result(dataEntry, res, inKey, -1, resB, helper, status);*/
                             if(helper) {
                               dataEntry = helper->fData;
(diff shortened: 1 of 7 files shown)


sffc (commented):
Wow!  This is a piece of work that has been needed for a long time!  Thank you!

dtptngen.cpp and number_longnames.cpp look good.  I also looked at the docs for your new function ures_getAllChildrenWithFallback.  I'll let Markus review the hard stuff in uresbund.cpp.

icu4c/source/common/uresbund.cpp
```diff
@@ -1031,8 +1031,12 @@ static UResourceBundle *init_resb_result(const ResourceData *rdata, Resource r,
                         /* this is an XPath alias, starting with "/LOCALE/" */
                         /* it contains the path to a resource which should be looked up */
                         /* starting in the requested locale */
-                        keyPath = locale; 
-                        locale = parent->fTopLevelData->fName; /* this is the requested locale's name */
+                        keyPath = locale;
+                        if (aliasBaseLocale != NULL) {
+                            locale = const_cast<char*>(aliasBaseLocale);
```
hugovdm, line 1036: Is this const_cast necessary?

(I see aliasBaseLocale is already a "const char*".)

richgillam: >I see aliasBaseLocale is already a "const char*".

Right.  I'm casting _away_ the const here.  Casting away a const is always kind of an iffy thing to do, but the original code seemed to be doing the moral equivalent of casting away the const (it's been a while, so I don't remember the details off the top of my head), and I couldn't think of a better approach.

>Is this const_cast necessary?

I was failing some of the automated PR checks without it.

sffc: Is it possible to change the variable `locale` itself to be `const char*` ?

richgillam: I think the thing that prevents me from just making `locale` const is that we're writing through it on line 1026 above ("`*locale = 0;`").  I could split this into two variables, a const one to convey the locale to the `ures_openDirect()` call on line 1069 and another, temporary, non-const one just for the munging of the locale ID in ll. 1022-1028, but it felt like it'd be a lot simpler to just cast away the const and not worry about it.

I'll admit my thinking here might be influenced by the fact that I found this after already getting it to work and didn't want to redo the whole thing because of a compile error, but it still seems like it's not really worth the trouble.

hugovdm: Aha: thanks a bunch for the explanation! Now it looks obvious. Some things only become obvious once I've bumped my head against them (or learned the language in a more thorough way).

I did not get very far in my review, but it looks like @sffc already took care of the parts where I could have been useful. :-) Bigger-picture appraisal of this work: thanks for this! I had struggled coming to terms with the existing inconsistencies between Java and C/C++.

richgillam: >Aha: thanks a bunch for the explanation! Now it looks obvious. Some things only become obvious once I've bumped my head against them (or learned the language in a more thorough way).

I understand.  It took me a while to go back through the code and figure that out again for myself!

>I did not get very far in my review, but it looks like @sffc already took care of the parts where I could have been useful. :-) Bigger-picture appraisal of this work: thanks for this! I had struggled coming to terms with the existing inconsistencies between Java and C/C++.

Thank you for the kind words and for taking the time to look!

richgillam:
The latest commit rebases my original fixes on top of all of the previous horizontal-inheritance changes.  Changes on this PR that were already made as part of one of the previous PRs have been removed.  Those changes took care of the problem for which I'd added all the "`aliasBaseLocale`" stuff in the original version of this PR, so the "`aliasBaseLocale`" stuff is also all gone now.  The code at ll. 2057-2064 (the `createPath()` stuff) got more complicated compared to the original version; I think the code here points to problems elsewhere where we're calling `init_resb_result()` and not setting `fResPath` correctly: The problem I'm trying to work around is where a resource shows up here with an `fResPath` of `"square-meter/"` and my original code ends up trying to pad it out to `"unitsShort/area/square-meter/"` but ends up with `"square-meter/area/square-meter/"`.  This needs another look and is probably justification for doing ICU-21715, but I think we can live without tackling that in ICU 70.

richgillam:
Apologies-- I forgot to commit my actual last set of changes...

markusicu:
Please update the PR description to reflect what's left here (e.g., removing parts that got done in earlier PRs).

markusicu:
You mentioned a Java version of a unit test. Please add that Java unit test.

markusicu:
Looking at the changes in number_longnames.cpp, it does look like you fixed ICU-13353 -- so that should be a fair ticket to use.

icu4c/source/common/uresbund.cpp
```diff
@@ -1928,24 +1992,32 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
...
-                dataEntry = dataEntry->fParent;
+            while(res == RES_BOGUS && (dataEntry->fParent != NULL || !didRootOnce)) { /* Otherwise, we'll look in parents */
+                if (dataEntry->fParent != NULL) {
+                    dataEntry = dataEntry->fParent;
+                } else {
+                    // We can't just stop when we get to a bundle whose fParent is NULL.  That'll work most of the time,
+                    // but if the bundle that the caller passed to us was "root" (which happens in getAllItemsWithFallback(),
+                    // this function will drop right out without doing anything if "root" doesn't contain the exact key path
+                    // specified.  In that case, we need one extra time through this loop to make sure we follow any
+                    // applicable aliases at the root level.
```
markusicu, line 2014: I believe you, but do you have a concrete example at hand that this change here fixes? getAllItemsWithFallback() starting from root with which path? Just trying to wrap my head around this...

icu4c/source/common/uresbund.cpp
```diff
@@ -1982,7 +2054,20 @@ ures_getByKeyWithFallback(const UResourceBundle *resB,
                     *status = U_USING_FALLBACK_WARNING;
                 }
 
-                fillIn = init_resb_result(dataEntry, res, inKey, -1, resB, fillIn, status);
+                fillIn = init_resb_result(dataEntry, res, key, -1, resB, fillIn, status);
+                if (resPath != nullptr) {
+                    createPath(path, resPath, len, origResPath, origResPathLen, inKey, status);
+                } else if (*(uprv_strchr(fillIn->fResPath, RES_PATH_SEPARATOR) + 1) != '\0') {
```
markusicu, line 2060: fResPath can be nullptr (if the fillIn is for the rootRes of some bundle), so you should guard against that, unless there is earlier code in this function that guarantees it not to be nullptr.
And if the strchr() finds no `/` then it returns nullptr and the code crashes.

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
-ures_getByKeyWithFallback(const UResourceBundle *resB, 
-                          const char* inKey, 
-                          UResourceBundle *fillIn, 
+static void createPath(CharString& path,
+                       const char* resPath,
+                       int32_t     resPathLen,
+                       const char* origResPath,
+                       int32_t     origResPathLen,
+                       const char* inKey,
+                       UErrorCode* status) {
```
markusicu, line 1919: nit/optional (that is, you can ignore me):
I would order the args like this:
1. origResPath (base)
2. resPath (input 1)
3. inKey (input 2)
4. path (output)
5. status (UErrorCode last)

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
+    // appended resPath and inKey to it, but that caused problems for horizontal inheritance.
+    //
+    // In normal cases, resPath is the same as origResPath, but if ures_getByKeyWithFallback() has followed an
+    // alias, resPath may be different from origResPath.  Not only may the existing path elements be different,
+    // but resPath may also have MORE path elements than origResPath did.  If it does, those additional path
+    // elements SUPERSEDE the corresponding elements of inKey.  So this code counts the number of elements in
+    // resPath and origResPath and, for each path element in resPath that doesn't have a counterpart in origResPath,
+    // deletes a path element from the beginning of inKey.  The remainder of inKey is then appended to
+    // resPath to form the result.  (We're not using uprv_strchr() here because resPath and origResPath may
+    // not be zero-terminated.)
```
markusicu, line 1931: This... sounds plausible, and I know you are fixing various bugs, so good.
However, it also sounds like we are repairing something that got messed up earlier, and it should be possible to properly track and adjust the actual keyPath as we go.
Maybe I am wrong, and the interplay between ures_getByKeyWithFallback() and createPath() is just right, but it feels dangerously convoluted. I should take time to step through the alias-following code in a debugger and see for myself...

Anyway, let's keep this as is for now. Quite possible that you have figured it all out and I am just not understanding it yet.

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
+    path.clear();
+    const char* key = inKey;
+    if (resPathLen > 0) {
+        path.append(resPath, resPathLen, *status);
+        if (U_SUCCESS(*status)) {
+            const char* resPathEnd = resPath + resPathLen;
+            const char* origResPathEnd = origResPath + origResPathLen;
```
markusicu, line 1938: optional (mostly): in ICU (back from your earlier ICU days) we still have the useful convention of using "limit" for an *exclusive* end, and "end" only for an *inclusive* end

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
+            const char* resPathPtr = resPath;
+            const char* origResPathPtr = origResPath;
+            
+            while (origResPathPtr < origResPathEnd && resPathPtr < resPathEnd) {
```
markusicu, line 1942: Please add a comment to explain this loop. I think it does this:
```
// Skip one keyPath segment at a time in parallel in both origResPath and resPath,
// until one or both run out of segments.
```

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
+                    ++origResPathPtr;
+                }
+                while (resPathPtr < resPathEnd && *resPathPtr != RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                if (resPathPtr < resPathEnd && *resPathPtr == RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+            }
+            while (resPathPtr < resPathEnd && *key != '\0') {
```
markusicu, line 1956: Please add a comment to explain this loop. I think it does this:
```
// Skip one keyPath segment at a time in parallel in both the remainder of resPath and inKey,
// until one or both run out of segments.
```

icu4c/source/common/uresbund.cpp
```diff
@@ -1910,10 +1910,74 @@ static Resource getTableItemByKeyPath(const ResourceData *pResData, Resource tab
...
+                while (resPathPtr < resPathEnd && *resPathPtr != RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                if (resPathPtr < resPathEnd && *resPathPtr == RES_PATH_SEPARATOR) {
+                    ++resPathPtr;
+                }
+                while (*key != '\0' && *key != RES_PATH_SEPARATOR) {
```
markusicu, line 1963: FYI only:
In Java, I had changed the storage of a keyPath into an array of String, one per segment. Simplified a lot of processing. A rough equivalent in C++ would be storing NUL-terminated segments consecutively in one buffer, maybe with an extra terminating NUL or else with the number of segments at the start or elsewhere. Then we could march through here just with uprv_strchr(keyPath, 0), and elsewhere we wouldn't need to copy the path so that we can NUL-terminate segments for strcmp()... Food for thought for 71.

markusicu:
Re https://unicode-org.atlassian.net/browse/ICU-13353 “I’ll have a new PR up soon…”

Please don't make a new PR. Just change the ticket number here, and when you eventually (at the end!) squash the commits, rewrite the ticket number there too.

richgillam:
>Please update the PR description to reflect what's left here (e.g., removing parts that got done in earlier PRs).

Done.

>You mentioned a Java version of a unit test. Please add that Java unit test.

Done.

>Please don't make a new PR. Just change the ticket number here, and when you eventually (at the end!) squash the commits, rewrite the ticket number there too.

I was scared either that the tools wouldn't handle a change of ticket number or that I'd screw it up, but okay.  I'll keep all my changes here and follow your instructions for fixing the ticket number.

icu4c/source/common/uresbund.cpp
richgillam: I don't remember precisely off the top of my head, but yes, you definitely have unit-test failures if you take out this logic.  As I remember it, the situation this addressed was where you ask for `unitsNarrow/area/square-foot` in `en_GB` and it falls back to `root`, which doesn't have `unitsNarrow/area/square-foot`; it only has `unitsNarrow`, which is an alias to `unitsShort`.

As I recall, the problem was that when we get to `root`, the `dataEntry->fParent != NULL` check in line 2006 causes us to bypass the whole loop, and it's the code in that loop (particularly ll. 2029-2039) that actually detects and follows the alias, so we don't follow the alias.

Faking an extra iteration through the loop to make sure we follow the alias is cheesy, especially since it forces us through the loop an extra time when we _don't_ need to do it, but I couldn't think of a cleaner way to do it without a lot more surgery to this function than I was comfortable making.  If you can recommend a better way to do this, I'm all ears.

icu4c/source/common/uresbund.cpp
richgillam: Good catch.  Fixed.

icu4c/source/common/uresbund.cpp
richgillam: Eh, okay.

icu4c/source/common/uresbund.cpp
richgillam: Again, it's been a while since I originally figured this out, but what I remember is that you'd get paths of different lengths in cases where the alias was at a different level than the thing you were actually looking for.  That is, if you were looking for `calendar/coptic` and it aliased over to `calendar/generic`, that worked fine, but if you were looking for `calendar/coptic/availableFormats` and followed the alias from `calendar/coptic` to `calendar/generic`, then the `resPath` you got back would be something like `generic/availableFormats` and if you just stuck things together naively, you'd end up with `calendar/generic/coptic/availableFormats`, which is of course wrong.  Or something like that-- I'm sure I have the details wrong.

In other words, this logic can probably be improved, but I think the need to match up the segments is inherent in what we're doing and not really a symptom of a bug.  But it's been a while, so maybe I'm wrong about this.

icu4c/source/common/uresbund.cpp
richgillam: I don't remember that, but these days I'm lucky if I can remember what I had for lunch yesterday.  Fixed.

icu4c/source/common/uresbund.cpp
richgillam: I'd hoped what I was doing was explained adequately by the block comment at the top of the function, but apparently not.  I added comments to the body in the hopes of explaining it better.  It's a lot of code to do something fairly simple and could probably be refactored to be clearer.

icu4c/source/common/uresbund.cpp
richgillam: The arrays of String in Java definitely are easier to understand, and the actual logic using them easier to follow (and significantly different from the logic in the C++ code).  I kind of like what you're suggesting here, and I wonder if we could go even further to make the actual algorithms match what we're doing on the Java side, but I was _way_ too scared to attempt anything like that for ICU 70.

markusicu (commented):
tnx lgtm pse squash
... and remember to change the squashed commit's comment to say ICU-13353

markusicu:
Please take a look at the "CI (C: Linux Clang-13 WarningsAsErrors (Ubuntu 20.04))" failure and see if it's a real problem.

richgillam:
>Please take a look at the "CI (C: Linux Clang-13 WarningsAsErrors (Ubuntu 20.04))" failure and see if it's a real problem.

I don't think it's a real problem.  I can't really make head nor tail out of that log info, but it looks like it's failing in the "Install Clang 13" stage, before it does anything with our code at all.

markusicu (approved):
lgtm merci beaucoup

jefgen:
> > Please take a look at the "CI (C: Linux Clang-13 WarningsAsErrors (Ubuntu 20.04))" failure and see if it's a real problem.
> 
> I don't think it's a real problem. I can't really make head nor tail out of that log info, but it looks like it's failing in the "Install Clang 13" stage, before it does anything with our code at all.

I'll try re-running the build. So far it looks like it's gotten past the "Install Clang-13" step.

PR #1791, opened 2021-08-11
ICU-21707 Fix LocaleBuilder if default locale has BCP47 extension tags

jefgen:
The `LocaleBuilder` class assumes that ICU's default locale will never have any BCP47 extension tags on it.

The issue is that it uses a `Locale` object internally to store/track any extensions/keywords, but it creates this by calling `new Locale()`, assuming that the default locale will never have any pre-existing extensions/keywords on it.

This change modifies the `LocaleBuilder` class so that when creating the internal `Locale` object it will clear any pre-existing keywords by calling `uloc_getBaseName` (which removes any keywords from the name).

It adds a new test case which changes the default locale, then runs the `TestLocaleBuilderBasic` test (which previously would have failed).

This change also adds a new CI build bot that has the `LANG` environment variable set to `en_US@calendar=gregorian;hours=h12` in order to test with the default locale containing extension tags on ICU first start-up.

I used the [newly added](https://github.com/unicode-org/icu/pull/1762) `PreflightingLocaleIDBuffer` helper class for this, but could use something else as well if preferred.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21707

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -118,6 +118,24 @@ jobs:
         CC: gcc
         CXX: g++
 #-------------------------------------------------------------------------
+- job: ICU4C_Clang_Ubuntu_2004_LANG
+  displayName: 'C: Linux Clang (Ubuntu 20.04) - LANG has extension tags'
+  timeoutInMinutes: 30
+  pool:
+    vmImage: 'ubuntu-20.04'
+  steps:
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
+    - script: |
+        cd icu4c/source && ./runConfigureICU Linux && make -j2 check
+      displayName: 'Build and Test'
+      env:
+        CC: clang
+        CXX: clang++
+        LANG: "en_US@calendar=gregorian;hours=h12"
+
+#-------------------------------------------------------------------------
 # VS 2019 Builds
 #-------------------------------------------------------------------------
 - job: ICU4C_MSVC_x64_Release_Distrelease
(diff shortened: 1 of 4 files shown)


icu4c/source/common/localebuilder.cpp
```diff
@@ -228,9 +246,8 @@ LocaleBuilder& LocaleBuilder::setExtension(char key, StringPiece value)
         return *this;
     }
     if (extensions_ == nullptr) {
-        extensions_ = new Locale();
```
FrankYFTang, line 231: I think the fix could be simplfy to just
extensions_ = Locale::getRoot().clone();

It does not really need to know the default locale but just try to get a locale. It is using that locale as a way to copy the extension only.

jefgen: Thanks Frank, that is a good idea. I'll update the PR to use that approach instead. Thank you!

jefgen: I pushed a new commit to use your suggestion instead.
Thanks!

.ci-builds/.azure-pipelines.yml
```diff
@@ -118,6 +118,24 @@ jobs:
...
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
+    - script: |
+        cd icu4c/source && ./runConfigureICU Linux && make -j2 check
+      displayName: 'Build and Test'
+      env:
+        CC: clang
+        CXX: clang++
+        LANG: "en_US@calendar=gregorian;hours=h12"
```
erik0686, line 136: Thanks for this! Really good idea :)

FrankYFTang (dismissed):
LGTM . Thanks for fixing this.

erik0686 (approved):

erik0686 (approved):

jefgen:
> LGTM . Thanks for fixing this.

Thanks Frank.
I squashed the commits, so will need one more approval.

.ci-builds/.azure-pipelines.yml
jefgen: Thanks!

PR #1792, opened 2021-08-11
ICU-21662 Improve UVector error handling.

aheninger:
- Add updated versions of UVector::addElement() and ensureCapacity() that respect
  incoming errors.
  Follow on to c26aebe, which renamed the original versions.

- Add UVector::adoptElement() as a replacement for addElement() when the vector
  has a deleter function set, meaning that it adopts ownership of its elements.

  The intent is to make the behavior clearer at the call sites when looking
  at unfamiliar code.

- Make all functions with an incoming failure, as indicated by a UErrorCode parameter,
  leave the vector unchanged.

- Change all functions that store object pointers into the vector such that,
  when the store cannot be completed for any reason _and_ the vector has a deleter function,
  then the incoming object is deleted.

  This change can simplify the error handling code around calls to the affected functions
  (addElement() and insertElementAt(), in particular)

- Add index bounds checking on functions where it was possible - that is, on functions
  with both U_ErrorCode and index parameters.

- Changed to more modern C++ idioms in some parts of the UVector implementation.

- Review & update as required all uses of the UVector functions
  setElementAt(), insertElementAt(), setSize(), sortedInsert()
  these being the functions with changed behavior on error conditions
  (aside from addElement()).

This PR will be followed by more, switching call sites in various ICU services
from UVector::addElementX() (old behavior on errors)
to   UVector::addElement()  (new behavior on errors)

##### Checklist

### DIFF
--- a/icu4c/source/common/serv.cpp
+++ b/icu4c/source/common/serv.cpp
@@ -257,20 +257,13 @@ public:
     }
 };
 
-// UObjectDeleter for serviceCache
+// Deleter for serviceCache
 U_CDECL_BEGIN
 static void U_CALLCONV
 cacheDeleter(void* obj) {
     U_NAMESPACE_USE ((CacheEntry*)obj)->unref();
 }
 
-/**
-* Deleter for UObjects
-*/
-static void U_CALLCONV
-deleteUObject(void *obj) {
-    U_NAMESPACE_USE delete (UObject*) obj;
-}
 U_CDECL_END
 
 /*
@@ -534,11 +527,10 @@ ICUService::getKey(ICUServiceKey& key, UnicodeString* actualReturn, const ICUSer
                 status = U_MEMORY_ALLOCATION_ERROR;
                 return NULL;
             }
-            cacheDescriptorList->addElementX(idToCache.getAlias(), status);
+            cacheDescriptorList->adoptElement(idToCache.orphan(), status);
             if (U_FAILURE(status)) {
                 return NULL;
             }
-            idToCache.orphan(); // cacheDescriptorList now owns the string.
         } while (key.fallback());
 outerEnd:
 
@@ -612,6 +604,7 @@ ICUService::getVisibleIDs(UVector& result, const UnicodeString* matchID, UErrorC
     if (U_FAILURE(status)) {
         return result;
     }
+    UObjectDeleter *savedDeleter = result.setDeleter(uprv_deleteUObject);
 
     {
         Mutex mutex(&lock);
@@ -619,7 +612,7 @@ ICUService::getVisibleIDs(UVector& result, const UnicodeString* matchID, UErrorC
         if (map != NULL) {
             ICUServiceKey* fallbackKey = createKey(matchID, status);
 
-            for (int32_t pos = UHASH_FIRST;;) {
+            for (int32_t pos = UHASH_FIRST; U_SUCCESS(status); ) {
                 const UHashElement* e = map->nextElement(pos);
                 if (e == NULL) {
                     break;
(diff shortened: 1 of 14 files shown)


icu4c/source/common/uvector.cpp
```diff
@@ -451,7 +486,8 @@ void UVector::sortedInsert(void* obj, UElementComparator *compare, UErrorCode& e
  * be sorted already.
  */
 void UVector::sortedInsert(int32_t obj, UElementComparator *compare, UErrorCode& ec) {
-    UElement e;
+    U_ASSERT(deleter == nullptr);
```
jefgen, line 489: Should the `sortedInsert` above maybe assert that there is a deleter?
`U_ASSERT(deleter != nullptr);`

icu4c/source/common/serv.cpp
```diff
@@ -845,13 +833,13 @@ ICUService::createSimpleFactory(UObject* objToAdopt, const UnicodeString& id, UB
...
-ICUService::registerFactory(ICUServiceFactory* factoryToAdopt, UErrorCode& status) 
+ICUService::registerFactory(ICUServiceFactory* factoryToAdopt, UErrorCode& status)
 {
     if (U_SUCCESS(status) && factoryToAdopt != NULL) {
         Mutex mutex(&lock);
 
         if (factories == NULL) {
-            factories = new UVector(deleteUObject, NULL, status);
+            factories = new UVector(uprv_deleteUObject, NULL, status);
             if (U_FAILURE(status)) {
```
jefgen, line 843: Might also want to check if `factories` is `nullptr` (in case of OOM), and set `status` to `U_MEMORY_ALLOCATION_ERROR` and return, so that it doesn't try to call `factories->insertElementAt` on nullptr below.

icu4c/source/i18n/rbt_pars.cpp
```diff
@@ -976,9 +976,12 @@ void TransliteratorParser::parseRules(const UnicodeString& rule,
             if (!parsingIDs) {
                 if (curData != NULL) {
                     if (direction == UTRANS_FORWARD)
-                        dataVector.addElementX(curData, status);
+                        dataVector.addElement(curData, status);
                     else
                         dataVector.insertElementAt(curData, 0, status);
+                    if (U_FAILURE(status)) {
+                        delete curData;
```
jefgen, line 983: I wonder if maybe adding a `U_ASSERT(!dataVector.hasDeleter());` here?

If I understand the new approach:
- `addElement` doesn't delete the object.
- `insertElementAt` does delete (if there is a deleter).

(in case somehow the `dataVector` does end up with a deleter, to prevent a double-delete ?)

icu4c/source/i18n/rbt_pars.cpp
```diff
@@ -1032,9 +1035,12 @@ void TransliteratorParser::parseRules(const UnicodeString& rule,
                     return;
                 }
                 if (direction == UTRANS_FORWARD)
-                    idBlockVector.addElementX(tempstr, status);
+                    idBlockVector.adoptElement(tempstr, status);
                 else
                     idBlockVector.insertElementAt(tempstr, 0, status);
+                if (U_FAILURE(status)) {
```
jefgen, line 1041: possibly also assert `idBlockVector` has no deleter.

icu4c/source/common/serv.cpp
jefgen: Thanks for cleaning all this up!

icu4c/source/common/uvector.cpp
```diff
@@ -359,30 +364,60 @@ UBool UVector::ensureCapacityX(int32_t minimumCapacity, UErrorCode &status) {
...
+}
 /**
  * Change the size of this vector as follows: If newSize is smaller,
  * then truncate the array, possibly deleting held elements for i >=
  * newSize.  If newSize is larger, grow the array, filling in new
  * slots with NULL.
  */
 void UVector::setSize(int32_t newSize, UErrorCode &status) {
-    int32_t i;
-    if (newSize < 0) {
```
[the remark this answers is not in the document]
jefgen: Ah, neverminded. I failed to notice that `ensureCapacity` now sets `U_ILLEGAL_ARGUMENT_ERROR` if the `newSize` is < 0.

jefgen (approved):
LGTM. Thank you very much for all of these changes!

(Thanks also for cleaning up some of the other OOM issues as well).

jefgen (approved):
lgtm (after the squash and additional changes).

Thanks again!

PR #1794, opened 2021-08-17
ICU-20973 Add C++20 build bots to Azure CI: Clang-13, GCC-11, MSVC with std:c++latest

jefgen:
Thanks to the work in PR https://github.com/unicode-org/icu/pull/1786, ICU4C now compiles with C++20.

This change adds 3 additional build bots to the Azure CI builds, in order to provide C++20 build coverage in the CI builds.

Notes:
- The Ubuntu 20.04 image doesn't have Clang-13 or GCC-11, so I needed to download and install it as part of the steps.
- For VS2019, there is no `/std:c++20` flag yet, so I used `/std:c++latest` as that includes many of the C++20 compiler and library changes.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20973

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -118,6 +118,28 @@ jobs:
         CC: gcc
         CXX: g++
 #-------------------------------------------------------------------------
+- job: ICU4C_GCC11_Cpp20_Ubuntu_2004
+  displayName: 'C: Linux GCC-11 C++20 (Ubuntu 20.04)'
+  timeoutInMinutes: 30
+  pool:
+    vmImage: 'ubuntu-20.04'
+  steps:
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
+    - script: |
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+        sudo apt-get update
+        sudo apt-get install gcc-11 g++-11
+      displayName: 'Install GCC-11'
+    - script: |
+        cd icu4c/source && ./runConfigureICU Linux && make -j2 check
+      displayName: 'Build and Test'
+      env:
+        CC: gcc-11
+        CXX: g++-11
+        CXXFLAGS: '-std=c++20 -fext-numeric-literals'
+#-------------------------------------------------------------------------
 - job: ICU4C_Clang_Ubuntu_2004_LANG
   displayName: 'C: Linux Clang (Ubuntu 20.04) - LANG has extension tags'
   timeoutInMinutes: 30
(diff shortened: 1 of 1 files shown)


markusicu (dismissed):
lgtm tnx!

.ci-builds/.azure-pipelines.yml
```diff
@@ -134,7 +134,29 @@ jobs:
...
+      fetchDepth: 10
+    # Install Clang-13 from https://apt.llvm.org/
+    - script: |
+        curl -Ls https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+        sudo apt-add-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main'
+        sudo apt update
+        sudo apt install -qy clang-13 lld-13 libc++-13-dev libc++abi-13-dev
+      displayName: 'Install Clang-13'
+    - script: |
+        export CXXFLAGS="-std=c++20" && cd icu4c/source && ./runConfigureICU Linux && make -j2 check
```
roubert, line 155: Why is `CXXFLAGS` not in the `env` section together with `CC` and `CXX`?

roubert (changes requested):
Could you add a C++20 build with GCC 11 too?

I think that'd be very valuable to have, for my experience is that GCC can behave rather differently from both Clang and MSVC particularly regarding details in newer standards.

.ci-builds/.azure-pipelines.yml
jefgen: Ah, the other build bots in this config file used `export` for the `CXXFLAGS`, so I did it here to be consistent.
However, I think you're right that it might be better to do it the same as `CC` and `CX`. 
I'll push an update in a moment.

jefgen:
> Could you add a C++20 build with GCC 11 too?

Good point. I'll try adding a GCC 11 build as well. I'll push an update in a moment.

daniel-ju (approved):

roubert (approved):
LGTM

It's a bit funny (but totally harmless) that one of the configurations uses `apt-get` while the other one uses `apt` for doing the same thing.

markusicu (approved):
lgtm

jefgen:
> It's a bit funny (but totally harmless) that one of the configurations uses `apt-get` while the other one uses `apt` for doing the same thing.

Yeah, they're aliases so it doesn't matter too much, but it would be nice to be consistent. I'll try to remember to clean it up next time I touch this file, or to comment if someone else touches it. :)

PR #1795, opened 2021-08-17
ICU-21460 Changed the ULocale initializers to allow locale IDs that use BCP47 syntax, but with _ as a field delimiter.

richgillam:
…core characters as field delimiters.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21460

This seemed fairly straightforward when I agreed to take on this bug, but was a little less so in practice.  The basic idea is that even though the `_` character is illegal in BCP47 language tags, because we support both the BCP47 syntax and the old-style locale-ID syntax (and a hybrid of the two), we'll still see locale IDs that use BCP47 syntax but with underscores between some of the fields.  So I just changed the BCP47 parser so that it'd accept either `-` or `_` as a field delimiter.

This works fine and solves the original symptom reported in JIRA, but I ended up having to change the expected results in a few existing unit tests.  I **think** the new results are okay, but I'm not absolutely certain I'm not having undesirable side effects.

On the Java side, there was an additional problem, where just changing the parser to accept `_` as a field delimiter resulted in "`x-klingon`" getting interpreted as "`@x=klingon`" (my change on the C side didn't have this problem).  I put in a workaround for this (because there didn't seem to be a good way to treat only `-` as a valid delimiter in some places and both `-` and `_` as valid in other places since they're all using the same token iterator), but it's kind of kludgey and I'm not sure it's the best solution to this problem.  Advice welcome.

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -1477,21 +1477,37 @@ _canonicalize(const char* localeID,
               ByteSink& sink,
               uint32_t options,
               UErrorCode* err) {
+    if (U_FAILURE(*err)) {
+        return;
+    }
+
     int32_t j, fieldCount=0, scriptSize=0, variantSize=0;
-    PreflightingLocaleIDBuffer tempBuffer;
+    PreflightingLocaleIDBuffer tempBuffer;  // if localeID has a BCP47 extension, tmpLocaleID points to this
     const char* origLocaleID;
     const char* tmpLocaleID;
     const char* keywordAssign = NULL;
     const char* separatorIndicator = NULL;
 
-    if (U_FAILURE(*err)) {
-        return;
-    }
-
     if (_hasBCP47Extension(localeID)) {
+        CharString localeIDWithHyphens;
+        const char* localeIDPtr = localeID;
+
+        // convert all underbars to hyphens, unless the "BCP47 extension" comes at the beginning of the string
+        if (uprv_strchr(localeID, '_') != nullptr && localeID[1] != '-' && localeID[1] != '_') {
+            localeIDWithHyphens.append(localeID, -1, *err);
+            if (U_SUCCESS(*err)) {
+                for (char* p = localeIDWithHyphens.data(); *p != '\0'; ++p) {
+                    if (*p == '_') {
+                        *p = '-';
+                    }
+                }
+                localeIDPtr = localeIDWithHyphens.data();
+            }
+        }
+        
         do {
-            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeID,
-                tempBuffer.getBuffer(), tempBuffer.getCapacity(), err);
+            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeIDPtr, tempBuffer.getBuffer(),
+                                                         tempBuffer.getCapacity(), err);
         } while (tempBuffer.needToTryAgain(err));
     } else {
         if (localeID==NULL) {
--- a/icu4c/source/test/cintltst/capitst.c
+++ b/icu4c/source/test/cintltst/capitst.c
@@ -90,6 +90,7 @@ void addCollAPITest(TestNode** root)
     addTest(root, &TestBengaliSortKey, "tscoll/capitst/TestBengaliSortKey");
     addTest(root, &TestGetKeywordValuesForLocale, "tscoll/capitst/TestGetKeywordValuesForLocale");
     addTest(root, &TestStrcollNull, "tscoll/capitst/TestStrcollNull");
+    addTest(root, &TestLocaleIDWithUnderscoreAndExtension, "tscoll/capitst/TestLocaleIDWithUnderscoreAndExtension");
 }
 
 void TestGetSetAttr(void) {
@@ -2563,6 +2564,20 @@ static void TestStrcollNull(void) {
     }
 
     ucol_close(coll);
+}
+
+static void TestLocaleIDWithUnderscoreAndExtension(void) {
+    UErrorCode err = U_ZERO_ERROR;
+    UCollator* c1 = ucol_open("en-US-u-kn-true", &err);
+    UCollator* c2 = ucol_open("en_US-u-kn-true", &err);
+    
+    if (assertSuccess("Failed to create collators", &err)) {
+        assertTrue("Comparison using \"normal\" collator failed", !ucol_greater(c1, u"2", -1, u"10", -1));
+        assertTrue("Comparison using \"bad\" collator failed", !ucol_greater(c2, u"2", -1, u"10", -1));
+    }
+    
+    ucol_close(c1);
+    ucol_close(c2);
 }
 
 #endif /* #if !UCONFIG_NO_COLLATION */
--- a/icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
@@ -1131,10 +1131,13 @@ public final class ULocale implements Serializable, Comparable<ULocale> {
      * @stable ICU 3.0
      */
     public static String getName(String localeID){
-        String tmpLocaleID;
+        String tmpLocaleID = localeID;
         // Convert BCP47 id if necessary
         if (localeID != null && !localeID.contains("@") && getShortestSubtagLength(localeID) == 1) {
-            tmpLocaleID = forLanguageTag(localeID).getName();
+            if (localeID.indexOf('_') >= 0 && localeID.charAt(1) != '_' && localeID.charAt(1) != '-') {
+                tmpLocaleID = localeID.replace('_', '-');
+            }
+            tmpLocaleID = forLanguageTag(tmpLocaleID).getName();
             if (tmpLocaleID.length() == 0) {
                 tmpLocaleID = localeID;
             }
--- a/icu4c/source/test/cintltst/capitst.h
+++ b/icu4c/source/test/cintltst/capitst.h
@@ -136,6 +136,11 @@
      * test strcoll with null arg
      */
     static void TestStrcollNull(void);
+ 
+    /**
+     * Simple test for ICU-21460.  The issue affects all components, but was originally reported against collation.
+     */
+    static void TestLocaleIDWithUnderscoreAndExtension(void);
 
 #endif /* #if !UCONFIG_NO_COLLATION */
 
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -3723,13 +3723,13 @@ const char* const basic_maximize_data[][2] = {
     ""
   }, {
      "de_u_co_phonebk",
-     "de_Latn_DE_U_CO_PHONEBK"
+     "de_Latn_DE@collation=phonebook"
   }, {
      "de_Latn_u_co_phonebk",
-     "de_Latn_DE_U_CO_PHONEBK"
+      "de_Latn_DE@collation=phonebook"
   }, {
      "de_Latn_DE_u_co_phonebk",
-     "de_Latn_DE_U_CO_PHONEBK"
+      "de_Latn_DE@collation=phonebook"
   }, {
     "_Arab@em=emoji",
     "ar_Arab_EG@em=emoji"
@@ -6377,7 +6377,7 @@ static const struct {
     {"hant-cmn-cn", "hant", 4},
     {"zh-cmn-TW", "cmn_TW", FULL_LENGTH},
     {"zh-x_t-ab", "zh", 2},
-    {"zh-hans-cn-u-ca-x_t-u", "zh_Hans_CN@calendar=yes",  15},
+    {"zh-hans-cn-u-ca-x_t-u", "zh_Hans_CN@calendar=yes", 15},
     /* #20140 dupe keys in U-extension */
     {"zh-u-ca-chinese-ca-gregory", "zh@calendar=chinese", FULL_LENGTH},
     {"zh-u-ca-gregory-co-pinyin-ca-chinese", "zh@calendar=gregorian;collation=pinyin", FULL_LENGTH},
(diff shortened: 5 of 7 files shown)


FrankYFTang (approved):
LGTM. Thanks for working on this.

markusicu:
ICU-TC discussion:
- We do want the Locale constructor and ucol_open() etc. to accept both `_` and `-`
- We *do not* want Locale::forLanguageTag() and LocaleBuilder::setLanguageTag() to accept underscores; we want them to be compatible with BCP 47 and Java class Locale
- @FrankYFTang will submit a new-API ticket for Locale::forUnicodeLocaleID() and LocaleBuilder::setUnicodeLocaleID() which validate according to https://www.unicode.org/reports/tr35/#Unicode_locale_identifier (and to check for LocaleBuilder discrepancies in C++ vs. Java)

FrankYFTang:
> * @FrankYFTang will submit a new-API ticket for Locale::forUnicodeLocaleID() and LocaleBuilder::setUnicodeLocaleID() which validate according to https://www.unicode.org/reports/tr35/#Unicode_locale_identifier (and to check for LocaleBuilder discrepancies in C++ vs. Java)

For fixing documentation https://unicode-org.atlassian.net/browse/ICU-21729
For new API https://unicode-org.atlassian.net/browse/ICU-21728

icu4c/source/common/uloc.cpp
```diff
@@ -1490,7 +1510,10 @@ _canonicalize(const char* localeID,
 
     if (_hasBCP47Extension(localeID)) {
         do {
-            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeID,
+            localeIDWithHyphens.requestedCapacity = _preprocessLocaleID(localeID, localeIDWithHyphens.getBuffer(), localeIDWithHyphens.getCapacity(), err);
+        } while (localeIDWithHyphens.needToTryAgain(err));
+        do {
+            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, U_SUCCESS(*err) ? localeIDWithHyphens.getBuffer() : localeID,
```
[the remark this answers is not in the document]
FrankYFTang: the problem is then the function would return different result sepending on the memory - but not failing. Both time it will success but return different result. 
I think it is better to return error while the allocation fail, at least you will have a consistent success and consistent failure result and cannot have two different successful results.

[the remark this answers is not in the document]
richgillam: Yeah, I think you're right.  I'll fix that.

FrankYFTang (approved):

markusicu:
Please adjust the PR description to what this change does now.

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
 static const char i_default[] = {'i', '-', 'd', 'e', 'f', 'a', 'u', 'l', 't'};
 #define I_DEFAULT_LENGTH UPRV_LENGTHOF(i_default)
 
+static int32_t _preprocessLocaleID(const char* inLocaleID, char* outBuffer, int32_t bufferCapacity, UErrorCode* err) {
```
markusicu, line 1468: In a function that takes a UErrorCode, normally the first line should be something like
```
if (U_FAILURE(*err)) { return 0; }
```

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+    // this function is to allow permissive parsing in locale IDs in general cases, where the locale ID can be in BCP
+    // 47 format, but with underbars as field delimiters
+    int32_t neededLength = static_cast<int32_t>(uprv_strlen(inLocaleID));
+    if (neededLength > bufferCapacity) {
```
markusicu, line 1472: Why not copy into a CharString?

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+        *err = U_BUFFER_OVERFLOW_ERROR;
+        return neededLength;
+    }
+    
+    uprv_strcpy(outBuffer, inLocaleID);
```
markusicu, line 1477: Possible to not allocate+copy when the string need not be modified?

icu4c/source/common/uloc.cpp
markusicu: Is it necessary to copy+modify the locale ID string? Wouldn't it be easier to parameterize the parser for allowing an underscore despite singleton extensions?

markusicu: Is it necessary to copy+modify the locale ID string? Wouldn't it be easier to parameterize the parser for allowing an underscore despite singleton extensions?

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+    if (uprv_strchr(inLocaleID, '_') != nullptr && ((inLocaleID[0] != 'x' && inLocaleID[0] != 'X') || (inLocaleID[0] != '\0' && inLocaleID[1] != '-'))) {
+        // convert all underbars to hyphens, unless the string starts with "x-" and contains an underbar
```
markusicu, line 1479: What if it starts with "x_" ?

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+    // this function is to allow permissive parsing in locale IDs in general cases, where the locale ID can be in BCP
+    // 47 format, but with underbars as field delimiters
+    int32_t neededLength = static_cast<int32_t>(uprv_strlen(inLocaleID));
+    if (neededLength > bufferCapacity) {
+        *err = U_BUFFER_OVERFLOW_ERROR;
+        return neededLength;
+    }
+    
+    uprv_strcpy(outBuffer, inLocaleID);
+    if (uprv_strchr(inLocaleID, '_') != nullptr && ((inLocaleID[0] != 'x' && inLocaleID[0] != 'X') || (inLocaleID[0] != '\0' && inLocaleID[1] != '-'))) {
```
markusicu, line 1478: No need to check for an empty string (`inLocaleID[0] != '\0'`) after finding that the string contains at least an underscore.

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+        // convert all underbars to hyphens, unless the string starts with "x-" and contains an underbar
+        for (char* p = outBuffer; *p != '\0'; ++p) {
+            *p = (*p == '_') ? '-' : *p;
```
markusicu, line 1481: Don't write to `*p` if you don't need to modify it.

icu4c/source/common/uloc.cpp
```diff
@@ -1465,6 +1465,25 @@ uloc_openKeywords(const char* localeID,
...
+        }
+    }
+    return bufferCapacity;
```
markusicu, line 1484: Confusing: Normally we return the string length, not the buffer capacity.

icu4c/source/test/cintltst/capitst.c
```diff
@@ -2565,4 +2566,18 @@ static void TestStrcollNull(void) {
     ucol_close(coll);
 }
 
+static void TestBogusLocaleID(void) {
+    UErrorCode err = U_ZERO_ERROR;
+    UCollator* c1 = ucol_open("en-US-u-kn-true", &err);
+    UCollator* c2 = ucol_open("en_US-u-kn-true", &err);
+    
+    if (assertSuccess("Failed to create collators", &err)) {
+        assertTrue("Comparison using \"normal\" collator failed", !ucol_greater(c1, u"2", -1, u"10", -1));
```
markusicu, line 2575: Double negative... Weird that we don't have a ucol_less() function, but I would assertFalse(..., ucol_greater(...))

icu4c/source/test/cintltst/capitst.c
```diff
@@ -2565,4 +2566,18 @@ static void TestStrcollNull(void) {
     ucol_close(coll);
 }
 
+static void TestBogusLocaleID(void) {
```
markusicu, line 2569: Clearer would be TestLocaleIDWithUnderscoreAndExtension()

icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java
```diff
@@ -1131,10 +1131,13 @@ private static int getShortestSubtagLength(String localeID) {
      * @stable ICU 3.0
      */
     public static String getName(String localeID){
-        String tmpLocaleID;
+        String tmpLocaleID = localeID;
         // Convert BCP47 id if necessary
         if (localeID != null && !localeID.contains("@") && getShortestSubtagLength(localeID) == 1) {
-            tmpLocaleID = forLanguageTag(localeID).getName();
+            if (!localeID.toLowerCase().startsWith("x-") || localeID.indexOf('_') < 0) {
```
markusicu, line 1137: Different condition from C++: There you test "if contains underscore and not..." but here you test "if ... *or does not contain underscore*". Pretty sure this is wrong :-)

markusicu (dismissed):
sorry needs work I think

icu4c/source/common/uloc.cpp
richgillam: I considered parameterizing the parser, but the logic that would need to change based on the parameter was about six steps away from here in the call chain, and I didn't want to have to change the signatures of all the intervening functions to account for this.  It seemed simpler to just fix up the string at the beginning.

icu4c/source/common/uloc.cpp
richgillam: You know, that actually _would_ be better.  I couldn't figure out a way to use CharString for the original problem I ended up creating PreflightingLocaleIDBuffer for, but I think it _will_ work here, and it's a lot less hassle.  Thanks for the suggestion.

icu4c/source/common/uloc.cpp
richgillam: This was driven by what the unit tests were telling me, and we didn't have any failing unit tests with an underbar in the second position.  Only hyphens.  So I concluded that there was a syntax rule that required that separator to be a hyphen.

However, in trying to get my revised version of this code working correctly, I encountered test failures where the first letter could be something other than "x" or "X", and where the separator could, indeed be an underbar.  I decided to loosen the condition here so that it just checks to see if the second character in the locale ID is `_` or `-`.  This seems to work without apparent side effects.  I'll make the same change in the Java version.

icu4c/source/common/uloc.cpp
richgillam: Why?  I've changed this, but I don't understand why writing to *p hurts anything.  I did it this way for brevity.

icu4c/source/test/cintltst/capitst.c
richgillam: There doesn't seem to be an assertFalse() function in cintltst.c.

icu4c/source/common/uloc.cpp
markusicu: This looks better now, but I still don't quite understand the conditions.
- "x-some-thing" is a valid BCP 47 language tag, and I think that writing it with underscores should work just as well as "de_AT-u-kn".
- "u-some-thing" is not valid BCP 47, nor really meaningful in legacy ICU syntax; we probably don't care how this is handled.

So... are there test cases (that look meaningful) which fail if you *only* check for `uprv_strchr(localeID, '_')` ?

icu4c/source/test/cintltst/capitst.c
markusicu: optional(!): You could add assertFalse() :-)

icu4c/source/common/uloc.cpp
```diff
@@ -1478,7 +1478,7 @@ _canonicalize(const char* localeID,
               uint32_t options,
               UErrorCode* err) {
     int32_t j, fieldCount=0, scriptSize=0, variantSize=0;
-    PreflightingLocaleIDBuffer tempBuffer;
+    PreflightingLocaleIDBuffer tempBuffer;  // if localeID has a BCP47 extension, tmpLocaleID points to this
```
markusicu, line 1481: Before the previous line (that is, as the first thing in the function body) please insert
```
if (U_FAILURE(*err)) { return 0; }
```

icu4c/source/common/uloc.cpp
markusicu: Oh, I think you mention this in the PR description. Without this, "x-klingon" becomes "@x=klingon".
But... @yumaoka isn't that what we would expect?

richgillam: All I can say here is that we had a bunch of unit tests that were specifically expecting us not to mess with "x-klingon" and friends.  I can't cite the specific ones off the top of my head, but if you want, I can back out that change and let you know specifically what failed.

markusicu: I am letting the "x-klingon" question slide. Not important enough to block this good work.

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm tnx

FrankYFTang:
This PR cause the  stack-use-after-scope. in https://unicode-org.atlassian.net/browse/ICU-21749 . I have a fix in https://github.com/unicode-org/icu/pull/1858 . Please either review / approve the fix in that PR or revert this PR ASAP.

icu4c/source/common/uloc.cpp
```diff
@@ -1477,21 +1477,37 @@ _canonicalize(const char* localeID,
...
+                    }
+                }
+                localeIDPtr = localeIDWithHyphens.data();
+            }
+        }
+        
         do {
-            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeID,
-                tempBuffer.getBuffer(), tempBuffer.getCapacity(), err);
+            tempBuffer.requestedCapacity = _ConvertBCP47(tmpLocaleID, localeIDPtr, tempBuffer.getBuffer(),
```
FrankYFTang, line 1509: This line cause tmpLocaleID may point to localeIDPtr which point to a memory inside localeIDWithHyphens. But tmpLocaleID is later used after localeIDWithHyphens is out of scope and cause bug!

PR #1797, opened 2021-08-18
ICU-21038 Run ICU4J localespiCheck as a part of CI build

yumaoka:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21038

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -13,7 +13,7 @@ jobs:
       lfs: true
       fetchDepth: 10
     - script: |
-        echo "Building ICU4J" && cd icu4j && ant init && ant check 
+        echo "Building ICU4J" && cd icu4j && ant init && ant check && ant localespiCheck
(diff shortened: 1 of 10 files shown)


yumaoka:
Java 9+ changed a way to load custom service provider. Until Java 8, it requires a provider jar must be in java's ext.dir. Java 9 no longer support ext.dir, and they can be placed in regular classpath.

Also, previously, Java only support subset of ICU supported locales. The old test code set the assumption. In newer release of Java, they implemented CLDR based locale provider. Unlike ICU, the implementation include all CLDR locales. It is possible to configure Java to skip CLDR based provider, but I prefer to use the default configuration of java runtime, including traditional Java locale implementation + CLDR provider. The test append ICU provider, but it is no longer true that ICU supported locale is super set of Java default (with CLDR provider). These test codes were updated.

markusicu (approved):
lgtm tnx

I didn't try to quite get the isIcuImpl changes, but they seem plausible and I trust you :-)

PR #1798, opened 2021-08-18
ICU-20973 Delete unnecessary deprecated TimeUnitFormat::operator!=().

roubert:
ICU-20973

An operator!=() is already defined by a base class so this class doesn't
need to define it again and not doing so avoids a C++20 ambiguity.

This simplifies the code.


markusicu (approved):
lgtm

markusicu:
In future PRs, please fill out the template rather than just deleting it :-)

PR #1800, opened 2021-08-18
ICU-21710 Remove BOYER_MOORE dead code from usearch.cpp

jefgen:
This is related to the changes in PR #1473.

In ICU 4.0 (see ticket ICU-5420), StringSearch was changed to use a simple linear search algorithm, and it no longer used the Boyer-Moore search algorithm.
In ICU 51.1 (see ticket ICU-9573), the Boyer-Moore string search implementation and test cases were retracted.

The Boyer-Moore search algorithm code in `usearch.cpp` hasn't actually compiled since the Collation V2 rewrite in ~2013/2014. 
(The code behind the `BOYER_MOORE` `#ifdef` refers to non-existent functions and methods that have been changed, making it no longer compile if set to `1`.)

This change completely removes the Boyer-Moore search algorithm code from the file `usearch.cpp`.

Note: The diff shown on GitHub is slightly confused. I didn't actually edit or change the code in the file -- I only made deletions/removals.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21710


docs/userguide/icu4j/faq.md
```diff
@@ -195,7 +195,9 @@ determine whether case and accents are ignored during a search.
 
 #### What algorithm are you using to perform the search?
 
-StringSearch uses a version of the Boyer-Moore search algorithm that has been
+As of ICU 53, StringSearch uses a simple linear search algorithm which
```
markusicu, line 198: Why does it say ICU 53 here, but ICU 4.0 in usearch.h?
I suspect that ICU 4.0 is closer to when B-M was turned off-by-default for hard-to-fix problems.

markusicu (commented):
nice getting rid of the non-functional *half* of the .cpp file!

docs/userguide/icu4j/faq.md
jefgen: I updated this to have a different date for ICU4J, as it looks like the J changes occurred after the C changes.

Based on looking at the tickets and code history, it appears as though B-M was turned off in ICU4C first, and later ported to ICU4J a few releases later.

The ICU4C changes to turn off B-M were done using ticket ICU-5420, which was marked for 4.0. This also aligns with existing comments in the [User Guide](https://github.com/unicode-org/icu/blob/2dce62892b6b00df02195a35ca09ebd8e082e80f/docs/userguide/collation/string-search.md#search-algorithm):

> In ICU4C 4.0, the string search service was updated with the simple linear search algorithm, which locates a match by shifting a cursor in the target text one by one, and these issues were fixed. In ICU4C 4.0.1, the Boyer-Moore search code was reintroduced as a separated API set as a technology preview. In a later release, this code was deleted.

The removal of the additional B-M implementation was done in ICU 51.1 using ticket ICU-9573.

And then later in ICU 53.1 (using ticket ICU-6288), the ICU4C StringSearch changes were ported to ICU4J.

This also was noted on the download page for ICU 53 http://site.icu-project.org/download/53

> **ICU4J Specific Changes**
> - StringSearch code re-implemented
>   - Ported ICU4C implementation to ICU4J, replacing the previous implementation

So I thought it might be good to update the documentation to call out that the J changes occurred later on (in version 53).

markusicu: Thanks for checking!
This is confusing if we just say "ICU 53 StringSearch", even in the *Java* FAQ.

Please change this here to say "As of ICU4J 53 / ICU4C 4.0, ..."
and then usearch.h to say "As of ICU4C 4.0 / ICU4J 53, ..."
and something similar in the User Guide.

jefgen: Thanks! 
I'll force push an update with the changes in a moment.

icu4c/source/i18n/usearch.cpp
```diff
@@ -545,2048 +541,79 @@ inline void setShiftTable(int16_t   shift[], int16_t backshift[],
...
-                // alas... no hope. failed at rearranged accent set
-                cleanUpSafeText(strsrch, safetext, safebuffer);
-                return USEARCH_DONE;
-            }
-            else {
-                if (isSafe) {
-                    failedoffset = safeoffset - failedoffset;
-                    cleanUpSafeText(strsrch, safetext, safebuffer);
-                }
+inline void initialize(UStringSearch *strsrch, UErrorCode *status)
```
aheninger, line 544: The comment for this `initialize()` function says "Building of the pattern collation element list and the boyer moore strsrch table."

Also, the preceding function, `setShiftTable()`, looks like it's setting up Boyer-Moore stuff.

But, in the interest of keeping this PR simple, quick and mechanical, Ietting these slide seems OK.

docs/userguide/collation/string-search.md
```diff
@@ -276,11 +276,12 @@ search service. There were some known issues in these previous releases.
...
-search service was updated with the simple linear search algorithm, which
-locates a match by shifting a cursor in the target text one by one, and these
-issues were fixed. In ICU4C 4.0.1, the Boyer-Moore search code was reintroduced
-as a separated API set as a technology preview. In a later release, this code was deleted.
+In ICU4C 4.0, the string search service was updated to use a simple linear search
+algorithm, which locates a match by shifting a cursor in the target text one by one,
+and these issues were fixed.
+
+In ICU4C 4.0.1, the Boyer-Moore search code was reintroduced as a separate API with
+technology preview status. In ICU 51.1, this code was deleted.
```
markusicu, line 284: ... Similarly, in ICU4J 53 (Java version) the Boyer-Moore search code was replaced by the simple linear search
algorithm.

?

icu4c/source/i18n/usearch.cpp
jefgen: Good point, thanks Andy!

I noticed a few other places in this file where things could be cleaned up / changed as well.
I think I'll create another PR after this one to make some additional changes (using the same ticket number, since they are related to the Boyer-Moore code). That way this PR is just a simple removal, without other changes mixed in with it.

docs/userguide/collation/string-search.md
jefgen: Ah, sorry for missing this... I'll force-push an update to change this.

markusicu (approved):
lgtm tnx

PR #1803, opened 2021-08-19
ICU-21429 Allow timezones with max 2 digits at the end to

messerkb:
This pull request fixes some chromium bugs related to [ICU-21429](https://unicode-org.atlassian.net/browse/ICU-21429).
Especially [Issue 1158789](https://bugs.chromium.org/p/chromium/issues/detail?id=1158789).

I also investigated why `TZ=$(readlink -m /etc/localtime) chromium` does't work, but this usage is not documented in page [Specifying the Time Zone with TZ](https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html) so I think making this work is not in the favor of the ICU project.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21429


yumaoka (approved):
LGTM

PR #1815, opened 2021-08-23
Fix coverity warnings

mike-fabian:
See: https://unicode-org.atlassian.net/browse/ICU-21667
(Issues found by coverity in icu-69.1)

##### Checklist


icu4c/source/common/brkiter.cpp
```diff
@@ -105,7 +105,9 @@ BreakIterator::buildInstance(const Locale& loc, const char *type, UErrorCode &st
         }
     }
 
+    /* coverity[incorrect_free] */
```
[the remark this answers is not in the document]
mike-fabian: I believe this is a false positive. `ures_close(brkName);` does do the right thing as far as I can tell.

`ures_initStackObject(brkName);`

creates an object on the stack.

`ures_close(brkName);` then calls

```
U_CAPI void  U_EXPORT2
ures_close(UResourceBundle* resB)
{
    ures_closeBundle(resB, TRUE);
}
```

and this seems to check correctly whehter it is on the stack or on the heap and does the right thing, I believe: 

```
static void
ures_closeBundle(UResourceBundle* resB, UBool freeBundleObj)
{
    if(resB != NULL) {
        if(resB->fData != NULL) {
            entryClose(resB->fData);
        }
        if(resB->fVersion != NULL) {
            uprv_free(resB->fVersion);
        }
        ures_freeResPath(resB);

        if(ures_isStackObject(resB) == FALSE && freeBundleObj) {
            uprv_free(resB);
        }
#if 0 /*U_DEBUG*/
        else {
            /* poison the data */
            uprv_memset(resB, -1, sizeof(UResourceBundle));
        }
#endif
    }
}
```

So probably this is just too complicated for coverity and is a false positive.

The comment silences coverity about this.

icu4c/source/common/serv.cpp
```diff
@@ -790,6 +790,7 @@ ICUService::getDisplayNames(UVector& result,
      */
     int32_t pos = UHASH_FIRST; 
     const UHashElement *entry = NULL;
+    /* coverity[deref_after_free] */
```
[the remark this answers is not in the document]
mike-fabian: This is also a false positive in my opinion.

Coverity thinks that dnCache might be freed already when this line is reached:

```
while ((entry = dnCache->cache.nextElement(pos)) != NULL) {
```

and then dereferencing the pointer would not be possible.

But when reading the code above, I see that there is a `delete dnCache` **but** then it is allocated again:

```
ncthis->dnCache = new DNCache(locale);
```

and it is properly checked whether the allocation was successfull.

So when the line coverity complains about is reached, there is no way that dnCache is NULL.

So this is a false positive.

Putting 

```
    /* coverity[deref_after_free] */
```
silences coverity about this. `deref_after_free` was the error coverity reported.

icu4c/source/common/uresbund.cpp
```diff
@@ -2515,7 +2515,9 @@ typedef struct ULocalesContext {
 static void U_CALLCONV
 ures_loc_closeLocales(UEnumeration *enumerator) {
     ULocalesContext *ctx = (ULocalesContext *)enumerator->context;
+    /* coverity[address_free] */
```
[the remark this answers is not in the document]
mike-fabian: Another false positive.

```
typedef struct ULocalesContext {
    UResourceBundle installed;
    UResourceBundle curr;
} ULocalesContext;

static void U_CALLCONV
ures_loc_closeLocales(UEnumeration *enumerator) {
    ULocalesContext *ctx = (ULocalesContext *)enumerator->context;
    ures_close(&ctx->curr);
    ures_close(&ctx->installed);
    uprv_free(ctx);
    uprv_free(enumerator);
}
```

This looks perfectly fine to me. 

In `ures_openAvailableLocales()`, these two objects are initialized like this

```
    ures_initStackObject(&myContext->installed);
    ures_initStackObject(&myContext->curr);
```

freeing them as above is OK.

FrankYFTang:
sorry, I was just busy on review the other PR recently. Will take a closer look about this soon.

icu4c/source/common/brkiter.cpp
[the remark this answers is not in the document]
FrankYFTang: By reading source/common/uresimp.h it seems the expected pattern should be
ures_initStackObject(&brkRulesStack);
ures_initStackObject(&brkNameStack);
...
ures_close(&brkRulesStack);
ures_close(&brkNameStack);

brkRules and brkNames could be changed by the code in between, is it correct to ures_close on them?

or should we try to fix this by using StackUResourceBundle instead?
@markusicu

icu4c/source/common/serv.cpp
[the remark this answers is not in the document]
FrankYFTang: I think the problem is 
```
    if (U_SUCCESS(status)) {
```
is FALSE in the beginning we may have dnCache == NULL
I think we should fix it by changing 
``` 
if (U_SUCCESS(status)) {
  // the code in the if block before
}
``` 
to
``` 
if (U_FAILURE(status)) {
  return result;
}
// the code in the if block before
```

icu4c/source/common/uloc_keytype.cpp
```diff
@@ -325,12 +325,20 @@ initFromResourceBundle(UErrorCode& sts) {
             }
         }
         if (U_FAILURE(sts)) {
+            if (typeDataMap != NULL) {
```
FrankYFTang, line 328: could we fix this by using LocalUHashtablePointer for typeDataMap instead?
See  source/common/uhash.h for document and
loadKnownCanonicalized in source/common/locid.cpp for sample usage. 
Replace 
        UHashtable* typeDataMap = uhash_open(uhash_hashIChars, uhash_compareIChars, NULL, &sts);
to
        LocalUHashtablePointer typeDataMap(uhash_open(uhash_hashIChars, uhash_compareIChars, NULL, &sts));

and change code like
                            uhash_put(typeDataMap, (void*)from, t, &sts);
to
                            uhash_put(typeDataMap.getAlias(), (void*)from, t, &sts);
and change
        keyData->typeMap.adoptInstead(typeDataMap);
to
        keyData->typeMap.adoptInstead(typeDataMap.orphan());

@markusicu  IS this right?

icu4c/source/common/uloc_keytype.cpp
```diff
@@ -325,12 +325,20 @@ initFromResourceBundle(UErrorCode& sts) {
...
+                uhash_close(typeDataMap);
+                typeDataMap = NULL;
+            }
             break;
         }
 
         LocExtKeyData* keyData = gLocExtKeyDataEntries->create();
         if (keyData == NULL) {
             sts = U_MEMORY_ALLOCATION_ERROR;
+            if (typeDataMap != NULL) {
```
FrankYFTang, line 338: by using LocalUHashtablePointer as I stated above we won't need these if/uhash_close

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2246,6 +2246,7 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
                 var = (VariantListEntry*)uprv_malloc(sizeof(VariantListEntry));
                 if (var == NULL) {
                     *status = U_MEMORY_ALLOCATION_ERROR;
+                    uprv_free(pExtension);
```
FrankYFTang, line 2249: Since ExtensionListEntry is a struct and defined as
```
struct ExtensionListEntry : public icu::UMemory {
```
could we change this by changing 
```
    ExtensionListEntry *pExtension;
```
to
```
LocalMemory<ExtensionListEntry> pExtension;
```
see  third_party/icu/source/common/cmemory.h
so just let the LocalMemory to handle the free for us.

@markusicu  Is this correct?

icu4c/source/common/umutablecptrie.cpp
```diff
@@ -1543,7 +1543,7 @@ int32_t MutableCodePointTrie::compactTrie(int32_t fastILimit, UErrorCode &errorC
     MixedBlocks mixedBlocks;
     int32_t newDataLength = compactData(fastILimit, newData, newDataCapacity,
                                         dataNullIndex, mixedBlocks, errorCode);
-    if (U_FAILURE(errorCode)) { return 0; }
+    if (U_FAILURE(errorCode)) { uprv_free(newData); return 0; }
```
FrankYFTang, line 1546: same as comment above, fix this by changing 
```
    uint32_t *newData = (uint32_t *)uprv_malloc(newDataCapacity * 4);
```
to
```
    LocalMemory<uint32_t> newData;
    newData.allocateInsteadAndCopy(newDataCapacity);
```
see similar code in https://source.chromium.org/chromium/chromium/src/+/main:third_party/icu/source/tools/genrb/parse.cpp;l=980?q=allocateInsteadAndCopy&ss=chromium%2Fchromium%2Fsrc

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
FrankYFTang: here is what I see, in ures_openAvailableLocales, it is possible that it return NULL, so UEnumeration * enumerator could be a NULL . so from compiler's point of view, the input to this function could be a NULL. and in the case where  ures_openAvailableLocales return NULL, it does not call ures_initStackObject on that null it return (it should not be) . 
I wonder would add a if (enumerator == NULL) return; on the top fix this instead?

icu4c/source/i18n/rbt_pars.cpp
```diff
@@ -557,6 +557,7 @@ int32_t RuleHalf::parseSection(const UnicodeString& rule, int32_t pos, int32_t l
                 // The next character MUST be a segment open
                 if (single == NULL ||
                     !ICU_Utility::parseChar(rule, iref, SEGMENT_OPEN)) {
+                    delete single;
```
FrankYFTang, line 560: could we change 
```
TransliteratorIDParser::SingleID* single =
                    TransliteratorIDParser::parseFilterID(rule, iref);
```
to
```
LocalMemory<TransliteratorIDParser::SingleID> single(
                    TransliteratorIDParser::parseFilterID(rule, iref));
```
to fix this?

icu4c/source/i18n/tridpars.cpp
```diff
@@ -136,6 +136,9 @@ TransliteratorIDParser::parseSingleID(const UnicodeString& id, int32_t& pos,
                 specsB = parseFilterID(id, pos, TRUE);
                 // Must close with a ')'
                 if (specsB == NULL || !ICU_Utility::parseChar(id, pos, CLOSE_REV)) {
+                    if (specsB != NULL) {
```
FrankYFTang, line 139: could we change 
```
Specs* specsA
Specs* specsB
```
to
```
LocalMemory<Specs> specsA;
LocalMemory<Specs> specsB;
```
to fix this?
@markusicu  any opinion?

icu4c/source/i18n/usearch.cpp
```diff
@@ -222,6 +222,9 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
         int32_t *temp = (int32_t *)allocateMemory(
                                          sizeof(int32_t) * newlength, status);
         if (U_FAILURE(*status)) {
+            if (temp != NULL) {
```
FrankYFTang, line 225: Same here. could we use
```
LocalMemory<int32_t> temp;
temp.allocateInsteadAndCopy(newlength);
```
to fix it?

icu4c/source/i18n/usearch.cpp
```diff
@@ -263,6 +266,9 @@ inline int64_t * addTouint64_tArray(int64_t    *destination,
                                          sizeof(int64_t) * newlength, status);
 
         if (U_FAILURE(*status)) {
+            if (temp != NULL) {
```
FrankYFTang, line 269: Change to use  LocalMemory<int32_t> temp above would remove the need to do so.

icu4c/source/i18n/uspoof_impl.cpp
```diff
@@ -196,6 +196,9 @@ void SpoofImpl::setAllowedLocales(const char *localesList, UErrorCode &status) {
     tmpSet = allowedChars.clone();
     const char *tmpLocalesList = uprv_strdup(localesList);
     if (tmpSet == NULL || tmpLocalesList == NULL) {
+        if (tmpLocalesList != NULL) {
```
FrankYFTang, line 199: If tmpSet is NULL then there are no point to call  uprv_strdup(localesList); right?
How about we change them to  
```
tmpSet = allowedChars.clone();
if (tmpSet == nullptr) { 
return; 
}
const char *tmpLocalesList = uprv_strdup(localesList);
if (tmpLocalesList == nullptr) { 
return; 
}
```

icu4c/source/tools/makeconv/genmbcs.cpp
```diff
@@ -173,7 +173,7 @@ MBCSOpen(UCMFile *ucm) {
     }
 
     MBCSInit(mbcsData, ucm);
-    return &mbcsData->newConverter;
+    return (NewConverter *)mbcsData;
```
FrankYFTang, line 176: This change will break if someone
swap the order of 
```
    NewConverter newConverter;
    UCMFile *ucm;
```
in MBCSData right?

shouldn't changing it to
```
return &(mbcsData->newConverter);
```
be enough?

FrankYFTang:
really appreciate you work on this. There are a lot of old code which use raw pointers we should migrate to use smart pointer instead.

FrankYFTang:
@mike-fabian  thank you for working on this. I will take a 15 days vacation Sep 19-Oct 4 and won't be able to review during that period of time. The freeze for ICU70.1 is coming soon. If you can revise this PR this week I may be able to review again and land it after you address all the issues. Thanks

icu4c/source/common/loclikely.cpp
```diff
@@ -1408,7 +1408,9 @@ ulocimp_getRegionForSupplementalData(const char *localeID, UBool inferRegion,
         }
     }
 
-    rgBuf[rgLen] = 0;
+    if (rgLen >= 0) {
+        rgBuf[rgLen] = 0;
+    }
     uprv_strncpy(region, rgBuf, regionCapacity);
```
markusicu, line 1414: There is no way that rgLen could be negative here.
However, I believe that there is a chance that rgLen == ULOC_RG_BUFLEN (from a bogus locale ID via uloc_getCountry()) in which case we must not write the terminating NUL here.
Also, we don't need the NUL if we use memcpy(), and u_terminateChars() will, well, terminate the region buffer if there is room.

```suggestion
    if (U_SUCCESS(*status) && rgLen <= regionCapacity) {
        uprv_memcpy(region, rgBuf, rgLen);
    }
```

icu4c/source/common/lsr.cpp
```diff
@@ -89,13 +89,15 @@ int32_t LSR::indexForRegion(const char *region) {
         int32_t b = region[1] - '0';
         if (b < 0 || 9 < b) { return 0; }
         c = region[2] - '0';
-        if (c < 0 || 9 < c || region[3] != 0) { return 0; }
+        if (uprv_strlen(region) > 3) { return 0; }
+        if (c < 0 || 9 < c) { return 0; }
```
markusicu, line 93: This change is unnecessary and inefficient. Please revert.

icu4c/source/common/lsr.cpp
```diff
@@ -89,13 +89,15 @@ int32_t LSR::indexForRegion(const char *region) {
...
         return (10 * a + b) * 10 + c + 1;
     } else {  // letters: "DE"
         a = uprv_upperOrdinal(c);
         if (a < 0 || 25 < a) { return 0; }
         int32_t b = uprv_upperOrdinal(region[1]);
-        if (b < 0 || 25 < b || region[2] != 0) { return 0; }
+        if (uprv_strlen(region) > 2) { return 0; }
+        if (b < 0 || 25 < b) { return 0; }
```
markusicu, line 100: This change is unnecessary and inefficient. Please revert.

icu4c/source/common/serv.cpp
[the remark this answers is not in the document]
markusicu: I agree with Frank's proposed change, but I also suspect that Coverity gets confused between the implicit `this`, which is const, and `ncthis`, which stands for “non-const this”.

So I suggest the following which hopefully makes the annotation comment unnecessary:
- make Frank's change
- then split the function into two
  - move all of the code into a new, non-const getDisplayNamesImpl(...)
  - remove all `ncthis->`
  - change the original function to just const_cast and call the impl function

```C++
UVector& 
ICUService::getDisplayNames(UVector& result, 
                            const Locale& locale, 
                            const UnicodeString* matchID, 
                            UErrorCode& status) const {
    // cast away semantic const
    return const_cast<ICUService *>(this)->getDisplayNamesImpl(result, locale, matchID, status);
}

UVector& 
ICUService::getDisplayNamesImpl(UVector& result, 
                            const Locale& locale, 
                            const UnicodeString* matchID, 
                            UErrorCode& status)
{
    if (U_FAILURE(status)) { return result; }
    result.removeAllElements();
    result.setDeleter(userv_deleteStringPair);
    Mutex mutex(&lock);
    ...
```

icu4c/source/common/uloc_keytype.cpp
markusicu: > @markusicu IS this right?

lgtm

For that last line, the following is nicer:
```
        keyData->typeMap = std::move(typeDataMap);
```

icu4c/source/common/uloc_tag.cpp
markusicu: > @markusicu Is this correct?

Almost. Since it's a UMemory and thus uses new/delete, we should use LocalPointer (unicode/localpointer.h).
LocalMemory is for raw types where we use uprv_malloc/uprv_free.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: I agree with adding
```
if (enumerator == nullptr) { return; }
```
at the very start of the function.
If we then still need the annotations, then ok.

icu4c/source/i18n/decNumber.h
```diff
@@ -86,7 +86,7 @@
                          /* range: -1999999997 through 999999999      */
     uint8_t bits;        /* Indicator bits (see above)                */
                          /* Coefficient, from least significant unit  */
-    decNumberUnit lsu[DECNUMUNITS];
+    decNumberUnit lsu[DECNUMUNITS+2];
```
markusicu, line 89: @sffc if this is necessary, then it should be upstreamed

icu4c/source/i18n/rbt_pars.cpp
markusicu: LocalPointer, not LocalMemory

Then
```
if (single.isNull() || !ICU_Util...()) {
    return syntaxError(...);
}
```
Following code also would no longer `delete single`.

icu4c/source/i18n/tridpars.cpp
markusicu: smart pointers are always safer

in this case, the change looks ok; i would actually `delete specsB` unconditionally

icu4c/source/i18n/usearch.cpp
markusicu: same as above

smart pointers good, but you could also unconditionally `uprv_free(temp)`

icu4c/source/tools/makeconv/genmbcs.cpp
markusicu: The parentheses in `return &(mbcsData->newConverter);` are redundant.

Why this change?

markusicu (changes requested):

icu4c/source/i18n/decNumber.h
sffc: ICU4C *is* the upstream for decNumber.

However, I would like to understand more about this change.

1. Could you provide a test case please that exercises the case where DECNUMUNITS is not sufficient?
2. Instead of adding `+2` here, I think it would be cleaner if you instead changed the definition of `DECNUMUNITS` itself, a few lines earlier, and updated the comment to explain why you need the extra 2 units.

sffc (changes requested):

FrankYFTang:
ping

mike-fabian:
> ping

I am trying do you requested changes now and rebase to current main.

icu4c/source/common/uloc_tag.cpp
mike-fabian: Something like @markusicu suggested seems to be already in current main.

icu4c/source/common/umutablecptrie.cpp
mike-fabian: @FrankYFTang like this?:
```
--- a/icu4c/source/common/umutablecptrie.cpp
+++ b/icu4c/source/common/umutablecptrie.cpp
@@ -1531,7 +1531,8 @@ int32_t MutableCodePointTrie::compactTrie(int32_t fastILimit, UErrorCode &errorC
         errorCode = U_MEMORY_ALLOCATION_ERROR;
         return 0;
     }
-    uint32_t* newData = static_cast<uint32_t*>(uprv_malloc(newDataCapacity * 4));
+    LocalMemory<uint32_t> buffer;
+    uint32_t *newData = buffer.allocateInsteadAndCopy(newDataCapacity * 4);
     if (newData == nullptr) {
         errorCode = U_MEMORY_ALLOCATION_ERROR;
         return 0;
```

mike-fabian: No, that doesn't work, this gives me an error when running `make check` ☹.

icu4c/source/i18n/rbt_pars.cpp
mike-fabian: @FrankYFTang like this?:
```
diff --git a/icu4c/source/i18n/rbt_pars.cpp b/icu4c/source/i18n/rbt_pars.cpp
index 2c2fd3a414f..b2ced2f1926 100644
--- a/icu4c/source/i18n/rbt_pars.cpp
+++ b/icu4c/source/i18n/rbt_pars.cpp
@@ -552,17 +552,15 @@ int32_t RuleHalf::parseSection(const UnicodeString& rule, int32_t pos, int32_t l
         case ALT_FUNCTION:
             {
                 int32_t iref = pos;
-                TransliteratorIDParser::SingleID* single =
-                    TransliteratorIDParser::parseFilterID(rule, iref);
+                LocalPointer<TransliteratorIDParser::SingleID> single(
+                    TransliteratorIDParser::parseFilterID(rule, iref));
                 // The next character MUST be a segment open
-                if (single == nullptr ||
+                if (single.isNull() ||
                     !ICU_Utility::parseChar(rule, iref, SEGMENT_OPEN)) {
-                    delete single;
                     return syntaxError(U_INVALID_FUNCTION, rule, start, status);
                 }
                 
                 Transliterator *t = single->createInstance();
-                delete single;
                 if (t == nullptr) {
                     return syntaxError(U_INVALID_FUNCTION, rule, start, status);
                 }
```

icu4c/source/i18n/uspoof_impl.cpp
mike-fabian: @FrankYFTang like this?:

```diff
--- a/icu4c/source/i18n/uspoof_impl.cpp
+++ b/icu4c/source/i18n/uspoof_impl.cpp
@@ -194,8 +194,12 @@ void SpoofImpl::setAllowedLocales(const char *localesList, UErrorCode &status) {
 
     // Store the updated spoof checker state.
     tmpSet = allowedChars.clone();
+    if (tmpSet == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return;
+    }
     const char *tmpLocalesList = uprv_strdup(localesList);
-    if (tmpSet == nullptr || tmpLocalesList == nullptr) {
+    if (tmpLocalesList == nullptr) {
         status = U_MEMORY_ALLOCATION_ERROR;
         return;
     }
```

icu4c/source/tools/makeconv/genmbcs.cpp
mike-fabian: I thought it fixes this coverity warning:

```
Error: RESOURCE_LEAK (CWE-772): [#def25](https://unicode-org.atlassian.net/browse/ICU-21667#def25)
icu/source/tools/makeconv/genmbcs.cpp:169: alloc_fn: Storage is returned from allocation function "uprv_malloc_69".
icu/source/tools/makeconv/genmbcs.cpp:169: var_assign: Assigning: "mbcsData" = storage returned from "uprv_malloc_69(456528UL)".
icu/source/tools/makeconv/genmbcs.cpp:175: noescape: Resource "mbcsData" is not freed or pointed-to in "MBCSInit".
icu/source/tools/makeconv/genmbcs.cpp:176: leaked_storage: Variable "mbcsData" going out of scope leaks the storage it points to.

     174|   
     175|       MBCSInit(mbcsData, ucm);
     176|>     return &mbcsData>newConverter;
     177|   }
     178|
```

icu4c/source/common/loclikely.cpp
mike-fabian: This part of the code does not seem to exist in current `main` anymore, so this can be omitted.

icu4c/source/common/lsr.cpp
mike-fabian: OK.

icu4c/source/common/lsr.cpp
mike-fabian: OK.

icu4c/source/common/serv.cpp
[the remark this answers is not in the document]
mike-fabian: @FrankYFTang @markusicu like this?:
```diff
--- a/icu4c/source/common/serv.cpp
+++ b/icu4c/source/common/serv.cpp
@@ -722,49 +722,57 @@ UVector&
 ICUService::getDisplayNames(UVector& result, 
                             const Locale& locale, 
                             const UnicodeString* matchID, 
-                            UErrorCode& status) const 
+                            UErrorCode& status) const
 {
+    // cast away semantic const
+    return const_cast<ICUService *>(this)->getDisplayNamesImpl(result, locale, matchID, status);
+}
+
+UVector&
+ICUService::getDisplayNamesImpl(UVector& result,
+                            const Locale& locale,
+                            const UnicodeString* matchID,
+                            UErrorCode& status)
+{
+    if (U_FAILURE(status)) { return result; }
     result.removeAllElements();
     result.setDeleter(userv_deleteStringPair);
-    if (U_SUCCESS(status)) {
-        ICUService* ncthis = const_cast<ICUService*>(this); // cast away semantic const
-        Mutex mutex(&lock);
+    Mutex mutex(&lock);
 
-        if (dnCache != nullptr && dnCache->locale != locale) {
-            delete dnCache;
-            ncthis->dnCache = nullptr;
-        }
+    if (dnCache != nullptr && dnCache->locale != locale) {
+        delete dnCache;
+        dnCache = nullptr;
+    }
 
+    if (dnCache == nullptr) {
+        const Hashtable* m = getVisibleIDMap(status);
+        if (U_FAILURE(status)) {
+            return result;
+        }
+        dnCache = new DNCache(locale);
         if (dnCache == nullptr) {
-            const Hashtable* m = getVisibleIDMap(status);
-            if (U_FAILURE(status)) {
-                return result;
-            }
-            ncthis->dnCache = new DNCache(locale); 
-            if (dnCache == nullptr) {
-                status = U_MEMORY_ALLOCATION_ERROR;
-                return result;
-            }
+            status = U_MEMORY_ALLOCATION_ERROR;
+            return result;
+        }
 
-            int32_t pos = UHASH_FIRST;
-            const UHashElement* entry = nullptr;
-            while ((entry = m->nextElement(pos)) != nullptr) {
-                const UnicodeString* id = static_cast<const UnicodeString*>(entry->key.pointer);
-                ICUServiceFactory* f = static_cast<ICUServiceFactory*>(entry->value.pointer);
-                UnicodeString dname;
-                f->getDisplayName(*id, locale, dname);
-                if (dname.isBogus()) {
-                    status = U_MEMORY_ALLOCATION_ERROR;
-                } else {
-                    dnCache->cache.put(dname, (void*)id, status); // share pointer with visibleIDMap
-                    if (U_SUCCESS(status)) {
-                        continue;
-                    }
+        int32_t pos = UHASH_FIRST;
+        const UHashElement* entry = nullptr;
+        while ((entry = m->nextElement(pos)) != nullptr) {
+            const UnicodeString* id = static_cast<const UnicodeString*>(entry->key.pointer);
+            ICUServiceFactory* f = static_cast<ICUServiceFactory*>(entry->value.pointer);
+            UnicodeString dname;
+            f->getDisplayName(*id, locale, dname);
+            if (dname.isBogus()) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+            } else {
+                dnCache->cache.put(dname, (void*)id, status); // share pointer with visibleIDMap
+                if (U_SUCCESS(status)) {
+                    continue;
                 }
-                delete dnCache;
-                ncthis->dnCache = nullptr;
-                return result;
             }
+            delete dnCache;
+            dnCache = nullptr;
+            return result;
         }
     }
 
diff --git a/icu4c/source/common/serv.h b/icu4c/source/common/serv.h
index 9aea548fc3a..3c53f228738 100644
--- a/icu4c/source/common/serv.h
+++ b/icu4c/source/common/serv.h
@@ -759,6 +759,11 @@ class U_COMMON_API ICUService : public ICUNotifier {
                              const UnicodeString* matchID, 
                              UErrorCode& status) const;
 
+    UVector& getDisplayNamesImpl(UVector& result,
+                             const Locale& locale,
+                             const UnicodeString* matchID,
+                             UErrorCode& status);
+
     /**
      * <p>A convenience override of registerInstance(UObject*, const UnicodeString&, UBool)
      * that defaults visible to true.</p>
```

icu4c/source/common/uloc_keytype.cpp
mike-fabian: @FrankYFTang @markusicu like this?:

```diff
--- a/icu4c/source/common/uloc_keytype.cpp
+++ b/icu4c/source/common/uloc_keytype.cpp
@@ -144,7 +144,7 @@ initFromResourceBundle(UErrorCode& sts) {
 
         bool isTZ = uprv_strcmp(legacyKeyId, "timezone") == 0;
 
-        UHashtable* typeDataMap = uhash_open(uhash_hashIChars, uhash_compareIChars, nullptr, &sts);
+        LocalUHashtablePointer typeDataMap(uhash_open(uhash_hashIChars, uhash_compareIChars, nullptr, &sts));
         if (U_FAILURE(sts)) {
             break;
         }
@@ -253,10 +253,10 @@ initFromResourceBundle(UErrorCode& sts) {
                 t->bcpId = bcpTypeId;
                 t->legacyId = legacyTypeId;
 
-                uhash_put(typeDataMap, (void*)legacyTypeId, t, &sts);
+                uhash_put(typeDataMap.getAlias(), (void*)legacyTypeId, t, &sts);
                 if (bcpTypeId != legacyTypeId) {
                     // different type value
-                    uhash_put(typeDataMap, (void*)bcpTypeId, t, &sts);
+                    uhash_put(typeDataMap.getAlias(), (void*)bcpTypeId, t, &sts);
                 }
                 if (U_FAILURE(sts)) {
                     break;
@@ -296,7 +296,7 @@ initFromResourceBundle(UErrorCode& sts) {
                                     from = fromBuf->data();
                                 }
                             }
-                            uhash_put(typeDataMap, (void*)from, t, &sts);
+                            uhash_put(typeDataMap.getAlias(), (void*)from, t, &sts);
                         }
                     }
                     if (U_FAILURE(sts)) {
@@ -318,7 +318,7 @@ initFromResourceBundle(UErrorCode& sts) {
                         // check if this is an alias of bcp type
                         if (uprv_compareInvWithUChar(nullptr, bcpTypeId, -1, to, toLen) == 0) {
                             const char* from = ures_getKey(bcpTypeAliasDataEntry.getAlias());
-                            uhash_put(typeDataMap, (void*)from, t, &sts);
+                            uhash_put(typeDataMap.getAlias(), (void*)from, t, &sts);
                         }
                     }
                     if (U_FAILURE(sts)) {
@@ -339,7 +339,7 @@ initFromResourceBundle(UErrorCode& sts) {
         keyData->bcpId = bcpKeyId;
         keyData->legacyId = legacyKeyId;
         keyData->specialTypes = specialTypes;
-        keyData->typeMap.adoptInstead(typeDataMap);
+        keyData->typeMap = std::move(typeDataMap);
 
         uhash_put(gLocExtKeyMap, (void*)legacyKeyId, keyData, &sts);
         if (legacyKeyId != bcpKeyId) {
```

icu4c/source/common/uloc_keytype.cpp
mike-fabian: OK.

icu4c/source/i18n/tridpars.cpp
mike-fabian: @FrankYFTang @markusicu like this?:

```diff
--- a/icu4c/source/i18n/tridpars.cpp
+++ b/icu4c/source/i18n/tridpars.cpp
@@ -136,6 +136,7 @@ TransliteratorIDParser::parseSingleID(const UnicodeString& id, int32_t& pos,
                 specsB = parseFilterID(id, pos, true);
                 // Must close with a ')'
                 if (specsB == nullptr || !ICU_Utility::parseChar(id, pos, CLOSE_REV)) {
+                    delete specsB;
                     delete specsA;
                     pos = start;
                     return nullptr;
```

icu4c/source/i18n/usearch.cpp
mike-fabian: @FrankYFTang @markusicu like this?:
```diff
--- a/icu4c/source/i18n/usearch.cpp
+++ b/icu4c/source/i18n/usearch.cpp
@@ -199,6 +199,7 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
         int32_t* temp = static_cast<int32_t*>(allocateMemory(
                                          sizeof(int32_t) * newlength, status));
         if (U_FAILURE(*status)) {
+            uprv_free(temp);
             return nullptr;
         }
         uprv_memcpy(temp, destination, sizeof(int32_t) * (size_t)offset);
@@ -240,6 +241,7 @@ inline int64_t * addTouint64_tArray(int64_t    *destination,
                                          sizeof(int64_t) * newlength, status));
 
         if (U_FAILURE(*status)) {
+            uprv_free(temp);
             return nullptr;
         }
 
```

icu4c/source/tools/makeconv/genmbcs.cpp
mike-fabian: Like this?:
```c++
index 774cdc6045d..a996ac5aa8d 100644
--- a/icu4c/source/tools/makeconv/genmbcs.cpp
+++ b/icu4c/source/tools/makeconv/genmbcs.cpp
@@ -173,6 +173,10 @@ MBCSOpen(UCMFile *ucm) {
     }
 
     MBCSInit(mbcsData, ucm);
+    /* The memory in the MBSData structure following
+     * newConverter will be properly freed in MBCSClose.
+     */
+    /* coverity[leaked_storage] */
     return &mbcsData->newConverter;
 }
```

mike-fabian:
@FrankYFTang @markusicu 

I have rebased this pull request on current main.

I have also addressed all things mentioned in your review comments.

Can you please review this again?

mike-fabian (commented):
@FrankYFTang @markusicu

I have rebased this pull request on current main.

I have also addressed all things mentioned in your review comments.

Can you please review this again?

PR #1817, opened 2021-08-24
ICU-21655 widen Java character APIs from char to int arguments

markusicu:
Widen the following APIs from `char` to `int`:
- class UTF16
  - public static boolean isSurrogate(int c) {
  - public static boolean isTrailSurrogate(int c) {
  - public static boolean isLeadSurrogate(int c) {

- class UCharacter
  - public static int getCodePoint(int lead, int trail)
  - public static boolean isHighSurrogate(int ch) {
  - public static boolean isLowSurrogate(int ch) {
  - public static final boolean isSurrogatePair(int high, int low) {
  - public static final int toCodePoint(int high, int low) {

I did not comprehensively look for call sites that precede such function calls with
“is BMP” or similar range checks that would now be unnecessary.

In most cases, the behavior should be the same.
In some cases, this change might fix obscure bugs when
such function calls _should have been preceded_ with “is BMP” checks.

I did not look for call sites like `Character.isHighSurrogate((char) c)`
that could now be simplified using ICU class UTF16 or UCharacter.

Some tests continue to use `char` values, for coverage where
someone might retain binary-compatibility overloads.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21655

### DIFF
--- a/docs/userguide/dev/codingguidelines.md
+++ b/docs/userguide/dev/codingguidelines.md
@@ -1551,6 +1551,22 @@ itself public can be placed in different places:
 4. If it is used by multiple packages, make it public and place the class in
    `the com.ibm.icu.impl` package.
 
+### ICU4J API Stability
+
+General discussion: See [ICU Design / ICU API compatibility](../icu/design.md#icu-api-compatibility).
+
+Occasionally, we “broaden” or “widen” a Java API by making a parameter broader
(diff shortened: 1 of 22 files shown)


yumaoka (approved):
LGTM

PR #1831, opened 2021-08-27
ICU-21737 replaced subtraction double with subtraction bigDecimal

VladimirZhdanov:
replaced subtraction double with subtraction bigDecimal

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/NFSubstitution.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/NFSubstitution.java
@@ -9,6 +9,7 @@
 package com.ibm.icu.text;
 
 import java.text.ParsePosition;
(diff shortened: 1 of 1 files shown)


grhoten (changes requested):
There are several issues with this change.

1. There is no test.
2. The same test and fix needs to be applied to ICU4C.
3. This is only a small part of the issue at hand. There are other math operations by
4. This should be reviewed for performance differences, since BigDecimal math is known to be much slower.
5. The equivalent math can probably be done without BigDecimal, but it probably requires a bit of bit twiddling on the IEEE754 format.

I don't recommend merging this.

grhoten:
I'm closing this for now. I'll probably get a fix done at some point. If a more complete solution is created by anyone else, a new pull request can be opened that provides a more complete solution.

PR #1826, opened 2021-08-26
ICU-20769 refactor UResourceBundle.fResData & init_resb_result()

markusicu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20769
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #1841, opened 2021-09-01
ICU-20769 getAliasTargetAsResourceBundle() only needs ResourceData

markusicu:
getAliasTargetAsResourceBundle() does not need the whole actual locale's UResourceDataEntry, only its ResourceData.

It looks up the alias string (from ResourceData+Resource) and then does not need anything from the alias source bundle at all. It used to also get the "path" (for locating the bundle "tree") from the source bundle dataEntry, but the path is the same for all bundles in the tree, so we can just as well get it from the validLocaleDataEntry which we need anyway for sideways "/LOCALE/keyPath" aliases.

(This should have been part of PR #1826)

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20769


richgillam (approved):
I like this; it's a lot cleaner.  I'll make the necessary changes to my PR.

PR #1825, opened 2021-08-26
ICU-21734 Add missing overrides

stampho:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21734
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #1842, opened 2021-09-01
ICU-21734 Add -Wsuggest-override to  Build only WarningsAsErrors buildbot  

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21734

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -62,7 +62,7 @@ jobs:
         sudo apt install -qy clang-13 lld-13 libc++-13-dev libc++abi-13-dev
       displayName: 'Install Clang-13'
     - script: |
-        export CPPFLAGS="-Werror -Wall -Wextra -Wextra-semi -Wundef -Wnon-virtual-dtor" && cd icu4c/source && ./runConfigureICU Linux && make -j2 tests
+        export CPPFLAGS="-Werror -Wall -Wextra -Wextra-semi -Wundef -Wnon-virtual-dtor -Wsuggest-override" && cd icu4c/source && ./runConfigureICU Linux && make -j2 tests
       displayName: 'Build only (WarningsAsErrors)'
       env:
         CC: clang-13


FrankYFTang:
@jefgen  @stampho @markusicu  This PR is for making sure once https://github.com/unicode-org/icu/pull/1825 is landed we will not regress from it . I would suggest we merge the change to the stampho's  2nd PR which have the unresolve conversial issues

jefgen:
Thanks for creating this PR Frank! 

Hm... it looks like the build bots aren't happy with the new flag though.

Doing some searching it looks like this flag was only very recently added to Clang (https://reviews.llvm.org/D82728 from Jun 2020).
However, GCC seems to have this flag for some time (GCC 5).

So it looks like our options for adding this are:
- Add a new GCC WarningsAsErrors build bot. (In addition to the clang one).
- Upgrade the existing Clang WarningsAsErrors to clang-13 (similar to how the [C++20 build bot](https://github.com/unicode-org/icu/blob/044ed90934a30f7add886a55e12ad5164334432f/.ci-builds/.azure-pipelines.yml#L160) uses the new clang version).

I don't have a strong preference either way though. What do you think?

markusicu:
> So it looks like our options for adding this are:
> 
> * Add a new GCC WarningsAsErrors build bot. (In addition to the clang one).
> * Upgrade the existing Clang WarningsAsErrors to clang-13 (similar to how the [C++20 build bot](https://github.com/unicode-org/icu/blob/044ed90934a30f7add886a55e12ad5164334432f/.ci-builds/.azure-pipelines.yml#L160) uses the new clang version).
> 
> I don't have a strong preference either way though. What do you think?

I have some preference for upgrading an existing build bot.

jefgen:
> I have some preference for upgrading an existing build bot.

Thanks. I feel like we might have some more (new) warnings generated as a result of updating the compiler, so I'll file a separate ticket for upgrading the WarningsAsErrors build bot to Clang-13 and Ubuntu 20.
(Once that is done we can add the new flag to it.)

jefgen (dismissed):
lgtm, thanks for adding this. :)

Minor suggestion: Maybe rename the commit/title to something like "Add buildbot" instead of "Try out buildbot".

FrankYFTang:
Please approv again my earlier commit/PR title are in full with ascii and github does not like it.

jefgen (approved):
lgtm, thanks for adding this

PR #1843, opened 2021-09-01
ICU-21581 integrate CLDR release-40-alpha2 to ICU trunk

pedberg-icu:
##### Checklist

Integrate CLDR data tagged as release-40-alpha2 to ICU. Notable changes (in CLDR and ICU):
- This changes the BCP47 short zone name "gaza" to "gazastrp", deprecating the old name
- This adds data mapping the CLDR metazone names to BCP47 short IDs of 4 characters. Per discussion, the ICU code that uses the BCP47 mapping will ignore these for the time being (they are the only mappings to 4-character BCP47 zone IDs), until in a future version of CLDR some of the CLDR metazoan names are changed to avoid conflicts with IANA zone aliases.
- This updates a lot of language data (which codes are deprecated, etc.).


icu4c/source/common/uloc_keytype.cpp
```diff
@@ -222,6 +222,11 @@ initFromResourceBundle(UErrorCode& sts) {
                 if (U_FAILURE(sts)) {
                     break;
                 }
+                // TODO: For now we skip loading entries mapping to metazone short ids,
+                // which have 4 characters.
+                if (isTZ && uBcpTypeId.length() == 4) {
+                    continue;
+                }
```
srl295, line 229: Should there be a followon issue noted here? (ICU known issue)

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -201,6 +201,16 @@ ant copy-cldr-testdata
...
+# 4e. For the time being, manually re-add the lstm entries in data/brkitr/root.txt
+
+open $ICU4C_DIR/source/data/brkitr/root.txt 
+
+# paste the following block at the end, after the dictionaries block>
+    lstm{
+        Thai{"Thai_graphclust_model4_heavy.res"}
+        Mymr{"Burmese_graphclust_model5_heavy.res"}
+    }
+
```
srl295, line 213: Presumably landed in https://unicode-org.atlassian.net/browse/ICU-21569 — is there a followon ticket for automating this?

icu4c/source/data/lang/en.txt
```diff
@@ -954,6 +954,13 @@ en{
             no{"Sort Digits Individually"}
             yes{"Sort Digits Numerically"}
         }
+        colReorder{
+            currency{"Currency"}
+            digit{"Digits"}
+            punct{"Punctuation"}
+            space{"Whitespace"}
+            symbol{"Symbol"}
+        }
```
srl295, line 963: I assume this was the revert you mentioned

icu4c/source/i18n/zonemeta.cpp
```diff
@@ -290,9 +290,13 @@ ZoneMeta::getCanonicalCLDRID(const UnicodeString &tzid, UErrorCode& status) {
...
     ures_getByKey(rb, gTimezoneTag, rb, &tmpStatus);
     ures_getByKey(rb, id, rb, &tmpStatus);
-    if (U_SUCCESS(tmpStatus)) {
+    int32_t bcp47StrLen = 0;
+    ures_getString(rb, &bcp47StrLen, &tmpStatus);
+    if (U_SUCCESS(tmpStatus) && bcp47StrLen!=4) {
         // type entry (canonical) found
         // the input is the canonical ID. resolve to const UChar*
+        // TODO: For now we skip entries mapping to metazone short ids,
+        // which have 4 characters.
```
srl295, line 299: TODO should link to a bug

icu4c/source/i18n/zonemeta.cpp
```diff
@@ -933,7 +937,11 @@ ZoneMeta::getShortIDFromCanonical(const UChar* canonicalID) {
     ures_getByKey(rb, gTimezoneTag, rb, &status);
     shortID = ures_getStringByKey(rb, tzidKey, NULL, &status);
     ures_close(rb);
-
+    if (U_FAILURE(status) || (shortID!=nullptr && u_strlen(shortID)==4)) {
+        // TODO: For now we skip entries mapping to metazone short ids,
+        // which have 4 characters.
+        return NULL;
+    }
```
srl295, line 944: ditto

srl295 (approved):
LGTM with suggestions for future

macchiati (approved):

icu4c/source/common/uloc_keytype.cpp
pedberg-icu: It is actually primarily a CLDR issue, filed https://unicode-org.atlassian.net/browse/CLDR-15023

icu4c/source/data/cldr-icu-readme.txt
pedberg-icu: Well, automating this would involve putting the relevant lines in data/xml/brkitr/root.xml, and I started doing that as part of CLDR 40 integrations. The problem is that then creates a dependency that requires adding the data to the ICU4J jars, and the desire was not to do that. The data is currently just for testing so it can be enabled with a build flag, but for that to work these lines have to be present without going through the usual mechanism for adding support for new brkitr resources.

icu4c/source/data/lang/en.txt
pedberg-icu: No, this was for a separate issue, https://unicode-org.atlassian.net/browse/CLDR-15009. The revert I mentioned does not appear in this PR because it was reverted before I ever committed anything. It was a temporary change in CollationRuleParser::parseSetting to look up keywords using both "co" and "collation"; I needed to do that to get ICU to build at all, in order to be able to figure out what was really wrong. Once I figured out and fixed the real problem (in CLDR bcp47/timezone.xml), then I reverted the CollationRuleParser::parseSetting change (which never made it to a commit).

icu4c/source/i18n/zonemeta.cpp
pedberg-icu: Yeah, that is https://unicode-org.atlassian.net/browse/CLDR-15023 which I had not filed at the time. I will add the reference during the next CLDR integration (for alpha3 or beta1)

sffc (commented):
LGTM

PR #1850, opened 2021-09-07
ICU-21730 Port `testGetPreferencesFor` ICU4C tests to ICU4J

younies:
closes: https://github.com/icu-units/icu/issues/129

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21730

### DIFF
--- a/icu4c/source/test/intltest/units_data_test.cpp
+++ b/icu4c/source/test/intltest/units_data_test.cpp
@@ -20,6 +20,7 @@ class UnitsDataTest : public IntlTest {
     void runIndexedTest(int32_t index, UBool exec, const char *&name, char *par = NULL);
 
     void testGetUnitCategory();
+    // This is a sanity check that only exists in ICU4C.
     void testGetAllConversionRates();
     void testGetPreferencesFor();
 };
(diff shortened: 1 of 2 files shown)


younies:
@sffc , testGetAllConversionRates it does not make sense, because of this I have deleted it from C++.

icu4c/source/test/intltest/units_data_test.cpp
```diff
@@ -60,21 +58,6 @@ void UnitsDataTest::testGetUnitCategory() {
     }
 }
 
-void UnitsDataTest::testGetAllConversionRates() {
```
sffc, line 63: This looks like a data sanity test.  I don't know exactly why this test existed, but I think we should probably keep it.  However, we don't need to port it to Java.  It's okay to have tests like this one in C++ but not in Java.

sffc (commented):
@hugovdm also mentioned `testGetUnitCategory` in https://github.com/icu-units/icu/issues/129.  Do we need to port that one as well?

icu4c/source/test/intltest/units_data_test.cpp
younies: @sffc , `testGetUnitCategory` is added in this PR: https://github.com/unicode-org/icu/pull/1839

For `testGetAllConversionRates`, we already testing this in testing the `UnitsConverter`, and the test should be always for the public API, except we need to make more deep testing.

sffc: I prefer that we keep the test only in C++ and add a comment explaining that.

sffc (changes requested):
Please restore the deleted test, but LGTM otherwise

sffc (approved):

PR #1856, opened 2021-09-10
ICU-21662 UVector Error Handling in Regex

aheninger:
Clean up some oversights from commit 0ec329c. This was triggered by fuzz testing finding
a memory leak with the original commit, see https://oss-fuzz.com/testcase-detail/4656781834452992

I was unable to replicate the fuzzing failure, but reviewing the nearby code showed
some likely problems.

In this commit,
- Fix UStack::pop() to not delete the popped element when a deleter function is present.
  This was a bug, but because there were no stacks with deleters, was not causing trouble.

- Change RegexCompile::compileSet() to delete the set if it cannot be added to the internal
  vector of sets. I suspect this is the cause of the fuzzing leak - 0ec329c changed the
  behavior of UVector in the presence of errors.

- Changed RegexCompile::fSetStack to use an object deleter function. This fixes the
  leak checking at the points new elements are pushed onto this stack.

##### Checklist

### DIFF
--- a/icu4c/source/i18n/regexcmp.cpp
+++ b/icu4c/source/i18n/regexcmp.cpp
@@ -53,7 +53,7 @@ U_NAMESPACE_BEGIN
 //
 //------------------------------------------------------------------------------
 RegexCompile::RegexCompile(RegexPattern *rxp, UErrorCode &status) :
-   fParenStack(status), fSetStack(status), fSetOpStack(status)
+   fParenStack(status), fSetStack(uprv_deleteUObject, nullptr, status), fSetOpStack(status)
 {
     // Lazy init of all shared global sets (needed for init()'s empty text)
     RegexStaticSets::initGlobals(&status);
@@ -278,11 +278,6 @@ void    RegexCompile::compile(
 
     if (U_FAILURE(*fStatus)) {
         // Bail out if the pattern had errors.
-        //   Set stack cleanup:  a successful compile would have left it empty,
-        //   but errors can leave temporary sets hanging around.
-        while (!fSetStack.empty()) {
-            delete (UnicodeSet *)fSetStack.pop();
-        }
         return;
     }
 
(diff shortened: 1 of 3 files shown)


markusicu:
Looks plausible, but you labeled this as "incomplete". I assume you are still debugging.

icu4c/source/i18n/regexcmp.cpp
```diff
@@ -53,7 +53,7 @@ U_NAMESPACE_BEGIN
 //
 //------------------------------------------------------------------------------
 RegexCompile::RegexCompile(RegexPattern *rxp, UErrorCode &status) :
-   fParenStack(status), fSetStack(status), fSetOpStack(status)
```
jefgen, line 56: Since the `fSetStack` stack has a deleter function now, and calling `pop` doesn't delete the element, does this also mean that the 3 calls to `fSetStack.pop()` on lines 4625, 4632, and 4638 will need to explicitly call `delete` on the popped element(s)?

https://github.com/unicode-org/icu/blob/main/icu4c/source/i18n/regexcmp.cpp#L4628-L4641

jefgen (commented):
Thanks for these fixes/changes!

Sorry for missing the issue with `UStack::pop()` in the review on 0ec329c.

icu4c/source/i18n/regexcmp.cpp
aheninger: I think everything is OK.

For non-error cases, the logic was OK before, when fSetStack did not have a deleter, and is unchanged. Either way, ``fSetStack.pop()`` doesn't delete.  ``pop()`` can't delete - it's defined to return the popped item, which would be meaningless if it were already deleted.

At line 4616 ``rightOperand = (UnicodeSet *)fSetStack.peek();``
grabs the top of stack. rightOperand is subsequently deleted when needed.

The main reason for adding the deleter function to fSetStack was to make the error recovery code around pushing items onto the stack work correctly.  At line 4656, where push wants to adopt, including delete-on-error.

markusicu (approved):
lgtm

icu4c/source/i18n/regexcmp.cpp
jefgen: Ah, I missed the `rightOperand` (which gets deleted in the 3 cases where `pop()` is called).
Thanks Andy!

jefgen (approved):
Thanks Andy!

PR #1857, opened 2021-09-11
ICU-21545 Add plumbing for running icuexportdata in BRS

sffc:
Rename the new icuwriteuprops tool to icuexportdata, in preparation for possible future changes for writing data other than for Unicode properties.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21545


sffc:
I added a CI build to run the tool in Azure Pipelines and upload the artifacts.  The artifacts can be found here:

https://dev.azure.com/ms/icu/_build/results?buildId=220775&view=artifacts&pathAsName=false&type=publishedArtifacts

I also added a few more features to icuwriteuprops while I was at it.

sffc:
CC @iainireland -- This PR pre-builds the property files and saves them in a zip file, which will be uploaded to the ICU download page, such that ICU4X clients don't need to build ICU4C and run the tool manually.

.ci-builds/.azure-pipelines.yml
```diff
@@ -582,3 +582,44 @@ jobs:
...
+    - script: |
+        cd icu4c/source && ./runConfigureICU Linux && make -j2
+      displayName: 'Build'
+      env:
+        CC: clang
+        CXX: clang++
+    - script: |
+        mkdir icu4c/source/full_out
+        cd icu4c/source && LD_LIBRARY_PATH=lib ./bin/icuwriteuprops -A -I -c -d full_out
+        zip -rj icuwriteuprops_full.zip full_out/
```
sffc, line 604: I picked the filename arbitrarily.  Do you think it should be something else, perhaps with the ICU version in it?

icu4c/source/tools/icuwriteuprops/icuwriteuprops.cpp
```diff
@@ -129,17 +163,36 @@ int main(int argc, char* argv[]) {
...
+    for (int i=1; i<argc; i++) {
+        propNames.push_back(argv[i]);
+    }
+    if (options[OPT_ALL].doesOccur) {
+        for (int i=UCHAR_BINARY_START; i<UCHAR_INT_LIMIT; i++) {
+            if (i == UCHAR_BINARY_LIMIT) {
+                i = UCHAR_INT_START;
+            }
+            UProperty uprop = static_cast<UProperty>(i);
+            const char* propName = u_getPropertyName(uprop, U_LONG_PROPERTY_NAME);
```
iainireland, line 176: Currently ICU4X uses the short name for the .toml file, so that we can determine the name of the file directly from the key. We can work around that if you think the long name is better for the download, though.

sffc: I used the long name because I realized that there are a few properties (the POSIX properties "alnum", "blank", etc) that don't have a short name.

markusicu: I recently stumbled on that, too. They should probably have a non-empty short name, and it should be the same as the long name -- like for other properties where we didn't invent unique short names. Care to submit an ICU ticket?

markusicu: In the meantime, that should not stop you from using the short name *where available*.

.ci-builds/.azure-pipelines.yml
markusicu: how about uprops_full.zip ? the "icuwrite" prefix is good for the tool but weird for its output.

.ci-builds/.azure-pipelines.yml
```diff
@@ -582,3 +582,44 @@ jobs:
...
+      displayName: 'Build full property dump file'
+    # In the sample file, include:
+    # - Basic binary properties: ASCII_Hex_Digit Basic_Emoji White_Space
+    # - Property of strings: RGI_Emoji
```
markusicu, line 608: Basic_Emoji also has strings. Therefore, you may want to not also add RGI_Emoji to samples.

.ci-builds/.azure-pipelines.yml
```diff
@@ -582,3 +582,44 @@ jobs:
...
+        cd icu4c/source && ./runConfigureICU Linux && make -j2
+      displayName: 'Build'
+      env:
+        CC: clang
+        CXX: clang++
+    - script: |
+        mkdir icu4c/source/full_out
+        cd icu4c/source && LD_LIBRARY_PATH=lib ./bin/icuwriteuprops -A -I -c -d full_out
+        zip -rj icuwriteuprops_full.zip full_out/
+      displayName: 'Build full property dump file'
```
markusicu, line 605: optional: I would s/dump/data/g

markusicu:
FYI: Some properties are ICU-specific (e.g., lccc & tccc), and some (alnum, blank, ...) are based on UTS 18 but may or may not match the ES regex spec for them. Output as is anyway?

.ci-builds/.azure-pipelines.yml
sffc: Hmm, I was just thinking that we may want to add additional data that is not just unicode properties (e.g., collation or segmentation data).  So maybe "structures_full.zip" ?

markusicu: I would much prefer separate uprops_full.zip vs. collation_data.zip etc.

icu4c/source/tools/icuexportdata/icuexportdata.cpp
```diff
@@ -129,28 +166,61 @@ int main(int argc, char* argv[]) {
...
+            "usage: %s -m uprops [-options] [-A | properties...]\n"
             "\tdump Unicode property data to .toml files\n"
             "options:\n"
             "\t-h or -? or --help  this usage text\n"
             "\t-V or --version     show a version message\n"
-            "\t-c or --copyright   include a copyright notice\n"
+            "\t-m or --mode        mode: currently only 'uprops', but more may be added\n"
             "\t-t or --trie-type   set the trie type (small or fast, default small)\n"
             "\t-d or --destdir     destination directory, followed by the path\n"
+            "\t-A or --all         write out all properties known to icuexportdata\n"
```
markusicu, line 210: Do you plan to keep "all" and have it write whatever that might mean for the selected mode?
Or should "all" be "all-props" and only work with "mode=uprops"?

icu4c/source/tools/icuexportdata/icuexportdata.cpp
```diff
@@ -129,28 +166,61 @@ int main(int argc, char* argv[]) {
...
+            "\t-I or --index       write an _index.toml summarizing all data exported\n"
```
markusicu, line 211: By the way, if you don't want to define a single-letter option, you can use something like `'\1'` in the UOPTION_DEF -- see other tools where I have done this. (Control code 1 is hard to type; and I don't advertise a single letter in the help text in such a case.)

icu4c/source/tools/icuexportdata/icuexportdata.cpp
```diff
@@ -129,28 +166,61 @@ int main(int argc, char* argv[]) {
...
+                    std::cerr << "Note: falling back to long name for: " << propName << std::endl;
+                }
+            }
+            if (propName != NULL) {
+                propNames.push_back(propName);
+            }
+        }
+    }
+
+    if(propNames.empty() || options[OPT_HELP_H].doesOccur || options[OPT_HELP_QUESTION_MARK].doesOccur || !options[OPT_MODE].doesOccur) {
```
markusicu, line 199:
```suggestion
    if(propNames.empty() || options[OPT_HELP_H].doesOccur || options[OPT_HELP_QUESTION_MARK].doesOccur ||
            !options[OPT_MODE].doesOccur) {
```

icu4c/source/tools/icuexportdata/icuexportdata.cpp
```diff
@@ -129,28 +166,61 @@ int main(int argc, char* argv[]) {
...
         FILE *stdfile=argc<0 ? stderr : stdout;
         fprintf(stdfile,
-            "usage: %s [-options] properties...\n"
+            "usage: %s -m uprops [-options] [-A | properties...]\n"
             "\tdump Unicode property data to .toml files\n"
             "options:\n"
             "\t-h or -? or --help  this usage text\n"
             "\t-V or --version     show a version message\n"
-            "\t-c or --copyright   include a copyright notice\n"
+            "\t-m or --mode        mode: currently only 'uprops', but more may be added\n"
```
markusicu, line 207: It looks like `--mode` is required but the CI bot config does not use it. Does it work?

sffc: Fixed.

icu4c/source/tools/icuexportdata/icuexportdata.cpp
sffc: I was thinking about that.  I think that if `--all` makes sense for some other mode added in the future, we would just re-use the same flag.  Likewise, if some other piece of data also contains CPTs, then `--trie-type` will be valid.  But if we are dumping some kind of data where those options are irrelevant, they would simply be ignored.

markusicu (commented):
lgtm pse squash

There are a couple of open suggestions (one is totally optional), but I won't hold the PR for them.

icu4c/source/tools/icuwriteuprops/icuwriteuprops.cpp
sffc: I was a little worried about using the short name and then falling back to the long name, because then the paths within the zip file might change between versions, but this should not be a major concern because code that looks up a particular property in the zip file should be using the `[binary_property].short_name` property in the TOML files.

So I have changed this to create files with the short name by default.

sffc:
> lgtm pse squash
> 
> There are a couple of open suggestions (one is totally optional), but I won't hold the PR for them.

I wasn't finished yet. :smiley: I have now addressed the remaining comments and pushed a new commit, and I'm ready for a re-review.

markusicu (commented):
tnx lgtm pse squash

markusicu (approved):
lgtm

PR #1858, opened 2021-09-11
ICU-21749 Fix stack-use-after-scope bug in uloc

FrankYFTang:
Change the _ConvertBCP47 function to return the new string
instead of modify the pointer of the first arg to make such behavior
obvious. Add a new pointer arg in the end to get the length instead.
Move the stack allocated memory to the right scope.

The bug is caused by https://github.com/unicode-org/icu/pull/1795 rich landed 10 days ago

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21749


FrankYFTang:
@richgillam @pedberg-icu @markusicu please review

pedberg-icu (approved):
Thanks and good catch

PR #1862, opened 2021-09-15
ICU-21683 ICU-21674 Clear state from UFormattedNumber[Range] in C API

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21683, https://unicode-org.atlassian.net/browse/ICU-21674

### DIFF
--- a/icu4c/source/i18n/formattedval_sbimpl.cpp
+++ b/icu4c/source/i18n/formattedval_sbimpl.cpp
@@ -96,6 +96,11 @@ void FormattedValueStringBuilderImpl::getAllFieldPositions(FieldPositionIterator
     }
 }
 
+void FormattedValueStringBuilderImpl::resetString() {
+    fString.clear();
+    spanIndicesCount = 0;
+}
+
 // Signal the end of the string using a field that doesn't exist and that is
 // different from kUndefinedField, which is used for "null field".
 static constexpr Field kEndField = Field(0xf, 0xf);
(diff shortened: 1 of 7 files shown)


icu4c/source/i18n/formattedval_sbimpl.cpp
```diff
@@ -283,14 +288,14 @@ bool FormattedValueStringBuilderImpl::nextPositionImpl(ConstrainedFieldPosition&
 
 void FormattedValueStringBuilderImpl::appendSpanInfo(UFieldCategory category, int32_t spanValue, int32_t start, int32_t length, UErrorCode& status) {
     if (U_FAILURE(status)) { return; }
-    U_ASSERT(spanIndices.getCapacity() >= spanValue);
-    if (spanIndices.getCapacity() == spanValue) {
-        if (!spanIndices.resize(spanValue * 2, spanValue)) {
+    U_ASSERT(spanIndices.getCapacity() >= spanIndicesCount);
+    if (spanIndices.getCapacity() == spanIndicesCount) {
+        if (!spanIndices.resize(spanIndicesCount * 2, spanIndicesCount)) {
```
FrankYFTang, line 293: Do you also need to changre the spanValue to spanIndicesCount in the following three lines inside the next function prependSpanInfo ?

```
    U_ASSERT(spanIndices.getCapacity() >= spanValue);
    if (spanIndices.getCapacity() == spanValue) {
        if (!spanIndices.resize(spanValue * 2, spanValue)) {
```

icu4c/source/i18n/formattedval_sbimpl.cpp
```diff
@@ -96,6 +96,11 @@ void FormattedValueStringBuilderImpl::getAllFieldPositions(FieldPositionIterator
     }
 }
 
+void FormattedValueStringBuilderImpl::reset() {
+    fString.clear();
+    spanIndicesCount = 0;
```
FrankYFTang, line 101: does it make sense to call it "reset"? If it's functionality is truely to "reset" FormattedValueStringBuilderImpl then should it also clear out the  fNumericField and spanIndices too? If fNumericField and spanIndices should not be clear out, then maybe the name "reset" does not truly reflect what you like it to do.

FrankYFTang: Also I see there are subclass adding more fields to the object, so if it is named "reset" I would expect the function also reset the extra fields the subclass added. Maybe call it resetXXX to "limit" what the scope it intend to reset of it's member data?

sffc:
@markusicu can you take this one?  @FrankYFTang was in the middle of the review but is now OOO.

markusicu (approved):
lgtm

PR #1863, opened 2021-09-15
ICU-21684 Fix performance issue in NumberRangeFormatter

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21684

### DIFF
--- a/icu4c/source/test/intltest/numbertest_range.cpp
+++ b/icu4c/source/test/intltest/numbertest_range.cpp
@@ -53,6 +53,7 @@ void NumberRangeFormatterTest::runIndexedTest(int32_t index, UBool exec, const c
         TESTCASE_AUTO(testCopyMove);
         TESTCASE_AUTO(toObject);
         TESTCASE_AUTO(testGetDecimalNumbers);
+        TESTCASE_AUTO(test21684_Performance);
         TESTCASE_AUTO(test21358_SignPosition);
     TESTCASE_AUTO_END;
 }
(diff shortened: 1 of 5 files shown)


icu4c/source/test/intltest/numbertest_range.cpp
```diff
@@ -910,6 +911,14 @@ void NumberRangeFormatterTest::testGetDecimalNumbers() {
     }
 }
 
+void NumberRangeFormatterTest::test21684_Performance() {
+    IcuTestErrorCode status(*this, "test21684_Performance");
+    LocalizedNumberRangeFormatter lnf = NumberRangeFormatter::withLocale("en");
+    // The following two lines of code should finish quickly.
+    lnf.formatFormattableRange({"-1e99999", status}, {"0", status}, status);
+    lnf.formatFormattableRange({"0", status}, {"1e99999", status}, status);
+}
```
pedberg-icu, line 920: This does not actually generate an error if the execution time is long. Is the idea for anyone running the tests to just look at the logged execution time and do something if it seems too long? Maybe we should at least have a infoln stating that; but that would not affect CI tests. Maybe better to call uprv_getRawUTCtime() at the beginning and end of the test and generate an error if the difference is significantly greater than expected (i.e. 20 seconds?)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberRangeFormatterTest.java
```diff
@@ -854,6 +855,14 @@ public void testNumberingSystemRangeData() {
         }
     }
 
+    @Test
+    public void test21684_Performance() {
+        LocalizedNumberRangeFormatter lnf = NumberRangeFormatter.withLocale(ULocale.ENGLISH);
+        // The following two lines of code should finish quickly.
+        lnf.formatRange(new BigDecimal("-1e99999"), new BigDecimal("0"));
+        lnf.formatRange(new BigDecimal("0"), new BigDecimal("1e99999"));
+    }
```
pedberg-icu, line 864: Same idea as above about generating an error if execution too long...

icu4c/source/test/intltest/numbertest_range.cpp
sffc: The problem with checking clock time is that the clock time really differs a lot depending on the environment. Not only CPU speed, but also whether the library was built with ASAN, run with Valgrind, etc. So as of now I basically use this as a manual qualitative runtime test.

I can either:

1. Remove the test, since the bug is fixed
2. Keep the test and add a flaky clock time checker
3. Keep the test as proposed

pedberg-icu: OK, I guess #3 seems best.

pedberg-icu (approved):

PR #1867, opened 2021-09-17
ICU-11891 make UnicodeSet.toPattern() well-formed & round-trip

markusicu:
UnicodeSet.toPattern() changes:
- Always escape surrogate code points, so that the pattern string is well-formed UTF-16.
  - For good measure (and after checking with the ICU team), toPattern() now also always escapes control codes (gc=Cc) and noncharacter code points, which could otherwise trip up consumers of the pattern string.
- When a range a-b has exactly the two code points a and b, then it is normally written without the dash (just ab). However, when these are the last lead surrogate and the first trail surrogate (U+DBFF and U+DC00), we do need to write the dash (a-b) so that the two do not combine into a surrogate pair.
- When a range a-f is followed by a range q-x, f is a lead surrogate, and q is a trail surrogate, then we have to do *something* so that f and q do not combine into a surrogate pair.
  - Merely escaping the surrogate code points is not sufficient, because the unescaping in applyPattern() (just like in the Java lexer) combines a pair of lead+trail surrogates into a supplementary code point *even when one or both are escaped*.
  - Mark suggested writing a space, but toPattern() has never added spaces before.
  - Instead, I am changing the order in which the ranges are written. (The order of elements in a set or its pattern is irrelevant.)
  - All ranges ending with a lead surrogate are postponed to after those that start with a trail surrogate.
  - If there is a range that starts with a lead surrogate and ends with a trail surrogate, it could be written immediately, or it could be postponed. Postponing it is easier to implement (and the code easier to understand).
- I moved the code for writing a range into a helper function, since there are now more places in the logic that write various ranges. (And this code got extended by the additional ab check mentioned above.)
- The toPattern() implementation used to have two code paths for whether the pattern was to be written with the actual ranges vs. as the complement of the not-contained ranges.
  - Keeping it that way would have required duplicating the new logic for avoiding surrogate pairs, with some confusing range iteration because the complement path used (getRangeEnd(previous)+1, getRangeStart(current)-1) pairs.
  - I changed the iteration to read directly from the inversion list. This by itself is less readable, but by reading (start, limit) pairs starting from index 1 instead of from index 0 we automatically get the complement of the ranges. Otherwise the logic is exactly the same.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-11891

### DIFF
--- a/icu4c/source/common/util.cpp
+++ b/icu4c/source/common/util.cpp
@@ -65,38 +65,52 @@ UnicodeString& ICU_Utility::appendNumber(UnicodeString& result, int32_t n,
     return result;
 }
 
-/**
- * Return true if the character is NOT printable ASCII.
- */
 UBool ICU_Utility::isUnprintable(UChar32 c) {
     return !(c >= 0x20 && c <= 0x7E);
 }
 
-/**
- * Escape unprintable characters using \uxxxx notation for U+0000 to
- * U+FFFF and \Uxxxxxxxx for U+10000 and above.  If the character is
- * printable ASCII, then do nothing and return FALSE.  Otherwise,
- * append the escaped notation and return TRUE.
- */
+UBool ICU_Utility::shouldAlwaysBeEscaped(UChar32 c) {
+    if (c < 0x20) {
+        return true;  // C0 control codes
+    } else if (c <= 0x7e) {
+        return false;  // printable ASCII
+    } else if (c <= 0x9f) {
+        return true;  // C1 control codes
+    } else if (c < 0xd800) {
+        return false;  // most of the BMP
+    } else if (c <= 0xdfff || (0xfdd0 <= c && c <= 0xfdef) || (c & 0xfffe) == 0xfffe) {
+        return true;  // surrogate or noncharacter code points
+    } else if (c <= 0x10ffff) {
+        return false;  // all else
+    } else {
+        return true;  // not a code point
+    }
+}
+
 UBool ICU_Utility::escapeUnprintable(UnicodeString& result, UChar32 c) {
     if (isUnprintable(c)) {
-        result.append(BACKSLASH);
-        if (c & ~0xFFFF) {
-            result.append(UPPER_U);
-            result.append(DIGITS[0xF&(c>>28)]);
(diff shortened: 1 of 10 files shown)


markusicu:
@poulsbo Do you enjoy a blast from the past? ;-)

markusicu:
Amended to fix a regex utility unit test expectation.

poulsbo:
> @poulsbo Do you enjoy a blast from the past? ;-)

Nice!

icu4c/source/common/util.cpp
```diff
@@ -65,38 +65,52 @@ UnicodeString& ICU_Utility::appendNumber(UnicodeString& result, int32_t n,
...
     return !(c >= 0x20 && c <= 0x7E);
 }
 
-/**
- * Escape unprintable characters using \uxxxx notation for U+0000 to
- * U+FFFF and \Uxxxxxxxx for U+10000 and above.  If the character is
- * printable ASCII, then do nothing and return FALSE.  Otherwise,
- * append the escaped notation and return TRUE.
- */
+UBool ICU_Utility::shouldAlwaysBeEscaped(UChar32 c) {
```
srl295, line 72: Clearer.

srl295 (approved):
Looks good

jefgen:
FYI, it looks like the ICU4J Exhaustive tests started failing:
https://dev.azure.com/ms/icu/_build/results?buildId=222672&view=logs&j=0a6b8993-9b96-51bf-2c08-98136845003b&t=2b6ec40e-5fb4-56b1-5b23-808ec3c6d9b0

```
BUILD FAILED
/home/vsts/work/1/s/icu4j/build.xml:493: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:500: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:488: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:674: The following error occurred while executing this line:
/home/vsts/work/1/s/icu4j/build.xml:383: test failed: translit
```

markusicu:
> FYI, it looks like the ICU4J Exhaustive tests started failing:

Sorry and thanks --> PR #1870

PR #1869, opened 2021-09-20
ICU-21141 Fix titlecase of accented Dutch ij digraph

echeran:
Currently done for ICU4J, will do C++ port(s) if logic & code seem correct.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21141

### DIFF
--- a/icu4c/source/common/ucasemap.cpp
+++ b/icu4c/source/common/ucasemap.cpp
@@ -420,6 +420,96 @@ void toUpper(int32_t caseLocale, uint32_t options,
 
 #if !UCONFIG_NO_BREAK_ITERATION
 
+namespace {
+
+constexpr uint8_t ACUTE_BYTE0 = u8"\u0301"[0];
+
+constexpr uint8_t ACUTE_BYTE1 = u8"\u0301"[1];
+
+/**
+ * Input: c is a letter I with or without acute accent.
+ * start is the index in src after c, and is less than segmentLimit.
+ * If a plain i/I is followed by a plain j/J,
+ * or an i/I with acute (precomposed or decomposed) is followed by a j/J with acute,
+ * then we output accordingly.
+ *
+ * @return the src index after the titlecased sequence, or the start index if no Dutch IJ
+ */
+int32_t maybeTitleDutchIJ(const uint8_t *src, UChar32 c, int32_t start, int32_t segmentLimit,
+                          ByteSink &sink, uint32_t options, icu::Edits *edits, UErrorCode &errorCode) {
+
+    int32_t index = start;
+    bool withAcute = false;
+
+    // If the conditions are met, then the following variables tell us what to output.
+    int32_t unchanged1 = 0;  // code units before the j, or the whole sequence (0..3)
+    bool doTitleJ = false;  // true if the j needs to be titlecased
+    int32_t unchanged2 = 0;  // after the j (0 or 1)
+
+    // next character after the first letter
+    UChar32 c2;
+    c2 = src[index++];
+
+    // Is the first letter an i/I with accent?
+    if (c == u'I') {
+        if (c2 == ACUTE_BYTE0 && index < segmentLimit && src[index++] == ACUTE_BYTE1) {
+            withAcute = true;
+            unchanged1 = 2;  // ACUTE is 2 code units in UTF-8
+            if (index == segmentLimit) { return start; }
+            c2 = src[index++];
+        }
+    } else {  // Í
+        withAcute = true;
+    }
+
+    // Is the next character a j/J?
+    if (c2 == u'j') {
+        doTitleJ = true;
+    } else if (c2 == u'J') {
+        ++unchanged1;
+    } else {
+        return start;
+    }
+
+    // A plain i/I must be followed by a plain j/J.
+    // An i/I with acute must be followed by a j/J with acute.
+    if (withAcute) {
+        if ((index + 1) >= segmentLimit || src[index++] != ACUTE_BYTE0 || src[index++] != ACUTE_BYTE1) {
+            return start;
+        }
+        if (doTitleJ) {
+            unchanged2 = 2;  // ACUTE is 2 code units in UTF-8
+        } else {
+            unchanged1 = unchanged1 + 2;    // ACUTE is 2 code units in UTF-8
+        }
+    }
+
+    // There must not be another combining mark.
+    if (index < segmentLimit) {
+        int32_t cp;
+        int32_t i = index;
+        U8_NEXT(src, i, segmentLimit, cp);
+        uint32_t typeMask = U_GET_GC_MASK(cp);
+        if ((typeMask & U_GC_M_MASK) != 0) {
+            return start;
+        }
+    }
+
+    // Output the rest of the Dutch IJ.
+    ByteSinkUtil::appendUnchanged(src + start, unchanged1, sink, options, edits, errorCode);
+    start += unchanged1;
+    if (doTitleJ) {
+        ByteSinkUtil::appendCodePoint(1, u'J', sink, edits);
+        ++start;
+    }
+    ByteSinkUtil::appendUnchanged(src + start, unchanged2, sink, options, edits, errorCode);
+
+    U_ASSERT(start + unchanged2 == index);
+    return index;
+}
+
+}  // namespace
+
 U_CFUNC void U_CALLCONV
 ucasemap_internalUTF8ToTitle(
         int32_t caseLocale, uint32_t options, BreakIterator *iter,
@@ -505,18 +595,13 @@ ucasemap_internalUTF8ToTitle(
 
                 /* Special case Dutch IJ titlecasing */
                 if (titleStart+1 < index &&
-                        caseLocale == UCASE_LOC_DUTCH &&
-                        (src[titleStart] == 0x0049 || src[titleStart] == 0x0069)) {
-                    if (src[titleStart+1] == 0x006A) {
-                        ByteSinkUtil::appendCodePoint(1, 0x004A, sink, edits);
-                        titleLimit++;
-                    } else if (src[titleStart+1] == 0x004A) {
-                        // Keep the capital J from getting lowercased.
-                        if (!ByteSinkUtil::appendUnchanged(src+titleStart+1, 1,
-                                                           sink, options, edits, errorCode)) {
-                            return;
-                        }
-                        titleLimit++;
+                    caseLocale == UCASE_LOC_DUTCH) {
+                    if (c < 0) {
+                        c = ~c;
+                    }
+
+                    if (c == u'I' || c == u'Í') {
+                        titleLimit = maybeTitleDutchIJ(src, c, titleLimit, index, sink, options, edits, errorCode);
                     }
                 }
 
--- a/icu4c/source/common/ustrcase.cpp
+++ b/icu4c/source/common/ustrcase.cpp
@@ -36,6 +36,12 @@
 #include "ustr_imp.h"
 #include "uassert.h"
 
+/**
+ * Code point for COMBINING ACUTE ACCENT
+ * @internal
+ */
+#define ACUTE u'\u0301'
+
 U_NAMESPACE_BEGIN
 
 namespace {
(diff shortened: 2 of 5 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
```diff
@@ -779,6 +822,57 @@ public static String toTitle(int caseLocale, int options, BreakIterator iter, Ch
         }
     }
 
+    /**
+     * Returns length of Latin i or I with acute accent, and 0 if not exact match.
+     */
+    public static int nextAccentedILength(CharSequence s, int index, int titleBreakLimit) {
+        if (index >= titleBreakLimit) {
```
markusicu, line 829: not possible in the call site, no need to check

icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
```diff
@@ -779,6 +822,57 @@ public static String toTitle(int caseLocale, int options, BreakIterator iter, Ch
...
+            return 0;
+        }
+        char charAtIndex = s.charAt(index);
+        if ((charAtIndex == LOWER_I_WITH_ACUTE)
+                || (charAtIndex == UPPER_I_WITH_ACUTE)){
+            return 1;
+        } else if ((index+1 < titleBreakLimit)
+                && (charAtIndex == 'i' || charAtIndex == 'I')
+                && (s.charAt(index+1) == COMBINING_ACCENT_ACUTE)) {
+                return 2;
```
markusicu, line 839:
```suggestion
            return 2;
```

markusicu (commented):
I think this mostly works, and it works like we discussed, but it's a frightening amount of code “just for Dutch IJ”. It's also reading and checking some characters multiple times. Let me try to think about how to do it with less.

markusicu (commented):
I hacked together a version that I think is more readable, and does not read and check characters multiple times. It replaces your version. There are TODOs for constants and for a new StringContextIterator method. Let me know what you think.

icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
```diff
@@ -733,25 +739,62 @@ public static String toTitle(int caseLocale, int options, BreakIterator iter, Ch
 
                         // Special case Dutch IJ titlecasing
```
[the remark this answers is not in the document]
echeran: Done, although I also added a `else if (~c == 'I' || ~c == 'Í')` condition because when the input started with a capital 'I', the integer value `c` seemed to be the bitwise-negation of the code point integer value.

icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
```diff
@@ -779,6 +822,57 @@ public static String toTitle(int caseLocale, int options, BreakIterator iter, Ch
         }
     }
 
+    /**
```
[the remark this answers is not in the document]
echeran: Done.  \u308 is the combining diaeresis, but the pre-existing class-level private constant `ACUTE` was already set to \u301. No changes needed for the code here.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UCharacterCaseTest.java
```diff
@@ -467,6 +470,72 @@ public void TestDutchTitle() {
...
                 "IJssel Igloo IJMUIdEN IPoD IJenough",
                 UCharacter.toTitleCase(LOC_DUTCH, "ijssel igloo IjMUIdEN iPoD ijenough", iter, options));
+
+        // Accented IJ testing
+
+        String[][] dutchIJCasesData = {
+                // input,           expected,         expOnlyChanged
+                {"ij",              "IJ",             "IJ"},
+                {"IJ",              "IJ",             ""},
+                {"íj́",              "ÍJ́",             "ÍJ"},
```
echeran, line 480: When the input is the lowercase accented digraph (`"íj́"`), the omit-unchanged string omits the accent mark after the capital 'J'. (`"ÍJ"`, whereas full titlecase string is `"ÍJ́"`) That means the changed region splits a grapheme cluster, right? If so, is that expected behavior for users?

echeran:
FYI: I'm running `./intltest utility/UnicodeStringTest/StringCaseTest/TestDutchTitle` for the new unit test. The test passed up until the next-to-last commit `dda4337` because it was just porting test cases for functionality prior to the change in this PR. The last commit introduces the full set of test cases, which break, and other tests in the larger set `./intltest utility/UnicodeStringTest/StringCaseTest` also show regressions.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UCharacterCaseTest.java
```diff
@@ -467,6 +470,72 @@ public void TestDutchTitle() {
         assertEquals("Dutch titlecase check in Dutch with nolowercase option",
```
markusicu, line 470: FYI -- I looked for the C/C++ test function equivalent to this. It looks like there was no C++ version, so it's good to just port it, and intltest/strcase.cpp is right for that.

What we do have is a C (cintltst) function, but it would be unnecessarily inconvenient to extend *that* with the new cases. https://github.com/unicode-org/icu/blob/main/icu4c/source/test/cintltst/cstrcase.c#L334

And we also have a data-driven test, but that is also hard for specific API coverage: https://github.com/unicode-org/icu/blob/main/icu4c/source/test/testdata/casing.txt#L46

markusicu:
When Java and C++ UTF-16 look good, there is also C++ UTF-8 :-)
https://github.com/unicode-org/icu/blob/main/icu4c/source/common/ucasemap.cpp#L506

markusicu (commented):
A few Java comments from this afternoon. Will look at C++ tomorrow.

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -667,6 +670,120 @@ StringCaseTest::TestTitleOptions() {
...
+        const UnicodeString expectedFull = cas.expectedFull;
+        const UnicodeString expectedOnlyChanged = cas.expectedOnlyChanged;
+
+        for (const auto& isTitlecaseWholeString : {true, false}) {
+            for (const auto& isOnlyChanged : {true, false}) {
+                uint32_t testOptions = U_TITLECASE_NO_LOWERCASE
+                    | (isTitlecaseWholeString ? U_TITLECASE_WHOLE_STRING : 0)
+                    | (isOnlyChanged ? U_OMIT_UNCHANGED_TEXT : 0);
+                
+                UnicodeString expected = isOnlyChanged ? expectedOnlyChanged : expectedFull;
```
markusicu, line 766:
```suggestion
                const UnicodeString &expected = isOnlyChanged ? expectedOnlyChanged : expectedFull;
```

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -667,6 +670,120 @@ StringCaseTest::TestTitleOptions() {
...
+        {u"íj́abc\xD834\xDD6E",  u"ÍJ́abc\xD834\xDD6E", u"ÍJ"},
+        {u"íjabc\u1ABE",        u"Íjabc\u1ABE",       u"Í"},
+    };
+
+    for (const auto& cas : dutchTitleTestCases) {
+        const UnicodeString input = cas.input;
+        const UnicodeString expectedFull = cas.expectedFull;
+        const UnicodeString expectedOnlyChanged = cas.expectedOnlyChanged;
+
+        for (const auto& isTitlecaseWholeString : {true, false}) {
```
markusicu, line 760: See question in the Java test file about this option loop.

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -667,6 +670,120 @@ StringCaseTest::TestTitleOptions() {
...
+                    TEST_TITLE,
+                    nullptr,
+                    "nl",
+                    testOptions
+                );
+            }
+        }
+        
+    }
+
```
markusicu, line 783: delete empty line

markusicu (changes requested):

icu4c/source/common/ustrcase.cpp
```diff
@@ -477,32 +479,65 @@ ustrcase_internalToTitle(int32_t caseLocale, uint32_t options, BreakIterator *it
...
+                UnicodeString dbg(dest, destIndex, destCapacity);
+                bufferLength = dbg.extract(0, dbg.length(), buffer, sizeof(buffer));
+                std::cout << "dest (dbg) = " << buffer << std::endl;
+
+                std::cout << "csc.cpStart = " << csc.cpStart << std::endl;
+                std::cout << "csc.cpLimit = " << csc.cpLimit << std::endl;
+                std::cout << "c = " << c << std::endl;
+
                 /* Special case Dutch IJ titlecasing */
                 if (titleStart+1 < index &&
```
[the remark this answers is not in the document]
echeran: I needed to use `titleLimit` in the UTF-8 because when the first character is U+00CD (pre-composed capital I with acute), `titleStart + 1 != titleLimit`.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UCharacterCaseTest.java
echeran: The new test in intltest/strcase.cpp passes now, after converting the C++ UTF-16 code to the UTF-8 version, because the strcase.cpp tests call both versions.

But now it seems like these fixes have triggered a different failure in a UnicodeString test now where it is calling title case methods.

icu4c/source/common/ucasemap.cpp
```diff
@@ -420,6 +426,96 @@ void toUpper(int32_t caseLocale, uint32_t options,
...
+ * Input: c is a letter I with or without acute accent.
+ * start is the index in src after c, and is less than segmentLimit.
+ * If a plain i/I is followed by a plain j/J,
+ * or an i/I with acute (precomposed or decomposed) is followed by a j/J with acute,
+ * then we output accordingly.
+ *
+ * @return the src index after the titlecased sequence, or the start index if no Dutch IJ
+ * @throws IOException
+ */
+int32_t maybeTitleDutchIJ(const uint8_t *src, UChar32 c, int32_t start, int32_t segmentLimit,
```
markusicu, line 440: Please surround this function with the anonymous namespace.
Like the previous code, but it's probably easier to start a new anonymous-namespace block to avoid dealing with the partial overlap of that with the UCONFIG_NO_BREAK_ITERATION section.
Please end the anonymous namespace exactly like line 425, with that comment.

icu4c/source/common/ucasemap.cpp
```diff
@@ -420,6 +426,96 @@ void toUpper(int32_t caseLocale, uint32_t options,
...
+    U8_NEXT(src, index, segmentLimit, c2);
+
+    // Is the first letter an i/I with accent?
+    if (c == u'I') {
+        if (c2 == ACUTE) {
+            withAcute = true;
+            unchanged1 = 2;  // ACUTE is 2 code units in UTF-8
+            if (index == segmentLimit) { return start; }
+            U8_NEXT(src, index, segmentLimit, c2);
+        }
```
markusicu, line 462: Optional: Faster check since we know what we are looking for...
```suggestion
    c2 = src[index++];

    // Is the first letter an i/I with accent?
    if (c == u'I') {
        if (c2 == u8"\u0301"[0] && index < segmentLimit && src[index++] == u8"\u0301"[1]) {
            withAcute = true;
            unchanged1 = 2;  // ACUTE is 2 code units in UTF-8
            if (index == segmentLimit) { return start; }
            c2 = src[index++];
        }
```
If you take my suggestions, then I think you won't need
```
#define ACUTE u'\u0301'
```
in this file any more.
Instead, you could then use
```
#define ACUTE_STR u8"\u0301"
```
and use that instead of the hardcoded literals.

icu4c/source/common/ucasemap.cpp
```diff
@@ -420,6 +426,96 @@ void toUpper(int32_t caseLocale, uint32_t options,
...
+    if (withAcute) {
+        if (index == segmentLimit) {
+            return start;
+        } else {
+            UChar32 c3;
+            U8_NEXT(src, index, segmentLimit, c3);
+            if (c3 != ACUTE) { 
+                return start;
+            }
+        }
```
markusicu, line 487: Optional
```suggestion
        if ((index + 1) >= segmentLimit || src[index++] != u8"\u0301"[0] || src[index++] != u8"\u0301"[1]) {
            return start;
        }
```

icu4c/source/common/ustrcase.cpp
```diff
@@ -396,6 +402,90 @@ U_NAMESPACE_USE
...
+ * Input: c is a letter I with or without acute accent.
+ * start is the index in src after c, and is less than segmentLimit.
+ * If a plain i/I is followed by a plain j/J,
+ * or an i/I with acute (precomposed or decomposed) is followed by a j/J with acute,
+ * then we output accordingly.
+ *
+ * @return the src index after the titlecased sequence, or the start index if no Dutch IJ
+ * @throws IOException
+ */
+int32_t maybeTitleDutchIJ(const UChar *src, UChar32 c, int32_t start, int32_t segmentLimit,
```
markusicu, line 415: ditto surround with anonymous namespace

That keeps these functions internal. Otherwise, on many platforms the function symbol is exported from the shared library. We don't want that.

markusicu (commented):
Looks good!
One issue/two places.
Otherwise, I added suggestions for making it slightly simpler/faster.

icu4c/source/common/ucasemap.cpp
markusicu: = leftover from https://github.com/unicode-org/icu/pull/1869#discussion_r753461595

echeran: Done.

icu4c/source/common/ustrcase.cpp
echeran: Done.

icu4c/source/common/ucasemap.cpp
echeran: Done.

icu4c/source/common/ucasemap.cpp
echeran: Done.  Change applied manually, and file-level #define removed.

icu4c/source/common/ucasemap.cpp
```diff
@@ -420,6 +420,97 @@ void toUpper(int32_t caseLocale, uint32_t options,
...
+
+/**
+ * Input: c is a letter I with or without acute accent.
+ * start is the index in src after c, and is less than segmentLimit.
+ * If a plain i/I is followed by a plain j/J,
+ * or an i/I with acute (precomposed or decomposed) is followed by a j/J with acute,
+ * then we output accordingly.
+ *
+ * @return the src index after the titlecased sequence, or the start index if no Dutch IJ
+ * @throws IOException
```
markusicu, line 437:
```suggestion
```

icu4c/source/common/ustrcase.cpp
```diff
@@ -396,6 +402,94 @@ U_NAMESPACE_USE
...
+
+/**
+ * Input: c is a letter I with or without acute accent.
+ * start is the index in src after c, and is less than segmentLimit.
+ * If a plain i/I is followed by a plain j/J,
+ * or an i/I with acute (precomposed or decomposed) is followed by a j/J with acute,
+ * then we output accordingly.
+ *
+ * @return the src index after the titlecased sequence, or the start index if no Dutch IJ
+ * @throws IOException
```
markusicu, line 415:
```suggestion
```

markusicu (commented):
lgtm except we don't throw exceptions in C++

please fix, and then squash right away

markusicu (approved):
lgtm tnx!

PR #1872, opened 2021-09-21
ICU-21662 UVector cleanup in rbtz.cpp

aheninger:
Revise uses of UVector in rbtz.cpp to better handle memory allocation failures.
This is one of an ongoing series of commits to address similar problems with
UVector usage throughout ICU.

The changes include

- Use LocalPointers and UVector deleter functions to simplify OOM checking and recovery.
- Fix RuleBasedTimeZone::addTransitionRule(rule) to have standard ICU adopt behavior
  when errors occur, meaning automatic deletion of the incoming rule. This simplifies
  both the implementation of the function and the code at the call sites.
- Update addTransitionRule() call sites. Includes modifying the Dangi calendar initialization
  to not silently ignore errors.
- struct Transition is changed to derive from UMemory, which allows the use of LocalPointers.

##### Checklist

### DIFF
--- a/icu4c/source/i18n/dangical.cpp
+++ b/icu4c/source/i18n/dangical.cpp
@@ -52,7 +52,7 @@ U_NAMESPACE_BEGIN
 //-------------------------------------------------------------------------
 
 DangiCalendar::DangiCalendar(const Locale& aLocale, UErrorCode& success)
-:   ChineseCalendar(aLocale, DANGI_EPOCH_YEAR, getDangiCalZoneAstroCalc(), success)
+:   ChineseCalendar(aLocale, DANGI_EPOCH_YEAR, getDangiCalZoneAstroCalc(success), success)
 {
 }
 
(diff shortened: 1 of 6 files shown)


icu4c/source/i18n/dangical.cpp
```diff
@@ -103,32 +103,41 @@ const char *DangiCalendar::getType() const {
...
-    TimeZoneRule* rule1897 = new TimeArrayTimeZoneRule(UNICODE_STRING_SIMPLE("Korean 1897"), 7*kOneHour, 0, millis1897, 1, DateTimeRule::STANDARD_TIME);
-    TimeZoneRule* rule1898to1911 = new TimeArrayTimeZoneRule(UNICODE_STRING_SIMPLE("Korean 1898-1911"), 8*kOneHour, 0, millis1898, 1, DateTimeRule::STANDARD_TIME);
-    TimeZoneRule* ruleFrom1912 = new TimeArrayTimeZoneRule(UNICODE_STRING_SIMPLE("Korean 1912-"), 9*kOneHour, 0, millis1912, 1, DateTimeRule::STANDARD_TIME);
-    UErrorCode status = U_ZERO_ERROR;
-    RuleBasedTimeZone* dangiCalZoneAstroCalc = new RuleBasedTimeZone(UNICODE_STRING_SIMPLE("KOREA_ZONE"), initialTimeZone); // adopts initialTimeZone
-    dangiCalZoneAstroCalc->addTransitionRule(rule1897, status); // adopts rule1897
-    dangiCalZoneAstroCalc->addTransitionRule(rule1898to1911, status);
-    dangiCalZoneAstroCalc->addTransitionRule(ruleFrom1912, status);
+    LocalPointer<InitialTimeZoneRule> initialTimeZone(new InitialTimeZoneRule(
+        UnicodeString(u"GMT+8"), 8*kOneHour, 0), status);
```
markusicu, line 112: optional: you should be able to simplify `UnicodeString(u"...")` to just `u"..."` because of the implicit constructor-from-`char16_t*`

icu4c/source/i18n/unicode/rbtz.h
```diff
@@ -89,17 +89,18 @@ class U_I18N_API RuleBasedTimeZone : public BasicTimeZone {
     virtual bool operator!=(const TimeZone& that) const;
 
     /**
-     * Adds the <code>TimeZoneRule</code> which represents time transitions.
-     * The <code>TimeZoneRule</code> must have start times, that is, the result
-     * of isTransitionRule() must be true. Otherwise, U_ILLEGAL_ARGUMENT_ERROR
+     * Adds the `TimeZoneRule` which represents time transitions.
```
markusicu, line 92: optional: I prefer not Markdown-ifying API docs that should be about the same between C++ and Java, because it makes it harder to keep them in sync.

icu4c/source/i18n/rbtz.cpp
```diff
@@ -114,32 +121,35 @@ RuleBasedTimeZone::operator!=(const TimeZone& that) const {
 
 void
 RuleBasedTimeZone::addTransitionRule(TimeZoneRule* rule, UErrorCode& status) {
+    LocalPointer<TimeZoneRule>lpRule(rule);
```
markusicu, line 124: good -- but we should document in the coding guidelines an exception to the “bail out immediately for incoming error” rule, saying that if a function adopts objects, it has to do so even if there is an incoming error.

separate PR or ticket?

markusicu (approved):
lgtm but with optional suggestions and a question

icu4c/source/i18n/dangical.cpp
aheninger: That constructor is marked explicit in library code, so it doesn't work.

jefgen (approved):
fwiw, lgtm. Thanks for this clean-up!

icu4c/source/i18n/unicode/rbtz.h
aheninger: I'd like to leave it markdown. The Java and C++ docs are sufficiently different that there's copy/paste isn't going to work anyhow.

icu4c/source/i18n/rbtz.cpp
aheninger: I think a separate PR, and if we have some generic user guide update ticket, do it using that.

I can do it.

markusicu: > I think a separate PR, and if we have some generic user guide update ticket, do it using that.

Separate PR is fine. If you can do it quickly, please use this same ticket.

PR #1875, opened 2021-09-22
ICU-21714 Use FixedDecimal instead of double for plural samples in tests

echeran:
In tests that parse plural rule sample values from strings (strings = the substrings of the plural rule), this PR parses those strings to FixedDecimal instead of to double. This enables carrying the exponent, which is needed when seeing numbers like 1.2c3 and 1.2e3.

(This also required fixing up the behavior `getPluralOperand` for FixedDecimal to work the same as in FormattedNumber, so that `PluralRules.select(FixedDecimal)` behaves like `PluralRules.select(FormattedNumber)`.)

@pedberg-icu I think this should fix the problems, but let me know what you observe. In ICU4J, when I removed the logKnownIssue line, the test failures I got seem to be due to a difference in the plural rule strings between expected (in `LOCALE_SNAPSHOT`) and actual (coming from data / a factory?). I added the `testCheckValue()` test to help verify.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21714

### DIFF
--- a/icu4c/source/i18n/plurrule.cpp
+++ b/icu4c/source/i18n/plurrule.cpp
@@ -1881,8 +1881,8 @@ void FixedDecimal::adjustForMinFractionDigits(int32_t minFractionDigits) {
 
 double FixedDecimal::getPluralOperand(PluralOperand operand) const {
     switch(operand) {
-        case PLURAL_OPERAND_N: return source;
-        case PLURAL_OPERAND_I: return static_cast<double>(intValue);
+        case PLURAL_OPERAND_N: return (exponent == 0 ? source : source * pow(10, exponent));
+        case PLURAL_OPERAND_I: return (double) longValue();
         case PLURAL_OPERAND_F: return static_cast<double>(decimalDigits);
         case PLURAL_OPERAND_T: return static_cast<double>(decimalDigitsWithoutTrailingZeros);
         case PLURAL_OPERAND_V: return visibleDecimalDigitCount;
@@ -1929,6 +1929,18 @@ UnicodeString FixedDecimal::toString() const {
         snprintf(buffer, sizeof(buffer), pattern, source);
     }
     return UnicodeString(buffer, -1, US_INV);
+}
+
+double FixedDecimal::doubleValue() const {
+    return (isNegative ? -source : source) * pow(10, exponent);
+}
+
+int64_t FixedDecimal::longValue() const {
+    if (exponent == 0) {
+        return intValue;
+    } else {
+        return (long) (pow(10, exponent) * intValue);
+    }
 }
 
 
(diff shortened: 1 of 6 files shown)


icu4c/source/i18n/plurrule.cpp
```diff
@@ -1881,8 +1881,8 @@ void FixedDecimal::adjustForMinFractionDigits(int32_t minFractionDigits) {
 
 double FixedDecimal::getPluralOperand(PluralOperand operand) const {
     switch(operand) {
-        case PLURAL_OPERAND_N: return source;
-        case PLURAL_OPERAND_I: return static_cast<double>(intValue);
+        case PLURAL_OPERAND_N: return (exponent == 0 ? source : source * pow(10, exponent));
+        case PLURAL_OPERAND_I: return (double) longValue();
```
echeran, line 1885: @sffc is `static_cast<double>` or `(double)` better to use in C++ here?

sffc: I prefer `static_cast<double>` but Markus prefers `(double)`.  The C++ standard is to use `static_cast<double>`.

pedberg-icu: I do not have a preference on the casting, but I am glad to see the exponent support added

pedberg-icu (approved):
This looks good to me

pedberg-icu:
> @pedberg-icu I think this should fix the problems, but let me know what you observe. In ICU4J, when I removed the logKnownIssue line, the test failures I got seem to be due to a difference in the plural rule strings between expected (in LOCALE_SNAPSHOT) and actual (coming from data / a factory?). I added the testCheckValue() test to help verify.

Thanks! I will do at a least one more CLDR-ICU integration pass before release, and at that point I will try removing the logKnownIssue and making necessary adjustments.

icu4c/source/i18n/plurrule.cpp
```diff
@@ -1881,8 +1881,8 @@ void FixedDecimal::adjustForMinFractionDigits(int32_t minFractionDigits) {
 
 double FixedDecimal::getPluralOperand(PluralOperand operand) const {
     switch(operand) {
-        case PLURAL_OPERAND_N: return source;
-        case PLURAL_OPERAND_I: return static_cast<double>(intValue);
+        case PLURAL_OPERAND_N: return (exponent == 0 ? source : source * pow(10, exponent));
```
sffc, line 1884: (as stated in chat)
I think that the correct place to multiply by pow(10) is in the parsing function, not in the getPluralOperand function.  I would be much more confident in this PR if the multiplication were confined to the constructors.

sffc (changes requested):

icu4c/source/i18n/plurrule.cpp
sffc: I chatted with Elango and he will prioritize https://unicode-org.atlassian.net/browse/ICU-21322 for ICU 71, and in the mean time this PR is fine for unblocking ICU 70.

sffc (dismissed):

sffc (approved):

PR #1876, opened 2021-09-22
ICU-21544 Throw argument error when the units are not convertible.

younies:
closes: https://github.com/icu-units/icu/issues/82

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21544

### DIFF
--- a/icu4c/source/test/intltest/numbertest_api.cpp
+++ b/icu4c/source/test/intltest/numbertest_api.cpp
@@ -88,6 +88,7 @@ void NumberFormatterApiTest::runIndexedTest(int32_t index, UBool exec, const cha
         TESTCASE_AUTO(unitCurrency);
         TESTCASE_AUTO(unitInflections);
         TESTCASE_AUTO(unitGender);
+        TESTCASE_AUTO(unitNotConvertible);
         TESTCASE_AUTO(unitPercent);
         if (!quick) {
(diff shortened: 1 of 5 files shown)


sffc (approved):

icu4c/source/test/intltest/numbertest_api.cpp
```diff
@@ -2699,6 +2700,31 @@ void NumberFormatterApiTest::unitGender() {
     assertEquals("getGender for a genderless language", "", fn.getGender(status));
 }
 
+void NumberFormatterApiTest::unitNotConvertible() {
+    IcuTestErrorCode status(*this, "unitNotConvertible");
+    const double randomNumber = 1234;
+
+    NumberFormatter::with()
+        .unit(MeasureUnit::forIdentifier("meter-and-liter", status))
```
FrankYFTang, line 2708: I think this test is not good enough, from this test, it cannot tell which line trigger the error.
Is the error happen in MeasureUnit::forIdentifier() or the error happen in formatDouble()?
If the error is expected to happen in MeasureUnit::forIdentifier(), then the .locale("en_US")
        .formatDouble(randomNumber, status); part should be removed. If the error should happen only in formatDouble() then we should assert the status is not error after the calling of  MeasureUnit::forIdentifier("meter-and-liter", status)

PR #1873, opened 2021-09-21
ICU-21581 BRSRC 70.1 Version update and regenerate configure for v70.1

erik0686:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21581
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [ ] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [X] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [X] API docs and/or User Guide docs changed or added, if applicable

In this PR, I am updating the version number from 70.0.1 for the BRS task.
Previously, we had frontloaded part of this, so the diff in this PR are not as numerous. 

It has also been decided that we should differentiate frontloaded tasks and RC tasks by having the version numbers being:
Frontload version number: XX.X.X (70.0.1 -> Major.minor.patch) 
RC/GA version number:      XX.X    (70.1   -> Major.minor)

I've added some documentation for this, for future releases :)

PR #1877, opened 2021-09-22
ICU-21581 ICU Change Reports for 70rc

sven-oly:
##### Checklist

### DIFF
--- a/icu4c/APIChangeReport.md
+++ b/icu4c/APIChangeReport.md
@@ -5,275 +5,659 @@
  License & terms of use: http://www.unicode.org/copyright.html
 -->
 
-# ICU4C API Comparison: ICU 68 (update #1: 68.2) with ICU 69
+# ICU4C API Comparison: ICU 69 with ICU 70
 
 > _Note_ Markdown format of this document is new for ICU 65.
 
-- [Removed from ICU 68](#removed)
-- [Deprecated or Obsoleted in ICU 69](#deprecated)
-- [Changed in  ICU 69](#changed)
-- [Promoted to stable in ICU 69](#promoted)
-- [Added in ICU 69](#added)
-- [Other existing drafts in ICU 69](#other)
+- [Removed from ICU 69](#removed)
+- [Deprecated or Obsoleted in ICU 70](#deprecated)
+- [Changed in  ICU 70](#changed)
+- [Promoted to stable in ICU 70](#promoted)
+- [Added in ICU 70](#added)
+- [Other existing drafts in ICU 70](#other)
 - [Signature Simplifications](#simplifications)
 
 ## Removed
 
-Removed from ICU 68
+Removed from ICU 69
   
-| File | API | ICU 68 | ICU 69 |
+| File | API | ICU 69 | ICU 70 |
 |---|---|---|---|
-| listformatter.h | <tt>static</tt> ListFormatter* icu::ListFormatter::createInstance(const Locale&amp;, const char*, UErrorCode&amp;) |  Internal | (missing)
-| measunit.h | MeasureUnit icu::MeasureUnit::withSIPrefix(UMeasureSIPrefix, UErrorCode&amp;) const |  DraftICU 67 | (missing)
-| measunit.h | UMeasureSIPrefix icu::MeasureUnit::getSIPrefix(UErrorCode&amp;) const |  DraftICU 67 | (missing)
-| numberformatter.h | Usage&amp; icu::number::impl::Usage::operator=(Usage&amp;&amp;) |  Internal | (missing)
-| numberformatter.h | Usage&amp; icu::number::impl::Usage::operator=(const Usage&amp;) |  Internal | (missing)
-| numberformatter.h | bool icu::number::impl::Usage::isSet() const |  Internal | (missing)
-| numberformatter.h | icu::number::impl::Usage::Usage(Usage&amp;&amp;) |  Internal | (missing)
-| numberformatter.h | icu::number::impl::Usage::Usage(const Usage&amp;) |  Internal | (missing)
-| numberformatter.h | icu::number::impl::Usage::~Usage() |  Internal | (missing)
-| numberformatter.h | int16_t icu::number::impl::Usage::length() const |  Internal | (missing)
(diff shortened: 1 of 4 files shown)


icu4j/APIChangeReport.html
```diff
@@ -4,260 +4,231 @@
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
 <!-- © 2021 and later: Unicode, Inc. and others. -->
 <!-- License & terms of use: http://www.unicode.org/copyright.html -->
-<title>ICU4J API Comparison: ICU4J 68.1 with ICU4J 69.1</title>
+<title>ICU4J API Comparison: ICU4J 69.1 with ICU4J 70.0.1</title>
```
yumaoka, line 7: I think we should set the version to 70.1 at this point, so we don't need to update it for 70.1 GA.

yumaoka (changes requested):
Can you update icu4j/build.properties to - api.doc.version = 70.1, maven.pom.ver = 70.1, icu4j/main/shared/build/common.properties - icu4j.impl.version = 70.1, then regenerate?

icu4j/APIChangeReport.html
yumaoka: The version issue - you should wat this -> https://github.com/unicode-org/icu/pull/1873

sven-oly: Thx. I've just added the ICU4C API Change report to this PR, squashing into a single commit.

I think PR 1873 will update the version to 70.1 - I can redo these API reports after that.

markusicu: Erik merged PR 1873 yesterday. Please update.

markusicu:
The PR title and the commit message no longer reflect the changes (no longer just ICU4J).
Please change both to “ICU-21581 API change reports for ICU 70rc”.

Changing the commit title is easiest when you update the reports anyway, using `git commit -a --amend` (without `-m`) which lets you change the commit message.

markusicu:
FYI: The PR description had problems; I just fixed them:
- The “Required: Issue filed:” was missing the ICU ticket number
- All of the checkboxes were ticked although not all of those things are done (nor need to be done). I unchecked the ones that aren't done.

icu4c/APIChangeReport.md
```diff
@@ -5,265 +5,1477 @@
...
-| numberformatter.h | icu::number::impl::Usage::Usage(Usage&amp;&amp;) |  Internal | (missing)
-| numberformatter.h | icu::number::impl::Usage::Usage(const Usage&amp;) |  Internal | (missing)
-| numberformatter.h | icu::number::impl::Usage::~Usage() |  Internal | (missing)
-| numberformatter.h | int16_t icu::number::impl::Usage::length() const |  Internal | (missing)
-| numberformatter.h | void icu::number::impl::Usage::set(StringPiece) |  Internal | (missing)
-| numberrangeformatter.h | UnicodeString icu::number::FormattedNumberRange::getFirstDecimal(UErrorCode&amp;) const |  DeprecatedICU 68 | (missing)
-| numberrangeformatter.h | UnicodeString icu::number::FormattedNumberRange::getSecondDecimal(UErrorCode&amp;) const |  DeprecatedICU 68 | (missing)
-| rbtz.h | void icu::RuleBasedTimeZone::getOffsetFromLocal(UDate, int32_t, int32_t, int32_t&amp;, int32_t&amp;, UErrorCode&amp;) const |  Internal | (missing)
-| simpletz.h | void icu::SimpleTimeZone::getOffsetFromLocal(UDate, int32_t, int32_t, int32_t&amp;, int32_t&amp;, UErrorCode&amp;) const |  Internal | (missing)
+| appendable.h | UBool icu::UnicodeStringAppendable::appendCodePoint(UChar32) |  StableICU 4.8 | (missing)
```
markusicu, line 26: The C++ API change report shows a misleadingly large number of changes from the addition of the `override` keyword that doesn't actually change the function signature.

@srl295 could you please point us to the code that simplifies function signatures for the report, so that irrelevant changes in annotations are not shown as a “change”? We should ignore both `virtual` and `override`.

srl295: @markusicu https://github.com/unicode-org/icu/blob/24b3cf182e848e7ea23d7107a2289bbafff1fb8b/tools/release/java/src/main/java/com/ibm/icu/dev/tools/docs/StableAPI.java#L543-L556

simplify here is in the sense of "create a canonical form for comparison".  You might want to make sure that override is at or near the end of the line, for example, so that some `u_overrideSomething()` does not get inadvertently transformed.

srl295: is this the successor to U_OVERRIDE anyway? then maybe  ` "[ ]*(override|U_OVERRIDE)", ""`

markusicu: Thanks Steven!
Yes, we have recently just used `override` directly. `U_OVERRIDE` is only necessary for compilers that do not fully support C++11. We should be able to retire `U_OVERRIDE` unless the EBCDIC platforms come back with good-enough C++ support but still missing `override` -- and if we agree that that is “good enough”.

@sven-oly please change the tool as Steven provided here for your next C++ change report generation.

sven-oly: Updated this line in the file, but the change report is still very large.

Line 511  "[ ]*(override|U_OVERRIDE)", "", // remove U_OVERRIDE

I also tried added a new line with just "override" but that also had no effect.

sven-oly: Any other ideas to fix all these extra removes and adds?

markusicu (commented):
Hi @sven-oly, thanks for updating after the version change, and for fixing the override problem.

However, we have another problem like this: @roubert changed many operator== and operator!= from returning UBool to returning bool. Now these are also cluttering up the C++ API change report, which makes it hard to see real API changes.

@srl295 et al., how about we also ignore UBool->bool changes? In StableAPI.java I think that should work with
```
    "\\bUBool\\b", "bool",
```

markusicu (commented):
otherwise lgtm AFAICT...

markusicu (approved):
lgtm tnx!

markusicu (approved):
still lgtm

PR #1879, opened 2021-09-24
ICU-21581 Adds GHA CI automation of BRS task 'Test ICU4J with only

gnrunge:
little-endian ICU4C data', cf.:
https://unicode-org.github.io/icu/processes/release/tasks/integration.html#test-icu4j-with-only-little-endian-icu4c-data

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21581

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -399,9 +399,34 @@ jobs:
       - uses: actions/checkout@v2
       - run: python3 tools/scripts/uconfig_vars_test.py -u
 
-# Run header tests with UCONFIG_NO_XXX variations.
+  # Run header tests with UCONFIG_NO_XXX variations.
   uconfig_variation-check-all-header-tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - run: python3 tools/scripts/uconfig_vars_test.py -p
+
+  # Test ICU4J with little-endian ICU4C data only
+  icu4j-little-endian-data-test:
+    needs: clang-release-build-and-test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+        with:
+          lfs: true
+      - name: Checkout lfs objects
+        run: git lfs pull
+      - uses: actions/setup-java@v1
+        with:
+          java-version: '11'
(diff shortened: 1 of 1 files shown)


.github/workflows/icu_ci.yml
```diff
@@ -399,9 +399,34 @@ jobs:
...
 
-# Run header tests with UCONFIG_NO_XXX variations.
+  # Run header tests with UCONFIG_NO_XXX variations.
   uconfig_variation-check-all-header-tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - run: python3 tools/scripts/uconfig_vars_test.py -p
+
+  # Test ICU4J with little-endian ICU4C data only
```
markusicu, line 409: lgtm

except: I suggest that we don't run it on every pending-PR commit, but only on merging into the main or maint/* branches. Just to reduce resource usage a bit for a test that rarely fails.

Can we just duplicate the "on:" lines 8-16 but omit the "pull_request:" part?

gnrunge: The "on" specification is for all jobs of the workflow. New workflow?
Do you have an observation (bottleneck or such) of yours in mind regarding resource usage? The overall elapse time of the GHA CI workflow doesn't seem longer than usual.

markusicu: I assumed that it was easy to make some actions happen on every commit, and some only after merging.
Maybe just in Azure? @jefgen 
If it's hard, then don't worry about it.

markusicu (approved):
lgtm tnx

.github/workflows/icu_ci.yml
jefgen: For the Azure pipelines exhaustive tests, we have another [file](https://github.com/unicode-org/icu/blob/main/.ci-builds/.azure-exhaustive-tests.yml) setup to run the tests. This file is set to only be triggered on pushes/new commits on the `main`/`maint/maint-*` branches, but not on any PRs. (So it runs *after* a PR gets merged).

I think we could do the same thing for the GitHub Actions, but we'd need to first create another file to define another (new) workflow, which would be setup to trigger in the same way.

gnrunge:
Replaced by PR#1884.

PR #1898, opened 2021-10-04
ICU-21781 suppress Wreturn-local-addr in formattedvalue

danbev:
This commit attempts to suppress the following warning when using GCC
11.2.1:
```console
In function ‘const UChar* icu_70::ufmtval_getString_70(
    const UFormattedValue*, int32_t*, UErrorCode*)’:
cc1plus: warning:
function may return address of local variable [-Wreturn-local-addr]
formattedvalue.cpp:207:19: note: declared here
207 |     UnicodeString readOnlyAlias = impl->fFormattedValue->
    |                   ^~~~~~~~~~~~~   toTempString(*ec);
```
This does not actually allow the warning to be ignored and I believe
the following issue has been opened in GCC for this:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93644
But adding this ignore would allow for this warning to be ignored when a
fix for this is included in GCC.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21781

### DIFF
--- a/icu4c/source/i18n/formattedvalue.cpp
+++ b/icu4c/source/i18n/formattedvalue.cpp
@@ -193,6 +193,9 @@ ucfpos_close(UConstrainedFieldPosition* ptr) {
 }
 
 
+#if defined(__clang__) || U_GCC_MAJOR_MINOR >= 1100
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wreturn-local-addr"
 U_CAPI const UChar* U_EXPORT2
 ufmtval_getString(
         const UFormattedValue* ufmtval,
@@ -211,6 +214,8 @@ ufmtval_getString(
     }
     return readOnlyAlias.getBuffer();
 }
+#pragma GCC diagnostic pop
+#endif
 
 
 U_CAPI UBool U_EXPORT2


icu4c/source/i18n/formattedvalue.cpp
```diff
@@ -193,6 +193,9 @@ ucfpos_close(UConstrainedFieldPosition* ptr) {
 }
 
 
+#if defined(__clang__) || U_GCC_MAJOR_MINOR >= 1100
```
FrankYFTang, line 196: I don't think we should suppress the warning. Warning is our friend who watch our back when we accidentially did the wrong thing. I think a simple fix is just
```
iff --git a/icu4c/source/i18n/formattedvalue.cpp b/icu4c/source/i18n/formattedvalue.cpp
index 1030661f22..a904b78a54 100644
--- a/icu4c/source/i18n/formattedvalue.cpp
+++ b/icu4c/source/i18n/formattedvalue.cpp
@@ -211,7 +211,7 @@ ufmtval_getString(
     }
     // Note: this line triggers -Wreturn-local-addr, but it is safe because toTempString is
     // defined to return memory owned by the ufmtval argument.
-    return readOnlyAlias.getBuffer();
+    return impl->fFormattedValue->toTempString(*ec).getBuffer();
 }
 
```

FrankYFTang:
ping

FrankYFTang:
ping, if the user do not response in 2 weeks, we will close this PR

PR #1910, opened 2021-10-21
ICU-21580 change site.icu-project.org to icu.unicode.org etc

markusicu:
Adjust most links from site.icu-project.org to icu.unicode.org.
Also adjust various other *.icu-project.org links as appropriate, for example ancient User Guide URLs.

I adjusted some but not all `#` fragment parts.

There are still some *.icu-project.org URLs for which I did not have time to find replacements, some in XML namespace URLs, etc.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21580


jefgen:
It looks like default compiler for MSYS2 was updated to use GCC 11 instead of GCC 10.
(from `mingw-w64-x86_64-gcc-10.3.0-5` to `mingw-w64-x86_64-gcc-11.2.0`)

Which is why this PR is now failing despite not touching any of the code. :/

It looks like the other PR also ran into the same issue as well, and made a fix as part of the change (though it's unrelated).
However, I think we'll want/need to fix this in the `maint` branch though, so I think we should file a separate ticket and make a PR to fix it.

jefgen:
FWIW, for this PR I think we can just bypass/override the failing MSYS2 check.

markusicu:
> FWIW, for this PR I think we can just bypass/override the failing MSYS2 check.

I just rebased on top of your PR #1911.

PR #1925, opened 2021-11-03
ICU-21793 Fix ucptrietest golden diff

kapouer:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21793

### DIFF
--- a/icu4c/source/tools/toolutil/toolutil.cpp
+++ b/icu4c/source/tools/toolutil/toolutil.cpp
@@ -228,18 +228,19 @@ uprv_compareGoldenFiles(
     std::ifstream ifs(goldenFilePath, std::ifstream::in);
     int32_t pos = 0;
     char c;
-    while ((c = ifs.get()) != std::char_traits<char>::eof() && pos < bufferLen) {
+    while (ifs.get(c) && pos < bufferLen) {
         if (c != buffer[pos]) {
             // Files differ at this position
-            return pos;
+            break;
         }
         pos++;
     }
-    if (pos < bufferLen || c != std::char_traits<char>::eof()) {
-        // Files are different lengths
-        return pos;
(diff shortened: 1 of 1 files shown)


icu4c/source/tools/toolutil/toolutil.cpp
```diff
@@ -228,14 +228,14 @@ uprv_compareGoldenFiles(
     std::ifstream ifs(goldenFilePath, std::ifstream::in);
     int32_t pos = 0;
     char c;
-    while ((c = ifs.get()) != std::char_traits<char>::eof() && pos < bufferLen) {
```
sffc, line 231: The [documentation](https://www.cplusplus.com/reference/istream/istream/get/) for `ifs.get()` says:

> returns the character read, or the end-of-file value (EOF) if no characters are available in the stream

In most cases, the end-of-file value is probably the NUL byte, which is what your code suggests, but my understanding is that the code as written before is *more* portable; assuming the EOF is NUL would be *less* portable.

markusicu: I think the problem in the existing code is that get() returns an int, but the code here immediately throws away most of the bits and casts it to a char. The EOF should be an out-of-band value, and by casting to char it becomes indistinguishable from a real contents byte.

In other words, the code should do this:
```
    int c;
    while ((c = ifs.get()) != std::char_traits<char>::eof() && pos < bufferLen) {
        if (((char)c) != buffer[pos]) {
```

I suspect that this is platform-sensitive in that the signed-ness of char and maybe the value of EOF differ.

See https://www.cplusplus.com/reference/istream/istream/get/
> `int get();`
> Return Value: The first signature returns the character read, or the end-of-file value (EOF) if no characters are available in the stream

markusicu (approved):
lgtm tnx

PR #1928, opened 2021-11-04
ICU-21823 Fixing charset detection to return error incase of no match…

shaobero:
… of charsets with input data

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21823

### DIFF
--- a/icu4c/source/test/intltest/csdetest.cpp
+++ b/icu4c/source/test/intltest/csdetest.cpp
@@ -109,6 +109,10 @@ void CharsetDetectionTest::runIndexedTest( int32_t index, UBool exec, const char
             if (exec) Ticket6954Test();
             break;
 
+       case 10: name = "Ticket21823Test";
+            if (exec) Ticket21823Test();
+            break;
+
         default: name = "";
             break; //needed to end loop
     }
@@ -838,4 +842,23 @@ void CharsetDetectionTest::Ticket6954Test() {
     TEST_ASSERT_SUCCESS(status);
     TEST_ASSERT(strcmp(name1, "windows-1252")==0);
 #endif
+}
+
+
+// Ticket 21823 - Issue with Charset Detector for ill-formed input strings. 
+//                Its fix involves returning a failure based error code 
+//                (U_INVALID_CHAR_FOUND) incase no charsets appear to match the input data.
+void CharsetDetectionTest::Ticket21823Test() {
+    UErrorCode status = U_ZERO_ERROR;
+    std::string str = "\x80";
+    UCharsetDetector* csd = ucsdet_open(&status);
+
+    ucsdet_setText(csd, str.data(), str.length(), &status);
+    const UCharsetMatch* match = ucsdet_detect(csd, &status);
+
+    if (match == NULL) {
+        TEST_ASSERT(U_FAILURE(status));
+    }
+
+    ucsdet_close(csd);
 }
--- a/icu4c/source/i18n/csdetect.cpp
+++ b/icu4c/source/i18n/csdetect.cpp
@@ -270,6 +270,11 @@ const CharsetMatch * const *CharsetDetector::detectAll(int32_t &maxMatchesFound,
 
     maxMatchesFound = resultCount;
 
+    if (maxMatchesFound == 0) {
+        status = U_INVALID_CHAR_FOUND;
+        return NULL;
+    }
+
     return resultArray;
 }
 
--- a/icu4c/source/test/cintltst/ucsdetst.c
+++ b/icu4c/source/test/cintltst/ucsdetst.c
@@ -402,6 +402,7 @@ static void TestBufferOverflow(void) {
     }
 
     for (idx = 0; idx < UPRV_LENGTHOF(testStrings); idx++) {
+        status = U_ZERO_ERROR;
         ucsdet_setText(csd, testStrings[idx], -1, &status);
         match = ucsdet_detect(csd, &status);
 
--- a/icu4c/source/test/intltest/csdetest.h
+++ b/icu4c/source/test/intltest/csdetest.h
@@ -33,6 +33,7 @@ public:
     virtual void IBM420Test();
     virtual void Ticket6394Test();
     virtual void Ticket6954Test();
+    virtual void Ticket21823Test();
 
 private:
     void checkEncoding(const UnicodeString &testString,


markusicu:
Sorry for the late reply...
Could you please amend your commit, adding a small piece of code to the unit tests to exercise this case?
That will also re-trigger the Continuous Integration testing, and hopefully the tests will pass this time.

Also, I submitted a follow-up ticket based on Andy's ticket comment: ICU-21934
In the meantime, I agree that your change is an improvement.

shaobero:
@markusicu It probably seems that log files corresp. to the above failed checks have been deleted (I see **Error: The log was not found. It may have been deleted based on retention settings.** upon viewing the failed checks). Could you please let me know in which files do I need to make changes for the unit tests?

markusicu:
> @markusicu It probably seems that log files corresp. to the above failed checks have been deleted

That's ok. I suspect that those failures were spurious.

> Could you please let me know in which files do I need to make changes for the unit tests?

Please add a small new test function here:
- https://github.com/unicode-org/icu/blob/main/icu4c/source/test/intltest/csdetest.h
- https://github.com/unicode-org/icu/blob/main/icu4c/source/test/intltest/csdetest.cpp

shaobero:
@markusicu I have added a new test function for testing this fix in the files mentioned above by you. Could you please review the same?

markusicu:
> @markusicu I have added a new test function for testing this fix in the files mentioned above by you. I have checked at my end that this test case fails without the fix and succeeds with this fix. Accordingly, could you please review the same?

Yes, thanks. The changes look ok. However, another test case fails, and that's probably what failed before.
```
  ucsdetst  {
   TestConstruction                          ---[OK]  
   TestUTF8                                  ---[OK]  
   TestUTF16                                 ---[OK]  
   TestC1Bytes                               ---[OK]  
   TestInputFilter                           ---[OK]  
   TestErrorChaining                         ---[OK]  
   TestBufferOverflow   {
!!   Unexpectedly got no results at index 7.
   } ---[1 ERRORS in /ucsdetst/TestBufferOverflow] 
```
From the CI logs, it's this function that fails:
https://github.com/unicode-org/icu/blob/main/icu4c/source/test/cintltst/ucsdetst.c#L371

Index 7 is for the last test case. I suspect that with your code change the detector sets an error code as early as the index=5 test case, and that it is ignored by the test code.

I would try to insert
```
        status = U_ZERO_ERROR;
```
just before the call to `ucsdet_setText()` on line 405.

----
**Actually,** I wonder now whether we shouldn't keep the behavior as is.
The [API docs](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucsdet_8h.html#af4880eb6b22328e118f698fd2ceaad98) do say "Returns: a UCharsetMatch representing the best matching charset, *or NULL if no charset matches the byte data*."

@aheninger @pedberg-icu should we keep the current behavior of returning a NULL UCharsetMatch with no error code?

icu4c/source/test/intltest/csdetest.cpp
```diff
@@ -839,3 +843,21 @@ void CharsetDetectionTest::Ticket6954Test() {
...
+void CharsetDetectionTest::Ticket21823Test() {
+    UErrorCode status = U_ZERO_ERROR;
+    std::string str = "\x80";
+    UCharsetDetector* csd = ucsdet_open(&status);
+
+    ucsdet_setText(csd, str.data(), str.length(), &status);
+    const UCharsetMatch* match = ucsdet_detect(csd, &status);
+
+    if (match == NULL)
+        TEST_ASSERT(U_FAILURE(status));
```
markusicu, line 860:
```suggestion
    if (match == NULL) {
        TEST_ASSERT(U_FAILURE(status));
    }
```

pedberg-icu:
> **Actually,** I wonder now whether we shouldn't keep the behavior as is. The [API docs](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucsdet_8h.html#af4880eb6b22328e118f698fd2ceaad98) do say "Returns: a UCharsetMatch representing the best matching charset, _or NULL if no charset matches the byte data_."
> 
> @aheninger @pedberg-icu should we keep the current behavior of returning a NULL UCharsetMatch with no error code?

The docs also say:
_The function will fail if
- no charset appears to match the data.
- no input text has been provided_

which suggests that we should set an error if no match.

icu4c/source/test/intltest/csdetest.cpp
aheninger: > should we keep the current behavior of returning a NULL UCharsetMatch with no error code?

We could, although it's not great for ease-of-use. User code would need to check both the error code and nullptr.

I think changing the implementation so that the ISO 8859-1 / Windows 1252 detector matches anything, with a low probability, would also be reasonable. Anything is valid. For short sequences that don't appear to match anything else, it's as good a guess as any. The place to make this change is probably in `CharsetRecog_8859_1::match()`, in csrsbcs.cpp:892

If we stick with setting an error code, would U_ILLEGAL_ARGUMENT_ERROR be better than U_INVALID_CHAR_FOUND? Neither is a perfect fit, but there is not an invalid character found. The input provided is to short or weird for meaningful detection. Returning to the original question, not being able to detect isn't really an error condition at all - a detector shouldn't choke on anything you throw at it.

markusicu:
Peter:
> The docs also say: “_The function will fail if no charset appears to match the data._ ...”

Andy:
> should we keep the current behavior ... We could, although it's not great for ease-of-use.

Yes. I agree that these arguments (and a user stumbling over the current behavior) justify this change.

Although it seems like Andy's final sentence contradicts that:
> Returning to the original question, not being able to detect isn't really an error condition at all - a detector shouldn't choke on anything you throw at it.

:shrug:

Andy:
> If we stick with setting an error code, would U_ILLEGAL_ARGUMENT_ERROR be better than U_INVALID_CHAR_FOUND? Neither is a perfect fit, but there is not an invalid character found. The input provided is to short or weird for meaningful detection.

I agree that U_INVALID_CHAR_FOUND is not a perfect fit, but a string that happens to be short is even less of an "illegal argument". And I don't think it's worth creating a whole new error code here.

Shashank: Please
- Fix the failing unit test; hopefully my suggested one-line addition will get you there.
- Apply my code style suggestion.
- Squash the commits into one.

shaobero:
Thanks @markusicu for your feedback. I have now made the suggested changes and have accordingly squashed the commits. I have checked at my end that now **cintltst** tests are successful. Could you please review these changes?

markusicu (approved):
lgtm tnx

PR #1934, opened 2021-11-13
ICU-21763 UVector cleanup, continued.

aheninger:
Revise uses of UVector in the next batch of files to better handle memory
allocation failures.  This is one of an ongoing series of commits to address
similar problems with UVector usage throughout ICU.

The changes primarily involve switching uses of UVector::addElementX() to the
new adoptElement() or addElement() functions, as appropriate, and using
LocalPointers for tracking memory ownership.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21763

### DIFF
--- a/icu4c/source/common/servnotf.cpp
+++ b/icu4c/source/common/servnotf.cpp
@@ -49,7 +49,11 @@ ICUNotifier::addListener(const EventListener* l, UErrorCode& status)
         if (acceptsListener(*l)) {
             Mutex lmx(&notifyLock);
             if (listeners == NULL) {
-                listeners = new UVector(5, status);
+                LocalPointer<UVector> lpListeners(new UVector(5, status), status);
+                if (U_FAILURE(status)) {
+                    return;
+                }
+                listeners = lpListeners.orphan();
             } else {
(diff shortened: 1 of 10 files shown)


icu4c/source/common/servnotf.cpp
```diff
@@ -59,7 +59,7 @@ ICUNotifier::addListener(const EventListener* l, UErrorCode& status)
                 }
```
jefgen, line 59: Ah, GitHub won't let comment on the line above... but I wonder if it would be worth adding a `LocalPointer<UVector>` on line 52 for the member `listeners`, in case OOM occurs.

jefgen (dismissed):
FWIW, LGTM (with a small optional comment/suggestion).
Thank you for all this clean up!

icu4c/source/common/servnotf.cpp
aheninger: Fixed. Sorry I missed that one.

jefgen (approved):
LGTM.
Thank you for all this clean up!

PR #1936, opened 2021-11-16
ICU-21699 Phrase based breaking(C++)

allensu05:
Implementation doc:
https://docs.google.com/document/d/1Lg50cEKCLowTL3JwCXlOXOYpDkYJAeKy0I8E7qk7mNs/edit#heading=h.6wpzwdae2hs1

Unit test step:

cd source
./runConfigureICU --enable-debug --disable-release Linux/gcc -disable-layoutex
make clean
make -j 30 tests
cd test/intltest
LD_LIBRARY_PATH=lib:stubdata:tools/ctestfw:../../lib:../../stubdata:../../tools/ctestfw:$LD_LIBRARY_PATH ./intltest rbbi/RBBITest/TestExtended

##### Checklist


icu4c/source/common/unicode/ubrk.h
```diff
@@ -132,6 +132,12 @@ typedef enum UBreakIteratorType {
  */
 #define UBRK_DONE ((int32_t) -1)
 
+/** Value indicating the text boundaries is open punctuation.
```
FrankYFTang, line 135: any changes into {common,i18n}/unicode/*.h are public header and need API review to add them. and this is surely not true "stable ICU 2.0" Please move such into your .cpp file instead of put that into public header?

icu4c/source/common/unicode/uchar.h
```diff
@@ -3110,6 +3118,12 @@ u_isxdigit(UChar32 c);
 U_CAPI UBool U_EXPORT2
 u_ispunct(UChar32 c);
 
+U_CAPI UBool U_EXPORT2
```
FrankYFTang, line 3121: u_isOpenPunct  / u_isClosePunct  are only used inside common/rbbi.cpp. I do not believe we need to add a new API for that unless you like to bring an API proposal to ICU TC for that. The other way is you just move these code into common/rbbi.cpp as a internal function in that file . Same to U_GC_OP_MASK / U_GC_CP_MASK

icu4c/source/common/unicode/rbbi.h
```diff
@@ -235,7 +240,7 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
      * @see #getBinaryRules
      * @stable ICU 2.8
      */
-    RuleBasedBreakIterator(UDataMemory* image, UErrorCode &status);
+    RuleBasedBreakIterator(UDataMemory* image, UErrorCode &status, const char* type);
```
FrankYFTang, line 243: This is not an acceptable change. Adding a new parameter will break pre-existing client. You could propose to add an new constructor which take one more arg but we need to keep the current one around to maintain backward compatability.

icu4c/source/common/unicode/rbbi.h
```diff
@@ -695,7 +700,7 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
      * @param c         A character in the dictionary set
      * @internal (private)
      */
-    const LanguageBreakEngine *getLanguageBreakEngine(UChar32 c);
```
FrankYFTang, line 698: This should not be changed, the return LanguageBreakEngine pointer have to be const. The returned LanguageBreakEngine should be immutable.

icu4c/source/common/rbbi.cpp
```diff
@@ -1187,7 +1196,7 @@ static void U_CALLCONV initLanguageFactories() {
 }
 
 
-static const LanguageBreakEngine*
```
FrankYFTang, line 1190: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created.

icu4c/source/common/rbbi.cpp
```diff
@@ -1196,7 +1205,7 @@ getLanguageBreakEngineFromFactory(UChar32 c)
     }
 
     int32_t i = gLanguageBreakFactories->size();
-    const LanguageBreakEngine *lbe = NULL;
+    LanguageBreakEngine *lbe = NULL;
```
FrankYFTang, line 1208: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created, you need to keep it as a const pointer.

icu4c/source/common/rbbi.cpp
```diff
@@ -1214,9 +1223,9 @@ getLanguageBreakEngineFromFactory(UChar32 c)
 //                          the character c.
 //
 //-------------------------------------------------------------------------------
-const LanguageBreakEngine *
```
FrankYFTang, line 1217: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created, you need to keep it as a const pointer.

icu4c/source/common/rbbi.cpp
```diff
@@ -1214,9 +1223,9 @@ getLanguageBreakEngineFromFactory(UChar32 c)
...
+LanguageBreakEngine *
 RuleBasedBreakIterator::getLanguageBreakEngine(UChar32 c) {
-    const LanguageBreakEngine *lbe = NULL;
```
FrankYFTang, line 1219: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created, you need to keep it as a const pointer.

icu4c/source/common/rbbi.cpp
```diff
@@ -1230,7 +1239,7 @@ RuleBasedBreakIterator::getLanguageBreakEngine(UChar32 c) {
 
     int32_t i = fLanguageBreakEngines->size();
     while (--i >= 0) {
-        lbe = (const LanguageBreakEngine *)(fLanguageBreakEngines->elementAt(i));
```
FrankYFTang, line 1233: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created, you need to keep it as a const pointer.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -158,11 +159,12 @@ void RuleBasedBreakIterator::DictionaryCache::populateDictionary(int32_t startPo
 
         // We now have a dictionary character. Get the appropriate language object
         // to deal with it.
-        const LanguageBreakEngine *lbe = fBI->getLanguageBreakEngine(c);
```
FrankYFTang, line 161: this is surely wrong, LanguageBreakEngine should be immutable and not changed once created, you need to keep it as a const pointer.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -158,11 +159,12 @@ void RuleBasedBreakIterator::DictionaryCache::populateDictionary(int32_t startPo
...
+        LanguageBreakEngine *lbe = fBI->getLanguageBreakEngine(c);
 
         // Ask the language object if there are any breaks. It will add them to the cache and
         // leave the text pointer on the other side of its range, ready to search for the next one.
         if (lbe != NULL) {
+            lbe->enablePhraseBreaking(fBI->fIsPhraseBreaking);
```
FrankYFTang, line 167: you should NOT change the LanguageBreakEngine after you create it. The LanguageBreakEngine is shared and used by multiple threads and multiple RBBIs currently, if you change it's internal state, the usage of it from a different thread will then be changes. you could choose to either
1. change the  getLanguageBreakEngine() to pass in fBI->fIsPhraseBreaking so it will return a different one OR
2. change the LanguageBreakEngine interface to somehow pass in the fBI->fIsPhraseBreaking as information in the breaking code (but not remembered in the internal state of LanguageBreakEngine itself)

icu4c/source/common/lstmbe.h
```diff
@@ -68,7 +68,7 @@ class LSTMBreakEngine : public DictionaryBreakEngine {
     const Vectorizer* fVectorizer;
 };
 
-U_CAPI const LanguageBreakEngine* U_EXPORT2 CreateLSTMBreakEngine(
```
FrankYFTang, line 71: You cannot remove the const here. The LanguageBreakEngine have to be immutable and the internal state cannot be changed after construction.

icu4c/source/common/brkeng.h
```diff
@@ -77,6 +77,11 @@ class LanguageBreakEngine : public UMemory {
                               UVector32 &foundBreaks,
                               UErrorCode &status) const = 0;
 
+  void enablePhraseBreaking(UBool enable);
```
FrankYFTang, line 80: It is ok to change the interface of LanguageBreakEngine since it is not considered as API. However, the design of LanguageBreakEngine is it need to be immutable and the internal state not changed after the construction because they will be shared by multiple thread. Adding a mutable method like enablePhraseBreaking() to change internal state is not acceptable.

FrankYFTang (changes requested):

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +422,128 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+        {0xFF10, 0xFF19}     //full width number 0~9
+    };
+    for(int32_t i = 0 ; i < 2 ; i++) {
+        if(ch >= range[i][0] && ch <= range[i][1])
+            return true;
+    }
+    return false;
+}
+
+static inline bool isCjk(UChar32 ch) {
```
macchiati, line 437: If possible, this would be better expressed as a UnicodeSet, using Unicode properties where possible.

icu4c/source/common/unicode/ubrk.h
allensu05: Fixed

icu4c/source/common/unicode/uchar.h
allensu05: Fixed

icu4c/source/common/unicode/rbbi.h
allensu05: Fix by adding a private constructor.

icu4c/source/common/unicode/rbbi.h
allensu05: Fixed

icu4c/source/common/rbbi.cpp
allensu05: Fixed

icu4c/source/common/rbbi.cpp
allensu05: Fixed

icu4c/source/common/rbbi.cpp
allensu05: Fixed

icu4c/source/common/rbbi.cpp
allensu05: Fixed

icu4c/source/common/rbbi_cache.cpp
allensu05: Fixed

icu4c/source/common/lstmbe.h
allensu05: Fixed

icu4c/source/common/brkeng.h
allensu05: Remove this function

icu4c/source/common/brkeng.cpp
```diff
@@ -79,6 +79,7 @@ UnhandledEngine::findBreaks( UText *text,
                              int32_t /* startPos */,
                              int32_t endPos,
                              UVector32 &/*foundBreaks*/,
+                             UBool isPhraseBreaking,
```
FrankYFTang, line 82: Change to 
                             UBool /* isPhraseBreaking */,
You only need to UBool in the function parameter but not the name isPhraseBreaking there. Same solution as for the startPos and foundBreaks 1 and 3 lines above.

icu4c/source/common/dictbe.cpp
```diff
@@ -230,6 +231,7 @@ ThaiBreakEngine::divideUpDictionaryRange( UText *text,
                                                 int32_t rangeStart,
                                                 int32_t rangeEnd,
                                                 UVector32 &foundBreaks,
+                                                UBool isPhraseBreaking,
```
FrankYFTang, line 234: If you got compiler error, change to 
                             UBool /* isPhraseBreaking */,

icu4c/source/common/dictbe.cpp
```diff
@@ -469,6 +471,7 @@ LaoBreakEngine::divideUpDictionaryRange( UText *text,
...
```
FrankYFTang, line 474: If you got compiler error, change to 
                             UBool /* isPhraseBreaking */,

icu4c/source/common/dictbe.cpp
```diff
@@ -662,6 +665,7 @@ BurmeseBreakEngine::divideUpDictionaryRange( UText *text,
...
```
FrankYFTang, line 668: If you got compiler error, change to 
                             UBool /* isPhraseBreaking */,

icu4c/source/common/dictbe.cpp
```diff
@@ -867,6 +871,7 @@ KhmerBreakEngine::divideUpDictionaryRange( UText *text,
...
```
FrankYFTang, line 874: If you got compiler error, change to 
                             UBool /* isPhraseBreaking */,

icu4c/source/common/lstmbe.cpp
```diff
@@ -639,6 +639,7 @@ LSTMBreakEngine::divideUpDictionaryRange( UText *text,
                                                 int32_t startPos,
                                                 int32_t endPos,
                                                 UVector32 &foundBreaks,
+                                                UBool isPhraseBreaking,
```
FrankYFTang, line 642: If you got compiler error, change to 
                             UBool /* isPhraseBreaking */,

icu4c/source/common/rbbi.cpp
```diff
@@ -322,6 +334,7 @@ void RuleBasedBreakIterator::init(UErrorCode &status) {
     fBreakCache           = nullptr;
     fDictionaryCache      = nullptr;
     fLookAheadMatches     = nullptr;
+    fIsPhraseBreaking     = FALSE;
```
FrankYFTang, line 337: user lower case false for now code.

icu4c/source/common/unicode/rbbi.h
```diff
@@ -163,6 +168,21 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
...
 
+    /**
+     * This constructor uses the udata interface to create a BreakIterator
+     * whose internal tables live in a memory-mapped file.  "image" is an
+     * ICU UDataMemory handle for the pre-compiled break iterator tables.
+     * @param image handle to the memory image for the break iterator data.
+     *        Ownership of the UDataMemory handle passes to the Break Iterator,
+     *        which will be responsible for closing it when it is no longer needed.
+     * @param status Information on any errors encountered.
+     * @param isPhraseBreaking TRUE if phrase based breaking is required, otherwise FALSE.
```
FrankYFTang, line 179: use lowercase true and false for new code.

icu4c/source/test/intltest/lstmbetst.cpp
```diff
@@ -153,7 +153,7 @@ void LSTMBETest::runTestFromFile(const char* filename) {
                     dataerrln("%s:%d Error %s Could not allocate UVextor32", __FILE__, __LINE__, u_errorName(status));
                     return;
                 }
-                engine->findBreaks(&ut, 0, value.length(), actual, status);
+                engine->findBreaks(&ut, 0, value.length(), actual, FALSE, status);
```
FrankYFTang, line 156: use lowercase false for new code

icu4c/source/test/intltest/lstmbetst.cpp
```diff
@@ -288,7 +288,7 @@ void LSTMBETest::runTestWithLargeMemory( const char* model, UScriptCode script)
             return;
         }
 
-        engine->findBreaks(&ut, 0, text.length(), actual, status);
+        engine->findBreaks(&ut, 0, text.length(), actual, FALSE, status);
```
FrankYFTang, line 291: use lowercase fasle for new code

icu4c/source/common/rbbi_cache.cpp
FrankYFTang: About Mark's comment, take a look at   UnicodeSet    fSet; in DictionaryBreakEngine :
and how is it used in 
```
UBool
DictionaryBreakEngine::handles(UChar32 c) const {
    return fSet.contains(c);
}
```

The code which construct them now in each LanguageBreakEngine seems has some problem (I see no point to keep them in the member data and please ignore those.

FrankYFTang: I found some of the UnicodeSet usage is wrong when I try to look at example for you. Try to correct them in https://github.com/unicode-org/icu/pull/1941

FrankYFTang:
```
UnicodeSet cjkCharacters(UNICODE_STRING_SIMPLE("[[:Han:][:Hiragana:][:Hangul:][:Katakana:]]"), status);
```
should match all the Han, Hiragana, Katakana, Hangul letters, if you want to add more punchation to this list you may need to add some more to the set. 

I am not sure they have [:Jamo:] or not . read the doc to figure out what should you put into the set.
See
https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeSet.html

FrankYFTang: read https://unicode-org.github.io/icu/userguide/strings/unicodeset.html
@markusicu  - is there a clear documentation to list Hiragana, Katakana and other acceptable id used inside UnicodeSet somewhere. I saw a lot of indirect link but it is rather hard to find the table which list all the acceptable keys. For example, how can I tell [:Jamo:] or [:FooBar:] is an acceptable expression or not from the documentation ?

allensu05: Please help review again. I integrated #1941 and used DictionaryBreakEngine#handles() to implement isCjk()

icu4c/source/common/rbbi.cpp
allensu05: Resolved

icu4c/source/common/brkiter.cpp
```diff
@@ -433,10 +433,18 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
             char lbKeyValue[kKeyValueLenMax] = {0};
             UErrorCode kvStatus = U_ZERO_ERROR;
             int32_t kLen = loc.getKeywordValue("lb", lbKeyValue, kKeyValueLenMax, kvStatus);
-            if (U_SUCCESS(kvStatus) && kLen > 0 && (uprv_strcmp(lbKeyValue,"strict")==0 || uprv_strcmp(lbKeyValue,"normal")==0 || uprv_strcmp(lbKeyValue,"loose")==0)) {
+            if (U_SUCCESS(kvStatus) && kLen > 0 && (uprv_strcmp(lbKeyValue,"strict")==0 || uprv_strcmp(lbKeyValue,"normal")==0 || uprv_strcmp(lbKeyValue,"loose")==0 || uprv_strcmp(lbKeyValue,"unit")==0)) {
```
FrankYFTang, line 436: so... we are adding a lb=unit and lw=phrase here?

icu4c/source/common/dictbe.cpp
```diff
@@ -1090,6 +1097,9 @@ static inline bool isKatakana(UChar32 value) {
             (value >= 0xFF66 && value <= 0xFF9f);
 }
 
+static inline bool isHiragana(UChar32 value) {
+    return value >= 0x3041 && value <= 0x309F;
```
FrankYFTang, line 1101: use UnicodeSet instead. See https://github.com/unicode-org/icu/commit/704415402a68be8bbc143e5c78c3102b0d40ba90
about the right way to set up Unicode set. If you need to check isHiragana. then you need to
1.  add   UnicodeSet                fHiraganaWordSet; back to CjkBreakEngine in the .h file
2.  in the constructor
fHiraganaWordSet.applyPattern(UnicodeString(u"[:Hiragana:]"), status);
fHiraganaWordSet.compact();
3. in the code where you need to check that char is a hiragana
instead of isHiragana(c), call fHiraganaWordSet.contains(c)

See https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeSet.html

icu4c/source/common/dictbe.cpp
```diff
@@ -1385,6 +1406,26 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
     // inputMap goes out of scope
     return numBreaks;
 }
+
+void CjkBreakEngine::loadJpParticleAndAuxVerbs() {
```
FrankYFTang, line 1410: Where do you call this function, I cannot findit.

icu4c/source/common/dictbe.cpp
```diff
@@ -1342,11 +1353,21 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
         t_boundary.addElement(numCodePts, status);
         numBreaks++;
     } else {
-        for (int32_t i = numCodePts; i > 0; i = prev.elementAti(i)) {
+        t_boundary.addElement(numCodePts, status);
```
FrankYFTang, line 1356: it seems your change will alternative the current behavior of word break , more than what your project intent to change, is that correct?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -207,6 +213,7 @@ void RuleBasedBreakIterator::DictionaryCache::populateDictionary(int32_t startPo
 RuleBasedBreakIterator::BreakCache::BreakCache(RuleBasedBreakIterator *bi, UErrorCode &status) :
         fBI(bi), fSideBuffer(status) {
     reset();
+    fCjkEngine = fBI->getLanguageBreakEngine(0x3041);
```
FrankYFTang, line 216: so, before your PR, if the user only use RuleBasedBreakIterator to break non CJK text, the CjBreakEngine will never be construct because there are no CJ text calling into RuleBasedBreakIterator and therefore we will never callfBI->getLanguageBreakEngine() with a unicode which in that CJ text range. But with your change, now you ALWAYS force to create a CjBreakEngine here , and therefore ALWAYS load megabytes of cj dictionaries into memory even if your caller is just try to break "Hello Word". This mean all the western language users now will load the cj dictionary and create the CjBreakEngine even they never come across any cj text. I do not think this practice is acceptable. It will slow down the performance for not target user as well as increase memory usage on user do not use CJ text.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +427,133 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
     return true;
 }
 
+static inline bool isNumber(UChar32 ch) {
+    static UChar32 range[2][2] = {
```
FrankYFTang, line 431: I think you should construct a UnicodeSet (private static) in RuleBasedBreakIterator::BreakCache to test the number. The cost for doing so won't be huge.  @markusicu What is the right way to specify all number? "[:N:]" ? or should he use "[:Nd:]" ?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +427,133 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+    };
+    for(int32_t i = 0 ; i < 2 ; i++) {
+        if(ch >= range[i][0] && ch <= range[i][1])
+            return true;
+    }
+    return false;
+}
+
+UBool RuleBasedBreakIterator::BreakCache::isCjk(UChar32 ch) {
+    return fCjkEngine->handles(ch);
```
FrankYFTang, line 443: As I explain before, this approach force RuleBasedBreakIterator::BreakCache in any condiction to call fBI->getLanguageBreakEngine() to construct a CjBreakEngine, even no text are in that range, this is not acceptable because that path will load mega bytes of dictionary into memory while not needed at all. Why don't you just create your own UnicodeSet for that. Just
UnicodeSet cjSet(UnicodeString(u"[[:Han:][:Hiragana:][:Katakana:]\\u30fc\\uff70\\uff9e\\uff9f]"), status);
but keep it a "private static" so you only need to create once.

FrankYFTang:
You need to find a way to measure the performance w/ and w/o of your changes, in particular, how it impact NON CJ users since your code change common code path, not just inside CJBreakEngine. Also what is the (quality+memory usage+performance) impact of your change to word break in CJ (not line break) or not the phrase based one

icu4c/source/common/dictbe.cpp
allensu05: Actually, it doesn't change.
Before: line-1345 and 1346: for loop and addElement
After: line-1355, 1367: add the first element, for loop, and add the remaining elements when necessary.
When isPhraseBreaking is false, everything is identical.
When isPhraseBreaking is true, skip JP particle and Hiragana.

icu4c/source/common/rbbi_cache.cpp
allensu05: Resolved

icu4c/source/common/rbbi_cache.cpp
allensu05: Resolved. 
(Both  Nd and N could work)

icu4c/source/common/rbbi_cache.cpp
allensu05: Thanks. Good suggestion.

icu4c/source/common/dictbe.cpp
allensu05: Resolved. It's in the constructor.
(Something lost as fixing conflict)

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/brkiter.cpp
allensu05: Resolved.

icu4c/source/common/dictbe.cpp
```diff
@@ -1385,6 +1405,26 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
     // inputMap goes out of scope
     return numBreaks;
 }
+
+void CjkBreakEngine::loadJpParticleAndAuxVerbs() {
+    UErrorCode error = U_ZERO_ERROR;
+    const char* tags[2] = {"particles", "auxVerbs"};
+
+    ResourceBundle ja(U_ICUDATA_BRKITR, "ja", error);
```
FrankYFTang, line 1413: so, in the way you read the resource, you get the UnicodeString from the resources, and then put into fParticleSet, which is a Hashtable and you always put in true for that. I think you should call fParticleSet.puti(str, 1, error);  instead, which will save some memory since the value is always number.

icu4c/source/common/dictbe.cpp
```diff
@@ -1385,6 +1405,26 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+    if (U_SUCCESS(error)) {
+       for(int32_t i = 0 ; i < 2 ; i++) {
+           ResourceBundle bundle = ja.get(tags[i], error);
+           if (U_SUCCESS(error)) {
```
FrankYFTang, line 1417: ICU api internally follow the check first fail pattern, so you do not need to check U_SUCCESS if there are no dependency other than the success code
I think you can simplify
```
           if (U_SUCCESS(error)) {
               while(bundle.hasNext()){
                   UnicodeString str=bundle.getNextString(error);
                   if(U_SUCCESS(error)){
                       fParticleSet.put(str, (void*)true, error);
                   }
               }
           }
```
to
```
```
               while(U_SUCCESS(error) && bundle.hasNext()){
                   fParticleSet.puti(bundle.getNextString(error), 1, error);
               }
```

File: icu4c/source/common/rbbi_cache.cpp [hunk 34]
```diff
@@ -420,50 +426,141 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+        if (U_SUCCESS(status)) {
+            numberSet.compact();
+        }
+    }
+    return numberSet.contains(ch);
+}
+
+static inline bool isCj(UChar32 ch) {
+    UErrorCode status = U_ZERO_ERROR;
+    static UnicodeSet cjSet;
```
FrankYFTang, line 443: I don't think this is thread safe. you have to use  static icu::UInitOnce with umtx_initOnce to create the UnicodeSet. Otherwise it will cause race condiction to break TSAN under multiple thread env.
Take a look at how is "static UnicodeSet *uni32Singleton" to created and used in common/uniset_props.cpp
In there the code also create a UnicodeSet, very similar to your use case. do the same thing here. 
you need to have a static UnicodeSet * and have a static icu::UInitOnce that = to  U_INITONCE_INITIALIZER
you need to have a static UBool U_CALLCONV  function to clean up your UnicodeSet when the library is unloaded in a dynamic loading env.

and look at 
```
// Cache some sets for other services -------------------------------------- ***
void U_CALLCONV createUni32Set(UErrorCode &errorCode) {
    U_ASSERT(uni32Singleton == NULL);
    uni32Singleton = new UnicodeSet(UNICODE_STRING_SIMPLE("[:age=3.2:]"), errorCode);
    if(uni32Singleton==NULL) {
        errorCode=U_MEMORY_ALLOCATION_ERROR;
    } else {
        uni32Singleton->freeze();
    }
    ucln_common_registerCleanup(UCLN_COMMON_USET, uset_cleanup);
}


U_CFUNC UnicodeSet *
uniset_getUnicode32Instance(UErrorCode &errorCode) {
    umtx_initOnce(uni32InitOnce, &createUni32Set, errorCode);
    return uni32Singleton;
}
```

icu4c/source/common/unicode/rbbi.h
```diff
@@ -163,6 +168,21 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
...
+     * @param image handle to the memory image for the break iterator data.
+     *        Ownership of the UDataMemory handle passes to the Break Iterator,
+     *        which will be responsible for closing it when it is no longer needed.
+     * @param status Information on any errors encountered.
+     * @param isPhraseBreaking true if phrase based breaking is required, otherwise false.
+     * @see udata_open
+     * @see #getBinaryRules
+     * @stable ICU 2.8
+     */
+    RuleBasedBreakIterator(UDataMemory* image, UErrorCode &status, UBool isPhraseBreaking);
```
aheninger, line 184: I'm unsure about surfacing isPhraseBreaking at this high of a level. It seems very specific to a one specialized type of break iterator, not something that would make sense for any rule based break iterator. And all do come through this path.  I haven't thought out possible alternatives, though.

Also, for functions with a UErrorCode parameter, it should be last.

icu4c/source/common/rbbi_cache.cpp
aheninger: Yes, what Frank said.
ICU code cannot have static instances of objects with constructors or destructors. Statics must be pointers or simple primitive types.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +426,141 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+    return (UBool)((CAT_MASK(props)&U_GC_OP_MASK)!=0);
+}
+
+static UBool u_isClosePunct(UChar32 c) {
+    uint32_t props;
+    GET_PROPS(c, props);
+    return (UBool)((CAT_MASK(props)&U_GC_CP_MASK)!=0);
+}
+
+UBool RuleBasedBreakIterator::BreakCache::handleNumber(int32_t pos) {
```
aheninger, line 474: I would like to figure out some way of moving logic like this, which is specific to one type of break iterator, out of this level (BreakCache), which is completely general to all break iterators. If we can figure out how, it would be better appearing in the Break Engine and/or the rules for its type of break iterator.

I have some partial thoughts, but they aren't worked out. I'll see if I can come up with something.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +426,141 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+        }
+    }
+    return cjSet.contains(ch);
+}
+
+/* getting a uint32_t properties word from the data */
+#define GET_PROPS(c, result) ((result)=UTRIE2_GET16(&propsTrie, c))
+
+/** Mask constant for multiple UCharCategory bits (OP Open Punctuation). @stable ICU 2.1 */
+#define U_GC_OP_MASK \
```
aheninger, line 457: #defines like this one are better done as a C++ `constexpr`

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -420,50 +426,141 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
     }
 
     fBI->fPosition = fromPosition;
-    pos = fBI->handleNext();
-    if (pos == UBRK_DONE) {
-        return FALSE;
-    }
+    do {
+        pos = fBI->handleNext();
+        while(fBI->fIsPhraseBreaking
```
aheninger, line 508: I suspect that the situation being addressed by this (and the following) logic may be a somewhat more wide-spread problem with any dictionary-based line breaking. The line breaking rules regarding open & close punctuation all fall apart once a dictionary is involved. Again, needs some further thought.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -4276,6 +4277,101 @@ void RBBITest::TestBug5532(void)  {
 }
 
 
+void RBBITest::TestJpParticleAndPunctuation(void)  {
+    //Expected breaking result:［京都▁観光］▁時雨▁殿に▁行った。
+
+    const char16_t *s16 = {u"［京都観光］時雨殿に行った。"};
```
aheninger, line 4283: These test cases can be added to the test file `source/test/testdata/rbbitst.txt`, without the need to write any code.

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.cpp
allensu05: Thanks. Resolved.

icu4c/source/common/rbbi_cache.cpp
allensu05: Thanks. Resolved

icu4c/source/data/brkitr/rules/line_unit_cj.txt
```diff
@@ -0,0 +1,371 @@
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
```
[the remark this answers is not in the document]
allensu05: Removed.
(Something lost as fixing conflict)

icu4c/source/data/brkitr/rules/line_phrase_cj.txt
```diff
@@ -0,0 +1,375 @@
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+# Copyright (c) 2002-2016  International Business Machines Corporation and
+# others. All Rights Reserved.
+#
+#  file:  line_cj.txt
```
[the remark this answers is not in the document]
allensu05: Resolved.

icu4c/source/common/rbbi_cache.cpp
allensu05: Move the logic to CjkBreakEngine.

icu4c/source/common/dictbe.cpp
```diff
@@ -1053,13 +1064,18 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
...
+    fNumberWordSet.applyPattern(UnicodeString(u"[:Nd:]"), status);
+    fNumberWordSet.compact();
 
     // handle Korean and Japanese/Chinese using different dictionaries
     if (type == kKorean) {
         if (U_SUCCESS(status)) {
             setCharacters(fHangulWordSet);
         }
     } else { //Chinese and Japanese
+        loadJpParticleAndAuxVerbs();
```
FrankYFTang, line 1078: Change function name to loadJapaneseParticleAndAuxVerbs please

icu4c/source/common/dictbe.cpp
```diff
@@ -17,14 +17,20 @@
...
 #include "uvectr32.h"
 #include "uvector.h"
 #include "uassert.h"
 #include "unicode/normlzr.h"
 #include "cmemory.h"
 #include "dictionarydata.h"
+#include "utrie2.h"
+#include "uprops.h"
+#define INCLUDED_FROM_UCHAR_C
+#include "uchar_props_data.h"
```
FrankYFTang, line 33: @markusicu  I am concern of this usage. Any opinion? should we just use UnicodeSet to replace such usage instead?

icu4c/source/common/dictbe.cpp
```diff
@@ -1090,14 +1106,31 @@ static inline bool isKatakana(UChar32 value) {
...
 }
 
+/** Mask constant for multiple UCharCategory bits (OP Open Punctuation). @stable ICU 2.1 */
+constexpr int32_t U_GC_OP_MASK = U_GC_PI_MASK | U_GC_PS_MASK;
+
+/** Mask constant for multiple UCharCategory bits (CP Close Punctuation). @stable ICU 2.1 */
+constexpr int32_t U_GC_CP_MASK =
+        U_GC_PD_MASK | U_GC_PE_MASK | U_GC_PC_MASK | U_GC_PO_MASK | U_GC_PF_MASK;
+
+static inline bool isOpenPunct(UChar32 c) {
```
FrankYFTang, line 1123: As I asked in the ICU4J review, should we use UnicodeSet here also the check the Punc mark instead? What is the benefit of using UTRIE2_GET16 and access propsTrie ?

icu4c/source/common/brkiter.cpp
```diff
@@ -437,6 +437,14 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
                 uprv_strcat(lbType, "_");
                 uprv_strcat(lbType, lbKeyValue);
             }
+            if(strcmp(lbType, "line") == 0) {
```
FrankYFTang, line 440: add " " between "if" and "(" please

icu4c/source/common/brkiter.cpp
```diff
@@ -437,6 +437,14 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
...
+                char lwKeyValue[kKeyValueLenMax] = {0};
+                kLen = loc.getKeywordValue("lw", lwKeyValue, kKeyValueLenMax, kvStatus);
+                if (U_SUCCESS(kvStatus) && kLen > 0 && uprv_strcmp(lwKeyValue,"phrase")==0) {
```
FrankYFTang, line 443: add " " before and after "==" please

icu4c/source/common/dictbe.cpp
```diff
@@ -1342,11 +1376,22 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
     } else {
-        for (int32_t i = numCodePts; i > 0; i = prev.elementAti(i)) {
+        t_boundary.addElement(numCodePts, status);
+        numBreaks++;
+        int32_t prevIdx = numCodePts;
+
+        for (int32_t i = prev.elementAti(numCodePts); i > 0; i = prev.elementAti(i)) {
+            int length = prevIdx -  i;
+            prevIdx = i;
+            //skip the breaking point if it belongs to the particle or Hiragana.
```
FrankYFTang, line 1386: Add " " after "//". Also, change "skip" to "Skip"

icu4c/source/common/dictbe.cpp
```diff
@@ -1342,11 +1376,22 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+            if(isPhraseBreaking
+                    && (fParticleSet.containsKey(inString.tempSubString(i, length))
```
FrankYFTang, line 1388: As I mentioned in ICU4J review, could we merge fParticleSet and  fHiraganaWordSet into one fSkipSet and check only once by calling fSkipSet.containsKey() ? 
1. Remove fHiraganaWordSet from member data and the check here.
2. Rename fParticleSet to fSkipSet
3. Construct a on stack hiraganaWordSet in the constructor as how you  construct fHiraganaWordSet now but could be destrory after the constructor.
4. Call https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeSetIterator.html w/ hiraganaWordSet and add the iteration to fSkipSet before or after loading the paractical from resources.

icu4c/source/common/dictbe.cpp
```diff
@@ -1342,11 +1376,22 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
         t_boundary.addElement(numCodePts, status);
         numBreaks++;
     } else {
```
FrankYFTang, line 1378: could we keep to original "} else {" block unchange but insert a } else if (isPhraseBreaking) {" block for the new code?

icu4c/source/common/dictbe.cpp
```diff
@@ -1342,11 +1376,22 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
-        for (int32_t i = numCodePts; i > 0; i = prev.elementAti(i)) {
+        t_boundary.addElement(numCodePts, status);
+        numBreaks++;
+        int32_t prevIdx = numCodePts;
+
+        for (int32_t i = prev.elementAti(numCodePts); i > 0; i = prev.elementAti(i)) {
+            int length = prevIdx -  i;
+            prevIdx = i;
+            //skip the breaking point if it belongs to the particle or Hiragana.
+            if(isPhraseBreaking
```
FrankYFTang, line 1387: add " " between "if" and "("

icu4c/source/common/dictbe.cpp
```diff
@@ -1369,7 +1415,12 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
         if (utextPos > prevUTextPos) {
             // Boundaries are added to foundBreaks output in ascending order.
             U_ASSERT(foundBreaks.size() == 0 || foundBreaks.peeki() < utextPos);
-            foundBreaks.push(utextPos, status);
+            if(!foundBreaks.contains(utextPos)
```
FrankYFTang, line 1418: Add " " between "if" and "(".

icu4c/source/common/dictbe.cpp
```diff
@@ -1369,7 +1415,12 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+                    && (utextPos != rangeStart
+                    || (utextPos == rangeStart  && isClosePunct(utext_char32At(inText,utextPos-1))))) {
```
FrankYFTang, line 1420: remove extra " " before "&&", you should have only one " " between "rangeStart" and "&&", not two.
add " " between "inText," and "utextPos".
add " " before and after "-"

icu4c/source/common/dictbe.cpp
FrankYFTang, line 1078: Move this line into the if (U_SUCCESS(status)) { block and pass the status by reference to that function.

icu4c/source/common/dictbe.cpp
```diff
@@ -1053,13 +1064,18 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
     // Korean dictionary only includes Hangul syllables
     fHangulWordSet.applyPattern(UnicodeString(u"[\\uac00-\\ud7a3]"), status);
     fHangulWordSet.compact();
+    fHiraganaWordSet.applyPattern(UnicodeString(u"[:Hiragana:]"), status);
```
FrankYFTang, line 1067: I think your fHiraganaWordSet do not need to be a member variable but just a on stack object in the constructor, after your call loadJapaneseParticleAndAuxVerbs you then create the hiraganaWordSet on the stack the same way and then you can call https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeSetIterator.html to get out the content and put that into fSkipSet. Since your fParticleSet and fHiraganaWordSet now are use for the same reason and there are no reason we have two different check, we should just have one HashSet named fSkipSet an load both the particle and verbs + the Hiragana into it and check that once only.

icu4c/source/common/dictbe.cpp
```diff
@@ -1381,9 +1432,33 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+    }
+
     // inString goes out of scope
     // inputMap goes out of scope
-    return numBreaks;
+    return correctedNumBreaks;
+}
+
+void CjkBreakEngine::loadJpParticleAndAuxVerbs() {
+    UErrorCode error = U_ZERO_ERROR;
```
FrankYFTang, line 1450: The error code should be pass in from the constructor by reference instead adding a new one in the constructor and got ignored later. If any error happen, the caller of the constructor should get the error.

icu4c/source/common/dictbe.cpp
```diff
@@ -1381,9 +1432,33 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+    const char* tags[2] = {"particles", "auxVerbs"};
+
+    ResourceBundle ja(U_ICUDATA_BRKITR, "ja", error);
+    if (U_SUCCESS(error)) {
+       for(int32_t i = 0 ; i < 2 ; i++) {
```
FrankYFTang, line 1455: add " " between "for" and (" and remove the " " between "0" and "; and "2" and ";".

icu4c/source/common/dictbe.cpp
```diff
@@ -1381,9 +1432,33 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+           ResourceBundle bundle = ja.get(tags[i], error);
+           while(U_SUCCESS(error) && bundle.hasNext()){
```
FrankYFTang, line 1457: add " " between "while" and "("

icu4c/source/common/dictbe.h
```diff
@@ -362,10 +369,19 @@ class CjkBreakEngine : public DictionaryBreakEngine {
      * @internal
      */
   UnicodeSet                fHangulWordSet;
+  UnicodeSet                fHiraganaWordSet;
```
FrankYFTang, line 372: As I mention elsewhere, since fHiraganaWordSet and fParticleSet are used in the same place, you should merge them into one Hashtable fSkipSet and load both content into it.

icu4c/source/common/dictbe.h
```diff
@@ -362,10 +369,19 @@ class CjkBreakEngine : public DictionaryBreakEngine {
...
+  UnicodeSet                fNumberWordSet;
 
   DictionaryMatcher        *fDictionary;
   const Normalizer2        *nfkcNorm2;
 
+ private:
+ /**
+  * <p> Load Japanese particles and auxiliary verbs into fParticleSet.
```
FrankYFTang, line 380: Rename to loadJapaneseParticleAndAuxVerbs

icu4c/source/common/unicode/rbbi.h
```diff
@@ -163,6 +168,21 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
...
+     * whose internal tables live in a memory-mapped file.  "image" is an
+     * ICU UDataMemory handle for the pre-compiled break iterator tables.
+     * @param image handle to the memory image for the break iterator data.
+     *        Ownership of the UDataMemory handle passes to the Break Iterator,
+     *        which will be responsible for closing it when it is no longer needed.
+     * @param status Information on any errors encountered.
+     * @param isPhraseBreaking true if phrase based breaking is required, otherwise false.
+     * @see udata_open
+     * @see #getBinaryRules
+     * @stable ICU 2.8
```
FrankYFTang, line 182: Remve the "@stable ICU 2.8" line, this is NOT true. You are adding this NEW function here for 71.1

FrankYFTang: Add "@internal (private)"

icu4c/source/data/brkitr/rules/line_phrase_cj.txt
```diff
@@ -0,0 +1,379 @@
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
```
FrankYFTang, line 1: Change to 2022

icu4c/source/data/brkitr/rules/line_phrase_cj.txt
```diff
@@ -0,0 +1,379 @@
...
+# License & terms of use: http://www.unicode.org/copyright.html
+# Copyright (c) 2002-2016  International Business Machines Corporation and
```
FrankYFTang, line 3: Remove these two line of IBM copyright since this is a new file

icu4c/source/common/rbbi.cpp
```diff
@@ -82,6 +82,18 @@ RuleBasedBreakIterator::RuleBasedBreakIterator(RBBIDataHeader* data, UErrorCode
     }
 }
 
+//-------------------------------------------------------------------------------
+//
+//   Constructor   from a UDataMemory handle to precompiled break rules
+//                 stored in an ICU data file.
```
FrankYFTang, line 88: Please also mention this is only intend for private internal usage

FrankYFTang (changes requested):

icu4c/source/common/rbbi.cpp
allensu05: Resolved

icu4c/source/data/brkitr/rules/line_phrase_cj.txt
allensu05: Resolved

icu4c/source/data/brkitr/rules/line_phrase_cj.txt
allensu05: Resolved

icu4c/source/common/unicode/rbbi.h
allensu05: Resolved

icu4c/source/common/dictbe.h
allensu05: Resolved

icu4c/source/common/brkiter.cpp
allensu05: Resolved

icu4c/source/common/brkiter.cpp
allensu05: Resolved

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/brkiter.cpp
```diff
@@ -437,6 +437,14 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
                 uprv_strcat(lbType, "_");
                 uprv_strcat(lbType, lbKeyValue);
             }
+            if (strcmp(lbType, "line") == 0) {
+                char lwKeyValue[kKeyValueLenMax] = { 0 };
```
FrankYFTang, line 441: CharString lwKeyValue;
CharStringByteSink valueSink(& lwKeyValue);
loc.getKeywordValue("lw", valueSink, kvStatus);
if (U_SUCCESS(kvStatus) && lwKeyValue == "phrase") {
   uprv_strcat(lbType, "_");
    uprv_strcat(lbType, lwKeyValue.data());
}

see similar usage in line 1732 of source/common/locid.cpp

FrankYFTang, line 441: Could you just reuse the lbKeyValue buffer  here? the content in that buffer is already no longer need after the if block above, right? you can reuse it here. Of course, you better rename that from lbKeyValue to just keyValue .
or you can rewrite whole block to
```
case UBRK_LINE:
        {
            UTRACE_ENTRY(UTRACE_UBRK_CREATE_LINE);
            uprv_strcpy(lbType, "line");
            UErrorCode kvStatus = U_ZERO_ERROR;
            CharString value;
            CharStringByteSink valueSink(&value);
            loc.getKeywordValue("lb", valueSink, kvStatus);
            if (U_SUCCESS(kvStatus) && (value == "strict" || value == "normal" || value == "loose")) {
                uprv_strcat(lbType, "_");
                uprv_strcat(lbType, value.data());
            } else {
               // lbType is still "line"
                value.clear();
                loc.getKeywordValue("lw", valueSink, kvStatus);
                if (U_SUCCESS(kvStatus) && value == "phrase") {
                    uprv_strcat(lbType, "_");
                    uprv_strcat(lbType, value.data());
                }
            }
            result = BreakIterator::buildInstance(loc, lbType, status);

            UTRACE_DATA1(UTRACE_INFO, "lb=%s", lbKeyValue);
            UTRACE_EXIT_STATUS(status);
        }
        break;
```

icu4c/source/common/dictbe.cpp
```diff
@@ -1341,6 +1374,21 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
         numBreaks++;
+    } else if (isPhraseBreaking) {
+        t_boundary.addElement(numCodePts, status);
+        numBreaks++;
+        int32_t prevIdx = numCodePts;
+
+        for (int32_t i = prev.elementAti(numCodePts); i > 0; i = prev.elementAti(i)) {
+            int length = prevIdx - i;
+            prevIdx = i;
+            // Skip the breakpoint if it belongs to the particle or Hiragana.
```
FrankYFTang, line 1385: Since the logic after the if block is very simple, how about flip the if condiction (by adding !  in front) and just put that part inside the new if block instead? 
```
            // Skip the breakpoint if it belongs to the particle or Hiragana.
            if (!fSkipSet.containsKey(inString.tempSubString(i, length))) {
                t_boundary.addElement(i, status);
                numBreaks++;
            }
```

icu4c/source/common/dictbe.cpp
```diff
@@ -1361,15 +1409,24 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
         int32_t cpPos = t_boundary.elementAti(i);
         U_ASSERT(cpPos > prevCPPos);
         int32_t utextPos =  inputMap.isValid() ? inputMap->elementAti(cpPos) : cpPos + rangeStart;
         U_ASSERT(utextPos >= prevUTextPos);
         if (utextPos > prevUTextPos) {
             // Boundaries are added to foundBreaks output in ascending order.
             U_ASSERT(foundBreaks.size() == 0 || foundBreaks.peeki() < utextPos);
-            foundBreaks.push(utextPos, status);
+            // In phrase breaking, there has to be a breakpoint between Cj character and
+            // close punctuation e.g.［京都観光］▁時雨
```
FrankYFTang, line 1422: It is good you include an example here, but your comment is not clear. Please write more to explain in that example, what you intend to do here. I cannot tell. what is _ here . Do you mean you need to add a break point between "]" and "時雨"?

icu4c/source/common/dictbe.cpp
```diff
@@ -1381,9 +1438,52 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
     }
     (void)prevCPPos; // suppress compiler warnings about unused variable
 
+    UChar32 lastChar = utext_char32At(inText, rangeEnd);
+
+    if (!foundBreaks.isEmpty() && foundBreaks.peeki() == rangeEnd) {
+        // In phrase breaking, there has to be a breakpoint between Cj character and
```
FrankYFTang, line 1444: Great to add example, but please rewrite your comment to make it clear what you intend to do here. Do you mean for the case of "１枚４０００円, 文字「そうだ...」" we need to insert a break between "枚" and "４" and a break between "字" and "「". If so, write down the example, and write dow where you need to insert the break. Your comment is too vague.

FrankYFTang: Also, how about also add that test case into test/testdata/rbbitst.txt

icu4c/source/common/dictbe.cpp
FrankYFTang: Also, how about also add that test case to test/testdata/rbbitst.txt

FrankYFTang (changes requested):

icu4c/source/common/brkiter.cpp
allensu05: Thanks. Resolved

icu4c/source/common/dictbe.cpp
allensu05: I could not understand. Could you specify more clearly?

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.cpp
allensu05: Resolved

FrankYFTang:
I saw the following errors in gcc-debug-build-and-test and clang-release-build-and-test
```
Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::next() but breakiterator does not depend on usetiter (for common/usetiter.o)
Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::~UnicodeSetIterator() but breakiterator does not depend on usetiter (for common/usetiter.o)
Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::UnicodeSetIterator(icu_70::UnicodeSet const&) but breakiterator does not depend on usetiter (for common/usetiter.o)
```

Please address them

FrankYFTang:
> I saw the following errors in gcc-debug-build-and-test and clang-release-build-and-test
> 
> ```
> Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::next() but breakiterator does not depend on usetiter (for common/usetiter.o)
> Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::~UnicodeSetIterator() but breakiterator does not depend on usetiter (for common/usetiter.o)
> Error: in group breakiterator: common/dictbe.o imports icu_70::UnicodeSetIterator::UnicodeSetIterator(icu_70::UnicodeSet const&) but breakiterator does not depend on usetiter (for common/usetiter.o)
> ```
> 
> Please address them

You need to change icu4c/source/test/depstest/dependencies.txt to fix that.

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.h
```diff
@@ -362,10 +369,23 @@ class CjkBreakEngine : public DictionaryBreakEngine {
      * @internal
      */
   UnicodeSet                fHangulWordSet;
+  UnicodeSet                fNumberWordSet;
```
FrankYFTang, line 372: as I mention in the Java PR, since fNumberWordSet  and fOpenPunctuationSet are use exactly in the same place, we should merge them into one UnicodeSet fNumberOrOpenPunctuationSet w/ u"[[:Nd:][:Pi:][:Ps:]]" pattern

icu4c/source/common/dictbe.cpp
```diff
@@ -1053,6 +1061,12 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
     // Korean dictionary only includes Hangul syllables
     fHangulWordSet.applyPattern(UnicodeString(u"[\\uac00-\\ud7a3]"), status);
     fHangulWordSet.compact();
+    fNumberWordSet.applyPattern(UnicodeString(u"[:Nd:]"), status);
```
FrankYFTang, line 1064: as I mention in the Java PR, since fNumberWordSet and fOpenPunctuationSet are use exactly in the same place, we should merge them into one UnicodeSet fNumberOrOpenPunctuationSet w/ u"[[:Nd:][:Pi:][:Ps:]]" pattern

icu4c/source/common/dictbe.cpp
```diff
@@ -1361,16 +1389,22 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
         if (utextPos > prevUTextPos) {
             // Boundaries are added to foundBreaks output in ascending order.
             U_ASSERT(foundBreaks.size() == 0 || foundBreaks.peeki() < utextPos);
-            if (!(foundBreaks.contains(utextPos) || utextPos == rangeStart)) {
+            // In phrase breaking, there has to be a breakpoint between Cj character and close
+            // punctuation.
+            // E.g.［携帯電話］正しい選択 -> ［携帯▁電話］▁正しい▁選択 -> breakpoint between ］ and 正
+            if (!foundBreaks.contains(utextPos)
+                && (utextPos != rangeStart
+                       || (isPhraseBreaking && utextPos == rangeStart && utextPos > 0
```
FrankYFTang, line 1406: No need for "&& utextPos == rangeStart" here because if  utextPos != rangeStart then it won't get into the part after ||.   so you only need
```
if (!foundBreaks.contains(utextPos)
                && (utextPos != rangeStart
                       || (isPhraseBreaking && utextPos > 0
                              && fClosePunctuationSet.contains(utext_char32At(inText, utextPos - 1))))) {

File: icu4c/source/common/dictbe.cpp [hunk 70]
```diff
@@ -1385,15 +1419,54 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+    UChar32 nextChar = utext_char32At(inText, rangeEnd);
     if (!foundBreaks.isEmpty() && foundBreaks.peeki() == rangeEnd) {
-        foundBreaks.popi();
-        correctedNumBreaks--;
+        // In phrase breaking, there has to be a breakpoint between Cj character and
+        // the number/open punctuation.
+        // E.g. る文字「そうだ、京都」->る▁文字▁「そうだ、▁京都」-> breakpoint between 字 and「
+        // E.g. 乗車率９０％程度だろうか -> 乗車▁率▁９０％▁程度だ▁ろうか -> breakpoint between 率 and ９
+        if (isPhraseBreaking) {
+            if (!fOpenPunctuationSet.contains(nextChar) && !fNumberWordSet.contains(nextChar)) {
```
FrankYFTang, line 1429: merge fOpenPunctuationSet and fNumberWordSet to fNumberOrOpenPunctuationSet
so you only need to have one test
```
if (!fNumberOrOpenPunctuationSet.contains(nextChar)) {
```
here

FrankYFTang (changes requested):

FrankYFTang:
I think the c++ part is mostly there except few minor comments I mention. I have not yet carefully review the rule file itself. @aheninger  could you help on that?

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.cpp
allensu05: Resolved

icu4c/source/common/dictbe.h
allensu05: Resolved

icu4c/source/common/dictbe.cpp
```diff
@@ -1053,6 +1061,9 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
     // Korean dictionary only includes Hangul syllables
     fHangulWordSet.applyPattern(UnicodeString(u"[\\uac00-\\ud7a3]"), status);
     fHangulWordSet.compact();
+    fNumberAndOpenPunctuationSet.applyPattern(UnicodeString(u"[[:Nd:][:Pi:][:Ps:]]"), status);
```
FrankYFTang, line 1064: This should be named fNumberOrOpenPunctuationSet not fNumberAndOpenPunctuationSet since it match an Unicode is a Number OR it is an Open Punctuation but not require it to be BOTH (which is actually impossible)

icu4c/source/common/dictbe.h
```diff
@@ -362,10 +369,22 @@ class CjkBreakEngine : public DictionaryBreakEngine {
      * @internal
      */
   UnicodeSet                fHangulWordSet;
+  UnicodeSet                fNumberAndOpenPunctuationSet;
```
FrankYFTang, line 372: rename to fNumberOrOpenPunctuationSet please

icu4c/source/common/dictbe.cpp
```diff
@@ -1385,15 +1416,54 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+    UChar32 nextChar = utext_char32At(inText, rangeEnd);
     if (!foundBreaks.isEmpty() && foundBreaks.peeki() == rangeEnd) {
-        foundBreaks.popi();
-        correctedNumBreaks--;
+        // In phrase breaking, there has to be a breakpoint between Cj character and
+        // the number/open punctuation.
+        // E.g. る文字「そうだ、京都」->る▁文字▁「そうだ、▁京都」-> breakpoint between 字 and「
+        // E.g. 乗車率９０％程度だろうか -> 乗車▁率▁９０％▁程度だ▁ろうか -> breakpoint between 率 and ９
+        if (isPhraseBreaking) {
+            if (!fNumberAndOpenPunctuationSet.contains(nextChar)) {
```
FrankYFTang, line 1426: rename to fNumberOrOpenPunctuationSet please

FrankYFTang (changes requested):
beside the naming of  fNumberAndOpenPunctuationSet other C++ code LGTM. need to review the rule changes carefully next.

FrankYFTang (approved):
LGTM

FrankYFTang (dismissed):

FrankYFTang:
To review the line_phrase_cj.txt, I did a diff with line_cj.txt and here are the diff
```
@@ -57,7 +55,7 @@
 $NL = [:LineBreak =  Next_Line:];
 # NS includes CJ for CSS strict line breaking.
 $NS = [[:LineBreak =  Nonstarter:] $CJ];
-$NU = [[:LineBreak =  Numeric:] \uFF10 - \uFF19];
+$NU = [:LineBreak =  Numeric:];
 $OP = [[:LineBreak =  Open_Punctuation:] \u201c];
 $PO = [:LineBreak =  Postfix_Numeric:];
 $PR = [:LineBreak =  Prefix_Numeric:];
```
Is it intentional to remove "\uFF10 - \uFF19" from $NU? (comparing to line_cj.txt) why is that?

The rest of changes in the rule all seems reasonable. (some are just making the rule easier to read).

FrankYFTang (approved):
All LTGM . Just have a question about the change on $NU

FrankYFTang (dismissed):
LGTM

icu4c/source/common/dictbe.cpp
```diff
@@ -1063,6 +1075,7 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
         UnicodeSet cjSet(UnicodeString(u"[[:Han:][:Hiragana:][:Katakana:]\\u30fc\\uff70\\uff9e\\uff9f]"), status);
         if (U_SUCCESS(status)) {
             setCharacters(cjSet);
+            initJapanesePhraseParameter(status);
```
markusicu, line 1078: This should only be called for -lw-phrase, right? Otherwise this is unnecessary work, and there won't be data if it's not the line_phrase_cj file?
You can check `if (isPhraseBreaking)` before calling this function, or inside. I prefer before.

icu4c/source/common/dictbe.cpp
```diff
@@ -1341,6 +1353,20 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
     if ((uint32_t)bestSnlp.elementAti(numCodePts) == kuint32max) {
         t_boundary.addElement(numCodePts, status);
         numBreaks++;
+    } else if (isPhraseBreaking) {
+        t_boundary.addElement(numCodePts, status);
```
markusicu, line 1357: Please handle `if (U_FAILURE(status))`

icu4c/source/common/dictbe.cpp
```diff
@@ -1341,6 +1353,20 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
...
+        numBreaks++;
+        int32_t prevIdx = numCodePts;
+
+        for (int32_t i = prev.elementAti(numCodePts); i > 0; i = prev.elementAti(i)) {
+            int length = prevIdx - i;
+            prevIdx = i;
+            // Skip the breakpoint if it belongs to the particle or Hiragana.
+            if (!fSkipSet.containsKey(inString.tempSubString(i, length))) {
```
markusicu, line 1365: The naming of `numCodePts` suggests that this code counts code points, and that the length is a number of code points, but UnicodeString (and most of ICU) counts UTF-16 *code units*.

Please check (review the code and add a test case) that this works right, and/or add a comment about code unit vs. code point counting.

For example, here are the Hiragana/Katakana supplementary code points (each of these uses two UTF-16 code units):
[`[[:Hiragana:][:Katakana:]-[\u0000-\uFFFF]]`](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5B%3AHiragana%3A%5D%5B%3AKatakana%3A%5D-%5B%5Cu0000-%5CuFFFF%5D%5D&g=&i=)

icu4c/source/common/dictbe.cpp
```diff
@@ -1341,6 +1353,20 @@ CjkBreakEngine::divideUpDictionaryRange( UText *inText,
     if ((uint32_t)bestSnlp.elementAti(numCodePts) == kuint32max) {
         t_boundary.addElement(numCodePts, status);
         numBreaks++;
+    } else if (isPhraseBreaking) {
```
markusicu, line 1356: FYI
1. The logic here looks plausible, and makes sense given the implementation doc, but I am not familiar with the BreakIterator implementation, and I trust Frank and Andy for having helped get this right. Praise to Frank & Andy!
2. Praise to the implementers: I am very impressed by how little code is necessary for this feature!

icu4c/source/common/dictbe.cpp
FrankYFTang: The initJapanesePhraseParameter is called during the constructor. This object follow the singleton pattern and will only be created once (actually, twice, one for Korean case one for CJ case) and are shared by both the case the isPhraseBreaking is true or false and shared between thread. If we lazy load this part we then need to mutex it to protect thread safety. The cost of that may not be smaller than always call initJapanesePhraseParameter

markusicu: makes sense tnx

icu4c/source/data/brkitr/ja.txt
```diff
@@ -7,5 +7,16 @@ ja{
...
+    }
+    particles{
+         か, かしら, から, が, くらい, けれども, こそ,
+         さ, さえ, しか, だけ, だに, だの, て, で, でも,
+         と, ところが, とも, な, など, なり, に, ね, の,
+         ので, のに, は, ば, ばかり, へ, ほど, まで, も,
+         や, やら, よ, より, わ, を
+    }
+    auxVerbs {
+        です, でしょ, でし, ます, ませ, まし
```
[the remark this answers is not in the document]
allensu05: Hi Markus
There is a credential issue.
I added you in a mail with the title "ICU code review"
Could you help me check that?
Thanks

icu4c/source/common/brkiter.cpp
```diff
@@ -430,16 +431,24 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
...
-            int32_t kLen = loc.getKeywordValue("lb", lbKeyValue, kKeyValueLenMax, kvStatus);
-            if (U_SUCCESS(kvStatus) && kLen > 0 && (uprv_strcmp(lbKeyValue,"strict")==0 || uprv_strcmp(lbKeyValue,"normal")==0 || uprv_strcmp(lbKeyValue,"loose")==0)) {
+            CharString value;
+            CharStringByteSink valueSink(&value);
+            loc.getKeywordValue("lb", valueSink, kvStatus);
+            if (U_SUCCESS(kvStatus) && (value == "strict" || value == "normal" || value == "loose")) {
                 uprv_strcat(lbType, "_");
-                uprv_strcat(lbType, lbKeyValue);
+                uprv_strcat(lbType, value.data());
+            } else {
```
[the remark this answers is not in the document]
markusicu: I looked yesterday and wasn't sure if they are orthogonal.

I found that I asked this in [August](https://docs.google.com/document/d/1il9KJH3MeSkghBZRD5XriGXk9gfxN3yrVXmRiiUR1q4/edit#heading=h.qqfnsikbpkct) and left it as a question in the to-do's: “Decide whether this is orthogonal to CSS line break strict/loose/... (This would be more complicated than if it could be just another value in the same namespace.)”

We had another meeting in November, and I think we discussed this again. Can someone please point me to the meeting minutes for that?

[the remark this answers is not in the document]
markusicu: So the [minutes for our 2021-nov-17 meeting](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#heading=h.6194od71cpe2) are in the ICU agenda & minutes doc. (Eventually they will move into an archive copy.)

I see that we decided that the new option fits into the -lw key's namespace, but apparently we didn't discuss -lb vs. -lw.

I am ok with not blocking this PR on this question, but we need to define the relationship between the -lb and -lw keys, for both the CLDR spec and the ICU code.

icu4c/source/data/brkitr/ja.txt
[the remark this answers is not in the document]
FrankYFTang: Please file another icu ticket for that. That issue could be solved later after merging this PR and do not need to be there approaching ICU 71.1 release. We should solve that before mid Feb though.

icu4c/source/common/dictbe.cpp
allensu05: Use U_SUCCESS

icu4c/source/common/dictbe.cpp
allensu05: Resolved. Change to use code unit

FrankYFTang (approved):

markusicu (approved):
lgtm tnx!

markusicu:
FYI: I will look at the Java code soon, probably tomorrow: PR #1955

PR #1937, opened 2021-11-18
Export case props data for ICU4X

iainireland:
Here's an initial draft of exporting the data needed for case mapping as TOML. I had to refactor icuexportprops.cpp a bit to support a second mode, and I reworked ucase_props_data.h slightly to make it possible to extract the exceptions/unfold data. 
(We'll want to either rename `INCLUDED_FROM_UCASE_CPP` to something else, or remove it entirely, but I wasn't sure which would be preferable.)

##### Checklist


iainireland:
Oops, had the wrong author email for these commits. Closing and opening a new PR.

PR #1939, opened 2021-11-22
ICU-21844 Fix variable shadowing

ngzhian:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21844

### DIFF
--- a/icu4c/source/common/unicode/localematcher.h
+++ b/icu4c/source/common/unicode/localematcher.h
@@ -461,13 +461,13 @@ public:
          * Option for whether to include or ignore one-way (fallback) match data.
          * By default, they are included.
          *
-         * @param direction the match direction to set.
+         * @param matchDirection the match direction to set.
          * @return this Builder object
          * @stable ICU 67
          */
-        Builder &setDirection(ULocMatchDirection direction) {
+        Builder &setDirection(ULocMatchDirection matchDirection) {
             if (U_SUCCESS(errorCode_)) {
-                direction_ = direction;
+                direction_ = matchDirection;
             }
             return *this;
         }


icu4c/source/common/unicode/localematcher.h
```diff
@@ -465,9 +465,9 @@ class U_COMMON_API LocaleMatcher : public UMemory {
          * @return this Builder object
          * @stable ICU 67
          */
-        Builder &setDirection(ULocMatchDirection direction) {
+        Builder &setDirection(ULocMatchDirection dir) {
```
markusicu, line 468: 1. You are renaming the parameter name here but not in the `@param` line above.
2. I don't see what's being shadowed here. The name of the class member ends with an underscore, making it different from the function parameter name. Why do we need to change anything here?

ngzhian: It shadows the `direction` member of the outer class, not in this builder class: https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/localematcher.h#L688

markusicu: I don't know why the outer class should matter, since the inner class has no access to that...

Please change the parameter name both in the `@param` and in the function signature to `matchDirection`.

ngzhian: Yea, good question, maybe because there's a `friend class LocaleMatcher` here? I tried this locally, removing "friend" also emits the warning. So maybe it's a bug in compiler diagnostics? https://godbolt.org/z/Wc7h9q7Ev I'll file an llvm bug

ngzhian: Also renamed to `matchDirection`

ngzhian: Oh, https://godbolt.org/z/ETc7jaE6e so because of the friend class, the `Builder` has visibility of `direction`, but it can't actually use it without a compile error, as `direction` will need to be static.

markusicu: I see; the outer class is a friend of the inner class so it has visibility... but it would have to qualify access like `innerObject.direction` which makes it unambiguous. I don't see an actual shadowing problem. Still seems like a bug in the diagnostic.

Thanks for your changes. Please squash into one commit.

ngzhian: Squashed, thanks for the review!

markusicu (approved):
lgtm

icu4c/source/common/unicode/localematcher.h
markusicu: Thanks for the fix!
(I will wait for the build bots before merging.)

ngzhian:
Btw I have no permissions to merge this, so please help me to do so, thanks!

PR #1941, opened 2021-11-24
ICU-21847 Move UnicodeSet to stack in constructor

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21847

### DIFF
--- a/icu4c/source/common/dictbe.cpp
+++ b/icu4c/source/common/dictbe.cpp
@@ -199,13 +199,13 @@ ThaiBreakEngine::ThaiBreakEngine(DictionaryMatcher *adoptDictionary, UErrorCode 
 {
     UTRACE_ENTRY(UTRACE_UBRK_CREATE_BREAK_ENGINE);
     UTRACE_DATA1(UTRACE_INFO, "dictbe=%s", "Thai");
-    fThaiWordSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Thai:]&[:LineBreak=SA:]]"), status);
+    UnicodeSet thaiWordSet(UnicodeString(u"[[:Thai:]&[:LineBreak=SA:]]"), status);
     if (U_SUCCESS(status)) {
-        setCharacters(fThaiWordSet);
+        setCharacters(thaiWordSet);
     }
-    fMarkSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Thai:]&[:LineBreak=SA:]&[:M:]]"), status);
+    fMarkSet.applyPattern(UnicodeString(u"[[:Thai:]&[:LineBreak=SA:]&[:M:]]"), status);
     fMarkSet.add(0x0020);
-    fEndWordSet = fThaiWordSet;
+    fEndWordSet = thaiWordSet;
     fEndWordSet.remove(0x0E31);             // MAI HAN-AKAT
     fEndWordSet.remove(0x0E40, 0x0E44);     // SARA E through SARA AI MAIMALAI
     fBeginWordSet.add(0x0E01, 0x0E2E);      // KO KAI through HO NOKHUK
@@ -441,13 +441,13 @@ LaoBreakEngine::LaoBreakEngine(DictionaryMatcher *adoptDictionary, UErrorCode &s
 {
     UTRACE_ENTRY(UTRACE_UBRK_CREATE_BREAK_ENGINE);
     UTRACE_DATA1(UTRACE_INFO, "dictbe=%s", "Laoo");
-    fLaoWordSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Laoo:]&[:LineBreak=SA:]]"), status);
+    UnicodeSet laoWordSet(UnicodeString(u"[[:Laoo:]&[:LineBreak=SA:]]"), status);
     if (U_SUCCESS(status)) {
-        setCharacters(fLaoWordSet);
+        setCharacters(laoWordSet);
     }
-    fMarkSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Laoo:]&[:LineBreak=SA:]&[:M:]]"), status);
+    fMarkSet.applyPattern(UnicodeString(u"[[:Laoo:]&[:LineBreak=SA:]&[:M:]]"), status);
     fMarkSet.add(0x0020);
-    fEndWordSet = fLaoWordSet;
+    fEndWordSet = laoWordSet;
     fEndWordSet.remove(0x0EC0, 0x0EC4);     // prefix vowels
     fBeginWordSet.add(0x0E81, 0x0EAE);      // basic consonants (including holes for corresponding Thai characters)
     fBeginWordSet.add(0x0EDC, 0x0EDD);      // digraph consonants (no Thai equivalent)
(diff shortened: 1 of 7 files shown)


icu4c/source/common/dictbe.cpp
```diff
@@ -199,13 +199,13 @@ ThaiBreakEngine::ThaiBreakEngine(DictionaryMatcher *adoptDictionary, UErrorCode
 {
     UTRACE_ENTRY(UTRACE_UBRK_CREATE_BREAK_ENGINE);
     UTRACE_DATA1(UTRACE_INFO, "dictbe=%s", "Thai");
-    fThaiWordSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Thai:]&[:LineBreak=SA:]]"), status);
+    UnicodeSet thaiWordSet(UNICODE_STRING_SIMPLE("[[:Thai:]&[:LineBreak=SA:]]"), status);
```
aheninger, line 202: (Optional) We could get rid of using the archaic UNICODE_STRING_SIMPLE macro now that we can use Unicode string literals. Use this instead: 
```UnicodeString(u"[[:Thai:]&[:LineBreak=SA:]]")```

Even more efficient, but more awkward, would be to build the set using properties directly, avoiding the need to parse the pattern string. Probably not worth the trouble and loss of easy readability.

Same comments apply to all the other changes.

Comment (markusicu) [reply icu4c/source/common/dictbe.cpp:202]: Yes, please change these string literals to using `u"..."` directly.
But keep the pattern strings. The parser is not *that* expensive, compared with computing the underlying sets and the intersection.

**Comment by aheninger** [comment]
A similar change could be done in ICU4J, so that UnicodeSets in question aren't forced to persist, but become collectible when they go out of scope.

**Comment by markusicu** [comment]
> A similar change could be done in ICU4J, so that UnicodeSets in question aren't forced to persist, but become collectible when they go out of scope.

Oh, yes, please keep C++ & Java in sync...

**Comment by FrankYFTang** [comment]
Change Java too. Java used to put those UnicodeSet as private static but that is not really needed since for most of them (HangulWordSet is an exception, but small) all these classes were used by the caller as a singleton in the system as only one instance (except CJ ) in the system by the calling function

File: icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/BurmeseBreakEngine.java [hunk 2]
```diff
@@ -31,23 +31,22 @@
     private static final byte BURMESE_MIN_WORD = 2;
 
     private DictionaryMatcher fDictionary;
-    private static UnicodeSet fBurmeseWordSet;
-    private static UnicodeSet fEndWordSet;
-    private static UnicodeSet fBeginWordSet;
-    private static UnicodeSet fMarkSet;
+    private UnicodeSet fEndWordSet;
```
markusicu, line 34: Note: The trade-off here is to permanently hold onto less memory, but to increase the time it takes to construct a break engine after the first time. I guess it's ok either way until someone complains about memory or about performance.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/BurmeseBreakEngine.java
```diff
@@ -31,23 +31,22 @@
...
+    public BurmeseBreakEngine() throws IOException {
         // Initialize UnicodeSets
-        fBurmeseWordSet = new UnicodeSet();
+        UnicodeSet burmeseWordSet = new UnicodeSet();
         fMarkSet = new UnicodeSet();
         fBeginWordSet = new UnicodeSet();
 
-        fBurmeseWordSet.applyPattern("[[:Mymr:]&[:LineBreak=SA:]]");
-        fBurmeseWordSet.compact();
+        burmeseWordSet.applyPattern("[[:Mymr:]&[:LineBreak=SA:]]");
```
markusicu, line 44: Optional: You can replace `new UnicodeSet()` + `applyPattern(pattern)` with `new UnicodeSet(pattern)`.

markusicu (approved):
lgtm but with notes

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/BurmeseBreakEngine.java
FrankYFTang: as I mention, for all these LanguageBreakEngine, since they are not part of the ICU API, they are only used internally in Java.
The only place construct them inside Java is inside
main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java 
and tracked by     private static final List<LanguageBreakEngine> gAllBreakEngines;
and all LanguageBreakEngine (except CjkBreakEngine) will only be created once and remember in gAllBreakEngines and will not call the constructor more than once. CjkBreakEngine constructor may be called twice- once for UScript.HAN UScript.KATAKANA or  UScript.HIRAGANA (shared by one) and another time for UScript.HANGUL . Therfore the so-called "construct a break engine after the first time" will never happen except the second time for CjkBreakEngine .

icu4c/source/common/dictbe.cpp
```diff
@@ -199,13 +199,13 @@ ThaiBreakEngine::ThaiBreakEngine(DictionaryMatcher *adoptDictionary, UErrorCode
 {
     UTRACE_ENTRY(UTRACE_UBRK_CREATE_BREAK_ENGINE);
     UTRACE_DATA1(UTRACE_INFO, "dictbe=%s", "Thai");
-    fThaiWordSet.applyPattern(UNICODE_STRING_SIMPLE("[[:Thai:]&[:LineBreak=SA:]]"), status);
+    UnicodeSet thaiWordSet(UnicodeString(u"[[:Thai:]&[:LineBreak=SA:]]"), status);
```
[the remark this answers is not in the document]
FrankYFTang: How?
I got the following if I don't.
```
make[1]: Entering directory '/usr/local/google/home/ftang/icu/icu/icu4c/source/common'
   (deps)	 dictbe.cpp
   g++	 ...  dictbe.cpp
dictbe.cpp: In constructor ‘icu_70::ThaiBreakEngine::ThaiBreakEngine(icu_70::DictionaryMatcher*, UErrorCode&)’:
dictbe.cpp:202:28: error: invalid conversion from ‘const char16_t*’ to ‘UChar32’ {aka ‘int’} [-fpermissive]
  202 |     UnicodeSet thaiWordSet(u"[[:Thai:]&[:LineBreak=SA:]]", status);
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                            |
      |                            const char16_t*
In file included from dictbe.h:14,
                 from dictbe.cpp:17:
unicode/uniset.h:392:24: note:   initializing argument 1 of ‘icu_70::UnicodeSet::UnicodeSet(UChar32, UChar32)’
  392 |     UnicodeSet(UChar32 start, UChar32 end);
      |                ~~~~~~~~^~~~~
```

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/BurmeseBreakEngine.java
markusicu: Sounds good. Thanks for the detailed explanation, all makes sense. I just didn't have the facts right.

icu4c/source/common/dictbe.cpp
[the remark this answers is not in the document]
markusicu: Never mind then. I guess we have enough constructor overloads to make this ambiguous.

markusicu (dismissed):

FrankYFTang:
PTAL

FrankYFTang:
BTW, should I replace "[\\uac00-\\ud7a3]" with "[:Hangul:]"? How about Jamo?

markusicu:
> BTW, should I replace "[\uac00-\ud7a3]" with "[:Hangul:]"?

No, Hangul is broader, it includes the Jamo characters and compatibility variants:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AHangul%3A%5D&abb=on&g=&i=

You could use Hangul_Syllable_Type, such as `[[:hst=LV:][:hst=LVT:]]` but the syllables are in one fixed range, so this kind of pattern is unnecessarily slow to compute.

> How about Jamo?

Depends on what you mean with that. `[[:hst=L:][:hst=V:][:hst=T:]]` has many characters outside of the U+11xx block.

See https://www.unicode.org/Public/UCD/latest/ucd/PropertyValueAliases.txt
```
# Hangul_Syllable_Type (hst)

hst; L                                ; Leading_Jamo
hst; LV                               ; LV_Syllable
hst; LVT                              ; LVT_Syllable
hst; NA                               ; Not_Applicable
hst; T                                ; Trailing_Jamo
hst; V                                ; Vowel_Jamo
```

markusicu:
> PTAL

Nice simplifications in the second commit!
Please squash when ready.

markusicu (approved):
lgtm

PR #1955, opened 2021-12-27
ICU-21699 Phrase based breaking(Java)

allensu05:
Implementation doc:
https://docs.google.com/document/d/1Lg50cEKCLowTL3JwCXlOXOYpDkYJAeKy0I8E7qk7mNs/edit#heading=h.6wpzwdae2hs1

Unit test step:

cd icu4j
ant checkTest -Dtestclass='com.ibm.icu.dev.test.rbbi.RBBITestExtended'

##### Checklist


icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -14,28 +14,64 @@
...
+    private static final int U_GC_CP_MASK = (1 << UCharacterCategory.DASH_PUNCTUATION)
+            | (1 << UCharacterCategory.END_PUNCTUATION)
+            | (1 << UCharacterCategory.CONNECTOR_PUNCTUATION)
+            | (1 << UCharacterCategory.OTHER_PUNCTUATION)
+            | (1 << UCharacterCategory.FINAL_PUNCTUATION);
 
     public CjkBreakEngine(boolean korean) throws IOException {
         fHangulWordSet = new UnicodeSet("[\\uac00-\\ud7a3]");
         fHangulWordSet.freeze();
+        fHiraganaWordSet = new UnicodeSet("[:Hiragana:]");
```
FrankYFTang, line 47: so, looks like both fHiraganaWordSet and fParticleSet are always used in exactly the same place to decide a break point should be skip or not. How about instead of checking a HashSet first then check a UnicodeSet, we only check one HashSet. so
1. Rename the fParticleSet to fSkipSet 
2. Remove fHiraganaWordSet from the member data
3. In the constructor, create a stack hiraganaWordSet  as how you currently create fHiraganaWordSet but then add the content of that hiraganaWordSet into fSkipSet by using https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/text/UnicodeSet.html#iterator-- and you won't need hiraganaWordSet after the constructor.
4. skip the checking of fHiraganaWordSet since the check will be done inside fSkipSet.contains()

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -14,28 +14,64 @@
...
+import com.ibm.icu.util.UResourceBundleIterator;
 
 public class CjkBreakEngine extends DictionaryBreakEngine {
     private UnicodeSet fHangulWordSet;
+    private UnicodeSet fHiraganaWordSet;
+    private UnicodeSet fNumberWordSet;
     private DictionaryMatcher fDictionary = null;
+    private HashSet<String> fParticleSet;
+    //open punctuation mask
+    private static final int U_GC_OP_MASK = (1 << UCharacterCategory.INITIAL_PUNCTUATION)
```
FrankYFTang, line 35: Should we just use UnicodeSet for checking puncation?  @markusicu  any opinion. I just don't see the reason we use UnicodeSet to check some char then use "UCharacter.getType()" to check other chars. Is there any benefit of using "UCharacter.getType()" instead of UnicodeSet?

FrankYFTang (changes requested):

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -84,6 +84,31 @@ public static RuleBasedBreakIterator getInstanceFromCompiledRules(InputStream is
...
+     *
+     * @param bytes a buffer supplying the compiled binary rules.
+     * @param phraseBreaking a flag indicating if phrase breaking is required.
+     * @throws IOException if there is an error while reading the rules from the buffer.
+     * @see    #compileRules(String, OutputStream)
+     * @internal
+     */
+    static RuleBasedBreakIterator getInstanceFromCompiledRules(ByteBuffer bytes,
+            boolean phraseBreaking) throws IOException {
+        RuleBasedBreakIterator  This = getInstanceFromCompiledRules(bytes);
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -84,6 +84,31 @@ public static RuleBasedBreakIterator getInstanceFromCompiledRules(InputStream is
...
+     * Binary break iterator rules are not guaranteed to be compatible between
+     * different versions of ICU.
+     *
+     * @param bytes a buffer supplying the compiled binary rules.
+     * @param phraseBreaking a flag indicating if phrase breaking is required.
+     * @throws IOException if there is an error while reading the rules from the buffer.
+     * @see    #compileRules(String, OutputStream)
+     * @internal
+     */
+    static RuleBasedBreakIterator getInstanceFromCompiledRules(ByteBuffer bytes,
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -197,11 +241,24 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
         } else {
-            for (int i = numCodePts; i > 0; i = prev[i]) {
+            t_boundary[numBreaks] = numCodePts;
+            numBreaks++;
+            int prevIdx = numCodePts;
+            StringBuffer sb = new StringBuffer();
+            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
+                int length = prevIdx -  i;
+                prevIdx = i;
+                if (isPhraseBreaking) {
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +268,34 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
         int correctedNumBreaks = 0;
         for (int i = numBreaks - 1; i >= 0; i--) {
             int pos = charPositions[t_boundary[i]] + startPos;
-            if (!(foundBreaks.contains(pos) || pos == startPos)) {
+            if (!foundBreaks.contains(pos)
+                    && (pos != startPos
+                    || pos ==0
```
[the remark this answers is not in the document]
allensu05: I use AndroidStudio to reformat the code.
&& and || do start at the same position.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -197,11 +241,24 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
             t_boundary[numBreaks] = numCodePts;
             numBreaks++;
         } else {
-            for (int i = numCodePts; i > 0; i = prev[i]) {
+            t_boundary[numBreaks] = numCodePts;
+            numBreaks++;
+            int prevIdx = numCodePts;
+            StringBuffer sb = new StringBuffer();
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +268,34 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
             correctedNumBreaks--;
         }
         if (!foundBreaks.isEmpty())
             inText.setIndex(foundBreaks.peek());
         return correctedNumBreaks;
     }
+
+    private String getPatternFromText(CharacterIterator text, StringBuffer sb, int start,
+            int length) {
+        for(int j = 0 ; j < length; j++) {
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +268,34 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
+            sb.append(text.setIndex(start + j));
+        }
+        String result = sb.toString();
+        sb.setLength(0);
```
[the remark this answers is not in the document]
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +268,34 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
+                    && (pos != startPos
+                    || pos ==0
+                    || (pos == startPos  && isClosePunct(inText.setIndex(pos-1))))) {
                 foundBreaks.push(charPositions[t_boundary[i]] + startPos);
                 correctedNumBreaks++;
             }
         }
 
-        if (!foundBreaks.isEmpty() && foundBreaks.peek() == endPos) {
+        if (!foundBreaks.isEmpty()
```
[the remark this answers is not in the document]
allensu05: For word break, even if the logic here pops the ending breakpoint, it will be added back in 
https://github.com/unicode-org/icu/blob/main/icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java#L1231

To keep the same logic for word breaking, I add isPhraseBreaking but actually no impact for word break.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -14,25 +14,73 @@
...
+            | (1 << UCharacterCategory.CONNECTOR_PUNCTUATION)
+            | (1 << UCharacterCategory.OTHER_PUNCTUATION)
+            | (1 << UCharacterCategory.FINAL_PUNCTUATION);
 
     public CjkBreakEngine(boolean korean) throws IOException {
         fHangulWordSet = new UnicodeSet("[\\uac00-\\ud7a3]");
         fHangulWordSet.freeze();
+        fNumberWordSet = new UnicodeSet("[:Nd:]");
+        fNumberWordSet.freeze();
+        fSkipSet = new HashSet<>();
```
FrankYFTang, line 49: fSkipSet = new HashSet<String>();

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -196,6 +252,23 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
             t_boundary[numBreaks] = numCodePts;
             numBreaks++;
+        } else if (isPhraseBreaking) {
+            t_boundary[numBreaks] = numCodePts;
+            numBreaks++;
+            int prevIdx = numCodePts;
+            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
+                int length = prevIdx - i;
+                prevIdx = i;
+                if (isPhraseBreaking) {
```
FrankYFTang, line 262: you already have isPhraseBreaking in the else if so this if (isPhraseBreaking) is always true.
and once we remove that check, the remaining logic is simpler so how about just flip the if by adding ! in front of contains check and place the two line into the if block as below
```
            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
                int length = prevIdx - i;
                prevIdx = i;
                // Skip the breakpoint if it is after the particle or Hiragana.
                if (!fSkipSet.contains(getPatternFromText(text, s, i, length))) {
                     t_boundary[numBreaks] = i;
                     numBreaks++;
                }
            }
```
is it "after" ? or before? I am not sure. Please correct if I made mistake

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +284,40 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
             }
         }
 
         if (!foundBreaks.isEmpty() && foundBreaks.peek() == endPos) {
-            foundBreaks.pop();
-            correctedNumBreaks--;
+            // In phrase breaking, there has to be a breakpoint between Cj character and
+            // number/open punctuation e.g. １枚▁４０００円, 文字▁「そうだ...」
+            if (isPhraseBreaking) {
+                if (!isOpenPunct(inText.setIndex(endPos)) && !fNumberWordSet.contains(
```
FrankYFTang, line 300: wrap the line before && to make it easier to read

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
```diff
@@ -135,11 +135,18 @@ private static BreakIterator createBreakInstance(ULocale locale, int kind) {
             if ( lbKeyValue != null && (lbKeyValue.equals("strict") || lbKeyValue.equals("normal") || lbKeyValue.equals("loose")) ) {
                 typeKeyExt = "_" + lbKeyValue;
             }
+            if(typeKeyExt == null) {
+                String lwKeyValue = locale.getKeywordValue("lw");
+                if("phrase".equals(lwKeyValue)) {
```
FrankYFTang, line 140: I think it is better 
```
if(lwKeyValue.equals("phrase")) {
```

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: You are right. The check here should be removed.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
allensu05: I think 
"phrase".equals(lwKeyValue) is shorter than
lwKeyValue != null && lwKeyValue.equals("phrase")

FrankYFTang:
your build fail on lstm-icu4j-build-and-test
The detail of the build is in https://github.com/unicode-org/icu/blob/b9311e2f547251807c3982e6cb003feec3ba5c83/.github/workflows/icu_ci.yml#L128

Notice, that build will rebuild the dat file and jar file under a different configuration during the build process so if your ICU4J tree is not based on your changes in the ICU4C branch the rebuilted jar file won't have the new thing you added to the ICU4C even you include the rebuilt jar file in the PR because during that build process the process will rebuilt the jar file from scratch and ignore your jar file (only in that particular testing build)

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -211,18 +268,45 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
         int correctedNumBreaks = 0;
         for (int i = numBreaks - 1; i >= 0; i--) {
             int pos = charPositions[t_boundary[i]] + startPos;
-            if (!(foundBreaks.contains(pos) || pos == startPos)) {
+            // In phrase breaking, there has to be a breakpoint between Cj character and close
+            // punctuation.
+            // E.g.［携帯電話］正しい選択 -> ［携帯▁電話］▁正しい▁選択 -> breakpoint between ］ and 正
+            if (!foundBreaks.contains(pos)
+                    && (pos != startPos
+                    || (isPhraseBreaking && pos == startPos && pos > 0
```
FrankYFTang, line 276: I believe you do NOT need to have "&& pos == startPos"  because that is ALWAYS TRUE after "pos != startPos  || " You only need 
if (!foundBreaks.contains(pos)
                    && (pos != startPos
                    || (isPhraseBreaking && pos > 0
                           && fClosePunctuationSet.contains(inText.setIndex(pos - 1))))) {

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -14,25 +14,68 @@
...
+import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.text.Normalizer;
 import com.ibm.icu.text.UnicodeSet;
+import com.ibm.icu.text.UnicodeSetIterator;
+import com.ibm.icu.util.UResourceBundle;
+import com.ibm.icu.util.UResourceBundleIterator;
 
 public class CjkBreakEngine extends DictionaryBreakEngine {
     private UnicodeSet fHangulWordSet;
+    private UnicodeSet fNumberWordSet;
```
FrankYFTang, line 29: so fNumberWordSet and fOpenPunctuationdSet are use exactly the same place together. Could we just use one UnicodeSet for that? 
fNumberOrOpenPunctuationdSet = new UnicodeSet("[[:Nd:][:Pi:][:Ps:]]");

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
```diff
@@ -131,15 +131,21 @@ private static BreakIterator createBreakInstance(ULocale locale, int kind) {
...
         if (kind == BreakIterator.KIND_LINE) {
-            String lbKeyValue = locale.getKeywordValue("lb");
-            if ( lbKeyValue != null && (lbKeyValue.equals("strict") || lbKeyValue.equals("normal") || lbKeyValue.equals("loose")) ) {
-                typeKeyExt = "_" + lbKeyValue;
+            String keyValue = locale.getKeywordValue("lb");
+            if ( keyValue != null && (keyValue.equals("strict") || keyValue.equals("normal") || keyValue.equals("loose")) ) {
+                typeKeyExt = "_" + keyValue;
+            } else {
+                keyValue = locale.getKeywordValue("lw");
+                if ("phrase".equals(keyValue)) {
```
FrankYFTang, line 139: Please use
```
                if (keyValue.equals("phrase")) {
```
in Java

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
```diff
@@ -151,7 +157,7 @@ private static BreakIterator createBreakInstance(ULocale locale, int kind) {
         // Create a normal RuleBasedBreakIterator.
         //
         try {
-            iter = RuleBasedBreakIterator.getInstanceFromCompiledRules(bytes);
+            iter = RuleBasedBreakIterator.getInstanceFromCompiledRules(bytes, brkfname.startsWith("line_phrase_cj"));
```
FrankYFTang, line 160: If in the previous try {} catch (Exception e) there are missing resources, then brkfname will be null, right? In that case brkfname will be null here.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
allensu05: Resolved

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
allensu05: Resolved

FrankYFTang (approved):

FrankYFTang (dismissed):

FrankYFTang (approved):
LGTM

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -212,19 +271,46 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
             inText.setIndex(foundBreaks.peek());
         return correctedNumBreaks;
     }
+
+    private String getPatternFromText(CharacterIterator text, StringBuffer sb, int start,
+            int length) {
+        sb.setLength(0);
+        for (int j = 0; j < length; j++) {
+            sb.append(text.setIndex(start + j));
+        }
```
markusicu, line 313: Please don't setIndex() for each char -- text is an iterator (although with unusual pre-increment semantics).
```suggestion
        if (length > 0) {
            sb.append(text.setIndex(start));
            for (int j = 1; j < length; ++j) {
                sb.append(text.next());
            }
        }
```

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIteratorFactory.java
```diff
@@ -131,15 +131,23 @@ private static BreakIterator createBreakInstance(ULocale locale, int kind) {
...
-                typeKeyExt = "_" + lbKeyValue;
+            String keyValue = locale.getKeywordValue("lb");
+            if ( keyValue != null && (keyValue.equals("strict") || keyValue.equals("normal") || keyValue.equals("loose")) ) {
+                typeKeyExt = "_" + keyValue;
+            }
+            String language = locale.getLanguage();
+            if (language != null && language.equals("ja")) {
+                keyValue = locale.getKeywordValue("lw");
+                if (keyValue != null && keyValue.equals("phrase")) {
+                    typeKeyExt = "_" + keyValue;
```
markusicu, line 142: It looks like the lw value *replaces* the lb value. Does this work? Shouldn't it append?
```suggestion
                    typeKeyExt += "_" + keyValue;
```

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -84,6 +84,31 @@ public static RuleBasedBreakIterator getInstanceFromCompiledRules(InputStream is
         return This;
     }
 
+    /**
+     * This constructor doesn't have access modifier, only accessible in the same package.
```
markusicu, line 88:
```suggestion
     * This factory method doesn't have an access modifier; it is only accessible in the same package.
```

markusicu (changes requested):

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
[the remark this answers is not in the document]
markusicu: It seems like it should nest because the second line continues a (parenthesis), but I don't care enough...

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
markusicu: Using a UnicodeSet adds to init time and heap memory, but is probably a little faster (if the set is "frozen", as I see now in the code) than getType(). Ok either way for General_Category values and groupings.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -212,19 +271,48 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
     }
+
+    private String getPatternFromText(CharacterIterator text, StringBuffer sb, int start,
+            int length) {
+        sb.setLength(0);
+        text.setIndex(start);
+        for (int j = 0; j < length; j++) {
+            sb.appendCodePoint(current32(text));
+            next32(text);
+        }
```
markusicu, line 315: Optional(!) optimization:
```suggestion
        text.setIndex(start);
        sb.appendCodePoint(current32(text));
        for (int j = 1; j < length; j++) {
            sb.appendCodePoint(next32(text));
        }
```

markusicu (approved):
lgtm tnx

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
allensu05: I am not sure if length is always > 0; so I keep it the loop.

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -196,6 +236,25 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
+            int prevIdx = numCodePts;
+            int codeUnitIdx = 0, length = 0;
+            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
+                codeUnitIdx = prenormstr.offsetByCodePoints(0, i);
+                length = prevIdx - i;
+                prevIdx = i;
+                String pattern = getPatternFromText(text, s, codeUnitIdx, length);
+                // Keep the breakpoint if the pattern is not in the fSkipSet and continuous Katakana
+                // characters don't occur.
+                text.setIndex(codeUnitIdx - 1);
```
markusicu, line 251: I am going to approve again, but could you take another look at this line and its `- 1`?
It looks like that is intended to "go one *code point* before codeUnitIdx" (see the following use of current32() and next32()) which could be one or two code units.

Also, is it possible here that codeUnitIdx == 0?

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -196,6 +236,25 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
...
+        } else if (isPhraseBreaking) {
+            t_boundary[numBreaks] = numCodePts;
+            numBreaks++;
+            int prevIdx = numCodePts;
+            int codeUnitIdx = 0, length = 0;
+            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
+                codeUnitIdx = prenormstr.offsetByCodePoints(0, i);
+                length = prevIdx - i;
+                prevIdx = i;
+                String pattern = getPatternFromText(text, s, codeUnitIdx, length);
```
markusicu, line 248: FYI  FYI
Thinking about the code points vs. code units...
- The pre-increment next32() function is not super efficient, because it first has to do most of what current32() does just to skip the current code point.
- Therefore, your getPatternText() from a couple of days ago (taking a code unit length)  *might* be faster.
- It should also help to compute the codeUnitIdx of the first `i` before the loop, and then offsetByCodePoints() backwards from there in small increments of `i - prevIdx` (negative=backwards), tracking two variables like codeUnitIdx and prevCodeUnitIdx.
- More interesting: If you knew that your fSkipSet never contained any strings with supplementary characters (e.g., offsetByCodePoints() moves by more code units than the given code point delta), then maybe you could just detect if there are any supplementary code points in the "length" span, and if so, skip the get-pattern and the matching (fSkipSet.contains() would be false anyway).

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
```diff
@@ -196,6 +236,25 @@ public int divideUpDictionaryRange(CharacterIterator inText, int startPos, int e
         if (bestSnlp[numCodePts] == kint32max) {
             t_boundary[numBreaks] = numCodePts;
             numBreaks++;
+        } else if (isPhraseBreaking) {
+            t_boundary[numBreaks] = numCodePts;
+            numBreaks++;
+            int prevIdx = numCodePts;
+            int codeUnitIdx = 0, length = 0;
+            for (int i = prev[numCodePts]; i > 0; i = prev[i]) {
+                codeUnitIdx = prenormstr.offsetByCodePoints(0, i);
```
markusicu, line 245: FYI
Please consider a follow-up task of making sure that you have good test coverage with strings with supplementary characters, for example [supplementary Katakana](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AKatakana%3A%5D-%5B%5Cu0000-%5CuFFFF%5D&g=&i=) characters etc. that would make a difference in this code.

markusicu (approved):
lgtm but one question

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
FrankYFTang: I asked the same question in the C++ review and Allen said line 244 and 245 will ensure codeUnitIdx > 0 always

FrankYFTang (approved):

icu4j/main/classes/core/src/com/ibm/icu/impl/breakiter/CjkBreakEngine.java
markusicu: Ok. What about one code point vs. one code unit?

allensu05: I made some modification here. Test case will be added soon.
https://github.com/unicode-org/icu/pull/1965

PR #1953, opened 2021-12-22
ICU-21878 Sync icu4j's CjkBreakEngine to icu4c's

allensu05:
Remove the first and the last breakpoints as dealing with word breaking.

- No impact on current implementation
- Make the implementation synchronous between icu4j and icu4c
- It will simplify the implementation of Japanese phrase breaking (ICU-21699)

Test:
icu4c: rbbi/RBBITest/TestExtended
icu4j: com.ibm.icu.dev.test.rbbi.RBBITestExtended

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21878
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #1956, opened 2022-01-02
ICU-21884 Revert PR-1953

allensu05:
PR-1953 caused fuzz testing failure.
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43173

Temporarily revert the change.
https://github.com/unicode-org/icu/pull/1953

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21884


aheninger (approved):
I confirmed that this PR accurately rolls back the change.
Now comes the hard part, which is figuring out why the change caused a fuzzing test failure.

FrankYFTang:
No need to revert because it is just slowness of ASAN build with the test case but not infinity loop. I tried the non ASAN build with the test case and it does not hang.

PR #1967, opened 2022-01-25
ICU-21893 Add some processing details to ubrk/brkiter documentation

litherum:
- [✅] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21893
- [✅] Required: The PR title must be prefixed with a JIRA Issue number. 
- [✅] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [✅] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [✅] Issue accepted (done by Technical Committee after discussion)
- [✅] Tests included, if applicable
- [✅] API docs and/or User Guide docs changed or added, if applicable

### DIFF
--- a/icu4c/source/common/unicode/brkiter.h
+++ b/icu4c/source/common/unicode/brkiter.h
@@ -62,7 +62,9 @@ U_NAMESPACE_BEGIN
  * The BreakIterator class implements methods for finding the location
  * of boundaries in text. BreakIterator is an abstract base class.
  * Instances of BreakIterator maintain a current position and scan over
- * text returning the index of characters where boundaries occur.
+ * text, returning the index of characters immediately after the locations
+ * where boundaries occur (because boundaries occur between adjacent
+ * characters).
  * <p>
  * Line boundary analysis determines where a text string can be broken
  * when line-wrapping. The mechanism correctly handles punctuation and
@@ -76,7 +78,7 @@ U_NAMESPACE_BEGIN
  * well as within text editing applications that allow the user to
  * select words with a double click. Word selection provides correct
  * interpretation of punctuation marks within and following
- * words. Characters that are not part of a word, such as symbols or
+ * words. Characters that are not part of a word, such as symbols, spaces or
  * punctuation marks, have word-breaks on both sides.
  * <p>
  * Character boundary analysis allows users to interact with
@@ -92,6 +94,34 @@ U_NAMESPACE_BEGIN
  * Unicode Standard Annex #14, Line Breaking Properties.  These
  * are available at http://www.unicode.org/reports/tr14/ and
  * http://www.unicode.org/reports/tr29/.
+ * <p>
+ * A boundary logically occurs between two adjacent code points. The boundary
+ * analysis functions in this file return the index of the code unit directly
+ * after a boundary - so a value of 0 means "the beginning of the string" and a
+ * value of u_strlen(string) means "the end of the string".
+ * <p>
+ * A span of text between two adjacent boundaries is called a "segment."
+ * The boundary analysis functions in this file behave correctly regardless of
+ * whether the current position is set to a boundary location or into the
+ * middle of a segment. When the current position points to a boundary, the two
+ * segments that straddle that boundary are called the "previous segment" and
+ * the "current segment."
+ * <p>
+ * If the current position is set to the middle of a surrogate pair, the
+ * boundary analysis functions in this file will move the current position to
+ * the beginning of the surrogate pair before any other processing. If the
+ * current position then points into the middle of a segment, these routines
+ * will move to the beginning or end of that segment. Otherwise, the current
+ * position points to a segment boundary, and these routines will move to the
+ * beginning of the previous segment or the end of the current segment.
+ * <p>
+ * The boundary analysis functions in this file may scan an unbounded number of
+ * characters around the current position or even the returned position. (One
+ * example is a long string of flag emoji. In order to know whether the current
+ * position is in the middle of a single flag or between adjacent flags, the
+ * implementation counts how many regional indicator symbols occur in the
+ * string before the current position, to determine if the count is odd or
+ * even.)
  * <p>
  * In addition to the C++ API defined in this header file, a
  * plain C API with equivalent functionality is defined in the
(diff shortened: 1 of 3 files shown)


icu4c/source/common/unicode/brkiter.h
```diff
@@ -93,6 +95,34 @@ U_NAMESPACE_BEGIN
  * are available at http://www.unicode.org/reports/tr14/ and
  * http://www.unicode.org/reports/tr29/.
  * <p>
+ * A boundary logically occurs between two adjacent code points. The boundary
+ * analysis functions in this file return the index of the code unit directly
+ * after a boundary - so a value of 0 means "the beginning of the string" and a
+ * value of u_strlen(string) means "the end of the string".
+ * <p>
+ * A span of text between two adjacent boundaries is called a "segment."
+ * The boundary analysis functions in this file behave correctly regardless of
```
aheninger, line 104: A break iterator's position is always on a boundary; there is no mechanism to set it to an arbitrary non-boundary position.
The functions `preceding(position)` and `following(position)` do take any position as an argument, but always leave the iterator on a nearby boundary.

icu4c/source/common/unicode/brkiter.h
```diff
@@ -93,6 +95,34 @@ U_NAMESPACE_BEGIN
...
+ * whether the current position is set to a boundary location or into the
+ * middle of a segment. When the current position points to a boundary, the two
+ * segments that straddle that boundary are called the "previous segment" and
+ * the "current segment."
+ * <p>
+ * If the current position is set to the middle of a surrogate pair, the
+ * boundary analysis functions in this file will move the current position to
+ * the beginning of the surrogate pair before any other processing. If the
+ * current position then points into the middle of a segment, these routines
+ * will move to the beginning or end of that segment. Otherwise, the current
```
FrankYFTang, line 114: "will move to the beginning or end of that segment."
how about change to
"will move to segment boundary before or after that segment."
to make it consistent w/ the following sentance.

icu4c/source/common/unicode/brkiter.h
```diff
@@ -93,6 +95,34 @@ U_NAMESPACE_BEGIN
...
+ * position points to a segment boundary, and these routines will move to the
+ * beginning of the previous segment or the end of the current segment.
```
FrankYFTang, line 116: how about change 
"beginning of the previous segment or the end of the current segment."
to
"segment boundary before the previous segment or the segment boundary after the current segment."
?

icu4c/source/common/unicode/brkiter.h
```diff
@@ -93,6 +95,34 @@ U_NAMESPACE_BEGIN
...
+ * <p>
+ * The boundary analysis functions in this file may scan an unbounded number of
+ * characters around the current position or even the returned position. (One
+ * example is a long string of flag emoji. In order to know whether the current
```
FrankYFTang, line 120: " (One example is a long string of flag emoji. In order to know whether the current position is in the middle of a single flag or between adjacent flags, the implementation counts how many regional indicator symbols occur in the string before the current position, to determine if the count is odd or even.)"
How about
"For example, while examing a long string of flag emoji, in order to decide whether the current position is in the middle of a single flag or between adjacent flags, the implementation need to counts how many regional indicator symbols occur in the string before the current position, to determine if the count is an odd or even number."

icu4c/source/common/unicode/ubrk.h
```diff
@@ -78,16 +79,40 @@
  * http://www.unicode.org/reports/tr29/ for additional information
  * on grapheme clusters and guidelines on their use.
  * <p>
- * Title boundary analysis locates all positions,
```
FrankYFTang, line 81: Why do we remove this documentation?

icu4j/main/classes/core/src/com/ibm/icu/text/BreakIterator.java
```diff
@@ -29,10 +29,8 @@
  * example, to break a piece of text into words, sentences, or logical characters
  * according to the conventions of some language or group of languages.
  *
- * We provide five built-in types of BreakIterator:
- * <ul><li>getTitleInstance() returns a BreakIterator that locates boundaries
```
FrankYFTang, line 33: why do we remove the title case documentation?

icu4c/source/common/unicode/brkiter.h
```diff
@@ -93,6 +95,34 @@ U_NAMESPACE_BEGIN
...
+ * value of u_strlen(string) means "the end of the string".
+ * <p>
+ * A span of text between two adjacent boundaries is called a "segment."
+ * The boundary analysis functions in this file behave correctly regardless of
+ * whether the current position is set to a boundary location or into the
+ * middle of a segment. When the current position points to a boundary, the two
+ * segments that straddle that boundary are called the "previous segment" and
+ * the "current segment."
+ * <p>
+ * If the current position is set to the middle of a surrogate pair, the
```
aheninger, line 110: There's an interesting difference between ICU4C and ICU4J here. See [ICU-13361](https://unicode-org.atlassian.net/browse/ICU-13361). The issue is with `preceding(index)` when index points to a trail surrogate, and there is a boundary preceding the code point. We never reached a consensus on which behavior to choose, and I'm concerned changing either one (C or J) will break somebody. My inclination would be to keep ICU4C's behavior, and change ICU4J to match.

icu4c/source/common/unicode/brkiter.h
aheninger, line 120: Because the current position is known to be a boundary `next()` can work efficiently with flag emojis - just move two forward. Backwards iteration, or positioning with `preceding()` or `following()` do run into the behavior described, though.

Combining mark sequences also can cause arbitrarily long scanning.

icu4c/source/common/unicode/ubrk.h
litherum: It's deprecated.

FrankYFTang:
ping

icu4c/source/common/unicode/brkiter.h
aheninger: Again, an ICU break iterators's current position is fully constrained to always be on a boundary. So neither of the proposed changes to this text works.

PR #1968, opened 2022-01-27
ICU-21699 Add breakpoint between Japanese and Latin char

allensu05:
##### Checklist


FrankYFTang:
The case you show is
// E.g. しかもロゴがｓｏｆｔｂａｎｋ！ -> しかも▁ロゴが▁ｓｏｆｔｂａｎｋ！-> breakpoint between が and ｓ

My question is why is "ｓｏｆｔｂａｎｋ" passing down to CjkBreakEngine but not break in rule level?

markusicu:
> My question is why is "ｓｏｆｔｂａｎｋ" passing down to CjkBreakEngine but not break in rule level?

Yes, good question. It looks like the phrase-specific expansion of the rule "set of dictionary characters" went too far, and the code change here is trying to compensate for that. Please try to instead adjust the sets of characters in the rule files.

It sounds like this is due to the addition of `$EAFWH = [\p{ea=F}\p{ea=W}\p{ea=H}]` to the $ALPlus set. This is discussed in the [implementation doc](https://docs.google.com/document/d/1Lg50cEKCLowTL3JwCXlOXOYpDkYJAeKy0I8E7qk7mNs/edit#heading=h.jwutcf3l387w).

I understand that we want to keep words together that are spelled with fullwidth letters (and I don't understand why the default rules don't do so), but (a) this should not cause those fullwidth letters to be handed to the dictionary-breaking code, and (b) the $EAFWH set contains a ton of characters that are not letters and should most likely not be added to $ALPlus.

I tried to figure out what this adds beyond the regular $ALPlus and the other $dictionary additions. I came up with almost 76000 additional code points; some 65000 of these are unassigned. Please review.
-->
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%5Cp%7Bea%3DF%7D%5Cp%7Bea%3DW%7D%5Cp%7Bea%3DH%7D%5D+-+%5B%5B%3Alb%3DAL%3A%5D+%5B%3Alb%3DAI%3A%5D+%5B%3Alb%3DSG%3A%5D+%5B%3Alb%3DXX%3A%5D+%5B%5B%3Alb%3DSA%3A%5D%5B%3AHan%3A%5D%5B%3AKatakana%3A%5D%5B%3AHiragana%3A%5D%5B%5Cuac00-%5Cud7a3%5D-%5B%5B%3AMn%3A%5D%5B%3AMc%3A%5D%5D%5D%5D&g=ea+gc&i=

markusicu:
GitHub does not let me add @hiroyuki-komatsu as a reviewer...?

aheninger:
This change is touching on areas that I strongly suspect are also broken in regular (non-phrase) word and line breaking.

The transition between ranges of text handled by rules and those handled by dictionaries is full of edge cases that have never been sorted out. For example, I'm pretty sure that Latin characters will glue together Thai words that would otherwise have a break between them, which makes no sense.

I guess what I am trying to say is, do whatever works to get phrase breaking up for now, but the whole area really wants some careful re-thinking, for all word and line breaking, not just phrase.

icu4c/source/common/dictbe.cpp
```diff
@@ -1061,8 +1061,10 @@ CjkBreakEngine::CjkBreakEngine(DictionaryMatcher *adoptDictionary, LanguageType
     // Korean dictionary only includes Hangul syllables
     fHangulWordSet.applyPattern(UnicodeString(u"[\\uac00-\\ud7a3]"), status);
     fHangulWordSet.compact();
-    fNumberOrOpenPunctuationSet.applyPattern(UnicodeString(u"[[:Nd:][:Pi:][:Ps:]]"), status);
-    fNumberOrOpenPunctuationSet.compact();
+    // Add number, open punctuation and a-z/A-Z with half width and full width.
+    fNumberOrOpenPunctuationOrLatinSet.applyPattern(
+        UnicodeString(u"[[:Nd:][:Pi:][:Ps:][:Lu:][:Ll:]]"), status);
```
markusicu, line 1066: Lu+Ll seem broad but incomplete. Should we use Alphabetic instead?
https://www.unicode.org/reports/tr44/#Alphabetic
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AAlphabetic%3A%5D&abb=on&g=&i=
```suggestion
    // Digits, open punctuation, and "Alphabetic" characters.
    fNumberOrOpenPunctuationOrAlphaSet.applyPattern(
        UnicodeString(u"[[:Nd:][:Pi:][:Ps:][:Alphabetic:]]"), status);
```

markusicu (commented):
Thanks a lot, @allensu05 & @aheninger for the explanation!
This makes sense to me.
Allen: Please take a look at my question/suggestion.

icu4c/source/common/dictbe.cpp
allensu05: Thanks. Resolved

markusicu:
Now that PR #1970 is merged, please rebase this PR, resolve conflicts (probably just need to make sure that the test additions from both PRs are present), and double-check that this is still doing the right thing and passing tests.

markusicu (approved):
lgtm tnx

PR #1970, opened 2022-01-28
ICU-21699 Revise phrase rule file

allensu05:
1. [U+30FC](https://www.compart.com/en/unicode/U+30FC) doesn't belong to Hira, Kana and Han. Add it into $dictionary
2. Revise $ALPlus by adding fullwidth char of Latin and digit rather than all East Asian Width char with class F, W and H

##### Checklist


markusicu:
This looks very much related to the discussion in https://github.com/unicode-org/icu/pull/1968#issuecomment-1024792431
I suspect that we need to figure out the right set of additional $ALPlus characters to fix both issues.

hiroyuki-komatsu:
> I'll confirm internally if there is any other char like U+30FC.

Here is the list of characters like U+30FC (ー).

https://docs.google.com/spreadsheets/d/1MYvDEuXqDeDSUM-EmS3d4DZytMF3aBfu6YTuU_wzDkQ/edit#gid=29114612

The reference is https://www.w3.org/TR/jlreq/#cl-09

Note, I suggested Allen to connect U+30FC (ー) to the previous and next chars, but the connection to the next character comes from the implementation limitation. We might want to start connecting the previous char only at first.

allensu05:
> > I'll confirm internally if there is any other char like U+30FC.
> 
> Here is the list of characters like U+30FC (ー).
> 
> https://docs.google.com/spreadsheets/d/1MYvDEuXqDeDSUM-EmS3d4DZytMF3aBfu6YTuU_wzDkQ/edit#gid=29114612
> 
> The reference is https://www.w3.org/TR/jlreq/#cl-09
> 
> Note, I suggested Allen to connect U+30FC (ー) to the previous and next chars, but the connection to the next character comes from the implementation limitation. We might want to start connecting the previous char only at first.

Thanks Hiro.
1. The element in the list belongs to either Hira or Kana  so it's unnecessary to add into $ALPlus.
2. [U+30FC](https://www.compart.com/en/unicode/U+30fc) doesn't belong to Hira, Kana and Han. Add it into $dictionary
3. Revise $ALPlus by adding fullwidth char of Latin and digit rather than all East Asian Width char with class F, W and H

icu4c/source/data/brkitr/rules/line_loose_phrase_cj.txt
```diff
@@ -127,7 +126,8 @@ $dictionary     = [$ComplexContext $dictionaryCJK];
 #                               XX  (Unknown, unassigned)
 #                         as $AL  (Alphabetic)
 #
-$ALPlus = [$AL $AI $SG $XX $EAFWH [$dictionary-[[:Mn:][:Mc:]]]];
+$EAW = [[:Lu:][:Ll:][\uff10-\uff19]];
```
markusicu, line 129: Thanks for reducing the additions. However, the `[:Lu:][:Ll:]` *look* large, and it's hard to see what really gets added.
Let's be specific, for example:
```suggestion
# Let fullwidth-ASCII digits and letters be part of words.
$FW_alphanum = [０-９Ａ-Ｚａ-ｚ];
```

markusicu:
The changes lgtm, thanks.

However, *almost all of the Windows builds* are failing.
@jefgen any idea why that would be?
@aheninger @sffc @FrankYFTang @pedberg is it possible that the BreakIterator rules files are not built with UTF-8 but assuming the platform encoding? The rules now include `[０-９Ａ-Ｚａ-ｚ]` but surely these are not the first non-ASCII characters in brkitr rule files?!
And if that was the problem, why would two Windows builds pass?
@allensu05 you *could try* to escape these six characters again... `[\uff10-\uff19\uff21-\uff3a\uff41-\uff5a]`

FrankYFTang:
I am not sure what encoding it is assming. but I believe other rule files always use \u escape text in the rule file so I agree with Markus we should change to that.

jefgen:
> almost all of the Windows builds are failing.
>@jefgen any idea why that would be?

I'm not sure off-hand, but it looks like the Windows builds that were passing/green were the ones that *don't* actually run any of the tests. Some of them just compile, and don't and/or can't run the tests. (ex: the ARM32/ARM64 builds just compile, but don't run tests as the pipeline doesn't support native ARM execution).

I wonder if maybe there is some encoding confusing going on somewhere. The file `line_loose_phrase_cj.txt` in two different editors, one says it's UTF-8 and the other confusingly thinks it is TIS-620 character set.

markusicu (approved):
lgtm tnx

@FrankYFTang @hiroyuki-komatsu ok?

markusicu:
> > almost all of the Windows builds are failing.
> > @jefgen any idea why that would be?
> 
> I'm not sure off-hand, but it looks like the Windows builds that were passing/green were the ones that _don't_ actually run any of the tests. Some of them just compile, and don't and/or can't run the tests. (ex: the ARM32/ARM64 builds just compile, but don't run tests as the pipeline doesn't support native ARM execution).

Thanks, that's useful.
I submitted https://unicode-org.atlassian.net/browse/ICU-21907 to follow up.

> I wonder if maybe there is some encoding confusing going on somewhere. The file `line_loose_phrase_cj.txt` in two different editors, one says it's UTF-8 and the other confusingly thinks it is TIS-620 character set.

I suspect that the editor that detected a Thai single-byte charset has a misbehaving heuristic based on just a few non-ASCII bytes. I doubt that Allen managed to copy-paste from the browser into his editor on a non-Windows machine and end up with something other than UTF-8. And, the tests passed on all of the Linux and Mac machines.

aheninger:
Looking into the genbrk tool, it does indeed assume the rule source files are in the system code page. It also detects an honors a BOM.

https://github.com/unicode-org/icu/blob/205144e3912d6bcf23f7fcba04f95fdfa2b4ea70/icu4c/source/tools/genbrk/genbrk.cpp#L246

This code hasn't been touched since its initial check-in of 20 years ago. "Blame" says it's mine, although I don't have much recollection of it.

The rbbi API for building from rules takes a UnicodeString. The code page stuff is in the genbrk tool.

hiroyuki-komatsu:
> @FrankYFTang @hiroyuki-komatsu ok?

LGTM. Thank you,

PR #1990, opened 2022-02-20
ICU-21919 Fix fuzzer bug (buffer overflow) in Dutch accented IJ titlecase

echeran:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21919

### DIFF
--- a/icu4c/source/common/ucasemap.cpp
+++ b/icu4c/source/common/ucasemap.cpp
@@ -437,6 +437,7 @@ constexpr uint8_t ACUTE_BYTE1 = u8"\u0301"[1];
  */
 int32_t maybeTitleDutchIJ(const uint8_t *src, UChar32 c, int32_t start, int32_t segmentLimit,
                           ByteSink &sink, uint32_t options, icu::Edits *edits, UErrorCode &errorCode) {
+    U_ASSERT(start < segmentLimit);
 
     int32_t index = start;
     bool withAcute = false;
@@ -594,7 +595,7 @@ ucasemap_internalUTF8ToTitle(
                 }
 
                 /* Special case Dutch IJ titlecasing */
-                if (titleStart+1 < index &&
+                if (titleLimit < index &&
                     caseLocale == UCASE_LOC_DUTCH) {
                     if (c < 0) {
                         c = ~c;
--- a/icu4c/source/common/ustrcase.cpp
+++ b/icu4c/source/common/ustrcase.cpp
@@ -416,6 +416,7 @@ namespace {
 int32_t maybeTitleDutchIJ(const UChar *src, UChar32 c, int32_t start, int32_t segmentLimit,
                           UChar *dest, int32_t &destIndex, int32_t destCapacity, uint32_t options,
                           icu::Edits *edits) {
+    U_ASSERT(start < segmentLimit);
 
     int32_t index = start;
     bool withAcute = false;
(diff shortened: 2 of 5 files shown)


echeran:
I was able to reproduce the oss-fuzz bug using the testcase data from Norbert and setting up the oss-fuzz image to [point to my local copy of ICU source](https://google.github.io/oss-fuzz/advanced-topics/reproducing/#reproduce-using-local-source-checkout).

I was able to show to myself that the change works for oss-fuzz. To do so, I had to remove and recreate the oss-fuzz image and re-run.

I wasn't able to reproduce the ASAN error through the added, test cases alone, though, but I've left in the test cases anyways.

icu4c/source/common/ustrcase.cpp
```diff
@@ -416,6 +416,9 @@ namespace {
 int32_t maybeTitleDutchIJ(const UChar *src, UChar32 c, int32_t start, int32_t segmentLimit,
                           UChar *dest, int32_t &destIndex, int32_t destCapacity, uint32_t options,
                           icu::Edits *edits) {
+    if (start >= segmentLimit) {
```
markusicu, line 419: This function is called with `(start, segmentLimit)` = `titleStart + 1, index` and this is inside the condition `if (titleStart+1 < index && ...` so it should not be possible for this "if" here to be true.

icu4c/source/common/ucasemap.cpp
```diff
@@ -437,6 +437,9 @@ constexpr uint8_t ACUTE_BYTE1 = u8"\u0301"[1];
  */
 int32_t maybeTitleDutchIJ(const uint8_t *src, UChar32 c, int32_t start, int32_t segmentLimit,
                           ByteSink &sink, uint32_t options, icu::Edits *edits, UErrorCode &errorCode) {
+    if (start >= segmentLimit) {
```
markusicu, line 440: This UTF-8 version is called with `(start, segmentLimit)` = `titleLimit, index` (different from UTF-16) and this is inside the condition `if (titleStart+1 < index && ...` (just like in the UTF-16 version).

So here the problem is that when `c == u'Í'` then titleLimit==titleStart+2 (while in UTF-16 it's just +1).

I think the call site should be changed so that https://github.com/unicode-org/icu/blob/main/icu4c/source/common/ucasemap.cpp#L597 has `if (titleLimit < index &&`

Then you need not change this function.

icu4c/source/test/intltest/strcase.cpp
```diff
@@ -763,7 +766,40 @@ void StringCaseTest::TestDutchTitle() {
                 testOptions
             );
         }
-        
+    }
+
+    // Bug ICU-21919
```
markusicu, line 771: You don't really need this when you have the additional test case line above on line 746, do you?

icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
```diff
@@ -794,6 +794,10 @@ public static String toTitle(int caseLocale, int options, BreakIterator iter, Ch
     private static <A extends Appendable> int maybeTitleDutchIJ(
             CharSequence src, int c, int start, int segmentLimit,
             A dest, int options, Edits edits) throws IOException {
+        if (start >= segmentLimit) {
```
markusicu, line 797: You should not need this, assuming it works like the UTF-16 code in C++.

optional:
If you like, you could
```
assert start < segmentLimit;
```
(and U_ASSERT(...) in C++)

markusicu (commented):
good debugging, thanks!

icu4c/source/common/ucasemap.cpp
echeran: Done.  That makes sense.

icu4c/source/test/intltest/strcase.cpp
echeran: Done. No, it's not necessary.

icu4j/main/classes/core/src/com/ibm/icu/impl/CaseMapImpl.java
echeran: Done.

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm tnx

PR #1991, opened 2022-02-21
ICU-21592 Update cj normal/loose linebreak per CSS

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21592

This updates the CJ normal & loose linebreak tailorings per a recent change agreed on for CSS:
- For normal, CSS no longer allows breaks before \u2010 HYPHEN or \u2013 EN DASH, i.e. it makes the behavior for these back to what standard linebreak does. These are normally in linebreak class BA but the cj normal tailoring has split them off into a separate class BAX to allow breaks; this PR retracts that.
- For loose, CSS allows breaks before  \u2010 HYPHEN or \u2013 EN DASH only after characters of class ID; for other characters it no longer does. These characters are still split into a separate class BAX, but the first part of rule LB 21 that formerly allowed a break before BAX now needs to be split into two rules, one for non-ID and one for ID. (BAX also needed to ba added to $CanFollowIS in rule LB 14a).

The separate rules for RBBIMonkeyTest in `icu4c/source/test/testdata/break_rules/` and `icu4j/main/tests/core/src/com/ibm/icu/dev/test/rbbi/break_rules/` needed to be updated accordingly. However they also needed one additional update for `line_loose_cj.txt`: They needed the following addition corresponding to something that has been in the standard rules for a while, but whose absence in the RBBIMonkeyTest did not previously cause a problem:
```
LB21.3:      CM+ [BA HY NS];
```

While the CI tests are running I will in parallel run an extended-duration version of RBBIMonkeyTest locally to check for any hard-to-find problems, and note the result below.

### DIFF
--- a/icu4c/source/test/testdata/break_rules/line_loose_cj.txt
+++ b/icu4c/source/test/testdata/break_rules/line_loose_cj.txt
@@ -20,7 +20,8 @@
 #         line-break=loose (BCP47 -u-lb-loose) as defined for Chinese & Japanese.
 #         It sets characters of class CJ to behave like ID.
 #         In addition, it allows breaks:
-#         * before hyphens 2010 & 2013 (both BA) and 301C, 30A0 (both NS)
+#         * between ID and hyphens 2010 & 2013 (both BA)
+#         * before 301C, 30A0 (both NS)
 #         * before iteration marks 3005, 303B, 309D, 309E, 30FD, 30FE (all NS)
 #         * between characters of LineBreak class IN such as 2026
 #         * before some centered punct 203C, 2047, 2048, 2049, 30FB, FF1A, FF1B,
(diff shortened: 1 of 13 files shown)


pedberg-icu:
I ran the ICU4C testMonkey for 400000 iterations (compared to 100 for a normal quick test), which took about 45 min; no problems found.

FrankYFTang:
@allensu05 please also take a look.

icu4c/source/test/testdata/break_rules/line_loose_cj.txt
```diff
@@ -200,8 +201,10 @@ LB20.09:     ^(HY | HH) CM* AL;
 
 LB21a:       HL CM* (HY | BA | BAX) CM* [^CM CB]?;
 
-LB21.1:      . CM* [BA HY NS];
-LB21.2:      BB CM* [^CM CB];
+LB21.1:      [^BK CR LF NL CM ZW SP CB ID] CM* [BA BAX HY NS];
```
aheninger, line 204: I don't think the long negated set is necessary; the hard breaks should never reach this  point, having been handled by earlier rules. Monkey test rules are handled sequentially, unlike the main production rules, which are run in parallel.

aheninger (approved):
The changes to the main rules look good.

I think the LB21 changes in the line_loose_cj.txt monkey test rules can be simplified, but they're probably good enough for the moment, since you're under time pressure.

I'll play with them, and make a followup PR if I come up with something better.

icu4c/source/test/testdata/break_rules/line_loose_cj.txt
pedberg-icu: Thanks Andy for looking at this and for approving!

PR #2004, opened 2022-02-25
ICU-21592 Linebreak loose cj rules cleanup

aheninger:
This is a followup to PR #1991, Update cj normal/loose linebreak per CSS

The original change to the line_loose_cj rules involved splitting hyphens out
of the BA (Break After) class, allowing a break when they follow an ID. This
change simplifies the the rules for doing that.

It also fixes a problem with the original change that had altered the behavior
of BAX hyphens that followed Regional Indicators or Unattached Combining Marks.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21592


pedberg-icu (approved):
Thanks for the fixes!

PR #2026, opened 2022-03-10
ICU-21840 Add alias units data

younies:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21840


younies:
@pedberg-icu , I still need to run the tool to generate the data. this should be done by tomorrow.

pedberg-icu:
> @pedberg-icu , I still need to run the tool to generate the data. this should be done by tomorrow.

@younies, please don't do that. I need to integrate CLDR beta1 to ICU starting this afternoon, and your doing a subset of that separately will create conflicting PRs (one or the other will have to be redone because of data jar conflict). What I would rather do is to pick up the change in this PR as part of my CLDR to ICU integration work.

tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_supplemental.txt
```diff
@@ -166,6 +166,10 @@
     ; /convertUnits/$1/factor ; values="$3"
     ; /convertUnits/$1/offset ; values=$4
 
+//supplementalData/metadata/alias/unitAlias[@type="(%W)"][@replacement="(%W)"][@reason="(%W)"]?
+    ; /alias/$1/replacement ; values=$2
+    ; /alias/$1/reason ; values="$3"
```
pedberg-icu, line 171: I do not think we can use just /alias/ here because we already have alias as the top-level bundle in metadata.txt. In my PR for CLDR-ICU integration I will use unitsAlias. Also, compare with how supplementalMetadata.xml is handled.

pedberg-icu: Also we should map usageAlias too.

younies: > > @pedberg-icu , I still need to run the tool to generate the data. this should be done by tomorrow.
> 
> @younies, please don't do that. I need to integrate CLDR beta1 to ICU starting this afternoon (as I mentioned in the ICU TC meeting), and your doing a subset of that separately will create conflicting PRs (one or the other will have to be redone because of data jar conflict). What I would rather do is to pick up the change in this PR as part of my CLDR to ICU integration work. Or else you could run that process after I am done and my integration PR is merged.

@pedberg-icu , I totally understand. You can pick the change and add it to you PR in the place and format you feel it is suitable. and I will focus in fixing Units Code.

pedberg-icu: > I do not think we can use just /alias/ here because we already have alias as the top-level bundle in metadata.txt. In my PR for CLDR-ICU integration I will use unitsAlias.

Actually never mind, I guess the idea is that this data should go into metadata.txt with all of the other alias data. So then it should be `alias` and not `unitsAlias`.

pedberg-icu: @younies, my https://github.com/unicode-org/icu/pull/2027 now has the alias mappings for the unit items (unitAlias, usageAlias). They go into metadata.txt with the other aliases for deprecated items.

pedberg-icu:
I think this PR can be closed, the necessary changes were done in #2027 instead.

tools/cldr/cldr-to-icu/src/main/resources/ldml2icu_supplemental.txt
younies: yes, I am going to close it.

PR #2028, opened 2022-03-11
ICU-21332 Add tests from ICU4C TestDelimiters() into related ICU4J test

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21332

Add tests from ICU4C test into related ICU4J test to verify proper inheritance of delimiters, especially in some cases where there are partial overrides.


richgillam (approved):
Looks good.  I take it the test passes and we don't need to make any fixes to the actual Java code?

pedberg-icu:
> Looks good. I take it the test passes and we don't need to make any fixes to the actual Java code?

Correct, I think the original problem may have been fixed by some of your resource manager fixes (or else by a a change to the sink for loading delimiter data); but in any case the original problem is gone.

PR #2039, opened 2022-03-16
ICU-21946 do not reset useful cache data to empty in populateNear()

llunak:
https://unicode-org.atlassian.net/browse/ICU-21946


markusicu:
@aheninger let us know if you don't want to be the assignee here...

markusicu:
Thanks for working on this. I just merged @aheninger 's replacement PR #2098.

PR #2050, opened 2022-03-25
ICU-20715 Fix fuzzing problem in CollationDataBuilder.

richgillam:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20715

Long explanation of what was wrong and how this fix addresses it:

The fuzzer was passing random garbage to the RuleBasedCollator constructor as a collation rule string.  If you try to parse the string as a set of collation rules, you wind up with a couple of very big contraction strings, each with a very large number of canonical-equivalent strings, and many also with before context.  Even though the crash (or ASAN failure) is in UCharsTrie, both UCharsTrie and its builder are working fine.  The problem is in the data structure that contains them.

The collator organizes all of its various before-context and after-context strings into a series of UCharsTrie structures, which are concatenated together and stored in a single UnicodeString (for simplicity).  We'll call this the "contexts array."  The collation element values that use these tries incorporate their absolute indexes into the contexts array.  For the final collator, this isn't a problem, but while the builder is building the collator, it puts lots of pressure on its contexts array.  Every time it encounters a new rule (or variant of an existing rule, as you get when expanding canonical equivalents) that has a context trie, it creates a new context trie and appends it to the end of the contexts array.  If the new context means updating an existing context trie, a new version of that trie is generated and appended to the end of the contexts array, effectively leaking the old version.

If the length of the contexts array exceeds Collation::MAX_INDEX, instead of bailing out and reporting an error, we were just clearing out the contexts array and re-adding the new trie to the now-empty contexts array.  In many cases, this would be fine, because everything in it was dead anyway, but in some cases (such as the fuzzing test), we'd have collation element values in other data structures that pointed to tries in the old contexts array and weren't rebuilt after that array was emptied.  The collation key would be pointing to a position in the contexts array that was either off the end or (in the fuzzing failure) no longer the root node of a UCharsTrie.  We'd try to traverse it as if it WERE the root note, and we'd run off the end of the array and either crash or trigger an ASAN failure.

So the fix was to change things so that if we exceeded Collation::MAX_INDEX, we'd dump out and report an error.  This involved adding "errorCode" parameters to a number of functions in the call chain that didn't already have them, so that the error could propagate back up to the RuleBasedCollator constructor.

Unfortunately, while this fix did solve the fuzzing problem (now we just dump out with an error, as we should), it caused genrb to fail (with the new error) while building the Japanese collation rules.  We were getting away with this before because everything in the contexts array at the point where it was cleared WAS dead and could be safely discarded.

An ideal solution to that problem would be to replace the builder's contexts array with some other type of structure that would let us throw out old tries when they get updated, or to rebuild the whole contexts array whenever one of the tries gets updated.  But both of these solutions would require us to use different collation element values that didn't depend on the physical location of the trie in the contexts array, or to update all the existing collation element values any time we regenerated the contexts array.  This was way too big a change for me to want to attempt without a lot of help from somebody who understands this code better than I do.

What I did instead was to add code that would check if the new trie being added to the contexts array is an update of the last trie that was added.  If so, we just overwrite the old version of the trie and reuse its collation-element value.  This saves a lot of space if you're repeatedly updating the same trie, which is what the Japanese rule set is actually doing, but it doesn't do much for you if there are multiple tries in the contexts array and they get updated in random order (or two get updated in alternating order, which happens if a rule has both before context AND a contraction).  It's enough to get the Japanese collator building, though, so I'm going with it and leaving a more comprehensive solution for later.

One more note about the fix: I'm detecting whether I'm updating the trie at the end of the contexts array by checking its defaultCE32 value.  This doesn't completely work, though-- if a rule has both before context and contractions, you get two different tries with the same defaultCE32 value: One for the before context (prefix) and a different one for the contraction information (suffixes).  I got around this by adding a flag that lets me differentiate between the two cases, but this is inelegant.

I didn't add a regression test for this fix because reusing the fuzzer test would give us a test that'd either crash or only show an error if you were using ASAN.  Explicitly running the fuzzer test using Frank's instructions above does show that the problem is fixed.

This is a C++-only fix.  I checked the Java code, and it DOES throw an exception in this situation, so it didn't have the same problem (and it was okay there because the Java code isn't used to build our collation data files).

### DIFF
--- a/icu4c/source/i18n/collationbuilder.cpp
+++ b/icu4c/source/i18n/collationbuilder.cpp
@@ -307,10 +307,13 @@ CollationBuilder::addReset(int32_t strength, const UnicodeString &str,
             parserErrorReason = "normalizing the reset position";
             return;
         }
-        cesLength = dataBuilder->getCEs(nfdString, ces, 0);
+        cesLength = dataBuilder->getCEs(nfdString, ces, 0, errorCode);
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason = "reset position maps to too many collation elements (more than 31)";
+            return;
+        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
+            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
             return;
         }
     }
@@ -734,11 +737,14 @@ CollationBuilder::addRelation(int32_t strength, const UnicodeString &prefix,
             parserErrorReason = "normalizing the relation extension";
             return;
         }
-        cesLength = dataBuilder->getCEs(nfdExtension, ces, cesLength);
+        cesLength = dataBuilder->getCEs(nfdExtension, ces, cesLength, errorCode);
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason =
                 "extension string adds too many collation elements (more than 31 total)";
+            return;
+        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
+            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
             return;
         }
     }
@@ -1182,10 +1188,12 @@ CollationBuilder::addTailComposites(const UnicodeString &nfdPrefix, const Unicod
                                      newNFDString, newString, errorCode)) {
             continue;
         }
-        int32_t newCEsLength = dataBuilder->getCEs(nfdPrefix, newNFDString, newCEs, 0);
+        int32_t newCEsLength = dataBuilder->getCEs(nfdPrefix, newNFDString, newCEs, 0, errorCode);
         if(newCEsLength > Collation::MAX_EXPANSION_LENGTH) {
             // Ignore mappings that we cannot store.
             continue;
+        } else if (U_FAILURE(errorCode)) {
+            return;
         }
         // Note: It is possible that the newCEs do not make use of the mapping
         // for which we are adding the tail composites, in which case we might be adding
@@ -1329,7 +1337,7 @@ CollationBuilder::closeOverComposites(UErrorCode &errorCode) {
     while(iter.next()) {
         U_ASSERT(!iter.isString());
         nfd.getDecomposition(iter.getCodepoint(), nfdString);
-        cesLength = dataBuilder->getCEs(nfdString, ces, 0);
+        cesLength = dataBuilder->getCEs(nfdString, ces, 0, errorCode);
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             // Too many CEs from the decomposition (unusual), ignore this composite.
             // We could add a capacity parameter to getCEs() and reallocate if necessary.
@@ -1347,7 +1355,7 @@ CollationBuilder::addIfDifferent(const UnicodeString &prefix, const UnicodeStrin
                                  UErrorCode &errorCode) {
     if(U_FAILURE(errorCode)) { return ce32; }
     int64_t oldCEs[Collation::MAX_EXPANSION_LENGTH];
-    int32_t oldCEsLength = dataBuilder->getCEs(prefix, str, oldCEs, 0);
+    int32_t oldCEsLength = dataBuilder->getCEs(prefix, str, oldCEs, 0, errorCode);
     if(!sameCEs(newCEs, newCEsLength, oldCEs, oldCEsLength)) {
         if(ce32 == Collation::UNASSIGNED_CE32) {
             ce32 = dataBuilder->encodeCEs(newCEs, newCEsLength, errorCode);
--- a/icu4c/source/i18n/collationbuilder.cpp
+++ b/icu4c/source/i18n/collationbuilder.cpp
@@ -308,12 +308,14 @@ CollationBuilder::addReset(int32_t strength, const UnicodeString &str,
             return;
         }
         cesLength = dataBuilder->getCEs(nfdString, ces, 0, errorCode);
+        if (U_FAILURE(errorCode)) {
+            // NOTE: as of right now (3/25/22), the only error getCEs() returns is INPUT_TOO_LONG_ERROR
+            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
+            return;
+        }
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason = "reset position maps to too many collation elements (more than 31)";
-            return;
-        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
-            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
             return;
         }
     }
@@ -738,13 +740,15 @@ CollationBuilder::addRelation(int32_t strength, const UnicodeString &prefix,
             return;
         }
         cesLength = dataBuilder->getCEs(nfdExtension, ces, cesLength, errorCode);
+        if (U_FAILURE(errorCode)) {
+            // NOTE: as of right now (3/25/22), the only error getCEs() returns is INPUT_TOO_LONG_ERROR
+            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
+            return;
+        }
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason =
                 "extension string adds too many collation elements (more than 31 total)";
-            return;
-        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
-            parserErrorReason = "exhausted maximum CONTRACTION_TAG value";
             return;
         }
     }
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/collationbuilder.cpp
```diff
@@ -307,11 +307,14 @@ CollationBuilder::addReset(int32_t strength, const UnicodeString &str,
             parserErrorReason = "normalizing the reset position";
             return;
         }
-        cesLength = dataBuilder->getCEs(nfdString, ces, 0);
+        cesLength = dataBuilder->getCEs(nfdString, ces, 0, errorCode);
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason = "reset position maps to too many collation elements (more than 31)";
             return;
+        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
```
pedberg-icu, line 315: Just a question/comment: Can getCEs return any other errors? (And stylistically I like putting the error handling first, though I assume it does not make any practical difference here or the other similar places since getsCEs returns 0 if error).

pedberg-icu (approved):
Wow, a nightmare to debug and fix, congratulations. Fix looks good to me but Markus should take a look.

icu4c/source/i18n/collationbuilder.cpp
```diff
@@ -1182,10 +1188,12 @@ CollationBuilder::addTailComposites(const UnicodeString &nfdPrefix, const Unicod
                                      newNFDString, newString, errorCode)) {
             continue;
         }
-        int32_t newCEsLength = dataBuilder->getCEs(nfdPrefix, newNFDString, newCEs, 0);
+        int32_t newCEsLength = dataBuilder->getCEs(nfdPrefix, newNFDString, newCEs, 0, errorCode);
         if(newCEsLength > Collation::MAX_EXPANSION_LENGTH) {
             // Ignore mappings that we cannot store.
             continue;
+        } else if (U_FAILURE(errorCode)) {
```
FrankYFTang, line 1195: no need to put the if after "else" since the previous block will always continue anyway.

icu4c/source/i18n/collationbuilder.cpp
```diff
@@ -734,12 +737,15 @@ CollationBuilder::addRelation(int32_t strength, const UnicodeString &prefix,
...
             return;
         }
-        cesLength = dataBuilder->getCEs(nfdExtension, ces, cesLength);
+        cesLength = dataBuilder->getCEs(nfdExtension, ces, cesLength, errorCode);
         if(cesLength > Collation::MAX_EXPANSION_LENGTH) {
             errorCode = U_ILLEGAL_ARGUMENT_ERROR;
             parserErrorReason =
                 "extension string adds too many collation elements (more than 31 total)";
             return;
+        } else if (errorCode == U_INPUT_TOO_LONG_ERROR) {
```
FrankYFTang, line 746: no need to put the if after else since the previous block will always return and no chance it will reach the if here anyway

icu4c/source/i18n/collationbuilder.cpp
FrankYFTang, line 315: no need to put the if after else here since the previous block will always return and no chance to reach the if check here anyway

richgillam: >Just a question/comment: Can getCEs return any other errors?

Come to think of it, it can't.  I checked the specific error code because I worried there might be more errors in the future, and then the error string this function is fabricating would be wrong.  But I'll change it.

>(And stylistically I like putting the error handling first, though I assume it does not make any practical difference here or the other similar places since getsCEs returns 0 if error).

I see your point.  I'll change this too.

richgillam: You're right: there's no need, but I like the "else" there anyway-- I think it's clearer.

But it's not a hill I want to die on, especially after what I've gone through to fix this bug.  I'm changing it.

icu4c/source/i18n/collationbuilder.cpp
richgillam: Same response as above, but I went ahead and made both your and Peter's suggested changes here too.

markusicu:
Hi Rich...

First: Sorry! Pretty much all bugs in the collation code are my fault. And sorry for not being more responsive...

Second: Thank you! Thanks for debugging and figuring this out! The collation builder is complex.

Too bad that because of Japanese we can't just bail out.

In principle, your workaround of "check if we are replacing the last trie" sounds ok. However, the `defaultCE32 == lastContextCE32` test seems a little fragile for that. The defaultCE32 is the CE that you get when none of the contraction suffixes match. It would be unusual but not impossible to construct rules where two different characters sort the same on their own and each start contraction strings.

I think that a reliable test would be to check the contraction index in the CE that is about to be rewritten with the index to the newly built context sub-array. If it points to lastContextIndex, then, well, we are replacing the last context. We should then not need any state other than lastContextIndex.

I will take a look at the caller of addContextTrie() to see if we can scrounge up enough information.

markusicu:
Also, I think it would actually be good to add a unit test with the rule string that the fuzzer came up with, just to make sure that the code can handle it -- succeed or fail with an error. The CI tests include ASAN, so they would find problems with that. Good builder stress test.

richgillam:
Markus—

> First: Sorry! Pretty much all bugs in the collation code are my fault. And sorry for not being more responsive...
> 

No apology necessary.  We all know how these things go.  :-)

> Second: Thank you! Thanks for debugging and figuring this out! The collation builder is complex.

Boy, was that an adventure!  I’m just glad I finally got there,

> Too bad that because of Japanese we can't just bail out.

Yeah...

> In principle, your workaround of "check if we are replacing the last trie" sounds ok. However, the defaultCE32 == lastContextCE32 test seems a little fragile for that. The defaultCE32 is the CE that you get when none of the contraction suffixes match. It would be unusual but not impossible to construct rules where two different characters sort the same on their own and each start contraction strings.

That makes sense, and my current fix is kind of ugly, so I’m not averse to changing it.

> I think that a reliable test would be to check the contraction index in the CE that is about to be rewritten with the index to the newly built context sub-array. If it points to lastContextIndex, then, well, we are replacing the last context. We should then not need any state other than lastContextIndex.
> 
> I will take a look at the caller of addContextTrie() to see if we can scrounge up enough information.

I like that.  If you can fill in the details on how to do that, I’ll fix it.

In fact, I’d love it if we could come up with a more robust solution overall than the one I have right now.  We’re getting by with my solution simply because of the way the Japanese collation rules happen to be structured, and if that changes substantially, all bets are off.  Unfortunately, I’m worried a more thorough solution will require rewriting a substantial chunk of the collation builder, and I’m scared to even think of that.  But if it’s easier than I think, please clue me in and I’ll take a shot (either in this PR [if it’s REALLY easy] or in a future one).

> Also, I think it would actually be good to add a unit test with the rule string that the fuzzer came up with, just to make sure that the code can handle it -- succeed or fail with an error. The CI tests include ASAN, so they would find problems with that. Good builder stress test.
> 

My worry was that having a test that crashes if the original problem recurs wouldn’t be a great idea, but if that’s okay, and the unit tests are already running under ASAN, I can add a unit test.

I’d appreciate it if you’d make these comments in the PR, so we have a public record and I don’t forget.

I’ll come back and fix this stuff on Monday; I’m going to try to take the rest of today to catch up on other things I’ve been letting slide.

Thanks again for all your help and patience!

—Rich

markusicu:
I turned the failing case into a proper unit test and tried to make it slightly readable. See https://github.com/unicode-org/icu/pull/2052

Note that that includes a Java unit test. Predictably, it fails in Java as well -- and I am set up better for debugging in Java than in C++.

----
Generally, the problem is that the builder has to be able to run the collation algorithm in order to make something sort like, or just after, something else; the collation algorithm is complex; and while building, the data is in flux. For easy stuff, the builder implementation can fetch collation elements directly from its mutable data, but for the complicated stuff I didn't want to write another implementation of the collation algorithm, and so it builds runtime-appropriate structures and lets the runtime implementation evaluate them.

----
I will take a look at the CollationBuilder::addIfDifferent() logic and what the builder does and see if I can swap some more of this stuff back into my brain.

richgillam:
Markus--

> I turned the failing case into a proper unit test and tried to make it slightly readable. See #2052 <https://github.com/unicode-org/icu/pull/2052>
The new test looks great.  I appreciate your doing this, but I would have gotten around to it on Monday.  Are you planning to check it in, or do you just want me to incorporate it into my PR?
> Note that that includes a Java unit test. Predictably, it fails in Java as well -- and I am set up better for debugging in Java than in C++.
> 
Well, we WANT it to fail, don’t we?  We just don’t want it to CRASH.  Or are you saying it crashes (or does some else terrible) in both C++ and Java?

> 
> Generally, the problem is that the builder has to be able to run the collation algorithm in order to make something sort like, or just after, something else; the collation algorithm is complex; and while building, the data is in flux. For easy stuff, the builder implementation can fetch collation elements directly from its mutable data, but for the complicated stuff I didn't want to write another implementation of the collation algorithm, and so it builds runtime-appropriate structures and lets the runtime implementation evaluate them.
> 
I understand, and I’m sympathetic.  I guess I was wondering if there was a way to get rid of all the redundancy without changing the whole data structure.

> 
> I will take a look at the CollationBuilder::addIfDifferent() logic and what the builder does and see if I can swap some more of this stuff back into my brain.
> 
Thank you!

—Rich

richgillam:
Markus has a much better fix for this problem in https://github.com/unicode-org/icu/pull/2052.  Closing my fix in favor of that one.

PR #2051, opened 2022-03-25
ICU-21965 fix utilities.jar module name

srl295:
https://unicode-org.atlassian.net/browse/ICU-21965

com.ibm.icu.utilities - otherwise the automatic module name is unusable

this is on maint-71

### DIFF
--- a/icu4j/build.xml
+++ b/icu4j/build.xml
@@ -1945,8 +1945,17 @@
         </javac>
 
         <mkdir dir="${cldr.util.out.dir}/lib"/>
+        <copy file="tools/build/manifest-utilities.stub" todir="${out.dir}">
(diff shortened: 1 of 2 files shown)


yumaoka (dismissed):
LGTM

markusicu (changes requested):
Is https://unicode-org.atlassian.net/browse/ICU-21425 the right ticket?
It does not mention the module name.
It does mention other things that I don't see done here, and that are likely too late for ICU 71.

srl295:
@markusicu  I could split the ticket.

srl295:
@markusicu  cloned ticket

markusicu (approved):
rslgtm

PR #2060, opened 2022-04-08
ICU-21984 Fix DateIntervalFormat.normalizeHourMetacharacters() so tha…

richgillam:
…t it doesn't require the hour and day-period

fields to appear in any particular order or position in the skeleton string.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21984

### DIFF
--- a/icu4c/source/i18n/dtitvfmt.cpp
+++ b/icu4c/source/i18n/dtitvfmt.cpp
@@ -966,23 +966,26 @@ DateIntervalFormat::normalizeHourMetacharacters(const UnicodeString& skeleton) c
     
     UChar hourMetachar = u'\0';
     UChar dayPeriodChar = u'\0';
-    int32_t metacharStart = 0;
-    int32_t metacharCount = 0;
+    int32_t hourFieldStart = 0;
+    int32_t hourFieldLength = 0;
+    int32_t dayPeriodStart = 0;
+    int32_t dayPeriodLength = 0;
     for (int32_t i = 0; i < result.length(); i++) {
         UChar c = result[i];
         if (c == LOW_J || c == CAP_J || c == CAP_C || c == LOW_H || c == CAP_H || c == LOW_K || c == CAP_K) {
             if (hourMetachar == u'\0') {
                 hourMetachar = c;
-                metacharStart = i;
+                hourFieldStart = i;
             }
-            ++metacharCount;
+            ++hourFieldLength;
         } else if (c == LOW_A || c == LOW_B || c == CAP_B) {
             if (dayPeriodChar == u'\0') {
                 dayPeriodChar = c;
+                dayPeriodStart = i;
             }
-            ++metacharCount;
+            ++dayPeriodLength;
         } else {
-            if (hourMetachar != u'\0') {
+            if (hourMetachar != u'\0' && dayPeriodChar != u'\0') {
                 break;
             }
         }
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/dtitvfmt.cpp
```diff
@@ -1023,29 +1026,23 @@ DateIntervalFormat::normalizeHourMetacharacters(const UnicodeString& skeleton) c
         }
         
         if (hourChar == CAP_H || hourChar == LOW_K) {
-            result.replace(metacharStart, metacharCount, hourChar);
+            result.replace(hourFieldStart, hourFieldLength, hourChar);
```
pedberg-icu, line 1029: OK, so I get that the above replace does not change the length of result, so the dayPeriodStart index is still valid for the remove below. But later on...

icu4c/source/i18n/dtitvfmt.cpp
```diff
@@ -1023,29 +1026,23 @@ DateIntervalFormat::normalizeHourMetacharacters(const UnicodeString& skeleton) c
...
+            } else if (dayPeriodLength >= 3 || hourFieldLength >= 3) {
+                newDayPeriodLength = 3;
+            } else {
+                newDayPeriodLength = 1;
+            }
+            for (int32_t i = 0; i < newDayPeriodLength; i++) {
+                hourAndDayPeriod.append(dayPeriodChar);
             }
-            result.replace(metacharStart, metacharCount, hourAndDayPeriod);
+            result.replace(hourFieldStart, hourFieldLength, hourAndDayPeriod);
```
pedberg-icu, line 1044: Doesn't the replace above here potentially change the length of result? So is the following remove valid, since dayPeriodStart might be after the section that was replaced above?

pedberg-icu (commented):
I had some questions about this...

macchiati (approved):
Nice!

icu4c/source/i18n/dtitvfmt.cpp
richgillam: Ugh.  You're absolutely right.  I'm an idiot.  Give me a little while, and I'll put up a new commit that addresses that.

richgillam:
>Also I assume that main is in fact the correct branch to target, since the ticket is marked for 72...

That was my assumption as well.  If we're going to put out an update to 71 with this fix in it, I assumed I'd be cherry-picking back to the maint branch.  But I don't know how important that is, especially since somebody (not me) put the ticket in 72.

pedberg-icu (approved):
Great, looks much better

pedberg-icu (approved):

FrankYFTang:
@richgillam  - could you help me out. I try to assess how important is this fix to decide I need to cherry pick this fix into chroumium or not. Could you tell me what will happen to date interval format if I do not cherry pick this PR but just stay with 71-1. What could happen ? what kind of buggy behavior will be there? I was wordering should I just ignore this PR and let chromium get the fix 6 months later with 72-1.

richgillam:
@FrankYFTang , I hate to say this, but you probably want to cherry-pick it.

On Java, you'll have the DateIntervalFormat sometimes fail to construct, even on skeleton strings that are perfectly valid, and on some of them that are actually in the CLDR data.  The issue is that if the skeleton string you pass to the constructor includes both hour and day period, but the day period doesn't immediately follow the hour (say, you have "Bh" or "hmsa"), it'll either choke on it and give you an error or mess up the pattern and give you weird-looking results.

In C++, you don't get the error and the effects of the problem are often benign, but you can get bad results.

FrankYFTang:
ok, great to know this. This is a regression, right?

richgillam:
Yes, totally a regression.

PR #2070, opened 2022-04-28
ICU-22005 Fix int32 overflow in FormattedStringBuilder

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22005

### DIFF
--- a/icu4c/source/i18n/formatted_string_builder.cpp
+++ b/icu4c/source/i18n/formatted_string_builder.cpp
@@ -6,6 +6,7 @@
 #if !UCONFIG_NO_FORMATTING
 
 #include "formatted_string_builder.h"
+#include "putilimp.h"
 #include "unicode/ustring.h"
 #include "unicode/utf16.h"
 #include "unicode/unum.h" // for UNumberFormatFields literals
@@ -197,6 +198,9 @@ FormattedStringBuilder::splice(int32_t startThis, int32_t endThis,  const Unicod
     int32_t thisLength = endThis - startThis;
     int32_t otherLength = endOther - startOther;
     int32_t count = otherLength - thisLength;
+    if (U_FAILURE(status)) {
+        return count;
+    }
     int32_t position;
     if (count > 0) {
         // Overall, chars need to be added.
@@ -221,6 +225,9 @@ int32_t FormattedStringBuilder::append(const FormattedStringBuilder &other, UErr
 
 int32_t
 FormattedStringBuilder::insert(int32_t index, const FormattedStringBuilder &other, UErrorCode &status) {
+    if (U_FAILURE(status)) {
+        return 0;
+    }
     if (this == &other) {
         status = U_ILLEGAL_ARGUMENT_ERROR;
         return 0;
@@ -255,12 +262,18 @@ int32_t FormattedStringBuilder::prepareForInsert(int32_t index, int32_t count, U
     U_ASSERT(index >= 0);
     U_ASSERT(index <= fLength);
     U_ASSERT(count >= 0);
+    U_ASSERT(fZero >= 0);
+    U_ASSERT(fLength >= 0);
+    U_ASSERT(getCapacity() - fZero >= fLength);
+    if (U_FAILURE(status)) {
+        return count;
+    }
     if (index == 0 && fZero - count >= 0) {
         // Append to start
         fZero -= count;
         fLength += count;
         return fZero;
-    } else if (index == fLength && fZero + fLength + count < getCapacity()) {
+    } else if (index == fLength && count <= getCapacity() - fZero - fLength) {
         // Append to end
         fLength += count;
         return fZero + fLength - count;
@@ -275,18 +288,26 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
     int32_t oldZero = fZero;
     char16_t *oldChars = getCharPtr();
     Field *oldFields = getFieldPtr();
-    if (fLength + count > oldCapacity) {
-        if ((fLength + count) > INT32_MAX / 2) {
-            // If we continue, then newCapacity will overflow int32_t in the next line.
+    int32_t newLength;
+    if (uprv_add32_overflow(fLength, count, &newLength)) {
+        status = U_INPUT_TOO_LONG_ERROR;
+        return -1;
+    }
+    int32_t newZero;
+    if (newLength > oldCapacity) {
+        if (newLength > INT32_MAX / 2) {
+            // We do not support more than 1G char16_t in this code because
+            // dealing with >2G *bytes* can cause subtle bugs.
             status = U_INPUT_TOO_LONG_ERROR;
             return -1;
         }
-        int32_t newCapacity = (fLength + count) * 2;
-        int32_t newZero = newCapacity / 2 - (fLength + count) / 2;
+        // Keep newCapacity also to at most 1G char16_t.
+        int32_t newCapacity = newLength * 2;
+        newZero = (newCapacity - newLength) / 2;
 
         // C++ note: malloc appears in two places: here and in the assignment operator.
-        auto newChars = static_cast<char16_t *> (uprv_malloc(sizeof(char16_t) * newCapacity));
-        auto newFields = static_cast<Field *>(uprv_malloc(sizeof(Field) * newCapacity));
+        auto newChars = static_cast<char16_t *> (uprv_malloc(sizeof(char16_t) * static_cast<size_t>(newCapacity)));
+        auto newFields = static_cast<Field *>(uprv_malloc(sizeof(Field) * static_cast<size_t>(newCapacity)));
         if (newChars == nullptr || newFields == nullptr) {
             uprv_free(newChars);
             uprv_free(newFields);
@@ -315,10 +336,8 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
         fChars.heap.capacity = newCapacity;
         fFields.heap.ptr = newFields;
         fFields.heap.capacity = newCapacity;
-        fZero = newZero;
-        fLength += count;
     } else {
-        int32_t newZero = oldCapacity / 2 - (fLength + count) / 2;
+        newZero = (oldCapacity - newLength) / 2;
 
         // C++ note: memmove is required because src and dest may overlap.
         // First copy the entire string to the location of the prefix, and then move the suffix
@@ -331,18 +350,20 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
         uprv_memmove2(oldFields + newZero + index + count,
                 oldFields + newZero + index,
                 sizeof(Field) * (fLength - index));
-
-        fZero = newZero;
-        fLength += count;
     }
-    U_ASSERT((fZero + index) >= 0);
+    fZero = newZero;
+    fLength = newLength;
     return fZero + index;
 }
 
 int32_t FormattedStringBuilder::remove(int32_t index, int32_t count) {
-    // TODO: Reset the heap here?  (If the string after removal can fit on stack?)
+     U_ASSERT(0 <= index);
+     U_ASSERT(index <= fLength);
+     U_ASSERT(count <= (fLength - index));
+     U_ASSERT(index <= getCapacity() - fZero);
+
     int32_t position = index + fZero;
-    U_ASSERT(position >= 0);
+    // TODO: Reset the heap here?  (If the string after removal can fit on stack?)
     uprv_memmove2(getCharPtr() + position,
             getCharPtr() + position + count,
             sizeof(char16_t) * (fLength - index - count));
--- a/icu4c/source/test/intltest/formatted_string_builder_test.cpp
+++ b/icu4c/source/test/intltest/formatted_string_builder_test.cpp
@@ -22,6 +22,7 @@ class FormattedStringBuilderTest : public IntlTest {
     void testFields();
     void testUnlimitedCapacity();
     void testCodePoints();
+    void testInsertOverflow();
 
     void runIndexedTest(int32_t index, UBool exec, const char *&name, char *par = 0) override;
 
@@ -50,6 +51,7 @@ void FormattedStringBuilderTest::runIndexedTest(int32_t index, UBool exec, const
         TESTCASE_AUTO(testFields);
         TESTCASE_AUTO(testUnlimitedCapacity);
         TESTCASE_AUTO(testCodePoints);
+        TESTCASE_AUTO(testInsertOverflow);
     TESTCASE_AUTO_END;
 }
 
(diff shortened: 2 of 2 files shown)


FrankYFTang:
The unit test take 60.8 sec to run in my computer. Most of the time is spent on the setting up.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -260,6 +260,11 @@ int32_t FormattedStringBuilder::prepareForInsert(int32_t index, int32_t count, U
         fZero -= count;
         fLength += count;
         return fZero;
+    } else if (fZero + fLength + count < 0 || fLength + count < 0 || fZero + fLength < 0) {
```
sffc, line 263: Suggestion: use `uprv_add32_overflow` and related methods from putilimp.h to check for integer overflow, rather than checking for negative values, which is not bulletproof

sffc (dismissed):

FrankYFTang:
PTAL

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -257,16 +258,37 @@ int32_t FormattedStringBuilder::prepareForInsert(int32_t index, int32_t count, U
...
+        int32_t newLength;
+        // Add and return U_INPUT_TOO_LONG_ERROR if overflow int32_t.
+        if (uprv_add32_overflow(fLength, count, &newLength)) {
+          status = U_INPUT_TOO_LONG_ERROR;
+          return count;
+        }
         fZero -= count;
-        fLength += count;
+        fLength = newLength;
         return fZero;
```
markusicu, line 269: You should not need to change this branch at all, because the condition `fZero - count >= 0` means that count is small (count <= fZero), and the new contents fits into the existing early part of the buffer. Please revert this part.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -257,16 +258,37 @@ int32_t FormattedStringBuilder::prepareForInsert(int32_t index, int32_t count, U
...
-    } else if (index == fLength && fZero + fLength + count < getCapacity()) {
-        // Append to end
-        fLength += count;
-        return fZero + fLength - count;
     } else {
+      int32_t position, requiredCapacity;
+      // Add and return U_INPUT_TOO_LONG_ERROR if overflow int32_t.
+      if (uprv_add32_overflow(fLength, fZero, &position) ||
```
markusicu, line 273: No need for a fancy check for `uprv_add32_overflow(fLength, fZero, &position)` because you would have caught an overflow there in a previous call. Just use `int32_t position = fZero + fLength;`

More importantly, see the comment on old lines 263..266 -- I think you can keep this whole case simple.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -257,16 +258,37 @@ int32_t FormattedStringBuilder::prepareForInsert(int32_t index, int32_t count, U
...
+          return count;
+        }
         fZero -= count;
-        fLength += count;
+        fLength = newLength;
         return fZero;
-    } else if (index == fLength && fZero + fLength + count < getCapacity()) {
-        // Append to end
-        fLength += count;
-        return fZero + fLength - count;
```
markusicu, line 266: I would preserve this append-to-end optimization as well, by refactoring the test slightly to simply avoid the addition:
```
    } else if (index == fLength && count <= getCapacity() - (fZero + fLength)) {
        // Append to end
        fLength += count;
        return fZero + fLength - count;
```

Only the need-to-grow-the-buffer "else" then needs more work.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -275,14 +297,19 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
     int32_t oldZero = fZero;
     char16_t *oldChars = getCharPtr();
     Field *oldFields = getFieldPtr();
+    int32_t newLength;
+    if (uprv_add32_overflow(fLength, count, &newLength)) {
+        status = U_INPUT_TOO_LONG_ERROR;
+        return -1;
+    }
     if (fLength + count > oldCapacity) {
```
markusicu, line 305:
```suggestion
    if (newLength > oldCapacity) {
```

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -275,14 +297,19 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
-        if ((fLength + count) > INT32_MAX / 2) {
-            // If we continue, then newCapacity will overflow int32_t in the next line.
+        int32_t newCapacity;
+        //  newCapacity = (fLength + count) * 2 w/ overflow checking.
+        if (uprv_add32_overflow(newLength, newLength, &newCapacity)) {
             status = U_INPUT_TOO_LONG_ERROR;
             return -1;
         }
```
markusicu, line 311: This is too strict. Both the old and new code fail when twice the newLength overflows, even though we can satisfy the current request with just newLength alone, which we now know didn't overflow. The exponential over-allocation is just an optimization, to reduce the number of memory allocations.
Suggestion:
```suggestion
        int32_t newCapacity;
        if (newLength <= INT32_MAX / 2) {
            newCapacity = newLength * 2;
        } else if (newLength <= INT32_MAX - 1000) {
            newCapacity = newLength + 1000;
        } else {
            newCapacity = INT32_MAX;
        }
```

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -275,14 +297,19 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
-        int32_t newCapacity = (fLength + count) * 2;
-        int32_t newZero = newCapacity / 2 - (fLength + count) / 2;
+        int32_t newZero = newLength / 2;
```
markusicu, line 312:
```suggestion
        int32_t newZero = (newCapacity - newLength) / 2;
```
... to center the contents in the buffer, as before.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -316,9 +343,9 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
         fFields.heap.ptr = newFields;
         fFields.heap.capacity = newCapacity;
         fZero = newZero;
-        fLength += count;
+        fLength = newLength;
```
markusicu, line 346: Same two lines in this branch and in the "else" branch. I suggest you combine and move them to after the if-else.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -333,16 +360,23 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
+    return result;
 }
 
-int32_t FormattedStringBuilder::remove(int32_t index, int32_t count) {
+int32_t FormattedStringBuilder::remove(int32_t index, int32_t count, UErrorCode &status) {
     // TODO: Reset the heap here?  (If the string after removal can fit on stack?)
-    int32_t position = index + fZero;
-    U_ASSERT(position >= 0);
+    int32_t position;
+    if (uprv_add32_overflow(fZero, index, &position)) {
```
markusicu, line 376: This seems unnecessary. Instead I would
```
    U_ASSERT(0 <= index && index <= fLength);
    U_ASSERT(count <= (fLength - index));
```
... and then you need not add the UErrorCode.

icu4c/source/test/intltest/formatted_string_builder_test.cpp
```diff
@@ -308,6 +310,45 @@ void FormattedStringBuilderTest::testCodePoints() {
     assertEquals("Code point count is 2", 2, nsb.codePointCount());
 }
 
+void FormattedStringBuilderTest::testInsertOverflow() {
+  if (quick) return;
```
markusicu, line 314: Please consistently indent by 4 spaces.

markusicu (changes requested):

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: we only know count is at most fZero per your logic, right? How could we be sure fLength + fZero won't overflow int32_t ?

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: how could we be sure that (fZero + fLength) does not overflow int32_t here?

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: how can we be sure fLength + fZero won't overflow here?

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: how can we be sure index + fZero won't overflow int32_t ?

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -333,16 +360,23 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
+        fLength = newLength;
+    }
+    int32_t result;
+    if (uprv_add32_overflow(fZero, index, &result)) {
+        status = U_INPUT_TOO_LONG_ERROR;
+        return -1;
     }
-    U_ASSERT((fZero + index) >= 0);
-    return fZero + index;
+    return result;
```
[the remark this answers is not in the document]
FrankYFTang: done

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: done

icu4c/source/test/intltest/formatted_string_builder_test.cpp
FrankYFTang: done

icu4c/source/test/intltest/formatted_string_builder_test.cpp
```diff
@@ -308,6 +310,45 @@ void FormattedStringBuilderTest::testCodePoints() {
...
+  // Setup the test fixture in sb, sb2, ustr.
+  UErrorCode status = U_ZERO_ERROR;
+  FormattedStringBuilder sb;
+  int32_t data_length = INT32_MAX / 2;
+  UnicodeString ustr(data_length, 'a', data_length);
```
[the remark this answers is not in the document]
FrankYFTang: done

FrankYFTang (commented):
PTAL

FrankYFTang:
ok, somehow we have segment fault on calling uprv_memcpy2 inside that functione while the third argument size_t len is more than the value of 31 bits can hold 2147483648. we may need to add a work around to check and return error before calling uprv_memcpy2

FrankYFTang:
With the current PR, we observe a crash in "./intltest format/ListFormatterTest/TestInt32Overflow -e"
under the follwoing line. The value of the third arg is > INT32_MAX
```
        uprv_memcpy2(newChars + newZero, oldChars + oldZero, sizeof(char16_t) * index);
```

aheninger:
> ok, somehow we have segment fault 

malloc() is probably failing in some UnicodeString creation or operation, leaving the string in a bogus state with no buffer.

Some time ago I tried to write some UText tests for large texts; it proved to be enough of a pain that I gave up.
On 32 bit platforms you'll never be able to allocate strings any where near the required size. I don't know whether we still support any 32 bit environments, though.

In 64 bit environments, the large memory allocations may or may not work, depending on specific platform configurations and limits. When they fail, you may get a clean NULL return from malloc, or you may get some kind of unrecoverable run-time debug exception or assertion. 

Probably the best you can do is add some sort of platform #ifdef, restricting the test to running somewhere that you know works.

icu4c/source/i18n/formatted_string_builder.cpp
FrankYFTang: somehow this relax of memory allocation size (for the case newLength > INT32_MAX / 2) will cause crash inside the first uprv_memcpy2 bewlow while running "format/ListFormatterTest/TestInt32Overflow -e" test. I have to revert back that change to the prior restriction to avoid create new problem.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -275,14 +288,26 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
+        // the first uprv_memcpy2() below when running intltest
+        // format/ListFormatterTest/TestInt32Overflow -e
+        if (newLength > INT32_MAX / 2) {
             status = U_INPUT_TOO_LONG_ERROR;
             return -1;
         }
-        int32_t newCapacity = (fLength + count) * 2;
-        int32_t newZero = newCapacity / 2 - (fLength + count) / 2;
+        newCapacity = newLength * 2;
+        newZero = newLength / 2;
```
markusicu, line 310: I suspect that we get into trouble when computing the number of bytes for the newCapacity or (fZero + newLength) number of char16_t. In those computations, it should work to cast the int32_t variables to size_t before multiplying. size_t is unsigned, and when we double something that's at most INT32_MAX *after casting it to size_t* then it fits into even a 32-bit size_t.

So I suggest that you try once more with the newCapacity code in your previous commit but add casts to size_t before multiplying int32_t values by 2.

For example, a couple of lines below:
```
        auto newChars = static_cast<char16_t *> (uprv_malloc(sizeof(char16_t) * newCapacity));
        auto newFields = static_cast<Field *>(uprv_malloc(sizeof(Field) * newCapacity));
```
-->
```
        auto newChars = static_cast<char16_t *> (uprv_malloc(sizeof(char16_t) * (size_t)newCapacity));
        auto newFields = static_cast<Field *>(uprv_malloc(sizeof(Field) * (size_t)newCapacity));
```
I would look at each uprv_memcpy2() and uprv_malloc().

markusicu: Second thought: Maybe my suggestion is too much work. We could just say that we don't support more than 1G char16_t in this code because dealing with >2G *bytes* can cause subtle bugs. However, your current code is not quite right for that either.

For that:
```
    if (newLength > oldCapacity) {
        if (newLength > INT32_MAX / 2) {
            // We do not support more than 1G char16_t in this code because
            // dealing with >2G *bytes* can cause subtle bugs.
            status = U_INPUT_TOO_LONG_ERROR;
            return -1;
        }
        int32_t newCapacity;
        // Keep newCapacity also to at most 1G char16_t.
        if (newLength <= INT32_MAX / 4) {
            newCapacity = newLength * 2;
        } else if (newLength <= (INT32_MAX / 2) - 1000) {
            newCapacity = newLength + 1000;
        } else {
            newCapacity = INT32_MAX / 2;
        }
```

FrankYFTang: this change cause the test setup to fail because the allocation size change. so I prefer not to take it otherwise, I will need to spend some more time to figure out how to set up the test to fail w/o the fix but pass w/ the fix.

icu4c/source/i18n/formatted_string_builder.cpp
```diff
@@ -295,23 +295,19 @@ int32_t FormattedStringBuilder::prepareForInsertHelper(int32_t index, int32_t co
...
-        // format/ListFormatterTest/TestInt32Overflow -e
         if (newLength > INT32_MAX / 2) {
+            // We do not support more than 1G char16_t in this code because
+            // dealing with >2G *bytes* can cause subtle bugs.
             status = U_INPUT_TOO_LONG_ERROR;
             return -1;
         }
-        newCapacity = newLength * 2;
-        newZero = newLength / 2;
+        // Keep newCapacity also to at most 1G char16_t.
```
markusicu, line 304: FYI: The code does not actually do so right now... fodder for a follow-up ticket.

markusicu (commented):
lgtm pse squash

markusicu (approved):
lgtm tnx!

PR #2071, opened 2022-04-28
ICU-22006 icupkg: %%ALIAS & %%Parent do not need truncation parent

markusicu:
When the icupkg tool checks for dependencies from one resource bundle on other bundles, it also checks for the truncation parent bundle.

However, when the bundle has a top-level %%ALIAS string, then the alias target *replaces* the current bundle; and when there is a top-level %%Parent string, then the explicit parent bundle is used. In either of these cases, the truncation parent bundle is ignored and is thus not a required dependency.

The dependency checking code is in the toolutil library, in pkgitems.cpp.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22006

### DIFF
--- a/icu4c/source/tools/toolutil/pkgitems.cpp
+++ b/icu4c/source/tools/toolutil/pkgitems.cpp
@@ -277,38 +277,40 @@ checkAlias(const char *itemName,
  * Enumerate one resource item and its children and extract dependencies from
  * aliases.
  */
-static void
+static UBool
 ures_enumDependencies(const char *itemName,
                       const ResourceData *pResData,
                       Resource res, const char *inKey, const char *parentKey, int32_t depth,
                       CheckDependency check, void *context,
                       Package *pkg,
                       UErrorCode *pErrorCode) {
+    UBool doCheckParent = TRUE;  // always remains TRUE if depth>1
     switch(res_getPublicType(res)) {
     case URES_STRING:
-        {
-            UBool useResSuffix = TRUE;
-            // Check for %%ALIAS
-            if(depth==1 && inKey!=NULL) {
-                if(0!=strcmp(inKey, "%%ALIAS")) {
-                    break;
-                }
-            }
-            // Check for %%DEPENDENCY
-            else if(depth==2 && parentKey!=NULL) {
-                if(0!=strcmp(parentKey, "%%DEPENDENCY")) {
-                    break;
-                }
-                useResSuffix = FALSE;
-            } else {
-                // we ignore all other strings
-                break;
-            }
-            int32_t length;
+        if(depth==1 && inKey!=NULL &&
+                (0==strcmp(inKey, "%%ALIAS") || 0==strcmp(inKey, "%%Parent"))) {
+            // Top-level %%ALIAS string:
+            //   The alias resource bundle will be used instead of this one.
+            // Top-level %%Parent string:
+            //   We use this bundle as well as the explicit parent bundle.
+            // Either way, the truncation parent is ignored.
+            doCheckParent = FALSE;
             // No tracing: build tool
+            int32_t length;
             const UChar *alias=res_getStringNoTrace(pResData, res, &length);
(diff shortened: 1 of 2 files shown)


pedberg-icu (approved):

markusicu:
Hm, the "GHA CI / lstm-icu4j-build-and-test" bot fails building the testdata.
```
make[1]: Entering directory '/home/runner/work/icu/icu/icu4c/source/test/testdata'
mkdir -p ../../data/out/icu4j/com/ibm/icu/dev/data/testdata
LD_LIBRARY_PATH=../../lib:../../stubdata:../../tools/ctestfw:$LD_LIBRARY_PATH  ../../bin/icupkg ../../test/testdata/out/testdata.dat -r test.icu -x '*' -tb -d ../../data/out/icu4j/com/ibm/icu/dev/data/testdata
Item structLocale.res depends on missing item .res
make[1]: *** [Makefile:165: ../../data/out/icu4j/testdata.jar] Error 2
make[1]: Leaving directory '/home/runner/work/icu/icu/icu4c/source/test/testdata'
make: *** [Makefile:369: icu4j-data-install] Error 2
```
@FrankYFTang is there something special about the LSTM test setup for building its testdata?

icu4c/source/tools/toolutil/pkgitems.cpp
```diff
@@ -277,38 +277,40 @@ checkAlias(const char *itemName,
...
-                if(0!=strcmp(parentKey, "%%DEPENDENCY")) {
-                    break;
-                }
-                useResSuffix = FALSE;
-            } else {
-                // we ignore all other strings
-                break;
-            }
+        if(depth==1 && inKey!=NULL &&
+                (0==strcmp(inKey, "%%ALIAS") || 0==strcmp(inKey, "%%Parent"))) {
```
sffc, line 291: Question: Is the change you made equivalent to if you had simply changed line 293 to check for either ALIAS or PARENT?

The rest of the refactoring of this if statement looks fine, but I want to make sure I'm not missing something

FrankYFTang:
could you state the IMPACT of this PR. I try to understand should I cherry pick that into chromium
1. Would this fix some bug in 71-1?
2. Would this PR cause some dat size reduction? (I spot we added some more top level %%ALIAS between 70-1 to 71-1.

markusicu:
> icupkg could change "" to "root" here, or it could ignore the %%Parent because of the nofallback, but I don't think we need to go out of our way to come up with what might be a reasonable way to handle this when there seems to be no good reason for the testdata to have this parent spec. I will remove it.

Removing it broke `cintltst /tsutil/cldrtest/TestLocaleStructure` as I should have expected.
Instead, I changed it to `%%Parent("root")` which is still bogus in a nofallback bundle, but it satisfies both cintltst and icupkg.

PTAL

sffc (approved):

markusicu:
> 1. Would this fix some bug in 71-1?

No. This bug has been here for a long time.
It's a bug in an offline data build tool, not in a runtime library.

The bug manifests itself when people try to build minimal data for, say, "hi_Latn", excluding "hi" which should not be necessary. Or data including "iw_IL" but not "iw".

> 2. Would this PR cause some dat size reduction? (I spot we added some more top level %%ALIAS between 70-1 to 71-1.

Not by itself. It does not change at all how data is packaged, only what dependencies it complains about.

With this PR, you will no longer have to include certain bundles in certain cases when they are unnecessary, as noted above. So if you have been including seemingly-unnecessary data files, you might be able to stop doing so now and reduce the data size.

However, also, icupkg now checks for `%%Parent` which it had ignored before. So if you were building a .dat package with "hi_Latn" and "hi" but not "en_IN" and its parents, then you would now get an error. Which is good, because you were omitting the explicitly specified parent bundle. If you wanted to proceed anyway, you could turn off dependency checking via the `--ignore-deps` option.

PR #2075, opened 2022-05-02
ICU-21960 Add missing "const" to kAttributeKey

agrieve:
https://unicode-org.atlassian.net/browse/ICU-21960 minor warning fixes

This variable was flagged by a chromium check that looks for variables named like constants that end up in the `.data` ELF section (rather than in `.rodata`).


markusicu:
Please prefix the commit with `ICU-21960 `
See https://unicode-org.atlassian.net/browse/ICU-21960 for minor warnings fixes.

agrieve:
Done

markusicu (approved):
lgtm tnx

PR #2077, opened 2022-05-03
ICU-22011 Fix leap years algorithm for Persian calendar

Sinakhx:
It's now using the arithmetic computational calendar system which is by far the most accurate algorithm for calculating Persian/Jalali leap years

external link: https://fa.wikipedia.org/wiki/%DA%AF%D8%A7%D9%87%E2%80%8C%D8%B4%D9%85%D8%A7%D8%B1%DB%8C_%D9%87%D8%AC%D8%B1%DB%8C_%D8%AE%D9%88%D8%B1%D8%B4%DB%8C%D8%AF%DB%8C_%D8%AD%D8%B3%D8%A7%D8%A8%DB%8C#%DA%AF%D8%A7%D9%87%D8%B4%D9%85%D8%A7%D8%B1%DB%8C_%D8%AD%D8%B3%D8%A7%D8%A8%DB%8C_%D8%B1%D8%A7%DB%8C%D8%A7%D9%86%D9%87%E2%80%8C%D8%A7%DB%8C

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22011

### DIFF
--- a/icu4c/source/i18n/persncal.cpp
+++ b/icu4c/source/i18n/persncal.cpp
@@ -109,9 +109,10 @@ int32_t PersianCalendar::handleGetLimit(UCalendarDateFields field, ELimitType li
  */
 UBool PersianCalendar::isLeapYear(int32_t year)
 {
-    int32_t remainder;
-    ClockMath::floorDivide(25 * year + 11, 33, remainder);
-    return (remainder < 8);
+    double cte = 683 / 2820;
+    double remainder = (year + 2346) * cte;
+    remainder -= ClockMath::floorDivide(remainder, 1);
+    return (remainder < cte);
 }
     
 /**


roozbehp:
Strong opposition. The 2820-year cycle is less accurate than the existing implementation in ICU. Will provide more details if necessary.

Sinakhx:
> Strong opposition. The 2820-year cycle is less accurate than the existing implementation in ICU. Will provide more details if necessary.

Please follow the links to see this is not the well-known simplistic 2820-year cycle approach (although it's using 2820 as coefficient). The existing implementation in icu doesn't confirm the results from any jalali libraries.
It would be nice if you could provide references to your claim (as the most accurate approach)

roozbehp:
In short:
1. Iranian law of 1925 clearly specifies that a "true solar year" should be used, which means astronomical calculations are needed.
2. The location used for computing the true solar noon which is a part of the calculations needed for the true solar noon is not specified in the law. Some use Tehran, and I've heard that the Iranian calendar authorities use some location on the 52.5°E meridian (the meridian that sets the Iranian standard timezone). This lack of exact locale makes the calendar ambiguous and underdefined for some years in the future, first of which happens in 2091 CE, 69 years from now, when the vernal equinox happens after the true noon at 52.5°E but before the true noon in Tehran.
3. The book Calendrical Calculations: The Ultimate Edition, which is the best known English language reference for various calendars, has compared the 33-year rule's results to actual astronomical calculations (which is the law). It says it agrees with the astronomical calendar until 1468 AP, 67 years from now. This is on page 265.
4. The same book claims that the 2820-year rule will generate the wrong dates starting with 1404 AP, just three years from now.

So in short, the 33-year rule works correctly until about the same time the Persian calendar is well-defined, while the 2820-year rule will stop working in just a few years.

As for other "jalali" libraries, please mention which ones you're taking about. They may have bugs. It's best if you post the list of years in the future where your proposed algorithm results in a different start of the year than the 33-year algorithm, so we can compare which one is doing better against the astronomical calculations.

Sinakhx:
> 3. The book Calendrical Calculations: The Ultimate Edition, which is the best known English language reference for various calendars, has compared the 33-year rule's results to actual astronomical calculations (which is the law). It says it agrees with the astronomical calendar until 1468 AP, 67 years from now. This is on page 265.

It's interesting to see how different references still use contradictory methods. The following link is from the Geophysics center of Tehran University which is officially considered as calendarial reference in Iran:
https://calendar.ut.ac.ir/Fa/News/Data/Doc/KabiseShamsi1206-1498-new.pdf

As you can see, the Jalali year of 1403 is officially considered as a leap year in Iran (which is the same thing that [time.ir](https://www.time.ir/) shows to its users and also what the famous [date-fns-jalali](https://github.com/date-fns-jalali/date-fns-jalali) & other jalali libraries are returning as their results). This complies with the algorithm provided in this pull-request but not with the current 33-year-rule algorithm.

I guess the decision on choosing a single single source of truth for Persian calendar dates should be based on the Iranian calendar authorities and not a general book which might be outdated in some cases.

roozbehp:
> As you can see, the Jalali year of 1403 is officially considered as a leap year in Iran [...]. This complies with the algorithm provided in this pull-request but not with the current 33-year-rule algorithm.

Looks like we have a disagreement over how basic math works then! According to the 33-year calendar currently implemented in ICU, 1403 AP *is* leap!

icu4c/source/i18n/persncal.cpp
```diff
@@ -109,9 +109,10 @@ int32_t PersianCalendar::handleGetLimit(UCalendarDateFields field, ELimitType li
  */
 UBool PersianCalendar::isLeapYear(int32_t year)
 {
-    int32_t remainder;
-    ClockMath::floorDivide(25 * year + 11, 33, remainder);
-    return (remainder < 8);
+    int32_t cte = 683 / 2820;
```
roozbehp, line 112: Apart from the problems with the algorithm, this code doesn't do what it tries to do. For example, `cte` is set to zero here, since it's an `int32_t`.

Sinakhx: Thanks for mentioning this problem. I don't know C++ well, so I guess changing it to `float32_t` might fix this problem. Please let me know if I'm right.

Sinakhx:
> Looks like we have a disagreement over how basic math works then! According to the 33-year calendar currently implemented in ICU, 1403 AP is leap!

Sorry, my playground code was buggy. I tried to fix it.

> 4\. The same book claims that the 2820-year rule will generate the wrong dates starting with 1403 AP, just three years from now.

As far as I can see neither the current implementation, nor the one in my pull-request (considering an upcoming fix of `int32_t`  to `float32_t`) is caculating wrong dates for 1403 AP. Please let me know if I'm still bad at math.
Both algorithms seem to have quite consistent results for many years from now (not just 3 years).

the current one is:
```
UBool PersianCalendar::isLeapYear(int32_t year)
{
    int32_t remainder;
    ClockMath::floorDivide(25 * year + 11, 33, remainder);
    return (remainder < 8);
}
```

while the one I'm offering is:
```
UBool PersianCalendar::isLeapYear(int32_t year)
{
  float32_t cte = 683 / 2820;
  float32_t remainder = (year + 2346) * cte;
  remainder -= ClockMath::floorDivide(remainder, 1);
  return (remainder < cte);
}
```

with a simple comparison of leap years calculations we can see both are quite consistent for the upcoming years (& not just in 3 years from now):

|        year (AP)         |  1400  |  1401  |  1402  |  1403  |  1404  |  1405  |  1406  |  1407  |
|        ---          |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| current algorithm |  false  |  false  |  false  |  **true**  |  false  |  false  |  false  |  false  |
| new algorithm      |  false  |  false  |  false  |  **true**  |  false  |  false  |  false  |  false  |

Actually the first inconsistency between the two algorithms happens in 1435 AP (around 34 years from now).

icu4c/source/i18n/persncal.cpp
roozbehp: This would still not work, since the division of 683 over 2820 is done as integers. You need to change 683 to 683.0 or something like that. Also there's no float32_t in C.

roozbehp:
> Actually the first inconsistency between the two algorithms happens in 1435 AP (around 34 years from now).

Ok, we are making some progress here. According to the calendar.ut.ac.ir source you referred us to, 1434 AP is not leap, 1435 AP is not leap, 1436 AP is leap, and 1437 AP is not leap. This matches ICU's current implementation.

What does your proposed algorithm say for these four years? (I think it says only 1437 AP is leap, which would be wrong, but I'll let you do the math yourself. I may have been misunderstanding your proposed algorithm.)

Sinakhx:
> It's interesting to see how different references still use contradictory methods. The following link is from the Geophysics center of Tehran University which is officially considered as calendarial reference in Iran:
> https://calendar.ut.ac.ir/Fa/News/Data/Doc/KabiseShamsi1206-1498-new.pdf

I'm going to close this issue because I've just noticed the algorithm that I have used doesn't consider 1436 AP as a leap year, but the Geophysics center of Tehran University has considered it as leap!

@roozbehp Thanks for your kind consideration & informative discussion & guidance.

PR #2078, opened 2022-05-06
ICU-22015 Add U_SUPPORTS_ATOMICS

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22015

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -347,6 +347,18 @@
 #   define U_HAVE_INTTYPES_H U_HAVE_STDINT_H
 #endif
 
+/**
+ * \def U_SUPPORTS_ATOMICS
+ * Defines whether the platform supports C++ atomics.
+ * Set this to 0 to disable certain code that requires atomics.
+ * @internal
+ */
+#ifdef U_SUPPORTS_ATOMICS
+    /* Use the predefined value. */
+#else
+#   define U_SUPPORTS_ATOMICS 1
+#endif
+
 /*===========================================================================*/
 /** @{ Compiler and environment features                                     */
 /*===========================================================================*/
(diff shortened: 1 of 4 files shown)


icu4c/source/common/unicode/platform.h
```diff
@@ -347,6 +347,18 @@
 #   define U_HAVE_INTTYPES_H U_HAVE_STDINT_H
 #endif
 
+/**
+ * \def U_SUPPORTS_ATOMICS
+ * Defines whether the platform supports C++ atomics.
+ * Set this to 0 to disable certain code that requires atomics.
```
markusicu, line 353: I would add another sentence like
Only a small part of the ICU code currently avoids C++ atomics based on this setting.

icu4c/source/common/unicode/platform.h
```diff
@@ -347,6 +347,18 @@
...
+ * @internal
```
sffc, line 354: Should it be `@internal`? It seems that many of the other flags are `@internal`, even if the have a `U_` prefix (not just `UPRV_`)

markusicu: good idea, especially since it's not used everywhere so you can't build all of ICU with it

markusicu:
FYI @aheninger

markusicu:
https://github.com/unicode-org/icu/pull/2080 might be a simpler solution for what Shane is trying to do.

aheninger:
I pretty strongly prefer just doing #2080, and withdrawing this PR.

There's no clear definition of what ICU services U_SUPPORTS_ATOMICS would enable or disable, and trying to tie it down would be hard.

sffc:
Sounds good to me!

PR #2086, opened 2022-05-10
ICU-22023 Fix Calendar::get() return out of bound value and SimpleDateTime::format assert while TimeZone is "UTC" and value is -1e-9

FrankYFTang:
This PR fix several issures related to Calendar::get() return value out of bound
defined by Calendar::getMinimum() and Calendar::getMaximum() while
the Date is a tiny negative value ( between 0 and -1e-9) and the TimeZone is
 "UTC". It also fix the issue while Calendar::get() return error under the same 
condition.

The issue is caused by a tiny value of x (0 .. -1e-9) and a huge value of y (> 1e8)
in the form of 
int32_t w = (int32_t)(x + y); while both x and y are double.
This produce a 1-off rounding error.
We fix it by change it by wrapping x with uprv_floor. This appeared in several places 
related to timezone code and we fix them all in this PR

We also add U_ASSERT to check the bound for most fields except fields of year which
has not real hard constrain of range.

We add two new test cases to demostrate the problem w/o the fix.
intltest format/DateFormatTest/Test22023_UTCWithMinusZero
intltest format/CalendarRegressionTest/TestUTCWrongAMPM22023

We also notice there are unit test to [test ZONE_OFFSET with 30 hours](https://github.com/unicode-org/icu/blob/369357aaa7ff13553ac9943de31d2e69f9313de4/icu4c/source/test/intltest/tzfmttst.cpp#L913) and [DST_OFFSET with -1 hour](https://github.com/unicode-org/icu/blob/369357aaa7ff13553ac9943de31d2e69f9313de4/icu4c/source/test/intltest/tzregts.cpp#L1230) so we change the limit to ensure it will not assert under those test condition since both thes two  bound should not impact the caller.

We also found in [time zone “Antarctica/Troll” the DST is 2 hours (7200s) ](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/misc/zoneinfo64.txt#:~:text=Antarctica/Troll)  so we also change the maximum DST_OFFSET
from 1 hour to 2 hours. (see https://unicode-org.atlassian.net/browse/ICU-22031)
 
Also change the function signature of the three parameters floorDivision so the output parameter
is pointer instead of ref to make it clear from the calling code that value could be changed instead of
implicit side effect via reference passing.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22023

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -639,8 +639,8 @@ static const int32_t kCalendarLimits[UCAL_FIELD_COUNT][4] = {
     {           0,            0,            59,            59  }, // MINUTE
     {           0,            0,            59,            59  }, // SECOND
     {           0,            0,           999,           999  }, // MILLISECOND
-    {-12*kOneHour, -12*kOneHour,   12*kOneHour,   15*kOneHour  }, // ZONE_OFFSET
-    {           0,            0,    1*kOneHour,    1*kOneHour  }, // DST_OFFSET
+    {-12*kOneHour, -12*kOneHour,   12*kOneHour,   30*kOneHour  }, // ZONE_OFFSET
+    { -1*kOneHour,  -1*kOneHour,    2*kOneHour,    2*kOneHour  }, // DST_OFFSET
     {/*N/A*/-1,       /*N/A*/-1,     /*N/A*/-1,       /*N/A*/-1}, // YEAR_WOY
     {           1,            1,             7,             7  }, // DOW_LOCAL
     {/*N/A*/-1,       /*N/A*/-1,     /*N/A*/-1,       /*N/A*/-1}, // EXTENDED_YEAR
@@ -1537,7 +1537,8 @@ void Calendar::computeFields(UErrorCode &ec)
     // JULIAN_DAY field and also removes some inelegant code. - Liu
     // 11/6/00
 
-    int32_t days =  (int32_t)ClockMath::floorDivide(localMillis, (double)kOneDay);
+    int32_t millisInDay;
+    int32_t days = ClockMath::floorDivide(localMillis, kOneDay, &millisInDay);
 
     internalSet(UCAL_JULIAN_DAY,days + kEpochStartAsJulianDay);
 
@@ -1561,19 +1562,47 @@ void Calendar::computeFields(UErrorCode &ec)
     // Compute time-related fields.  These are independent of the date and
     // of the subclass algorithm.  They depend only on the local zone
     // wall milliseconds in day.
-    int32_t millisInDay =  (int32_t) (localMillis - (days * kOneDay));
+
     fFields[UCAL_MILLISECONDS_IN_DAY] = millisInDay;
+    U_ASSERT(getMinimum(UCAL_MILLISECONDS_IN_DAY) <=
+             fFields[UCAL_MILLISECONDS_IN_DAY]);
+    U_ASSERT(fFields[UCAL_MILLISECONDS_IN_DAY] <=
+             getMaximum(UCAL_MILLISECONDS_IN_DAY));
+
     fFields[UCAL_MILLISECOND] = millisInDay % 1000;
+    U_ASSERT(getMinimum(UCAL_MILLISECOND) <= fFields[UCAL_MILLISECOND]);
+    U_ASSERT(fFields[UCAL_MILLISECOND] <= getMaximum(UCAL_MILLISECOND));
+
     millisInDay /= 1000;
     fFields[UCAL_SECOND] = millisInDay % 60;
+    U_ASSERT(getMinimum(UCAL_SECOND) <= fFields[UCAL_SECOND]);
+    U_ASSERT(fFields[UCAL_SECOND] <= getMaximum(UCAL_SECOND));
+
     millisInDay /= 60;
     fFields[UCAL_MINUTE] = millisInDay % 60;
+    U_ASSERT(getMinimum(UCAL_MINUTE) <= fFields[UCAL_MINUTE]);
+    U_ASSERT(fFields[UCAL_MINUTE] <= getMaximum(UCAL_MINUTE));
+
     millisInDay /= 60;
     fFields[UCAL_HOUR_OF_DAY] = millisInDay;
+    U_ASSERT(getMinimum(UCAL_HOUR_OF_DAY) <= fFields[UCAL_HOUR_OF_DAY]);
+    U_ASSERT(fFields[UCAL_HOUR_OF_DAY] <= getMaximum(UCAL_HOUR_OF_DAY));
+
     fFields[UCAL_AM_PM] = millisInDay / 12; // Assume AM == 0
+    U_ASSERT(getMinimum(UCAL_AM_PM) <= fFields[UCAL_AM_PM]);
+    U_ASSERT(fFields[UCAL_AM_PM] <= getMaximum(UCAL_AM_PM));
+
     fFields[UCAL_HOUR] = millisInDay % 12;
+    U_ASSERT(getMinimum(UCAL_HOUR) <= fFields[UCAL_HOUR]);
+    U_ASSERT(fFields[UCAL_HOUR] <= getMaximum(UCAL_HOUR));
+
     fFields[UCAL_ZONE_OFFSET] = rawOffset;
+    U_ASSERT(getMinimum(UCAL_ZONE_OFFSET) <= fFields[UCAL_ZONE_OFFSET]);
+    U_ASSERT(fFields[UCAL_ZONE_OFFSET] <= getMaximum(UCAL_ZONE_OFFSET));
+
     fFields[UCAL_DST_OFFSET] = dstOffset;
+    U_ASSERT(getMinimum(UCAL_DST_OFFSET) <= fFields[UCAL_DST_OFFSET]);
+    U_ASSERT(fFields[UCAL_DST_OFFSET] <= getMaximum(UCAL_DST_OFFSET));
 }
 
 uint8_t Calendar::julianDayToDayOfWeek(double julian)
@@ -1699,11 +1728,20 @@ void Calendar::computeWeekFields(UErrorCode &ec) {
     }
     fFields[UCAL_WEEK_OF_YEAR] = woy;
     fFields[UCAL_YEAR_WOY] = yearOfWeekOfYear;
+    // min/max of years are not constrains for caller, so not assert here.
     // WEEK_OF_YEAR end
 
     int32_t dayOfMonth = fFields[UCAL_DAY_OF_MONTH];
     fFields[UCAL_WEEK_OF_MONTH] = weekNumber(dayOfMonth, dayOfWeek);
+    U_ASSERT(getMinimum(UCAL_WEEK_OF_MONTH) <= fFields[UCAL_WEEK_OF_MONTH]);
+    U_ASSERT(fFields[UCAL_WEEK_OF_MONTH] <= getMaximum(UCAL_WEEK_OF_MONTH));
+
     fFields[UCAL_DAY_OF_WEEK_IN_MONTH] = (dayOfMonth-1) / 7 + 1;
+    U_ASSERT(getMinimum(UCAL_DAY_OF_WEEK_IN_MONTH) <=
+             fFields[UCAL_DAY_OF_WEEK_IN_MONTH]);
+    U_ASSERT(fFields[UCAL_DAY_OF_WEEK_IN_MONTH] <=
+             getMaximum(UCAL_DAY_OF_WEEK_IN_MONTH));
+
 #if defined (U_DEBUG_CAL)
     if(fFields[UCAL_DAY_OF_WEEK_IN_MONTH]==0) fprintf(stderr, "%s:%d: DOWIM %d on %g\n",
         __FILE__, __LINE__,fFields[UCAL_DAY_OF_WEEK_IN_MONTH], fTime);
--- a/icu4c/source/i18n/gregoimp.cpp
+++ b/icu4c/source/i18n/gregoimp.cpp
@@ -33,28 +33,33 @@ int64_t ClockMath::floorDivide(int64_t numerator, int64_t denominator) {
 }
 
 int32_t ClockMath::floorDivide(double numerator, int32_t denominator,
-                          int32_t& remainder) {
-    double quotient;
-    quotient = uprv_floor(numerator / denominator);
-    remainder = (int32_t) (numerator - (quotient * denominator));
+                          int32_t* remainder) {
+    // For an integer n and representable ⌊x/n⌋, ⌊RN(x/n)⌋=⌊x/n⌋, where RN is
+    // rounding to nearest.
+    double quotient = uprv_floor(numerator / denominator);
+    // For doubles x and n, where n is an integer and ⌊x+n⌋ < 2³¹, the
+    // expression `(int32_t) (x + n)` evaluated with rounding to nearest
+    // differs from ⌊x+n⌋ if 0 < ⌈x⌉−x ≪ x+n, as `x + n` is rounded up to
+    // n+⌈x⌉ = ⌊x+n⌋ + 1.  Rewriting it as ⌊x⌋+n makes the addition exact.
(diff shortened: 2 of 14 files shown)


FrankYFTang:
I currently got the following result
```

=== Handling test: format/CalendarRegressionTest/TestUTCWrongAMPM22023: ===
   format {
      CalendarRegressionTest {
         TestUTCWrongAMPM22023 {
         UTC UCAL_AM_PM should be either UCAL_AM or UCAL_PM  but is 2
         FAIL: GMT get(UCAL_AM_PM, status) failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         GMT get(3) < getMinimum(3) : 0 < 1
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         GMT get(5) < getMinimum(5) : 0 < 1
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         GMT get(4) < getMinimum(4) : 0 < 1
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         UTC getMaximum(21) < get(21) : 86399999 < 86400000
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         UTC getMaximum(11) < get(11) : 23 < 24
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         UTC getMaximum(9) < get(9) : 1 < 2
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
         FAIL: GMT get() failed, error U_ILLEGAL_ARGUMENT_ERROR
      
         } ERRORS (22) in TestUTCWrongAMPM22023
      
   
      } ERRORS (22) in CalendarRegressionTest
   

   } ERRORS (22) in format


--------------------------------------
Errors in total: 22.
            TestUTCWrongAMPM22023
         CalendarRegressionTest
      format
   
--------------------------------------
Elapsed Time: 00:00:00.002
```

FrankYFTang:
```
-    int32_t millisInDay =  (int32_t) (localMillis - (days * kOneDay));
+    int32_t millisInDay =  (int32_t) (uprv_floor(localMillis) - (days * kOneDay));
```
will fix the am pm issues but we still have other out of bound

richgillam (approved):
This looks great, as far as I can tell (that's an awful lot of U_ASSERTs, but you can never be too careful).  After we talked about this on Wednesday, I was thinking maybe uprv_floor() was the answer to your problems.  Really glad that worked!

FrankYFTang:
> This looks great, as far as I can tell (that's an awful lot of U_ASSERTs, but you can never be too careful). After we talked about this on Wednesday, I was thinking maybe uprv_floor() was the answer to your problems. Really glad that worked!

I am very suprise those U_ASSERT actually catch that many cases. It seems unncessary to add those U_ASSERT but those U_ASSERT did break in many cases prior to my code changes under our current unit tests.

FrankYFTang:
@yumaoka - need your feedback about TimeZone related changes (of the limit and the code)
@sffc I think you know the double vs int32_t issue much better than I do so please review this and let me know is there anything I missed (or is my comments inaccurate)

eggrobin (changes requested):

icu4c/source/test/intltest/calregts.cpp
```diff
@@ -3089,6 +3090,105 @@ void CalendarRegressionTest::TestIslamicCalOverflow(void) {
...
+        }
+    }
+}
+
+void CalendarRegressionTest::TestUTCWrongAMPM22023(void) {
+    VerifyGetStayInBound(-1);
+    VerifyGetStayInBound(0);
+    VerifyGetStayInBound(-1e-8);
+    VerifyGetStayInBound(-1e-9);
+    VerifyGetStayInBound(-1e-15);
```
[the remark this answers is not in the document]
FrankYFTang: I agree w/ you in principle. It is tricky to figure out how to set the parameter input to produce that situration when it reach x and n though. Any suggestion?

FrankYFTang:
PTAL

icu4c/source/i18n/simpletz.cpp
```diff
@@ -519,10 +519,13 @@ SimpleTimeZone::getOffsetFromLocal(UDate date, UTimeZoneLocalOption nonExistingT
...
+    // For an integer n and representable ⌊x/n⌋, ⌊RN(x/n)⌋=⌊x/n⌋, where RN is
+    // rounding to nearest.
     double day = uprv_floor(date / U_MILLIS_PER_DAY);
-    // "(int32_t) (x + y)" will produce a 1-off rounding error while
-    // y > 1e8 and x is beteen (0, -1e-9).
-    // Fix it by wrapping x with uprv_flooring()
+    // For doubles x and n, where n is an integer and ⌊x+n⌋ < 2³¹, the
+    // expression `(int32_t) (x + n)` evaluated with rounding to nearest
+    // differs from ⌊x+n⌋ if 0 < ⌈x⌉−x ≪ x+n, as `x + n` is rounded up to
+    // n+⌈x⌉ = ⌊x+n⌋ + 1.  Rewriting it as ⌊x⌋+n makes the addition exact.
```
richgillam, line 528: Two comments:
1) Maybe it's because I'm still trying to wake up this morning, or maybe it's because I'm a music major, but I can't make head nor tail out of either of these comments.  I think they either need to be made more general, go into a lot more detail, or point to an external source (some general explanation of IEEE floating-point math?) that might explain better what's wrong and how this fixes it.
2) It doesn't help that these same indecipherable descriptions appear in four different places in our code base.  Maybe it's worth moving the actual math into a utility library somewhere-- not because the code is that complicated, but because the explanation is.

eggrobin: I definitely agree with 2, and we could then address 1 in the « go into more detail » direction, rather than handwaving a ≪ while desperately trying to squeeze it all in four lines. I guess `floorDivide` is already trying to be that general-purpose utility?

icu4c/source/i18n/calendar.cpp
```diff
@@ -1561,19 +1561,53 @@ void Calendar::computeFields(UErrorCode &ec)
     // Compute time-related fields.  These are independent of the date and
     // of the subclass algorithm.  They depend only on the local zone
     // wall milliseconds in day.
-    int32_t millisInDay =  (int32_t) (localMillis - (days * kOneDay));
+
+    // For doubles x and n, where n is an integer and ⌊x+n⌋ < 2³¹, the
+    // expression `(int32_t) (x + n)` evaluated with rounding to nearest
+    // differs from ⌊x+n⌋ if 0 < ⌈x⌉−x ≪ x+n, as `x + n` is rounded up to
+    // n+⌈x⌉ = ⌊x+n⌋ + 1.  Rewriting it as ⌊x⌋+n makes the addition exact.
+    int32_t millisInDay =  (int32_t) (uprv_floor(localMillis) - (days * kOneDay));
```
sffc, line 1569: Suggestion: To get `days`, up above, there is

```cpp
    int32_t days =  (int32_t)ClockMath::floorDivide(localMillis, (double)kOneDay);
```

Note that `millisInDay` is just getting the "remainder" of that division operation. It would be cleaner if you could do that all in one step (get the quotient and remainder). There are C++ standard library functions for this, such as

https://www.cplusplus.com/reference/cstdlib/div/

I think we have `uprv_div` but unsure.

icu4c/source/i18n/gregoimp.cpp
```diff
@@ -34,9 +34,14 @@ int64_t ClockMath::floorDivide(int64_t numerator, int64_t denominator) {
...
-    quotient = uprv_floor(numerator / denominator);
-    remainder = (int32_t) (numerator - (quotient * denominator));
+    // For an integer n and representable ⌊x/n⌋, ⌊RN(x/n)⌋=⌊x/n⌋, where RN is
+    // rounding to nearest.
+    double quotient = uprv_floor(numerator / denominator);
+    // For doubles x and n, where n is an integer and ⌊x+n⌋ < 2³¹, the
+    // expression `(int32_t) (x + n)` evaluated with rounding to nearest
+    // differs from ⌊x+n⌋ if 0 < ⌈x⌉−x ≪ x+n, as `x + n` is rounded up to
+    // n+⌈x⌉ = ⌊x+n⌋ + 1.  Rewriting it as ⌊x⌋+n makes the addition exact.
+    remainder = (int32_t) (uprv_floor(numerator) - (quotient * denominator));
```
sffc, line 44: Same suggestion: I suggest combining the division and remainder into one operation.

sffc (changes requested):

icu4c/source/i18n/calendar.cpp
eggrobin: `ClockMath` has a `floorDivide` with a remainder as an out parameter, so we should use that (and that would address the replication issue mentioned by Rich.

sffc: Note that `localMillis` is a double, which is garbage (timestamps should not be doubles). Do you know if there is a limit on the size of the double?

Note that the max safe integer-valued double, 9007199254740991, is less than the capacity of a uint64_t, 18446744073709551615.

icu4c/source/i18n/gregoimp.cpp
FrankYFTang: This function is THE combined ONE operation.

markusicu:
> `UDate` is `double`

Yes, *in ICU4C*. Much of the original ICU4C code was ported from ICU4J (actually from the code that Taligent contributed to Java 1.1.4) more than 25 years ago. `UDate` is the C++ version of `java.util.Date`, which is a glorified Java `long`, which corresponds to C++ `int64_t`. In the 90s, many (most?) C++ compilers did not support integer types with more than 32 bits, so the Taligent i18n team chose `UDate` as a typedef for `double`.

@FrankYFTang I suggest you submit an ICU ticket for comparing the Calendar math code between Java and C++, and consider changing the C++ implementation code (not the API) to working with `int64_t`, with whatever rounding (nearest/floor/...) is appropriate for double->int, and range-check assertions for int->double.

FYI @macchiati @pedberg-icu @yumaoka

FrankYFTang:
> @FrankYFTang I suggest you submit an ICU ticket...
https://unicode-org.atlassian.net/browse/ICU-22038

icu4c/source/i18n/simpletz.cpp
FrankYFTang: yes, I think ClockMath::floorDivide should the place. I really dislike it take the 3rd arg as reference and modify the ref value. I change it to a pointer here so in the caller call it is obvious that value could be changed after the call.

richgillam (dismissed):
Looks great to me...

sffc (approved):

FrankYFTang:
I just squash, PTAL

sffc (approved):

richgillam (approved):

icu4c/source/test/intltest/calregts.cpp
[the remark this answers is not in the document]
eggrobin: Honestly I don’t understand how we are getting a large n in this test to start with.
Do you know which of the (many) `floorDivide`s was breaking things?

FrankYFTang:
the fix is not good enough. It assert in v8 test under https://chromium-review.googlesource.com/c/v8/v8/+/3659713
see https://ci.chromium.org/p/v8/builders/try.triggered/v8_linux64_dbg_ng_triggered/b8813399630698400273?
and 
https://logs.chromium.org/logs/v8/buildbucket/cr-buildbucket/8813399630698400273/+/u/Test262_-_default/timeZones-accepted-by-DateTimeFo...
It assert in 
d8: ../../third_party/icu/source/i18n/calendar.cpp:1600: virtual void icu_71::Calendar::computeFields(UErrorCode &): Assertion `getMinimum(UCAL_ZONE_OFFSET) <= fFields[UCAL_ZONE_OFFSET]' failed.

somehow the fFields[UCAL_ZONE_OFFSET] is < getMinimum(UCAL_ZONE_OFFSET) 

The test is https://github.com/tc39/test262/blob/main/test/intl402/Intl/supportedValuesOf/timeZones-accepted-by-DateTimeFormat.js
which is 
```
const timeZones = Intl.supportedValuesOf("timeZone");

for (let timeZone of timeZones) {
  let obj = new Intl.DateTimeFormat("en", {timeZone});
  assert.sameValue(obj.resolvedOptions().timeZone, timeZone,
                   `${timeZone} is supported by DateTimeFormat`);
}
```
It just create all supported timezone . I will try to figure out which timezone cause that and what is the value.

FrankYFTang:
It assert on Asia/Makassar

FrankYFTang:
-43200000 -57360000
d8: ../../third_party/icu/source/i18n/calendar.cpp:1604: virtual void icu_71::Calendar::computeFields(UErrorCode &): Assertion `getMinimum(UCAL_ZONE_OFFSET) <= fFields[UCAL_ZONE_OFFSET]' failed.

getMinimum(UCAL_ZONE_OFFSET) is -43200000
but fFields[UCAL_ZONE_OFFSET] is -57360000

icu4c/source/i18n/gregoimp.cpp
```diff
@@ -33,28 +33,33 @@ int64_t ClockMath::floorDivide(int64_t numerator, int64_t denominator) {
 }
 
 int32_t ClockMath::floorDivide(double numerator, int32_t denominator,
-                          int32_t& remainder) {
-    double quotient;
-    quotient = uprv_floor(numerator / denominator);
-    remainder = (int32_t) (numerator - (quotient * denominator));
+                          int32_t* remainder) {
```
markusicu, line 36: FYI
I see that some of these output parameters got changed from references to pointers, even though they are still required.

We should use references when they are required, and pointers when they are optional (can be nullptr), as we have always done.

FrankYFTang: I think what we have always done is very bad due to the fact that by reading the code it is very difficult to realize the call will have side effect to change the value of that parameter. I understand that is what we always did but I think that is a good receipe for buggy code. The worst case for passing as pointer is the caller pass in nullptr and the code just crash right away, but the worst case for passing in as reference is the reader cannot tell that function call will change the value easily unless the reader have to read all the header for every single function call in the code.

markusicu: > I think what we have always done is very bad due to the fact that by reading the code it is very difficult to realize the call will have side effect to change the value of that parameter. I understand that is what we always did but I think that is a good receipe for buggy code. The worst case for passing as pointer is the caller pass in nullptr and the code just crash right away, but the worst case for passing in as reference is the reader cannot tell that function call will change the value easily unless the reader have to read all the header for every single function call in the code.

The rule that we have followed is that if it's nullable, use a pointer; otherwise, use a reference.
We have not really had any problems with this convention. I would rather not churn the code in this way.

richgillam: I'm going to chime in here to side with Markus.  I agree with Frank that the current reference syntax is not self-documenting and can lead to errors, but pointers in general are a pain in the butt and a source of errors, and consistency across the library is vital.

FrankYFTang: > We have not really had any problems with this convention. I would rather not churn the code in this way.

The problem happen when people read the code and debugging the code . The fact that it fool me while I debug this problem and cost me more time to address this issue itself is the problem. I would spot the issue much earlier if the function call is not an output parameter passing by reference.

FrankYFTang: I file https://unicode-org.atlassian.net/browse/ICU-22058 . Let's follow up there.

PR #2087, opened 2022-05-13
ICU-22028 Export collation and normalization data for ICU4X

hsivonen:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22028

This PR doesn't include tests. In practice, the test is that the end-to-end data transfer to ICU4X works. (To be documented on the ICU4X side.)

Exporting the normalization data is completely in the export tool. Exporting the collation data involves different behavior in the collation builder, which is why this patch adds an `icu4xMode` flag to the builder code. In this patch, it's always a run-time flag, since avoiding those branches in the builder doesn't seem like a worthwhile optimization to build compile-time option for.


hsivonen:
Marking as draft until the corresponding ICU4X change lands.

hsivonen:
I can repro the Mac failure locally. 

On Linux, after a successful build, `rg --binary UDataMemory_createNewInstance_72` finds:
```
lib/libicui18n.so.72.0.1: binary file matches (found "\0" byte around offset 7)

lib/libicuuc.so.72.0.1: binary file matches (found "\0" byte around offset 7)

i18n/collationroot.o: binary file matches (found "\0" byte around offset 7)

common/udata.o: binary file matches (found "\0" byte around offset 7)

common/udatamem.o: binary file matches (found "\0" byte around offset 7)
```

The result on Mac differs by not finding libicui18n, but the build errors out while linking that one. libicui18n is trying to link with `-licuuc`, so the symbols should come from there. However, adding `U_EXPORT2` between the return type and the function name in the two relevant function declarations does not remedy the problem.

Why?

hsivonen:
`unicode-update-tools ` is about a failure when bazelisk runs `genuca` in the ICU4X mode.

hsivonen:
Looks like MSVC in unhappy for the same reason as clang on Mac.

hsivonen:
I'm marking this ready for review despite the Mac and Windows linkage errors in CI, because
1. The linkage error is probably super-simple to resolve for someone who has seen it before.
2. The need for symbol visibility relates to forcing the root collation to be loaded from a file before ICU picks up the root collation that's already inside the .so/dll/dylib, and I expect the reviewer to have better ideas for how to accomplish that.

icu4c/source/common/udatamem.h
```diff
@@ -44,7 +44,7 @@ struct UDataMemory {
     int32_t           length;      /* Length of the data in bytes; -1 if unknown.     */
 };
 
-U_CFUNC UDataMemory *UDataMemory_createNewInstance(UErrorCode *pErr);
+U_CFUNC UDataMemory U_EXPORT2 *UDataMemory_createNewInstance(UErrorCode *pErr);
```
jefgen, line 47: RE: linker errors

This is going to sound really silly, but I think you need to move the `*`.

The data type returned by the function is `UDataMemory*`. When you add the `U_EXPORT2`, this ends up (unintentionally) changing things.

```suggestion
U_CFUNC UDataMemory* U_EXPORT2 UDataMemory_createNewInstance(UErrorCode *pErr);
```

hsivonen: Thank you! This alone wasn't sufficient, but locally on Mac also changing `U_CFUNC` to `U_CAPI` was needed also. Let's see what happens in CI.

hsivonen:
I expect to be away from GitHub for a couple of weeks soon, so if this gets approval to land, please land this without waiting for me, and if changes are needed to address review comments, I defer to @sffc, @echeran , or someone they delegate to to make the appropriate adjustments.

There's a backport to the 71 branch in #2115 in order to get a data export as of Unicode 14 as opposed to Unicode 15 pre-release. If this PR requires changes that need to be reflected in backporting, the actions I took for backport were resolving a trivial failure for the patch to apply as-is to one header file and re-running the `genuca` `bazelisk` tasks (see `.azure-pipelines.yml`) to have the checked-in root collation serializations in sync with the checked-in collation data.

icu4c/source/tools/icuexportdata/icuexportdata.cpp
```diff
@@ -387,6 +862,46 @@ int exportUprops(int argc, char* argv[]) {
...
+            || options[OPT_HELP_QUESTION_MARK].doesOccur
+            || !options[OPT_MODE].doesOccur) {
+        FILE *stdfile=argc<0 ? stderr : stdout;
+        fprintf(stdfile,
+            "usage: %s -m uprops [-options] [--all | properties...]\n"
+            "\tdump Unicode property data to .toml files\n"
+            "options:\n"
+            "\t-h or -? or --help  this usage text\n"
+            "\t-V or --version     show a version message\n"
+            "\t-m or --mode        mode: currently only 'uprops', but more may be added\n"
```
robertbastian, line 876: update

echeran (approved):
LGTM

PR #2088, opened 2022-05-13
ICU-22029 Avoid using ATOMIC_VAR_INIT if the library has deprecated it.

pkasting:
This fixes warnings with -std=c++20 -Wdeprecated-pragma.

Fixes https://unicode-org.atlassian.net/browse/ICU-22029 .

Bug: chromium:1284275

### DIFF
--- a/icu4c/source/common/umutex.h
+++ b/icu4c/source/common/umutex.h
@@ -71,7 +71,12 @@ U_NAMESPACE_BEGIN
  ****************************************************************************/
 
 typedef std::atomic<int32_t> u_atomic_int32_t;
+#if !defined(__cpp_lib_atomic_value_initialization) || \
+    __cpp_lib_atomic_value_initialization < 201911L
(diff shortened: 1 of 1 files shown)


aheninger (approved):
I think this change looks fine.

I'm somewhat inclined to look into completely removing the ATOMIC_INT32_T_INITIALIZER macro, though. It's existence dates from the C++03 era, when atomic support in ICU had to be pieced together from a mishmash of non-portable compiler and OS functions.

Removal of the macro would be a more intrusive change, but would leave things simpler overall.

aheninger:
#2090 is the bigger change, removing ICU's ATOMIC_INT32_T_INITIALIZER macro altogether.

aheninger:
#2090 is having some kind of odd MSVC build problem. If that can't be easily resolved, this PR may be the better fix.

pkasting:
Looks like PR #2090 was merged, so this is moot.  Closing.

aheninger:
Thank you so much for pointing out the problem in the first place.

PR #2090, opened 2022-05-16
ICU-21960 C++20 Warnings from ATOMIC_VAR_INIT

aheninger:
Remove the ICU macros ATOMIC_INT32_T_INITIALIZER and U_INITONCE_INITIALIZER,
which made use of C++ ATOMIC_VAR_INIT, which has been removed from C++20.

With modern C++ features being available, these macros no longer served
any real need.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21960

### DIFF
--- a/icu4c/source/common/brkiter.cpp
+++ b/icu4c/source/common/brkiter.cpp
@@ -279,7 +279,7 @@ ICUBreakIteratorService::~ICUBreakIteratorService() {}
 // defined in ucln_cmn.h
 U_NAMESPACE_END
 
-static icu::UInitOnce gInitOnceBrkiter = U_INITONCE_INITIALIZER;
+static icu::UInitOnce gInitOnceBrkiter {};
 static icu::ICULocaleService* gService = NULL;
 
 
--- a/icu4c/source/common/characterproperties.cpp
+++ b/icu4c/source/common/characterproperties.cpp
@@ -40,7 +40,7 @@ constexpr int32_t NUM_INCLUSIONS = UPROPS_SRC_COUNT + (UCHAR_INT_LIMIT - UCHAR_I
 
 struct Inclusion {
     UnicodeSet  *fSet = nullptr;
-    UInitOnce    fInitOnce = U_INITONCE_INITIALIZER;
+    UInitOnce    fInitOnce {};
 };
 Inclusion gInclusions[NUM_INCLUSIONS]; // cached getInclusions()
 
--- a/icu4c/source/common/emojiprops.cpp
+++ b/icu4c/source/common/emojiprops.cpp
@@ -22,7 +22,7 @@ U_NAMESPACE_BEGIN
 namespace {
 
 EmojiProps *singleton = nullptr;
-icu::UInitOnce emojiInitOnce = U_INITONCE_INITIALIZER;
+icu::UInitOnce emojiInitOnce {};
 
 UBool U_CALLCONV emojiprops_cleanup() {
     delete singleton;
--- a/icu4c/source/common/loadednormalizer2impl.cpp
+++ b/icu4c/source/common/loadednormalizer2impl.cpp
@@ -134,14 +134,14 @@ U_CDECL_END
 
 #if !NORM2_HARDCODE_NFC_DATA
 static Norm2AllModes *nfcSingleton;
-static icu::UInitOnce nfcInitOnce = U_INITONCE_INITIALIZER;
+static icu::UInitOnce nfcInitOnce {};
 #endif
 
 static Norm2AllModes *nfkcSingleton;
-static icu::UInitOnce nfkcInitOnce = U_INITONCE_INITIALIZER;
+static icu::UInitOnce nfkcInitOnce {};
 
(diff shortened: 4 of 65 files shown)


markusicu:
Changes lgtm, thanks!
If ATOMIC_VAR_INIT is gone from C++20, then why do our two C++20 build bots not fail?
Should we add some warning or something to one of our build bots?

aheninger:
> If ATOMIC_VAR_INIT is gone from C++20, then why do our two C++20 build bots not fail?

My clang:
```
$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
My configure:
```CXXFLAGS='-std=c++20 -stdlib=libc++' CPPFLAGS="-fsanitize=$SANITIZER -DREGEX_DEBUG=1 -DRBBI_DEBUG=1" LDFLAGS="-stdlib=libc++ -fsanitize=$SANITIZER" ICU_DATA_BUILDTOOL_OPTS=--include_uni_core_data ./runConfigureICU --enable-debug --disable-release Linux --disable-renaming```

It turns out that to see the failures, both clang 14 and `-stdlib=libc++` are needed.

From the clang 14 release notes
https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#c-20-feature-support

C++20 Feature Support
-  The ATOMIC_VAR_INIT and ATOMIC_FLAG_INIT macros from the C standard library <stdatomic.h> header are now diagnosed as deprecated in C++20 and later. Note, the behavior is specific to the inclusion of <stdatomic.h> in C++ code; the STL <atomic> header also controls the behavior of these macros and is not affected by these changes. The <stdatomic.h> diagnostic can be disabled by defining the _CLANG_DISABLE_CRT_DEPRECATION_WARNINGS macro prior to including the header.

When building the current main branch with C++20, the warnings look like this (one of many):
```
brkiter.cpp:282:42: warning: macro 'ATOMIC_VAR_INIT' has been marked as deprecated [-Wdeprecated-pragma]
static icu::UInitOnce gInitOnceBrkiter = U_INITONCE_INITIALIZER;
                                         ^
./umutex.h:108:33: note: expanded from macro 'U_INITONCE_INITIALIZER'
#define U_INITONCE_INITIALIZER {ATOMIC_INT32_T_INITIALIZER(0), U_ZERO_ERROR}
                                ^
./umutex.h:74:41: note: expanded from macro 'ATOMIC_INT32_T_INITIALIZER'
#define ATOMIC_INT32_T_INITIALIZER(val) ATOMIC_VAR_INIT(val)
                                        ^
/usr/lib/llvm-14/bin/../include/c++/v1/atomic:2713:43: note: macro marked 'deprecated' here
#  pragma clang deprecated(ATOMIC_VAR_INIT)
                                          ^
```

The clang C++ 20 build bot is running clang 13.
The gcc bot is running gcc 11. 
Neither is using llvm's libc++; both use gnu's libstdc++

markusicu:
Thanks Andy!
@jefgen is clang 14 available for Azure build bots?
And what do you think about switching the clang C++20 bot to using `-stdlib=libc++` ?

markusicu (approved):
lgtm tnx!

jefgen:
Thanks Andy!

> is clang 14 available for Azure build bots?

Unfortunately, it looks like the latest version that is available by default in the images is clang 12.0.1 (from Ubuntu 22.04), so we would need to manually install clang-14 (similar to what we currently do with clang 13).

> And what do you think about switching the clang C++20 bot to using -stdlib=libc++ ?

I wonder if we could add another (additional) build bot to build with libc++ (so we'd have two that build with clang-14 C++20)?

PR here to update the build bots: https://github.com/unicode-org/icu/pull/2095

aheninger:
libc++ is default on Macintosh. On Ubuntu, it's not installed with clang by
default, but requires an explicit
sudo apt install libc++1 libc++-dev libc++abi-dev libc++abi1

On my Ubuntu 22.04, clang 14 is the default version, so it seems curious
that you're getting 12.0.1.

I originally switched to libc++ several years ago because support for some
of the sanitizers was better.

jefgen:
> On my Ubuntu 22.04, clang 14 is the default version, so it seems curious that you're getting 12.0.1.

Yeah, it looks like you're right that the default version for clang on Ubuntu 22.04 (jammy jellyfish) looks to be version 14.
Ex: https://packages.ubuntu.com/jammy/devel/clang lists 1:14.0-55~exp2

However, I was going based on the list of "Installed Software" for the Azure Pipelines image "Ubuntu 22.04 Preview" from here:
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md
which only lists clang-11 and clang-12:

> Installed Software
> Language and Runtime
> Clang 11.1.0, 12.0.1

So perhaps they change the version of the installed compiler for the Azure Pipelines image?
Though the image is still in "preview" state, so perhaps the list of software isn't finalized yet. I don't know if we'd want to move over to the preview image just yet.

PR #2098, opened 2022-05-21
ICU-21946 RBBI Break Cache Optimizations

aheninger:
Adjust RuleBasedBreakIterator::BreakCache::populateNear() to retain
the cache contents in additional cases where they are still useful, resulting in
improved performance.

This change is related to PR #2039, which addressed the same problem. This one
retains the cache contents in more situations.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21946

### DIFF
--- a/icu4c/source/common/rbbi_cache.cpp
+++ b/icu4c/source/common/rbbi_cache.cpp
@@ -341,41 +341,87 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
     }
     U_ASSERT(position < fBoundaries[fStartBufIdx] || position > fBoundaries[fEndBufIdx]);
 
-    // Find a boundary somewhere in the vicinity of the requested position.
-    // Depending on the safe rules and the text data, it could be either before, at, or after
-    // the requested position.
-
+    // Add boundaries to the cache near the specified position.
+    // The given position need not be a boundary itself.
+    // The input position must be within the range of the text, and
+    // on a code point boundary.
+    // If the requested position is a break boundary, leave the iteration
+    // position on it.
+    // If the requested position is not a boundary, leave the iteration
+    // position on the preceding boundary and include both the
+    // preceding and following boundaries in the cache.
+    // Additional boundaries, either preceding or following, may be added
+    // to the cache as a side effect.
 
     // If the requested position is not near already cached positions, clear the existing cache,
     // find a near-by boundary and begin new cache contents there.
 
-    if ((position < fBoundaries[fStartBufIdx] - 15) || position > (fBoundaries[fEndBufIdx] + 15)) {
-        int32_t aBoundary = 0;
-        int32_t ruleStatusIndex = 0;
-        if (position > 20) {
-            int32_t backupPos = fBI->handleSafePrevious(position);
+    // Threshold for a text position to be considered near to existing cache contents.
+    // TODO: See issue ICU-22024 "perf tuning of Cache needed."
+    //       This value is subject to change. See the ticket for more details.
+    static constexpr int32_t CACHE_NEAR = 15;
 
-            if (backupPos > 0) {
-                // Advance to the boundary following the backup position.
-                // There is a complication: the safe reverse rules identify pairs of code points
-                // that are safe. If advancing from the safe point moves forwards by less than
-                // two code points, we need to advance one more time to ensure that the boundary
-                // is good, including a correct rules status value.
-                //
-                fBI->fPosition = backupPos;
-                aBoundary = fBI->handleNext();
-                if (aBoundary <= backupPos + 4) {
-                    // +4 is a quick test for possibly having advanced only one codepoint.
-                    // Four being the length of the longest potential code point, a supplementary in UTF-8
-                    utext_setNativeIndex(&fBI->fText, aBoundary);
-                    if (backupPos == utext_getPreviousNativeIndex(&fBI->fText)) {
-                        // The initial handleNext() only advanced by a single code point. Go again.
-                        aBoundary = fBI->handleNext();   // Safe rules identify safe pairs.
-                    }
+    int32_t aBoundary = -1;
+    int32_t ruleStatusIndex = 0;
+    bool retainCache = false;
+    if ((position > fBoundaries[fStartBufIdx] - CACHE_NEAR) && position < (fBoundaries[fEndBufIdx] + CACHE_NEAR)) {
+        // Requested position is near the existing cache. Retain it.
+        retainCache = true;
+    } else if (position <= CACHE_NEAR) {
+        // Requested position is near the start of the text. Fill cache from start, skipping
+        // the need to find a safe point.
+        retainCache = false;
+        aBoundary = 0;
+    } else {
+        // Requested position is not near the existing cache.
+        // Find a safe point to refill the cache from.
+        int32_t backupPos = fBI->handleSafePrevious(position);
+
+        if (fBoundaries[fEndBufIdx] < position && fBoundaries[fEndBufIdx] >= (backupPos - CACHE_NEAR)) {
+            // The requested position is beyond the end of the existing cache, but the
+            // reverse rules produced a position near or before the cached region.
+            // Retain the existing cache, and fill from the end of it.
+            retainCache = true;
+        } else if (backupPos < CACHE_NEAR) {
+            // The safe reverse rules moved us to near the start of text.
+            // Take that (index 0) as the backup boundary, avoiding the complication
+            // (in the following block) of moving forward from the safe point to a known boundary.
+            //
+            // Retain the cache if it begins not too far from the requested position.
+            aBoundary = 0;
+            retainCache = (fBoundaries[fStartBufIdx] <= (position + CACHE_NEAR));
+        } else {
+            // The safe reverse rules produced a position that is neither near the existing
+            // cache, nor near the start of text.
+            // Advance to the boundary following.
+            // There is a complication: the safe reverse rules identify pairs of code points
+            // that are safe. If advancing from the safe point moves forwards by less than
+            // two code points, we need to advance one more time to ensure that the boundary
+            // is good, including a correct rules status value.
+            retainCache = false;
+            fBI->fPosition = backupPos;
+            aBoundary = fBI->handleNext();
+            if (aBoundary != UBRK_DONE && aBoundary <= backupPos + 4) {
+                // +4 is a quick test for possibly having advanced only one codepoint.
+                // Four being the length of the longest potential code point, a supplementary in UTF-8
+                utext_setNativeIndex(&fBI->fText, aBoundary);
+                if (backupPos == utext_getPreviousNativeIndex(&fBI->fText)) {
+                    // The initial handleNext() only advanced by a single code point. Go again.
+                    aBoundary = fBI->handleNext();   // Safe rules identify safe pairs.
                 }
-                ruleStatusIndex = fBI->fRuleStatusIndex;
             }
+            if (aBoundary == UBRK_DONE) {
+                // Note (Andy Heninger): I don't think this condition can occur, but it's hard
+                // to prove that it can't. We ran off the end of the string looking a boundary
+                // following a safe point; choose the end of the string as that boundary.
+                aBoundary = utext_nativeLength(&fBI->fText);
+            }
+            ruleStatusIndex = fBI->fRuleStatusIndex;
         }
+    }
+
+    if (!retainCache) {
+        U_ASSERT(aBoundary != -1);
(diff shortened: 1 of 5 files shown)


markusicu:
Does this subsume/replace PR #2039 or does it go on top of that?

aheninger:
> Does this subsume/replace PR #2039 or does it go on top of that?

This PR subsumes/replaces #2039. This one covers the case #2039 fixes, plus additional cases where a useful break cache was being discarded for no good reason.

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
-                    // Four being the length of the longest potential code point, a supplementary in UTF-8
-                    utext_setNativeIndex(&fBI->fText, aBoundary);
-                    if (backupPos == utext_getPreviousNativeIndex(&fBI->fText)) {
-                        // The initial handleNext() only advanced by a single code point. Go again.
-                        aBoundary = fBI->handleNext();   // Safe rules identify safe pairs.
-                    }
+    int32_t aBoundary = -1;
+    int32_t ruleStatusIndex = 0;
+    bool retainCache = false;
+    if ((position > fBoundaries[fStartBufIdx] - 15) && position < (fBoundaries[fEndBufIdx] + 15)) {
```
FrankYFTang, line 362: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+        // Requested position is near the existing cache. Retain it.
+        retainCache = true;
+    } else if (position <= 20) {
```
FrankYFTang, line 365: What is this magic number 20 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+        // Requested position is near the start of the text. Fill cache from start, skipping
+        // the need to find a safe point.
+        retainCache = false;
+        aBoundary = 0;
+    } else {
+        // Requested position is not near the existing cache.
+        // Find a safe point to refill the cache from.
+        int32_t backupPos = fBI->handleSafePrevious(position);
+
+        if (fBoundaries[fEndBufIdx] < position && fBoundaries[fEndBufIdx] >= (backupPos - 15)) {
```
FrankYFTang, line 375: again, what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+            // The requested position is beyond the end of the existing cache, but the
+            // reverse rules produced a position near or before the cached region.
+            // Retain the existing cache, and fill from the end of it.
+            retainCache = true;
+        } else if (backupPos < 15) {
```
FrankYFTang, line 380: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+            // The safe reverse rules moved us to near the start of text.
+            // Take that (index 0) as the backup boundary, avoiding the complication
+            // (in the following block) of moving forward from the safe point to a known boundary.
+            //
+            // Retain the cache if it begins not too far from the requested position.
+            aBoundary = 0;
+            retainCache = (fBoundaries[fStartBufIdx] <= (position + 15));
```
FrankYFTang, line 387: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
```diff
@@ -341,41 +341,82 @@ UBool RuleBasedBreakIterator::BreakCache::populateNear(int32_t position, UErrorC
...
+            // cache, nor near the start of text.
+            // Advance to the boundary following.
+            // There is a complication: the safe reverse rules identify pairs of code points
+            // that are safe. If advancing from the safe point moves forwards by less than
+            // two code points, we need to advance one more time to ensure that the boundary
+            // is good, including a correct rules status value.
+            retainCache = false;
+            fBI->fPosition = backupPos;
+            aBoundary = fBI->handleNext();
+            if (aBoundary != UBRK_DONE && aBoundary <= backupPos + 4) {
```
FrankYFTang, line 399: what is the magic number 4 here? Could we add a constant for that and add a note about how we pick that number?

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1436,42 +1436,84 @@ int current() {
...
-                    // that are safe. If advancing from the safe point moves forwards by less than
-                    // two code points, we need to advance one more time to ensure that the boundary
-                    // is good, including a correct rules status value.
-                    //
-                    fPosition = backupPos;
+        int startOfText = fText.getBeginIndex();
+        int aBoundary = -1;
+        int ruleStatusIndex = 0;
+        boolean retainCache = false;
+        if ((position > fBoundaries[fStartBufIdx] - 15) && position < (fBoundaries[fEndBufIdx] + 15)) {
```
FrankYFTang, line 1458: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1436,42 +1436,84 @@ int current() {
...
+            // Requested position is near the existing cache. Retain it.
+            retainCache = true;
+        } else if (position <= startOfText + 20) {
```
FrankYFTang, line 1461: what is the magic number 20 here? Could we add a constant for that and add a note about how we pick that number?

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1436,42 +1436,84 @@ int current() {
...
+            // Requested position is near the start of the text. Fill cache from start, skipping
+            // the need to find a safe point.
+            retainCache = false;
+            aBoundary = startOfText;
+        } else {
+            // Requested position is not near the existing cache.
+            // Find a safe point to refill the cache from.
+            int backupPos = handleSafePrevious(position);
+
+            if (fBoundaries[fEndBufIdx] < position && fBoundaries[fEndBufIdx] >= (backupPos - 15)) {
```
FrankYFTang, line 1471: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1436,42 +1436,84 @@ int current() {
...
+                // The requested position is beyond the end of the existing cache, but the
+                // reverse rules produced a position near or before the cached region.
+                // Retain the existing cache, and fill from the end of it.
+                retainCache = true;
+            } else if (backupPos < startOfText + 15) {
```
FrankYFTang, line 1476: what is the magic number 15 here? Could we add a constant for that and add a note about how we pick that number?

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1436,42 +1436,84 @@ int current() {
...
+                // There is a complication: the safe reverse rules identify pairs of code points
+                // that are safe. If advancing from the safe point moves forwards by less than
+                // two code points, we need to advance one more time to ensure that the boundary
+                // is good, including a correct rules status value.
+                //
+                retainCache = false;
+                fPosition = backupPos;
+                aBoundary = handleNext();
+                if (aBoundary == backupPos + 1 ||
+                        (aBoundary == backupPos + 2 &&
```
FrankYFTang, line 1497: what is the magic number 2 here? Could we add a constant for that and add a note about how we pick that number?

icu4c/source/common/rbbi_cache.cpp
markusicu: Sounds reasonable -- I am ok with keeping magic numbers, together with a comment about them.

markusicu:
FYI  I just glanced at the changes. They look roughly plausible, and no better person to know what they are doing here than Andy... let me know if you want me to look more deeply. I could try to squeeze it in.

icu4c/source/common/rbbi_cache.cpp
FrankYFTang: so... 
1. how do you come out with the guess of each.
2. Is there a requirement to keep some of these guessed magic number in sync? so if I later change one 15 to 18, I have to change another 15 to 18 together or otherwise it will cause bug?  for example, there aer two "15" in line 362, if we later change one of them but not both of them to 18 , would that cause bug? How about the 15 in line 375?  I cannot tell the "relationship" between these "15"s . Are they all independent or are some of them need to be the same value even currently they are all 15? any constrain of these guess? could we later guess 14 or 16 and won't cause problem?

aheninger: These numbers all represent a concept of "close". Is a requested position close to existing boundary cache contents, or close to a known boundary, like the start of text. If close, the code will prepend or append to existing cache contents to pick up the needed content. If not close, existing cache contents are dumped and a new cache is started around the needed position. The cache can only hold a contiguous set of boundary positions. It can't handle boundaries from discontinuous regions of the input text

There is no required relationship between any of "close" numbers. Any of them can be adjusted independently of the others. Any values will work - the break iterator will still produce correct results. What will change is the performance.

Choosing optimum "close" numbers will depend on the scenario being measured. And probably the break iterator type.  It's going to be tricky to do. For now, I think we just want to verify that the various tests and code paths do what the comments say, retaining the cache when its content is "close" and discarding it when not. You don't even need to worry about off-by-one errors.

I added the new random access test specifically to tickle all of the code paths through this function. And stepped through a bunch of them in the debugger to verify that the code was doing what I expected. There's no other obvious easy way to test it; the break iterator still produces correct results if the cache management is broken, just with worse performance.

icu4j/main/classes/core/src/com/ibm/icu/text/RuleBasedBreakIterator.java
aheninger: backupPos and aBoundary are both string indexes into the input text. The if statement is testing if the difference between them is a single code point. So a difference of 1, or a difference of 2 with the text contents being supplemental pair.

This is Java, so the input is known to be UTF-16. The equivalent function in C++ looks different because it could be either UTF-16 or UTF-8, thanks to UText.

icu4c/source/common/rbbi_cache.cpp
FrankYFTang: >There is no required relationship between any of "close" numbers. Any of them can be adjusted independently of the others.
ok. My concern is such "no required connection" is not clear to the reader of the code 10 years later. How could we land the code in a way that clearly indicating one 15 not need to be the same value as another 15 in the code. Should we add 
// The following value do NOT need to be the same. There is no required relationship between any of 
// these radom guess. They could be changed independently even if they currently share the same value.
constexpr int32_t kRandomGuess1  = 15;
constexpr int32_t kRandomGuess2 = 15;
constexpr int32_t kRandomGuess3  = 15;
constexpr int32_t kRandomGuess4 = 15;
constexpr int32_t kRandomGuess5 = 4;
constexpr int32_t kRandomGuess6 = 20;

and use them in different spots?
(or maybe come out same better name for these random guess.)

aheninger:
Can I ping Markus for a review? I'd like to finish off this PR.

icu4c/source/common/rbbi_cache.cpp
markusicu: Andy added a comment. Works for me.

markusicu (approved):
lgtm tnx!

Looks plausible (I didn't try to follow every nuance), and improved with added comments and named constants. And I trust Andy with this code. (And all tests pass.)

PR #2100, opened 2022-05-24
ICU-22043 Change minimum ZONE_OFFSET value from -12h to -16h

FrankYFTang:
GMT offset for Asia/Manila before  Mon Dec 30 23:59:59 1844 is  -15.9333333333 hours
See "zdump Asia/Manila"
so we need to change the minimum to -16 hours.
Also add unit tests to ensure the assertion won't happen.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22043

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -639,7 +639,7 @@ static const int32_t kCalendarLimits[UCAL_FIELD_COUNT][4] = {
     {           0,            0,            59,            59  }, // MINUTE
     {           0,            0,            59,            59  }, // SECOND
     {           0,            0,           999,           999  }, // MILLISECOND
-    {-12*kOneHour, -12*kOneHour,   12*kOneHour,   30*kOneHour  }, // ZONE_OFFSET
+    {-16*kOneHour, -16*kOneHour,   12*kOneHour,   30*kOneHour  }, // ZONE_OFFSET
     { -1*kOneHour,  -1*kOneHour,    2*kOneHour,    2*kOneHour  }, // DST_OFFSET
     {/*N/A*/-1,       /*N/A*/-1,     /*N/A*/-1,       /*N/A*/-1}, // YEAR_WOY
     {           1,            1,             7,             7  }, // DOW_LOCAL
--- a/icu4c/source/test/intltest/calregts.cpp
+++ b/icu4c/source/test/intltest/calregts.cpp
@@ -98,6 +98,7 @@ CalendarRegressionTest::runIndexedTest( int32_t index, UBool exec, const char* &
         CASE(54,TestWeekOfYear13548);
         CASE(55,Test13745);
         CASE(56,TestUTCWrongAMPM22023);
+        CASE(57,TestAsiaManilaAfterSetGregorianChange22043);
     default: name = ""; break;
     }
 }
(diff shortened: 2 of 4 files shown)


icu4c/source/i18n/calendar.cpp
```diff
@@ -639,7 +639,7 @@ static const int32_t kCalendarLimits[UCAL_FIELD_COUNT][4] = {
     {           0,            0,            59,            59  }, // MINUTE
     {           0,            0,            59,            59  }, // SECOND
     {           0,            0,           999,           999  }, // MILLISECOND
-    {-12*kOneHour, -12*kOneHour,   12*kOneHour,   30*kOneHour  }, // ZONE_OFFSET
+    {-16*kOneHour, -16*kOneHour,   12*kOneHour,   30*kOneHour  }, // ZONE_OFFSET
```
sffc, line 642: Question: What is the difference between "minimum" and "greatest min"?

icu4c/source/i18n/calendar.cpp
```diff
@@ -639,7 +639,7 @@ static const int32_t kCalendarLimits[UCAL_FIELD_COUNT][4] = {
...
     { -1*kOneHour,  -1*kOneHour,    2*kOneHour,    2*kOneHour  }, // DST_OFFSET
```
sffc, line 643: Question: Do you need this change in Java, too? Maybe at least add tests for it?

FrankYFTang: good point. Let me check.

icu4c/source/i18n/calendar.cpp
FrankYFTang: [UCAL_LIMIT_MINIMUM](https://source.chromium.org/chromium/chromium/src/+/main:third_party/icu/source/i18n/unicode/calendar.h;drc=4f130855d9bc6894ecf77a8230c8071c3a2c96be;l=1614) is used to return in 
https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1Calendar.html#a1a75a6ed6500cf7e36977ada0cc4412f


[UCAL_LIMIT_GREATEST_MINIMUM](https://source.chromium.org/chromium/chromium/src/+/main:third_party/icu/source/i18n/unicode/calendar.h;drc=4f130855d9bc6894ecf77a8230c8071c3a2c96be;l=1615)) is used to return in 
https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1Calendar.html#a03311976955b27d0ba49d31059619fb6

sffc: OK, a typical ICU function with little helpful documentation. :)

If the min and greatest-min were equal before, it makes sense to keep them equal now

FrankYFTang:
will add Java test tomorrw. Do you think the C++ change is ok. Currently the ICU rolling of my last PR got stuck because of this so if you are ok with the C++ part I will cherry them first

sffc (dismissed):
Yes it's fine to do Java in another PR

icu4c/source/i18n/calendar.cpp
yumaoka: This also leads a question that the data for least max/greatest max. These are now +16 hours and +30 hours.

There are some Pacific islands actively using UTC+14 offset. I don't really know any positive offset beyond +14 has been ever used (or such data in tz database exists.

sffc (approved):

PR #2103, opened 2022-05-26
ICU-21957 integrate CLDR release-42-m1 (early milestone) to ICU main for ICU 72

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21957

This integrates CLDR release-42-m1 (milestone release as of the start of general submission) to ICU. Several significant changes:
- CLDR version is updated from 41 to 42, notably in the icu4c/source/data/coll/* files and in the /icu4c/source/data/*/LOCALE_DEPS.json files.
- CLDR 42 is changing the way spaces are used in patterns to be more “Unicode-like”, moving away from just ASCII-range space to use things like U+2009 thin space (on either side of en-dash in intervalFormats for Latin script), and U+202F narrow no-break-space in narrow unit patterns, to separate short/narrow AM/PM markers for Latin/Cyrillic/Greek, and to separate the year marker abbreviation in Cyrillic. These changes are not yet complete in this milestone 1 integration. However, they affect many unit tests. They also required some updates to the SimpleDateFormat parsing code (mostly in ICU4J) to allow for  the non-breaking spaces in patterns. Many C tests are switched over to using u"xxx" strings.
- For the date-time combining patterns: The forms that specify date "at" time (and thus require particular grammatical inflections in many languages) are moved to a new type="atTime" variant in CLDR, and a new DateTimePatterns%atTime bundle in ICU. What remains for the standard date-time combining patterns are the versions that do not generally require any special grammatical inflection (e.g. just using comma or space instead of literal text). This requires changes to the code that loads these patterns for SimpleDateFormat, DateTimePatternGenerator, RelativeDateTimeFormatter in order to keep using the atTime forms for these. For DateIntervalFormat we are now (with this PR) using the standard forms (to accommodate patterns like "May 25, 3:00-5:00 PM" for which the "atTime" combining patterns would be inappropriate. All of this is awaiting spec guidance from CLDR on which patterns to use in which situations.
- CLDR has also added some alternate currency patterns. For locales other than English, these currently have draft="provisional" status and so are not converted to ICU (after vetters verify them the draft status should get updated); in the PR the alternates only appear in English:
    - currencyFormat%alphaNextToNumber: For currency formats that have no space between ¤ and the number, this alternate pattern provides a different pattern (usually with space) to use when the currency symbol has a letter on the side adjacent to the number. May be used for both standard and compact decimal formats; also for accountingFormat%alphaNextToNumber.
    - currencyFormat%noCurrency: Provides a currency-style format to use for currency-style formats without the currency symbol. This format may be used, for example, in a table of values all for the same currency (in which repeating the symbol for each value is unnecessary). Typically this is not needed for compact currency formats, since the compact decimals can be used instead, but it may be provided if necessary.
- Other more specific changes:
    - There is a new measurement unit, duration-quarter.
    - Two new currency codes: SLE, VED
    - icu4c/source/data/coll/root.txt emoji collation is updated for Unicode 15 emoji
    - Maltese adds plural category two (and Catalan adds category many for special use mainly in compact decimals, like French, Italian, etc.)
    - The English name of QAR changes to use riyal instead of rial
    - English adds/updates display names for several languages
- This does not yet have ICU versions of the new CLDR personName formats data (need to have a discussion about the desired ICU format for that).

Note: I added a logKnownIssue skip for part of FormattedStringBuilderTest::testInsertOverflow() to address a crash that only happens running Ci exhaustive tests for Clang/Linux (not on my local system) which appears to have nothing to do with this PR; see https://unicode-org.atlassian.net/browse/ICU-22047. When the problematic code is skipped, the test uses an alternative way of generating a long enough UnicodeString so the remaining tests in that function are valid.


macchiati (approved):

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -161,8 +161,8 @@ static const int8_t kTimeFieldsCount = 10;
 // of a resource.
 static const UChar gDefaultPattern[] =
 {
-    0x79, 0x79, 0x79, 0x79, 0x4D, 0x4D, 0x64, 0x64, 0x20, 0x68, 0x68, 0x3A, 0x6D, 0x6D, 0x20, 0x61, 0
-};  /* "yyyyMMdd hh:mm a" */
+    0x79, 0x4D, 0x4D, 0x64, 0x64, 0x20, 0x68, 0x68, 0x3A, 0x6D, 0x6D, 0x202F, 0x61, 0
+};  /* "yMMdd hh:mm\u202Fa" */
```
srl295, line 165: note that the gDefaultPattern changed.

srl295 (approved):

pedberg-icu:
> Review tip:
> 
> ```shell
> $ gh pr checkout 2103
> $ git diff fa702a450462925d520dc86a372c77827e82aa38..e17219582ed0396d993927066e95af0c9199f8db -IcldrVersion -IVersion
> ```
> 
> (ignore version-number changes)

Probably need to update that to
$ git diff edfbe1704a7fbe758ff77f43e471a36b727c4c5b..e17219582ed0396d993927066e95af0c9199f8db -IcldrVersion -IVersion

icu4c/source/i18n/smpdtfmt.cpp
pedberg-icu: @srl295 I decided to revert the use of \u202F in gDefaultPattern, so now the only thing that has changed is yyyy -> y.

pedberg-icu:
@srl295 and @macchiati, for an exhaustive test failure (nothing to do with this PR) I had to add some logging and a logKnownIssue skip in FormattedStringBuilderTest::testInsertOverflow(). Also I reverted part of the change to gDefaultPattern in ICU4C SimpleDateFormat, so the only change there is now just yyyy -> y. Other that those two things, this PR is the same as what you previously approved, but needs re-approval. Thanks!

srl295 (approved):
diff lgtm

pedberg-icu:
Hmm, C: Linux Clang Exhaustive Tests (Ubuntu 18.04) still crashing in FormattedStringBuilderTest:testInsertOverflow(), will submit a separate PR with a fix for that.

pedberg-icu:
> Hmm, C: Linux Clang Exhaustive Tests (Ubuntu 18.04) still crashing in FormattedStringBuilderTest:testInsertOverflow(), will submit a separate PR with a fix for that.

The PR for that is https://github.com/unicode-org/icu/pull/2104

But the crash itself may be related to the following change in FormattedStringBuilder per https://unicode-org.atlassian.net/browse/ICU-22005 (Integer overflow leading to OOB/CHECK in icu_71::FormattedStringBuilder::prepareForInsertHelper): PR https://github.com/unicode-org/icu/pull/2070

PR #2118, opened 2022-06-22
ICU-22068 Cleanup inconsistent annotations between declarations and d…

cyndyishida:
…efinitions

This cleans up inconsistent annotations between declared APIs in headers
vs defined implementations in cpp's. This better ensures the API's
referenceable in headers represent what is exposed and defined in the
ultimate binary library's symbol table.

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22068
- [x] Tests included, if applicable (current regression tests should suffice) 

### DIFF
--- a/icu4c/source/common/utf_impl.cpp
+++ b/icu4c/source/common/utf_impl.cpp
@@ -55,7 +55,7 @@
  * -finish:
  * (BSR: Bit Scan Reverse, scans for a 1-bit, starting from the MSB)
  */
-extern "C" U_EXPORT const uint8_t
+U_CAPI const uint8_t
 utf8_countTrailBytes[256]={
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
--- a/icu4c/source/i18n/collation.cpp
+++ b/icu4c/source/i18n/collation.cpp
@@ -20,15 +20,6 @@
 
 U_NAMESPACE_BEGIN
 
-// Some compilers don't care if constants are defined in the .cpp file.
-// MS Visual C++ does not like it, but gcc requires it. clang does not care.
-#ifndef _MSC_VER
-const uint8_t Collation::LEVEL_SEPARATOR_BYTE;
-const uint8_t Collation::MERGE_SEPARATOR_BYTE;
-const uint32_t Collation::ONLY_TERTIARY_MASK;
-const uint32_t Collation::CASE_AND_TERTIARY_MASK;
-#endif
-
 uint32_t
 Collation::incTwoBytePrimaryByOffset(uint32_t basePrimary, UBool isCompressible, int32_t offset) {
     // Extract the second byte, minus the minimum byte value,
--- a/icu4c/source/i18n/collationtailoring.h
+++ b/icu4c/source/i18n/collationtailoring.h
@@ -24,6 +24,8 @@
 #include "collationsettings.h"
 #include "uhash.h"
 #include "umutex.h"
+#include "unifiedcache.h"
+ 
 
 struct UDataMemory;
 struct UResourceBundle;
@@ -105,6 +107,10 @@ struct U_I18N_API CollationCacheEntry : public SharedObject {
     const CollationTailoring *tailoring;
 };
 
+template<> U_I18N_API
+const CollationCacheEntry *
+LocaleCacheKey<CollationCacheEntry>::createObject(const void *creationContext,
+                                                  UErrorCode &errorCode) const;
 U_NAMESPACE_END
 
 #endif  // !UCONFIG_NO_COLLATION
--- a/icu4c/source/stubdata/stubdata.cpp
+++ b/icu4c/source/stubdata/stubdata.cpp
@@ -19,26 +19,8 @@
 *   for running the data building tools.
 *
 */
-#include "unicode/utypes.h"
-#include "unicode/udata.h"
-#include "unicode/uversion.h"
 
-
-typedef struct {
-    uint16_t headerSize;
-    uint8_t magic1, magic2;
-    UDataInfo info;
-    char padding[8];
-    uint32_t count, reserved;
-    /*
-    const struct {
-    const char *const name; 
-    const void *const data;
-    } toc[1];
-    */
-   int   fakeNameAndData[4];       /* TODO:  Change this header type from */
-                                   /*        pointerTOC to OffsetTOC.     */
-} ICU_Data_Header;
+#include "stubdata.h"
 
 extern "C" U_EXPORT const ICU_Data_Header U_ICUDATA_ENTRY_POINT = {
     32,          /* headerSize */
--- a/icu4c/source/common/filteredbrk.cpp
+++ b/icu4c/source/common/filteredbrk.cpp
@@ -58,7 +58,7 @@ static int32_t U_CALLCONV compareUnicodeString(UElement t1, UElement t2) {
 /**
  * A UVector which implements a set of strings.
  */
-class U_COMMON_API UStringSet : public UVector {
+class UStringSet : public UVector {
  public:
   UStringSet(UErrorCode &status) : UVector(uprv_deleteUObject,
                                            uhash_compareUnicodeString,
@@ -482,7 +482,7 @@ SimpleFilteredSentenceBreakIterator::last(void) {
 /**
  * Concrete implementation of builder class.
  */
-class U_COMMON_API SimpleFilteredBreakIteratorBuilder : public FilteredBreakIteratorBuilder {
+class SimpleFilteredBreakIteratorBuilder : public FilteredBreakIteratorBuilder {
 public:
   virtual ~SimpleFilteredBreakIteratorBuilder();
   SimpleFilteredBreakIteratorBuilder(const Locale &fromLocale, UErrorCode &status);
--- a/icu4c/source/common/unicode/utf_old.h
+++ b/icu4c/source/common/unicode/utf_old.h
@@ -293,10 +293,10 @@ typedef int32_t UTextOffset;
 #ifdef U_UTF8_IMPL
 // No forward declaration if compiling utf_impl.cpp, which defines utf8_countTrailBytes.
 #elif defined(U_STATIC_IMPLEMENTATION) || defined(U_COMMON_IMPLEMENTATION)
-U_CFUNC const uint8_t utf8_countTrailBytes[];
+U_CAPI const uint8_t utf8_countTrailBytes[];
 #else
-U_CFUNC U_IMPORT const uint8_t utf8_countTrailBytes[];    /* U_IMPORT2? */ /*U_IMPORT*/
-#endif
+U_CFUNC U_IMPORT const uint8_t utf8_countTrailBytes[];
+#endif 
 
 /**
  * Count the trail bytes for a UTF-8 lead byte.
--- a/icu4c/source/common/unifiedcache.cpp
+++ b/icu4c/source/common/unifiedcache.cpp
@@ -45,20 +45,20 @@ U_CDECL_END
 
 U_NAMESPACE_BEGIN
 
-U_CAPI int32_t U_EXPORT2
+int32_t U_EXPORT2
 ucache_hashKeys(const UHashTok key) {
     const CacheKeyBase *ckey = (const CacheKeyBase *) key.pointer;
     return ckey->hashCode();
 }
 
-U_CAPI UBool U_EXPORT2
+UBool U_EXPORT2
 ucache_compareKeys(const UHashTok key1, const UHashTok key2) {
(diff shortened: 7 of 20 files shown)


icu4c/source/common/utf_impl.cpp
```diff
@@ -55,7 +55,7 @@
  * -finish:
  * (BSR: Bit Scan Reverse, scans for a 1-bit, starting from the MSB)
  */
-extern "C" U_EXPORT const uint8_t
+U_CAPI const uint8_t
```
richgillam, line 58: Kind of curious about this one-- if this is just in the .cpp file, what's U_CAPI doing for you?  Is somebody else actually using this despite its not being in a header?

icu4c/source/i18n/collationtailoring.h
```diff
@@ -105,6 +107,10 @@ struct U_I18N_API CollationCacheEntry : public SharedObject {
     const CollationTailoring *tailoring;
 };
 
+template<> U_I18N_API
+const CollationCacheEntry *
+LocaleCacheKey<CollationCacheEntry>::createObject(const void *creationContext,
+                                                  UErrorCode &errorCode) const;
```
richgillam, line 113: Pardon my ignorance, but where is this function actually implemented?  What's going on here?

I see this happens in several other classes below, so the same question applies to all of them.  Is the idea that this doesn't actually _need_ to be implemented, but it _does_ need to be declared?  Or is the implementation getting inherited from somewhere?

richgillam (dismissed):
FWIW, this all looks good to me (modulo a couple questions).  I'm interested in what Markus has to say, though...

icu4c/source/i18n/collationtailoring.h
cyndyishida: Template specializations essentially tell the compiler that the code defines its own implementation for a specific type parameter (unlike template instantiations). This function itself is defined and exposed in https://github.com/unicode-org/icu/blob/main/icu4c/source/i18n/ucol_res.cpp#L139 

For a tool like TAPI, that attempts to discover all linkable symbols from headers, this function looks hidden. Declaring this function template specialization now resolves this. 

I thought it was reasonable to put these specializations next to the class declaration for the type parameter, in this case `CollationCacheEntry`.

icu4c/source/common/utf_impl.cpp
cyndyishida: I think `utf8_countTrailBytes` is used in tests. `U_CAPI` expands to `U_CFUNC U_EXPORT` which expands to `extern "C" U_EXPORT` so its the same annotation. I updated this to match with [the updated header declaration in this PR](https://github.com/unicode-org/icu/pull/2118/files#diff-5cc093838cf40a8ed1ba9f84d4c32e22789d1c77dd86209d366c02f683024a3fR296).

icu4c/source/i18n/collationtailoring.h
richgillam: Thanks for the pointers here; I don't know how I missed that function.  Obviously I was searching for the wrong thing.  Thanks for the explanation; your actual changes look fine.

icu4c/source/common/utf_impl.cpp
richgillam: I see.  Thanks for the explanation!

cyndyishida:
Waiting on @markusicu feedback before moving forward.

pedberg-icu (dismissed):

pedberg-icu:
@markusicu You probably want to have a look at this one.

icu4c/source/i18n/collation.cpp
```diff
@@ -20,15 +20,6 @@
...
 U_NAMESPACE_BEGIN
 
-// Some compilers don't care if constants are defined in the .cpp file.
-// MS Visual C++ does not like it, but gcc requires it. clang does not care.
-#ifndef _MSC_VER
-const uint8_t Collation::LEVEL_SEPARATOR_BYTE;
-const uint8_t Collation::MERGE_SEPARATOR_BYTE;
-const uint32_t Collation::ONLY_TERTIARY_MASK;
-const uint32_t Collation::CASE_AND_TERTIARY_MASK;
-#endif
```
markusicu, line 30: The comment says that gcc requires these definitions. Shouldn't we keep these then? On the other hand, the build bots seem happy.

icu4c/source/stubdata/stubdata.cpp
```diff
@@ -19,26 +19,8 @@
...
-    /*
-    const struct {
-    const char *const name; 
-    const void *const data;
-    } toc[1];
-    */
-   int   fakeNameAndData[4];       /* TODO:  Change this header type from */
-                                   /*        pointerTOC to OffsetTOC.     */
-} ICU_Data_Header;
+#include "stubdata.h"
```
markusicu, line 23: Why is it necessary to pull these definitions into a new header file that is only included from here?

icu4c/source/i18n/collation.cpp
cyndyishida: I don't understand that comment since these are defined in a header file unconditionally. https://github.com/unicode-org/icu/blob/3cefbd55c7aafbb29bc67aa9a27df9bf13293a5a/icu4c/source/i18n/collation.h#L32

icu4c/source/stubdata/stubdata.cpp
cyndyishida: `U_ICUDATA_ENTRY_POINT` is already exported and referenced when building `libicuuc`. My understanding is that all exported symbols should have a matching declaration in a header file. This is specifically a requirement for TAPI, which is enabled in Apple's ICU fork.

icu4c/source/i18n/collation.cpp
markusicu: C++ lets you declare class constants, and assign values, in the class declaration, often in a .h file. That goes together with a *definition*, normally in a .cpp file. It *at least used to be* that gcc required these while MSVC failed with these definitions. I see that our current build bots are green, but I am nervous about older MSVC versions. @jefgen ?

icu4c/source/stubdata/stubdata.cpp
markusicu: stubdata is an artificial construct. Its structure matches other definitions, and matches expectations by the runtime code, but the precise struct is unique. Moving it to a header file seems harmless, but also unnecessary, since no one else is including it.

I don't remember quite why this struct definition is unique. It's been a long time...

markusicu:
Is it possible to add "TAPI" to a build bot, so that these kinds of things don't backslide?

icu4c/source/i18n/collation.cpp
cyndyishida: If a value is assigned to `TERMINATOR_BYTE` in a header file, what is left to define in the implementation file?

markusicu: > If a value is assigned to `TERMINATOR_BYTE` in a header file, what is left to define in the implementation file?

The storage location. The value should mostly be inlined, but a compiler may generate code that refers to its storage address.

jefgen: > I see that our current build bots are green, but I am nervous about older MSVC versions.

Well, FWIW, in terms of older versions of the MSVC compiler I think that the version that ships with VS 2015 (19.0) would be the oldest that would work for ICU now that it requires and uses C++11.

For example, support for the new C++11 type `char16_t` was [added](https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170#c-compiler-features) in [VS 2015 (19.0)](https://en.cppreference.com/w/cpp/compiler_support/11) -- so that would be the minimum needed to compile and test ICU now.

It might be possible to hack/modify the code to compile with an older version of MSVC, but I think that would likely be "not supported" (at least for newer versions of the ICU library).

We do have a [build bot ](https://github.com/unicode-org/icu/blob/main/.ci-builds/.azure-pipelines.yml#L467)that uses the VS 2015 compiler toolchain, so if the builds are green, then this works on VS 2015.

markusicu: Thanks Jeff... Actually, I asked the wrong question :-(
Some version of MSVC failed with these definitions, but removing them altogether removes them for all compilers, even gcc which used to require them. So the real question is about old versions of *gcc*.

Maybe we are sufficiently covered. If this shows up again, then we should probably go full C++11 and change the declarations to using `constexpr` not merely `const`.

icu4c/source/stubdata/stubdata.cpp
jefgen: > ... why this struct definition is unique...

FWIW, I found an older ticket that might give some insight: https://unicode-org.atlassian.net/browse/ICU-8043

From the ticket (emphasis added):

> U_ICUDATA_ENTRY_POINT is defined at 2 different places as 2 differents structure instance. It prevents a compiler (Visual C++) from inlining across different compilation units.

> The two mismatched struct types in question, one is a data header only (the declaration that the library code sees) and one is the same data header followed by some canned data.  **That canned data won't be present in all cases, so I don't want to make its declaration visible to the main library code.**

cyndyishida: Thanks for the background! If I understand correctly, this object is declared additionally https://github.com/unicode-org/icu/blob/85705f04e03f9cb41d4406bcd38c3e419eb7889d/icu4c/source/common/udata.cpp#L646 and also referenced in that file. 
I would naively suggest that this is where `stubdata.h` should be included. Attempting to do so reflects 
```
/icu/icuSources/common/udata.cpp:702:52: error: comparison of distinct pointer types ('const DataHeader *' and 'const ICU_Data_Header *')
                if(gCommonICUDataArray[i]->pHeader == &U_ICUDATA_ENTRY_POINT) {
````

However `ICU_Data_Header` ends up being the type of `U_ICUDATA_ENTRY_POINT` at runtime since it links against subdata's definition (at least on Darwin). It sounds like it's intentionally this way, but it seems a bit concerning that the compiler is tricked, and sees a different type for the object.

cyndyishida:
> Is it possible to add "TAPI" to a build bot, so that these kinds of things don't backslide?

I would love to add TAPI to ICU's CI! Unfortunately the tool itself only works out of the box with Xcode. The tool hasn't been completely upstreamed as apart of LLVM (slowly working on this). I'd love to touch base again after we've finished open sourcing TAPI as apart of LLVM's toolchain to get this supported.

icu4c/source/stubdata/stubdata.cpp
markusicu: > However `ICU_Data_Header` ends up being the type of `U_ICUDATA_ENTRY_POINT` at runtime since it links against subdata's definition (at least on Darwin). It sounds like it's intentionally this way, but it seems a bit concerning that the compiler is tricked, and sees a different type for the object.

It's a semi-arbitrary type for an engineered blob of bytes. The data format is variable-width to accommodate anything that we throw at it. When you build the ICU data into a shared library, then you link with that instead of with the stubdata. The contents of the shared library changes with the input data. The shared library itself can be created directly from a .o file (at least on Linux with ELF objects, and similarly on Windows), where there is no type info at all, just a linker symbol pointing to the bytes, or it can be created from a C or assembler source file.

https://unicode-org.github.io/icu/userguide/icu_data/#data-in-shared-librariesdlls-vs-dat-package-files

The compiler should not care about the exact C++ data type. If it does, it's best to silence that for the data entry point.

icu4c/source/i18n/collation.cpp
cyndyishida: Marking this as resolved, since it sounds like its alright to remove this moving forward.

icu4c/source/stubdata/stubdata.cpp
cyndyishida: I see, thanks for the explanation. Though in principle I believe its good to have declarations in headers reflect what is exposed in a binary's symbol table, I don't feel strongly about keeping this header if it will just become a maintenance burden without a clear benefit. Maybe when/if ICU adopts TAPI validation this could be revisited. What do you think is best for moving forward with this PR?

markusicu:
Also, there is a merge conflict. Please rebase & resolve.

icu4c/source/stubdata/stubdata.cpp
cyndyishida: I'll update with a comment to specify this situation. 

> If it's possible to add a line that silences the "declarations should be in a header file" warning for the struct in stubdata.cpp, then that would be best.

This is possible with TAPI, but the true power of tapi is using these header declarations to generate artifacts that can be directly consumed by our system linker (ld64). This allows for increased parallelization across shared object dependencies. Demoting this error into a warning means this symbol would no longer be reference-able when that artifact exists.

cyndyishida: Added in: https://github.com/unicode-org/icu/pull/2118/files#diff-dfab250a2804b93e2e4135dfc82d4df6f5f2d55e5490379d2af12cbe962602f4R11

markusicu (approved):
lgtm tnx

PR #2119, opened 2022-06-23
ICU-22072 Update Darwin Specific Macros

cyndyishida:
* Update `U_PF_IPHONE` to be 0 when building for macOS/macCatalyst.
* add macro definition for `attribute((visibility("hidden")))` for cases
  where internal structs exist within exposed classes.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22072
- [X] Tests included, if applicable (current tests should suffice)

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -168,7 +168,7 @@
 #   define U_PLATFORM U_PF_LINUX
 #elif defined(__APPLE__) && defined(__MACH__)
 #   include <TargetConditionals.h>
-#   if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  /* variant of TARGET_OS_MAC */
+#   if (defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE) && (defined(TARGET_OS_MACCATALYST) && !TARGET_OS_MACCATALYST)   /* variant of TARGET_OS_MAC */
 #       define U_PLATFORM U_PF_IPHONE
 #   else
 #       define U_PLATFORM U_PF_DARWIN
@@ -845,6 +845,14 @@ namespace std {
 #   define U_IMPORT __declspec(dllimport)
 #else
 #   define U_IMPORT 
+#endif
+
+#ifdef U_HIDDEN
+    /* Use the predefined value. */
+#elif defined(__GNUC__)
+#   define U_HIDDEN __attribute__((visibility("hidden")))
+#else
+#   define U_HIDDEN 
 #endif
 
 /**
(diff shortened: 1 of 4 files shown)


richgillam (approved):
This all looks good to *me*, but you might want to wait for Peter to weigh in to be sure.

I like the U_HIDDEN thing as a way of hiding the internal structs we were discussing before.  And the U_PF_IPHONE thing seems completely reasonable, too, as long as we don't have anybody who was depending on the old behavior.

pedberg-icu (approved):

icu4c/source/common/unicode/platform.h
```diff
@@ -847,6 +847,14 @@ namespace std {
 #   define U_IMPORT 
 #endif
 
+#ifdef U_HIDDEN
```
markusicu, line 850: This is in a public API header, so it's public API unless we say otherwise.

It needs a comment block like the other definitions here (with a `\def U_HIDDEN` line to connect it with whatever is the active definition), and it should probably be marked as `@internal`.

icu4c/source/i18n/unicode/dtitvinf.h
```diff
@@ -343,7 +343,7 @@ class U_I18N_API DateIntervalInfo U_FINAL : public UObject {
     /**
      * Internal struct used to load resource bundle data.
      */
-    struct DateIntervalSink;
+    struct U_HIDDEN DateIntervalSink;
```
markusicu, line 346: Why is this important? It's a forward declaration in the private section of the class.
Same for U_HIDDEN in the other classes.

icu4c/source/common/unicode/platform.h
```diff
@@ -168,7 +168,7 @@
 #   define U_PLATFORM U_PF_LINUX
 #elif defined(__APPLE__) && defined(__MACH__)
 #   include <TargetConditionals.h>
-#   if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  /* variant of TARGET_OS_MAC */
+#   if (defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE) && (defined(TARGET_OS_MACCATALYST) && !TARGET_OS_MACCATALYST)   /* variant of TARGET_OS_MAC */
```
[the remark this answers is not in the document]
cyndyishida: No, my intention is `U_PF_DARWIN` to be set when building for maccatalyst since that slice has to match with the macOS slice and maccatalyst triple also sets  `TARGET_OS_IPHONE`. 
Its a little confusing but `TARGET_OS_IPHONE != TARGET_OS_IOS`. 

MacCatalyst basically means the same macOS binary can run on macOS and iOS but there are specific api differences already set in ICU. Internally this change exists and hasn't caused issues so those differences may not be exposed to clients. The other option is to make MacOS == iOS == MacCatalyst so there is no need specify Macro distinctions.

icu4c/source/common/unicode/platform.h
pedberg-icu: Made the above fix in https://github.com/unicode-org/icu/pull/2201

PR #2123, opened 2022-06-25
ICU-22073 Do not throw away CompactDecimalFormat's affixes

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22073

### DIFF
--- a/icu4c/source/i18n/number_formatimpl.cpp
+++ b/icu4c/source/i18n/number_formatimpl.cpp
@@ -353,7 +353,9 @@ NumberFormatterImpl::macrosToMicroGenerator(const MacroProps& macros, bool safe,
     }
     fPatternModifier.adoptInstead(patternModifier);
     const AffixPatternProvider* affixProvider =
-        macros.affixProvider != nullptr
+        macros.affixProvider != nullptr && (
+                // For more information on this condition, see ICU-22073
+                !isCompactNotation || isCurrency == macros.affixProvider->hasCurrencySign())
             ? macros.affixProvider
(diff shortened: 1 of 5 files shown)


macchiati (dismissed):

macchiati:
Thanks!!

sffc:
I pushed an extra test demonstrating that it works with CompactCustomData. Note that I needed to pass the third argument to the pattern parser:

```java
                PatternStringParser.parseToExistingProperties(
                    "0 foo", props, PatternStringParser.IGNORE_ROUNDING_ALWAYS);
```

Otherwise the pattern started overriding the rounding behavior, which is undesirable.

macchiati:
Great, I'll add that to the CLDR code.

Note: am now debugging another CompactNumber bug:
https://unicode-org.atlassian.net/browse/CLDR-15762

"Upon review of https://st.unicode.org/cldr-apps/v#r_compact/cs// our Czech vetter noted that some of the plural formats are incorrect. For example, “1,1 milionů” should be “1,1 milionu”, as directed by https://st.unicode.org/cldr-apps/v#/cs/Compact_Decimal_Formatting/6cbd4b7d742691c (“many”). 

It appears that the examples in the report do not take the appropriate plural rules into account."

I looked into it, and verified that "1,1 milionu" (pattern for 'many') should be generated, not “1,1 milionů” (pattern for 'other'). So I'll try to find out where the wrong answer is coming from.

macchiati:
I fixed the 1,1 milionů problem on the CLDR side, so that doesn't need ICU changes.

macchiati (approved):

PR #2135, opened 2022-07-15
ICU-21939 Fix bogus "conflicting fields" error in DateIntervalFormat.

richgillam:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21939


richgillam:
The only substantive change here is on the Java side; the C side wasn't showing this error, but I went ahead and added a unit test for it just in case.

On the Java side, I think the error checking was just wrong; I simply changed it to allow "r" and "y" to coexist in the pattern.  On the C side, the code that was throwing the error on the Java side simply doesn't exist.  I don't know if we want to port those checks over from Java to C++, get rid of them altogether on the Java side, or leave things alone.  I took the path of least resistance and maintained the status quo.

FrankYFTang (approved):
LGTM

PR #2138, opened 2022-07-21
ICU-22075 Adds a random waiting period (0 - 179 secs) to each test th…

gnrunge:
…at runs as

part of a high concurrency performance test setup. This will disperse commits
of performance test results into the data branch over a wider time frame and
drastically reduces the chance of data uploads failing after ten unsuccesful
commit attempts.

Test runs showed a huge drop in failed commits/retry, from a high of 113
without wait down to only 4 with the extra wait.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22075

### DIFF
--- a/.github/workflows/icu_merge_ci.yml
+++ b/.github/workflows/icu_merge_ci.yml
@@ -363,6 +363,10 @@ jobs:
           ant perf-tests;
           cd perf-tests;
           mkdir -p perf/results/j_decimalformatperf/${{ matrix.locale }}/${{ matrix.perf }};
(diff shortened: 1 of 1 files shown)


markusicu (approved):
looks... hacky but plausible
anyone opposed?

echeran (dismissed):
It is hacky, but not opposed, on the condition that if the problem reoccurs too much in the future, then we replace this fix with a solution that directly addresses the original problem.

echeran (approved):

PR #2139, opened 2022-07-23
ICU-22088 Various fixes to make dealing with NUMBERING_SYSTEM formatt…

richgillam:
…ers easier.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22088


icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -2489,6 +2490,21 @@ IntlTestRBNF::doLenientParseTest(RuleBasedNumberFormat* formatter, const char* t
     }
 }
 
+void
+IntlTestRBNF::TestNumberingSystem() {
```
richgillam, line 2494: NOTE: This test is not duplicated on the Java side because there is no NUMBERING_FORMAT_STYLE (or any other RBNF-related style constants) in NumberFormat, so you can't pass those types to NumberFormat.getInstance() like you can in C.  (I checked, and the style parameter is an int, but the relevant constants on RuleBasedNumberFormat don't have distinct numeric values from the ones on NumberFormat, so they're clearly not intended to be used with NumberFormat.getInstance() ).

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1839,4 +1839,14 @@ public void testRoundingUnrealNumbers() {
         assertEquals("infinity", rbnf.format(Double.POSITIVE_INFINITY));
         assertEquals("not a number", rbnf.format(Double.NaN));
     }
+
+    @Test
+    public void TestNumberingSystem() {
+        RuleBasedNumberFormat rbnf = new RuleBasedNumberFormat(ULocale.US, RuleBasedNumberFormat.NUMBERING_SYSTEM);
+
+        assertEquals("Wrong result with default rule set", "123", rbnf.format(123));
```
richgillam, line 1847: This test currently fails because I didn't know how to rebuild icudata.jar with the rule change I made in icu4c/source/data/rbnf/root.txt.  If somebody can clue me in, I'll fix this.

icu4c/source/data/rbnf/root.txt
```diff
@@ -661,6 +661,8 @@ root{
             "%%tamil-thousands:",
             "0: =%tamil=;",
             "1000: <<\u0BF2[>>];",
+            "%zDefault:",
+            "0: =#,##0=;",
```
richgillam, line 665: Three things: 1) I'm still not sure this is the right thing to do.  It produces a more meaningful result if you forget to set the default result set on the NumberingSystem formatter, but that might make it harder to tell what you've done wrong (or that you've done anything wrong) if you get this in your app.  2) I didn't know how to get this change into icudata.jar for use on the Java side.  3) Do we also need to make this change in CLDR, and if we don't, how will we keep it here with future integrations of data from CLDR?

richgillam (commented):
I have a few questions for the reviewers:

icu4c/source/data/rbnf/root.txt
pedberg-icu: Yes, there should be a CLDR ticket for this, all RBNF data comes from CLDR. As for moving this to the ICU4J jars, the procedure for that is part of icu4c/source/data/cldr-icu-readme.txt

richgillam: Thank you.  I think I've fixed `icudata.jar` now, and I filed https://unicode-org.atlassian.net/browse/CLDR-15863 for the CLDR side of the fix.

macchiati (commented):
LGTM, but I'd feel better if Shane reviewed.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
sffc: Looks like you fixed it?

icu4c/source/i18n/unum.cpp
```diff
@@ -112,10 +113,6 @@ unum_open(  UNumberFormatStyle    style,
     case UNUM_DURATION:
         retVal = new RuleBasedNumberFormat(URBNF_DURATION, Locale(locale), *status);
         break;
-
-    case UNUM_NUMBERING_SYSTEM:
-        retVal = new RuleBasedNumberFormat(URBNF_NUMBERING_SYSTEM, Locale(locale), *status);
```
sffc, line 117: Suggestion: I think it would ideally be better to "fix" this in the RuleBasedNumberFormat constructor itself if possible. It seems like it could be disruptive if `unum_open` started returning something that might be a DecimalFormat instead of a RuleBasedNumberFormat.

icu4c/source/i18n/numfmt.cpp
```diff
@@ -1310,6 +1310,14 @@ NumberFormat::makeInstance(const Locale& desiredLocale,
...
         return NULL;
     }
+    
+    // For the purposes of general number formatting, UNUM_NUMBERING_SYSTEM should behave the same
+    // was as UNUM_DECIMAL.  In both cases, you get either a DecimalFormat or a RuleBasedNumberFormat
+    // depending on the locale's numbering system (either the default one for the locale or a specific
+    // one specified by using the "@numbers=" or "-u-nu-" parameter in the locale ID.
+    if (style == UNUM_NUMBERING_SYSTEM) {
+        style = UNUM_DECIMAL;
+    }
```
sffc, line 1320: As suggested below: I feel like this is a bug in the RBNF constructor, not in the `unum_open` or `NumberFormat::makeInstance` wrappers. I'm a little worried about changing around the old behavior.

sffc (commented):
It seems strange that we're essentially making `UNUM_NUMBERING_SYSTEM` behave the same as `UNUM_DECIMAL`. Surely the original author added the two separate options because they are intended to be used differently.

So while I think we should fix the odd bug, the high-order bit for me is that we provide proper documentation, e.g., pointing people toward `rbnf.setDefaultRuleSet("%ethiopic")`.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/RbnfTest.java
richgillam: I did fix it.  Thanks.

icu4c/source/i18n/unum.cpp
richgillam: As I thought I explained in the JIRA ticket, I don't think there's a good way to do that.  There are two fundamental problems: 1) A decent number of numbering systems are not implemented with a `RuleBasedNumberFormat`, and 2) there's no way to set the default rule set of an RBNF through the C interface.

So if you ask the RBNF constructor for a numbering system that's not RBNF-backed, you have two choices: The constructor can fail, which seems unnecessarily user-hostile, or you can get a rule set that consists of `0: =#,##0=;`, which is kludgey and may or may not produce the right thing (I put that in anyway, but I'm still not sure I think it's a good idea).

If we leave the RBNF constructor as it is and rely on the caller calling `setDefaultRuleSet()`, the caller has to know the name of the rule set.  I don't believe they're documented or guaranteed to be stable right now, and I'm not sure we want to change that.

It seems safer to say that you have to go through `NumberFormat::getInstance()` or `unum_open()` if you want to pick up the numbering system from the locale ID, and to leave the RBNF constructor as an implementation detail (in fact, if you're working in C, you have no choice).  That leaves us with the question of what the difference is between `UNUM_DECIMAL` and `UNUM_NUMBERING_SYSTEM`.  I agree it's weird to have them do the same thing, but I don't know what the original intent was or whether it really holds up at this point.  Frankly, it's equally weird to have `UNUM_DECIMAL` sometimes return something that's not a `DecimalFormat`.

>So while I think we should fix the odd bug...

Not sure what you meant by "the odd bug" here.  The fact that you get Tamil if you don't specify something else?

>...the high-order bit for me is that we provide proper documentation, e.g., pointing people toward rbnf.setDefaultRuleSet("%ethiopic").

I thought I did that.  There are additional places in the docs where I need to add verbiage?

sffc: OK, I'm convinced.

> I agree it's weird to have them do the same thing, but I don't know what the original intent was or whether it really holds up at this point.

OK.

> I thought I did that. There are additional places in the docs where I need to add verbiage?

Actually your docs look good and I didn't see all of them on my first pass.

sffc (approved):

icu4c/source/i18n/unum.cpp
richgillam: Thank you!  That's a relief.

richgillam:
@pedberg-icu , @macchiati , @markusicu , before I squash and merge, do any of you want to take another crack at this, or are we good to go?

pedberg-icu (approved):

pedberg-icu:
> @pedberg-icu , @macchiati , @markusicu , before I squash and merge, do any of you want to take another crack at this, or are we good to go?

I took another looks, seems good!

pedberg-icu (approved):

PR #2147, opened 2022-08-04
ICU-21125 Improvements to resource fallback:

richgillam:
- Added code to use the parentLocales data in supplementalData.xml to determine the "parent locale ID" to use when
  the requested resource bundle is not present (ICU-21126).
- Added code to change the parent-chain search path to handle the script better (ICU-21125; algorithm was described
  in CLDR-15265):
  - The base search patch is now ll_Ssss_RR -> ll_RR -> ll_Ssss -> ll -> root
  - If the requested script is not the default script for the requested language and region, we automatically
    avoid fallbacks that will implicitly change the script.
- Added new code to the CLDR-to-ICU data generation tool to generate source code, and used it to generate the lookup
  tables for the new resource-fallback logic (we can't use the existing resource files for this, since that would
  involve opening a resource bundle while trying to open another resource bundle).  The data-generation stuff is
  intended to be generic enough to allow for us to generate more static data tables in the future.
- Commented out a few collator tests, and changed one resource bundle tests, because they're incompatible with the
  new fallback logic (specifically, the default-script logic).

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21125


icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
```diff
@@ -242,25 +242,25 @@ void checkService(String requestedLocale, ServiceFacade svc,
...
-                    {"Norvégien (latin)", "Norsk (latinsk)", "no", "no_Latn"},
-                },
+//                {{"iw", "iw-US", "no", "no-Cyrl", "in", "in-YU"},
+//                    {"Hébreu (États-Unis)", "עברית (ארצות הברית)", "iw_US", "iw_US"},
+//                    {"Hébreu (Israël)", "עברית (ישראל)", "iw", "iw_IL"},
+//                    {"Indonésien (Indonésie)", "Indonesia (Indonesia)", "in", "in_ID"},
+//                    {"Indonésien (Serbie)", "Indonesia (Serbia)", "in_YU", "in_YU"},
+//                    {"Norvégien (cyrillique)", "Norsk (kyrillisk)", "no_Cyrl", "no_Cyrl"},
+//                    {"Norvégien (latin)", "Norsk (latinsk)", "no", "no_Latn"},
+//                },
```
richgillam, line 263: I want to specifically point out my changes to this file, where I simply commented out several tests that were failing.  The tests depend on the old locale-fallback behavior ("just lop components off the end") and don't work with the new default-script logic.  I couldn't figure out enough about how these tests worked to know what a more appropriate fix was, and am hoping somebody familiar with the Java Collator stuff can guide me.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
```diff
@@ -357,7 +357,7 @@ public void TestCurrencyDisplayNames() {
         assertNull("de_FOO_plural_foo", cdn.getPluralName("FOO", "foo"));
 
         // unknown locale with no substitute
-        cdn = CurrencyDisplayNames.getInstance(ULocale.forLanguageTag("zz-Gggg-YY"), true);
+        cdn = CurrencyDisplayNames.getInstance(ULocale.forLanguageTag("zz-YY"), true);
```
richgillam, line 360: I had to take the script code out of both instances of "zz-Gggg-YY" in this file-- thanks to the new default-script logic, these would fall straight back to "root", rather than to the system default locale, so I fixed it by removing the script code.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ICUResourceBundleTest.java
```diff
@@ -1141,4 +1141,32 @@ public void TestUResourceBundleCoverage() {
...
     }
+
+    @Test
+    public void TestAlgorithmicParentFallback() {
+        // Test for ICU-21125 and ICU-21126 -- cases where resource fallback isn't determined by lopping fields off
+        // the end of the locale ID (or following a %%Parent directive in a resource bundle)
+        // first column is input locale, second column is expected output locale
+        String[][] testCases = {
+            { "de_Latn_LI", "de_LI" },
+//            { "en_VA", "en_150" },  // put this back in after https://unicode-org.atlassian.net/browse/CLDR-15893 is fixed
```
richgillam, line 1152: CLDR-15893 is asking to add a bunch of new English and Spanish entries to the parent-locale table.  I couldn't find an existing locale ID that would test the parent-locale-table behavior-- all of the ones that might have actual resource bundles.

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
macchiati: Hmmm. You could see where the expected value != Actual value, see if the actual value is what you expect, and modify the test. Or did you try that?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
macchiati: Ok

macchiati (dismissed):
LGTM. My only question is whether the list of pairs of strings is the best days structure

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,172 @@ static UBool chopLocale(char *name) {
...
 
+static UBool hasScript(const char* localeID) {
+    UErrorCode err = U_ZERO_ERROR;
+    int32_t scriptLength = uloc_getScript(localeID, NULL, 0, &err);
+    return scriptLength != 0;
+}
+
+static UBool hasRegion(const char* localeID) {
+    UErrorCode err = U_ZERO_ERROR;
+    int32_t scriptLength = uloc_getCountry(localeID, NULL, 0, &err);
```
pedberg-icu, line 102: scriptLength is confusing. Maybe regionLength?

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,172 @@ static UBool chopLocale(char *name) {
...
+    return scriptLength != 0;
+}
+
+static UBool hasVariant(const char* localeID) {
+    UErrorCode err = U_ZERO_ERROR;
+    int32_t scriptLength = uloc_getVariant(localeID, NULL, 0, &err);
```
pedberg-icu, line 108: maybe variantLength, not scriptLength?

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,172 @@ static UBool chopLocale(char *name) {
...
+
+static void removeScript(const char* localeID, char* localeIDWithoutScript) {
+    char workingLocaleID[ULOC_FULLNAME_CAPACITY];
+    char fieldBuffer[ULOC_FULLNAME_CAPACITY];
+    int32_t fieldLength = 0;
+    UErrorCode err = U_ZERO_ERROR;
+    
+    *localeIDWithoutScript = '\0';
+    uloc_getLanguage(localeID, fieldBuffer, ULOC_FULLNAME_CAPACITY, &err);
+    uprv_strcpy(workingLocaleID, fieldBuffer);
```
pedberg-icu, line 149: Do we need to check err here? What if someone passes in a localeID that is longer than ULOC_FULLNAME_CAPACITY, or one that is malformed in some way? Is that checked at a higher level?

pedberg-icu (dismissed):
Looks pretty good to me, I especially like all of the CodeGenerator stuff in tools/, that will be very useful. I had some concerns about protection against array overwriting in C...

markusicu:
Sorry for being late, looking...

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
markusicu: This, and the last bullet in the description (“Commented out a few collator tests ... because they're incompatible with the
new fallback logic (specifically, the default-script logic).”) doesn't sound right.

We use different fallback behavior for different “services” / “locale trees”.

For normal locale trees (the main one, curr, lang, zone, ...), we have normal CLDR fallbacks where we do things like skip over a truncation locale that has a different script. For a long time, we have done this via data in the resource bundles using `%%Parent` entries, because we didn't want to hardcode the CLDR fallback behavior. Example: https://github.com/unicode-org/icu/blob/main/icu4c/source/data/locales/zh_Hant.txt gets generated with `%%Parent{"root"}`

As a result, users can write their own resource bundles with their own parent data, and customize ICU data fallbacks by modifying the `%%Parent` entries.

For some locale trees, at least the `coll` tree, we just use truncation fallbacks, because we are not concerned about going to a different script -- the data is used for processing text, not for displaying stuff. Example: https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/zh_Hant.txt has no `%%Parent` entry. It has almost no data at all: It only overrides the default collation type (there can be multiple tailorings per language), and *falls back to zh for the actual tailoring data*.

This is specified in the LDML spec: https://www.unicode.org/charts/collation/chart_Ignored.html see “Collation data, however, is an exception. ...”

The CLDR-to-ICU converter clearly has a switch for generating the parent locale data for some trees but not others.

And of course we have supplemental data files that use no fallback at all.

Now, I know that we decided that there are benefits from hardcoding the CLDR parent locale data and fallback behavior, and we have essentially decided that CLDR is now an industry-wide standard, and we don't really know of anyone writing ICU-format resource bundles for their application data. But we need to apply this fallback mechanism only for data trees where this makes sense.

Note that we have an `enum OpenType`, for example in ICUResourceBundle.java. So far, we have used it for other distinctions, such as not using the default locale for `coll`, `brkitr`, and `translit`, see https://www.unicode.org/reports/tr35/tr35.html#411-bundle-vs-item-lookup “The default locale is not even always used ...”

For these trees, we should be loading bundles with OpenType.LOCALE_ROOT.

We could add a new value like LOCALE_ROOT_NO_PARENTS to skip the newly hardcoded CLDR parent locale fallbacks for collation.

Actually, maybe we should treat coll/brkitr/translit the same for both no-default and no-parents. If so, then we could just use LOCALE_ROOT for both. Indeed, https://github.com/unicode-org/icu/blob/main/icu4c/source/data/brkitr/zh_Hant.txt also has no `%%Parent` entry. The translit bundles don't use locale IDs for their names, so the usual CLDR fallback makes no sense there anyway. We should update the LDML spec to align the types of data on no-default = no-parents.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
markusicu: Is this right? It seems like zz-Gggg should skip zz and *go to the next thing in the fallback chain*, which for the "normal" locale trees would be the default locale.

markusicu (changes requested):
Sorry for being both late and a bit negative... I just went through the discussion, and the behavior doesn't sound right.

I will read through the code today.

icu4c/source/common/localeFallbackTables.h
```diff
@@ -0,0 +1,629 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+//
+// Internal static data tables used by uresbund.cpp
+// WARNING: This file is mechanically generated by the CLDR-to-ICU tool
```
markusicu, line 5: Please change the name of this file so that it ends with ..._data.h like we have with some existing generated files.

Also, since this is intended to be text-included by one single .cpp file, we usually don't make it a stand-alone header file (with its own includes, but which would also want an include-only-once guard), and instead guard it against inclusion by anyone else (such as the header test script).

Example: https://github.com/unicode-org/icu/blob/main/icu4c/source/common/propname_data.h
- filename suffix
- no includes of cstdint etc.
- yes include-by-one-cpp guard: `#ifdef INCLUDED_FROM_PROPNAME_CPP`

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
richgillam: The way I have this coded right now, no.  If the parent locale table says "root", I go to root, not to the default locale. Do we have consensus that this is the wrong thing to do, and I should go through the default locale in this case?

richgillam: Never mind.  I changed this, and it didn't hurt anything, so I'm just going to make this change.

icu4c/source/common/uresbund.cpp
richgillam: Bonehead copy/paste error.  Fixed.

icu4c/source/common/uresbund.cpp
richgillam: Yup.  Copy/paste error.  Fixed.

icu4c/source/common/uresbund.cpp
richgillam: This is a tough one.  All of my new code depends on the locale ID being well-formed, and I'm kind of relying on whatever existing mechanisms are present in the resource-bundle code to ensure that's true.  If we can be sure the existing resource-bundle code is giving us a well-formed locale ID, then what I have should be okay, since no external callers should be calling it.

But just in case, I went back and fixed this function to make sure the language and region codes are no longer than they're supposed to be, and to check the error codes coming back from both uloc_getLanguage() and uloc_getCountry().  If we get errors from either, this function will end up returning the original locale ID unchanged (i.e., it might still have a script code), which may cause additional problems, but at least we should be protected against writing off the end of the buffer.  Please take another look-- if you think there's a better way to do this, I'm all for it.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
markusicu: I was actually questioning myself on this one. The current behavior for when we have the data, like `%%Parent{"root"}`, is to go straight to root, and not via the default locale, right? So what you did was equivalent, just extending it to all locales for which CLDR has data, or where the CLDR algorithm holds (skip the language if it implies a different script). Sorry if I am causing you to flip-flop.

Does anyone else want to chime in here?

richgillam: What I originally did is what you're describing here, and it broke those tests.  To be honest, I'm not sure what the Right Answer is supposed to be in this situation, but if the `%%Parent` directives in the resource bundles themselves send us straight to `root` , that's an argument for what I originally had (and changing the tests in some way).  Unless we tend to believe the `%%Parent` directives were doing the wrong thing...

I think you're right-- I'd like other people to chime in and for us to come to some kind of consensus as to what the proper behavior here is.  The change I made is easy enough to undo, but I don't want to have to redo it again if the discussion goes that way.  Is carrying on the discussion here good enough, or should I sent out an email?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
```diff
@@ -0,0 +1,93 @@
...
+    public void execute() throws BuildException {
+        for (GeneratedFileDef task : generatedFileDefs) {
+            File cOutPath = new File(cOutDir, task.cRelativePath);
+            File javaOutPath = new File(javaOutDir, task.javaRelativePath);
+
+            if (this.action != null && this.action.equals("clean")) {
+                cOutPath.delete();
+                javaOutPath.delete();
+            } else {
+                log("Deleting " + cOutPath + " and " + javaOutPath + "...");
```
markusicu, line 73: It looks like this comment belongs into the previous "if" branch.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
```diff
@@ -0,0 +1,93 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+// Note: Auto-magical Ant methods are listed as "unused" by IDEs, unless the warning is suppressed.
+public final class GenerateCodeTask extends Task {
+    private Path cldrPath;
+    private File cOutDir;
+    private File javaOutDir;
```
markusicu, line 20: Curious: Why do you use both the newer Path and the ancient File?
https://www.baeldung.com/java-path-vs-file

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
```diff
@@ -0,0 +1,93 @@
...
+                cOutPath.delete();
+                javaOutPath.delete();
+            } else {
+                log("Deleting " + cOutPath + " and " + javaOutPath + "...");
+                cOutPath.getParentFile().mkdirs();
+                javaOutPath.getParentFile().mkdirs();
+
+                try {
+                    PrintWriter cOut = new PrintWriter(new OutputStreamWriter(new FileOutputStream(cOutPath)));
+                    PrintWriter javaOut = new PrintWriter(new OutputStreamWriter(new FileOutputStream(javaOutPath)));
```
markusicu, line 79: Could you please use try-with-resources?
- https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html
- https://www.baeldung.com/java-try-with-resources

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/generator/ResourceFallbackCodeGenerator.java
```diff
@@ -0,0 +1,181 @@
...
+    private void handleLikelySubtag(CldrValue value) {
+        String from = value.get(AttributeKey.keyOf("likelySubtag", "from"));
+        String to = value.get(AttributeKey.keyOf("likelySubtag", "to"));
+
+        String[] fromPieces = localeIDSplitter.splitToList(from).toArray(new String[] {});
+        String[] toPieces = localeIDSplitter.splitToList(to).toArray(new String[] {});
+
+        if (toPieces.length != 3) {
+            System.err.println("Didn't get 3 segments in 'to' value: from=" + from + ", to=" + to);
+            return;
```
markusicu, line 52: Why not throw an exception? What if someone runs the converter and misses the console output?

icu4j/main/classes/core/src/com/ibm/icu/impl/LocaleFallbackTables.java
```diff
@@ -0,0 +1,562 @@
...
+        { "zh_US", "Hant" },
+        { "zh_VN", "Hant" },
+        { "zhx", "Nshu" },
+        { "zkt", "Kits" },
+    };
+
+    //======================================================================
+    // Parent locale table
+    public static final String[][] PARENT_LOCALE_TABLE = {
+        { "az_Arab", "root" },
```
markusicu, line 387: Shouldn't we use "" instead of "root" ?

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1261,8 +1263,90 @@ private static boolean localeIDStartsWithLangSubtag(String localeID, String lang
...
+            if (getDefaultScript(language, region).equals(script)) {
+                return language + "_" + region;
+            } else {
+                return language + "_" + script;
+            }
+        } else if (!region.isEmpty()) {
+            // if "name" has region but not script, did the original locale ID specify a script?
+            // - if yes, replace the region with the script from the original locale ID
+            // - if no, replace the region with the default script for that language and region
+            String origNameScript = ULocale.getScript(origName);
```
markusicu, line 1327: It looks like there is a fair bit of conversion back & forth between String and ULocale.
Maybe add a TODO to optimize this later, such as using a struct like class LSR in the LocaleMatcher?

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1304,17 +1388,16 @@ public ICUResourceBundle load() {
 
             // fallback to locale ID parent
             if(b == null){
-                int i = localeName.lastIndexOf('_');
-                if (i != -1) {
-                    // Chop off the last underscore and the subtag after that.
-                    String temp = localeName.substring(0, i);
-                    b = instantiateBundle(baseName, temp, defaultID, root, openType);
+                String origLocaleName = (origLocaleID != null) ? origLocaleID : localeName;
+                String fallbackLocaleID = getParentLocaleID(localeName, origLocaleName);
```
markusicu, line 1392: This looks like the place where we should check the openType.
If it's LOCALE_DEFAULT_ROOT or LOCALE_ONLY then call getParentLocaleID().
Otherwise, use the old truncation; maybe wrap it into a little helper function that also returns null if there is no more underscore.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1304,17 +1388,16 @@ public ICUResourceBundle load() {
...
+                if (fallbackLocaleID != null) {
+                    b = instantiateBundle(baseName, fallbackLocaleID, origLocaleName, defaultID, root, openType);
                 }else{
                     // No underscore, only a base language subtag.
```
markusicu, line 1396: This comment is now out of date.

markusicu (commented):
I made it through a good read of the Java code.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1304,17 +1388,16 @@ public ICUResourceBundle load() {
...
             // fallback to locale ID parent
             if(b == null){
-                int i = localeName.lastIndexOf('_');
-                if (i != -1) {
-                    // Chop off the last underscore and the subtag after that.
-                    String temp = localeName.substring(0, i);
-                    b = instantiateBundle(baseName, temp, defaultID, root, openType);
+                String origLocaleName = (origLocaleID != null) ? origLocaleID : localeName;
+                String fallbackLocaleID = getParentLocaleID(localeName, origLocaleName);
+                if (fallbackLocaleID != null) {
```
markusicu, line 1393: I am not seeing enough context here... If there is an explicit `%%Parent`, is that handled at the right time? Probably before the CLDR parent fallback?

icu4c/source/common/localeFallbackTables.h
richgillam: At one point I needed the "include <cstdint>" to get the file to build, but now it seems to build fine without it.  🤷‍♂️  I question the need for the include guard, but I added it for consistency (and, of course, I renamed the file).  I made no corresponding changes on the Java side-- do I need to do something with the name of that generated file too?

markusicu: > I renamed the file

Would you mind making it all-lowercase? I don't remember us having any mixed-case .c/.cpp/.h filenames so far.

> I made no corresponding changes on the Java side-- do I need to do something with the name of that generated file too?

I thought about it, but we don't have an established convention in Java, since in general it is much less efficient in Java to hardcode data, and so we rarely do it. Maybe rename it to LocalFallbackData.java for a little more consistency?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
markusicu: > For resource bundle lookup we insert the default locale into the chain. That's basically saying that if we fail to find your locale, we'll try the system's default.

We really have two related algorithms here:
- Looking for the first (most specific) bundle for a given locale ID. If we find nothing at all, then we go through the default locale (for display-oriented services).
- Once we have a bundle, also load the parent locale bundle and its parent etc., down to root. This never goes through the default locale.

In ICU so far, we get the parent locale information from inside the bundles, which means that the two algorithms visit different parents when we are missing files. When we find zh_Hant, `%%Parent` sends us to root (and we have *not* been going to the default locale). When zh_Hant is missing, we truncate to zh.

With Rich's change, we should be able to remove the `%%Parent` entries, because they will be redundant with the CLDR parent locale data. I think this should also mean that we now build parent-locale chains (down from a found bundle) consistently like we look for the first (most specific) available bundle.

So if we find zh_Hant, then we skip zh, and only root is the parent in the runtime bundle chain.
If we don't find zh_Hant, then we also skip zh, and... I think load root, not default?
And if so, then if we don't find zz_Gggg, then we skip zz and also load root, not default?

markusicu: > https://unicode.org/reports/tr35/#Bundle_vs_Item_Lookup That's also different than inherited item lookup, where we are pulling in a particular item, and looking in the parent locale. For that, we don't want to involve the default locale at all, otherwise we can get a severe ransom-note effect.

The point is, for services like collation and break iteration, we don't go through the default locale *even in bundle lookup*.

markusicu:
Another thought about collation: Even though we don't/shouldn't use parent-locale data there, we should still get the other bundle-finding improvements. In particular, we have de_AT and fr_CA bundles. I am pretty sure that we won't find them if you open a collator for de_Latn_AT or fr_Latn_CA. This work here should fix that. In other words, we should not turn off all new code for the OpenType for coll & brkitr.

markusicu:
A thought about unit testing: You need not rely only on data generated from CLDR. If you want to test unusual scenarios, or make sure that the data is stable for testing, you can add source bundles with a small amount of custom-designed data in [icu4c/source/test/testdata](https://github.com/unicode-org/icu/tree/main/icu4c/source/test/testdata).

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ICUResourceBundleTest.java
```diff
@@ -1144,4 +1144,32 @@ public void TestUResourceBundleCoverage() {
...
+            // or sr_Latn to root, but all of these resource bundle files actually exist in the project
+        };
+
+        for (String[] testCase : testCases) {
+            ULocale locale = new ULocale(testCase[0]);
+            ICUResourceBundle rb = ICUResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME, locale, ICUResourceBundle.OpenType.LOCALE_ROOT);
+
+            String actual = rb.getULocale().toString();
+            if (testCase[1].equals("root")) {
+                assertEquals("Got wrong locale", "", actual);
```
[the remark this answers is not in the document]
richgillam: I thought "root" was clearer, but I went ahead and changed this.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
richgillam: You're right.  Must have been a copy/paste error.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
richgillam: Ignorance.  I didn't know how the new Path object worked and couldn't find methods on it that did what I wanted.  Thanks for the pointer to the article; that helped a lot.  I've updated this class to use Path everywhere.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/GenerateCodeTask.java
richgillam: Oh, that's slick.  I didn't know that existed either.  Fixed.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/generator/ResourceFallbackCodeGenerator.java
richgillam: I don't know.  I think that felt too drastic, and I wasn't sure when I wrote this that I was guaranteed that the likely-subtags resources would all have three-segment locale IDs (yeah, I know that's the whole point, but I was worried there might be exceptions to it).  In other words, it was as much about my confirming my own assumptions as anything.

If I change this to throw an exception, what exception should I throw?  And would I also do it on line 59 (the "'from' entry has a non-'und' language..." thing)?

icu4j/main/classes/core/src/com/ibm/icu/impl/LocaleFallbackTables.java
richgillam: I like "root."  It matches what's in the original resource files, and I think it makes it clearer what we're doing.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: The idea originally was to keep it all in the String domain, but the easiest way I could find to break the string up into language, script, region, and variant.  I think I considered LSR, but it didn't have that functionality.  If you've got a better way of doing this, I'm all ears.  I did go ahead and add a TODO.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: You're right.  Removed.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+static UBool hasVariant(const char* localeID) {
+    UErrorCode err = U_ZERO_ERROR;
+    int32_t variantLength = uloc_getVariant(localeID, NULL, 0, &err);
+    return variantLength != 0;
+}
+
+// This file contains the tables for doing locale fallback, which are generated
+// by the CLDR-to-ICU process directly from the CLDR data.  This file should only
+// ever be included from here.
+#include "localeFallbackTables.h"
```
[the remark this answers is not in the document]
richgillam: Already done.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+
+static const char* performFallbackLookup(const char* key,
+                                         const char* keyStrs,
+                                         const char* valueStrs,
+                                         const int32_t* lookupTable,
+                                         int32_t lookupTableSize) {
```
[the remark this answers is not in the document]
richgillam: Done.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+    const int32_t* bottom = lookupTable;
+    const int32_t* top = lookupTable + lookupTableSize;
+
+    while (bottom < top) {
+        const int32_t* middle = bottom + ((top - bottom) / 2);
+        middle -= (middle - lookupTable) % 2;
```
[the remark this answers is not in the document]
richgillam: >You could also write it directly with string literals, instead of the index indirection; I would hope that with enough const keywords the compiler will de-duplicate read-only string literals.

After I wrote this, I thought a lot about that too-- am I really gaining anything by doing all the string management myself-- wouldn't I get basically the same thing as what I'm getting right now by jumping through all these hoops?

Your comment and this thought of mine set off a giant wild goose chase that ate up most of my day and left me tearing my hair out.

Just getting rid of the separate `char*` arrays and making the main tables into arrays of `char*` looked good and worked fine, but blew up the size of the generated file by 13K, from 53K to 66K.  I'm not sure why.  I don't think it's because the compiler isn't deduping all those string literals-- I checked that in a small test app.  I think part of it may by that the main tables went from being arrays of 32-bit integers to arrays of 64-bit pointers, but I don't think those tables are big enough for a doubling in their size to add 13K.

I also couldn't figure out a good way to make the table into an array of two-entry arrays.  C++ apparently doesn't like this, or I simply couldn't figure out the declarator syntax.  I tried making the whole thing into an array of `std::pair`, but this gave me a "global constructor" warning-- according to StackOverflow, this is a warning that I'm introducing static-init code I didn't need before.  I was really trying to avoid this kind of thing, so I gave up, stuck with what I had, and just made the change you suggest above.

If somebody smarter than I wants to come up with some better here, great, but I'm reluctant at this point to throw good time after bad.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+        } else if (strcmpResult < 0) {
+            top = middle;
+        } else {
+            bottom = middle + 2;
+        }
+    }
+    return nullptr;
+}
+
+static void removeScript(const char* localeID, char* localeIDWithoutScript) {
```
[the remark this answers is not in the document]
richgillam: See below.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+    char workingLocaleID[ULOC_FULLNAME_CAPACITY];
+    char fieldBuffer[ULOC_FULLNAME_CAPACITY];
+    int32_t fieldLength = 0;
+    UErrorCode err = U_ZERO_ERROR;
+    
+    *localeIDWithoutScript = '\0';
+    uloc_getLanguage(localeID, fieldBuffer, ULOC_LANG_CAPACITY, &err);
```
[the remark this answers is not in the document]
richgillam: I had forgotten about CharString and didn't realize there were internal methods in uloc.cpp that returned CharStrings (although they were a little more complicated to use than I expected).  I rewrote this whole thing to use those methods and CharString and (incidentally) to look a lot more like the Java version of this algorithm.  I think it's a lot better now.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+    const char* parentID = performFallbackLookup(name, parentLocaleChars, parentLocaleChars, parentLocaleTable, UPRV_LENGTHOF(parentLocaleTable));
+    if (parentID != NULL) {
+        uprv_strcpy(name, parentID);
+        return TRUE;
+    }
+
+    // if it's not in the parent locale table, figure out the fallback script algorithmically
+    // (see CLDR-15265 for an explanation of the algorithm)
+    char defaultScript[ULOC_SCRIPT_CAPACITY];
+    char workingLocaleID[ULOC_FULLNAME_CAPACITY];
```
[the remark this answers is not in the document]
richgillam: See above.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+ *  Internal function, determines the search path for resource bundle files.
+ *  Currently, this function is used only by findFirstExisting() to help search for resource bundle files when a bundle for the specified
+ *  locale doesn't exist.  The code that supports inheritance of resources between existing resource bundle files continues to
+ *  use chopLocale() below.
+ *  @param name In-out parameter: On input, the locale ID to get a parent locale ID for (this is a locale's base name, without keywords); on output, the
+ *  requested parent locale ID.
+ *  @param origName The original locale ID the caller of findFirstExisting() requested.  This is the same as `name` on the first call to this function,
+ *  but as findFirstExisting() ascends the resource bundle's parent tree, this parameter will continue to be the original locale ID requested.
+ */
+static UBool getParentLocaleID(char *name, const char *origName) {
```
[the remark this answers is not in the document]
richgillam: You're right.  Fixed.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,173 @@ static UBool chopLocale(char *name) {
...
+    uloc_getScript(localeID, originalScript, ULOC_SCRIPT_CAPACITY, &err);
+    removeScript(localeID, localeIDWithoutScript);
+    
+    const char* newScript = performFallbackLookup(localeIDWithoutScript, dsLocaleIDChars, scriptCodeChars, defaultScriptTable, UPRV_LENGTHOF(defaultScriptTable));
+    if (newScript == NULL && uprv_strchr(localeIDWithoutScript, '_') != NULL) {
+        // we had language and country; chop off the country and try again
+        *(uprv_strchr(localeIDWithoutScript, '_')) = '\0';
+        newScript = performFallbackLookup(localeIDWithoutScript, dsLocaleIDChars, scriptCodeChars, defaultScriptTable, UPRV_LENGTHOF(defaultScriptTable));
+    }
+    if (newScript == NULL) {
```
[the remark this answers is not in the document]
richgillam: I rewrote all this, and I think in the process I managed to replace all the `NULL`s with `nullptr`s.  Please let me know if I missed any.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: Unless I'm badly mistaken, all of this code is about locating the first resource bundle file that matches the requested locale ID and actually exists.  You don't get to `%%Parent` until you find an actual resource file, and at that point you're not in this code anymore.

richgillam:
@macchiati 

>LGTM. My only question is whether the list of pairs of strings is the best days structure

In Java, right?  And you're talking about the new default-script and parent-locale tables, right?  I didn't work to come up with an optimized structure because I didn't think it'd make much of a difference.  What would you recommend instead?

icu4c/source/common/localeFallbackTables.h
richgillam: >Would you mind making it all-lowercase? I don't remember us having any mixed-case .c/.cpp/.h filenames so far.

Sure.  Done.

>I thought about it, but we don't have an established convention in Java, since in general it is much less efficient in Java to hardcode data, and so we rarely do it. Maybe rename it to LocalFallbackData.java for a little more consistency?

Sure, why not?  Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ICUResourceBundleTest.java
macchiati: We discussed that today, and were thinking that rather than adding a bunch of empty stub locales, it would be better to add parentLocale relationships. We left this in design because we wanted to talk to you and Peter about it.

icu4j/main/classes/core/src/com/ibm/icu/impl/LocaleFallbackData.java
```diff
@@ -0,0 +1,562 @@
...
+        { "zh_PF", "Hant" },
+        { "zh_PH", "Hant" },
+        { "zh_SR", "Hant" },
+        { "zh_TH", "Hant" },
+        { "zh_TW", "Hant" },
+        { "zh_US", "Hant" },
+        { "zh_VN", "Hant" },
+        { "zhx", "Nshu" },
+        { "zkt", "Kits" },
+    };
```
macchiati, line 382: About the string pairs. When Java has an array, the initialization code builds it item by item. If you are going to want a map anyway, and you are generating the code, then it is just as easy to avoid the extra work and generate the code to build the map in the first place. I'll phrase this as using an ImmutableMap builder.

OLD
    public static final String[][] DEFAULT_SCRIPT_TABLE = {
            { "ab", "Cyrl" },
...
            { "zkt", "Kits" },
    };

NEW
    public static final Map<String, String> DEFAULT_SCRIPT_MAP = ImmutableMap.<String, String>builder()
            .put("ab", "Cyrl")
...
            .put("zkt", "Kits")
            .build();

If ImmutableMap isn't available to you, then you can add a simple helper function:

    public static class ImmutableMapBuilder<K,V> {
        private Map<K,V> data;
        public ImmutableMapBuilder(Map<K,V> data) {
            this.data = data;
        }
        public ImmutableMapBuilder<K,V> put(K key, V value) {
            data.put(key, value);
            return this;
        }
        public Map<K, V> build() {
            return Collections.unmodifiableMap(data);
        }
    }

and change 
ImmutableMap.<String, String>builder() 
to 
new ImmutableMapBuilder<String,String>(new HashMap<>())

The other advantage is that an array in Java is mutable, so you really never want to expose it as public; that's not a problem if you just directly build an immutable map.

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
richgillam: >Actually, maybe we should treat coll/brkitr/translit the same for both no-default and no-parents. If so, then we could just use LOCALE_ROOT for both.

I thought about it, and I agree with this.  I changed the ICU4J code to check the open type and bypass the special parent lookup when it's something other than LOCALE_DEFAULT_ROOT (my implementation does still retain the special handling of the script field, as you recommended).  There are two problems with this, though:

1) The unit test failures I commented out still fail, so this change didn't fix that.

This is because that particular test isn't _doing_ parent fallback in the "coll" resource tree, even though it's in the collation tests.  That particular test is testing the _display name_ functionality on Collator.  That's the LocaleDisplayNames stuff, which is in the "locales" resource tree and DOES want the new parent-fallback stuff.  I could find a way to hack the collator stuff to use straight truncation fallback, but I don't think we want to do that; I think we need to fix the tests.  I didn't do that originally because the assumption of how the lookup works is baked so deeply into the design of that test, so I'm hoping you (or somebody else) can give me some advice on the best way to fix this test.  Or I can leave it commented out a filed another ticket.

2) Given that the unit test above doesn't actually exercise the new functionality, I'm not sure anything in our test suite does.  I can (and probably will) update my own new unit test to exercise it a little, but that won't really tell us whether this does what we want in real-world conditions.  Yes, I can add dummy resources over in testData.jar, but that's not really verifying under real-world conditions that it does what we want, either.  Any thoughts on this?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
richgillam: >For resource bundle lookup we insert the default locale into the chain. That's basically saying that if we fail to find your locale, we'll try the system's default.

I read this to mean the logic I have right now is correct, and when the parent locale table says "root" or we have a non-default script, we want to fall back through the default locale, instead of directly to root.  @macchiati and @markusicu , do you agree?

>With Rich's change, we should be able to remove the %%Parent entries, because they will be redundant with the CLDR parent locale data. I think this should also mean that we now build parent-locale chains (down from a found bundle) consistently like we look for the first (most specific) available bundle.

I don't actually think this is true, without more work.  My change enables us to get rid of resource bundles that have nothing in them but a `%%Parent` resource, but if a resource bundle has a `%%Parent` resource _and other resources_, we'd need extra code to make _inherited item lookup_ use the parent locale table instead of the `%%Parent` resources.  I think this is a good idea, but I also think this PR is getting really big, and we should handle that as a separate ticket after I land this one, maybe in a future version of ICU.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: Yup.

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
richgillam: To clarify what I'm asking for, `ULocaleCollationTest.TestNameList()` has a list of test cases, each of which has a list of locales, followed by a list of expected values for each of the locales.  The expected-values list has four expected values: name in French, name in self, and the minimized and maximized versions of the locale ID.  In the test cases I've commented out, I believe the problem is the "name in self" value.  If "self" doesn't exist and it uses a non-default script, like `fr_Cyrl_BE`, we fall all the way back to root (or, actually, to the system default locale-- when I run, I actually get "French (Cyrillic, Belgium)").  I guess I could just change the expected values in these tests to look for that, but that feels weird, especially since it'd depend on the default locale.  What do you all think?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ICUResourceBundleTest.java
richgillam: Well, that's certainly what I was proposing-- just add entries to the `parentLocales` resource.  I know that makes that resource a lot bigger and slower to search, so maybe we'd need to change the structure in ICU to preserve performance (or size), although I don't think that'd affect CLDR.  I also know Shane has concerns about just adding a bunch of new `parentLocales` entries without proper vetting to make sure we're falling back to the right things, and I agree, provided we have a good way to get good answers to those questions.

I'm game to discuss all this whenever.

icu4c/source/common/uresbund.cpp
```diff
@@ -494,8 +663,14 @@ findFirstExisting(const char* path, char* name, const char* defaultLocale,
         *isRoot = (UBool)(uprv_strcmp(name, kRootLocaleName) == 0);
 
         /*Fallback data stuff*/
-        *hasChopped = chopLocale(name);
-        if (*hasChopped && *name == '\0') {
+        if (!hasRealData) {
+            *foundParent = getParentLocaleID(name, origName);
```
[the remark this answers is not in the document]
richgillam: Done.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
macchiati: Sounds like that is best left for a future possible optimization.

icu4j/main/classes/core/src/com/ibm/icu/impl/LocaleFallbackData.java
richgillam: You know, I like that a lot better.  I went ahead and made this change, although since I don't have ImmutableMap, instead of creating a helper class like you show above (which _is_ nice), I just did the builder code the old-fashioned way.

richgillam:
>A thought about unit testing: You need not rely only on data generated from CLDR. If you want to test unusual scenarios, or make sure that the data is stable for testing, you can add source bundles with a small amount of custom-designed data in [icu4c/source/test/testdata](https://github.com/unicode-org/icu/tree/main/icu4c/source/test/testdata).

I have mixed feelings about this.  On the one hand, you're right, and I should definitely do something like this to make sure everything's working right (especially since we don't have en_VA in the parent-locales list right now).  On the other hand, messing with testdata.dat is such a pain in the butt.

I think I'm going to take the middle ground-- I'm going to suspend work on this PR for a day or two and try to crank out my other ICU obligations, and then I'll come back and try to beef up the tests.

macchiati (dismissed):

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
markusicu: Thanks for the changes you made.
And yes, I didn't appreciate that these are display name tests, and display names come from regular, display-oriented resource bundles, so the behavior changes make sense.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/CurrencyTest.java
markusicu: Thanks for the explanation. I had thought that the goal was to put the new parent-locale data to work in both bundle lookup and bundle-parent-chain building. I now understand that it's only the former. So, good.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/generator/ResourceFallbackCodeGenerator.java
markusicu: How about IllegalArgumentException in both places? I am not particular about which exception to throw here.

icu4c/source/common/uresbund.cpp
```diff
@@ -91,6 +91,202 @@ static UBool chopLocale(char *name) {
...
+static CharString getDefaultScript(const CharString& language, const CharString& region) {
+    const char* defaultScript = nullptr;
+    UErrorCode err = U_ZERO_ERROR;
+    
+    // if we were passed both language and region, make them into a locale ID and look that up in the default
+    // script table
+    if (!region.isEmpty()) {
+        CharString localeID;
+        localeID.append(language, err).append("_", err).append(region, err);
+        if (U_SUCCESS(err)) {
```
markusicu, line 139: Error handling: Normally we would have this function take & check & update a `UErrorCode &` argument. If you don't want to do that, then we should at least try to return a non-broken string object, and not clobber an earlier failure like the later `err = U_ZERO_ERROR` currently does.

I suggest:
- pull the `result` declaration to near the top
- this line: `if (U_FAILURE(err)) { return result; }`
- don't reset `err` later (and you won't have to after bailing out on failure)

I think that should do. I believe that appending to result won't fail because the CharString has a stack buffer of 40 chars, so the script code will definitely fit.

icu4c/source/common/uresbund.cpp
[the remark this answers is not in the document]
markusicu: Thank you for figuring out all of the CharString functions and handling -- much nicer now! :-)

markusicu (commented):
a couple more things -- otherwise lgtm, and thanks for the nice improvements!

markusicu:
... and, there is a build bot failure, with a build bot with "LANG has extension tags" in the name. Looks related.
Hopefully just something about testing with a funky default locale ID, such as the default having an extension and the getLocale() omitting that. If so, then just fetch the expected default locale once and remove extensions.
Please take a look.

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
richgillam: Still not sure how to fix these tests.  The easy fix is just to change the expected results to whatever I'm actually getting (presumably English), but that hard-codes an expectation that the system default locale is English, which seems brittle.  I'm leaving those tests commented out for now, but I can make that change if you think that's the right one to make.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/generator/ResourceFallbackCodeGenerator.java
richgillam: Okay.

icu4c/source/common/uresbund.cpp
richgillam: I agree.  I made the change you requested (or at least I think I did).

richgillam:
>Oh, and the merge conflicts in uresbund.cpp are probably from Fredrik replacing all FALSE with false and all TRUE with true. If you do the same in your workspace file, hopefully the conflict goes away.

Yup.  Fixed.

>... and, there is a build bot failure, with a build bot with "LANG has extension tags" in the name. Looks related.
Hopefully just something about testing with a funky default locale ID, such as the default having an extension and the getLocale() omitting that. If so, then just fetch the expected default locale once and remove extensions.
Please take a look.

This one was weird.  It was failing because the locale ID it got back from uloc_getDefault() was actually malformed: It was getting back `en_US_CALENDAR=GREGORIAN;HOURS=H12`.  Neither `uloc_canonicalize()` nor `uloc_getBaseName()` could handle this (the problem is the `_` instead of `@` between `US` and `CALENDAR`).  I ended up changing my code to check if the resource locale is a prefix of the system default locale, which I think will take care of this, but I'm concerned that returning `en_US_CALENDAR=GREGORIAN;HOURS=H12` is flat wrong and might be causing problems elsewhere.  But that's not really my problem...

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
markusicu: Ok. We do test ICU with default locales other than English, so we can't have unguarded tests that only work when it's English.

markusicu:
> > ... and, there is a build bot failure, with a build bot with "LANG has extension tags" in the name. Looks related.
> 
> This one was weird. It was failing because the locale ID it got back from uloc_getDefault() was actually malformed: It was getting back `en_US_CALENDAR=GREGORIAN;HOURS=H12`. Neither `uloc_canonicalize()` nor `uloc_getBaseName()` could handle this (the problem is the `_` instead of `@` between `US` and `CALENDAR`). I ended up changing my code to check if the resource locale is a prefix of the system default locale, which I think will take care of this, but I'm concerned that returning `en_US_CALENDAR=GREGORIAN;HOURS=H12` is flat wrong and might be causing problems elsewhere. But that's not really my problem...

Yikes!
Please submit a bug ticket for uloc_getDefault() returning an ill-formed locale ID. Or maybe it's the test (CI workflow?) passing in a bad environment variable value?

markusicu (commented):
lgtm tnx please squash

icu4j/main/tests/collate/src/com/ibm/icu/dev/test/util/ULocaleCollationTest.java
richgillam: I filed https://unicode-org.atlassian.net/browse/ICU-22152 for somebody to do something smarter with this unit test, so the problem doesn't just fall on the floor.

richgillam:
>Please submit a bug ticket for uloc_getDefault() returning an ill-formed locale ID. Or maybe it's the test (CI workflow?) passing in a bad environment variable value?

I filed https://unicode-org.atlassian.net/browse/ICU-22153 for this problem.  It might be the CI stuff setting the locale wrong; I mentioned that possibility in the ticket description.

richgillam:
>lgtm tnx please squash

Squashed and ready for your final approval (as I write this, all but 6 tests have completed and passed).

markusicu (approved):
lgtm tnx

FrankYFTang:
@richgillam I think this PR cause a UBSAN breakage in v8. See 
https://bugs.chromium.org/p/chromium/issues/detail?id=1394654

FrankYFTang:
Bug file about UBSAN issue in https://unicode-org.atlassian.net/browse/ICU-22224

PR #2149, opened 2022-08-05
ICU-22105 Fixed the unit-conversion logic to work correctly with nega…

richgillam:
…tive temperature values.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22105


sffc (commented):
I have to say that I don't really understand how this change fixes the issue. I'd prefer if @hugovdm or @younies could take a look as well.

hugovdm:
I feel this solution is good. When not doing mixed units, the flooring logic, for extracting integer numbers of larger units, isn't at play.

- I see UnitConversionHandler (number_usageprefs.h) is always using a ComplexUnitsConverter.
- I remain wondering whether it could use a simple UnitsConverter when not dealing with Mixed units (e.g. paired with LongNameHandler when a MixedUnitLongNameHandler is not needed) - but there are also trade-offs regarding code complexity and comprehensibility to keep in mind.

PR #2150, opened 2022-08-08
ICU-22109 icu-config hardcodes build paths unnecessarily

rpurdie:
https://unicode-org.atlassian.net/browse/ICU-22109

The makefile hardcodes paths to the build directory into icu-config. It doesn’t
need to do this and it unnecessarily breaks build reproducibility. This patch
makes a simple change to avoid this.

Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>

### DIFF
--- a/icu4c/source/Makefile.in
+++ b/icu4c/source/Makefile.in
@@ -337,7 +337,7 @@ $(top_builddir)/config/icu-config: $(top_builddir)/Makefile $(top_srcdir)/config
 	chmod u+w $@
 	@echo "# Following from icu/icu4c/source/config/Makefile.inc" >> $@
 	LC_ALL=C $(SED) -f $(top_srcdir)/config/make2sh.sed < $(top_builddir)/config/Makefile.inc | grep -v '#M#' | uniq >> $@
-	@echo "# Following from @platform_make_fragment@" >> $@
+	@echo "# Following from @platform_make_fragment_name@" >> $@
 	LC_ALL=C $(SED) -f $(top_srcdir)/config/make2sh.sed < @platform_make_fragment@ | grep -v '#M#' | uniq >> $@
 	cat $(top_srcdir)/config/icu-config-bottom >> $@
 	chmod u-w $@


markusicu:
@srl295 could you please take a look at this?

srl295 (approved):
@rpurdie  this change is fair enough. 

But in general, the [deprecated icu-config](https://unicode-org.github.io/icu/userguide/icu/howtouseicu.html#notes-on-icu-config) does tend to hardcode paths and tools.  It's recommended to use the pkg-config files instead.

rpurdie:
> @rpurdie this change is fair enough.
> 
> But in general, the [deprecated icu-config](https://unicode-org.github.io/icu/userguide/icu/howtouseicu.html#notes-on-icu-config) does tend to hardcode paths and tools. It's recommended to use the pkg-config files instead.

Thanks, I appreciate it is deprecated and we (as in Yocto Project) very much prefer pkg-config too but until you drop it, we also need to ship it!

markusicu:
Hi @rpurdie our commit checker complains about the commit message. It looks like you have a line break after the ICU ticket number. Could you please try to amend your commit so that there is just a simple space between the ticket number and the first line of text?

srl295:
> > @rpurdie this change is fair enough.
> > But in general, the [deprecated icu-config](https://unicode-org.github.io/icu/userguide/icu/howtouseicu.html#notes-on-icu-config) does tend to hardcode paths and tools. It's recommended to use the pkg-config files instead.
> 
> Thanks, I appreciate it is deprecated and we (as in Yocto Project) very much prefer pkg-config too but until you drop it, we also need to ship it!

why does yocto need to ship it?

rpurdie:
We've had people building software on target which has wanted it from icu in the past. Probably not for a while but we tend to follow the lead of the upstream with provision of things like this as a path of least resistance. If your plan is to remove it, we could delete it and see what issues are raised but it will take a while for users to switch to new versions and notice
.

srl295:
> We've had people building software on target which has wanted it from icu in the past. Probably not for a while but we tend to follow the lead of the upstream with provision of things like this as a path of least resistance. If your plan is to remove it, we could delete it and see what issues are raised but it will take a while for users to switch to new versions and notice
> 
> .

I think what I wanted to do with it was to make it only built if you explicitly requested

rpurdie:
> > We've had people building software on target which has wanted it from icu in the past. Probably not for a while but we tend to follow the lead of the upstream with provision of things like this as a path of least resistance. If your plan is to remove it, we could delete it and see what issues are raised but it will take a while for users to switch to new versions and notice
> > .
> 
> I think what I wanted to do with it was to make it only built if you explicitly requested

Looking at the configure code, it defaults to being installed so we've just run with that. I dislike config scripts like this enough I'm happy to try disabling it!

markusicu:
> Hi @rpurdie our commit checker complains about the commit message. It looks like you have a line break after the ICU ticket number. Could you please try to amend your commit so that there is just a simple space between the ticket number and the first line of text?

Thanks for fixing that!

srl295 (approved):

rpurdie:
> > > > We've had people building software on target which has wanted it from icu in the past. Probably not for a while but we tend to follow the lead of the upstream with provision of things like this as a path of least resistance. If your plan is to remove it, we could delete it and see what issues are raised but it will take a while for users to switch to new versions and notice
> > > > .
> > > 
> > > 
> > > I think what I wanted to do with it was to make it only built if you explicitly requested
> > 
> > 
> > Looking at the configure code, it defaults to being installed so we've just run with that. I dislike config scripts like this enough I'm happy to try disabling it!
> 
> If pkg-config was as widespread then (2002?) as now, I would never have written icu-config. At the time we didn't want to have to require it as a dependency.

Indeed! I totally understand how we end up here. I'm really happy there is a config option. We're now turning off icu-config in Yocto Project/OpenEmbedded and I've not had any complaints so far. If there are we'll tell them to use pkg-config :).

srl295:
Great ! If you want to help a PR to make it off by default would be great.

PR #2154, opened 2022-08-12
ICU-22104 Add tests for CPP alphabetic index set/get maximum label count

sven-oly:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22104


icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+    AlphabeticIndex aindex(locale, status);
+
+    int32_t defaultMax = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 99", 99, defaultMax);
+
+    aindex.setMaxLabelCount(237, status);
+    if (U_FAILURE(errorCode)) {
+      dataerrln("%s %d  Error in index setMaxLabelCount",  __FILE__, __LINE__);
+      return;
+    }
```
markusicu, line 774: Why do you use assertEquals() above but not here?

If you do want it like this, then use errln() here not dataerrln().

dataerrln() is appropriate when a failure at that point might have to do with data loading.
Please check the `status` right after the constructor, and use the `UBool errDataIfFailureAndReset(const char *fmt, ...);` method on the errorCode.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
             uscript_getScript(bucket->getLabel().char32At(0), errorCode));
 }
 
+void AlphabeticIndexTest::checkMaxLabelCount() {
+    IcuTestErrorCode errorCode(*this, "checkMaxLabelCount");
+    UErrorCode status = U_ZERO_ERROR;
```
markusicu, line 762: Why do you have both `errorCode` and `status`? Just use the higher-function `errorCode`.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+    int32_t defaultMax = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 99", 99, defaultMax);
+
+    aindex.setMaxLabelCount(237, status);
+    if (U_FAILURE(errorCode)) {
+      dataerrln("%s %d  Error in index setMaxLabelCount",  __FILE__, __LINE__);
+      return;
+    }
+
+    // Verify that all bucket are cleared after setting max bucket count.
```
markusicu, line 776: All buckets? The code checks only the first one.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+    const UnicodeString bucketLabel = aindex.getBucketLabel();
+    assertEquals("Expect empty string", "", bucketLabel);
+
+    int32_t maxChanged = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 237", 237, maxChanged);
+
+    int32_t bigMax = 100000;
+    aindex.setMaxLabelCount(bigMax, errorCode);
```
markusicu, line 784: I don't know that it's useful to set it more than once without doing anything interesting in between.

I suggest you
- set it once to a value a bit higher than 1000
- add almost that many labels
- buildImmutableIndex() and check that you got all of the expected buckets
- add a bunch more labels, so that you exceed the max
- buildImmutableIndex() and check that the number of buckets does not exceed the max, and that the buckets you get are a  subset of those you set

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+    Locale locale("en");
+    UnicodeString loc = locale.getName();
+    AlphabeticIndex aindex(locale, status);
+
+    int32_t defaultMax = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 99", 99, defaultMax);
+
+    aindex.setMaxLabelCount(237, status);
+    if (U_FAILURE(errorCode)) {
+      dataerrln("%s %d  Error in index setMaxLabelCount",  __FILE__, __LINE__);
```
markusicu, line 772: We always indent by multiples of 4.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+      return;
+    }
+
+    // Verify that all bucket are cleared after setting max bucket count.
+    const UnicodeString bucketLabel = aindex.getBucketLabel();
```
markusicu, line 777: The function returns a reference.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+      dataerrln("%s %d  Error in index setMaxLabelCount to bigMax",
+                __FILE__, __LINE__);
+      return;
+    }
+    maxChanged = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 100000", bigMax, maxChanged);
+
+    int32_t numBuckets = aindex.getBucketCount(status);
+    int32_t expectedBuckets = 0x1C;
+    assertEquals("Expect numBuckets = 1", expectedBuckets, numBuckets);
```
markusicu, line 795: You are not actually expecting it to be 1. I suggest you don't hardcode the expected value in the message, since they are easily out of sync.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
 
+void AlphabeticIndexTest::checkMaxLabelCount() {
+    IcuTestErrorCode errorCode(*this, "checkMaxLabelCount");
+    UErrorCode status = U_ZERO_ERROR;
+    Locale locale("en");
+    UnicodeString loc = locale.getName();
+    AlphabeticIndex aindex(locale, status);
+
+    int32_t defaultMax = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 99", 99, defaultMax);
```
markusicu, line 768: I would not test that it's a specific number (which is not guaranteed to not change), but that it's, say, fewer than 200.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
...
+    if (U_FAILURE(errorCode)) {
+      dataerrln("%s %d  Error in index setMaxLabelCount to bigMax",
+                __FILE__, __LINE__);
+      return;
+    }
+    maxChanged = aindex.getMaxLabelCount();
+    assertEquals("Expect max count = 100000", bigMax, maxChanged);
+
+    int32_t numBuckets = aindex.getBucketCount(status);
+    int32_t expectedBuckets = 0x1C;
```
markusicu, line 794: This magic number deserves a comment, and probably wants to be in decimal.

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -756,4 +757,52 @@ void AlphabeticIndexTest::checkHasBuckets(const Locale &locale, UScriptCode scri
             uscript_getScript(bucket->getLabel().char32At(0), errorCode));
 }
 
+void AlphabeticIndexTest::checkMaxLabelCount() {
+    IcuTestErrorCode errorCode(*this, "checkMaxLabelCount");
+    UErrorCode status = U_ZERO_ERROR;
+    Locale locale("en");
+    UnicodeString loc = locale.getName();
```
markusicu, line 764: unused?

icu4c/source/test/intltest/alphaindextst.cpp
aheninger: I'm fond of using just the test framework macro WHERE for expect messages. It gives the file name and line number, which makes it much easier to zero in on a failing test, compared with searching for the text of the error message in  the test source code. Plus it's quick and easy to write; it's all you really need.

PR #2159, opened 2022-08-23
ICU-22112 word break updates for @,colon; colon tailorings for fi,sv

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22112

Word break updates:
- Per https://unicode-org.atlassian.net/browse/CLDR-15910, move colons out of MidLetter class for standard word break, but add tailorings for `fi` and `sv` that keep colons in MidLetter class
- Per https://unicode-org.atlassian.net/browse/CLDR-15767, add @ to ALetter class


richgillam (dismissed):

macchiati (dismissed):

pedberg-icu:
Hmm, failures on MSVC and Cygwin, likely the non-ASCII in the MidLetter set, will change to \u notation.

pedberg-icu:
> Hmm, failures on MSVC and Cygwin for C, likely the non-ASCII in the MidLetter set, will change to \u notation (not necessary for J but changing there too for consistency)

And that did fix the MSVC etc. failures.

pedberg-icu:
@macchiati @richgillam Thanks for prior approvals, I had to fix CI build failures on MSVC due to the non-ASCII colons in the MidLetter set removals, changed to using \uNNNN notation. Please re-approve when you have a chance...

richgillam (approved):

srl295 (approved):

pedberg-icu:
@macchiati @richgillam Thanks for prior approvals, I had to fix CI build failures on MSVC due to the non-ASCII colons in the MidLetter set removals, changed to using \uNNNN notation, then I had to resolve a merge conflict (just rebuilding the jars). Please re-approve when you have a chance... (now ready)

FrankYFTang:
we now got complain from chrome uses on this https://bugs.chromium.org/p/chromium/issues/detail?id=1410331

pedberg-icu:
> we now got complain from chrome users on this https://bugs.chromium.org/p/chromium/issues/detail?id=1410331

@FrankYFTang That is interesting, since the change was initially promoted by Google, see https://unicode-org.atlassian.net/browse/CLDR-15767

FrankYFTang:
> initially promoted by Google

hum? who from Google lobby for that?

pedberg-icu:
> hum? who from Google lobby for that?

@FrankYFTang Mark and Markus as I recall. This was not something that came from me, or from Apple.

markmentovai:
https://github.com/unicode-org/icu/pull/2159#issuecomment-1421560453:
> we now got complain from chrome users on this https://bugs.chromium.org/p/chromium/issues/detail?id=1410331

This is covered by [Chromium bug 1410331](https://crbug.com/1410331).

The problem in Chromium was that we already had a local patch to break at `'.'`, so in Chromium, ICU 72’s new behavior of not breaking at `'@'` didn’t have the intended effect of not breaking e-mail addresses. Instead, it made e-mail addresses break incredibly awkwardly.

Previously in Chromium, `"user.name@example.com"` would break into `{"user", ".", "name", "@", "example", ".", "com"}`. Not breaking at `'@'` meant that this address began breaking into `{"user", ".", "name@example", ".", "com"}`. This was unexpected and undesirable.

Given that the problematic breaking in Chromium was triggered by the interaction of ICU’s new don’t-break-at-`'@'` behavior and Chromium’s _local_ break-at-`'.'` behavior, I don’t think that there’s any action necessary in ICU, _with one exception_:

word_POSIX.txt appears to have the same behavior as Chromium of breaking at `'.'` ([lines 44, 46](https://github.com/unicode-org/icu/blob/49d192fefe09fcc38547203487cf4e63d2dad61f/icu4c/source/data/brkitr/rules/word_POSIX.txt#L44)), in addition to the new ICU behavior of not breaking at `'@'` ([line 41](https://github.com/unicode-org/icu/blob/49d192fefe09fcc38547203487cf4e63d2dad61f/icu4c/source/data/brkitr/rules/word_POSIX.txt#L44)), so it’s possible that it’s breaking e-mail addresses in the same undesirable way that we saw in Chromium. More attention may be needed to word_POSIX.txt’s handling, even in upstream ICU.

<sub>Chromium’s break-at-`'.'` behavior was a change made to restore previous ICU behavior when upgrading from 3.8 to 4.2.1 (https://codereview.chromium.org/202011, https://crbug.com/21142, https://src.chromium.org/viewvc/chrome?revision=25651&view=revision, 2009-09-08). It’s unclear whether Chromium still needs to break at `'.'`, and if we’re able to stop doing so, we can likely also start breaking at `'@'`.</sub>

PR #2166, opened 2022-08-30
ICU-22116 Updating minimum Java runtime support to Java 8

yumaoka:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22116


icu4j/build.xml
```diff
@@ -813,17 +801,13 @@
         <condition property="disp.is.java8" value="${java.version}" else="${java.version} - Must be 1.8!">
             <isset property="is.java8"/>
         </condition>
-        <condition property="disp.java7.bootclasspath" value="${java7.bootclasspath}" else="Not Defined!">
```
yumaoka, line 816: If we decided to use JDK newer than 8 for release build, we need to cross-compile. But bootclasspath option is deprecated in a new version and need a different way to handle it. So this won't be used in future.

icu4j/build.xml
```diff
@@ -1308,7 +1292,11 @@
         <ant dir="${icu4j.samples.dir}" inheritAll="false"/>
     </target>
 
-    <target name="build-tools" description="Build build-tool classes">
+    <target name="_checkBuildToolsSupported" unless="is.java8">
+        <fail>JDK8 is required to build build-tools.</fail>
+    </target>
+
+    <target name="build-tools" depends="_checkBuildToolsSupported" description="Build build-tool classes">
```
yumaoka, line 1299: We depends on unstable javadoc taglet APIs. Our current source works with JDK 8, but not compatible with newer JDK version. I added a check and a build target requiring javadoc tooling will be failed when JDK 9 or newer version is used.

icu4j/build.xml
```diff
@@ -1730,52 +1706,6 @@
         </java>
     </target>
 
-    <target name="gatherapiOld" depends="info, build-tools" description="Run API database generator tool (Pre Java 5 style)">
```
yumaoka, line 1733: This is for old API signature file format. I think we stopped to producing API signature file in old format since more than 10 versions ago. I think we don't need to maintain this target at this point (this means, you cannot compare APIs between ICU 4.4 and ICU 72.1 - but who would want this at this point?)

icu4j/demos/.classpath
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
 	<classpathentry kind="src" path="src"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
```
yumaoka, line 4: Same changes in all eclipse project files.

icu4j/tools/build/build.xml
```diff
@@ -35,25 +24,11 @@
...
     <target name="clean" depends="@clean" description="Clean up the build outputs"/>
 
-    <target name="compile" depends="copy-src-pre8, @compile" description="Compile java source files"/>
+    <target name="compile" depends="@compile" description="Compile java source files"/>
 
     <target name="copy" depends="@copy" description="Copy non-java runtime files to the project's binary directory"/>
 
     <target name="jar" depends="compile, copy, @jar" description="Create the project's jar file"/>
 
-    <target name="src-jar" depends="copy-src-pre8, @src-jar" description="Create the project's source jar file"/>
```
yumaoka, line 44: We used to support building docs with both JDK 7 and JDK 8. They need different version of source files. I completely dropped JDK 7 support, so no alternative source is required.

However, if we want to support newer version of JDK other than 8, we need to do something similar, because taglet API is not compatible.

yumaoka (commented):
I added some comments for explaining key changes.

markusicu:
@mihnita FYI

icu4j/main/classes/currdata/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -6,9 +6,9 @@ org.eclipse.jdt.core.compiler.annotation.nonnullbydefault=org.eclipse.jdt.annota
 org.eclipse.jdt.core.compiler.annotation.nullable=org.eclipse.jdt.annotation.Nullable
 org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled
 org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
-org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.7
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
```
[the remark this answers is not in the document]
yumaoka: No - I walked through all project files and changed it. Probably copying one to others may work, but I did not check if there are any project specific configuration.

Eclipse project file in a repository is not a bad idea if people are using the same env. But for longer term, we'd manage things through maven plugin and some other mechanisms.

icu4j/build.xml
yumaoka: I think I want to create another PR for further clean up. There are some targets not working at this point. Also some other target seems probably no use/not really maintained. I will add some more comments at that time.

PR #2170, opened 2022-08-31
ICU-22124 Adding a tech preview implementation of MessageFormat v2

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22124

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedMessage.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedMessage.java
@@ -1,0 +1,120 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import java.text.AttributedCharacterIterator;
+
+import com.ibm.icu.text.ConstrainedFieldPosition;
+import com.ibm.icu.text.FormattedValue;
+
+/**
+ * Not yet implemented: The result of a message formatting operation.
+ *
+ * <p>This contains information about where the various fields and placeholders
+ * ended up in the final result.</p>
+ * <p>This class allows the result to be exported in several data types,
+ * including a {@link String}, {@link AttributedCharacterIterator}, more (TBD).</p>
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class FormattedMessage implements FormattedValue {
+
+    /**
+     * Not yet implemented.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedMessage() {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public int length() {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public char charAt(int index) {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public CharSequence subSequence(int start, int end) {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public <A extends Appendable> A appendTo(A appendable) {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public boolean nextPosition(ConstrainedFieldPosition cfpos) {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * Not yet implemented.
+     *
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public AttributedCharacterIterator toCharacterIterator() {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
@@ -1,0 +1,79 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import com.ibm.icu.text.FormattedValue;
+
+/**
+ * An immutable, richer formatting result, encapsulating a {@link FormattedValue},
+ * the original value to format, and we are considering adding some more info.
+ * Very preliminary.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class FormattedPlaceholder {
+    private final FormattedValue formattedValue;
+    private final Object inputValue;
+
+    /**
+     * Constructor creating the {@code FormattedPlaceholder}.
+     *
+     * @param inputValue the original value to be formatted.
+     * @param formattedValue the result of formatting the placeholder.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedPlaceholder(Object inputValue, FormattedValue formattedValue) {
+        if (formattedValue == null) {
+            throw new IllegalAccessError("Should not try to wrap a null formatted value");
+        }
+        this.inputValue = inputValue;
+        this.formattedValue = formattedValue;
+    }
+
+    /**
+     * Retrieve the original input value that was formatted.
+     *
+     * @return the original value to be formatted.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public Object getInput() {
+        return inputValue;
+    }
+
+    /**
+     * Retrieve the formatted value.
+     *
+     * @return the result of formatting the placeholder.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedValue getFormattedValue() {
+        return formattedValue;
+    }
+
+    /**
+     * Returns a string representation of the object.
+     * It can be null, which is unusual, and we plan to change that.
+     *
+     * @return a string representation of the object.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public String toString() {
+        return formattedValue.toString();
+    }
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/message2/IdentityFormatterFactory.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/message2/IdentityFormatterFactory.java
@@ -1,0 +1,39 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Creates a {@link Formatter} that simply returns the String non-i18n aware representation of an object.
+ */
+class IdentityFormatterFactory implements FormatterFactory {
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Formatter createFormatter(Locale locale, Map<String, Object> fixedOptions) {
+        return new IdentityFormatterImpl();
+    }
+
+    private static class IdentityFormatterImpl implements Formatter {
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public FormattedPlaceholder format(Object toFormat, Map<String, Object> variableOptions) {
+            return new FormattedPlaceholder(toFormat, new PlainStringFormattedValue(Objects.toString(toFormat)));
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public String formatToString(Object toFormat, Map<String, Object> variableOptions) {
+            return format(toFormat, variableOptions).toString();
+        }
+    }
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
@@ -1,0 +1,338 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * {@code MessageFormatter} is the next iteration of {@link com.ibm.icu.text.MessageFormat}.
+ *
+ * <p>This new version builds on what we learned from using {@code MessageFormat} for 20 years
+ * in various environments, either exposed "as is" or as a base for other public APIs.</p>
+ *
+ * <p>It is more modular, easier to backport, and provides extension points to add new
+ * formatters and selectors without having to modify the specification.</p>
+ *
+ * <p>We will be able to add formatters for intervals, relative times, lists, measurement units,
+ * people names, and more, and support custom formatters implemented by developers
+ * outside of ICU itself, for company or even product specific needs.</p>
+ *
+ * <p>MessageFormat 2 will support more complex grammatical features, such as gender, inflections,
+ * and tagging parts of the message for style changes or speech.</p>
+ *
+ * <p>The reasoning for this effort is shared in the
+ * <a target="github" href="https://github.com/unicode-org/message-format-wg/blob/main/docs/why_mf_next.md">“Why
+ * MessageFormat needs a successor”</a> document.</p>
+ *
+ * <p>The “MessageFormat 2” project, which develops the new data model, semantics, and syntax,
+ * is hosted on <a target="github" href="https://github.com/unicode-org/message-format-wg">GitHub</a>.</p>
+ *
+ * <p>The current specification for the syntax and data model can be found
+ * <a target="github" href="https://github.com/unicode-org/message-format-wg/blob/develop/spec/syntax.md">here</a>.</p>
+ *
+ * <p>This tech preview implements enough of the {@code MessageFormat} functions to be useful,
+ * but the final set of functions and the parameters accepted by those functions is not yet finalized.</p>
+ *
+ * <p>These are the functions interpreted right now:</p>
+ *
+ * <table border="1">
+ * <tr>
+ *   <td rowspan="4">{@code datetime}</td>
+ *   <td>Similar to the ICU {@code "date"} and {@code "time"}.</td>
+ * </tr>
+ *
+ *   <tr><td>{@code datestyle} and {@code timestyle}<br>
+ *   Similar to {@code argStyle : short | medium | long | full}.<br>
+ *   Same values are accepted, but we can use both in one placeholder,
+ *   for example <code>{$due :datetime datestyle=full timestyle=long}</code>.
+ *   </td></tr>
+ *
+ *   <tr><td>{@code pattern}<br>
+ *   Similar to {@code argStyle = argStyleText}.<br>
+ *   This is bad i18n practice, and will probably be dropped.<br>
+ *   This is included just to support migration to MessageFormat 2.
+ *   </td></tr>
+ *
+ *   <tr><td>{@code skeleton}<br>
+ *   Same as {@code argStyle = argSkeletonText}.<br>
+ *   These are the date/time skeletons as supported by {@link com.ibm.icu.text.SimpleDateFormat}.
+ *   </td></tr>
+ *
+ * <tr>
+ *   <td rowspan="4">{@code number}</td>
+ *   <td>Similar to the ICU "number".</td>
+ * </tr>
+ *
+ *   <tr><td>{@code skeleton}<br>
+ *   These are the number skeletons as supported by {@link com.ibm.icu.number.NumberFormatter}.</td></tr>
+ *
+ *   <tr><td>{@code minimumFractionDigits}<br>
+ *   Only implemented to be able to pass the unit tests from the ECMA tech preview implementation,
+ *   which prefers options bags to skeletons.<br>
+ *   TBD if the final {@number} function will support skeletons, option backs, or both.</td></tr>
+ *
+ *   <tr><td>{@code offset}<br>
+ *   Used to support plural with an offset.</td></tr>
+ *
+ * <tr><td >{@code identity}</td><td>Returns the direct string value of the argument (calling {@code toString()}).</td></tr>
+ *
+ * <tr>
+ *   <td rowspan="3">{@code plural}</td>
+ *   <td>Similar to the ICU {@code "plural"}.</td>
+ * </tr>
+ *
+ *   <tr><td>{@code skeleton}<br>
+ *   These are the number skeletons as supported by {@link com.ibm.icu.number.NumberFormatter}.<br>
+ *   Can also be indirect, from a local variable of type {@code number} (recommended).</td></tr>
+ *
+ *   <tr><td>{@code offset}<br>
+ *   Used to support plural with an offset.<br>
+ *   Can also be indirect, from a local variable of type {@code number} (recommended).</td></tr>
+ *
+ * <tr>
+ *   <td>{@code selectordinal}</td>
+ *   <td>Similar to the ICU {@code "selectordinal"}.<br>
+ * For now it accepts the same parameters as "plural", although there is no use case for them.<br>
+ * TBD if this will be merged into "plural" (with some {@code kind} option) or not.</td></tr>
+ *
+ * <tr><td>{@code select}</td><td>Literal match, same as the ICU4 {@code "select"}.</td></tr>
+ * </table>
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class MessageFormatter {
+    private final Locale locale;
+    private final String pattern;
+    private final Mf2FunctionRegistry functionRegistry;
+    private final Mf2DataModel dataModel;
+    private final Mf2DataModelFormatter modelFormatter;
+
+    private MessageFormatter(Builder builder) {
+        this.locale = builder.locale;
+        this.functionRegistry = builder.functionRegistry;
+        if ((builder.pattern == null && builder.dataModel == null)
+                || (builder.pattern != null && builder.dataModel != null)) {
+            throw new IllegalArgumentException("You need to set either a pattern, or a dataModel, but not both.");
+        }
+
+        if (builder.dataModel != null) {
+            this.dataModel = builder.dataModel;
+            this.pattern = Mf2Serializer.dataModelToString(this.dataModel);
+        } else {
+            this.pattern = builder.pattern;
+            Mf2Serializer tree = new Mf2Serializer();
+            Mf2Parser parser = new Mf2Parser(pattern, tree);
+            try {
+                parser.parse_Message();
+                dataModel = tree.build();
+            } catch (Mf2Parser.ParseException pe) {
+                throw new IllegalArgumentException(
+                        "Parse error:\n"
+                        + "Message: <<" + pattern + ">>\n"
+                        + "Error:" + parser.getErrorMessage(pe) + "\n");
+            }
+        }
+        modelFormatter = new Mf2DataModelFormatter(dataModel, locale, functionRegistry);
+    }
+
+    /**
+     * Creates a builder.
+     *
+     * @return the Builder.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Get the locale to use for all the formatting and selections in
+     * the current {@code MessageFormatter}.
+     *
+     * @return the locale.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public Locale getLocale() {
+        return locale;
+    }
+
+    /**
+     * Get the pattern (the serialized message in MessageFormat 2 syntax) of
+     * the current {@code MessageFormatter}.
+     *
+     * <p>If the {@code MessageFormatter} was created from an {@link Mf2DataModel}
+     * the this string is generated from that model.</p>
+     *
+     * @return the pattern.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public String getPattern() {
+        return pattern;
+    }
+
+    /**
+     * Give public access to the message data model.
+     *
+     * <p>This data model is similar to the functionality we have today
+     * in {@link com.ibm.icu.text.MessagePatternUtil} maybe even a bit more higher level.</p>
+     *
+     * <p>We can also imagine a model where one parses the string syntax, takes the data model,
+     * modifies it, and then uses that modified model to create a {@code MessageFormatter}.</p>
+     *
+     * @return the data model.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public Mf2DataModel getDataModel() {
+        return dataModel;
+    }
+
+    /**
+     * Formats a map of objects by iterating over the MessageFormat's pattern,
+     * with the plain text “as is” and the arguments replaced by the formatted objects.
+     *
+     * @param arguments a map of objects to be formatted and substituted.
+     * @return the string representing the message with parameters replaced.
+     *
+     * @throws IllegalArgumentException when something goes wrong
+     *         (for example wrong argument type, or null arguments, etc.)
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public String formatToString(Map<String, Object> arguments) {
+        return modelFormatter.format(arguments);
+    }
+
+    /**
+     * Not yet implemented: formats a map of objects by iterating over the MessageFormat's
+     * pattern, with the plain text “as is” and the arguments replaced by the formatted objects.
+     *
+     * @param arguments a map of objects to be formatted and substituted.
+     * @return the {@link FormattedMessage} class representing the message with parameters replaced.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedMessage format(Map<String, Object> arguments) {
+        throw new RuntimeException("Not yet implemented.");
+    }
+
+    /**
+     * A {@code Builder} used to build instances of {@link MessageFormatter}.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public static class Builder {
+        private Locale locale = Locale.getDefault(Locale.Category.FORMAT);
+        private String pattern = null;
+        private Mf2FunctionRegistry functionRegistry = Mf2FunctionRegistry.builder().build();
+        private Mf2DataModel dataModel = null;
+
+        // Prevent direct creation
+        private Builder() {
+        }
+
+        /**
+         * Sets the locale to use for all formatting and selection operations.
+         *
+         * @param locale the locale to set.
+         * @return the builder, for fluent use.
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public Builder setLocale(Locale locale) {
+            this.locale = locale;
+            return this;
+        }
+
+        /**
+         * Sets the pattern (in MessageFormat 2 syntax) used to create the message.<br>
+         * It conflicts with the data model, so it will reset it (the last call on setter wins).
+         *
+         * @param pattern the pattern to set.
+         * @return the builder, for fluent use.
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public Builder setPattern(String pattern) {
+            this.pattern = pattern;
+            this.dataModel = null;
+            return this;
+        }
+
+        /**
+         * Sets an instance of {@link Mf2FunctionRegistry} that should register any
+         * custom functions used by the message.
+         *
+         * <p>There is no need to do this in order to use standard functions
+         * (for example date / time / number formatting, plural / ordinal / literal selection).<br>
+         * The exact set of standard functions, with the types they format and the options
+         * they accept is still TBD.</p>
+         *
+         * @param functionRegistry the function registry to set.
+         * @return the builder, for fluent use.
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public Builder setFunctionRegistry(Mf2FunctionRegistry functionRegistry) {
+            this.functionRegistry = functionRegistry;
+            return this;
+        }
+
+        /**
+         * Sets the data model used to create the message.<br>
+         * It conflicts with the pattern, so it will reset it (the last call on setter wins).
+         *
+         * @param dataModel the pattern to set.
+         * @return the builder, for fluent use.
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public Builder setDataModel(Mf2DataModel dataModel) {
+            this.dataModel = dataModel;
+            this.pattern = null;
+            return this;
+        }
+
+        /**
+         * Builds an instance of {@link MessageFormatter}.
+         *
+         * @return the {@link MessageFormatter} created.
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public MessageFormatter build() {
+            return new MessageFormatter(this);
+        }
+    }
+}
--- a/icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
@@ -1,0 +1,280 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import com.ibm.icu.message2.Mf2DataModel.Expression;
+import com.ibm.icu.message2.Mf2DataModel.Part;
+import com.ibm.icu.message2.Mf2DataModel.Pattern;
+import com.ibm.icu.message2.Mf2DataModel.SelectorKeys;
+import com.ibm.icu.message2.Mf2DataModel.Text;
+import com.ibm.icu.message2.Mf2DataModel.Value;
+import com.ibm.icu.util.Calendar;
+import com.ibm.icu.util.CurrencyAmount;
+
+/**
+ * Takes an {@link Mf2DataModel} and formats it to a {@link String}
+ * (and later on we will also implement formatting to a {@code FormattedMessage}).
+ */
+// TODO: move this in the MessageFormatter
+class Mf2DataModelFormatter {
+    private final Locale locale;
+    private final Mf2DataModel dm;
+
+    final Mf2FunctionRegistry standardFunctions;
+    final Mf2FunctionRegistry customFunctions;
+    private static final Mf2FunctionRegistry EMPTY_REGISTY = Mf2FunctionRegistry.builder().build();
+
+    Mf2DataModelFormatter(Mf2DataModel dm, Locale locale, Mf2FunctionRegistry customFunctionRegistry) {
+        this.locale = locale;
+        this.dm = dm;
+        this.customFunctions = customFunctionRegistry == null ? EMPTY_REGISTY : customFunctionRegistry;
+
+        standardFunctions = Mf2FunctionRegistry.builder()
+                // Date/time formatting
+                .setFormatter("datetime", new DateTimeFormatterFactory())
+                .setDefaultFormatterNameForType(Date.class, "datetime")
+                .setDefaultFormatterNameForType(Calendar.class, "datetime")
+
+                // Number formatting
+                .setFormatter("number", new NumberFormatterFactory())
+                .setDefaultFormatterNameForType(Integer.class, "number")
+                .setDefaultFormatterNameForType(Double.class, "number")
+                .setDefaultFormatterNameForType(Number.class, "number")
+                .setDefaultFormatterNameForType(CurrencyAmount.class, "number")
+
+                // Format that returns "to string"
+                .setFormatter("identity", new IdentityFormatterFactory())
+                .setDefaultFormatterNameForType(String.class, "identity")
+                .setDefaultFormatterNameForType(CharSequence.class, "identity")
+
+                // Register the standard selectors
+                .setSelector("plural", new PluralSelectorFactory("cardinal"))
+                .setSelector("selectordinal", new PluralSelectorFactory("ordinal"))
+                .setSelector("select", new TextSelectorFactory())
+                .setSelector("gender", new TextSelectorFactory())
+
+                .build();
+    }
+
+    private static Map<String, Object> mf2OptToFixedOptions(Map<String, Value> options) {
+        Map<String, Object> result = new HashMap<>();
+        for (Entry<String, Value> option : options.entrySet()) {
+            Value value = option.getValue();
+            if (value.isLiteral()) {
+                result.put(option.getKey(), value.getLiteral());
+            }
+        }
+        return result;
+    }
+
+    private Map<String, Object> mf2OptToVariableOptions(Map<String, Value> options, Map<String, Object> arguments) {
+        Map<String, Object> result = new HashMap<>();
+        for (Entry<String, Value> option : options.entrySet()) {
+            Value value = option.getValue();
+            if (value.isVariable()) {
+                result.put(option.getKey(), variableToObjectEx(value, arguments));
+            }
+        }
+        return result;
+    }
+
+    FormatterFactory getFormattingFunctionFactoryByName(Object toFormat, String functionName) {
+        // Get a function name from the type of the object to format
+        if (functionName == null || functionName.isEmpty()) {
+            if (toFormat == null) {
+                // The object to format is null, and no function provided.
+                return null;
+            }
+            Class<?> clazz = toFormat.getClass();
+            functionName = standardFunctions.getDefaultFormatterNameForType(clazz);
+            if (functionName == null) {
+                functionName = customFunctions.getDefaultFormatterNameForType(clazz);
+            }
+            if (functionName == null) {
+                throw new IllegalArgumentException("Object to format without a function, and unknown type: "
+                        + toFormat.getClass().getName());
+            }
+        }
+
+        FormatterFactory func = standardFunctions.getFormatter(functionName);
+        if (func == null) {
+            func = customFunctions.getFormatter(functionName);
+            if (func == null) {
+                throw new IllegalArgumentException("Can't find an implementation for function: '"
+                        + functionName + "'");
+            }
+        }
+        return func;
+    }
+
+    String format(Map<String, Object> arguments) {
+        List<Expression> selectors = dm.getSelectors();
+        Pattern patternToRender = selectors.isEmpty()
+                ? dm.getPattern()
+                : findBestMatchingPattern(selectors, arguments);
+
+        StringBuilder result = new StringBuilder();
+        for (Part part : patternToRender.getParts()) {
+            if (part instanceof Text) {
+                result.append(part);
+            } else if (part instanceof Expression) { // Placeholder is an Expression
+                FormattedPlaceholder fp = formatPlaceholder((Expression) part, arguments, false);
+                result.append(fp.toString());
+            } else {
+                throw new IllegalArgumentException("Unknown part type: " + part);
+            }
+        }
+        return result.toString();
+    }
+
+    private Pattern findBestMatchingPattern(List<Expression> selectors, Map<String, Object> arguments) {
+        Pattern patternToRender = null;
+
+        // Collect all the selector functions in an array, to reuse
+        List<Selector> selectorFunctions = new ArrayList<>(selectors.size());
+        for (Expression selector : selectors) {
+            String functionName = selector.getFunctionName();
+            SelectorFactory funcFactory = standardFunctions.getSelector(functionName);
+            if (funcFactory == null) {
+                funcFactory = customFunctions.getSelector(functionName);
+            }
+            if (funcFactory != null) {
+                Map<String, Object> opt = mf2OptToFixedOptions(selector.getOptions());
+                selectorFunctions.add(funcFactory.createSelector(locale, opt));
+            } else {
+                throw new IllegalArgumentException("Unknown selector type: " + functionName);
+            }
+        }
+        // This should not be possible, we added one function for each selector, or we have thrown an exception.
+        // But just in case someone removes the throw above?
+        if (selectorFunctions.size() != selectors.size()) {
+            throw new IllegalArgumentException("Something went wrong, not enough selector functions, "
+                    + selectorFunctions.size() + " vs. " + selectors.size());
+        }
+
+        // Iterate "vertically", through all variants
+        for (Entry<SelectorKeys, Pattern> variant : dm.getVariants().entrySet()) {
+            int maxCount = selectors.size();
+            List<String> keysToCheck = variant.getKey().getKeys();
+            if (selectors.size() != keysToCheck.size()) {
+                throw new IllegalArgumentException("Mismatch between the number of selectors and the number of keys: "
+                        + selectors.size() + " vs. " + keysToCheck.size());
+            }
+            boolean matches = true;
+            // Iterate "horizontally", through all matching functions and keys
+            for (int i = 0; i < maxCount; i++) {
+                Expression selector = selectors.get(i);
+                String valToCheck = keysToCheck.get(i);
+                Selector func = selectorFunctions.get(i);
+                Map<String, Object> options = mf2OptToVariableOptions(selector.getOptions(), arguments);
+                if (!func.matches(variableToObjectEx(selector.getOperand(), arguments), valToCheck, options)) {
+                    matches = false;
+                    break;
+                }
+            }
+            if (matches) {
+                patternToRender = variant.getValue();
+                break;
+            }
+        }
+
+        // TODO: check that there was an entry with all the keys set to `*`
+        // And should do that only once, when building the data model.
+        if (patternToRender == null) {
+            // If there was a case with all entries in the keys `*` this should not happen
+            throw new IllegalArgumentException("The selection went wrong, cannot select any option.");
+        }
+
+        return patternToRender;
+    }
+
+    /*
+     * Pass a level to prevent local variables calling each-other recursively:
+     *
+     * <code><pre>
+     * let $l1 = {$l4 :number}
+     * let $l2 = {$l1 :number}
+     * let $l3 = {$l2 :number}
+     * let $l4 = {$l3 :number}
+     * </pre></code>
+     *
+     * We can keep track of the calls (complicated and expensive).
+     * Or we can forbid the use of variables before they are declared, but that is not in the spec (yet?).
+     */
+    private Object variableToObjectEx(Value value, Map<String, Object> arguments) {
+        if (value == null) { // function only
+            return null;
+        }
+        // We have an operand. Can be literal, local var, or argument.
+        if (value.isLiteral()) {
+            return value.getLiteral();
+        } else if (value.isVariable()) {
+            String varName = value.getVariableName();
+            Expression localPh = dm.getLocalVariables().get(varName);
+            if (localPh != null) {
+                return formatPlaceholder(localPh, arguments, false);
+            }
+            return arguments.get(varName);
+        } else {
+            throw new IllegalArgumentException("Invalid operand type " + value);
+        }
+    }
+
+    private FormattedPlaceholder formatPlaceholder(Expression ph, Map<String, Object> arguments, boolean localExpression) {
+        Object toFormat;
+        Value operand = ph.getOperand();
+        if (operand == null) { // function only, "...{:currentOs option=value}..."
+            toFormat = null;
+        } else {
+            // We have an operand. Can be literal, local var, or argument.
+            if (operand.isLiteral()) { // "...{(1234.56) :number}..."
+                // If it is a literal, return the string itself
+                toFormat = operand.getLiteral();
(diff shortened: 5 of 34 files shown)


icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+
+package com.ibm.icu.message2;
+
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * {@code MessageFormatter} is the next iteration of {@link com.ibm.icu.text.MessageFormat}.
+ *
+ * <p>For more than two years the Message Format Working Group (MFWG) worked on
```
macchiati, line 12: The discussion of the background should not talk about the MFWG, but rather just focus on the feature comparison to MF1.0

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+ *
+ *   <tr><td>{@code datestyle} and {@code timestyle}<br>
+ *   Similar to {@code argStyle : short | medium | long | full}.<br>
+ *   Same values are accepted, but we can use both in one placeholder,
+ *   for example <code>{$due :datetime datestyle=full timestyle=long}</code>.
+ *   </td></tr>
+ *
+ *   <tr><td>{@code pattern}<br>
+ *   Similar to {@code argStyle = argStyleText}.<br>
+ *   This is bad i18n practice, and will probably be dropped.<br>
```
macchiati, line 53: Change to: This is included just to support migration to MF2.0.

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
```diff
@@ -0,0 +1,73 @@
...
+package com.ibm.icu.message2;
+
+/**
+ * An exception thrown when parsing a {@link MessageFormatter} syntax fails.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class Mf2ParseException extends RuntimeException {
```
markusicu, line 13: Discuss: Should this be a different exception class from MessageFormatterException or IllegalArgumentException?

markusicu, line 13: Note: The exceptions are public API but not in the design doc / proposal. We will need to amend the proposal to include all public API.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    protected Mf2ParseException(String message, Throwable cause,
+                               boolean enableSuppression,
+                               boolean writableStackTrace) {
```
markusicu, line 70: Same question as for the other exception, about whether we need/want to override this one.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2Parser.java
```diff
@@ -0,0 +1,758 @@
...
+
+/**
+ * Class generated from EBNF.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+@SuppressWarnings("all") // Disable all warnings in the generated file
+class Mf2Parser
```
markusicu, line 16: Please remove the internal/deprecated/tech preview annotations+comments from non-public classes and methods.

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Very-very rough implementation of FormattedValue, packaging a string.
+ * Expect it to change.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class PlainStringFormattedValue implements FormattedValue {
```
markusicu, line 20: Can we keep this internal / only-package-visible?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Args.java
```diff
@@ -0,0 +1,187 @@
...
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.dev.test.message2;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Convenience class that provides the same functionality as
+ * <code>Map.of</code> introduced in JDK 11, which can't be used for ICU4J.
```
markusicu, line 11:
```suggestion
 * Convenience class that provides the same functionality as
 * <code>Map.of</code> introduced in JDK 11, which can't be used yet for ICU4J.
```

markusicu, line 11: Question: Should the returned Map be unmodifiable? Maybe to prove that the formatter does not modify it?

markusicu (commented):
I did a run through the public API.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2.functions;
```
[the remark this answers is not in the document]
mihnita: The reason I did it was for modularity.

And to enforce that the standard functions really built on top of public APIs
(because they can't access package APIs).
So it "proves" that one can replace them (with ones implemented on top of Android APIs, for example)

And they are public because they are in a different package.

---

I moved them.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+    @Deprecated
+    @Override
+    public Formatter createFormatter(Locale locale, Map<String, Object> fixedOptions) {
+        DateFormat df;
+
+        Object opt = fixedOptions.get("skeleton");
+        if (opt != null) {
+            String skeleton = Objects.toString(opt);
+            df = DateFormat.getInstanceForSkeleton(skeleton, locale);
+            return new DateTimeFormatter(df);
```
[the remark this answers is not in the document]
mihnita: It is function dependent.
I would expect that part of the function "registration" in the standard to describe the options, priorities, etc.
Since the functions are not standard, I made up my own rules (skeleton takes priority).

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+        /**
+         * {@inheritDoc}
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        @Override
+        public FormattedPlaceholder format(Object toFormat, Map<String, Object> variableOptions) {
+            String result = toFormat == null ? null : icuFormatter.format(toFormat);
```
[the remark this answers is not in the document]
mihnita: To represent functions without something to format.
For example `"...{:platform}..."`

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public Formatter createFormatter(Locale locale, Map<String, Object> fixedOptions) {
+        DateFormat df;
+
+        Object opt = fixedOptions.get("skeleton");
```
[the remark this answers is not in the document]
mihnita: I think that would be function dependent.

All the current function implementations are "made up", to get something functional,
and to show that it is possible to support all kinds of scenarios.
But there is no "polishing"

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
```diff
@@ -0,0 +1,76 @@
...
+     * Constructor creating the immutable {@code FormattedPlaceholder}.
+     *
+     * @param formattedValue the result of formatting the placeholder.
+     * @param inputValue the original value to be formatted.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedPlaceholder(FormattedValue formattedValue, Object inputValue) {
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
```diff
@@ -0,0 +1,76 @@
...
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class FormattedPlaceholder {
+    private final FormattedValue formattedValue;
+    private final Object inputValue;
+
+    /**
+     * Constructor creating the immutable {@code FormattedPlaceholder}.
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @return a string representation of the object.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public String toString() {
+        return formattedValue == null ? null : formattedValue.toString();
```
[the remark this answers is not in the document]
mihnita: Kind of yes.

I tried to emulate what MessageFormat does.
When something goes wrong (for certain kinds of "wrong" :-) MF1 returns the placeholder un-replaced.

For example `"Hello {user}"`, if there was no parameter named "user", the result is `"Hello {user}"`
So it does not throw.

Since the functions in placeholders don't have access to their own name, or to the name of the argument they format (only the value of the arg to format), there is no way for a function that goes wrong to
"fallback" to "...{$foo}..." or "...{:bar}...".

But the the "engine" knows that. The function returns null in the formatted value, and the engine
decides what to do (throw, show `null` or `{$foo}`).

Plus we get consistency: if all functions return null when they fail, all failing placeholders look the same.
You don't get empty string from one, "null" from another, and so on ("Hello null, tomorrow is {$date}")

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+     *
+     * @param arguments a map of objects to be formatted and substituted.
+     * @return the {@link FormattedMessage} class representing the message with parameters replaced.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public FormattedMessage format(Map<String, Object> arguments) {
+        throw new RuntimeException("Not yet implemented.");
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2Parser.java
mihnita: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Args.java
mihnita: Done.
Updated the javadoc, and changed all returns to
`return Collections.unmodifiableMap(result);`

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
mihnita: I use it in the unit tests when I implement some custom functions.

But I can make a copy of it in tests, and make them both package visible?

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+     *
+     * @param arguments a map of objects to be formatted and substituted.
+     * @return the string representing the message with parameters replaced.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public String formatToString(Map<String, Object> arguments) {
+        Mf2DataModelFormatter mfr = new Mf2DataModelFormatter(dataModel, locale, functionRegistry);
```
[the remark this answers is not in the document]
mihnita: Done.

Created a final field, initialized in constructor (at the same time with the `dataModel`), and reused.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+@Deprecated
+public class DateTimeFormatterFactory implements FormatterFactory {
+
+    private static int stringToStyle(String option) {
+        switch (option.toUpperCase(Locale.US)) {
+            case "FULL": return DateFormat.FULL;
+            case "LONG": return DateFormat.LONG;
+            case "MEDIUM": return DateFormat.MEDIUM;
+            case "SHORT": return DateFormat.SHORT;
+            default: return DateFormat.DEFAULT;
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        opt = fixedOptions.get("pattern");
+        if (opt != null) {
+            String pattern = Objects.toString(opt);
+            // I don't think there is a way anymore to create a date formatter with a pattern,
+            // only with a skeleton.
+            // So we create it with an empty string, and then we apply the pattern we want.
+            SimpleDateFormat sf = (SimpleDateFormat) SimpleDateFormat.getPatternInstance("", locale);
+            sf.applyPattern(pattern);
```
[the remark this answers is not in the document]
mihnita: Done.

---

I didn't see it :-)
I was conditioned (by most ICU formatters) to use a get???Instance.
And `getPatternInstance` takes a skeleton, although there is a `getInstanceForSkeleton` already.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,285 @@
...
+import com.ibm.icu.message2.functions.IdentityFormatterFactory;
+import com.ibm.icu.message2.functions.NumberFormatterFactory;
+import com.ibm.icu.message2.functions.PluralSelectorFactory;
+import com.ibm.icu.message2.functions.TextSelectorFactory;
+import com.ibm.icu.text.FormattedValue;
+import com.ibm.icu.util.Calendar;
+import com.ibm.icu.util.CurrencyAmount;
+
+/**
+ * Takes a {@link Mf2DataModel} and formats it to a {@link String}
```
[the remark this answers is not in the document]
mihnita: Done (2 instances)

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2FormattingException.java
```diff
@@ -0,0 +1,73 @@
...
+package com.ibm.icu.message2;
+
+/**
+ * An exception thrown when formatting a {@link MessageFormatter} fails.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class Mf2FormattingException extends RuntimeException {
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2FunctionRegistry.java
```diff
@@ -0,0 +1,348 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
```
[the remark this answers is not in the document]
mihnita: Done.

I also like it. This was Eclipse "helping".

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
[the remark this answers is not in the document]
mihnita: Done.

mihnita: I checked what MessageFormat does: `java.lang.IllegalArgumentException`
I can do that, and remove `Mf2ParseException`.
Your call, I don't mind.

mihnita: Done.
Replaced all instances with `IllegalArgumentException` and removed `Mf2ParseException`

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2FormattingException.java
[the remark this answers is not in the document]
mihnita: Renamed.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2FormattingException.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    protected Mf2FormattingException(String message, Throwable cause,
+                               boolean enableSuppression,
+                               boolean writableStackTrace) {
```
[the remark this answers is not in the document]
mihnita: I don't know if we have to.
But since the parent had it, I thought it is a good thing to do it.

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+        if (builder.dataModel != null) {
+            this.dataModel = builder.dataModel;
+            this.pattern = Mf2Serializer.dataModelToString(this.dataModel);
+        } else {
+            this.pattern = builder.pattern;
+            Mf2Serializer tree = new Mf2Serializer();
+            Mf2Parser parser = new Mf2Parser(pattern, tree);
+            try {
+                parser.parse_Message();
+                dataModel = tree.build();
```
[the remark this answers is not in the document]
mihnita: The idea was that the data model is syntax independent.

One can write a MF1 serializer parsing the MF1 syntax to the data model, and generate MF1 syntax from the data model.
Similar if one wants a Fluent serializer.

---

That is also the reason why the formatting proper is on the model and outside MF2.
One can (for example) write a MF1 serializer, implement MF1 compatible functions ("date", "time", etc), and re-implement MF1 on top of that. Or implement FluentFormatter on top of that.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
mihnita: I've checked how I use the exceptions.

I used the `Mf2ParseException` to report parsing errors (missing `{` and so on), when one creates a `MessageFormatter` from a string.
But now it is replaced with `IllegalArgumentException`, same as MF1

And I used `Mf2FormatException` (now `MessageFormatterException`) when the `mf.format` is invoked.
Thinks like type problems, or bad values for the function options.

But it looks like MF1 uses `IllegalArgumentException` for that too (for example I pass a String to a "...{foo, date}...")
Which kind of is true.

Idea: should I use `IllegalArgumentException` instead of `MessageFormatterException` and remove it?
That would make it more consistent with MF1.
And makes it easier to "split" into a backport that is not ICU4J dependent :-)

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
mihnita: The Mf2ParseException is removed.

But otherwise the same answer: inheriting from `ICUException` means I don't need to override this,
and adding it to `ICUException` is a decision that is not part of MF2.

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedMessage.java
```diff
@@ -0,0 +1,108 @@
+// © 2022 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.message2;
```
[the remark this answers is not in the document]
mihnita: Done

Added the proper `<include name="com/ibm/icu/message2/**"/>` to the `build.xml` file.

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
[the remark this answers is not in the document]
mihnita: TODO: Serializer => find a better name
TODO: move the formatting work in the MessageFormatter

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
```diff
@@ -0,0 +1,333 @@
...
+     * @param arguments a map of objects to be formatted and substituted.
+     * @return the string representing the message with parameters replaced.
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    public String formatToString(Map<String, Object> arguments) {
+        Mf2DataModelFormatter mfr = new Mf2DataModelFormatter(dataModel, locale, functionRegistry);
+        return mfr.format(arguments);
```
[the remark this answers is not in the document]
mihnita: TODO: move the formatting work in the MessageFormatter

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: TODO: how to handle conflicts

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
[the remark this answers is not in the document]
mihnita: TODO: FormattedValue should never be null.
In case of error the functions should throw.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: TODO: singleton / const => a special type ?(with if ... instanceof) 
Not a 72 blocker.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: Not a blocker

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: Sometimes handled in the factory, sometimes in the formatter.

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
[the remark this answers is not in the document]
mihnita: Try doing it.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
mihnita: OK, use IllegalArgumentException and remove this class.

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
mihnita: Internal, deprecated, but not tech preview.
visible for testing, maybe.
Search examples.

icu4j/main/classes/core/src/com/ibm/icu/message2/MessageFormatter.java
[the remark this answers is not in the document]
mihnita: Done.
Added TODO in the Mf2DataModelFormatter file

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: Done
Added TODO in `DateTimeFormatterFactory`.
Although the decision is not date-time specific, it is general, for all formatters.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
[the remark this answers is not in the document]
mihnita: Done.
Added TODO.
Doing it now would change already approved public API.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2ParseException.java
mihnita: Done.

Throwing IllegalArgumentException, added that info to javadoc, removed this class.

icu4j/main/classes/core/src/com/ibm/icu/message2/functions/DateTimeFormatterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+ * <code>{exp, date}</code> and <code>{exp, time}</code> in {@link com.ibm.icu.text.MessageFormat}.
+ *
+ * @internal ICU 72 technology preview
+ * @deprecated This API is for ICU internal use only.
+ */
+@Deprecated
+public class DateTimeFormatterFactory implements FormatterFactory {
+
+    private static int stringToStyle(String option) {
+        switch (option.toUpperCase(Locale.US)) {
```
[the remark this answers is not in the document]
mihnita: Done
Used `AsciiUtil.toUpperString`

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
mihnita: Used `@internal ICU 72 technology preview. Visible For Testing.` as a compromise.

`PluralRules` use `@internal Visible For Testing`, and it is the only class doing that, so I used it as a model.

Not sure it is 100% ok.
But it looks like what's in @internal is not that strictly defined.
I though it is good to keep the "tech preview" part (for example to grep).
There are all kind of "unstructured" things, for example:
```
@internal Used in UnicodeTools
@internal but could become public
@internal, only public for access by callback.
@internal For specific users only until proposed publicly.
```

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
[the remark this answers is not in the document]
mihnita: Done.
A lot less painful than I was afraid of.
Only one test failed, the JS one with `"hello {null}"`

Throwing a NPE as this is what MF1 does, but let me know if you would rather see `IllegalArgumentException`.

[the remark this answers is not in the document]
markusicu: Now that we no longer allow a null formattedValue, please stop checking for it.
```suggestion
        return formattedValue.toString();
```

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
```diff
@@ -0,0 +1,76 @@
...
+@Deprecated
+public class FormattedPlaceholder {
+    private final FormattedValue formattedValue;
+    private final Object inputValue;
+
+    /**
+     * Constructor creating the {@code FormattedPlaceholder}.
+     *
+     * @param formattedValue the result of formatting the placeholder.
+     * @param inputValue the original value to be formatted.
```
markusicu, line 25: swap docs like params
```suggestion
     * @param inputValue the original value to be formatted.
     * @param formattedValue the result of formatting the placeholder.
```

icu4j/main/classes/core/src/com/ibm/icu/message2/IdentityFormatterFactory.java
```diff
@@ -0,0 +1,51 @@
...
+class IdentityFormatterFactory implements FormatterFactory {
+    /**
+     * {@inheritDoc}
+     *
+     * @internal ICU 72 technology preview
+     * @deprecated This API is for ICU internal use only.
+     */
+    @Deprecated
+    @Override
+    public Formatter createFormatter(Locale locale, Map<String, Object> fixedOptions) {
```
markusicu, line 22: Could you please remove the internal/deprecated annotations also from methods that are no longer public because their classes no longer are?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                .setDefaultFormatterNameForType(Calendar.class, "datetime")
+
+                // Number formatting
+                .setFormatter("number", new NumberFormatterFactory())
+                .setDefaultFormatterNameForType(Integer.class, "number")
+                .setDefaultFormatterNameForType(Double.class, "number")
+                .setDefaultFormatterNameForType(Number.class, "number")
+                .setDefaultFormatterNameForType(CurrencyAmount.class, "number")
+
+                // Format that returns "to sting"
```
markusicu, line 54:
```suggestion
                // Format that returns "to string"
```

icu4j/main/classes/core/src/com/ibm/icu/message2/OptUtils.java
```diff
@@ -0,0 +1,50 @@
...
+            } catch (NumberFormatException e) {
+            }
+        }
+        return null;
+    }
+
+    static Integer getInteger(Map<String, Object> options, String key) {
+        Object value = options.get(key);
+        if (value == null)
+            return null;
```
markusicu, line 32: Please always use curly braces for if, else, etc.

icu4j/main/classes/core/src/com/ibm/icu/message2/PluralSelectorFactory.java
```diff
@@ -0,0 +1,117 @@
...
+ */
+class PluralSelectorFactory implements SelectorFactory {
+    private final PluralType pluralType;
+
+    /**
+     * Creates a {@code PluralSelectorFactory} of the desired type.
+     *
+     * @param type the kind of plural selection we want
+     */
+    PluralSelectorFactory(String type) {
```
markusicu, line 26: Please add a TODO to use an enum -- probably PluralType -- not a String type.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Args.java
```diff
@@ -0,0 +1,193 @@
...
+ * does not change it</p>
+ */
+@SuppressWarnings("javadoc")
+public class Args {
+
+    public static final Map<String, Object> NONE = new HashMap<>();
+
+    public static Map<String, Object> of(
+            String argName0, Object argValue0
+    ) {
```
markusicu, line 26: nit/optional: I would pull these `) {` to the end of the previous line (multiple instances)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
```diff
@@ -0,0 +1,412 @@
...
+                .pattern("{hello {(world)}}")
+                .expected("hello world")
+                .build(),
+            new TestCase.Builder()
+                .pattern("{hello {()}}")
+                .expected("hello ")
+                .build(),
+            new TestCase.Builder()
+                .pattern("{hello {$place}}")
+                .arguments(Args.of( "place", "world" ))
```
markusicu, line 33: nit/optional: no spaces immediately inside the argument list; e.g.:
```suggestion
                .arguments(Args.of("place", "world"))
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
```diff
@@ -0,0 +1,412 @@
...
+                .build(),
+            new TestCase.Builder()
+                .pattern("{{$one} et {$two}}")
+                .locale("fr")
+                .arguments(Args.of( "one", 1.3, "two", 4.2 ))
+                .expected("1,3 et 4,2")
+                .build(),
+            new TestCase.Builder()
+                .pattern("{hello {(4.2) :number}}")
+                .expected("hello 4.2")
```
markusicu, line 54: 🎉 !
... and looking forward to seeing this with locale "ar-EG" :-)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
```diff
@@ -0,0 +1,412 @@
...
+                .arguments(Args.of( "bar", 4.2))
+                .expected("bar 4.20")
+                .build(),
+            new TestCase.Builder().ignore("Maybe") // Because minimumFractionDigits=foo
+                .pattern("let $foo = {$bar :number minimumFractionDigits=foo} {bar {$foo}}")
+                .arguments(Args.of( "bar", 4.2))
+                .expected("bar 4.2")
+                .errors("invalid-type")
+                .build(),
+            new TestCase.Builder().ignore("Maybe. Function speciffic behavior.")
```
markusicu, line 118:
```suggestion
            new TestCase.Builder().ignore("Maybe. Function specific behavior.")
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
```diff
@@ -0,0 +1,412 @@
...
+                .expected("bar NaN")
+                .build(),
+            new TestCase.Builder()
+                .pattern("let $foo = {$bar} let $bar = {$baz} {bar {$foo}}")
+                .arguments(Args.of( "baz", "foo"))
+                // expectedJs: "bar foo"
+                // It is currently undefined if a local variable (like $foo)
+                // can reference a local variable that was not yet defined (like $bar).
+                // That is called hoisting and it is valid in JavaScript or Python.
+                // Not allowing that would prevent circular references.
```
markusicu, line 130: opinion: let's not allow that

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
```diff
@@ -0,0 +1,412 @@
...
+                .build(),
+            new TestCase.Builder()
+                .patternJs("match {$foo} when (1) {one} when * {other}")
+                .pattern("match {$foo :select} when (1) {one} when * {other}")
+                .arguments(Args.of( "foo", "1"))
+                .expected("one")
+                .build(),
+            new TestCase.Builder()
+                .pattern("match {$foo :plural} when 1 {one} when * {other}")
+                .arguments(Args.of( "foo", "1")) // Should this be error? Plural on string?
```
markusicu, line 141: opinion: error -- we shouldn't go overboard trying to auto-convert input values, such as from string to number

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    @Test
+    public void testPluralWithOffset() {
+        String message = ""
+                + "match {$count :plural offset=2}\n"
+                + " when 1 {Anna}\n"
+                + " when 2 {Anna and Bob}\n"
+                + " when one {Anna, Bob, and {$count :number offset=2} other guest}\n"
+                + " when * {Anna, Bob, and {$count :number offset=2} other guests}\n";
```
markusicu, line 325: question: Is there no syntax at all like MF `#` to get the output of the formatted `$count`?
Users should not have to repeat the placeholder details like this.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
```diff
@@ -0,0 +1,461 @@
...
+@RunWith(JUnit4.class)
+@SuppressWarnings("javadoc")
+public class Mf2FeaturesTest extends TestFmwk {
+
+    // November 23, 2022 at 7:42:37.123 PM
+    static final Date TEST_DATE = new Date(1669261357123L);
+
+    @Before
+    public void prepareDefaultLocale() {
+        Locale.setDefault(Locale.US);
```
markusicu, line 29: Is this necessary?
If it is, then you should also later restore the default locale to what it used to be. Otherwise you change the environment for other ICU4J unit tests.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
```diff
@@ -0,0 +1,461 @@
...
+                .expected(expectedEn)
+                .build());
+        TestUtils.runTestCase(new TestCase.Builder()
+                .pattern(message)
+                .arguments(arguments)
+                .locale("ro")
+                .expected(expectedRo)
+                .build());
+
+        Locale.setDefault(Locale.forLanguageTag("ro"));
```
markusicu, line 106: ditto restore the original

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
```diff
@@ -0,0 +1,461 @@
...
+                .expected("From literal: 123.456.789,531!")
+                .build());
+        TestUtils.runTestCase(new TestCase.Builder()
+                .pattern("{From literal: {(123456789.531) :number}!}")
+                .locale("my")
+                .arguments(Args.of("val", value))
+                .expected("From literal: \u1041\u1042\u1043,\u1044\u1045\u1046,\u1047\u1048\u1049.\u1045\u1043\u1041!")
+                .build());
+
+        // Testing that the detection works for various types works (without specifying :number)
```
markusicu, line 233:
```suggestion
        // Testing that the detection works for various types (without specifying :number)
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
```diff
@@ -0,0 +1,396 @@
...
+import com.ibm.icu.util.BuddhistCalendar;
+import com.ibm.icu.util.Calendar;
+import com.ibm.icu.util.GregorianCalendar;
+import com.ibm.icu.util.Measure;
+import com.ibm.icu.util.MeasureUnit;
+
+
+@RunWith(JUnit4.class)
+@SuppressWarnings("javadoc")
+public class MessageFormat2Test extends TestFmwk {
```
markusicu, line 34: Please add a brief javadoc comment (`/** ... */`) for each of the `extends TestFmwk` test classes, saying briefly what they do.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/TestCase.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.Locale.Category;
+import java.util.Map;
+import java.util.StringJoiner;
+
+/** Utility class encapsulating what wee need for a simple test. */
```
markusicu, line 14:
```suggestion
/** Utility class encapsulating what we need for a simple test. */
```

icu4j/main/classes/core/src/com/ibm/icu/message2/FormattedPlaceholder.java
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/IdentityFormatterFactory.java
mihnita: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Args.java
mihnita: Done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
mihnita: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
mihnita: Added it, just for fun :-)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
mihnita: Done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
mihnita: Mentioned in https://github.com/unicode-org/message-format-wg/issues/292
I proposed not to support hoisting, and I don't think there is much resistance to the idea.

mihnita: Added a link to the issue in the code.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/FromJsonTest.java
mihnita: I agree with you.
That is why my implementation returns `"other"`

This Monday we discussed these differences, because I think they reflect either differences in opinions, or unclear spec. So we should address them.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
mihnita: There is no need to do it.
The `#` is replaced by the locale variables.
The next test (`testPluralWithOffsetAndLocalVar`) checks that it works:
```
let $foo = {$count :number offset=2}
match {$foo :plural}
  when 1 {Anna}
  when 2 {Anna and Bob}
  when one {Anna, Bob, and {$foo} other guest}
  when * {Anna, Bob, and {$foo} other guests}
```

It is unclear from the spec if the placeholders in the messages would need to also inherit from the condition in the match.
It was proposed, but I argued against, thinking it would have too many side-effects (it would be as if every condition in the match would declare a "fake" local variable)

We can revisit, if you think we should do that.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
mihnita: Done. Removed the `@Before` and doing set + restore in one method only.

icu4j/main/classes/core/src/com/ibm/icu/message2/OptUtils.java
mihnita: Done, and checked all other files.
Except for `Mf2Parser.java`, which is generated, they are all good.

icu4j/main/classes/core/src/com/ibm/icu/message2/PluralSelectorFactory.java
mihnita: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
mihnita: Done.
Setting and restoring default locale in one method only.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
mihnita: Done.

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
```diff
@@ -81,39 +81,43 @@ public <A extends Appendable> A appendTo(A appendable) {
         try {
             appendable.append(value);
         } catch (IOException e) {
-            throw new RuntimeException("problem appending");
+            throw new RuntimeException("problem appending", e);
```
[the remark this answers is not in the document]
mihnita: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
markusicu: 👍 tnx

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
```diff
@@ -28,7 +28,11 @@
 import com.ibm.icu.util.Measure;
 import com.ibm.icu.util.MeasureUnit;
 
-
+/**
+ * Tests migrated from {@link com.ibm.icu.text.MessageFormat}, to show they look like and that they work.
```
markusicu, line 32:
```suggestion
 * Tests migrated from {@link com.ibm.icu.text.MessageFormat}, to show what they look like and that they work.
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/message2/Mf2FeaturesTest.java
```diff
@@ -115,6 +117,7 @@ public void testDefaultLocale() {
                 .arguments(arguments)
                 .expected(expectedRo)
                 .build());
+        Locale.setDefault(Locale.US);
```
markusicu, line 120: Please restore the original default, rather than setting a hardcoded en-US locale.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+            if (functionName == null) {
+                throw new IllegalArgumentException("Object to format without a function, and unknown type: "
+                        + toFormat.getClass().getName());
+            }
+        }
+
+        FormatterFactory func = standardFunctionRegistry.getFormatter(functionName);
+        if (func == null) {
+            func = customFunctionRegistry.getFormatter(functionName);
+        }
```
markusicu, line 111: ditto which one to look in first? probably same elsewhere for selectors etc.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+        return func;
```
markusicu, line 112: question: exception if null?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+ * Takes an {@link Mf2DataModel} and formats it to a {@link String}
+ * (and later on we will also implement formatting to a {@code FormattedMessage}).
+ */
+// TODO: move this in the MessageFormatter
+class Mf2DataModelFormatter {
+    private final Locale locale;
+    private final Mf2DataModel dm;
+
+    final Mf2FunctionRegistry standardFunctionRegistry;
+    final Mf2FunctionRegistry customFunctionRegistry;
```
markusicu, line 34: totally optional: s/...FunctionRegistry/...Functions/ to make the code a little shorter, which I think will make it a little more readable

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+        return func;
+    }
+
+    String format(Map<String, Object> arguments) {
+        Pattern patternToRender;
+        List<Expression> selectors = dm.getSelectors();
+        if (selectors.isEmpty()) {
+            patternToRender = dm.getPattern();
+        } else {
```
markusicu, line 120: optional: How about moving this "else" branch into a function like
```
    patternToRender = selectPattern(...);
```
?
I think that might make both the format() method and the selection code easier to follow.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                String functionName = selector.getFunctionName();
+                SelectorFactory funcFactory = standardFunctionRegistry.getSelector(functionName);
+                if (funcFactory == null) {
+                    funcFactory = customFunctionRegistry.getSelector(functionName);
+                }
+                if (funcFactory != null) {
+                    Map<String, Object> opt = new HashMap<>();
+                    for (Entry<String, Value> e : selector.getOptions().entrySet()) {
+                        opt.put(e.getKey(), variableToObjectEx(e.getValue(), arguments));
+                    }
```
markusicu, line 133: Question 1: This looks a bit like mf2OptToVariableOptions() except that that also checks for value.isVariable(). Should you use that helper method?

Question 2: Shouldn't we collect the fixed options at this stage? Or is this intentionally not calling the helper method so that it collects all of the options?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                    throw new IllegalArgumentException("Mismatch between the number of selectors and the number of keys: "
+                            + selectors.size() + " vs. " + keysToCheck.size());
+                }
+                boolean matches = true;
+                // Iterate "horizontally", through all matching functions and keys
+                for (int i = 0; i < maxCount; i++) {
+                    Expression selector = selectors.get(i);
+                    String valToCheck = keysToCheck.get(i);
+                    Selector func = selectorFunctions.get(i);
+                    Map<String, Object> options = mf2OptToFixedOptions(selector.getOptions());
```
markusicu, line 161: Questions... I thought we fetched fixed options for the factory and variable options for the actual selector. ??

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                    break;
+                }
+            }
+        }
+
+        // TODO: check that there was an entry with all the keys set to `*`
+        if (patternToRender == null) {
+            // If there was a case with all entries in the keys `*` this should not happen
+            throw new IllegalArgumentException("The selection went wrong, cannot select any option.");
+        }
```
markusicu, line 178: This can only be null if there are selectors and nothing matched, right? If so, then please move it into the selector-handling code (into the "else" branch above, or into the new method if you agree it makes sense to move this code).

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                    }
+                }
+                if (matches) {
+                    patternToRender = variant.getValue();
+                    break;
+                }
+            }
+        }
+
+        // TODO: check that there was an entry with all the keys set to `*`
```
markusicu, line 174: That should probably be checked while building the data model, not while formatting, right?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+            // If there was a case with all entries in the keys `*` this should not happen
+            throw new IllegalArgumentException("The selection went wrong, cannot select any option.");
+        }
+
+        StringBuilder result = new StringBuilder();
+        for (Part part : patternToRender.getParts()) {
+            if (part instanceof Text) {
+                result.append(part);
+            } else if (part instanceof Expression) { // Placeholder is an Expression
+                FormattedPlaceholder fr = formatPlaceholder((Expression) part, arguments, false);
```
markusicu, line 185: optional: s/fr/fp/  ?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+                FormattedValue fv = fr.getFormattedValue();
+                if (fr.getInput() == null && (fv == null || fv.toString() == null)) {
```
markusicu, line 187: fv cannot be null any more; and why would its .toString() be null?

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+     *
+     * <code><pre>
+     * let $l1 = {$l4 :number}
+     * let $l2 = {$l1 :number}
+     * let $l3 = {$l2 :number}
+     * let $l4 = {$l3 :number}
+     * </pre></code>
+     *
+     * We can keep track of the calls (complicated and expensive).
+     * Or we can forbid the use of variables before they are declared, but that is not in the spec (yet?).
```
markusicu, line 214: +1 "KISS"

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+     */
+    private Object variableToObjectEx(Value value, Map<String, Object> arguments, int level) {
```
markusicu, line 216: I don't see any call site where level != 0.  ??

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+            if (operand.isLiteral()) { // "...{(1234.56) :number}..."
+                // If it is a literal, return the string itself
+                toFormat = operand.getLiteral();
+            } else if (operand.isVariable()) {
+                String varName = operand.getVariableName();
+                if (!localExpression) {
+                    Expression localPh = dm.getLocalVariables().get(varName);
+                    if (localPh != null) {
+                        // If it is a local variable, we need to format that (recursive)
+                        return formatPlaceholder(localPh, arguments, true);
```
markusicu, line 251: I am surprised that the local variables aren't evaluated immediately when they are defined, before the actual pattern is formatted. Could you please add a TODO to think about it once more?

markusicu (commented):
reviewed the formatter

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+    private final Locale locale;
+    private final Mf2DataModel dm;
+
+    final Mf2FunctionRegistry standardFunctionRegistry;
+    final Mf2FunctionRegistry customFunctionRegistry;
+
+    Mf2DataModelFormatter(Mf2DataModel dm, Locale locale, Mf2FunctionRegistry customFunctionRegistry) {
+        this.locale = locale;
+        this.dm = dm;
+        this.customFunctionRegistry = customFunctionRegistry == null ? Mf2FunctionRegistry.builder().build() : customFunctionRegistry;
```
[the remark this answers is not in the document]
mihnita: Done
(went with the singleton)

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
```diff
@@ -0,0 +1,277 @@
...
+        if (functionName == null || functionName.isEmpty()) {
+            if (toFormat == null) {
+                // The object to format is null, and no function provided.
+                return null;
+            }
+            Class<?> clazz = toFormat.getClass();
+            functionName = standardFunctionRegistry.getDefaultFormatterNameForType(clazz);
+            if (functionName == null) {
+                functionName = customFunctionRegistry.getDefaultFormatterNameForType(clazz);
+            }
```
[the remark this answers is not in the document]
mihnita: Q1: in is not specified who wins. I can open a ticket.

I think one can argue both ways...
I decided to have the standard win on the idea that the standard is the standard and one should not mess with it.
If you want a different behavior, make a custom.

Q2. I think we need both.
For example you format `"... {$someone :person} ... was born on {$dob :date skeleton=yMMMd}..."`

But the I also wanted to talk with you (after the release pressure is gone) about being able to separate the standard formatters for backport (for example implement them on top of Android APIs).

But yes, the current implementation ties them together, because it is in ICU anyway.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Same answer.
I don't feel strongly one way or another, and I can see pros and cons both ways.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done.
Moved in the new method.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Yes.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: ACK. I'm all for it.
We have an issue opened for MFWG to put that in the spec.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done.
Removed (it was possible to call this recursively at some point, but it went through some refactoring.
Added unit test to be 100% I don't loop.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done.
Added TODO.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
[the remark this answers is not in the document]
mihnita: Discussed.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Added TODO

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Added TODO

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Added TODO(before 72)

mihnita: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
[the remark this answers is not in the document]
markusicu: Part of the discussion: There should not be a name collision, because a custom registry should not use standard function names. The MFWG has discussed a naming convention, such as custom function names containing dots (e.g., "com.google.somegooglefunction").

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
markusicu: Discussed: Evaluating local variables (an unfortunate name) immediately might waste time because not all of them may be used by the selected pattern. Lazy-once eval should be possible. The ICU User Guide (future chapter about MF2) and MFWG spec should say that the time of evaluation may vary. This might be observable if a variable expression yields an error when evaluated, which means that depending on the strategy some implementation may throw an error and another might ignore a locale variable.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done
And added unit tests.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Done
And added unit tests.

markusicu:
lgtm pse squash

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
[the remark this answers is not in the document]
mihnita: Created issue: https://github.com/unicode-org/message-format-wg/issues/298

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModelFormatter.java
mihnita: Created issue: https://github.com/unicode-org/message-format-wg/issues/299

PR #2173, opened 2022-09-02
ICU-22116 Update CI job for ICU4J to use Java 8 instead of Java 7

echeran:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22116

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -30,8 +30,8 @@ jobs:
            # Regex note: (?! ... ) is a negative lookahead. Succeed if the pattern is not present. 
           set +o pipefail && make doc 2>&1 | tee doxygen.log && ( ! grep -P 'warning:(?! .* file .?Doxyfile)' doxygen.log )
 
-  # Java7 ICU4J build and unit test
-  java7-icu4j-build-and-test:
+  # Java8 ICU4J build and unit test
+  java8-icu4j-build-and-test:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout and setup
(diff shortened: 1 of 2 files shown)


echeran:
Note: This is a straightforward change from Java 7 to Java 8, and things seem to pass.

I can also update the CI job step that automatically installs Java for us.  Newer versions of that Java setup job step want us to also specify the [name of the distribution of OpenJDK](https://github.com/actions/setup-java#supported-distributions) that we use (not just the version number).  If you want, I can do the update of the Java setup step, including that preference, here too. I don't think people had an opinion on the OpenJDK distribution when I asked previously, and given the advice of Adopt OpenJDK users to use Eclipse Temurin OpenJDK, I would just use 

cc @mihnita

markusicu:
> I can also update the CI job step that automatically installs Java for us. Newer versions of that Java setup job step want us to also specify the [name of the distribution of OpenJDK](https://github.com/actions/setup-java#supported-distributions) that we use (not just the version number). If you want, I can do the update of the Java setup step, including that preference, here too. I don't think people had an opinion on the OpenJDK distribution when I asked previously, and given the advice of Adopt OpenJDK users to use Eclipse Temurin OpenJDK, I would just use

Ok. I didn't actually know that there are multiple "distributions" of OpenJDK... given the explanation, and the GH actions docs, temurin sounds fine.

.github/workflows/icu_ci.yml
```diff
@@ -43,27 +43,6 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: '8'
```
markusicu, line 45: I am surprised that this line already said '8' ...

markusicu (approved):
lgtm tnx!

Happy to re-approve for more changes, such as the distribution.

We should probably think about some piece of code that would benefit from Java 8, to demonstrate that it actually works. Maybe something in the tests first?

.github/workflows/icu_ci.yml
gnrunge: Further down in that file we test ICU4J with java-version: '11' and java-version: '16', respectively.

gnrunge (approved):

echeran:
> We should probably think about some piece of code that would benefit from Java 8, to demonstrate that it actually works. Maybe something in the tests first?

Done. I updated a test using the Java 8 Streams/Functions constructs.  

In general, there are tradeoffs of this slightly more functional style, of course.  The benefits are more readable code, less code, less code with manual iteration, so less surface area for mistakes (ex: off-by-one errors).  The costs would be performance -- you're (probably?) creating and GC'ing a lot of objects.  Even though a lot of the Java GCs coming out since Java 8 are tuned to perform better for lots of small objects, to help with Java 8 and the many new FP languages built on the JVM, it makes sense to start making such Java 8 changes by putting them in tests.  I don't know what the perf impacts would be for such changes in hot code paths in the main code (ex: would function/closure literals get translated into local variables at compile time, and would that mitigate the risks much?)

markusicu (approved):
lgtm / nice! / tnx!

PR #2175, opened 2022-09-06
ICU-21148 Consistently use standard lowercase true/false everywhere.

roubert:
This is the normal standard way in C, C++ as well as Java and there's no
longer any reason for ICU to be different. The various internal macros
providing custom boolean constants can all be deleted and code as well
as documentation can be updated to use lowercase true/false everywhere.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21148
- [X] Required: Each commit message must be prefixed with a JIRA Issue number.>


icu4c/source/test/testdata/te.txt
```diff
@@ -96,7 +96,7 @@ te
     collations {
       standard { 
         Version { "1.0" }
-        Override { "FALSE" }
+        Override { "false" }
```
markusicu, line 99: This is not C/C++ code. Safer to leave this one as is.

markusicu (commented):
lgtm tnx!

One nit, and I sent a question to the team about removing even the option to define the formerly-stable FALSE & TRUE macros.

icu4c/source/test/testdata/te.txt
roubert: On the contrary, I think there's value in changing to lowercase absolutely everywhere we can, so that no-one ever has to remember that there is some particular corner of ICU where it should be uppercase for no obvious reason.

icu4c/source/common/unicode/umachine.h
```diff
@@ -268,50 +268,6 @@
...
-#elif defined(U_COMBINED_IMPLEMENTATION) || \
-        defined(U_COMMON_IMPLEMENTATION) || defined(U_I18N_IMPLEMENTATION) || \
-        defined(U_IO_IMPLEMENTATION) || defined(U_LAYOUTEX_IMPLEMENTATION) || \
-        defined(U_TOOLUTIL_IMPLEMENTATION)
-    // Inside ICU: Keep FALSE & TRUE available.
-#   define U_DEFINE_FALSE_AND_TRUE 1
-#else
-    // Outside ICU: Avoid collision with non-macro definitions of FALSE & TRUE.
-#   define U_DEFINE_FALSE_AND_TRUE 0
-#endif
```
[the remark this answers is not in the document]
roubert: I've now instead simply reverted the public facing code here to what it used to be, so that people using this feature won't experience any change in behaviour at all.

icu4c/source/common/unicode/umachine.h
```diff
@@ -268,50 +268,6 @@
...
-#ifndef FALSE
-/**
- * The FALSE value of a UBool.
- *
- * @deprecated ICU 68 Use standard "false" instead.
- */
-#   define FALSE 0
-#endif
-#endif  // U_DEFINE_FALSE_AND_TRUE
```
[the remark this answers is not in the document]
roubert: What do you mean? What change do you propose?

icu4c/source/test/testdata/te.txt
markusicu: But this is a string value, not code. I haven't looked at the code that parses this string to see if it handles other spellings. Have you?

roubert: No, I have not looked, I have simply changed them all, that's an advantage too, no need to look.

markusicu: Well, it kind of doesn't make sense, since this is not a C/C++ constant nor talking about one -- but I did look, and this Override value seems to be never read, so it shouldn't matter.

markusicu (approved):
lgtm

PR #2176, opened 2022-09-08
ICU-21957 integrate CLDR release-42-alpha3 to ICU main for 72

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21957
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

Integrate CLDR release-42-alpha3 to ICU main for 72. Some notes:
- Emoji collation updates per CLDR-15965
- Unicode 15 collation data updates per CLDR-15516
- Arabic currency formats for 'latn' digits have been further updated so currency symbol follows number, per CLDR-15992
- Several updates to personNames data per CLDR-15977, including removing some spaces in formats for ja/lo/yue, and other changes to formats for zh/zh_Hant
- Compact decimal data for Japanese has been extended to values above 100000000000000 (i.e. above COMPACT_MAX_DIGITS=15) per CLDR-15966. There is nothing in the CLDR spec that disallows this; however ICU code cannot currently handle these new large values, so I have changed to code to skip loading them.
- Some formerly-provisional data has been promoted to contributed in certain locales, so it gets picked up by ICU.

PR #2181, opened 2022-09-09
ICU-22143 Increase COMPACT_MAX_DIGITS from 15 to 20, needed for new ja data

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22143

Well this indeed turned out to be very straightforward, just increasing the value of the constant; then everything else was handled (including locales that do not have data that covers all the way up to COMPACT_MAX_DIGITS). Added tests for ja.

This will be useful to have back in the ICU4J libs in CLDR (e.g. for chart generation), will update those soon.


pedberg-icu:
@sffc @macchiati This should be a very easy review...thanks

sffc (changes requested):
Please undo the change from [#2176](https://github.com/unicode-org/icu/pull/2176) because the assertion did its job of alerting us that there was a problem. Without the assertion we may never know that there is a problem.

pedberg-icu:
> Please undo the change from [#2176](https://github.com/unicode-org/icu/pull/2176) because the assertion did its job of alerting us that there was a problem. Without the assertion we may never know that there is a problem.

I am not going to just undo the change from #2176 because (a) the asset was in the wrong place, after we would already have had a crash from writing past an array boundary, and (b) I strongly believe we should not have asserts in production code, especially for conditions that do not cause a fatal problem with the operation of ICU (and this condition does not rise to that level). I agree we should have some kind of notification that there is an issue; do you have a suggestion for another more benign kind of notification?

sffc:
I think a debug assertion is the right behavior. `U_ASSERT` is a debug assertion.

sffc:
A unit test is also fine, but more work.

pedberg-icu:
> I think a debug assertion is the right behavior. `U_ASSERT` is a debug assertion.

OK, I can restore the U_ASSERT for C The Java assert is not necessarily debug-only, depends on the runtime and/or other parameters. In any case, we only need one of these to check the data, and it is probably more useful in ICU4C since that is where we first check the data.

pedberg-icu:
@sffc OK restored the U_ASSERT in a better place, PTAL, thanks.

sffc (approved):

PR #2182, opened 2022-09-09
ICU-22122 Support Locale Tags (ms, mu and rg)

younies:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22122

### DIFF
--- a/icu4c/source/i18n/units_data.cpp
+++ b/icu4c/source/i18n/units_data.cpp
@@ -5,10 +5,12 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include "bytesinkutil.h"
 #include "cstring.h"
 #include "number_decimalquantity.h"
 #include "resource.h"
 #include "uassert.h"
+#include "unicode/locid.h"
 #include "unicode/unistr.h"
 #include "unicode/ures.h"
 #include "units_data.h"
@@ -387,24 +389,97 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
     ures_getAllItemsWithFallback(unitsBundle.getAlias(), "unitPreferenceData", sink, status);
 }
 
-// TODO: make outPreferences const?
-//
-// TODO: consider replacing `UnitPreference **&outPreferences` with slice class
-// of some kind.
-void U_I18N_API UnitPreferences::getPreferencesFor(StringPiece category, StringPiece usage,
-                                                   StringPiece region,
-                                                   const UnitPreference *const *&outPreferences,
-                                                   int32_t &preferenceCount, UErrorCode &status) const {
-    int32_t idx = getPreferenceMetadataIndex(&metadata_, category, usage, region, status);
-    if (U_FAILURE(status)) {
-        outPreferences = nullptr;
-        preferenceCount = 0;
-        return;
+CharString getKeyWordValue(const Locale &locale, StringPiece kw, UErrorCode &status) {
+    CharString result;
+    if (U_FAILURE(status)) { return result; }
+    {
+        CharStringByteSink sink(&result);
+        locale.getKeywordValue(kw, sink, status);
     }
+    if (U_SUCCESS(status) && result.isEmpty()) {
+        status = U_MISSING_RESOURCE_ERROR;
+    }
+    return result;
+}
+
+MaybeStackVector<UnitPreference>
+    U_I18N_API UnitPreferences::getPreferencesFor(StringPiece category, StringPiece usage,
+                                                  const Locale &locale, UErrorCode &status) const {
+
+    MaybeStackVector<UnitPreference> result;
+
+    // TODO: remove this once all the categories are allowed.
+    UErrorCode internalMuStatus = U_ZERO_ERROR;
+    if (category.compare("temperature") == 0) {
+        CharString localeUnitCharString = getKeyWordValue(locale, "mu", internalMuStatus);
+        if (U_SUCCESS(internalMuStatus)) {
+            // TODO: use the unit category as Java especially when all the categories are allowed..
+            if (localeUnitCharString == "celsius"       //
+                || localeUnitCharString == "fahrenheit" //
+                || localeUnitCharString == "kelvin"     //
+            ) {
+                UnitPreference unitPref;
+                unitPref.unit.append(localeUnitCharString, status);
+                result.emplaceBackAndCheckErrorCode(status, unitPref);
+                return result;
+            }
+        }
+    }
+
+    CharString region(locale.getCountry(), status);
+
+    // Check the locale system tag, e.g `ms=metric`.
+    UErrorCode internalMeasureTagStatus = U_ZERO_ERROR;
+    CharString localeSystem = getKeyWordValue(locale, "measure", internalMeasureTagStatus);
+    bool isLocaleSystem = false;
+    if (U_SUCCESS(internalMeasureTagStatus)) {
+        if (localeSystem == "metric") {
+            region.clear();
+            region.append("001", status);
+            isLocaleSystem = true;
+        } else if (localeSystem == "ussystem") {
+            region.clear();
+            region.append("US", status);
+            isLocaleSystem = true;
+        } else if (localeSystem == "uksystem") {
+            region.clear();
+            region.append("GB", status);
+            isLocaleSystem = true;
+        }
+    }
+
+    // Check the region tag, e.g. `rg=uszzz`.
+    if (!isLocaleSystem) {
+        UErrorCode internalRgTagStatus = U_ZERO_ERROR;
+        CharString localeRegion = getKeyWordValue(locale, "rg", internalRgTagStatus);
+        if (U_SUCCESS(internalRgTagStatus) && localeRegion.length() >= 3) {
+            if (localeRegion == "default") {
+                region.clear();
+                region.append(localeRegion, status);
+            } else if (localeRegion[0] >= '0' && localeRegion[0] <= '9') {
+                region.clear();
+                region.append(localeRegion.data(), 3, status);
+            } else {
+                // Take the first two character and capitalize them.
+                region.clear();
+                region.append(uprv_toupper(localeRegion[0]), status);
+                region.append(uprv_toupper(localeRegion[1]), status);
+            }
+        }
+    }
+
+    int32_t idx =
+        getPreferenceMetadataIndex(&metadata_, category, usage, region.toStringPiece(), status);
+    if (U_FAILURE(status)) {
+        return result;
+    }
+
     U_ASSERT(idx >= 0); // Failures should have been taken care of by `status`.
     const UnitPreferenceMetadata *m = metadata_[idx];
-    outPreferences = unitPrefs_.getAlias() + m->prefsOffset;
-    preferenceCount = m->prefsCount;
+    for (int32_t i = 0; i < m->prefsCount; i++) {
+        result.emplaceBackAndCheckErrorCode(status, *(unitPrefs_[i + m->prefsOffset]));
+    }
+    return result;
 }
 
 } // namespace units
--- a/icu4c/source/i18n/units_router.cpp
+++ b/icu4c/source/i18n/units_router.cpp
@@ -43,17 +43,17 @@ Precision UnitsRouter::parseSkeletonToPrecision(icu::UnicodeString precisionSkel
     return result;
 }
 
-UnitsRouter::UnitsRouter(StringPiece inputUnitIdentifier, StringPiece region, StringPiece usage,
+UnitsRouter::UnitsRouter(StringPiece inputUnitIdentifier, const Locale &locale, StringPiece usage,
                          UErrorCode &status) {
-    this->init(MeasureUnit::forIdentifier(inputUnitIdentifier, status), region, usage, status);
+    this->init(MeasureUnit::forIdentifier(inputUnitIdentifier, status), locale, usage, status);
 }
 
-UnitsRouter::UnitsRouter(const MeasureUnit &inputUnit, StringPiece region, StringPiece usage,
+UnitsRouter::UnitsRouter(const MeasureUnit &inputUnit, const Locale &locale, StringPiece usage,
                          UErrorCode &status) {
-    this->init(std::move(inputUnit), region, usage, status);
+    this->init(std::move(inputUnit), locale, usage, status);
 }
 
-void UnitsRouter::init(const MeasureUnit &inputUnit, StringPiece region, StringPiece usage,
+void UnitsRouter::init(const MeasureUnit &inputUnit, const Locale &locale, StringPiece usage,
                        UErrorCode &status) {
 
     if (U_FAILURE(status)) {
@@ -73,22 +73,19 @@ void UnitsRouter::init(const MeasureUnit &inputUnit, StringPiece region, StringP
         return;
     }
 
-    const UnitPreference *const *unitPreferences;
-    int32_t preferencesCount = 0;
-    prefs.getPreferencesFor(category.toStringPiece(), usage, region, unitPreferences, preferencesCount,
-                            status);
-
-    for (int i = 0; i < preferencesCount; ++i) {
-        U_ASSERT(unitPreferences[i] != nullptr);
-        const auto &preference = *unitPreferences[i];
+    const MaybeStackVector<UnitPreference> unitPrefs =
+        prefs.getPreferencesFor(category.toStringPiece(), usage, locale, status);
+    for (int32_t i = 0, n = unitPrefs.length(); i < n; ++i) {
+        U_ASSERT(unitPrefs[i] != nullptr);
+        const auto preference = unitPrefs[i];
 
         MeasureUnitImpl complexTargetUnitImpl =
-            MeasureUnitImpl::forIdentifier(preference.unit.data(), status);
+            MeasureUnitImpl::forIdentifier(preference->unit.data(), status);
         if (U_FAILURE(status)) {
             return;
         }
 
-        UnicodeString precision = preference.skeleton;
+        UnicodeString precision = preference->skeleton;
 
         // For now, we only have "precision-increment" in Units Preferences skeleton.
         // Therefore, we check if the skeleton starts with "precision-increment" and force the program to
@@ -103,7 +100,7 @@ void UnitsRouter::init(const MeasureUnit &inputUnit, StringPiece region, StringP
         outputUnits_.emplaceBackAndCheckErrorCode(status,
                                                   complexTargetUnitImpl.copy(status).build(status));
         converterPreferences_.emplaceBackAndCheckErrorCode(status, inputUnitImpl, complexTargetUnitImpl,
-                                                           preference.geq, std::move(precision),
+                                                           preference->geq, std::move(precision),
                                                            conversionRates, status);
 
         if (U_FAILURE(status)) {
--- a/icu4c/source/test/intltest/units_data_test.cpp
+++ b/icu4c/source/test/intltest/units_data_test.cpp
@@ -5,7 +5,9 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include "cstring.h"
 #include "measunit_impl.h"
+#include "unicode/locid.h"
 #include "units_data.h"
 
 #include "intltest.h"
@@ -118,8 +120,7 @@ void UnitsDataTest::testGetPreferencesFor() {
         {"Unknown usage US", "length", "foobar", "US", USLenMax, USLenMin},
         {"Unknown usage 001", "length", "foobar", "XX", WorldLenMax, WorldLenMin},
         {"Fallback", "length", "person-height-xyzzy", "DE", "centimeter", "centimeter"},
-        {"Fallback twice", "length", "person-height-xyzzy-foo", "DE", "centimeter",
-         "centimeter"},
+        {"Fallback twice", "length", "person-height-xyzzy-foo", "DE", "centimeter", "centimeter"},
         // Confirming results for some unitPreferencesTest.txt test cases
         {"001 area", "area", "default", "001", "square-kilometer", "square-centimeter"},
         {"GB area", "area", "default", "GB", "square-mile", "square-inch"},
@@ -140,18 +141,20 @@ void UnitsDataTest::testGetPreferencesFor() {
 
     for (const auto &t : testCases) {
         logln(t.name);
-        const UnitPreference *const *prefs;
-        int32_t prefsCount;
-        preferences.getPreferencesFor(t.category, t.usage, t.region, prefs, prefsCount, status);
+        CharString localeID;
+        localeID.append("und-", status); // append undefined language.
+        localeID.append(t.region, status);
+        Locale locale(localeID.data());
+        auto unitPrefs = preferences.getPreferencesFor(t.category, t.usage, locale, status);
         if (status.errIfFailureAndReset("getPreferencesFor(\"%s\", \"%s\", \"%s\", ...", t.category,
                                         t.usage, t.region)) {
             continue;
         }
-        if (prefsCount > 0) {
+        if (unitPrefs.length() > 0) {
             assertEquals(UnicodeString(t.name) + " - max unit", t.expectedBiggest,
-                         prefs[0]->unit.data());
+                         unitPrefs[0]->unit.data());
             assertEquals(UnicodeString(t.name) + " - min unit", t.expectedSmallest,
-                         prefs[prefsCount - 1]->unit.data());
+                         unitPrefs[unitPrefs.length() - 1]->unit.data());
         } else {
             errln(UnicodeString(t.name) + ": failed to find preferences");
         }
--- a/icu4c/source/test/intltest/units_test.cpp
+++ b/icu4c/source/test/intltest/units_test.cpp
@@ -948,7 +948,11 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
         return;
     }
 
-    UnitsRouter router(inputMeasureUnit, region, usage, status);
+    CharString localeID;
+    localeID.append("und-", status); // append undefined language.
+    localeID.append(region, status);
+    Locale locale(localeID.data());
+    UnitsRouter router(inputMeasureUnit, locale, usage, status);
     if (status.errIfFailureAndReset("UnitsRouter(<%s>, \"%.*s\", \"%.*s\", status)",
                                     inputMeasureUnit.getIdentifier(), region.length(), region.data(),
                                     usage.length(), usage.data())) {
@@ -976,7 +980,7 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
 
     // Test UnitsRouter created with CLDR units identifiers.
     CharString inputUnitIdentifier(inputUnit, status);
-    UnitsRouter router2(inputUnitIdentifier.data(), region, usage, status);
+    UnitsRouter router2(inputUnitIdentifier.data(), locale, usage, status);
     if (status.errIfFailureAndReset("UnitsRouter2(<%s>, \"%.*s\", \"%.*s\", status)",
                                     inputUnitIdentifier.data(), region.length(), region.data(),
                                     usage.length(), usage.data())) {
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
@@ -4,14 +4,28 @@ package com.ibm.icu.impl.units;
 
 import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
 
 import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.impl.ICUResourceBundle;
 import com.ibm.icu.impl.UResource;
+import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.UResourceBundle;
 
 public class UnitPreferences {
+    private static final Map<String, String> measurementSystem;
+
+    static {
+        Map tempMS = new HashMap<String, String>();
+        tempMS.put("metric", "001");
+        tempMS.put("ussystem", "US");
+        tempMS.put("uksystem", "GB");
+        measurementSystem = Collections.unmodifiableMap(tempMS);
+    }
+
 
     private HashMap<String, HashMap<String, UnitPreference[]>> mapToUnitPreferences = new HashMap<>();
 
@@ -56,7 +70,48 @@ public class UnitPreferences {
         return result.toArray(new String[0]);
     }
 
-    public UnitPreference[] getPreferencesFor(String category, String usage, String region) {
+    public UnitPreference[] getPreferencesFor(String category, String usage, ULocale locale, UnitsData data) {
+        // TODO: remove this condition when all the categories are allowed.
+        if (category.equals("temperature")) {
+            String localeUnit = locale.getKeywordValue("mu");
+            String localeUnitCategory;
+            try {
+                localeUnitCategory = localeUnit == null ? null : data.getCategory(MeasureUnitImpl.forIdentifier(localeUnit));
+            } catch (Exception e) {
+                localeUnitCategory = null;
+            }
+
+            if (localeUnitCategory != null && category.equals(localeUnitCategory)) {
+                UnitPreference[] preferences = {new UnitPreference(localeUnit, null, null)};
+                return preferences;
+            }
+        }
+
+        String region = locale.getCountry();
+
+        // Check the locale system tag, e.g `ms=metric`.
+        String localeSystem = locale.getKeywordValue("measure");
+        boolean isLocaleSystem = false;
+        if (measurementSystem.containsKey(localeSystem)) {
+            isLocaleSystem = true;
+            region = measurementSystem.get(localeSystem);
+        }
+
+        // Check the region tag, e.g. `rg=uszzz`.
+        if (!isLocaleSystem) {
+            String localeRegion = locale.getKeywordValue("rg");
+            if (localeRegion != null && localeRegion.length() >= 3) {
+                if (localeRegion.equals("default")) {
+                    region = localeRegion;
+                } else if (Character.isDigit(localeRegion.charAt(0))) {
+                    region = localeRegion.substring(0, 3); // e.g. 001
+                } else {
+                    // Capitalize the first two character of the region, e.g. ukzzzz or usca
+                    region = localeRegion.substring(0, 2).toUpperCase(Locale.ROOT);
+                }
+            }
+        }
+
         String[] subUsages = getAllUsages(usage);
         UnitPreference[] result = null;
         for (String subUsage :
@@ -64,6 +119,7 @@ public class UnitPreferences {
             result = getUnitPreferences(category, subUsage, region);
             if (result != null) break;
         }
+
         // TODO: if a category is missing, we get an assertion failure, or we
         // return null, causing a NullPointerException. In C++, we return an
         // U_MISSING_RESOURCE_ERROR error.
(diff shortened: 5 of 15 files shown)


icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -489,6 +489,45 @@ public void notationCompact() {
                 "Kun");
     }
 
+    @Test
+    public void unitUMinLocale() {
+        assertFormatSingleMeasure(
```
macchiati, line 494: Should test all combinations of:
2 alternates for the input measure: Fahrenheit and celsius
3 alternates for usage: none, "default", and "other".
2 alternates for base locale: en, en_AU, and de
each combination of +/- mu, +/- ms and +/- rg, eg 8 combinations (including none).

Easiest if driven from array of test cases + expected output

macchiati (changes requested):
Looks like a good approach.
Some things to fix: additional test cases and other tests are failing.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -489,6 +489,45 @@ public void notationCompact() {
...
+                "0°F");
+
+        assertFormatSingleMeasure(
+                "Measure format for `en-US` take mu as celsius with usage",
+                "measure-unit/temperature-fahrenheit usage/default",
+                "unit/fahrenheit usage/default",
+                NumberFormatter.with().unit(MeasureUnit.FAHRENHEIT).usage("default"),
+                ULocale.forLanguageTag("en-US-u-mu-celsius"),
+                new Measure(0, MeasureUnit.FAHRENHEIT),
+                "-18°C");
```
sffc, line 510: Please also test the outputUnit on the FormattedNumber.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -56,9 +61,18 @@ private static String[] getAllUsages(String usage) {
         return result.toArray(new String[0]);
     }
 
-    public UnitPreference[] getPreferencesFor(String category, String usage, String region) {
+    public UnitPreference[] getPreferencesFor(String category, String usage, ULocale locale) {
+        if (category.equals("temperature")) {
```
sffc, line 65: Question: Do we support `-u-ms` yet?

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -56,9 +61,18 @@ private static String[] getAllUsages(String usage) {
...
+            String localeUnit = locale.getKeywordValue("measurement-unit");
```
sffc, line 66: Question: when did the "measurement-unit" identifier get added to ICU?

sffc (approved):
Praise: Good, clean solution!

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
younies: now, yes

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
younies: I do not know, but it is added to this file: ../icu4c/source/data/misc/keyTypeData.txt

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -5,17 +5,27 @@
...
+    private final HashMap<String, String> measurementSystem;
+
 
     private HashMap<String, HashMap<String, UnitPreference[]>> mapToUnitPreferences = new HashMap<>();
 
     public UnitPreferences() {
+        HashMap<String, String> tempMS = new HashMap<>();
+        tempMS.put("metric", "001");
+        tempMS.put("ussystem", "US");
+        tempMS.put("uksystem", "UK");
```
sffc, line 26: GB is the actual region code

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -56,14 +66,46 @@ private static String[] getAllUsages(String usage) {
...
+        boolean isLocaleSystem = false;
+        if (measurementSystem.containsKey(localeSystem)){
+            isLocaleSystem = true;
+            region = measurementSystem.get(localeSystem);
+        }
+
+        String localeRegion = locale.getKeywordValue("rg");
+        if (!isLocaleSystem && localeRegion != null) {
+            // we need to check the region and the capitalization.
+            region = localeRegion.replaceAll("zzzz", "").toUpperCase();
```
sffc, line 98: The following are all potentially valid:

- `gbzzzz`
- `150zzzz`
- `usca`

You want to select the first 2 chars if the first char is a letter, and the first 3 chars if the first char is a number.

sffc: In C++, you can make a stack string: `UChar region[] = {0, 0, 0, 0};`

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
macchiati: doooh, I missed that!

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
macchiati: That was imported from CLDR, a long alias.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
macchiati: That code is inefficient. It is simpler to just extract the first two characters, eg 

localeRegion.substring(0,2).toUpperCase(Locale.ROOT);

Also, the upper/lowercase methods have to have an explicit locale. Otherwise they can screwup if the default locale is, say, Turkish.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsRouter.java
```diff
@@ -47,16 +48,16 @@ public class UnitsRouter {
     private ArrayList<ConverterPreference> converterPreferences_ = new ArrayList<>();
 
     public UnitsRouter(String inputUnitIdentifier, String region, String usage) {
-        this(MeasureUnitImpl.forIdentifier(inputUnitIdentifier), region, usage);
+        this(MeasureUnitImpl.forIdentifier(inputUnitIdentifier), ULocale.forLanguageTag(region), usage);
```
macchiati, line 51: This will not work. forLanguageTag('US') will think it is a language called 'us'. You would need to use ULocale.forLanguageTag("und-" + region) or use the ULocale builder.

macchiati: Note, I suspect that longer term it would be better to push these up to the caller, eg make the call sites supply a locale.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -686,7 +686,8 @@ public String toString() {
         }
 
         for (TestCase testCase : tests) {
-            UnitsRouter router = new UnitsRouter(testCase.inputUnit.second, testCase.region, testCase.usage);
+            UnitsRouter router = new UnitsRouter(testCase.inputUnit.second, ULocale.forLanguageTag(testCase.region),
```
macchiati, line 689: Same problem with forLanguageTag from just a reason. Look for all those cases.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
macchiati: You would also want to check that localeRegion.size() > 2 before substringing.

macchiati (changes requested):

icu4c/source/common/uloc.cpp
```diff
@@ -566,6 +566,7 @@ static int32_t locale_canonKeywordName(char *buf, const char *keywordName, UErro
   int32_t keywordNameLen = 0;
 
   for (; *keywordName != 0; keywordName++) {
+    // TODO: fix when the keyword = "measurement-unit".
     if (!UPRV_ISALPHANUM(*keywordName)) {
```
younies, line 570: @sffc , @markusicu 

Because the keywords can only contains numbers or letters, I could not use "measurement-unit" as a key in `Locale::getKeywordValue`. even that "measurement-unit" is the key to get the desired value.

Do you have a solution?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: Done.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: There is a bug, https://unicode-org.atlassian.net/browse/ICU-22154

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
younies: Done

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsRouter.java
younies: This an internal class that only used for exposing UnitsRouter as a class for the end users. Therefore, it is okay to leave this constructor taking the locale as a string. I used `new Ulocale(locale)`

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
markusicu: That alias should not have been defined in CLDR. It is unnecessary, excessively long, and, as @younies found out, violates ICU locale keyword syntax. We are lucky that Younies found this before CLDR 42 was released, because we can still retract this alias. I submitted CLDR-16022 to hopefully prevent new key/value aliases going forward.

icu4c/source/common/uloc.cpp
younies: Sure, I have removed it.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
younies: This is an internal class constructor that is only used for exposing UnitsRouter to the end users. Therefore, it is okay to leave this constructor taking the locale as a string. I used new Ulocale(locale).

The other constructor, which used for NumberFormatter, take the locale as it is provided by the users.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
younies: Thanks Markus, and I have updated this PR to use "mu"

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
markusicu: @macchiati I see what looks like a good set of test cases but not the full set of all combinations that you seemed to ask for. Are you satisfied?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -650,7 +651,12 @@ private void insertData(String category,
                                     ArrayList<Pair<String, String>> outputs /* Unit Identifier, expected value */) {
                 this.category = category;
                 this.usage = usage;
-                this.region = region;
+                if (Character.isDigit(region.charAt(0)) || region.equals("default")) {
+                    this.region = region;
+                } else {
+                    this.region = "en-" + region;
+                }
+                this.locale = ULocale.forLanguageTag(this.region);
```
markusicu, line 659: Same problem as in C++: You need to decide whether you just need a region, or you need a locale ID. Name the parameter accordingly (and parallel with the C++ struct field). If it's just a region, you can use `new ULocale("und", region)`. If it's a locale ID, you can use `new ULocale(localeID)` or `forLanguageTag(lang)`. Make sure that all of the test values are consistently one or the other.

I think in your current Java version, when region="en-GB" you create a ULocale for "en-en-GB".

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+  
+
+    if (U_FAILURE(status) || kwLen == 0) {
+        status = U_MISSING_RESOURCE_ERROR;
+        return result;
+    }
+
+    result.append(charBuffer, kwLen, status);
+    return result;
+}
```
markusicu, line 406:
```suggestion
CharString getKeyWordValue(const Locale &locale, StringPiece kw, UErrorCode &status) {
    CharString result;
    if (U_FAILURE(status)) { return result; }
    {
        CharStringByteSink sink(&result);
        locale.getKeywordValue(kw, sink, status);
    }
    if (U_SUCCESS(status) && kw.isEmpty()) {
        status = U_MISSING_RESOURCE_ERROR;
    }
    return result;
}
```
- easier to use a sink than an output buffer
- any function that takes `UErrorCode` should first check for failure; you can defer that only if you call other functions that take `UErrorCode`
- don't clobber an error code

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+
 // TODO: make outPreferences const?
 //
 // TODO: consider replacing `UnitPreference **&outPreferences` with slice class
 // of some kind.
```
markusicu, line 411: TODOs look obsolete now
```suggestion
```

markusicu (changes requested):

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsRouter.java
markusicu: You still need to be consistent about whether the test uses a region subtag or a locale ID, and construct a Locale/ULocale accordingly. I added more verbose comments on the current version of the code, both in C++ and in Java.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
markusicu: That's not what Mark is talking about (and with "reason" he meant "region"). See above, and see my new comments.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -56,14 +67,52 @@ private static String[] getAllUsages(String usage) {
...
+
+        String localeRegion = locale.getKeywordValue("rg");
+        if (!isLocaleSystem && localeRegion != null && localeRegion.length() >= 3) {
+            if (localeRegion.equals("default")) {
+                region = localeRegion;
+            } else if (Character.isDigit(localeRegion.charAt(0))) {
+                region = localeRegion.substring(0, 3); // e.g. 001
+            } else {
+                // Capitalize the first two character of the region, e.g. ukzzzz or usca
+                region = localeRegion.substring(0, 2).toUpperCase(Locale.ROOT);
```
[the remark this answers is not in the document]
macchiati: There is a check a few lines up that localeRegion.length() >= 3, so we should be ok here.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
macchiati: Yes, you never want to assume the language is English. "und-" is the right value for indeterminant language.

macchiati (changes requested):
Just a few more items to fix. You will also need to squash the commits.

icu4c/source/i18n/units_data.cpp
younies: thanks Markus, done!

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+            if (uprv_strcmp(localeUnitCharString.data(), "celsius") == 0 ||
+                uprv_strcmp(localeUnitCharString.data(), "fahrenheit") == 0 ||
+                uprv_strcmp(localeUnitCharString.data(), "kelvin") == 0) {
+                UnitPreference unitPref;
+                unitPref.unit.append(localeUnitCharString.data(), localeUnitCharString.length(), status);
+                result.emplaceBackAndCheckErrorCode(status, unitPref);
+                return result;
+            }
+        } else {
+            localeTagsStatus = U_ZERO_ERROR;
```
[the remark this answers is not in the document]
younies: Okay, I will do this, I meant `Locale`, which mean is coming from the locale or from checking the `locale`. but I think `local` or `internal` is better prefix.

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+            region.append("GB", status);
+            isLocaleSystem = true;
+        }
+    } else {
+        localeTagsStatus = U_ZERO_ERROR;
+    }
+
+    // Check the region tag, e.g. `rg=uszzz`.
+    CharString localeRegion = getKeyWordValue(locale, "rg", localeTagsStatus);
+    if (!isLocaleSystem && U_SUCCESS(localeTagsStatus) && localeRegion.length() >= 3) {
```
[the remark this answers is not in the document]
younies: good catch, done!

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+        if (uprv_strcmp(localeRegion.data(), "default") == 0) {
```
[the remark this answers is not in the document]
younies: done

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+            region.clear();
+            region.append(localeRegion, status);
+        } else if (localeRegion.data()[0] >= '0' && localeRegion.data()[0] <= '9') {
+            region.clear();
+            region.append(localeRegion.data(), 3, status);
```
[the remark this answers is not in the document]
younies: I have already checked above `if (U_SUCCESS(internalRgTagStatus) && localeRegion.length() >= 3) {`

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +388,104 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+        } else {
+            // Take the first two character and capitalize them.
+            region.clear();
+            region.append(uprv_toupper(localeRegion.data()[0]), status);
+            region.append(uprv_toupper(localeRegion.data()[1]), status);
```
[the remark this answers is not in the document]
younies: done

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -5,17 +5,28 @@
...
+import java.util.Map;
 
 import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.impl.ICUResourceBundle;
 import com.ibm.icu.impl.UResource;
+import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.UResourceBundle;
 
 public class UnitPreferences {
+    private final HashMap<String, String> measurementSystem;
```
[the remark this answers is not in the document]
younies: done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
younies: Done! and I have fixed the test cases.

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -948,7 +948,12 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
         return;
     }
 
-    UnitsRouter router(inputMeasureUnit, region, usage, status);
+    std::string localeStr(region.data(), region.length());
+    if (!((localeStr[0] >= '0' && localeStr[0] <= '9') || localeStr == "default"))
```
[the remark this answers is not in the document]
younies: no, if `default`, I do not change it.

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -948,7 +948,12 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
...
+        localeStr = "en-" + localeStr; // Testing the region, so, it is okay to use any language.
```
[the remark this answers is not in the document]
younies: done!

is it okay to not add curly braces in the testing :)

markusicu:
> You will also need to squash the commits.

Squashing is best at the end, when all feedback is resolved. Now that Younies squashed early, it's impossible to see what changed since my last review :-(

GitHub code reviews are harder than they need to be anyway. Let's not make them even harder.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -56,14 +70,54 @@ private static String[] getAllUsages(String usage) {
...
+                localeUnitCategory = null;
+            }
+
+            if (localeUnitCategory != null && category.equals(localeUnitCategory)) {
+                UnitPreference[] preferences = {new UnitPreference(localeUnit, null, null)};
+                return preferences;
+            }
+        }
+
+        String region = locale.getCountry();
```
sffc, line 90: Suggestion (optional): Make this variable a `CharSequence` and use `localeRegion.subSequence` below so that we can avoid allocating extra strings.

sffc (approved):

icu4c/source/test/intltest/units_test.cpp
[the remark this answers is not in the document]
markusicu: I prefer curly braces because they can prevent bugs. Not as many as in C++, but still.

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +389,96 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
-                                                   int32_t &preferenceCount, UErrorCode &status) const {
-    int32_t idx = getPreferenceMetadataIndex(&metadata_, category, usage, region, status);
+CharString getKeyWordValue(const Locale &locale, StringPiece kw, UErrorCode &status) {
+    CharString result;
+    if (U_FAILURE(status)) { return result; }
+    {
+        CharStringByteSink sink(&result);
+        locale.getKeywordValue(kw, sink, status);
+    }
+    if (U_SUCCESS(status) && kw.empty()) {
```
markusicu, line 399: Actually, is this supposed to test `result.isEmpty()` rather than `kw.empty()`?
This function is never called with an empty kw, right?

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +389,96 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+
+    MaybeStackVector<UnitPreference> result;
+
+    // TODO: remove this once all the categories are allowed.
+    UErrorCode internalMuStatus = U_ZERO_ERROR;
+    if (category.compare("temperature") == 0) {
+        CharString localeUnitCharString = getKeyWordValue(locale, "mu", internalMuStatus);
+        if (U_SUCCESS(internalMuStatus)) {
+            // TODO: use the unit category as Java especially when all the categories are allowed..
+            if (localeUnitCharString.toStringPiece().compare("celsius") == 0 ||
```
markusicu, line 417: Sorry for misremembering the internal API :-(
Could you please change all of these from e.g. `localeUnitCharString.toStringPiece().compare("celsius")` to `localeUnitCharString == "celsius"`?

https://github.com/unicode-org/icu/blob/main/icu4c/source/common/charstr.h
```
    bool operator==(StringPiece other) const
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -785,7 +792,14 @@ public TestCase(String name, String category, String usage, String region, Strin
 
         UnitsData data = new UnitsData();
         for (TestCase t : testCases) {
-            UnitPreferences.UnitPreference prefs[] = data.getPreferencesFor(t.category, t.usage, t.region);
+            ULocale testCaseLocale;
+            if (t.region.equals("default") || Character.isDigit(t.region.charAt(0))) {
+                testCaseLocale = new ULocale(t.region);
+            } else {
+                testCaseLocale = new ULocale("und-" + t.region);
+            }
```
sffc, line 800: Here and in C++: you should never create a ULocale with the region as the first subtag.

```suggestion
            ULocale testCaseLocale = new ULocale("und-" + t.region);
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
```diff
@@ -650,7 +651,12 @@ private void insertData(String category,
                                     ArrayList<Pair<String, String>> outputs /* Unit Identifier, expected value */) {
                 this.category = category;
                 this.usage = usage;
-                this.region = region;
+                if (Character.isDigit(region.charAt(0)) || region.equals("default")) {
+                    this.region = region;
+                } else {
+                    this.region = "und-" + region;
+                }
+                this.locale = new ULocale(this.region);
```
sffc, line 659:
```suggestion
                this.region = region;
                this.locale = new ULocale("und-" + region);
```

sffc (dismissed):
Some funny stuff going on with regions and locales

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
markusicu, line 800: I still don't see how this works.
- If region == "default", you do `new ULocale("default")`. That's bogus.
- Did you mean to check if usage == "default"? If so, how should that inform how to construct the testCaseLocale?
- If region starts with a digit, as in "001", you do `new ULocale("001")`. That's also bogus.
- If region starts with a letter (and it's not "default"), then you do `new ULocale("und-GB")` which is good.
- Now that all of the region inputs are back to really being region codes, I think you need to prepend "und-" whether there are digits or not.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
markusicu, line 659: Same problems as in testGetPreferencesFor(), see my comment there (below).

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitPreferences.java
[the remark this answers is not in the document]
markusicu: You are right, sorry, I overlooked that. Never mind!

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -948,7 +948,12 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
         return;
     }
 
-    UnitsRouter router(inputMeasureUnit, region, usage, status);
+    CharString localeID;
+    if (uprv_isASCIILetter(region.data()[0]) && region.compare("default") != 0)
+        localeID.append("und-", status);
+    localeID.append(region, status);
+    Locale locale(localeID.data());
```
markusicu, line 955: Same problems as in the Java test. See my comments there.

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -948,7 +948,12 @@ void unitPreferencesTestDataLineFn(void *context, char *fields[][2], int32_t fie
         return;
     }
 
-    UnitsRouter router(inputMeasureUnit, region, usage, status);
+    CharString localeID;
+    if (uprv_isASCIILetter(region.data()[0]) && region.compare("default") != 0)
+        localeID.append("und-", status);
```
markusicu, line 953: And please remember to use curly braces.

icu4c/source/test/intltest/units_data_test.cpp
```diff
@@ -140,18 +141,22 @@ void UnitsDataTest::testGetPreferencesFor() {
...
         logln(t.name);
-        const UnitPreference *const *prefs;
-        int32_t prefsCount;
-        preferences.getPreferencesFor(t.category, t.usage, t.region, prefs, prefsCount, status);
+        CharString localeID;
+        if (uprv_isASCIILetter(t.region[0]) && uprv_strcmp(t.region, "default") != 0) {
+            localeID.append("und-", status);
+        }
+        localeID.append(t.region, status);
+        Locale locale(localeID.data());
```
markusicu, line 149: Same problems as in the Java test. See my comments there.

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +389,96 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+                localeUnitCharString.toStringPiece().compare("fahrenheit") == 0 ||
+                localeUnitCharString.toStringPiece().compare("kelvin") == 0) {
```
markusicu, line 419: Please indent the condition-continuation lines by another 4 spaces, so that one can distinguish the if head from the if body.

icu4j/main/classes/core/src/com/ibm/icu/impl/units/UnitsRouter.java
markusicu: This code is slightly different now but still doesn't work. I am resolving this comment in favor of newer ones.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
markusicu: This code is slightly different now but still doesn't work. I am resolving this comment in favor of newer ones.

icu4c/source/test/intltest/units_test.cpp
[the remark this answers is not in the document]
markusicu: This code is somewhat different now but still doesn't work. I am resolving this comment in favor of newer ones.

icu4c/source/test/intltest/units_test.cpp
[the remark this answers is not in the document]
markusicu: This code is somewhat different now but still doesn't work. I am resolving this comment in favor of newer ones.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/impl/UnitsTest.java
markusicu: This code is somewhat different now but still doesn't work. I am resolving this comment in favor of newer ones.

icu4c/source/i18n/units_router.cpp
```diff
@@ -43,17 +43,18 @@ Precision UnitsRouter::parseSkeletonToPrecision(icu::UnicodeString precisionSkel
     return result;
 }
 
-UnitsRouter::UnitsRouter(StringPiece inputUnitIdentifier, StringPiece region, StringPiece usage,
+UnitsRouter::UnitsRouter(StringPiece inputUnitIdentifier, StringPiece localeString, StringPiece usage,
                          UErrorCode &status) {
-    this->init(MeasureUnit::forIdentifier(inputUnitIdentifier, status), region, usage, status);
+    Locale locale(localeString.data());
```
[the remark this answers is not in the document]
markusicu: Ping: Please check if this is safe.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: For the test cases:
-- 2 alternates for the input measure: Fahrenheit and celsius
**I have added them**

-- 3 alternates for usage: none, "default", and "other".
**I have used `none` and `default`. `Other` will give the same result as default, I did not tested it.**

-- 2 alternates for base locale: en, en_AU, and de
**I have used en_US for the test cases, shall I test other locales with different languages?  it will be robust, but I think the main outcome of the test cases that we need is to test the effect of the tags.**

icu4c/source/test/intltest/units_data_test.cpp
younies: done

icu4c/source/i18n/units_data.cpp
markusicu: Not my preferred style (I prefer Google's extra indentation, and keeping the `) {` at the end of the last condition line, but this is readable, so ok.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
markusicu: @macchiati ok?

icu4c/source/i18n/units_router.cpp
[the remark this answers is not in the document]
younies: Now, we pass the locale, so, no locale as string anymore.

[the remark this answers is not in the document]
markusicu: Great, thanks for checking! (GitHub doesn't show this comment in context any more...)

markusicu (dismissed):
lgtm tnx!
If Mark is happy with the set of test cases, then it's all good I think.

macchiati (approved):

icu4j/main/tests/core/src/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
markusicu: Mark says test cases are ok now.

sffc (dismissed):

sffc (approved):

icu4c/source/i18n/units_data.cpp
```diff
@@ -387,24 +389,97 @@ U_I18N_API UnitPreferences::UnitPreferences(UErrorCode &status) {
...
+    MaybeStackVector<UnitPreference> result;
+
+    // TODO: remove this once all the categories are allowed.
+    UErrorCode internalMuStatus = U_ZERO_ERROR;
+    if (category.compare("temperature") == 0) {
+        CharString localeUnitCharString = getKeyWordValue(locale, "mu", internalMuStatus);
+        if (U_SUCCESS(internalMuStatus)) {
+            // TODO: use the unit category as Java especially when all the categories are allowed..
+            if (localeUnitCharString == "celsius"       //
+                || localeUnitCharString == "fahrenheit" //
```
sffc, line 418: Should this be `fahrenhe` instead of `fahrenheit`?

CC @mihnita

mihnita: +10

Yes, I think it should `fahrenhe`, everywhere.

 So i would add a unit test like this:

```
String defaultTempUnit = NumberFormatter
        .withLocale(Locale.US)
        .usage("weather")
        .unit(MeasureUnit.CELSIUS)
        .format(1)
        .getOutputUnit()
        .getIdentifier();
assertEquals("fahrenhe", defaultTempUnit);
```

It currently returns `fahrenheit`, and I think that is incorrect.

PR #2187, opened 2022-09-16
ICU-20512 Add extra matchers to handle empty currency symbols

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20512
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2188, opened 2022-09-16
ICU-20512 ICU4J: just add test of parse with empty curr symbol, code already works as desired

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20512

In ICU4J, just add a test of parsing with empty currency symbol (corresponds to ICU4C test in #2187), code already behaves as desired per ICU-20512

### DIFF
--- a/icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
+++ b/icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
@@ -6966,4 +6966,94 @@ public class NumberFormatTest extends TestFmwk {
             assertEquals("Via applyPattern: field position end", 3, fp.getEndIndex());
         }
     }
+
+    @Test
+    public void TestParseWithEmptyCurr() {
+        DecimalFormat df = (DecimalFormat)NumberFormat.getInstance(new ULocale("en_US"), NumberFormat.CURRENCYSTYLE);
+        DecimalFormatSymbols dfs = df.getDecimalFormatSymbols();
+        dfs.setCurrencySymbol("");
+        df.setDecimalFormatSymbols(dfs);
+ 
+        String textToParse = "3";    
+        ParsePosition ppos = new ParsePosition(0);
+
+        Number num = df.parse(textToParse, ppos);
+        if (ppos.getIndex() != 1 || num.doubleValue() != 3.0) {
+            errln("parse with empty curr symbol, expect pos 1, value 3.0; get "
+                                + ppos.getIndex() + ", " + num.doubleValue());
+        }
+
+        ppos.setIndex(0);
+        CurrencyAmount currAmt = df.parseCurrency(textToParse, ppos);
+        if (currAmt != null) {
+            errln("parseCurrency with empty curr symbol, expect null but succeeds");
+        }
+
+        //                        "¤#,##0.00" "¤ #,##0.00" "#,##0.00 ¤" "#,##,##0.00¤"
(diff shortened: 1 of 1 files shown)


icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
```diff
@@ -6966,4 +6966,94 @@ public void Test21556_CurrencyAsDecimal() {
...
+        if (ppos.getIndex() != 1 || num.doubleValue() != 3.0) {
+            errln("parse with empty curr symbol, expect pos 1, value 3.0; get "
+                                + ppos.getIndex() + ", " + num.doubleValue());
+        }
+
+        ppos.setIndex(0);
+        CurrencyAmount currAmt = df.parseCurrency(textToParse, ppos);
+        if (currAmt != null) {
+            errln("parseCurrency with empty curr symbol, expect null but succeeds");
+        }
```
richgillam, line 6990: This kind of threw me.  Intuitively, it seems like if the currency symbol is the empty string and you parse a string that doesn't have a currency symbol, it should succeed-- the "currency symbol" in the string matches the "currency symbol" in the formatter.  Why is this not the expected behavior?

IIRC, we see this in practice in a couple of locales that have an infix currency symbol, such as the Cape Verdean escudo.  The "currency symbol" is blank, and the actual currency symbol is the currency decimal separator.  In that case, you'd get a parse error if the currency decimal separator wasn't the expected one.  Is this related to that kind of use case?

macchiati (approved):

richgillam (approved):
This looks like it does what the ticket and discussion say it should do.  I gather the issue here was that parse() was doing the wrong thing, and this looks like it tests that well, but I'm a little puzzled by the idea that parseCurrency() should fail in this case, even though that's apparently what it has always done.  But I'm coming to this cold, so I'm going to assume you guys all know what you're doing.  But would you mind filling in the background for me?

sffc (commented):
I'm surprised that Java passes this test; the C++ and Java behavior are supposed to be identical. It may be worth investigating further why Java works this way but not C++.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
pedberg-icu: @richgillam parseCurrency is intended to parse an arbitrary currency, not the specific currency of the locale associated with the formatter. So if there is no currency symbol itcnnot report which currency it parsed. The standard parse is only intended to parse a specific currency, either the one for the locale or some other symbol that has been specifically set (which might be empty).

pedberg-icu:
> I'm surprised that Java passes this test; the C++ and Java behavior are supposed to be identical. It may be worth investigating further why Java works this way but not C++.

I will file a separate ticket for that. Meanwhile this test will catch any regressions in that behavior.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatTest.java
richgillam: Oh, okay.  That makes total sense then.  Thanks for the explanation.

pedberg-icu:
> > I'm surprised that Java passes this test; the C++ and Java behavior are supposed to be identical. It may be worth investigating further why Java works this way but not C++.
> 
> I will file a separate ticket for that. Meanwhile this test will catch any regressions in that behavior.

@sffc : https://unicode-org.atlassian.net/browse/ICU-22161

PR #2196, opened 2022-09-21
ICU-21957 BRS72 J API signature file

yumaoka:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21957


icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2DataModel.java
```diff
@@ -664,6 +664,17 @@ public Variable build() {
...
         private static final long serialVersionUID = -7049361727790825496L;
+
+        /**
+         * {@inheritDoc}
+         *
+         * @internal ICU 72 technology preview
+         * @deprecated This API is for ICU internal use only.
+         */
+        @Deprecated
+        public OrderedMap() {
```
markusicu, line 675: FYI Mihai also added this in https://github.com/unicode-org/icu/pull/2194
Looks like the change is exactly the same, so there is no conflict :-)

icu4j/main/classes/core/src/com/ibm/icu/text/PersonName.java
```diff
@@ -92,17 +101,21 @@ private NameField(String name) {
 
         /**
          * Returns the NameField's display name.
-         * @internal
+         * @internal ICU 72 technology preview
```
markusicu, line 104: @richgillam @macchiati Should the PersonName* APIs that only have `@internal` remain like that (i.e., never intended to become public), or should they be "tech preview" like Yoshito is tagging them here?

icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java
```diff
@@ -286,7 +311,9 @@ private PersonNameFormatter(Locale locale, Length length, Usage usage, Formality
 
     /**
      * @internal For unit testing only!
+     * @deprecated This API is for technology preview only.
```
markusicu, line 314:
```suggestion
     * @deprecated This API is for unit testing only.
```

icu4j/main/classes/core/src/com/ibm/icu/text/PersonName.java
richgillam: If I had something as just "@internal" without the "tech preview" verbiage, it's because I intended it to really be internal (i.e., never to be part of the public API), not just internal because it was part of a tech preview.  In other words, don't change this.

icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java
richgillam: Same issue here.  This really is just for unit testing, never to be public.  I support Markus's proposed change.

icu4j/main/classes/core/src/com/ibm/icu/text/PersonName.java
yumaoka: @richgillam This specific one is for toString(). I think it's not a good idea to make toString() @internal if a class becomes a public ICU API.

icu4j/main/classes/core/src/com/ibm/icu/message2/Formatter.java
```diff
@@ -10,7 +10,7 @@
  * that can be used from {@link MessageFormatter}.
  *
  * @internal ICU 72 technology preview
- * @deprecated This API is for ICU internal use only.
+ * @deprecated This API is for technology preview only.
```
yumaoka, line 13: Changed "This API is for ICU internal use only." to "This API is for technology preview only." for technology preview APIs.

icu4j/main/classes/core/src/com/ibm/icu/message2/Mf2Serializer.java
```diff
@@ -117,52 +117,27 @@ public String toString() {
 
     Mf2Serializer() {}
 
-    /**
```
yumaoka, line 120: @internal/@deprecated tags are not required in package private classes. I deleted these unnecessary tags.

icu4j/main/classes/core/src/com/ibm/icu/text/PersonName.java
```diff
@@ -102,7 +113,9 @@ public String toString() {
         /**
          * Returns the appropriate NameField for its display name.
          * @internal
+         * @deprecated This API is for ICU internal use only.
```
yumaoka, line 116: I kept PersonName#forString(String) as regular ICU internal use API.

icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java
yumaoka: Done

icu4j/main/classes/core/src/com/ibm/icu/message2/Formatter.java
markusicu: Thanks! And sorry that I missed those :-(
FYI  @mihnita

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
```diff
@@ -15,7 +15,7 @@
  * Expect it to change.
  *
  * @internal ICU 72 technology preview. Visible For Testing.
- * @deprecated This API is for ICU internal use only.
+ * @deprecated This API is for technology preview only.
```
markusicu, line 18: So... this class is only public to be "Visible For Testing". It's probably better if it doesn't say "technology preview" at all. @mihnita

yumaoka: PlainStringFormatterValue is a class implementing com.ibm.icu.text.FormattedValue. This is the concrete class used by the message formatter2. Although we have a design option not to expose such class as public API for ICU users, it implies our user cannot create an equivalent object to the output of the formatter.

If we really want to hide this, we don't even need to say "technology preview". If our intent is only for ICU internal testing, I'd drop the term.

I think "Visible For Testing" is good for user experimenting the technology preview MessgeFormatter. So I prefer to keep "technology preview only".

markusicu: I don't want to block the release candidate, but we are not planning for this to eventually become public.
Ok for rc. I am asking @mihnita to change this class back to truly `@internal` right after rc.

I think these should say
```
 * @internal Visible For Testing.
 * @deprecated This API is for unit testing only.
 *
```

markusicu (approved):
lgtm

icu4j/main/classes/core/src/com/ibm/icu/message2/PlainStringFormattedValue.java
yumaoka: OK. It's still @internal and deprecated comment should not affect signature data file.

PR #2199, opened 2022-09-22
ICU-10752 Spread (s|g)etRelativeYear to subclass

FrankYFTang:
Remove the switch statment implementaiton in
Calendar::(g|s)etRelatedYear
and move the code into each subclass as proper OOP style.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-10752

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -1267,130 +1267,14 @@ Calendar::set(int32_t year, int32_t month, int32_t date, int32_t hour, int32_t m
 }
 
 // -------------------------------------
-// For now the full getRelatedYear implementation is here;
-// per #10752 move the non-default implementation to subclasses
-// (default implementation will do no year adjustment)
-
-static int32_t gregoYearFromIslamicStart(int32_t year) {
-    // ad hoc conversion, improve under #10752
-    // rough est for now, ok for grego 1846-2138,
-    // otherwise occasionally wrong (for 3% of years)
-    int cycle, offset, shift = 0;
-    if (year >= 1397) {
-        cycle = (year - 1397) / 67;
-        offset = (year - 1397) % 67;
-        shift = 2*cycle + ((offset >= 33)? 1: 0);
-    } else {
-        cycle = (year - 1396) / 67 - 1;
-        offset = -(year - 1396) % 67;
-        shift = 2*cycle + ((offset <= 33)? 1: 0);
-    }
-    return year + 579 - shift;
-}
-
 int32_t Calendar::getRelatedYear(UErrorCode &status) const
 {
-    if (U_FAILURE(status)) {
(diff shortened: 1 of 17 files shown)


FrankYFTang:
I don't think we need to rush for 72, we can wait till 73.

pedberg-icu:
> I don't think we need to rush for 72, we can wait till 73.

Well, this *must* wait until 73, since it is after the code freeze. So it cannot be merged until after the maintenance branch is forked for 72. I will wait to review until then.

richgillam (dismissed):
This looks a lot better than what was there before, and I'm happy to see it go in, but...

Most of the implementation of getRelatedYear() and setRelatedYear() appear to be the same, and the ones that are a little more complicated are more complicated because they're doing extra work to calculate the proper offset.  Would it make sense to have getRelatedYear() and setRelatedYear() still be implemented only at the Calendar level, but call through to a separate virtual getRelatedYearOffset() that just calculates (or returns) the offset value?

In fact, it might even make sense to store the related year offset in a private member variable at the Calendar level and have each subclass calculate the right value for that variable in their constructors.  Then you wouldn't need the virtual function call here at all.  (Of course, that comes at the expense of making the Calendar object bigger and requiring implementers to know they have to calculate this value at construction time, so maybe I'm wrong...)

At any rate, it seems like either of these suggestions would be cleaner, but what you have in this PR is already considerably cleaner than the original code.

pedberg-icu (approved):

richgillam (approved):
Again, this looks good, but do you need to override getRelatedYear() in all the Calendar subclasses?  It seems like you could have a single getRelatedYear() at the Calendar level that looks the same as all your overrides, but calls out to a separate internal getRelatedYearOffset() method that you DO override at each level, and which supplies the offset for that particular calendar.  Or are these existing constraints that prevent that?

FrankYFTang:
> Or are these existing constraints that prevent that?

Introducing a getRelatedYearOffset() function means I will need to make an API change, since that will mandate all the subclassed to add a new method, right? (consider those subclass of Clander which is not shipped with ICU.

richgillam:
>Introducing a getRelatedYearOffset() function means I will need to make an API change, since that will mandate all the subclassed to add a new method, right?

I didn't really think about that.  But no, not necessarily.  I'd think you could introduce that new method as an internal method that we use without opening it up to anyone else.  You couldn't make it a pure virtual method, of course-- you'd have to have an implementation at the top level that either did something trivial (like return 0) or threw an exception.

Of course, if the `geRelatedYear()` method is currently declared as pure virtual at the top level of Calendar, changing it to have an implementation might have negative API-compatibility effects; I'm not completely sure about this.

But you merged, so it's all moot.  I don't think this is worth filing a new ticket over or asking you to do more work.

PR #2216, opened 2022-10-03
does this lib support vs2022?

dangdkhanh:
I need to compile on visual studio 2022 however I get the following error
image

I'm trying to fix it but first i want to ask if it will be affected if the compile is successful or not?


markusicu:
I don't see an error message.
We can confirm that ICU works on Visual Studio 2022.

PR #2231, opened 2022-10-13
ICU-21877 Fixed it so that getAllChildrenWithFallback() correctly calls its sink with every possible...

richgillam:
...resource the one the user requested might be inheriting elements from.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21877

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -2351,7 +2351,66 @@ struct GetAllChildrenSink : public ResourceSink {
                     aliasedValue.setData(aliasRB->getResData());
                     aliasedValue.setValidLocaleDataEntry(aliasRB->fValidLocaleDataEntry);
                     aliasedValue.setResource(aliasRB->fRes, ResourceTracer(aliasRB));
-                    dest.put(key, aliasedValue, isRoot, errorCode);
+                    
+                    if (aliasedValue.getType() != URES_TABLE) {
+                        dest.put(key, aliasedValue, isRoot, errorCode);
+                    } else {
+                        // if the resource we're aliasing over to is a table, the sink might iterate over its contents.
+                        // If it does, it'll get only the things defined in the actual alias target, not the things
+                        // the target inherits from its parent resources.  So we walk the parent chain for the *alias target*,
+                        // calling dest.put() for each of the parent tables we could be inheriting from.  This means
+                        // that dest.put() has to iterate over the children of multiple tables to get all of the inherited
+                        // resource values, but it already has to do that to handle normal vertical inheritance.
+                        UResType aliasedValueType = URES_TABLE;
+                        CharString tablePath;
+                        tablePath.append(aliasRB->fResPath, errorCode);
+                        const char* parentKey = key; // dest.put() changes the key
+                        dest.put(parentKey, aliasedValue, isRoot, errorCode);
+                        UResourceDataEntry* entry = aliasRB->fData;
+                        Resource res = aliasRB->fRes;
+                        while (aliasedValueType == URES_TABLE && entry->fParent != nullptr) {
+                            CharString localPath;
+                            localPath.copyFrom(tablePath, errorCode);
+                            char* localPathAsCharPtr = localPath.data();
+                            const char* childKey;
+                            entry = entry->fParent;
+                            res = entry->fData.rootRes;
+                            Resource newRes = res_findResource(&entry->fData, res, &localPathAsCharPtr, &childKey);
+                            if (newRes != RES_BOGUS) {
+                                aliasedValue.setData(entry->fData);
+                                // TODO: do I also need to call aliasedValue.setValueLocaleDataEntry() ?
+                                aliasedValue.setResource(newRes, ResourceTracer(aliasRB)); // probably wrong to use aliasRB here
+                                aliasedValueType = aliasedValue.getType();
+                                if (aliasedValueType == URES_ALIAS) {
+                                    // in a few rare cases, when we get to the root resource bundle, the resource in question
+                                    // won't be an actual table, but will instead be an alias to a table.  That is, we have
+                                    // two aliases in the inheritance path.  (For some locales, such as Zulu, we see this with
+                                    // children of the "fields" resource: "day-narrow" aliases to "day-short", which aliases
+                                    // to "day".)  When this happens, we need to make sure we follow all the aliases.
+                                    ResourceDataValue& rdv2 = static_cast<ResourceDataValue&>(aliasedValue);
+                                    aliasRB = getAliasTargetAsResourceBundle(rdv2.getData(), rdv2.getResource(), nullptr, -1,
+                                                                             rdv2.getValidLocaleDataEntry(), nullptr, 0,
+                                                                             stackTempBundle.getAlias(), &errorCode);
+                                    tablePath.clear();
+                                    tablePath.append(aliasRB->fResPath, errorCode);
+                                    entry = aliasRB->fData;
+                                    res = aliasRB->fRes;
+                                    aliasedValue.setData(entry->fData);
+                                    // TODO: do I also need to call aliasedValue.setValueLocaleDataEntry() ?
+                                    aliasedValue.setResource(res, ResourceTracer(aliasRB)); // probably wrong to use aliasRB here
+                                    aliasedValueType = aliasedValue.getType();
+                                }
+                                if (aliasedValueType == URES_TABLE) {
+                                    dest.put(parentKey, aliasedValue, isRoot, errorCode);
+                                } else {
+                                    // once we've followed the alias, the resource we're looking at really should
+                                    // be a table
+                                    errorCode = U_INTERNAL_PROGRAM_ERROR;
+                                    return;
+                                }
+                            }
+                        }
+                    }
                 }
             } else {
                 dest.put(key, value, isRoot, errorCode);
--- a/icu4c/source/test/intltest/restest.cpp
+++ b/icu4c/source/test/intltest/restest.cpp
@@ -15,10 +15,17 @@
 #include "unicode/resbund.h"
 #include "restest.h"
 
+#include "uresimp.h"
+#include "ureslocs.h"
+
 #include <stdlib.h>
 #include <time.h>
 #include <string.h>
 #include <limits.h>
+
+#include <map>
+#include <set>
+#include <string>
 
 //***************************************************************************************
 
@@ -219,6 +226,8 @@ void ResourceBundleTest::runIndexedTest( int32_t index, UBool exec, const char* 
 #endif
 
     case 4: name = "TestExemplar"; if (exec) TestExemplar(); break;
+    case 5: name = "TestPersonUnits"; if (exec) TestPersonUnits(); break;
+    case 6: name = "TestZuluFields"; if (exec) TestZuluFields(); break;
         default: name = ""; break; //needed to end loop
     }
 }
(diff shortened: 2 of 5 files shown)


icu4c/source/common/uresbund.cpp
```diff
@@ -2350,7 +2350,84 @@ struct GetAllChildrenSink : public ResourceSink {
...
+////                                aliasedResource = ICUResourceBundle.this.findWithFallback(tablePath);
+////                            }
+//                            aliasedResourceImpl = (ICUResourceBundleImpl) aliasedResource;
+//                            aliasedValue = new ReaderValue();
+//                            aliasedValue.reader = aliasedResourceImpl.wholeBundle.reader;
+//                            aliasedValue.res = aliasedResourceImpl.getResource();
+//                            aliasedValueType = aliasedValue.getType(); // sink.put() messes up the value
+//                            keyCopy = key.clone(); // sink.put() messes up the key
+//                            sink.put(keyCopy, aliasedValue, noFallback);
+//                        }
```
richgillam, line 2429: I'm sorry.  I'll take out all this crap in a future commit after I've received review feedback.

richgillam:
Some explanation: This addresses the last (according to my testing) horizontal-inheritance problem in ICU, which is that if you use `getAllChildrenWithFallback()` to iterate over all the children of a resource and the things you're iterating over are _tables_, the tables that get passed to the sink are incomplete if `getAllChildrenWithFallback()` followed an alias.  If you follow an alias, you get the contents of the exact table the alias pointed to, but not anything it's inheriting from its parent resource bundle.  This means you can get different results by going straight to a particular element of the table with `getByKeyWithFallback()` than you get with `getAllChildrenWithFallback()`.

At first I thought I was going to have to fix this by either fabricating a table to pass to the sink that actually contained all the inherited elements, or by adding machinery to the table class that would fetch inherited items from parent bundles (machinery that doesn't exist now for good reason).  What I ended up doing was simpler: If `getAllChildrenWithFallback()` follows an alias to a table resource, not only does it call the sink's `put()` method with that table, but it ascends the parent chain, calling the sink's `put()` method with the corresponding resource in every bundle in the containing bundle's parent chain (following _more_ aliases when necessary).  This produces a patten of calls to the sink that looks a little weird, but upholds the sink's expectations for how things should be passed to it.

This wasn't too hard to get up and running in Java, but the port to C++ was complicated.  The code I have works, but probably includes some logic that doesn't actually need to be there, or which can be done in a simpler way.  As always, I invite suggestions for making this code better.

richgillam:
Two additional comments:

1) In addition to the unit tests included in this PR, I've written a comprehensive horizontal inheritance test that checks to make sure`getByKeyWithFallback()` returns the correct values for everything, and also spot-checks that `getAllChildrenWithFallback()` returns the same thing as `getByKeyWithFallback()`.  It's in three pieces: a "resource tree flattener", written in Java, that converts the data in the ICU4C data source files into a text files that just contain key paths and resource values, and two "resource tree verifiers", in C++ and Java, that iterate over the text files created by the flattener, calling ICU with the various paths, and comparing the results to what the text files contain.  I don't know where to check this in or what I need to do to make it useful to anyone but me.  I'd love some feedback here.

2) Some sinks have a "`fillInMissing()` method to work around the problem I'm trying to fix here, and lots of places in the code are still calling `getAllItemsWithFallback()`, which has lots of problems with aliases, instead of the new `getAllChildrenWithFallback()`.  I want to fix all this, but it seems prudent to do that on another ticket after these changes land.

macchiati:
Is this still active?

richgillam:
>Is this still active?

Yes!  I'm hoping Markus will find time to review it for ICU 74.

markusicu:
> 1. In addition to the unit tests included in this PR, I've written a comprehensive horizontal inheritance test that checks to make sure`getByKeyWithFallback()` returns the correct values for everything, and also spot-checks that `getAllChildrenWithFallback()` returns the same thing as `getByKeyWithFallback()`.  It's in three pieces: a "resource tree flattener", written in Java, that converts the data in the ICU4C data source files into a text files that just contain key paths and resource values, and two "resource tree verifiers", in C++ and Java, that iterate over the text files created by the flattener, calling ICU with the various paths, and comparing the results to what the text files contain.  I don't know where to check this in or what I need to do to make it useful to anyone but me.  I'd love some feedback here.

This sounds very useful!

We have test and tools sub-trees and both icu4c & icu4j, and we also have a top-levels tools tree. I am sure we can figure out places for these to go.

Running such tests is probably slow? We could do it post-merge, and/or once a week etc. as needed, like "exhaustive" tests and similar.

markusicu:
Please rebase & resolve conflicts.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/util/ICUResourceBundleTest.java
```diff
@@ -1179,4 +1174,101 @@ public void TestAlgorithmicParentFallback() {
             assertEquals("Got wrong locale with LOCALE_ROOT", localeRootExpected, localeRootActual);
         }
     }
+
+    @Test
+    public void TestPersonUnits() {
```
markusicu, line 1179: Would this test actually fail with the current code & data?
I assume that either the current data has "redundant" values to work around these problems, or the current PersonNameFormatter code has workarounds so that it gets complete data.

I am just worried that the problem is masked by workarounds, and the test may not find the problems that are described below.

Would it be better to reconstruct this problem in testdata, independent of what the real locale data looks like? (Could be in a follow-up PR.)

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -459,7 +459,43 @@ public void put(UResource.Key key, UResource.Value value, boolean noFallback) {
...
+                                    tablePath = tablePath.substring(0, tablePath.lastIndexOf('/'));
+                                    tablePath += "/" + newAliasedResource.key;
+                                    aliasedResource = ICUResourceBundle.this.findWithFallback(tablePath);
+                                }
+                                aliasedResourceImpl = (ICUResourceBundleImpl) aliasedResource;
+                                aliasedValue = new ReaderValue();
+                                aliasedValue.reader = aliasedResourceImpl.wholeBundle.reader;
+                                aliasedValue.res = aliasedResourceImpl.getResource();
+                                aliasedValueType = aliasedValue.getType(); // sink.put() messes up the value
+                                keyCopy = key.clone(); // sink.put() messes up the key
```
markusicu, line 495: Please add a comment like
```
// TODO: keyCopy.copyFrom(key); -- to reduce object creations a little
```

icu4c/source/test/intltest/restest.cpp
```diff
@@ -655,3 +664,160 @@ ResourceBundleTest::TestGetLocaleByType(void)
         }
     }
 }
+
+void
+ResourceBundleTest::TestPersonUnits(void) {
```
markusicu, line 669:
```suggestion
ResourceBundleTest::TestPersonUnits() {
```

icu4c/source/test/intltest/restest.h
```diff
@@ -34,6 +34,9 @@ class ResourceBundleTest: public IntlTest {
 
     void TestGetSize(void);
     void TestGetLocaleByType(void);
+    
+    void TestPersonUnits(void);
+    void TestZuluFields(void);
```
markusicu, line 39:
```suggestion
    void TestPersonUnits();
    void TestZuluFields();
```

icu4c/source/test/intltest/restest.cpp
```diff
@@ -655,3 +664,160 @@ ResourceBundleTest::TestGetLocaleByType(void)
...
+
+    // getAllChildrenWithFallback() wasn't bringing all of those things back, however.  When en_CA/units/year-person
+    // aliased over to en_CA/unitsShort/year, the sink would only be called on en_CA/unitsShort/year, which means it'd
+    // only see en_CA/unitsShort/year/per, because "per" was the only thing contained in en_CA/unitsShort/year.  But
+    // en_CA/unitsShort/year should be inheriting "dnam", "one", and "other" from en/unitsShort/year.
+    // getAllChildrenWithFallback() had to be modified to walk the parent chain and call the sink again with
+    // en/unitsShort/year and root/unitsShort/year.
+    struct TestPersonUnitsSink : public ResourceSink {
+        typedef std::set<std::string> FoundKeysType;
+        typedef std::map<std::string, FoundKeysType > FoundResourcesType;
```
markusicu, line 706: totally optional: this smells like you might have been looking for [std::multimap](https://en.cppreference.com/w/cpp/container/multimap) ...

icu4c/source/test/intltest/restest.cpp
```diff
@@ -655,3 +664,160 @@ ResourceBundleTest::TestGetLocaleByType(void)
...
+        FoundResourcesType foundResources;
+        IntlTest* owningTestCase;
+        
+        TestPersonUnitsSink(IntlTest* owningTestCase) : owningTestCase(owningTestCase) {}
```
markusicu, line 710: Please use a reference, not a pointer. (Much nicer if it can't be nullptr.)

icu4c/source/test/intltest/restest.cpp
```diff
@@ -655,3 +664,160 @@ ResourceBundleTest::TestGetLocaleByType(void)
...
+                errln("%s doesn't contain 'other'!", foundResourcesKey.c_str());
+            }
+            if (foundKeys.find("per") == foundKeys.end()) {
+                errln("%s doesn't contain 'per'!", foundResourcesKey.c_str());
+            }
+        }
+    }
+}
+
+#include "charstr.h" // DEBUGGING ONLY!!! TAKE OUT!!!
```
markusicu, line 759: DEBUGGING ONLY!!! TAKE OUT!!!

icu4c/source/common/uresbund.cpp
```diff
@@ -2350,7 +2350,61 @@ struct GetAllChildrenSink : public ResourceSink {
...
+                                    ResourceDataValue& rdv2 = static_cast<ResourceDataValue&>(aliasedValue);
+                                    aliasRB = getAliasTargetAsResourceBundle(rdv2.getData(), rdv2.getResource(), nullptr, -1,
+                                                                             rdv2.getValidLocaleDataEntry(), nullptr, 0,
+                                                                             stackTempBundle.getAlias(), &errorCode);
+                                    tablePath.clear();
+                                    tablePath.append(aliasRB->fResPath, errorCode);
+                                    entry = aliasRB->fData;
+                                    res = aliasRB->fRes;
+                                    aliasedValue.setData(entry->fData);
+                                    // do I also need to call aliasedValue.setValueLocaleDataEntry() ?
```
markusicu, line 2398: I don't know... please add TODOs for these questions...

icu4c/source/common/uresbund.cpp
```diff
@@ -2350,7 +2350,61 @@ struct GetAllChildrenSink : public ResourceSink {
...
+                                    aliasedValue.setResource(res, ResourceTracer(aliasRB)); // probably wrong to use aliasRB here
+                                    aliasedValueType = aliasedValue.getType();
+                                }
+                                if (aliasedValueType == URES_TABLE) {
```
markusicu, line 2402: Set an error if not?
You don't check this in Java, leaving it to the sink to check & complain. Why check in C++ but not in Java?

markusicu (commented):
I would really like to spend a couple of days traipsing through the code and thinking about whether we can do this better.

Failing that (and I have clearly been failing), your explanations sound very plausible, and the implementation changes... also look plausible.

Want to try to get this into 73?

icu4c/source/test/intltest/restest.cpp
```diff
@@ -655,3 +664,160 @@ ResourceBundleTest::TestGetLocaleByType(void)
...
+
+void
+ResourceBundleTest::TestZuluFields(void) {
```
markusicu, line 762:
```suggestion
ResourceBundleTest::TestZuluFields() {
```

richgillam:
> Want to try to get this into 73?

I'm buried under crises in my day job right now, so I don't think I'm going to be able to turn this around in the next couple of weeks.  Let's shoot for (hopefully early in) 74.

Thank you for all the feedback!  I'll try to look at it in detail and figure out what to do sometime in the next few weeks, when I come up for air.

richgillam:
> This sounds very useful!
> 
> We have test and tools sub-trees and both icu4c & icu4j, and we also have a top-levels tools tree. I am sure we can figure out places for these to go.
> 
> Running such tests is probably slow? We could do it post-merge, and/or once a week etc. as needed, like "exhaustive" tests and similar.

I filed https://unicode-org.atlassian.net/browse/ICU-22490 to add this later; I don't want this PR hung up on that, but I agree we should let it completely fall through the cracks.  I'm kind of guessing we'd end up with something like the PersonNameConsistencyTest, where a CLDR build generates the text dumps of the resources, the CLDR-to-ICU tool copies the text dumps over into the ICU tree, and we have tests in ICU that use that (and yes, probably are part of an exhaustive-test run).  But I think that'd also be a fairly radical change from what I have right now, which (if I remember right) actually iterates over the raw text files in icu4c/source/data instead of the CLDR sources.

I think the rules for inheritance on the CLDR side have changed or been refined since I originally did all this work, so I'm a little worried things don't totally line up, but that seems like all the more reason to have this kind of test.  I wonder if I can list somebody to help with the CLDR side of things?

richgillam:
I also filed https://unicode-org.atlassian.net/browse/ICU-22491 for the other loose-end comment above.

richgillam:
> Want to try to get this into 73?

Too late for that now.  😉  But I really _would_ like to get this into 74, if at all possible.

icu4c/source/common/uresbund.cpp
richgillam: (It's been so long that I might misremember some of this, but...)

The code here isn't parallel.  In C, I had to add extra code to deal with the situation where the alias pointed to another alias; the Java code handled that situation for me, so I was sure I was coming back from `findWithFallback()` with something other than an alias.

I think in C, the `if (URES_TABLE)` was there mostly to be parallel to the `if (URES_ALIAS)` at the top of the clause above, more than that I wanted to really make sure it was a table.  On the other hand, I think I may have also had it there to guard against hitting a _third_ alias.  (I thought about making the if above a while loop, but it seemed like overkill, and I think it might have been a lot more complicated, so I let it go.). I'll change this to signal an error if the alias target isn't a table.

icu4c/source/test/intltest/restest.cpp
richgillam: I don't think `std::multimap` does quite what I'm trying to do here.  What I was after was something that would tell me that, for example, `units/duration` (the key in `FoundResourcesType`) had child resources named `one` and `other`, but not `per` or `dnam`.  So I'm mapping from a single string (e.g., `units/duration`) to a collection of strings (`one`, `many`).

richgillam:
[**NOTE:** The above commit rebases, squashes, and addresses all of Markus's comments.]

icu4c/source/common/uresbund.cpp
```diff
@@ -2351,7 +2351,66 @@ struct GetAllChildrenSink : public ResourceSink {
...
+                                aliasedValue.setData(entry->fData);
+                                // TODO: do I also need to call aliasedValue.setValueLocaleDataEntry() ?
+                                aliasedValue.setResource(newRes, ResourceTracer(aliasRB)); // probably wrong to use aliasRB here
+                                aliasedValueType = aliasedValue.getType();
+                                if (aliasedValueType == URES_ALIAS) {
+                                    // in a few rare cases, when we get to the root resource bundle, the resource in question
+                                    // won't be an actual table, but will instead be an alias to a table.  That is, we have
+                                    // two aliases in the inheritance path.  (For some locales, such as Zulu, we see this with
+                                    // children of the "fields" resource: "day-narrow" aliases to "day-short", which aliases
+                                    // to "day".)  When this happens, we need to make sure we follow all the aliases.
```
pedberg-icu, line 2389: ... And there is likely a lot more of this sort of multiple aliasing in CLDR 44, so this is good.

pedberg-icu (approved):

macchiati:
> I wonder if I can list somebody to help with the CLDR side of things?

Rich, I can help with that this week. Chat on Monday?

richgillam:
> Rich, I can help with that this week. Chat on Monday?

Sure.  Send me an invite.

richgillam:
@markusicu , are you okay with this going in?

icu4c/source/test/intltest/restest.cpp
markusicu: I think that's what a multimap does -- but, as I wrote earlier, "totally optional".

markusicu (approved):
let's get it in!

richgillam:
> let's get it in!

Awesome!  Huge thanks!

PR #2236, opened 2022-10-20
ICU-22190 Add public PGP Key

rp9-next:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22190


markusicu (dismissed):
rslgtm

@echeran @roubert @srl295 could one of you please verify that this works with the ICU 72 Windows binaries?

roubert:
> @echeran @roubert @srl295 could one of you please verify that this works with the ICU 72 Windows binaries?

I've now verified that this public key indeed belongs to the key used to sign the ICU 72 Windows binaries. But I don't think it's a particularly good idea to add an unsigned public key to the ICU `KEYS` file.

rp9-next:
@roubert, do the keys change after signing ?

roubert:
> @roubert, do the keys change after signing ?

The key itself won't change, but the key block here in the `KEYS` file includes both key and signatures, so that'll change.

markusicu:
@roubert can you please help @rp9-next here with whatever needs to be done? And then help @sven-oly in a hopefully-following-soon PR?

roubert:
> @roubert can you please help @rp9-next here with whatever needs to be done? And then help @sven-oly in a hopefully-following-soon PR?

Done.

roubert (approved):
LGTM, key block with signature verified.

rp9-next:
Thanks @roubert for helping with generating the signed key block.
I have updated the KEYS to include a signed key block (instead of an unsigned one).
@sven-oly, apologize for partial instructions, but you might also need to use a signed block from @roubert!
(On the plus side, it involves having a nice conversation with him)

roubert (approved):

KEYS
```diff
@@ -960,3 +960,34 @@ uVrQ13BH7D1TWlzM7jN1dToIySXIjGRcMpqT/2LqYDwHfrJCdQZGKDBl5vNWXwNs
 OC7PFtJwtXNl8p7Jvk3lplU5rNSSkTF8hSFgtxu4cpJ5VAY8nQ==
 =FdGa
 -----END PGP PUBLIC KEY BLOCK-----
+
+pub   ed25519 2022-10-19 [SC] [expires: 2024-10-19]
```
markusicu, line 964: Why does the key expire? Will this mean that people can't verify ICU 72 binaries in two years?

markusicu (approved):
rslgtm

KEYS
roubert: It'll always remain possible to verify the signature, it'll then just say that the key has since expired.

PR #2239, opened 2022-10-21
ICU-22191 writesrc.cpp: fix cinttypes header

barracuda156:
Fixes build failure on macOS with GCC.
`<cinttypes>` should be used instead of `<inttypes.h>`: https://en.cppreference.com/w/cpp/header/cinttypes

P. S. Using <inttypes.h> requires defining `__STDC_FORMAT_MACROS`: https://opensource.apple.com/source/gcc/gcc-926.1/inttypes.h.auto.html – but it is not defined, which results in build failure.
Alternative would be adding something like:
```
#if defined __APPLE__
#define __STDC_FORMAT_MACROS
#endif
```
before problematic header.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22191

### DIFF
--- a/icu4c/source/tools/toolutil/writesrc.cpp
+++ b/icu4c/source/tools/toolutil/writesrc.cpp
@@ -19,8 +19,8 @@
 */
 
 #include <stdio.h>
-#include <inttypes.h>
 #include <time.h>
+#include <cinttypes>
 #include "unicode/utypes.h"
 #include "unicode/putil.h"
 #include "unicode/ucptrie.h"


srl295 (approved):
100%. i actually hit this myself on another project.

richgillam (approved):
It bothers me a little that this one include is now a C++-style include, while all the other ones are still .h-style includes, giving me a slight preference for the `__STDC_FORMAT_MACROS` solution, but it's not a strong enough preference to make a stink over.  It might, however, not be a bad idea to move this include down into a separate group of includes, just to help highlight that it's different from the ones around it.

icu4c/source/tools/toolutil/writesrc.cpp
```diff
@@ -19,7 +19,7 @@
 */
 
 #include <stdio.h>
-#include <inttypes.h>
+#include <cinttypes>
 #include <time.h>
```
markusicu, line 23: optional, to put the C++ standard header after the C standard headers:
```suggestion
#include <time.h>
#include <cinttypes>
```

markusicu (approved):
lgtm tnx

markusicu:
> It might, however, not be a bad idea to move this include down into a separate group of includes, just to help highlight that it's different from the ones around it.

I made an optional suggestion to that effect, in case @barracuda156 would be amenable to *amending* his commit. (It needs to be a single commit to pass all checks.)

barracuda156:
@markusicu Sure, I can make this change and rebase.

icu4c/source/tools/toolutil/writesrc.cpp
barracuda156: @markusicu Done.

markusicu (approved):
lgtm tnx

richgillam (approved):
LGTM too.

barracuda156:
@markusicu @srl295 Turned out that we still need to add `__STDC_FORMAT_MACROS` define, i.e.:
```
#define __STDC_FORMAT_MACROS 1
#include <cinttypes>
```

Specifically, while builds with GCC are all good with just fixing a header, as well as Clang builds on 10.7+, one configuration was found to fail: 10.6 with Clang.
Discussion here: https://github.com/macports/macports-ports/commit/2a477ac3558ce6342ecb151f1c683657e7558349

Potentially may be relevant for some other, non-Apple, systems: https://en.cppreference.com/w/cpp/types/integer (see Notes).

markusicu:
> @markusicu @srl295 Turned out that we still need to add `__STDC_FORMAT_MACROS` define, i.e.:

Ok. I just sent https://github.com/unicode-org/icu/pull/2250

PR #2242, opened 2022-10-27
ICU-21180 RuleBasedBreakIterator, refactor init.

aheninger:
In class RuleBasedBreakIterator, refactor how object initialization is handled by the various constructors, taking advantage of C++11's ability to directly initialize data members in the class declaration.

This will simplify ongoing maintenance of the code by eliminating the need to keep initialization lists synchronized with the class data members. This is being done now in preparation for additional changes to fix problems with the handling of memory allocation failures.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21180

### DIFF
--- a/icu4c/source/common/rbbi.cpp
+++ b/icu4c/source/common/rbbi.cpp
@@ -63,9 +63,8 @@ UOBJECT_DEFINE_RTTI_IMPLEMENTATION(RuleBasedBreakIterator)
  * tables object that is passed in as a parameter.
  */
 RuleBasedBreakIterator::RuleBasedBreakIterator(RBBIDataHeader* data, UErrorCode &status)
- : fSCharIter(UnicodeString())
+ : RuleBasedBreakIterator(status)
 {
-    init(status);
     fData = new RBBIDataWrapper(data, status); // status checked in constructor
     if (U_FAILURE(status)) {return;}
     if(fData == nullptr) {
@@ -102,9 +101,8 @@ RuleBasedBreakIterator::RuleBasedBreakIterator(UDataMemory* udm, UBool isPhraseB
 RuleBasedBreakIterator::RuleBasedBreakIterator(const uint8_t *compiledRules,
                        uint32_t       ruleLength,
                        UErrorCode     &status)
- : fSCharIter(UnicodeString())
+ : RuleBasedBreakIterator(status)
 {
-    init(status);
     if (U_FAILURE(status)) {
         return;
     }
@@ -141,9 +139,8 @@ RuleBasedBreakIterator::RuleBasedBreakIterator(const uint8_t *compiledRules,
 //
 //-------------------------------------------------------------------------------
 RuleBasedBreakIterator::RuleBasedBreakIterator(UDataMemory* udm, UErrorCode &status)
- : fSCharIter(UnicodeString())
+ : RuleBasedBreakIterator(status)
 {
-    init(status);
     fData = new RBBIDataWrapper(udm, status); // status checked in constructor
     if (U_FAILURE(status)) {return;}
     if(fData == nullptr) {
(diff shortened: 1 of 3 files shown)


icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -5231,7 +5231,7 @@ void RBBITest::TestTraceCreateBreakEngine(void) {
 
     // To word break the following text, BreakIterator will create 5 dictionary
     // break engine internally.
-    brkitr->setText(
+    UnicodeString text(
```
aheninger, line 5234: Note for reviewers: this change fixes a pre-existing bug that began causing CI failures with this PR. With the original code, brkitr->setText(UnicodeString &s) was being passed a temporary string that immediately went out of scope. But setText() retains a reference to it for use during iteration, which becomes a dangling reference.

icu4c/source/common/unicode/rbbi.h
```diff
@@ -71,32 +71,38 @@ class U_COMMON_API RuleBasedBreakIterator /*U_FINAL*/ : public BreakIterator {
...
      */
-    RBBIDataWrapper    *fData;
+    RBBIDataWrapper    *fData = nullptr;
+
 private:
+    /**
+      * The saved error code associated with this break iterator.
+      * This is the value to be returned by copyErrorTo().
+      */
+    UErrorCode      fErrorCode = U_ZERO_ERROR;
```
aheninger, line 81: Note for Reviewers: talking about copyErrorTo() already is jumping the gun a bit on an upcoming PR, which will introduce it as its primary point. But adding fErrorCode is needed now to allow the default constructor to delegate to the one taking a UErrorCode.

icu4c/source/test/intltest/rbbitst.cpp
markusicu: lgtm but I am curious why this is failing only now.
Does this have something to do with the switch to a UCharCharacterIterator?

markusicu (approved):
lgtm tnx

icu4c/source/test/intltest/rbbitst.cpp
aheninger: Good question, I was curious too. It does indeed have to do with the switch to UCharCharIterator.

StringCharacterIterator::setText() makes a copy of the incoming UnicodeString.
Clever reference counted buffer sharing in UnicodeString ends up keeping the heap allocated buffer alive after the original UnicodeString goes out of scope, for use by the copied string. And it is that buffer that the break iterator operates on, via UText.

The original code would still fail (I think) if the UnicodeString was short enough to use the string's internal buffer instead of the heap buffer.

It's all kind of funny, that the CharacterIterator stuff is only there for compatibility with obsolete API functions that probably no one is using, but its implementation has this unintended side effect on fundamental rbbi APIs that everyone uses.

There's a real chance that this change will break some existing users. What, if anything, to do about it warrants further discussion. Copy the input string all of the time, because that's the least error prone? Never copy, because that's what we've always said we do? Copy some of the time, to avoid potential new breakages? (But what would we be promising, and how would we ever explain it?)

No matter what we decide, I'd like to separate the question from this PR, and from this Jira issue, so that it gets better visibility.

markusicu: Ok, so we have done more than we promise:
> The BreakIterator will retain a reference to the supplied string. The caller must not modify or delete the text while the BreakIterator retains the reference.

What we say is consistent with other APIs, so no surprise. And at least some of our users will be calling the C API which can only take pointer+length.

icu4c/source/common/rbbi.cpp
```diff
@@ -194,10 +190,35 @@ RuleBasedBreakIterator::RuleBasedBreakIterator( const UnicodeString  &rules,
...
+/**
+ * Simple Constructor with an error code.
+ * Handles common initialization for all other constructors.
+ */
+RuleBasedBreakIterator::RuleBasedBreakIterator(UErrorCode &status) {
+    utext_openUChars(&fText, nullptr, 0, &status);
+    LocalPointer<DictionaryCache> lpDictionaryCache(new DictionaryCache(this, status), status);
+    LocalPointer<BreakCache> lpBreakCache(new BreakCache(this, status), status);
+    if (U_FAILURE(status)) {
+        fErrorCode = status;
```
grhoten, line 206: This data member is never read. Isn't the parameter sufficient for the status?

grhoten: Never mind. This is mentioned later on as being used with copyErrorTo in a future pull request.

richgillam (approved):
LGTM, AFAICT.

roubert (approved):
LGTM

PR #2249, opened 2022-11-06
ICU-7747 CMake Port

clemenswasser:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-7747

##### Summary

This adds CMake and CTest support to ICU4C. I haven't yet tested this on all platforms and there is still some issue with the generation of `icudt.dll`, which causes a `ICU Initialization returned: U_FILE_ACCESS_ERROR` when running `icuinfo` and most tests, hence this is marked as a draft. Since I'm not so familiar with the icudt generation, I hope that you can help me with fixing this issue.


clemenswasser:
Hi @markusicu @srl295,
you both seem to be involved with the Jira Ticket [ICU-7747](https://unicode-org.atlassian.net/browse/ICU-7747), is there something I can do to move this PR forward?
As I've stated in the initial description, in the current form it already builds on some platforms, but there is currently still some issue with icudt dll generation, which causes most tests to fail. I tried to find what's missing, but with my limited ICU expertise I didn't get very far with debugging this issue, hence I would love to get some feedback/suggestions from you.
My current guess is, that there is some sort of binary patching of the `U_ICUDATA_ENTRY_POINT` variable happening during the "makedata" step, which doesn't happen in my CMake port.

theoparis:
Can you add a cmake install target as well? I'm looking to use this in my own build files but I need to be able to install ICU to a prefix/sysroot.

srl295:
Hi. By the way, as I noted https://github.com/mesonbuild/wrapdb/issues/950#issuecomment-1490972572 there's a sources.txt and dependencies files, which should be used for getting the source files.  My cmake branch used it.

srl295:
@clemenswasser ^ that link to gencmake.py goes to my old branch.  And by old i mean old.

ceztko:
> 1. source/stubdata is built into a library. It's like a data library but with no data.
> 2. common and i18n and others are built, linked against stubdata
> 3. now the tools can be built, linked against above
> 4. now data can be generated. This creates a new library
> 

Is this the reason why ICU needs a pre-compilation in the host architecture for cross-compiling? It's something I really hate about the current build system of ICU.

srl295:
@ceztko this is for a bootstrap from source (i.e. git).  I hate doing rework too, but what else would you do?

if you have a release tarball, it has `source/data/in/icudt72l.dat` which can be used directly (in mmap or fread mode).  In that case, you can skip the tools part and just use that as your icu data dir. or provide it another way/ That may make more sense for cross compilation.

But, at least what it means is that if you have such a file, you can skip most of the tools, you only need enough to convert icudt72l.dat to libicudt.so

ceztko:
> But, at least what it means is that if you have such a file, you can skip most of the tools, you only need enough to convert icudt72l.dat to libicudt.so. (That's how I have node.js do its icu build) [...] but what else would you do?

I would depend on an highly available scripting language (eg. python) to generate the compilation units. Maybe that's what you were trying already with gencmake.py?

srl295:
These are two different issues. gencmake.py is for the code structure.  It's a way to slice up which .cpp's need to be included with finer granularity than the library. 

The reason for the double build in the case of cross platform has to do with the original data, which needs ICU source to run.  It's really a bootstrap issue.

Converting a binary .dat file to a .so with a large pure text segment could be done a number of ways. There's probably some other tool out there to do it by now. (Now as in, not 2002) 

The dependencies among the data generators themselves are already implemented in python, that's BUILDRULES.py

ceztko:
> The reason for the double build in the case of cross platform has to do with the original data, which needs ICU source to run. It's really a bootstrap issue.

Yes, I understood. As you say, it would be easier to trim pre-compilation when the .dat is available, which it would make the life easier for most people that just want to use ICU. Anyway, I would welcome a CMake port a lot, with or without the pre-build step, because it makes cross-compilation much easier in general.

srl295:
> > The reason for the double build in the case of cross platform has to do with the original data, which needs ICU source to run. It's really a bootstrap issue.
> 
> Yes, I understood. As you say, it would be easier to trim pre-compilation when the .dat is available, which it would make the life easier for most people that just want to use ICU. Anyway, I would welcome a CMake port a lot, with or without the pre-build step, because it makes cross-compilation much easier in general.

the .dat file is already there in the .tgz and .zip files.  And for that matter, building data should be able to use the python script for the data deps.

icu4c/source/data/makedata.mak
```diff
@@ -284,7 +284,13 @@ ICU_DATA_BUILD_VERBOSE=
...
 # Keep track of whether they are built via timestamp files.
 
-$(TOOLS_TS): "$(ICUTOOLS)\genrb\$(CFGTOOLS)\genrb.exe" "$(ICUTOOLS)\gencnval\$(CFGTOOLS)\gencnval.exe" "$(ICUTOOLS)\gencfu\$(CFGTOOLS)\gencfu.exe" "$(ICUTOOLS)\icupkg\$(CFGTOOLS)\icupkg.exe" "$(ICUTOOLS)\makeconv\$(CFGTOOLS)\makeconv.exe" "$(ICUPBIN)\pkgdata.exe"
+GENRB_DIR=$(ICUTOOLS)\genrb\$(CFGTOOLS)
+GENCNVAL_DIR=$(ICUTOOLS)\gencnval\$(CFGTOOLS)
+GENCFU_DIR=$(ICUTOOLS)\gencfu\$(CFGTOOLS)
+ICUPKG_DIR=$(ICUTOOLS)\icupkg\$(CFGTOOLS)
+MAKECONV_DIR=$(ICUTOOLS)\makeconv\$(CFGTOOLS)
+
+$(TOOLS_TS): "$(GENRB_DIR)\genrb.exe" "$(GENCNVAL_DIR)\gencnval.exe" "$(GENCFU_DIR)\gencfu.exe" "$(ICUPKG_DIR)\icupkg.exe" "$(MAKECONV_DIR)\makeconv.exe" "$(ICUPBIN)\pkgdata.exe"
```
srl295, line 293: flattened out the build directories?

clemenswasser: Yeah these are some "hacks" with which I wanted to get the real, non-stub, icudt dll to generate. I'll look into cleaning up and making this PR actually functioning tomorrow.

eli-schwartz:
> Is this the reason why ICU needs a pre-compilation in the host architecture for cross-compiling? It's something I really hate about the current build system of ICU.

As a matter of curiosity -- any particular challenge this causes, other than the general need to possess both a host and a build compiler?

eli-schwartz:
Ah right, the cmake cross compile situation ugliness, how could I forget... You can cross build the tools you need, it's just painfully manual.

I'm used to doing this -- from the meson port for ICU that @srl295 cross referenced above:
```meson
genccode_exe = executable(
  'genccode',
  sources,
  dependencies: toolutil_dep,
  install: true,
)

if meson.can_run_host_binaries()
  genccode_native_exe = genccode_exe
else
  genccode_native_exe = executable(
    'genccode-native',
    'genccode.c',
    dependencies: toolutil_native_dep,
    native: true,
  )
endif
```

The `native: true` kwarg here is key. It says that you're building a tool that *must* be built by the build compiler, not the cross one, meson tries to find both for you. A native target can be run by other build steps but not installed, and can be linked to dependencies that are either looked up by `dependency('...', native: true)` or built internally with the same kwarg. The same way this native genccode is defined, toolutil_dep and toolutil_native_dep are defined.

In the event that meson detects you can run cross-compiled binaries anyway, it doesn't bother to build additional copies of the installable genccode. This could happen for example when building mingw copies of ICU under Linux, and define `wine` as a cross wrapper, or when cross compiling from Linux x86-64 to aarch64 but with qemu-user set as a cross wrapper.

You can do the same thing with autotools although it's not as simple to author the necessary configure/make bits. The basic trick is to compile the files with $(CC_FOR_BUILD).

srl295:
we'd all rather forget ! 

Also you only need the non-native genccode if the end user wants to build/rebuild their own packages on the target arch.  I would put that in a developer tools package if i were packaging icu4c such as for a platform, separate from the library itself.

ceztko:
> The only reason that icudata is a .so is for user convenience (at the expense of INconvenience of [us builders](https://github.com/nodejs/node/commit/ac2857b12cd819b68405b15c3f8e95e48bcc32d8), sorry :-) …  so that the system library loader can find ICU's data. But if you have another mechanism, that can be used too. Lots of other ways. I hear you, you just want to use it. I think the only way you could avoid that 2nd ICU build is if there was some other tool to build the data .so from the binary file.

I don't know exactly how icudata is turned into .so. Sorry if I am saying something very obvious, but if it's just a single blob of data that must be available from a symbol at runtime (but I really don't know) then I would create a shim header and optionally generate the source calling of the compilation unit with the defined data in octal form as the next sample:

```
const char s_srgb_icc[] = "\
\000\000\014\110\114\151\156\157\002\020\000\000\155\156\164\162\
\122\107\102\040\130\131\132\040\007\316\000\002\000\011\000\006\
...
```

There's no ubiquitous scripting language that can be picked for the task, but one may assume that this could be done in python **if** python interpreter is found and a .dat file is specified. If requested I can write the python script to convert a blob to a cpp file, but I suspect that's not really problem (unless you welcome a small help). Other [methods](https://stackoverflow.com/a/4158997/213871) may be not portable (eg. in Windows), or require non ubiquitous utilities,.

eli-schwartz:
> Other methods may be not portable (eg. in Windows), or require non ubiquitous utilities,.

The epitome of a non ubiquitous utility would be a c23 compiler with support for `#embed`, but we can dream...

srl295:
c23…  that's available in Ubuntu 14.04 LTS right??

@ceztko what you described is what the `genccode` utility does. 

it generates assembly where it can, because that's a lot faster and doesn't run into compiler memory issues.

and on windows it goes directly from a .dat file to a PE because there are utilities for that.

ceztko:
> @ceztko what you described is what the `genccode` utility does.
> 
> and on windows it goes directly from a .dat file to a PE because there are utilities for that.

I could imagine you had an utility doing exactly that task, but again this a problem of not falling into chicken-and-egg situation and to skip a build you'll have to do it using something more ubiquitous. You could for example benefit if genccode is already [installed](https://packages.ubuntu.com/kinetic/icu-devtools) in Unix platforms, and use directly PE utilities in Windows. Maybe this could be a good compromise?

srl295:
@ceztko Are there PE utilities which will do this? Yes. you could detect genccode (and other ICU utils) in the environment.  Would take some rework but that could be possible. You need the ICU utils to be able to create the _target_ environment's format. 

I did find this https://stackoverflow.com/questions/4158900/embedding-resources-in-executable-using-gcc

ceztko:
> Are there PE utilities which will do this?

Maybe I misunderstood a previous message of yours where you talked about use of PE utilities? Anyway, assuming ICU utils can be easily installed on Unix, which is true in most situations, an alternative in Windows may be use of of [Windows Resource Compiler](https://learn.microsoft.com/en-us/windows/win32/menurc/resource-compiler). I can have a look if it fits enough: for sure with CMake it's quite easy to use .rc files using MSVC as the compiler (I did it multiple times), but I never tried to use them to embed files.

srl295:
> > Are there PE utilities which will do this?
> 
> Maybe I misunderstood a previous message of yours where you talked about use of PE utilities? 

I guess it's Windows and also ELF on linux. search icu codebase for `CAN_GENERATE_OBJECTS` - genccode writes directly to a .o file, which is then linkable.

> Anyway, assuming ICU utils can be easily installed on Unix, which I think it is true in most situations, an alternative in Windows may be use of of [Windows Resource Compiler](https://learn.microsoft.com/en-us/windows/win32/menurc/resource-compiler). I can have a look if it fits enough: for sure with CMake it's quite easy to use .rc files using MSVC as the compiler (I did it multiple times), but I never tried to use them to embed files.

Would it be happy with a 30Mb .dat file though? not sure

eli-schwartz:
Ultimately the goal here is, I think, to explicitly eschew the use of official solutions such as producing a C source file containing octal data. Because... it can get *extremely* slow depending on compiler, and is never really fast. With MSVC it's a common experience to simply run out of memory.

Instead, sideloading your own objects can be extremely advantageous for build speed. It's also why `#embed` is so useful (or will be once compilers implement it, now that we're guaranteed by the newest standard that it shall exist).

Here's some discussion on it by the author:
https://thephd.dev/full-circle-embed#speed-and-space-is-everything
https://thephd.dev/embed-the-details
https://thephd.dev/finally-embed-in-c23

srl295:
> Ultimately the goal here is, I think, to explicitly eschew the use of official solutions such as producing a C source file containing octal data. Because... it can get _extremely_ slow depending on compiler, and is never really fast. With MSVC it's a common experience to simply run out of memory. The task starts getting hopeless past like 4mb.

but to be clear:  a c source file with octal data is _only_ done when no other mechanism is available… i.e. writing a .o file (linux elf, windows PE), generating asm (most others…) 

So yes it's super slow. But it hasn't been common for maybe 15 years here.

> Instead, sideloading your own objects can be extremely advantageous for build speed. It's also why `#embed` is so useful (or will be once compilers implement it, now that we're guaranteed by the newest standard that it shall exist).
> 
> Here's some discussion on it by the author: https://thephd.dev/full-circle-embed#speed-and-space-is-everything https://thephd.dev/embed-the-details https://thephd.dev/finally-embed-in-c23

i saw that one also. if there are *any* compilers which support this, then we could start to have a PR to allow it when available.

ceztko:
> > I can have a look if it fits enough: for sure with CMake it's quite easy to use .rc files using MSVC as the compiler (I did it multiple times), but I never tried to use them to embed files.
> 
> Would it be happy with a 30Mb .dat file though? not sure

I just tried and yes it handles it and it's very fast. Give me a second to craft a full example.

srl295:
> > > I can have a look if it fits enough: for sure with CMake it's quite easy to use .rc files using MSVC as the compiler (I did it multiple times), but I never tried to use them to embed files.
> > 
> > 
> > Would it be happy with a 30Mb .dat file though? not sure
> 
> I just tried and yes it handles it and it's very fast. Give me a second to craft a full example.

that's great and definitely simpler.

ceztko:
Ok, you can find it attached here:
[TestResourcesWin.zip](https://github.com/unicode-org/icu/files/11127353/TestResourcesWin.zip)

On Windows I just do the usual:
```
md build
cd build
cmake ..
```

The resource is declared with a simple `Resources.rc`
```
#include <winres.h>

MyRes UserResources data.res
```

That loads `data.res` which is just the string "Hello World" followed by 30mb or null characters.

I test it with the following source:

```
#include <Windows.h>
#include <cassert>
#include <string_view>

using namespace std;

int main()
{
    HRSRC hResInfo = FindResourceA(NULL, "MyRes", "UserResources");
    HGLOBAL hRes = LoadResource(NULL, hResInfo);
    LPVOID lpRes = LockResource(hRes);
    DWORD dwSize = SizeofResource(NULL, hResInfo);
    string_view view((const char*)lpRes);
    assert(view == "Hello World");
    return 0;
}
```

It was much more easier than I expected, but this was aided by CMake that does know exactly how to deal with .rc files. I have no idea how this should implemented for the current build system of ICU on Windows, but after all we are talking about this topic in a PR about a CMake port, which should definitely ease builds on Windows as well.

srl295:
Probably icu would have to be able to load using the find resource call to take advantage of this. But still it's definitely fewer tools. And could be worth it for the benefit for cross compile. 

Would you have to link with the DLL or not ?

eli-schwartz:
The paper for `#embed` talks about rc files in the same "non-portable platform specific tricks" sentence as objcopy, which I assume without any knowledge that it's the moral equivalent of. I'd assume it will do the job, yeah -- which is good when the dream of `#embed` is not yet here. I'm pretty sure GCC hasn't merged an implementation, and clang has a status page for more than just CXX -- their C page says it's not yet implemented. So there's a way to go here, and .rc tricks will help.

(Incidentally, meson can handle rc files for you too.)

ceztko:
> Would you have to link with the DLL or not ?

In the case of dll is also easy enough. You just have to discover the module from an exported symbol as it follows:

```
_declspec(dllexport) string Foo()
{
    // Get the module handle out out this exported method
    HMODULE hModule;
    GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
        GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
        (LPCSTR)&Foo, &hModule);

    HRSRC hResInfo = FindResourceA(hModule, "MyRes", "UserResources");
    HGLOBAL hRes = LoadResource(hModule, hResInfo);
    LPVOID lpRes = LockResource(hRes);
    DWORD dwSize = SizeofResource(hModule, hResInfo);
    string str((const char*)lpRes);
    UnlockResource(hRes);
    return str;
}
```

Attached the fully working example. 
[TestResourcesWinDll.zip](https://github.com/unicode-org/icu/files/11128827/TestResourcesWinDll.zip)

ceztko:
Any chance CMake port will be considered again for integration in the near future?

aberaud:
Please consider moving to CMake and/or Meson.
Autotools is obsolete and causing issues.

srl295:
> Please consider moving to CMake and/or Meson.

Would love to, it's not able to do the data build. That might be easier now that the data build is partly in Python. 

> Autotools is obsolete and causing issues.

Can you be more specific? How "obsolete"?  Autoconf 2.73 was released not 7 months ago, so it's certainly not abandoned.  What issues?

eli-schwartz:
autotools can be objectively described as "slow" and also as "annoying to set up a build environment on windows". And subjectively described as "painful to be a developer for or with".

I certainly haven't heard any news about it being obsolete though -- many people remain convinced of its superiority, such as more or less the entire GNU project, which is determined to continue maintaining it and continue using it.

"Issues" is such a vague term. autotools is technically capable of doing anything the others can do, even if projects may have bugs because they used it improperly (and that can be fixed as, well, a bugfix).

In fact, autotools can do a lot more than any other build system. The usual counterargument is simply that many of those things shouldn't be done or aren't useful to do or only one in a thousand projects actually needs to do it.

The reasons to change are generally "better support for windows without a mingw development environment prerequisite" and "speed". Those aren't "issues", merely conveniences. Even if I do personally think they are worth the change. ;)

ceztko:
@srl295 Autoconf still creates friction for Windows users, and and there is a generation of developers who refutes it for precisely this reason and for the unacceptably steep (by today's standards) learning curve, especially if you want to modify the build itself. Nowadays you are either part of the CMake team or the Meson team, and most people will tell you that they are both excellent tools. Meson looks nicer from the outside, but carries an implicit dependency on python and pushes for some orthodox design choices, like [prohibiting](https://github.com/mesonbuild/meson/issues/3234) user's defined functions. I tend to prefer CMake because it has more pragmatic approach, and still looks nice and easy to understand.

eli-schwartz:
> Meson looks nicer from the outside, but carries an implicit dependency on python and pushes for some orthodox design choices, like [prohibiting](https://github.com/mesonbuild/meson/issues/3234) user's defined functions.

Meson doesn't require python -- the reference implementation does. https://muon.build exists and bootstraps from c99, and was created precisely because not everyone wants to depend on python. It is essentially feature-complete ( == parity with the python reference implementation) for C/C++, although for things like Meson's rust support it is very "patches welcome but none seem forthcoming". I guess people who want to write rust code with Meson, don't have a problem with python. :)

> I tend to prefer CMake because it has more pragmatic approach, and still looks nice and easy to understand.

The world is full of opinions on that, clearly. I can give you an endless list of people who think that cmake both looks bad and is impossible to understand due to fourth-order effects. I usually cite the ";" rule, myself, when talking about "looks nice and easy to understand". Of course, I am biased. :D

aberaud:
> Can you be more specific? How "obsolete"? Autoconf 2.73 was released not 7 months ago, so it's certainly not abandoned. What issues?

The main problems with autotools, which will likely never get fixed because the project is mostly maintenance-only now:
* Windows support
* The absurdly long configure time, running every single build, wasting build machine and developers time and resources, to support build paradigms which turned out to be bad ideas and were never embraced (shipping configured tarballs and multiplatform support using config.h and macros). This means that projects using autotools (like ICU) become an unnecessary burden and ongoing cost for downstream projects building them. 
* The fact that the autotoools language is so obscure, even for experienced users. The fact that a project build system relies on such an obscure language is a security hazard. This was shown during the xz ssh supply chain attack, where the attack payload was deployed in production using an obscure autotools script which looked even to experienced eyes just like any other autotools gibberish.

ceztko:
> The world is full of opinions on that, clearly. I can give you an endless list of people who think that cmake both looks bad and is impossible to understand due to fourth-order effects.

Oh, but I also have a lot of things I don't like about CMake. In my opinion, both CMake and Meson are excellent products; only a couple of deciding factors (which are still relevant and not subjected to change anytime soon) make me prefer the former. By the way: it seems [#embed](https://en.cppreference.com/cpp/preprocessor/embed) is finally on its way for C++26.

eli-schwartz:
> The main problems with autotools, which will likely never get fixed because the project is mostly maintenance-only now:
> 
> * Windows support

Supported, not a problem, merely a convenience.

> * The absurdly long configure time, running every single build, wasting build machine and developers time and resources, to support build paradigms which turned out to be bad ideas and were never embraced (shipping configured tarballs and multiplatform support using config.h and macros). This means that projects using autotools (like ICU) become an unnecessary burden and ongoing cost for downstream projects building them.

config.h is optional in autotools, and is just as present in CMake and Meson -- widely used, in other words. Whether or not it is necessary is a property of the source code, not the build system. Anyone who thinks that config.h is useless has never written cross platform software, and yes, I do mean cross platform in the sense of "supports Windows, Linux, macOS and FreeBSD" -- I'm not even talking about Haiku, BeOS, SCO OpenServer, OpenVMS, Tru64 etc. 

Simply supporting the "big three" all on its own has a strong tendency to need config.h for anything nontrivial. And actually supporting software with build options and optional features, needs to pass feature gates for the enabled features. Welcome to config.h. 

It's not particularly slow in cmake or meson, and it's not the major reason autotools is slow.

But also I really really really think you need to start using build caches. On linux we call that cache, a "package manager". They are super useful and you can get primitive versions of them on windows too -- it's called vcpkg / conan.

Or just use msys2.

> * The fact that the autotoools language is so obscure, even for experienced users. The fact that a project build system relies on such an obscure language is a security hazard. This was shown during the xz ssh supply chain attack, where the attack payload was deployed in production using an obscure autotools script which looked even to experienced eyes just like any other autotools gibberish.

... The xz issue had nothing to do with autotools unless you get all your information from digital tabloids.

It was a social engineering attack. The xz maintainer was socially engineered into trusting someone until he stopped actually looking at the autotools script. Jia Tan spent a significant part of the social engineering time simply convincing the original xz maintainer that he was competent to manage releases, including by sending him prepared tarballs that were compared to the original maintainer's tarballs, and was regularly interrogated for getting "different results". The actual hack only appeared after people decided it was finally time to blindly trust Jia Tan. It turns out that virtually nobody actually looks at project source code when they are "pretty sure it's trustworthy, after all it's used by loads of people".

What you want is anyways just https://reproducible-builds.org, this is totally trivial to verify.

Regardless it's all an extra-big red herring. Jia Tan committed traps to the experimental cmake files too.  Just nobody cares because Debian didn't use cmake to build. It turns out that it is indeed possible to hide malicious changes in many places, if you care to do so.

eli-schwartz:
> By the way: it seems [#embed](https://en.cppreference.com/cpp/preprocessor/embed) is finally on its way for C++26.

Yup, it's very exciting, although we can I think assume that sufficiently new compilers to support c23 will support this automatically in C++ frontends too, even without official C++26 support.

aberaud:
> > 
> > * Windows support
> 
> Supported, not a problem, merely a convenience.

It doesn't natively support building with MSVC, unlike CMake. For us (downstream users of ICU) that's really the main actual limitation.

> 
> config.h is optional in autotools, and is just as present in CMake and Meson -- widely used, in other words. Whether or not it is necessary is a property of the source code, not the build system. Anyone who thinks that config.h is useless has never written cross platform software, and yes, I do mean cross platform in the sense of "supports Windows, Linux, macOS and FreeBSD" -- I'm not even talking about Haiku, BeOS, SCO OpenServer, OpenVMS, Tru64 etc.

That's reasonable. Also, for ICU the configure script is well optimized and runs reasonably fast.

> ... The xz issue had nothing to do with autotools unless you get all your information from digital tabloids.

It did use autotools as part of the attack chain, and did conceal it using autotools gibberish, but of course you're right: the main part of the attack was the social engineering.

eli-schwartz:
> It did use autotools as part of the attack chain, and did conceal it using autotools gibberish, but of course you're right: the main part of the attack was the social engineering.

I'm making a much stronger point: that the "autotools gibberish" was incidental to the attack, that there were other ways to do a malicious attack (if you know that people trust you and don't actually look at what you do) which didn't rely on it, and traces of those other ways were discovered in the CMake files and reverted.

Autotools was simply the first attack run. Avoiding autotools doesn't save anybody.

If you're genuinely concerned about "autotools gibberish" attacks then it is trivial to have a workflow for verifying the reproducibility of the compiled outputs of configure.ac (which is quite readable). It's like any other compiler. People trust the vast majority of compiled ELF and PE outputs without even checking md5sums, let alone a collision-resistant hash, let alone code signing certificates, let alone verifying the reproducibility of the compiler.

...

Seriously, I would love to see non-FUD arguments for or against various choices in build systems. The "b-b-b-but the xz-utils hack" thing got really old approximately exactly when it first got started.

aberaud:
> Seriously, I would love to see non-FUD arguments for or against various choices in build systems. The "b-b-b-but the xz-utils hack" thing got really old approximately exactly when it first got started.

The point here is that the syntax is highly unpleasant and obscured by multiple layers of text replacement, for no good reasons today.
Call it "steep learning curve" or anything else. The xz thing is just a example of that being exploited IRL.

Yes, that point is mostly about convenience, not something that would require a build system change.

PR #2257, opened 2022-11-20
ICU-22211 macOS PPC should use `.p2align 4` instead `.balign 16` 

catap:
An attempt to build ICU on old macOS with PowerPC leads to an issue:
```
./out/tmp/icudt72b_dat.S:7:Unknown pseudo-op: .balign
./out/tmp/icudt72b_dat.S:7:Rest of line ignored. 1st junk character valued 49 (1).
```

Why? Because `as` is too old.

Anyway, switc to `.p2align 4` fix a build and allows to pass all tests.

See: https://trac.macports.org/ticket/66258

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22211

### DIFF
--- a/icu4c/source/tools/toolutil/pkg_genc.cpp
+++ b/icu4c/source/tools/toolutil/pkg_genc.cpp
@@ -158,6 +158,22 @@ static const struct AssemblyType {
 
         ".long ","",HEX_0X
     },
+    /* macOS PPC should use `.p2align 4` instead `.balign 16` because is
+     * unknown pseudo ops for such legacy system*/
+    {"gcc-darwin-ppc",
+        /*"\t.section __TEXT,__text,regular,pure_instructions\n"
+        "\t.section __TEXT,__picsymbolstub1,symbol_stubs,pure_instructions,32\n"*/
+        ".globl _%s\n"
+        "#ifdef U_HIDE_DATA_SYMBOL\n"
+        "\t.private_extern _%s\n"
+        "#endif\n"
+        "\t.data\n"
+        "\t.const\n"
+        "\t.p2align 4\n"
+        "_%s:\n\n",
+
+        ".long ","",HEX_0X
+    },
     {"gcc-cygwin",
         ".globl _%s\n"
         "\t.section .rodata\n"
--- a/icu4c/source/acinclude.m4
+++ b/icu4c/source/acinclude.m4
@@ -77,6 +77,7 @@ x86_64-*-cygwin)
 *-*ibm-openedition*|*-*-os390*)	icu_cv_host_frag=mh-os390 ;;
 *-*-os400*)	icu_cv_host_frag=mh-os400 ;;
 *-apple-rhapsody*)	icu_cv_host_frag=mh-darwin ;;
+powerpc*-apple-darwin*)	icu_cv_host_frag=mh-darwin-ppc ;;
 *-apple-darwin*)	icu_cv_host_frag=mh-darwin ;;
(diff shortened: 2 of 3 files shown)


pedberg-icu (dismissed):

icu4c/source/tools/toolutil/pkg_genc.cpp
```diff
@@ -153,7 +153,12 @@ static const struct AssemblyType {
         "#endif\n"
         "\t.data\n"
         "\t.const\n"
+        /* macOS PPC should use .align instead .balign because is unknown pseudo ops for it*/
+        "#if defined(__APPLE__) && defined(__POWERPC__)\n"
+        "\t.align 16\n"
+        "#else\n"
         "\t.balign 16\n"
+        "#endif\n"
```
markusicu, line 161: Would it make sense to provide both versions at the same time, and just give them different names?

markusicu:
FYI @richgillam catap signed the CLA on Nov 20, 2022. I don't know why the CLA UI does not show here on the PR. @srl295 ?

icu4c/source/tools/toolutil/pkg_genc.cpp
catap: Thanks, just pushed an updated version.

catap:
[rebased to the last master with hope that it makes CI green]

kencu:
NB: 

".align" in PowerPC assembly most probably refers to .p2align (power-of-two).
".balign" in PowrPC assembly refers to byte alignment.

So they may not be equivalent. Please check and verify before committing this change.

Ref: <https://reviews.llvm.org/D16549>

catap:
FYI: it was changed from `.align` to `.balign` not that long ago: https://unicode-org.atlassian.net/browse/ICU-10305

kencu:
if it is changed back to “.align” it would be important to use the right factor.

I believe for PowerPC, “.balign 16” equals “.align 4” equals “.p2align 4”.

I am looking for the official ibm PPC assembly reference, this is from llvm and Google.

catap:
@kencu thanks to catch it up. I've made a bit digging in history of changes and had discovered that before `.balign 16` it was `.align 8`: https://github.com/unicode-org/icu/commit/ac2fed5f3002b0418a8ec78501dc1cbbe7ad4daa

Let me quote your link:
> The way the required alignment is specified varies from system to system. For the a29k, hppa, m68k, m88k, w65, sparc, and Hitachi SH, and i386 using ELF format, the first expression is the alignment request in bytes. For example `.align 8' advances the location counter until it is a multiple of 8. If the location counter is already a multiple of 8, no change is needed.

> For other systems, including the i386 using a.out format, it is the number of low-order zero bits the location counter must have after advancement. For example `.align 3' advances the location counter until it a multiple of 8. If the location counter is already a multiple of 8, no change is needed.

=> this is "other system" and as far as I understand it the right command to zero 16 bytes should be `.align 128`.

From another hand `.p2align` seems cleaner:
> Pad the location counter (in the current subsection) to a particular storage boundary. The first expression (which must be absolute) is the number of low-order zero bits the location counter must have after advancement. For example `.p2align 3' advances the location counter until it a multiple of 8. If the location counter is already a multiple of 8, no change is needed.

=> I've updated my patch.

Thanks!

catap:
Base on http://personal.denison.edu/~bressoud/cs281-s07/Assembler.pdf (page 44) I think that GAS has incorrect wording and here should be `.p2align 4` as Ken pointed.

catap:
@pedberg-icu, @markusicu do you have anything in mind about this one?

PR #2264, opened 2022-12-02
ICU-22222 Add explicit instantiation declarations

tkoeppe:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22222

### DIFF
--- a/icu4c/source/i18n/unicode/numberformatter.h
+++ b/icu4c/source/i18n/unicode/numberformatter.h
@@ -2418,6 +2418,13 @@ class U_I18N_API NumberFormatterSettings {
     friend class impl::NumberRangeFormatterImpl;
 };
 
+// Explicit instantiations in source/i18n/number_fluent.cpp.
+// (MSVC treats imports/exports of explicit instantiations differently.)
+#ifndef _MSC_VER
+extern template class NumberFormatterSettings<UnlocalizedNumberFormatter>;
+extern template class NumberFormatterSettings<LocalizedNumberFormatter>;
+#endif
+
 /**
  * A NumberFormatter that does not yet have a locale. In order to format numbers, a locale must be specified.
  *
--- a/icu4c/source/i18n/unicode/numberrangeformatter.h
+++ b/icu4c/source/i18n/unicode/numberrangeformatter.h
@@ -381,6 +381,13 @@ class U_I18N_API NumberRangeFormatterSettings {
     friend class UnlocalizedNumberRangeFormatter;
 };
 
+// Explicit instantiations in source/i18n/numrange_fluent.cpp.
+// (MSVC treats imports/exports of explicit instantiations differently.)
+#ifndef _MSC_VER
+extern template class NumberRangeFormatterSettings<UnlocalizedNumberRangeFormatter>;
+extern template class NumberRangeFormatterSettings<LocalizedNumberRangeFormatter>;
+#endif
+
 /**
  * A NumberRangeFormatter that does not yet have a locale. In order to format, a locale must be specified.
  *


tkoeppe:
I think https://unicode-org.atlassian.net/browse/ICU-20455 shoud be reopened. It's not done, and the referenced https://unicode-org.atlassian.net/browse/ICU-9749 is something separate (though related).

tkoeppe:
Hm, sorry, I don't know how to make explicit template instantiation declarations work in MSVC.

@CaseyCarter, on the off-chance that you might have seen ICU before, would you happen to have an idea?

tkoeppe:
And changing `U_I18N_API` to `U_IMPORT` on the `extern template` declaration results in:

```
Error LNK2001: unresolved external symbol "__declspec(dllimport) public: __thiscall icu_72::number::NumberFormatterSettings<class icu_72::number::LocalizedNumberFormatter>::~NumberFormatterSettings<class icu_72::number::LocalizedNumberFormatter>(void)" (__imp_??1?$NumberFormatterSettings@VLocalizedNumberFormatter@number@icu_72@@@number@icu_72@@QAE@XZ)
```

tkoeppe:
Done!

srl295:
@markusicu @tkoeppe I'm seeing this pattern in many other sites:

```c
./source/i18n/erarules.h-// Export an explicit template instantiation of LocalMemory used as a data member of EraRules.
./source/i18n/erarules.h-// When building DLLs for Windows this is required even though no direct access leaks out of the i18n library.
./source/i18n/erarules.h-// See digitlst.h, pluralaffix.h, datefmt.h, and others for similar examples.
./source/i18n/erarules.h-#if U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN
./source/i18n/erarules.h:#if defined(_MSC_VER)
./source/i18n/erarules.h-// Ignore warning 4661 as LocalPointerBase does not use operator== or operator!=
./source/i18n/erarules.h-#pragma warning(push)
./source/i18n/erarules.h-#pragma warning(disable: 4661)
./source/i18n/erarules.h-#endif
./source/i18n/erarules.h-template class U_I18N_API LocalPointerBase<int32_t>;
./source/i18n/erarules.h-template class U_I18N_API LocalMemory<int32_t>;
./source/i18n/erarules.h:#if defined(_MSC_VER)
./source/i18n/erarules.h-#pragma warning(pop)
./source/i18n/erarules.h-#endif
./source/i18n/erarules.h-#endif
```

note it is checking `if U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN` and not just _MSC_VER 

there are some cases where ONLY `U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN` is checked and not _MSC_VER

srl295:
update: I'm noticing these are if NOT def _MSC_VER so it's a different situation.   so maybe it's OK.

markusicu:
I just looked through ICU4C for `MaybeStackArray<`. The spiffiest version seems to be the one in https://github.com/unicode-org/icu/blob/main/icu4c/source/common/charstr.h#L24-L28 :
```
// Windows needs us to DLL-export the MaybeStackArray template specialization,
// but MacOS X cannot handle it. Same as in digitlst.h.
#if !U_PLATFORM_IS_DARWIN_BASED
template class U_COMMON_API MaybeStackArray<char, 40>;
#endif
```

(FYI: digitlst.h no longer exists.)

Observations:
- We have a variety of conditions, and presumably they all work. Just `#ifndef _MSC_VER` is a new one. I wonder whether this is more of a compiler issue or a platform issue.
- We seem to never have `extern template`, just `template`. Do we need the `extern`?
- We always have the `U_I18N_API` in there, as in `template class U_I18N_API MaybeStackArray<int64_t, CEBUFFER_INITIAL_CAPACITY>;` ([collationiterator.h](https://github.com/unicode-org/icu/blob/main/icu4c/source/i18n/collationiterator.h#L34-L44))

Suggestion:
```
// Windows needs us to DLL-export this template specialization, but MacOS X cannot handle it.
#if !U_PLATFORM_IS_DARWIN_BASED
template class U_I18N_API NumberFormatterSettings<UnlocalizedNumberFormatter>;
template class U_I18N_API NumberFormatterSettings<LocalizedNumberFormatter>;
#endif
```

tkoeppe:
Well, `extern template` is the entire point of this change: this is a non-defining declaration of an explicit instantiation, which only became available in C++11, and which I'd like to roll out consistently guided by Clang's `-Wundefined-func-template` warning: basically, if a user names template specialization that has an explicit instantiation, the user should see a declaration of that explicit instantiation, too.

The problem I found with the `U_I18N_API` macro is that I couldn't find a consistent way to apply the macro to a primary class template that would work in MSVC when explicit instantiations are involved. If you _don't_ put a macro on the primary, and make the non-defining explicit instantiation declaration (the "extern template") imported, and the definition (non-extern just "template") exported, that works, but it all goes wrong if you also add the macro the primary.

I tried figuring this out for a few years (see issue history in Jira), but I'm giving up, and would just like to do this for the non-MSVC rest of us for now, until an MSVC expert comes along and can tell me how to declare explicit instantiations correctly in MSVC.

markusicu:
I give up. I don't know enough about the standard, and compilers disagree.

I realize that every other ICU template specialization is done *for MSVC* or *to the exclusion of macOS/iOS*, while your change is the opposite, it's *not for MSVC*.

Hacks like https://github.com/unicode-org/icu/blob/main/icu4c/source/i18n/unicode/numberformatter.h#L1602-L1610
```
#if (U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN) && defined(_MSC_VER)
// Ignore MSVC warning 4661. This is generated for NumberFormatterSettings<>::toSkeleton() as this method
// is defined elsewhere (in number_skeletons.cpp). The compiler is warning that the explicit template instantiation
// inside this single translation unit (CPP file) is incomplete, and thus it isn't sure if the template class is
// fully defined. However, since each translation unit explicitly instantiates all the necessary template classes,
// they will all be passed to the linker, and the linker will still find and export all the class members.
#pragma warning(push)
#pragma warning(disable: 4661)
#endif
```
look related, but again, those are *for MSVC*.

So you need these lines, and our CI tests all pass. Let's do it.

markusicu (approved):
rslgtm see ticket & PR discussion

tkoeppe:
Many thanks!

It's quite possible that the MSVC code can be cleaned up nowadays if you expect a sufficiently modern MSVC, but it would probably require a few changes, e.g. removing the API macro from the primary class templates. Maybe some toy example could be demonstrated somewhere, if you're interested in such a cleanup.

PR #2265, opened 2022-12-02
ICU-9749 Move explicit instantiation definition, add declaration

tkoeppe:
Moves the definition of an explicit template instantiation from the header number_decnum.h into the .cpp file number_decimalquantity.cpp, and adds a declaration of the instantiation to the header instead.

Removes old guard macros for specific platforms. The new construction is portable C++11.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-9749


tkoeppe:
Actually, I cannot figure out how to get this to work. Abandoning.

PR #2266, opened 2022-12-12
ICU-22193 fix some CI test failures

markusicu:
Fixing some CI test failures that I noticed on other pending PRs.
- API doc warnings about parameters in function typedefs
- dependency checker: mutex/init-once code calling pthread_* functions

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22193

### DIFF
--- a/icu4c/source/test/depstest/dependencies.txt
+++ b/icu4c/source/test/depstest/dependencies.txt
@@ -48,6 +48,9 @@ group: std_mutex
     std::condition_variable::~condition_variable()
     std::condition_variable_any::condition_variable_any()
     std::condition_variable_any::~condition_variable_any()
+    pthread_once
+    pthread_mutex_lock
+    pthread_mutex_unlock
 
 group: ubsan
     # UBSan=UndefinedBehaviorSanitizer, clang -fsanitize=bounds
--- a/icu4c/source/common/unicode/icuplug.h
+++ b/icu4c/source/common/unicode/icuplug.h
@@ -187,8 +187,12 @@ typedef enum {
 
 /**
  * Entrypoint for an ICU plugin.
- * @param plug the UPlugData handle. 
- * @param status the plugin's extended status code.
+ * @param plug the UPlugData handle.
+ * @param reason the reason code for the entrypoint's call.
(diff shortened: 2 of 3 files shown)


richgillam (approved):
I don't completely know what I'm looking at, but the changes seem to be not unreasonable...

icu4c/source/test/depstest/dependencies.txt
```diff
@@ -48,6 +48,9 @@ group: std_mutex
     std::condition_variable::~condition_variable()
     std::condition_variable_any::condition_variable_any()
     std::condition_variable_any::~condition_variable_any()
+    pthread_once
```
aheninger, line 51: This is kind of odd, because there is no direct use of anything pthread from ICU code; everything should be using C++ std library functions.

I grepped the ICU source, and didn't find any use of  pthread functions. Maybe it's indirect usage, via the C++ std library. There are mentions of pthread stuff in configure; whether that is obsolete, I don't know.

It would be best to somehow keep the pthread functions out of the std_mutex group, so that any attempt to use pthread stuff directly from ICU code remains an error.

roubert (approved):
LGTM

icu4c/source/test/depstest/dependencies.txt
markusicu: I have no idea what changed that we get these and other test failures now. Maybe the OS on GHA machines got upgraded to the next level (next ubuntu version)?

My guess on this one: Maybe the clang standard library got updated and now uses inline implementations of std:: threading functions.

> It would be best to somehow keep the pthread functions out of the std_mutex group, so that any attempt to use pthread stuff directly from ICU code remains an error.

I can move them to a different group, but other than hacking special knowledge about one special group into the Python script I don't know how to prevent other ICU code from calling these. It seems like the system-symbols groups are already special in that they specify symbols, not .o files.

aheninger: Or just let it be (as proposed in this PR). The potential problem of unwanted use of pthreads seems very unlikely to actually happen, and would break Windows builds.

aheninger (approved):

PR #2272, opened 2022-12-16
ICU-22233 Use separate Bazel cache keys per OS

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22233

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -486,7 +486,7 @@ jobs:
 
   # Build Unicode update tools
   unicode-update-tools:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
     - uses: bazelbuild/setup-bazelisk@v1
@@ -494,7 +494,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: "~/.cache/bazel"
-        key: bazel
+        key: ${{ runner.os }}-bazel
 
     - name: Generate the data
       run: |


roubert:
Hm, apparently not transient after all, just seemingly impossible for me to reproduce when running locally.

icu4c/source/data/unidata/generate.sh
```diff
@@ -30,6 +30,9 @@ rm $ICU4C_DATA_IN/coll/*.icu
 # Exit this shell script when a command fails.
 set -e
 
+# Workaround for obscure bug with the Bazel cache (ICU-22193).
```
markusicu, line 33: Would you mind moving this comment / commit message / PR title to ICU-22233?

markusicu (commented):
This is great, thanks!

One question below, plus: Is this a known Bazel bug? If not, worth reporting?

.github/workflows/icu_ci.yml
```diff
@@ -486,7 +486,7 @@ jobs:
 
   # Build Unicode update tools
   unicode-update-tools:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
echeran, line 489: The cache key is configured in this job. If the cache has worked for ubuntu-20.04, but the crossover to ubuntu-22.04 has caused problems, then maybe we can update the key so that caching is per-OS version. There are lots of [examples of that](https://github.com/actions/cache/blob/main/examples.md#rust---cargo).

So below where it says `key: bazel`, it would change to `key: ${{ runner.os }}-bazel`.

icu4c/source/data/unidata/generate.sh
```diff
@@ -30,6 +30,9 @@ rm $ICU4C_DATA_IN/coll/*.icu
...
+bazelisk clean --expunge
```
echeran, line 34: Interesting! Since this change works, then this tells us what the problem is: the compiled code in the cache from the tools in Ubuntu 20.04 is interfering with the code compilation with the tools in Ubutunu 22.04.

Of course, force-clearing the cache cancels out any benefit of caching in the first place... 

My other comment in the Github Actions suggests a possibility in which we can keep the cache mechanism (if it works, then it should have some benefit) by tying it to the OS version.

If fixing the caching works, then we wouldn't need to always clear the cache first.

echeran:
> One question below, plus: Is this a known Bazel bug? If not, worth reporting?

I filed https://github.com/bazelbuild/setup-bazelisk/issues/70

echeran (approved):
Great, LGTM!

(btw, you might want to edit the PR title prefix to ICU-22233 to match the commit message prefix)

icu4c/source/data/unidata/generate.sh
roubert: Done.

icu4c/source/data/unidata/generate.sh
roubert: I had simply assumed that this was run hermetically, the idea that the cache might be shared between different operating system versions didn't even cross my mind.

PR #2273, opened 2022-12-16
ICU-12811 Make CI jobs for Maven run serially

echeran:
There are 2 CI jobs that run a Maven command at the root.  The jobs are configured to persist any downloaded Maven artifact dependencies, and cache them (save them at the end of the job, load the cache at the beginning of the next instantiation).

We want the cache to work properly to make the jobs more resilient in case of ephemeral network connections, etc., like what happened in this [recent workflow run](https://github.com/unicode-org/icu/actions/runs/3700019863/jobs/6268025244#logs).

But because these jobs run in parallel, there is probably a race condition.  One job invokes a target (`verify`) that is later phase in the Maven lifecycle than the other job's target phase (`test`). This means the first jobs required set of dependencies is a superset of the other.

We could have a separate cache for each job, or we could have the job that needs fewer dependencies always execute after the other, and then the latter job can reuse the former job's cache.  This PR takes the latter approach.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-12811

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -63,6 +63,10 @@ jobs:
   java8-icu4j-test-maven:
     name: Run unit tests with Maven using JDK 8
     runs-on: ubuntu-latest
+    # Make this unit test target job depend on a later phase target job to prevent race condition when
+    # trying to persist the Maven cache to the Github cache, knowing that artifacts needed for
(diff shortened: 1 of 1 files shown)


markusicu (approved):
looks plausible, and Elango explained a bit more about it -- tnx

PR #164, opened 2018-09-21
WIP ICU-12968 remove icudata.jar from icu4j tree

srl295:
- improve build of data
- :construction: does NOT work on windows
- add a convenience script on the icu4j side
- probably some other holes here

##### Checklist

- [x] Issue filed at https://unicode-org.atlassian.net/browse/ICU-12968
- [x] Update PR title to include Issue number
- [ ] Issue accepted (kind of)
- [ ] Documentation is changed or added (kind of)

PR #2275, opened 2022-12-19
ICU-12811 Replace local-maven-repo with data jar contents

echeran:
It's better to review this PR commit-by-commit, since the addition of the data file contents is very numerous and noisy (and makes Github's UI slow down a lot).

I think [cldr-icu-readme.txt](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/cldr-icu-readme.txt) is the file that contains the stpes for integrating new CLDR data versions into ICU, so it would need to be updated here in this PR, too.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-12811


icu4j/maven-build/extract-data-files.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Copyright (C) 2016 and later: Unicode, Inc. and others.
```
mihnita, line 2: Nitpick: 2023

icu4j/maven-build/extract-data-files.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
mihnita, line 1: Nitpick: to be complete should also have a Windows script.

icu4c/source/data/cldr-icu-readme.txt
```diff
@@ -307,12 +314,24 @@ make icu4j-data-install
 cd $ICU4C_DIR/source/test/testdata
 make icu4j-data-install
 
+# 12c. Replace the extracted {main, test} data files in the Maven build
+
+cd $ICU4J_ROOT/maven-build
+sh extract-data-files.sh
```
mihnita, line 320: Nitpick :-)
```
./extract-data-files.sh
```

mihnita (commented):
Thank you.
Only nitpicks.

But my main worry is the GitHub inability of dealing with many changed files.
I was unable to see the 2 commits that updated `.res` files (even when looking at one commit at the time)
I got an "angry unicorn" page
https://user-images.githubusercontent.com/95055527/200951443-84459209-d082-4f08-ae33-59632b92da30.png

I had to `git clone` your repo locally, `git co also-expand-data-jars`, and review there.
That is quite inconvenient...

icu4c/source/data/cldr-icu-readme.txt
echeran: Done.

icu4j/maven-build/extract-data-files.sh
echeran: Done.

icu4j/maven-build/extract-data-files.sh
echeran: This script is just to simplify `cldr-icu-readme.txt`, which is a set of steps that need to be run by a person (who can double check discrepencies, etc.), and those steps are written for a POSIX on *nix environment. So I think this is okay for now (until there exists an easy cross-platform solution to do things).

echeran:
> Thank you. Only nitpicks.

Thanks for the review. I made changes to address the comments that were addressable, and comments on the others.

I also pushed a commit to remove more configs (ones that I had missed previously) that are now obsolete after the removal of the local-maven-repo setup.

> But my main worry is the GitHub inability of dealing with many changed files. I was unable to see the 2 commits that updated `.res` files (even when looking at one commit at the time) I got an "angry unicorn" page https://user-images.githubusercontent.com/95055527/200951443-84459209-d082-4f08-ae33-59632b92da30.png
> 
> I had to `git clone` your repo locally, `git co also-expand-data-jars`, and review there. That is quite inconvenient...

All of the non-uncommon CLDR data integration PRs end up like this, too (they're updating all the same files). But for specific purposes of PR reviews, you can use [File Filter drop down](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/filtering-files-in-a-pull-request) to ignore all the binary files if they have a distinct file extension. FWIW, [here is what this PR looks like](https://github.com/unicode-org/icu/pull/2275/files?file-filters%5B%5D=.jar&file-filters%5B%5D=.pom&file-filters%5B%5D=.sh&file-filters%5B%5D=.txt&file-filters%5B%5D=.xml&show-deleted-files=true&show-viewed-files=true) with such filters applied.

mihnita (approved):

srl295:
@mihnita @echeran late to the party but … it's a good step, but why even have the .res files in the repo? Might consider just rebuilding from the icu4c source similar to https://github.com/unicode-org/icu/pull/164#issuecomment-424872304 - and aggressively caching the result for CI.

PR #2279, opened 2023-01-09
ICU-22220 integrate CLDR release-43-m0 to ICU main for 73

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22220

This integrates to ICU main data from CLDR release-43-m0 (after the start of general data submission but without any of the result from Survey Tool data collection), _**except**_ (per Markus request) the adding to root of 4 exemplar city names per part of [CLDR PR-2583](https://github.com/unicode-org/cldr/pull/2583), which will be handled as a separate PR. Some notes:
- Before Survey Tool data collection, CLDR reset the draft status for most personName data (except in `root`, `en`, and some regional locales) to draft="provisional"; thus it is not picked up by CLDR. Consequently, some unit tests had to be temporarily reverted to use root behavior..
- CLDR has changed some aspects of the spec for PersonName behavior. While some of corresponding ICU4J code was updated so it would build, other changes related to name spacing have not been made in ICU yet because they cannot be tested until the real non-English data is available again. That should happen with the next CLDR integration.
- An early version of the CLDR 43-m0 data had very aggressive minimization of data that was expected to be inherited. This demonstrated some problems with fallback data loading in ICU4C DateFormatSymbols,  which has been updated. Subsequently the minimization was found to have some problems and was reverted, with planned less-aggressive minimization still coming in CLDR 43. This reversion was done before ICU4J was tested with the more aggressive version, so any fallback issues in ICU4J DateFormatSymbols have not yet been fixed.
- There are some round-trip problems with the updated CLDR Cyrillic-Latin transform; the corresponding tests are currently skipped with a logKnownIssue on a CLDR ticket about the issue.
- CLDR has moved all locales formerly in seed to common. However, el_POLYTON is not at a sufficient coverage level for ICU, but the ICU build normally picks up all variants of a given language code. I needed to add special code  to exclude this (in the future the CLDR-to-ICU build will use CLDR coverage levels as the determinant for whether a locale is picked up by ICU).
- Also regenerate the files in maven-build/local-maven-repo/, deleting the generated md5/sha1 files, and restoring the copyright notice to the generated pom/xml files.


macchiati (approved):

pedberg-icu:
@macchiati Can you re-stamp please? The only change was to also add the re-generated data jars in icu4j/maven-build/local-maven-repo/ needed for the new CI test for mavenized ICU4J.

macchiati (approved):

PR #2287, opened 2023-01-18
ICU-22100 Improve Japanese phrase breaking performance

allensu05:
1. Use UnicodeString directly instead of type casing between char16_t* and UnicodeString
2. Because Hashtable.geti(key) returns 0 when key doesn't exist, it doesn't need to check whether the key exists.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22100

### DIFF
--- a/icu4c/source/common/mlbe.cpp
+++ b/icu4c/source/common/mlbe.cpp
@@ -34,20 +34,6 @@ MlBreakEngine::~MlBreakEngine() {}
 
 namespace {
     const char16_t INVALID = u'|';
-    const int32_t MAX_FEATURE = 13;
-    const int32_t MAX_FEATURE_LENGTH = 11;
-
-    void concatChar(const char16_t *str, const UChar32 *arr, int32_t length, char16_t *feature, UErrorCode &status) {
-        if (U_FAILURE(status)) {
-            return;
-        }
-        UnicodeString result(str);
-        for (int i = 0; i < length; i++) {
-            result.append(arr[i]);
-        }
-        U_ASSERT(result.length() < MAX_FEATURE_LENGTH);
-        result.extract(feature, MAX_FEATURE_LENGTH, status);  // NUL-terminates
-    }
 }
 
 int32_t MlBreakEngine::divideUpRange(UText *inText, int32_t rangeStart, int32_t rangeEnd,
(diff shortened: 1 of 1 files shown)


icu4c/source/common/mlbe.cpp
```diff
@@ -144,96 +131,75 @@ int32_t MlBreakEngine::divideUpRange(UText *inText, int32_t rangeStart, int32_t
...
+        listLength++;
     }
+
     int32_t score = fNegativeSum;
     for (int32_t j = 0; j < listLength; j++) {
-        UnicodeString key(featureList[j]);
-        if (fModel.containsKey(key)) {
-            score += (2 * fModel.geti(key));
-        }
+        score += (2 * fModel.geti(featureList[j]));
```
FrankYFTang, line 202: Please add a comment above this line.
```
// Because when the key doesn't exist geti(key) returns 0  and 2 * 0 = 0,
//  the following line will not be changed, so we can skip checking fModel
// includes key featureList[j] or not.
```

FrankYFTang (approved):

FrankYFTang (changes requested):
please add that comment line

icu4c/source/common/mlbe.cpp
allensu05: Done

icu4c/source/common/mlbe.cpp
```diff
@@ -144,96 +131,87 @@ int32_t MlBreakEngine::divideUpRange(UText *inText, int32_t rangeStart, int32_t
...
-    const UChar32 w3 = elementList[2];
-    const UChar32 w4 = elementList[3];
-    const UChar32 w5 = elementList[4];
-    const UChar32 w6 = elementList[5];
-
-    length = 1;
-    if (w1 != INVALID) {
-        arr[0] = w1;
-        concatChar(u"UW1:", arr, length, featureList[listLength++], status);
+    UnicodeString featureList[MAX_FEATURE];
```
FrankYFTang, line 138: why do we need to allocate MAX_FEATURE UnicodeString on the stack and then later on go through a loop to calculate the score by looping through it. How about just allocate one as feature and reuse it?
```
UnicodeString feature;
int32_t score = fNegativeSum;

if (elementList[0] != INVALID) {
    // Because when the key doesn't exist geti(key) returns 0  and 2 * 0 = 0,
    // the following line will not be changed, so we can skip checking fModel
    // includes key featureList[j] or not.
    score += (2 * fModel.geti(feature.setTo(u"UW1:", 4).append(elementList[0])));
}
 if (elementList[1] != INVALID) {
    score += (2 * fModel.geti(feature.setTo(u"UW2:", 4).append(elementList[1])));
}
.... and similar to the rest of if
```

allensu05: Done. Thanks

FrankYFTang:
I fixed and landed the MacOS Latest Build + WarningAsError build breakage in 
https://github.com/unicode-org/icu/pull/2291
Please rebase to see the problem is fixed or not.

FrankYFTang (approved):

PR #2289, opened 2023-01-19
ICU-12725 Update u_isIDStart and u_isIDPart to TR31

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-12725

1. Adjust the API doc in the uchar.h to refer to TR31 
2. Move u_isIDStart and u_isIDPart to uprops.cpp
3. Add tests for both C and Java API
4. Fix the C and Java implementation.
5. Adjust the dependency

### DIFF
--- a/icu4c/source/common/uchar.cpp
+++ b/icu4c/source/common/uchar.cpp
@@ -304,30 +304,6 @@ u_ispunct(UChar32 c) {
     return (UBool)((CAT_MASK(props)&U_GC_P_MASK)!=0);
 }
 
-/* Checks if the Unicode character can start a Unicode identifier.*/
-U_CAPI UBool U_EXPORT2
-u_isIDStart(UChar32 c) {
-    /* same as u_isalpha() */
-    uint32_t props;
-    GET_PROPS(c, props);
-    return (UBool)((CAT_MASK(props)&(U_GC_L_MASK|U_GC_NL_MASK))!=0);
-}
-
-/* Checks if the Unicode character can be a Unicode identifier part other than starting the
- identifier.*/
-U_CAPI UBool U_EXPORT2
-u_isIDPart(UChar32 c) {
-    uint32_t props;
-    GET_PROPS(c, props);
-    return (UBool)(
-           (CAT_MASK(props)&
-            (U_GC_ND_MASK|U_GC_NL_MASK|
-             U_GC_L_MASK|
-             U_GC_PC_MASK|U_GC_MC_MASK|U_GC_MN_MASK)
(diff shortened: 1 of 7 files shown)


FrankYFTang:
Ready for Review

icu4c/source/common/uchar.cpp
```diff
@@ -304,30 +304,6 @@ u_ispunct(UChar32 c) {
...
-    GET_PROPS(c, props);
-    return (UBool)(
-           (CAT_MASK(props)&
-            (U_GC_ND_MASK|U_GC_NL_MASK|
-             U_GC_L_MASK|
-             U_GC_PC_MASK|U_GC_MC_MASK|U_GC_MN_MASK)
-           )!=0 ||
-           u_isIDIgnorable(c));
-}
-
```
richgillam, line 330: I take it you moved these functions from `uchar.cpp` to `uprops.cpp` because that's where `u_hasBinaryProperty()` happens to be defined?  Why move the functions to a new file instead of adding an `#include`?  Are there any side effects from doing this?

icu4c/source/common/uprops.cpp
```diff
@@ -423,6 +423,19 @@ u_hasBinaryProperty(UChar32 c, UProperty which) {
...
+    return u_hasBinaryProperty(c, UCHAR_ID_START);
+}
+
+/* Checks if the Unicode character can be a Unicode identifier part other than starting the
+ identifier.*/
+U_CAPI UBool U_EXPORT2
+u_isIDPart(UChar32 c) {
+    return u_hasBinaryProperty(c, UCHAR_ID_CONTINUE);
+}
+
```
richgillam, line 438: So we already had the right property information in the property database?  It was just a matter of actually using it?

richgillam (approved):
Maybe I should have reviewed this more closely, but it all seems to make sense...

icu4c/source/common/uchar.cpp
FrankYFTang: if we do that we need to change the dependency graph and cause any code depends on the old cpp to increase more dependent .o

icu4c/source/common/uprops.cpp
FrankYFTang: yes

FrankYFTang:
@markusicu any feedback?

FrankYFTang:
@markusicu  in case you have concern please reopen https://unicode-org.atlassian.net/browse/ICU-12725

PR #2291, opened 2023-01-21
ICU-22251 Move sprintf to snprintf.

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22251


richgillam (approved):
This all looks good.  Looks mechanical, except for the calculation of the buffer size.  I didn't review all of those carefully, but the ones I did all looked right.

richgillam (approved):
New stuff looks good too.

roubert (dismissed):
I really don't think that you should be hardcoding any of these numbers, that wouldn't get us any real improved memory safety, it would just move the places where things could go wrong.

Instead, existing code like this:

```
char tmp[16];
sprintf(tmp, "%d", value);
```

Should be replaced with new code like this:

```
char tmp[16];
snprintf(tmp, sizeof tmp, "%d", value);
```

Or, when that's not possible, some kind of logic / argument passing that ensures that the value used is ultimately calculated by the compiler from the actual buffer size.

FrankYFTang:
> I really don't think that you should be hardcoding any of these numbers, that wouldn't get us any real improved memory safety, it would just move the places where things could go wrong.
> 
> Instead, existing code like this:
> 
> ```
> char tmp[16];
> sprintf(tmp, "%d", value);
> ```
> 
> Should be replaced with new code like this:
> 
> ```
> char tmp[16];
> snprintf(tmp, sizeof tmp, "%d", value);
> ```
> 
> Or, when that's not possible, some kind of logic / argument passing that ensures that the value used is ultimately calculated by the compiler from the actual buffer size.

FrankYFTang:
> Should be replaced with new code like this:
> 
> ```
> char tmp[16];
> snprintf(tmp, sizeof tmp, "%d", value);
> ```



> ```
> 
> ```
> 
> 
>     
>   
> 
> Should be replaced with new code like this:
> 
> 
> ```
> char tmp[16];
> snprintf(tmp, sizeof tmp, "%d", value);
> ```

DONE

icu4c/source/test/cintltst/udatatst.c
```diff
@@ -1157,7 +1157,7 @@ static void TestICUDataName()
     }
 
     /* Only major number is needed. */
-    sprintf(expectDataName, "%s%d%c",
+    snprintf(expectDataName, 20, "%s%d%c",
```
richgillam, line 1160: Is there a reason why you didn't use `sizeof` here?

icu4c/source/test/cintltst/udatatst.c
```diff
@@ -1626,7 +1626,7 @@ TestSwapCase(UDataMemory *pData, const char *name,
 
 static void U_CALLCONV
 printErrorToString(void *context, const char *fmt, va_list args) {
-    vsprintf((char *)context, fmt, args);
+    vsnprintf((char *)context, 100, fmt, args);
```
richgillam, line 1629: Is there a reason you didn't use `sizeof` here?

icu4c/source/test/intltest/alphaindextst.cpp
```diff
@@ -698,7 +698,7 @@ void AlphabeticIndexTest::TestJapaneseKanji() {
     int32_t overflowIndex = immIndex->getBucketCount() - 1;
     for(int32_t i = 0; i < UPRV_LENGTHOF(kanji); ++i) {
         char msg[40];
-        sprintf(msg, "kanji[%d]=U+%04lX in overflow bucket", (int)i, (long)kanji[i]);
+        snprintf(msg, 40, "kanji[%d]=U+%04lX in overflow bucket", (int)i, (long)kanji[i]);
```
richgillam, line 701: `sizeof(msg)`?

richgillam (dismissed):
Looks good, and I think Fredrik's recommendation is a good one and makes the code more readable and bulletproof.  But I think you missed a couple spots...

icu4c/source/test/cintltst/udatatst.c
FrankYFTang: somehow I missed this

icu4c/source/test/cintltst/udatatst.c
FrankYFTang: this is much hader to get the size and I cannot simply use sizeof here

FrankYFTang:
PTAL

icu4c/source/test/cintltst/udatatst.c
richgillam: Well, it's unit-test code, so I'm not terribly concerned.

richgillam (approved):
It all looks fine.  Cut and print.

FrankYFTang:
> I really don't think that you should be hardcoding any of these numbers, that wouldn't get us any real improved memory safety, it would just move the places where things could go wrong.
> 
> Instead, existing code like this:
> 
> ```
> char tmp[16];
> sprintf(tmp, "%d", value);
> ```
> 
> Should be replaced with new code like this:
> 
> ```
> char tmp[16];
> snprintf(tmp, sizeof tmp, "%d", value);
> ```
> 
> Or, when that's not possible, some kind of logic / argument passing that ensures that the value used is ultimately calculated by the compiler from the actual buffer size.

done

PR #2293, opened 2023-01-25
ICU-22226 Fix Calendar.getFirstDayOfWeek to honor -u-fw

mihnita:
This CL fixes only ICU-22226, not [ ICU-11870](https://unicode-org.atlassian.net/browse/ICU-11870)

Properly fixing ICU-11870 (the `DateFormat.getTimeInstance` honoring `hc`) does require a redesign (and design document)

This is independent, and a simple bug, does not require design docs.

##### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22226

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -3948,6 +3948,29 @@ Calendar::setWeekData(const Locale& desiredLocale, const char *type, UErrorCode&
         } else {
             status = U_INVALID_FORMAT_ERROR;
         }
+
+        // Check if the locale has a "fw" u extension and we honor it if present.
+        // And we don't change the overal status, as the presence / lack of "fw" is not an error.
+        UErrorCode fwStatus = U_ZERO_ERROR;
+        char fwExt[ULOC_FULLNAME_CAPACITY] = "";
+        desiredLocale.getKeywordValue("fw", fwExt, ULOC_FULLNAME_CAPACITY, fwStatus);
+        if (U_SUCCESS(fwStatus)) {
+            if (uprv_strcmp(fwExt, "sun") == 0) {
+                fFirstDayOfWeek = UCAL_SUNDAY;
+            } else if (uprv_strcmp(fwExt, "mon") == 0) {
+                fFirstDayOfWeek = UCAL_MONDAY;
+            } else if (uprv_strcmp(fwExt, "tue") == 0) {
+                fFirstDayOfWeek = UCAL_TUESDAY;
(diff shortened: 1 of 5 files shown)


icu4c/source/i18n/calendar.cpp
```diff
@@ -3948,6 +3948,29 @@ Calendar::setWeekData(const Locale& desiredLocale, const char *type, UErrorCode&
         } else {
             status = U_INVALID_FORMAT_ERROR;
         }
+
+        // Check if the locale has a "fw" u extension and we honor it if present.
+        // And we don't change the overal status, as the presence / lack of "fw" is not an error.
+        UErrorCode fwStatus = U_ZERO_ERROR;
+        char fwExt[ULOC_FULLNAME_CAPACITY] = "";
+        desiredLocale.getKeywordValue("fw", fwExt, ULOC_FULLNAME_CAPACITY, fwStatus);
+        if (!U_FAILURE(fwStatus)) {
```
FrankYFTang, line 3957: how about U_SUCCESS() instead of !U_FAILURE() ?

mihnita: Done.

---

I know there are (used to be?) warnings states.

So in my mind:
```
if (U_SUCCESS) { ... } => SUCCESS
if (!U_FAILURE) { ... } => SUCCESS || WARNING
```
There are other instances of code doing `if (!U_FAILURE) { ... }`

But it looks like the warning levels were deprecated (protected with `#ifdef`)` and now `U_SUCCESS === ! U_FAILURE`

FrankYFTang (dismissed):
LGTM

FrankYFTang:
please squashed

mihnita:
> LGTM

Thank you.

Can you also accept the ICU-22226 ticket?
It was initially reported and accepted, then folded into [ICU-11870](https://unicode-org.atlassian.net/browse/ICU-11870) and "un-accepted" .

But although they sound similar on the surface, they don't require the same fix and extended changes.
This is a simple bug fix, the `hc` part indeed requires bigger changes, and a design document.

So I think the two bugs should be "untangled".

I will update both bugs with this info.

Thank you,
Mihai

FrankYFTang (approved):
LGTM

mihnita:
Thanks for the LGTM.

Sorry, still blocked with "Jira ticket ICU-22226 is not accepted; it has status New"
I can't change it to Accepted.

And even after that I will have no right to merge it.
So, can you please change the issue to Accepted then merge? If you can't I can ask Markus.

Thank you,
M.

PR #2295, opened 2023-01-28
ICU-22249 Fixed endless loop in ICUResourceBundle when you ask for a …

richgillam:
…locale with a nonstandard parent and that locale

is also the system default locale.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22249

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
@@ -1396,16 +1396,20 @@ public  class ICUResourceBundle extends UResourceBundle {
 
             // fallback to locale ID parent
             if(b == null){
+                OpenType localOpenType = openType;
+                if (openType == OpenType.LOCALE_DEFAULT_ROOT && localeName.equals(defaultID)) {
+                    localOpenType = OpenType.LOCALE_ROOT;
+                }
                 String origLocaleName = (origLocaleID != null) ? origLocaleID : localeName;
                 String fallbackLocaleID = getParentLocaleID(localeName, origLocaleName, openType);
                 if (fallbackLocaleID != null) {
-                    b = instantiateBundle(baseName, fallbackLocaleID, origLocaleName, defaultID, root, openType);
+                    b = instantiateBundle(baseName, fallbackLocaleID, origLocaleName, defaultID, root, localOpenType);
                 }else{
-                    if(openType == OpenType.LOCALE_DEFAULT_ROOT &&
+                    if(localOpenType == OpenType.LOCALE_DEFAULT_ROOT &&
                             !localeIDStartsWithLangSubtag(defaultID, localeName)) {
                         // Go to the default locale before root.
-                        b = instantiateBundle(baseName, defaultID, null, defaultID, root, openType);
-                    } else if(openType != OpenType.LOCALE_ONLY && !rootLocale.isEmpty()) {
+                        b = instantiateBundle(baseName, defaultID, null, defaultID, root, localOpenType);
+                    } else if(localOpenType != OpenType.LOCALE_ONLY && !rootLocale.isEmpty()) {
                         // Ultimately go to root.
                         b = ICUResourceBundle.createBundle(baseName, rootLocale, root);
                     }
--- a/icu4j/main/tests/translit/src/com/ibm/icu/dev/test/translit/TransliteratorTest.java
+++ b/icu4j/main/tests/translit/src/com/ibm/icu/dev/test/translit/TransliteratorTest.java
@@ -2586,7 +2586,6 @@ public class TransliteratorTest extends TestFmwk {
         }
     }
 
-
     static final String[][] registerRules = {
         {"Any-Dev1", "x > X; y > Y;"},
         {"Any-Dev2", "XY > Z"},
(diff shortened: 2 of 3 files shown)


richgillam:
There's a fairly full explanation of the problem and solution in the JIRA ticket...

icu4j/main/tests/translit/src/com/ibm/icu/dev/test/translit/TransliteratorTest.java
```diff
@@ -2585,6 +2585,16 @@ public void TestScriptAllCodepoints(){
             }
         }
     }
+    
+    @Test
+    public void TestResourceBundleCrash() {
+        ULocale oldDefaultLocale = ULocale.getDefault();
+		ULocale.setDefault(ULocale.forLanguageTag("nb"));
```
mihnita, line 2592: Suggestion 1: also tests for "nb", "ht" (maps to "fr-HT") and "hi-Latn" (maps to "en-IN").
Maybe "mo", in case it ever comes back :-)
Might be an array and iterate, I don't think it matters which one fails, as long as we detect a regression.

Suggestion 2: also test `Transliterator.getDisplayName("Any-Latin", ULocale.GERMAN);`
The second locale can be anything that does not have a .res file, so no ENGLISH or US.
Can be the same as default, but I used GERMAN to show that it is not in the same class with the failing locales
(where the parent.language <> getDefault.language)

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1396,16 +1396,20 @@ public ICUResourceBundle load() {
 
             // fallback to locale ID parent
             if(b == null){
+                OpenType newOpenType = openType;
+                if (openType == OpenType.LOCALE_DEFAULT_ROOT && localeName.equals(defaultID)) {
```
mihnita, line 1400: > localeName.equals(defaultID)

Is there a chance that this test fails for locales that have different string ID, but they are in fact the same locale?
Like `he` / `iw`, or `id` / `in` ?
Or even `FR` <> `fr` or `it_IT` <> `it-IT`?

In other words, are we sure things are normalized / canonicalized somewhere, before we get here?

mihnita (commented):
Thank you,
Mihai

PS: no rush to do anything tonight or over the week-end from my side :-)

markusicu (commented):
lgtm except for what Mihai said...

icu4j/main/tests/translit/src/com/ibm/icu/dev/test/translit/TransliteratorTest.java
richgillam: I decided to move the test into ICUResourceBundleTest, since the thing that's failing is the resource bundle-- it doesn't have anything to do with the transliterator, and it fails before we even try to get a specific resource out of the resource bundle.

I tried the locale IDs you suggested above: `nb` and `nn` fail, as we know.  `ht` did too.  For some reason, `hi-Latn` doesn't-- I'm not sure why, but it might be because the script code is changing the lookup pattern.  `mo` also didn't cause any problems.  But I changed the unit test to test all of these.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
richgillam: I tried all these things, and they worked fine.  The code that deals with obsolete language tags is incorporated into the code that normalizes the locale IDs for things like case and punctuation, so they weren't affected by the problem.  I could add unit tests to prove this, but decided against it.

icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
```diff
@@ -1396,16 +1396,20 @@ public ICUResourceBundle load() {
 
             // fallback to locale ID parent
             if(b == null){
+                OpenType newOpenType = openType;
```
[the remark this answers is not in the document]
richgillam: I like that.  Done.

icu4j/main/tests/translit/src/com/ibm/icu/dev/test/translit/TransliteratorTest.java
mihnita: > For some reason, hi-Latn doesn't-- I'm not sure why

I think it is because it maps to en-IN, and there is an `en.res` file.
This is one of the experiments I did. When I deleted that res file then `hi-Latn` also failed.

mihnita: > `mo` also didn't cause any problems.

Yes, it does not cause any problems today.

I know that the treatment of `mo` changed several times in the last 10 years.
As an experiment I've added a `ro` as a parent of `mo`, and then `mo` failed.

So the suggestion to add it to the test was as a abuse of caution :-), to detect a possible addition of `mo` => `ro`

mihnita: LGTM.

I will approve, and if you want to also add a a test for `Transliterator.getDisplayName("Any-Latin", ...);` fine.
If not, that's fine too, no need to explain why not :-)

Thanks,
Mihai

mihnita (approved):
Thank you very much!
Mihai

markusicu (commented):
lgtm tnx! pse squash

markusicu:
> lgtm tnx! pse squash

ping

richgillam:
>>lgtm tnx! pse squash

>ping

Sorry, it's been a crazy week...

markusicu (approved):
lgtm

PR #2300, opened 2023-02-03
ICU-21833 Replace UChar with char16_t in all C++ code.

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21833


markusicu (approved):
(largely rs) lgtm, spot-check ok, tnx!

aheninger:
Spot checking, this looks good to me.

Has there been any discussion of moving the plain C minimum up to C11, which also supports char16_t, and would let us finish this job? I don't think there would be any binary compatibility issues.

There are a few C++ headers that look like they were missed
  - common/normalizer2impl.h
  - common/lstmbe.h
  - common/dictionarydata.h

There are also several places where a C style function is declared in an implementation header and defined in a .cpp file, where the definition is updated but the declaration is not. For example,

i18n/rbt_pars.h:352
```
U_CAPI int32_t
utrans_stripRules(const UChar *source, int32_t sourceLen, UChar *target, UErrorCode *status);
```
i18n/rbt_pars.cpp:1675
```
U_CAPI int32_t
utrans_stripRules(const char16_t *source, int32_t sourceLen, char16_t *target, UErrorCode *status) {
```

The same thing happens with plain C API functions, but for them the mismatch seems unavoidable. For internal functions (declared in non-public headers) updating the declaration is probably safe.

I think it's fine to merge this PR as is.

roubert:
> I think it's fine to merge this PR as is.

OK, let's do that then to get the bulk of the changes out of the way, then we can fix what's found later on a case-by-case basis.

PR #1884, opened 2021-09-27
ICU-21581 Creates a new workflow to be activated upon merge into main or

gnrunge:
maintenance branches only and adds GHA CI automation of BRS task 'Test ICU4J
with only little-endian ICU4C data', cf.:
https://unicode-org.github.io/icu/processes/release/tasks/integration.html#test-icu4j-with-only-little-endian-icu4c-data.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21581
- [ ] Required: The PR title must be prefixed with a JIRA Issue number. 
- [ ] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [ ] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [ ] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2302, opened 2023-02-04
ICU-22257 Add "Test ICU4J with only little-endian ICU4C data"

FrankYFTang:
Automate "Test ICU4J with only little-endian ICU4C data" as stated in https://unicode-org.github.io/icu/processes/release/tasks/integration.html#test-icu4j-with-only-little-endian-icu4c-data

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22257

### DIFF
--- a/.ci-builds/.azure-exhaustive-tests.yml
+++ b/.ci-builds/.azure-exhaustive-tests.yml
@@ -71,4 +71,34 @@ jobs:
         CC: clang
         CXX: clang++
 #-------------------------------------------------------------------------
-
+# Based on https://unicode-org.github.io/icu/processes/release/tasks/integration.html#test-icu4j-with-only-little-endian-icu4c-data
+- job: ICU4J_with_only_little_endian_ICU4C_data
+  displayName: 'CJ: ICU4J Linux Clang Tests little-endian ICU4C data (Ubuntu 18.04)'
+  timeoutInMinutes: 50
+  pool:
+    vmImage: 'ubuntu-18.04'
+    demands: ant
+  steps:
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
+    - script: |
+        cd icu4c/source && ICU_DATA_BUILDTOOL_OPTS=--include_uni_core_data ./runConfigureICU Linux && make -j 6 check
+      displayName: 'Build ICU4C with ICU_DATA_BUILDTOOL_OPTS'
+      env:
+        CC: clang
+        CXX: clang++
+    - script: |
+        mkdir -p /tmp/icu4j_data_test && cp icu4c/source/data/out/tmp/icudt*l.dat /tmp/icu4j_data_test
+      displayName: 'Copy ICU4C data'
+    - script: |
+        echo "test ICU4J" && cd icu4j && ant init && ant -Dicu4c.data.path=/tmp/icu4j_data_test check
+      displayName: 'Build and Test ICU4J with little-endian data'
+      env:
+        BUILD: ICU4J
+# exit with a non-zero status in order to make this step show as a red X in the UI.
+    - script: |
+        [ -d icu4j/out/junit-results ] && cd icu4j && cat `find out/junit-results -name "*.txt" -exec grep -l FAILED {} \;` && exit 1
+      condition: failed() # only run if the build fails.
+      displayName: 'List failures (if any)'
+#-------------------------------------------------------------------------


FrankYFTang:
see the success execution on https://dev.azure.com/ms/icu/_build/results?buildId=424795&view=logs&j=c4e2fbc5-02b9-52ef-7a41-f5c664de1614

.ci-builds/.azure-exhaustive-tests.yml
```diff
@@ -71,4 +71,34 @@ jobs:
         CC: clang
         CXX: clang++
 #-------------------------------------------------------------------------
-
+# Based on https://unicode-org.github.io/icu/processes/release/tasks/integration.html#test-icu4j-with-only-little-endian-icu4c-data
+- job: ICU4J_with_only_little_endian_ICU4C_data
+  displayName: 'CJ: ICU4J Linux Clang Tests little-endian ICU4C data (Ubuntu 18.04)'
+  timeoutInMinutes: 50
+  pool:
+    vmImage: 'ubuntu-18.04'
```
markusicu, line 79: why such an old OS?

.ci-builds/.azure-exhaustive-tests.yml
```diff
@@ -71,4 +71,34 @@ jobs:
...
+    demands: ant
+  steps:
+    - checkout: self
+      lfs: true
+      fetchDepth: 10
```
markusicu, line 84: 1 ?

markusicu (commented):
thanks!

.ci-builds/.azure-exhaustive-tests.yml
FrankYFTang: I keep it as the same as the value of other build in this file above. Moving to a new version should be done in a separate ticket with the rest of the build in this file all together.

.ci-builds/.azure-exhaustive-tests.yml
FrankYFTang: Again, it is "fetchDepth: 10" in other builds in this file. I want to keep it as the same as the rest unless there is a good reason to change from it.

FrankYFTang:
As someone who is not very familar with the azure and yml file, I just copy and paste the setting from others builds config on top of that file and change the part which make sense and keep other value unchange. In this way we know the extends of the differences between the newly added one to the other pre-existing builds. Using other value would be a total different discussion to the task. My understanding is while Norbert run this task manaually before, he is not target toward a particular version of OS or need to use a newer OS in his instruction, therefore I do not believe that should be a requirement for this set up to be differ from the rest of the pre-existing configuraiton in this file. If you feel we should use a newer version of OS to test, I think we should open a different ticket and change all the version under that ticket instead. But that will be yet a different discussio and require a differrent PR since such change is not guarantee without new issues. And I do not believe such new issues should be mixed with this simple task.

.ci-builds/.azure-exhaustive-tests.yml
```diff
@@ -71,4 +71,34 @@ jobs:
...
+        cd icu4c/source && ICU_DATA_BUILDTOOL_OPTS=--include_uni_core_data ./runConfigureICU Linux && make -j 6 check
+      displayName: 'Build ICU4C with ICU_DATA_BUILDTOOL_OPTS'
+      env:
+        CC: clang
+        CXX: clang++
+    - script: |
+        mkdir -p /tmp/icu4j_data_test && cp icu4c/source/data/out/tmp/icudt*l.dat /tmp/icu4j_data_test
+      displayName: 'Copy ICU4C data'
+    - script: |
+        echo "test ICU4J" && cd icu4j && ant init && ant -Dicu4c.data.path=/tmp/icu4j_data_test && ant localespiCheck
```
markusicu, line 95:
```suggestion
        echo "test ICU4J" && cd icu4j && ant init && ant -Dicu4c.data.path=/tmp/icu4j_data_test check
```
Let's change the ant test invocation to what the BRS task docs say.

markusicu (commented):
Frank, you convinced me. I was just waiting for @echeran's feedback.

But now I compared your build script with the docs and found a discrepancy. Please fix.

markusicu (approved):
lgtm tnx

markusicu:
I don't see the new check in the pending list. Am I missing something?

FrankYFTang:
It only trigger on the check if I put in the following in the comment

/azp run CI-Exhaustive

FrankYFTang:
ok, it seems that line to trigger the run cannot contains any other text

echeran (approved):
LGTM. Since most of this is just an appropriate adaptation of other similar jobs in the same workflow file, it seems fine as it is.

markusicu:
Uhm... this is a little embarrassing...

I found @gnrunge's spreadsheet of BRS tasks that could be automated. (His Google account owned it, so I now made [a copy owned by Unicode](https://docs.google.com/spreadsheets/d/13-U_YRmEUmLyK-NiONkV1EZC-YeJvqCqzCoO7iIZy9A/edit#gid=0).)

He had already marked this task here as “Done with GHA, post-merge workflow”!

So now I looked, and I found that in fact he added a GHA in time for ICU 70 which does the same thing. See [ICU-21581](https://unicode-org.atlassian.net/browse/ICU-21581) https://github.com/unicode-org/icu/pull/1884 (2021-sep).

We just didn't have this documented in the BRS tasks, nor in the BRS spreadsheet :-(

Should we remove this new but redundant Azure CI check in favor of the pre-existing GHA check?

We should keep the documentation saying that it's been automated, but update it to say ICU 70 not 73.

@echeran @FrankYFTang

markusicu:
https://github.com/unicode-org/icu/blob/main/.github/workflows/icu_merge_ci.yml#L16

PR #2303, opened 2023-02-06
ICU-22260 Allow relative datetime formatting without context adjustment with UCONFIG_NO_BREAK_ITERATION

hsivonen:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22260

### DIFF
--- a/icu4c/source/i18n/reldatefmt.cpp
+++ b/icu4c/source/i18n/reldatefmt.cpp
@@ -12,7 +12,7 @@
 
 #include "unicode/reldatefmt.h"
 
-#if !UCONFIG_NO_FORMATTING && !UCONFIG_NO_BREAK_ITERATION
+#if !UCONFIG_NO_FORMATTING
 
 #include <cmath>
 #include <functional>
@@ -701,7 +701,7 @@ static UBool getDateTimePattern(
     return getStringByIndex(topLevel.getAlias(), dateTimeFormatOffset, result, status);
 }
 
-template<> 
+template<>
 const RelativeDateTimeCacheData *LocaleCacheKey<RelativeDateTimeCacheData>::createObject(const void * /*unused*/, UErrorCode &status) const {
     const char *localeId = fLoc.getName();
     LocalUResourceBundlePointer topLevel(ures_open(nullptr, localeId, &status));
@@ -761,6 +761,7 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(UErrorCode& status) :
         fStyle(UDAT_STYLE_LONG),
         fContext(UDISPCTX_CAPITALIZATION_NONE),
         fOptBreakIterator(nullptr) {
+    (void)fOptBreakIterator; // suppress unused field warning
     init(nullptr, nullptr, status);
 }
 
@@ -809,11 +810,16 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(
         return;
     }
     if (capitalizationContext == UDISPCTX_CAPITALIZATION_FOR_BEGINNING_OF_SENTENCE) {
+#if !UCONFIG_NO_BREAK_ITERATION
         BreakIterator *bi = BreakIterator::createSentenceInstance(locale, status);
         if (U_FAILURE(status)) {
             return;
         }
         init(nfToAdopt, bi, status);
+#else
+        status = U_UNSUPPORTED_ERROR;
+        return;
+#endif // !UCONFIG_NO_BREAK_ITERATION
     } else {
         init(nfToAdopt, nullptr, status);
     }
@@ -832,9 +838,11 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(
     fCache->addRef();
     fNumberFormat->addRef();
     fPluralRules->addRef();
+#if !UCONFIG_NO_BREAK_ITERATION
     if (fOptBreakIterator != nullptr) {
       fOptBreakIterator->addRef();
     }
+#endif // !UCONFIG_NO_BREAK_ITERATION
 }
 
 RelativeDateTimeFormatter& RelativeDateTimeFormatter::operator=(
@@ -843,7 +851,9 @@ RelativeDateTimeFormatter& RelativeDateTimeFormatter::operator=(
         SharedObject::copyPtr(other.fCache, fCache);
         SharedObject::copyPtr(other.fNumberFormat, fNumberFormat);
         SharedObject::copyPtr(other.fPluralRules, fPluralRules);
+#if !UCONFIG_NO_BREAK_ITERATION
         SharedObject::copyPtr(other.fOptBreakIterator, fOptBreakIterator);
+#endif // !UCONFIG_NO_BREAK_ITERATION
         fStyle = other.fStyle;
         fContext = other.fContext;
         fLocale = other.fLocale;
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/reldatefmt.cpp
```diff
@@ -817,6 +839,9 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(
     } else {
         init(nfToAdopt, nullptr, status);
     }
+#else
```
FrankYFTang, line 842:
```
#else
     if (capitalizationContext == UDISPCTX_CAPITALIZATION_FOR_BEGINNING_OF_SENTENCE) {
        status =   U_UNSUPPORTED_ERROR;
        return ;
    }
    init(nfToAdopt, status);
#endif // !UCONFIG_NO_BREAK_ITERATION
```

icu4c/source/i18n/unicode/reldatefmt.h
```diff
@@ -696,11 +698,15 @@ class U_I18N_API RelativeDateTimeFormatter : public UObject {
     const SharedPluralRules *fPluralRules;
     UDateRelativeDateTimeFormatterStyle fStyle;
     UDisplayContext fContext;
+#if !UCONFIG_NO_BREAK_ITERATION
     const SharedBreakIterator *fOptBreakIterator;
```
markusicu, line 702: Please initialize the fOptBreakIterator with nullptr here in the class declaration and remove the constructor initializers. That will remove clutter because you won't need to ifdef each of those any more.

icu4c/source/i18n/unicode/reldatefmt.h
```diff
@@ -696,11 +698,15 @@ class U_I18N_API RelativeDateTimeFormatter : public UObject {
...
+#endif // !UCONFIG_NO_BREAK_ITERATION
     Locale fLocale;
     void init(
             NumberFormat *nfToAdopt,
+#if !UCONFIG_NO_BREAK_ITERATION
             BreakIterator *brkIter,
+#endif // !UCONFIG_NO_BREAK_ITERATION
```
markusicu, line 709:
```suggestion
#if UCONFIG_NO_BREAK_ITERATION
            std::nullptr_t,
#else
            BreakIterator *brkIter,
#endif // !UCONFIG_NO_BREAK_ITERATION
```
(may need `<cstddef>`)

here and in the .cpp function implementation so that call sites that already always give this a nullptr need not be made conditional.

icu4c/source/i18n/reldatefmt.cpp
markusicu: I agree. Modification: I would move the `#if !UCONFIG_NO_BREAK_ITERATION` inside the `if (capitalizationContext == ...) {`

icu4c/source/i18n/unicode/reldatefmt.h
FrankYFTang: but the type of  fOptBreakIterator is SharedBreakIterator  which is ifdef out

FrankYFTang:
ping

icu4c/source/i18n/unicode/reldatefmt.h
markusicu: I see. How about
```
#if UCONFIG_NO_BREAK_ITERATION
    std::nullptr_t fOptBreakIterator;
#else
    const SharedBreakIterator *fOptBreakIterator;
#endif // !UCONFIG_NO_BREAK_ITERATION
```
?
That will declutter *some* of the code, especially the initialization.
You should "only" need ifdefs where there is code with addRef(), removeRef(), and SharedObject::someThing().

hsivonen:
> ping

Review ack. Thanks. I'll try to get back to this.

icu4c/source/i18n/reldatefmt.cpp
hsivonen: Fixed.

icu4c/source/i18n/unicode/reldatefmt.h
hsivonen: Used `std::nullptr_t`.

icu4c/source/i18n/unicode/reldatefmt.h
hsivonen: Done.

FrankYFTang (approved):
LGTM @markusicu any additional concerns?

icu4c/source/i18n/reldatefmt.cpp
```diff
@@ -809,11 +810,16 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(
...
     if (capitalizationContext == UDISPCTX_CAPITALIZATION_FOR_BEGINNING_OF_SENTENCE) {
+#if !UCONFIG_NO_BREAK_ITERATION
         BreakIterator *bi = BreakIterator::createSentenceInstance(locale, status);
         if (U_FAILURE(status)) {
             return;
         }
         init(nfToAdopt, bi, status);
+#else
+        status = U_UNSUPPORTED_ERROR;
+        return;
```
markusicu, line 821: nit: unnecessary line

icu4c/source/i18n/reldatefmt.cpp
```diff
@@ -827,14 +833,18 @@ RelativeDateTimeFormatter::RelativeDateTimeFormatter(
           fPluralRules(other.fPluralRules),
           fStyle(other.fStyle),
           fContext(other.fContext),
+#if !UCONFIG_NO_BREAK_ITERATION
```
markusicu, line 836: Does this still need to be ifdef'ed? We should be able to copy a nullptr.

icu4c/source/i18n/unicode/reldatefmt.h
```diff
@@ -248,15 +248,17 @@ typedef enum UDateDirection {
 #endif  // U_HIDE_DEPRECATED_API
 } UDateDirection;
 
-#if !UCONFIG_NO_BREAK_ITERATION
-
 U_NAMESPACE_BEGIN
 
+#if !UCONFIG_NO_BREAK_ITERATION
```
markusicu, line 253: nit: forward decls should be harmless and should not need ifdef

icu4c/source/i18n/unicode/reldatefmt.h
```diff
@@ -248,15 +248,17 @@ typedef enum UDateDirection {
...
 class BreakIterator;
+#endif // !UCONFIG_NO_BREAK_ITERATION
 class RelativeDateTimeCacheData;
 class SharedNumberFormat;
 class SharedPluralRules;
+#if !UCONFIG_NO_BREAK_ITERATION
```
markusicu, line 259: ditto

icu4c/source/i18n/reldatefmt.cpp
FrankYFTang: Markus, I think it is unnecessary if no one ever will add more lines after the else block. But would that be safer for him to put the return here in case years later someone may add some more line after the else block?

markusicu: I called it a "nit" indicating that I could live with it as it is, but it jumped out as weird and unnecessary.

It *could* be safer to return immediately when we think we are done, or it *could* be a subtle bug on a future change that intends for something else to happen at the end of the function even when there is an error. We won't know until someone wants to add stuff there.

FrankYFTang (approved):

FrankYFTang:
I modified the branch based on Markus comments (except one)

hsivonen:
> Please Squash I do not have the right to do so in your branch.

Done. Thanks.

FrankYFTang:
merged. Thanks

PR #2309, opened 2023-02-07
ICU-22262 Automate environment test

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22262

Configure 17+21 = 38 jobs to run icu tests under
500+ locales with LC_ALL= environment variable, and
600+ timezone with TZ= environment variable.

Fix broken tests discovered:
1. Ignore error in cintltst/cdtrgtst.c if the calendar of the default
   locale is not "gregorian" because the expectation is reasonable
   for 'gregorian' calendar only.
2. Consider the string are the same if the numerical value of the digit
   are the same for tests inside cintltst/cdtrgtst.c.
3. Skip the TestCDefaultLocale() error in cintltst/cloctst.c
   if LC_ALL env is set.
4. Fix the test which incorrectly assume the string iteration is not end
   (under "wo" locale") and avoid incorrect expectation in TestAssignment()
   of intltest/itercoll.cpp (see [ICU-22273](https://unicode-org.atlassian.net/browse/ICU-22273) )
5. Skip Test4139940() error in intltest/loctest.cpp
   if the calendar is not "gregorian".
6. Log known issue [ICU-22274](https://unicode-org.atlassian.net/browse/ICU-22274) for failure of detectHostTimeZone()'s raw offset
   != host timezone's offset while the default TimeZones is one of the
   following troublesome TimeZones:
   "Africa/Casablanca", "Europe/Dublin", "America/Godthab", "America/Nuuk",
   or "Africa/El_Aaiun".

Following
https://unicode-org.github.io/icu/processes/release/tasks/integration.html#run-the-environment-tests

[ICU-22273]: https://unicode-org.atlassian.net/browse/ICU-22273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/icu_envtest.yml
+++ b/.github/workflows/icu_envtest.yml
@@ -1,0 +1,136 @@
+# Copyright (C) 2023 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+#
+# GitHub Action configuration script for ICU environment test tasks.
+#
+
+name: GHA EnvTest
+
+on:
+  workflow_dispatch:
+    # To trigger the Env Test workflow manually, follow the instructions in
+    # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+
+  schedule:
+    # This cron schedule is set to run 10:23 UTC every SAT
+    - cron: '23 10 * * SAT'
+
+jobs:
+  #=================================================================
+  # locale env tests.
+  env-test-locale:
+    runs-on: ubuntu-latest
+    strategy:
+      # "fail-fast: false" let other jobs keep running even if the test break in some locales.
+      fail-fast: false
+      matrix:
+        # Since we have total 500+ locales to run on three set of test, we create
+        # many jobs to test concurrently.
+        # shard is used to bucket the lines from the locale list into jobs.
+        # Currently we run testing of 30 locales per shard, and we have total 17 shards.
+        # 17x30 = 510 > 502 (the number of locales in 'locale -a').
+        tests_per_shard: [30]
+        shard: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
+    steps:
+      - name: Install all locales by apt-get
+        run: |
+          sudo apt-get update -y;
+          sudo apt-get install -y locales-all;
+      - name: Show installed locales post apt-get
+        run: |
+          locale -a;
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
+          cd icu4c/source/;
+          ./runConfigureICU Linux;
+          make -j10 tests;
+      - name: Test different locales as LC_ALL
+        run: |
+          echo "This job run the environment test of the following Locales";
+          begin=$(( ${{matrix.shard}} * ${{matrix.tests_per_shard}} + 1));
+          end=$(( (${{matrix.shard}}+1) * ${{matrix.tests_per_shard}} ));
+          sedarg=${begin},${end}p;
+          locale -a |sed -n ${sedarg};
+
+          cd icu4c/source/test;
+          for loc in `locale -a |sed -n ${sedarg}`;
+          do
+              echo "============================================";
+              echo "Start Tests under LC_ALL=$loc locale";
+              for test_dir in iotest cintltst intltest
+              do
+                  cd $test_dir;
+                  LC_ALL=$loc make check;
+                  cd ..;
+              done
+
+              echo "Complete Tests under LC_ALL=$loc locale";
+              echo "============================================";
+          done
+
+          echo "This job completed the test of the following Locales";
+          locale -a |sed -n ${sedarg};
+
+  #=================================================================
+  # tz env tests.
+  env-test-tz:
+    runs-on: ubuntu-latest
+    strategy:
+      # "fail-fast: false" let other jobs keep running even if the test break in some timezones.
+      fail-fast: false
+      matrix:
+        # Since we have total 600+ timezones to run on three set of test, we create
+        # many jobs to test concurrently.
+        # shard is used to bucket the lines from the timezone list into jobs.
+        # Currently we run testing of 30 timezones per shard, and we have total 21 shards.
+        # 21x30 = 630 > 604 (the number of known timezones).
+        tests_per_shard: [30]
+        shard: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
+    steps:
+      - name: Install all locales by apt-get
+        run: |
+          sudo apt-get update -y;
+          sudo apt-get install -y tzdata;
+      - name: Show all the TimeZones
+        run: |
+          echo "Total number of TimeZone is "
+          find /usr/share/zoneinfo/ -type f,l|egrep -v "/(right|posix)/"|egrep -v "\.tab"|wc -l;
+          find /usr/share/zoneinfo/ -type f,l|egrep -v "/(right|posix)/"|egrep -v "\.tab"|cut -d '/' -f5-50;
+      - name: Checkout and setup
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
+          cd icu4c/source/;
+          ./runConfigureICU Linux;
+          make -j10 tests;
+      - name: Test different locales as TZ
+        run: |
+          echo "This job run the environment test of the following TimeZones";
+          begin=$(( ${{matrix.shard}} * ${{matrix.tests_per_shard}} + 1));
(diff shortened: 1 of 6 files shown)


FrankYFTang:
with 30 tests (locales or timezone) per job (eg shard), each job take about 36-45 minutes to run, faster than two othe pre-existing ci . 
My intention is to land this PR as is, and after the landing, create another PR to comment out (not remove, but just comment out) the two lines
```
  pull_request:
    branches: '**'
```
under the line "For BRS" on the top of file .github/workflows/icu_envtest.yml
so Env Test will stop running on every pull_request but only run by cron on every Sat 10:23 Sat UTC (which is about every Friday night PST) because we do not need to run this so frequently and Weekly run should be good enough. 
In the future if people want to run it on demand, just uncomment that two line and push a PR and such push (no need to merge, just push) to github will trigger gihub to run the Env tests.

FrankYFTang:
currently, the env test set up (30 tests per job) require less time (which is about 45mins) to run than
uconfig_variation-check-unit-tests and uconfig_variation-check-all-header-tests (which both took about 55mins)
If we speed up these two tests, we may want to reconfigure the number of tests per job in the all env test to make it complete faster.

markusicu:
> My intention is to land this PR as is, and after the landing, create another PR to comment out (not remove, but just comment out) the two lines ... so Env Test will stop running on every pull_request but only run by cron ... Weekly run should be good enough. In the future if people want to run it on demand, just uncomment that two line and push a PR and such push (no need to merge, just push) to github will trigger gihub to run the Env tests.

@mihnita found that it's possible to add a few config lines that define a command line interface to manually trigger a GitHub Action, no pending PR needed. That seems better.

.github/workflows/icu_envtest.yml
```diff
@@ -0,0 +1,137 @@
...
+on:
+  schedule:
+    # this cron schedule is set to run 10:23 UTC every SAT
+    - cron: '23 10 * * SAT'
+
+  # For BRS:
+  # Uncomment the following two lines in a commit and push that to a PR on github
+  # will trigger the testing in addition to the schedule weekly run above.
+  pull_request:
+    branches: '**'
```
markusicu, line 17: Please ask @mihnita for instructions for setting up a command line interface for a manual trigger.

.github/workflows/icu_envtest.yml
```diff
@@ -0,0 +1,137 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      # "fail-fast: false" let other jobs keep running even if the test break in some locales.
+      fail-fast: false
+      matrix:
+        # Since we have total 500+ locales to run on three set of test, we create
+        # many jobs to test concurrently.
+        # shard is used to bucket the lines from the locale list into jobs.
+        # Currently we run testing of 30 locales per shard, and we have total 17 shards.
+        # 17x30 = 510 > 502 (the number of locales in 'locale -a'.
```
markusicu, line 32:
```suggestion
        # 17x30 = 510 > 502 (the number of locales in 'locale -a').
```

icu4c/source/test/cintltst/cdtrgtst.c
```diff
@@ -241,6 +247,26 @@ void Test4059917()
...
+    int32_t len1 = u_strlen(a);
+    int32_t len2 = u_strlen(b);
+    UChar32 ch1, ch2;
+    int32_t idx1, idx2;
+    idx1=idx2=0;
+    // U16_NEXT will increment idx1 and idx2 inside the loop.
+    while (idx1 < len1 && idx2 < len2) {
+        U16_NEXT(a, idx1, len1, ch1);
+        U16_NEXT(b, idx2, len2, ch2);
+        if (ch1 != ch2 && u_charDigitValue(ch1) != u_charDigitValue(ch2)) return false;
```
markusicu, line 264: This passes if the characters differ and neither is a digit. -->

```suggestion
        if (ch1 != ch2) {
            int32_t digit1 = u_charDigitValue(ch1);
            int32_t digit2 = u_charDigitValue(ch2);
            if (digit1 < 0 || digit1 != digit2) { return false; }
        }
```

icu4c/source/test/cintltst/cdtrgtst.c
```diff
@@ -266,7 +292,14 @@ void aux917( UDateFormat *fmt, UChar* str)
     status = U_ZERO_ERROR;
     formatted = myFormatit(fmt, d1);
     if( u_strcmp(formatted,str)!=0) {
-        log_err("Fail: Want %s Got: %s\n", austrdup(str),  austrdup(formatted) );
+        // We may get non-ASCII result back for some locale such as ne_NP
+        // which use numbering systems other than latn
+        if (strcmp("gregorian", ucal_getType(udat_getCalendar(fmt), &status)) != 0) {
+          log_verbose("Skipping Test4059917 when the default date time format is not using gregorian calendart.");
```
markusicu, line 298:
```suggestion
          log_verbose("Skipping Test4059917 when the default date time format is not using gregorian calendar.");
```

.github/workflows/icu_envtest.yml
FrankYFTang: @mihnita Is workflow_dispatch stated in  https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch what you are talking about?

FrankYFTang:
Ready for Review

.github/workflows/icu_envtest.yml
FrankYFTang: it is hard to debug how to set up that manual trigger without this yaml file in the repo. Could we resolve that part later? I put in workflow_dispatch but I am not 100% sure it work unless we land this. (not sure would that only show up after the config is landing or not)

mihnita: Yes, it is `workflow_dispatch`

> it is hard to debug how to set up that manual trigger without this yaml file in the repo

I was able to play with it in a fork + branch. I can help resolve it later, or in this PR, your call.

.github/workflows/icu_envtest.yml
```diff
@@ -0,0 +1,147 @@
+# Copyright (C) 2023 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+#
+# GitHub Action configuration script for ICU environment test tasks.
+#
+
+name: GHA EnvTest
+
+on:
```
mihnita, line 9: Yes, this should be enough for manual trigger

.github/workflows/icu_envtest.yml
echeran: FYI, after configuring `workflow_dispatch`, you are not limited to the CLI for a manual trigger. You can also use the UI, which seems much less cumbersome and error-prone: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow?tool=webui

icu4c/source/test/cintltst/cdtrgtst.c
```diff
@@ -241,6 +247,30 @@ void Test4059917()
     free(myDate);
 }
 
+UBool EqualIgnoreNumberingSystem(UChar* a, UChar* b);
+// This function return true if the sting contains the same value or the
+// numerical values are the same if we ignore the
```
echeran, line 252:
```suggestion
// numerical values are the same if we ignore the numbering system.
```

.github/workflows/icu_envtest.yml
FrankYFTang: how to trigger manually?

FrankYFTang:
PTAL

FrankYFTang:
ping

markusicu (approved):
lgtm tnx!

As a follow-up after merging this PR, please edit the BRS task description and add near its top a link to this .yml file and say that it runs periodically and can also be triggered manually / on demand.

FrankYFTang:
the manual trigger up show up in https://github.com/unicode-org/icu/actions/workflows/icu_envtest.yml after I landed the PR

markusicu:
> the manual trigger up show up in https://github.com/unicode-org/icu/actions/workflows/icu_envtest.yml after I landed the PR

Very nice! :-)

markusicu:
> lgtm tnx!
> 
> As a follow-up after merging this PR, please edit the BRS task description and add near its top a link to this .yml file and say that it runs periodically and can also be triggered manually / on demand.

At https://github.com/unicode-org/icu/blob/main/docs/processes/release/tasks/integration.md?plain=1#L399
after the “Run the environment tests” heading, please add something like

> Since ICU 73, this test has been automated via a [GitHub Action](https://github.com/unicode-org/icu/actions/workflows/icu_envtest.yml) which runs periodically. You can also trigger it manually from that UI page.

PR #2322, opened 2023-02-14
ICU-6065 UnicodeSet::closeOver(simple case folding)

markusicu:
Adds a new option for UnicodeSet::closeOver(int options) and other functions with this options parameter:
- C (unicode/uset.h) USET_SIMPLE_CASE_INSENSITIVE
- Java UnicodeSet.SIMPLE_CASE_INSENSITIVE

With the new option, the UnicodeSet is expanded so that for each code point the other code points in its Simple_Case_Folding (scf) equivalence class are added, and each multi-character string is replaced with a version where each code point is replaced with its scf mapping.

The implementation uses the same data structure as for the full case folding (UnicodeSet.CASE_INSENSITIVE) but only adds the “simple” (single-code point) mappings, and also omits a hardcoded list of mappings where there is no scf equivalence. (I will propose additional scf mappings in a future version of Unicode to fix those.)

There is a new test function which computes the scf equivalence classes and compares closeOver(SIMPLE_CASE_INSENSITIVE) with applying that test-code closure data.

As an attempted optimization, when the UnicodeSet is large, the closeOver() code only looks at the code points with the Case_Sensitive property. It looks at fewer code points at the cost of lazy-initializing the Case_Insensitive set and computing the intersection.

There are other changes for some code cleanup and minor refactoring. For example, in the UnicodeSet implementation, there are now two separate functions for CASE_INSENSITIVE/SIMPLE_CASE_INSENSITIVE vs. ADD_CASE_MAPPINGS. In the low-level code, the hardcoded mappings for letters like ‘i’ are moved into the slower case-exception-handling path. The Java closeOver() test had been missing some test data that the C++ test had.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-6065

### DIFF
--- a/icu4c/source/common/characterproperties.cpp
+++ b/icu4c/source/common/characterproperties.cpp
@@ -377,22 +377,30 @@ UCPMap *makeMap(UProperty property, UErrorCode &errorCode) {
 
 }  // namespace
 
-U_NAMESPACE_USE
+U_NAMESPACE_BEGIN
 
-U_CAPI const USet * U_EXPORT2
-u_getBinaryPropertySet(UProperty property, UErrorCode *pErrorCode) {
-    if (U_FAILURE(*pErrorCode)) { return nullptr; }
+const UnicodeSet *CharacterProperties::getBinaryPropertySet(UProperty property, UErrorCode &errorCode) {
+    if (U_FAILURE(errorCode)) { return nullptr; }
     if (property < 0 || UCHAR_BINARY_LIMIT <= property) {
-        *pErrorCode = U_ILLEGAL_ARGUMENT_ERROR;
+        errorCode = U_ILLEGAL_ARGUMENT_ERROR;
         return nullptr;
     }
     Mutex m(&cpMutex);
     UnicodeSet *set = sets[property];
     if (set == nullptr) {
-        sets[property] = set = makeSet(property, *pErrorCode);
+        sets[property] = set = makeSet(property, errorCode);
     }
-    if (U_FAILURE(*pErrorCode)) { return nullptr; }
-    return set->toUSet();
+    return set;
+}
+
+U_NAMESPACE_END
+
+U_NAMESPACE_USE
+
+U_CAPI const USet * U_EXPORT2
+u_getBinaryPropertySet(UProperty property, UErrorCode *pErrorCode) {
+    const UnicodeSet *set = CharacterProperties::getBinaryPropertySet(property, *pErrorCode);
+    return U_SUCCESS(*pErrorCode) ? set->toUSet() : nullptr;
 }
 
 U_CAPI const UCPMap * U_EXPORT2
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
@@ -459,7 +459,9 @@ public class UnicodeSet extends UnicodeFilter implements Iterable<String>, Compa
      * for the syntax of the pattern language.
      * @param pattern a string specifying what characters are in the set
      * @param options a bitmask indicating which options to apply.
-     * Valid options are IGNORE_SPACE and CASE.
+     * Valid options are {@link #IGNORE_SPACE} and
+     * at most one of {@link #CASE_INSENSITIVE}, {@link #ADD_CASE_MAPPINGS},
+     * {@link #SIMPLE_CASE_INSENSITIVE}. These case options are mutually exclusive.
      * @exception java.lang.IllegalArgumentException if the pattern contains
      * a syntax error.
      * @stable ICU 3.8
@@ -495,7 +497,9 @@ public class UnicodeSet extends UnicodeFilter implements Iterable<String>, Compa
      * @param symbols a symbol table mapping variables to char[] arrays
      * and chars to UnicodeSets
      * @param options a bitmask indicating which options to apply.
-     * Valid options are IGNORE_SPACE and CASE.
+     * Valid options are {@link #IGNORE_SPACE} and
+     * at most one of {@link #CASE_INSENSITIVE}, {@link #ADD_CASE_MAPPINGS},
+     * {@link #SIMPLE_CASE_INSENSITIVE}. These case options are mutually exclusive.
      * @exception java.lang.IllegalArgumentException if the pattern
      * contains a syntax error.
      * @stable ICU 3.2
@@ -587,7 +591,9 @@ public class UnicodeSet extends UnicodeFilter implements Iterable<String>, Compa
      * See the class description for the syntax of the pattern language.
      * @param pattern a string specifying what characters are in the set
      * @param options a bitmask indicating which options to apply.
-     * Valid options are IGNORE_SPACE and CASE.
+     * Valid options are {@link #IGNORE_SPACE} and
+     * at most one of {@link #CASE_INSENSITIVE}, {@link #ADD_CASE_MAPPINGS},
+     * {@link #SIMPLE_CASE_INSENSITIVE}. These case options are mutually exclusive.
      * @exception java.lang.IllegalArgumentException if the pattern
      * contains a syntax error.
      * @stable ICU 3.8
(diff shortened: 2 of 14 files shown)


macchiati (commented):
Looks nice!

markusicu:
@macchiati @richgillam please take a look at the "better docs" commit here.

FYI: C++ is complete (modulo feedback) but I have not yet started the Java port, so I am keeping this PR draft+incomplete.

richgillam (approved):
The doc changes look good to me.

markusicu:
I think I finished the Java port. It should be complete. Please review, again probably best one commit at a time. I will look it over myself once more as well. I will squash after addressing feedback.

icu4c/source/common/characterproperties.cpp
```diff
@@ -377,22 +377,30 @@ UCPMap *makeMap(UProperty property, UErrorCode &errorCode) {
 
 }  // namespace
 
-U_NAMESPACE_USE
+U_NAMESPACE_BEGIN
 
-U_CAPI const USet * U_EXPORT2
-u_getBinaryPropertySet(UProperty property, UErrorCode *pErrorCode) {
-    if (U_FAILURE(*pErrorCode)) { return nullptr; }
+const UnicodeSet *CharacterProperties::getBinaryPropertySet(UProperty property, UErrorCode &errorCode) {
```
markusicu, line 382: FYI  This is just creating an internal C++ API which is easier to use (less casting) from C++ impl code than the public C API.

markusicu (commented):
I looked over the changes myself and made very minor docs & test changes. I think I am done. Please review :-)

markusicu:
I wrote a lengthy description for this pull request. Hope this helps with reviewing as well as documentation for posterity.

richgillam (dismissed):
The API and documentation changes look good to me.  I don't think I'm qualified to review the implementation changes, but as far as I could figure them out, they looked good too.

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -3861,58 +3869,81 @@ private void applyPropertyPattern(RuleCharacterIterator chars,
...
+     *
+     * <p>This value is an options bit set value for some
+     * constructors, applyPattern(), and closeOver().
+     * It can be ORed together with other, unrelated options.
+     *
+     * @draft ICU 73
+     */
+    public static final int SIMPLE_CASE_INSENSITIVE = 6;
+
+    private static final int CASE_MASK = CASE_INSENSITIVE | ADD_CASE_MAPPINGS;
```
macchiati, line 3945: Question: would it be useful to note which of these don't make sense in combination? eg what would 0xFF do?

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -3930,97 +3961,185 @@ private static final void addCaseMapping(UnicodeSet set, int result, StringBuild
         // see UCaseProps
     }
 
+    /** For case closure on a large set, look only at code points with relevant properties. */
+    UnicodeSet maybeOnlyCaseSensitive(UnicodeSet src) {
+        if (src.size() < 30) {
+            return src;
+        }
+        return CharacterProperties.getBinaryPropertySet(UProperty.CASE_SENSITIVE).
```
macchiati, line 3969: Not a blocker.
Would it be faster to clone the src and then retainAll CharacterProperties.getBinaryPropertySet(UProperty.CASE_SENSITIVE), because the src is likely to be smaller? Or maybe clone the smaller of the two?

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
macchiati: Or are they really just singletons?

macchiati:
I didn't review the C++, just the java. I had no objections — just wanted to get a read from you on some of my comments.

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -3861,58 +3869,81 @@ private void applyPropertyPattern(RuleCharacterIterator chars,
...
 
     /**
-     * Bitmask for constructor, applyPattern(), and closeOver()
-     * indicating letter case.  This may be ORed together with other
-     * selectors.
+     * Alias for {@link #CASE_INSENSITIVE}.
      *
+     * @stable ICU 3.8
+     */
+    public static final int CASE = 2;
```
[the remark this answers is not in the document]
markusicu: could do --> https://unicode-org.atlassian.net/browse/ICU-22299

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
markusicu: In the public-API docs, I documented
- the valid options
- that the case-related ones are mutually exclusive
- (in a couple of places) that undefined bits are ignored/reserved

I don't think we need to add more on this internal constant.

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
markusicu: Hard to say without a benchmark. It's probably cheaper to clone the one with fewer *ranges*, and if the src.hasStrings() then cloning src unnecessarily clones the structure for that too.

Do you want me to do something like
```
UnicodeSet sensitive = ...
if (src.hasStrings() || src.getRangeCount() > sensitive.getRangeCount()) {
    return sensitive.cloneAsThawed().retainAll(src);
} else {
    return src.clone().retainAll(sensitive);
}
```
?

icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -1323,38 +1325,98 @@ public void TestScriptNames() {
...
+        String SIMPLE_CASE_INSENSITIVE = String.valueOf(UnicodeSet.SIMPLE_CASE_INSENSITIVE);
         String[] DATA = {
                 // selector, input, output
                 CASE,
                 "[aq\u00DF{Bc}{bC}{Fi}]",
                 "[aAqQ\u00DF\u1E9E\uFB01{ss}{bc}{fi}]", // U+1E9E LATIN CAPITAL LETTER SHARP S is new in Unicode 5.1
 
+                SIMPLE_CASE_INSENSITIVE,
+                "[aq\u00DF{Bc}{bC}{Fi}]",
+                "[aAqQ\u00DF\u1E9E{bc}{fi}]",
```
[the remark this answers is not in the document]
markusicu: I can see if I can make a quick change to that effect. I would probably keep combining marks, and hard-to-distinguish characters like Kelvin, escaped.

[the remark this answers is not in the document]
markusicu: I tried this, and I don't like it. Half of the characters want to remain unescaped as noted above. But for several of the others it's also more obfuscating what's going on than helpful.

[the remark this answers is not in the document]
macchiati: I agree with combining marks, di, controls, etc. 

Not sure about combining marks or confusables. I find 
```
"[\u0390]",          
"[\u0390\u1FD3{\u03B9\u0308\u0301}]", 

```
more opaque than
```
"[ΐ]",
"[ΐ{ΐ}ΐ]",
```

But maybe one or the other in comments?

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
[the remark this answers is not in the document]
macchiati: great, thanks.

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
macchiati: ok, no problem.

icu4j/main/classes/core/src/com/ibm/icu/text/UnicodeSet.java
macchiati: You have a good point. Probably not worth doing without a benchmark.

markusicu: I just did this in the latest commit.

macchiati (dismissed):

markusicu:
Thanks for reviewing! I just squashed, please re-approve.

macchiati (approved):

PR #2323, opened 2023-02-14
ICU-21386 uprv_tzname() should find the correct Olson ID when /etc/localtime is a "double" link

Robbos:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21386

### DIFF
--- a/icu4c/source/common/putil.cpp
+++ b/icu4c/source/common/putil.cpp
@@ -722,7 +722,7 @@ extern U_IMPORT char *U_TZNAME[];
 #include <dirent.h>  /* Needed to search through system timezone files */
 #endif
 static char gTimeZoneBuffer[PATH_MAX];
-static char *gTimeZoneBufferPtr = nullptr;
+static const char *gTimeZoneBufferPtr = nullptr;
 #endif
 
 #if !U_PLATFORM_USES_ONLY_WIN32_API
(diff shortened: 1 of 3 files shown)


Robbos:
By using "realpath" instead of "readlink" we can still get the proper timezone when there's a soft link to a soft link.
We ended up in this case because /etc/localtime was readonly and we couldn't let it directly link to /usr/share/zoneinfo/<OlsonID>

In previous commits, I was able to create a test case for reading the Olson ID, but unfortunately could not get it to work with the CI system, because the test should only run on platforms that use Olson IDs.
If you have any input on how to get the testcase to run properly, I'd like to know, but currently the code is less complex and the change is smaller.

icu4c/source/common/putil.cpp
```diff
@@ -1171,16 +1171,16 @@ uprv_tzname(int n)
         because the tzfile contents is underspecified.
         This isn't guaranteed to work because it may not be a symlink.
         */
-        int32_t ret = (int32_t)readlink(TZDEFAULT, gTimeZoneBuffer, sizeof(gTimeZoneBuffer)-1);
-        if (0 < ret) {
+        char *ret = realpath(TZDEFAULT, gTimeZoneBuffer);
```
markusicu, line 1174: Should the preceding comment be adjusted?

icu4c/source/common/putil.cpp
```diff
@@ -1171,16 +1171,16 @@ uprv_tzname(int n)
...
-            gTimeZoneBuffer[ret] = 0;
             char *  tzZoneInfoTailPtr = uprv_strstr(gTimeZoneBuffer, TZZONEINFOTAIL);
-
-            if (tzZoneInfoTailPtr != nullptr
-                && isValidOlsonID(tzZoneInfoTailPtr + tzZoneInfoTailLen))
-            {
-                return (gTimeZoneBufferPtr = tzZoneInfoTailPtr + tzZoneInfoTailLen);
+            if (tzZoneInfoTailPtr != nullptr) {
+                tzZoneInfoTailPtr += tzZoneInfoTailLen;
+                skipZoneIDPrefix(const_cast<const char **>(&tzZoneInfoTailPtr));
```
markusicu, line 1180: Is this added line related to the same issue?
(I don't know my way around this part of the code...)

roubert, line 1180: I suggest that you update the signature of the skipZoneIDPrefix() helper function as needed, instead of using a const_cast here.

icu4c/source/common/putil.cpp
roubert: No, what that comment says would still be true after these changes.

roubert (changes requested):
I think this change looks good, realpath() is a standard function and is promised to not write beyond PATH_MAX.

icu4c/source/common/putil.cpp
Robbos: Yes.
The path might be something like /usr/share/zoneinfo/posix/<OlsonID>.
The tzZoneInfoTailPtr will be put behind the "zoneinfo" part of the path. However, to fetch the OlsonID, we also need to get rid of the "posix" part. That's what skipZoneIDPrefix does. This function could be used without any changes.

Robbos: I think with my new commit it's not necessary to adjust the signature and I can keep it const.

yumaoka (dismissed):
Thanks for updates. LGTM.

roubert (approved):
LGTM

yumaoka:
@Robbos Can you go to https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/icu/2323 and click "Squash Branch", then perform squash? This should resolve the check failure - "single-commit". I cannot squash these commits in your fork. Once it's completed, I will merge the PR. Thanks!

yumaoka (approved):
LGTM

FrankYFTang:
syg@google.com found the landing of this PR cause a lot of problem on some platforms.
See https://unicode-org.atlassian.net/browse/ICU-22541
We would like to revert it for 74 now.

FrankYFTang:
Some other bugs that may related to the landing of this PR
https://bugs.chromium.org/p/chromium/issues/detail?id=1473422
https://bugs.chromium.org/p/v8/issues/detail?id=14343

FrankYFTang:
I chat with Markus. It seems reverting this will not solve the whole problem. While it may be a temp solution for chrome w/ 73-1. We still have time to fix it real for 74.

PR #2324, opened 2023-02-14
ICU-22224 Enable UBSAN and fix breakage

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22224

This is the preparation work to fix other ubsan breakage reported. Some of the bugs cannot be tested without a running ubsan env. So we need to add ubsan support first. 

1. Add Github Action to run ubsan
2. Add a macro NO_SANITIZE_UNDEFINED to annotate some function which take
enum but the test put in out of bound int. So the unit test run will not
trigger ubsan breakage.
3. For some C function that may take output buffer as nullptr and
   internally do pointer math of that without using it, only perform the
   pointer math if the value is not nullptr to avoid ubsan breakage.
4. For function which take possible null pointer of ParserError and
   internally deref it to pass into another function as ref, use an
   internal stack ParseError to replace it if the pointer is nullptr to
   avoid ubsan breakage and use reference of object in null location.

### DIFF
--- a/icu4c/source/common/propname.cpp
+++ b/icu4c/source/common/propname.cpp
@@ -303,7 +303,10 @@ u_getPropertyEnum(const char* alias) {
 U_CAPI const char* U_EXPORT2
 u_getPropertyValueName(UProperty property,
                        int32_t value,
-                       UPropertyNameChoice nameChoice) {
+                       UPropertyNameChoice nameChoice) UPRV_NO_SANITIZE_UNDEFINED {
+    if (nameChoice < 0 || U_PROPERTY_NAME_CHOICE_COUNT <= nameChoice) {
+       return nullptr;
+    }
     U_NAMESPACE_USE
     return PropNameData::getPropertyValueName(property, value, nameChoice);
 }
--- a/icu4c/source/i18n/number_capi.cpp
+++ b/icu4c/source/i18n/number_capi.cpp
@@ -131,7 +131,8 @@ unumf_openForSkeletonAndLocaleWithError(const char16_t* skeleton, int32_t skelet
     }
     // Readonly-alias constructor (first argument is whether we are NUL-terminated)
     UnicodeString skeletonString(skeletonLen == -1, skeleton, skeletonLen);
-    impl->fFormatter = NumberFormatter::forSkeleton(skeletonString, *perror, *ec).locale(locale);
+    UParseError tempParseError;
+    impl->fFormatter = NumberFormatter::forSkeleton(skeletonString, (perror == nullptr) ? tempParseError : *perror, *ec).locale(locale);
     return impl->exportForC();
 }
 
--- a/icu4c/source/i18n/numrange_capi.cpp
+++ b/icu4c/source/i18n/numrange_capi.cpp
@@ -97,8 +97,9 @@ unumrf_openForSkeletonWithCollapseAndIdentityFallback(
     }
     // Readonly-alias constructor (first argument is whether we are NUL-terminated)
     UnicodeString skeletonString(skeletonLen == -1, skeleton, skeletonLen);
+    UParseError tempParseError;
     impl->fFormatter = NumberRangeFormatter::withLocale(locale)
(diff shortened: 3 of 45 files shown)


icu4c/source/common/propname.cpp
```diff
@@ -303,7 +303,10 @@ u_getPropertyEnum(const char* alias) {
 U_CAPI const char* U_EXPORT2
 u_getPropertyValueName(UProperty property,
                        int32_t value,
-                       UPropertyNameChoice nameChoice) {
+                       UPropertyNameChoice nameChoice) NO_SANITIZE_UNDEFINED {
```
aheninger, line 306: Would casting the incoming enum values to ints in the range checks work as an alternative to NO_SANITIZE_UNDEFINED, for this kind of check?

It might be slightly safer, in that it would avoid the undefined behavior, instead of suppressing checking for it. Deals with the off chance that some perverse future compiler recognizes that the if condition can only be true for undefined behavior conditions and decides to optimize the whole check away.

FrankYFTang: good idea. Let me check would that work.

FrankYFTang: I tried, UBSAN is not happen the moment the code access the incoming enum, in the moment if I try to cast it to int. UBSAN detect the value is out of defined values before the casting, triggered by the casting.

icu4c/source/i18n/number_capi.cpp
```diff
@@ -131,6 +131,10 @@ unumf_openForSkeletonAndLocaleWithError(const char16_t* skeleton, int32_t skelet
     }
     // Readonly-alias constructor (first argument is whether we are NUL-terminated)
     UnicodeString skeletonString(skeletonLen == -1, skeleton, skeletonLen);
+    UParseError temp_parse_error;
```
markusicu, line 134:
```suggestion
    UParseError tempParseError;
```

icu4c/source/i18n/numrange_capi.cpp
```diff
@@ -97,6 +97,10 @@ unumrf_openForSkeletonWithCollapseAndIdentityFallback(
     }
     // Readonly-alias constructor (first argument is whether we are NUL-terminated)
     UnicodeString skeletonString(skeletonLen == -1, skeleton, skeletonLen);
+    UParseError temp_parse_error;
```
markusicu, line 100: ditto

FrankYFTang:
PTAL

markusicu (commented):
lgtm

For the UParseError workarounds, if the parse errors are truly optional, then the functions we call should take pointers not references, but if they are public APIs then we can't change them. And I don't want to block your PR on that. FYI @sffc @pedberg-icu @richgillam

(rant) undefined behavior of shifts of signed ints that should be defined behavior... (/rant)

aheninger:
I sent a fix  to Frank for the undefined behavior in the RuleBasedBreakIterator default constructor, as a PR to his branch.
https://github.com/FrankYFTang/icu/pull/4

I see multiple plausible ways to  proceed:
1.  Frank merges my PR into his branch, then pushes the result, thus updating this PR.
2.  Frank merges this PR as-is, then I submit a followup PR with my changes. Using the same ticket.
3.  Punt my fix until the next release

FrankYFTang:
> I sent a fix to Frank for the undefined behavior in the RuleBasedBreakIterator default constructor, as a PR to his branch. [FrankYFTang#4](https://github.com/FrankYFTang/icu/pull/4)
> 
> I see multiple plausible ways to proceed:
> 
> 1. Frank merges my PR into his branch, thus updating this PR.
> 2. Frank merges this PR as-is, then I submit a followup PR with my changes. Using the same ticket.
> 3. Punt my fix until the next release

I merge your PR into my branch. Thanks
PTAL

markusicu (approved):
lgtm tnx!

aheninger (approved):
LGTM

PR #2329, opened 2023-02-16
ICU-22283 Add additional ERoundingMode variants

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22283


richgillam (commented):
Most of this looks fine, but I couldn't follow what was going on the unit tests.

There were a couple of things: First, I didn't understand what the third column was, the one that sometimes had a "K" or some other letter (or sequence of letters) in it.  Second, it seems like you're modifying a lot of existing tests, not just adding new ones to cover the new rounding modes, which suggests the rounding-mode change has some kind of side effect (or that I'm just misinterpreting what you're doing).  Could you elucidate?

sffc:
I change numberformattestspecification.txt to run on Java 11 instead of Java 8 so that I could reproduce test failures, which resulted in my having to update more of the tests in that file whose behavior changed between 8 and 11. This change is in the last file in the PR.

richgillam (approved):

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/NumberFormatDataDrivenTest.java
```diff
@@ -718,7 +718,7 @@ public void TestDataDrivenJDK() {
         // #13373: Since not all JDK implementations are the same, test only whitelisted JDKs
         // with known behavior. The JDK version should be occasionally updated.
```
markusicu, line 719: This test runs only on one version, right? -->
```suggestion
        // #13373: Since not all JDK implementations are the same, test only on a JDK
        // with known behavior. The JDK version should be occasionally updated.
```

On other JDKs/versions, does it fail? Or does `assumeTrue()` somehow exit this test function?

sffc: I think that is how assumeTrue works; the test shows as skipped in JUnit out

markusicu (commented):
lgtm tnx
(ignoring the numberformattestspecification.txt changes...)

sffc:
@FrankYFTang lgty?

FrankYFTang (dismissed):
LGTM
Thanks for working on this.

FrankYFTang:
Please squash the branch

sffc:
I think this can be merged once someone approves the squash

richgillam (approved):
LGTM

FrankYFTang (approved):

PR #2330, opened 2023-02-19
ICU-22286 Speed up substring equality comparison

ivan-tymoshenko:
The main goal of this PR is to speed up and simplify the `UnicodeString::startsWith` and `UnicodeString::endsWith` functions implementations. Those functions use `UnicodeString::doCompare` under the hood. `doCompare` compares strings lexicographically, which is unnecessary when we want to know if two substrings are equal. Based on that, we can do two main optimizations:
1. Return `false` if the substrings we want to compare have different lengths. (If we use `UnicodeString::doCompare`, we can't do this because we want to know the order of the strings).
2. We can use `uprv_memcmp` for both little-endian and big-endian systems because we care only about string equality.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22286

### DIFF
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -3489,6 +3489,19 @@ private:
    */
   UBool doEquals(const UnicodeString &text, int32_t len) const;
 
+  inline UBool
+  doEqualsSubstring(int32_t start,
+           int32_t length,
+           const UnicodeString& srcText,
+           int32_t srcStart,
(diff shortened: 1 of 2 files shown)


markusicu:
@ivan-tymoshenko the CI checks are mostly happy :-)
Please squash your follow-up commits into the first one (and make sure to retain its commit message, especially with the `ICU-22286 ` prefix).

markusicu (approved):

PR #2336, opened 2023-02-22
ICU-22293 Add bot for gcc and -Werror

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22293

### DIFF
--- a/icu4c/source/common/ubiditransform.cpp
+++ b/icu4c/source/common/ubiditransform.cpp
@@ -499,7 +499,8 @@ ubiditransform_transform(UBiDiTransform *pBiDiTransform,
     /* Checking for U_SUCCESS() within the loop to bail out on first failure. */
     for (action = pBiDiTransform->pActiveScheme->actions; *action && U_SUCCESS(*pErrorCode); action++) {
         if ((*action)(pBiDiTransform, pErrorCode)) {
-            if (action + 1) {
+            const UBiDiAction * next = action+1;
+            if (next != nullptr) {
                 updateSrc(pBiDiTransform, pBiDiTransform->dest, *pBiDiTransform->pDestLength,
                         *pBiDiTransform->pDestLength, pErrorCode);
             }
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -856,8 +856,8 @@ Calendar::operator=(const Calendar &right)
         fWeekendCease            = right.fWeekendCease;
         fWeekendCeaseMillis      = right.fWeekendCeaseMillis;
         fNextStamp               = right.fNextStamp;
-        uprv_strncpy(validLocale, right.validLocale, sizeof(validLocale));
-        uprv_strncpy(actualLocale, right.actualLocale, sizeof(actualLocale));
+        uprv_memcpy(validLocale, right.validLocale, sizeof(validLocale));
+        uprv_memcpy(actualLocale, right.actualLocale, sizeof(actualLocale));
         validLocale[sizeof(validLocale)-1] = 0;
         actualLocale[sizeof(validLocale)-1] = 0;
     }
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -856,10 +856,10 @@ Calendar::operator=(const Calendar &right)
         fWeekendCease            = right.fWeekendCease;
         fWeekendCeaseMillis      = right.fWeekendCeaseMillis;
         fNextStamp               = right.fNextStamp;
-        uprv_memcpy(validLocale, right.validLocale, sizeof(validLocale));
-        uprv_memcpy(actualLocale, right.actualLocale, sizeof(actualLocale));
-        validLocale[sizeof(validLocale)-1] = 0;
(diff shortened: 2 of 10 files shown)


icu4c/source/i18n/calendar.cpp
```diff
@@ -856,8 +856,8 @@ Calendar::operator=(const Calendar &right)
         fWeekendCease            = right.fWeekendCease;
         fWeekendCeaseMillis      = right.fWeekendCeaseMillis;
         fNextStamp               = right.fNextStamp;
-        uprv_strncpy(validLocale, right.validLocale, sizeof(validLocale));
-        uprv_strncpy(actualLocale, right.actualLocale, sizeof(actualLocale));
+        uprv_memcpy(validLocale, right.validLocale, sizeof(validLocale));
+        uprv_memcpy(actualLocale, right.actualLocale, sizeof(actualLocale));
```
markusicu, line 860: Does it not complain about reading bytes from right.valid/actualLocale that may be uninitialized? I assume that the constructor does not usually touch every last byte there.

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -983,43 +983,61 @@ static void createFileNames(UPKGOptions *o, const char mode, const char *version
...
 #else
         if (noVersion && !reverseExt) {
-            snprintf(libFileNames[LIB_FILE_VERSION_TMP], sizeof(libFileNames[LIB_FILE_VERSION_TMP]), "%s%s%s",
+            if (snprintf(libFileNames[LIB_FILE_VERSION_TMP], sizeof(libFileNames[LIB_FILE_VERSION_TMP]), "%s%s%s",
                     libFileNames[LIB_FILE],
                     FILE_SUFFIX,
-                    pkgDataFlags[SOBJ_EXT]);
+                    pkgDataFlags[SOBJ_EXT]) < 0) {
+                 fprintf(stderr, "Buffer overflow");
+                 exit(-1);
```
markusicu, line 991:
```suggestion
                fprintf(stderr, "Buffer overflow");
                exit(-1);
```

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -983,43 +983,61 @@ static void createFileNames(UPKGOptions *o, const char mode, const char *version
...
-            snprintf(libFileNames[LIB_FILE_VERSION_TMP], sizeof(libFileNames[LIB_FILE_VERSION_TMP]), "%s%s%s%s%s",
+            if (snprintf(libFileNames[LIB_FILE_VERSION_TMP], sizeof(libFileNames[LIB_FILE_VERSION_TMP]), "%s%s%s%s%s",
                     libFileNames[LIB_FILE],
                     FILE_SUFFIX,
                     reverseExt ? version : pkgDataFlags[SOBJ_EXT],
                     FILE_EXTENSION_SEP,
-                    reverseExt ? pkgDataFlags[SOBJ_EXT] : version);
+                    reverseExt ? pkgDataFlags[SOBJ_EXT] : version) < 0) {
+                 fprintf(stderr, "Buffer overflow");
+                 exit(-1);
```
markusicu, line 1001:
```suggestion
                fprintf(stderr, "Buffer overflow");
                exit(-1);
```

icu4c/source/i18n/calendar.cpp
FrankYFTang: https://stackoverflow.com/questions/1453876/why-does-strncpy-not-null-terminate suggest set the dec to empty string then use strncat instead

FrankYFTang: does not work. try to use -Wno-stringop-truncation
 instead

icu4c/source/common/ubiditransform.cpp
```diff
@@ -499,7 +499,8 @@ ubiditransform_transform(UBiDiTransform *pBiDiTransform,
     /* Checking for U_SUCCESS() within the loop to bail out on first failure. */
     for (action = pBiDiTransform->pActiveScheme->actions; *action && U_SUCCESS(*pErrorCode); action++) {
         if ((*action)(pBiDiTransform, pErrorCode)) {
-            if (action + 1) {
+            const UBiDiAction * next = action+1;
```
aheninger, line 502: It looks like the original code may be incorrect; `(action + 1)` can never be null here. Probably what was intended was `if (*(action + 1)) {`, but this would change the behavior, so check with someone who knows this code.

icu4c/source/test/intltest/transtst.cpp
```diff
@@ -219,9 +219,10 @@ void TransliteratorTest::TestInstantiation() {
     assertSuccess("count()", ec);
     UnicodeString name;
     for (int32_t i=0; i<n; ++i) {
-        const UnicodeString& id = *avail->snext(ec);
+        const UnicodeString* snext = avail->snext(ec);
+        const UnicodeString& id = *snext;
```
aheninger, line 223: Move this line that dereferences `snext` after the checking for `snext!=nullptr`.

aheninger (changes requested):

PR #2356, opened 2023-03-04
ICU-22265 Update PersonNameFormatter and its associated classes so th…

richgillam:
…at the behavior matches that of the

PersonNameFormatter in CLDR.  Added a new test that tests the ICU PersonNameFormatter against a comprehensive set of test results from the CLDR PersonNameFormatter.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22265


richgillam:
This PR includes a test designed to test ICU PersonNameFormatter against the files Mark created in `cldr/common/testData/personNameTest`, in the hopes of making sure the ICU PersonNameFormatter and the CLDR PersonNameFormatter do the same thing.  This PR also includes all the changes I needed to make to get the new test to pass (some of those changes were bringing the ICU stuff up to date with changes on the CLDR side, and some were just fixing preexisting bugs on the ICU side), plus new unit tests for some of the new logic and some general cleanup (including a few changes Mark asked for).

There are still some important issues:
- I'm skipping five files in `cldr/common/testData/personNameTest`: The `dsb` and `hsb` locales exist in ICU, but most of the person-name data is missing from them.  The `gaa`, `syr` and `lij` locales didn't have files in `data/locales` in ICU at all (although a couple of them did have files in the other directories).  I checked to make sure this wasn't just in my local repo-- I didn't see these things in the main Unicode repo either.
- In `tk.txt`, I'm getting a bunch of errors where the field order is wrong, and I don't know why.  As far as I can tell, this locale wants GN-first ordering for all locales, and yet the test file's expected results are all in SN-first order.  I don't understand what's going on here.
- In `to.txt`, we're getting weird results in a number of test cases because the locale data file is incomplete and is inheriting some of its name formatting patterns from `root`.  I don't know whether this is a case of some stuff not making it from CLDR to ICU, or whether there's a difference in resource fallback behavior between CLDR and ICU.
- I'm not sure I have the foreign-space-replacement logic right.  All of the stuff in `personNameTest` seems to work now, but some of my existing unit tests didn't.  Take a close look at the changes in `PersonNameFormatterTest.TestNameSpacing()`.  I'm particularly concerned about the "Albert Einstein" example.  There isn't anything in `personNameTest` that looks like this, and I don't know if I'm deviating from CLDR or if CLDR has this wrong too.  Either way, I think we're going to need to fix it.
- Mark had a list of demands at the top of `tools/cldr-code/src/test/java/org/unicode/cldr/unittest/CheckPersonNamesTest.java`.  I did my best to address them, but might have missed a few.  If there's anything else you still want, please comment in this PR.

macchiati:
Thanks Rich! I'll look it over this weekend.

I think what's going on in the funny locales is that we remove the
provisional data in ICU, so the test generator sees different data. Let me
think of the best way to handle that; I'm thinking maybe we just skip
locales that have it.

"a list of demands"... (I guess you meant the French 'demander'). thanks
for that too.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameFormatterTest.java
```diff
@@ -355,22 +364,34 @@ public void TestNameSpacing() {
...
+                { "zh_CN", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "宮崎·駿" },
             }),
 
             // if the formatter locale doesn't use spaces and the name's locale does, use the name locale's formatter,
             // but if the name is still using the formatter locale's script, use the native formatter's
             // "foreign space replacement" character instead of spaces
             new NameAndTestCases("locale=en_US,given=Albert,surname=Einstein", new String[][] {
-                { "ja_JP", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "Albert Einstein" },
-                { "zh_CN", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "Albert Einstein" },
+// NOTE TO REVIEWERS: These two tests fail with "Albert・Einstein", which is clearly wrong.  The files in
```
macchiati, line 374: Good catch.

I think what is happening is if the name is in a different script, it formats with a different locale's formatter. That is in the LDML spec but not in CLDR's tooling, because for gathering data we don't need that.

So you need to add the logic to:
1. get the script of a PersonName
2. reroute PersonNames with a different script to a different locale.

I think this is clear in the spec, but if not we can discuss.

macchiati (changes requested):
Looks really good. There is only one area that I see that needs to be fixed.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameFormatterTest.java
```diff
@@ -355,7 +364,7 @@ public void TestNameSpacing() {
             // the native formatter
             new NameAndTestCases("locale=ja_JP,given=駿,surname=宮崎", new String[][] {
                 { "ja_JP", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "宮崎駿" },
-                { "zh_CN", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "宮崎 駿" },
+                { "zh_CN", "LONG",   "REFERRING",  "FORMAL",   "DEFAULT", "", "宮崎駿" },
```
richgillam, line 367: I want to call your attention specifically to this test case.  The earlier change to always use the formatter's own patterns and to use `nativeSpaceReplacement` or `foreignSpaceReplacement` messed up this example, with my getting back "宮崎・駿".  And, of course, we were also getting the dot in "Albert Einstein".  I fixed "Albert Einstein" to not get the dot, but the logic I added (using the native formatter for the name) doesn't apply to this case, which still got the dot.  I changed the `hasSameScript()` method so that Japanese and Chinese would return true, and that got me "宮崎駿", with no dot and also no space.  I don't think there's a good way to get the space in this ("宮崎 駿"), and I'm not sure we want it.  But if we do, there's more work to do here.  Please let me know what you think.

richgillam (commented):
I restored some of the logic I had before, but modified it a bit.  This is similar to what I outlined in my earlier comment, but not exactly the same: If the formatter locale doesn't use spaces, the name locale does, and the actual name is using a different script from the formatter locale, we use the formatter for the name's locale.  This passes everything in this unit test and everything in the exhaustive test, with the exception of the one case called out below.  My own hunch is that what we were getting for that locale before was wrong, but I'd like you all to confirm or refute that.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -142,12 +138,23 @@ public String formatToString(PersonName name) {
             result = getBestPattern(snFirstPatterns, name).format(name);
         }
 
+        // if the formatter locale doesn't use spaces, the name locale does, and the name data is not in
+        // the formatter locale's script, throw out the result, create a formatter for the name's locale,
+        // and use THAT to format the name
+        boolean nameLocaleUsesSpaces = !LOCALES_THAT_DONT_USE_SPACES.contains(nameLocale.getLanguage());
+        boolean useNativeFormatter = !formatterLocaleUsesSpaces && nameLocaleUsesSpaces && !scriptMatchesLocale(result, this.locale);
```
macchiati, line 145: only check for !scriptMatches

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -142,12 +138,23 @@ public String formatToString(PersonName name) {
...
+            result = nativeFormatter.formatToString(name);
+        }
+
         // if either of the space-replacement characters is something other than a space,
         // check to see if the name locale's script matches the formatter locale's script.
         // If they match, replace all spaces with the native space-replacement character,
         // and if they don't, replace all spaces with the foreign space-replacement character
-        if (!nativeSpaceReplacement.equals(" ") || !foreignSpaceReplacement.equals(" ")) {
-            if (hasSameScript(nameLocale, this.locale)) {
+        if (!useNativeFormatter && (!nativeSpaceReplacement.equals(" ") || !foreignSpaceReplacement.equals(" "))) {
```
macchiati, line 156: optional: use } else if (!nativeSpaceReplacement.equals(" ") {...

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameFormatterTest.java
richgillam: We discussed this in the Person Names meeting, and we all agreed this is indeed the result we want.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -128,6 +128,13 @@ public String toString() {
     public String formatToString(PersonName name) {
         // TODO: Should probably return a FormattedPersonName object
 
+        if (!nameScriptMatchesLocale(name, this.locale)) {
```
macchiati, line 131: It looks good to me, and I can align the spec with it. 

The hack where we say that ja == zh we can document, and even point to the ticket

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -310,40 +306,48 @@ private Locale getNameLocale(PersonName name) {
...
-    private boolean hasCommonScript(Locale nameLocale, Locale formatterLocale) {
-        int[] nameLocaleScripts = UScript.getCode(nameLocale);
-        int[] formatterLocaleScripts = UScript.getCode(formatterLocale);
-
-        for (int nameScript : nameLocaleScripts) {
-            for (int formatterScript : formatterLocaleScripts) {
-                if (nameScript == formatterScript) {
-                    return true;
-                }
+    private boolean nameScriptMatchesLocale(PersonName name, Locale formatterLocale) {
```
macchiati, line 311: I'll change the documentation to align with this.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -310,40 +306,48 @@ private Locale getNameLocale(PersonName name) {
...
-     * so it probably won't work right on mixed-script strings.
+     * Returns true if the two locales should be considered equivalent for space-replacement purposes.
      */
-    private boolean scriptMatchesLocale(String s, Locale locale) {
-        int[] localeScripts = UScript.getCode(locale);
-        int stringScript = UScript.COMMON;
-        for (int i = 0; stringScript == UScript.COMMON && i < s.length(); i++) {
-            char c = s.charAt(i);
-            stringScript = UScript.getScript(c);
+    private boolean localesMatch(Locale nameLocale, Locale formatterLocale) {
```
macchiati, line 337: And this.

macchiati (dismissed):

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: I'm not sure which ticket you're referring to, but that all sounds fine to me.  As I said, if you've got a better idea, I'm all ears.  Not only is the hack ugly, but it ignores things (like `yue`) that we'd probably also want to treat as the same.

One alternative I thought about was to go back to a more conservative reading of "same script" that would cause `ja` and `zh` to compare as different (because `Jpan` wouldn't match `Hans` or `Hant`), and then a Chinese name formatted in Japanese would just end up using the Chinese formatter.  I don't know if this would have undesirable side effects on other locales, though.  (The different flavors of Chinese might also be a problem, but I'm guessing their patterns are all pretty much the same...)  Of course, we can file a ticket (or maybe that's the ticket you're talking about) and revisit this post 43/73.

richgillam:
>It looks good to me, and I can align the spec with it.

Thanks a lot, Mark!  I just squashed my commits; would you mind rubber-stamping the squashed version?

macchiati (approved):
sure

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
macchiati: In https://unicode-org.atlassian.net/browse/CLDR-16457, there is currently proviso for which locales have native vs foreign spacing behavior. I just added another section on having the "use other formatter" be data-driven; needs more work, but that way we could have more customized behavior.

richgillam: Okay.  Sounds good.  (Thanks for the final stamp, by the way!)

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/ExhaustivePersonNameFormatterTest.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.*;
+
+/**
+ * This is a test designed to parse the files generated by GeneratePersonNameTestData.java in
+ * the CLDR project.  It takes one command-line parameter-- the path to the directory that
+ * contains the test files (common/testData/personNameTest in the CLDR source tree).
+ * This isn't set up as a unit test because of the dependency on the CLDR files (I didn't
+ * want to copy all of those over into the ICU tree) and because I thought the test would
+ * take too long to run.
+ */
```
pedberg-icu, line 19: There are several other test files that are copied from the CLDR tree to the ICU tree as part of the CLDR-ICU integration process, soothes would not be a problem. See the "copy-cldr-testdata" target in `tools/cldr/build.xml`

richgillam:
> Belated LGTM. I had a few questions/comments, nothing that would have blocked approval.

Thanks for looking at this!  Those are all good comments.  Since I already merged, would you mind filing a new ticket to clean up that stuff?

pedberg-icu:
> > Belated LGTM. I had a few questions/comments, nothing that would have blocked approval.
> 
> Thanks for looking at this! Those are all good comments. Since I already merged, would you mind filing a new ticket to clean up that stuff?

Sure, I will file a ticket for those.

PR #2365, opened 2023-03-14
ICU-22285 omit the gb2312 & big5han collation tailorings by default

markusicu:
This reduces the size of coll/zh.res from 918660 bytes to 682904 bytes.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22285
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [x] API docs and/or User Guide docs changed or added, if applicable

PR #2371, opened 2023-03-15
ICU-22285 how to include gb2312 & big5han collation tailorings

markusicu:
Explains how to undo the omit-by-default change from PR #2365.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22285

### DIFF
--- a/docs/userguide/icu_data/buildtool.md
+++ b/docs/userguide/icu_data/buildtool.md
@@ -104,6 +104,27 @@ The default configuration is equivalent to a filter file like this:
       ]
     }
 
+If you do want to include these collation tailorings,
+you can configure ICU with a filter file which has the opposite effect:
+
+    {
+      "resourceFilters": [
+        {
+          "categories": [
+            "coll_tree"
+          ],
+          "rules": [
+            "+/collations/big5han",
+            "+/collations/gb2312han"
+          ]
+        }
+      ]
+    }
+
+Alternatively, if you want to make sure to include *all* collation tailorings,
+you can use a single filter rule like `+/collations` or `+/collations/*`
+rather than listing each tailoring by its full, explicit path.
+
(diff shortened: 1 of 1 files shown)


docs/userguide/icu_data/buildtool.md
```diff
@@ -104,6 +104,23 @@ The default configuration is equivalent to a filter file like this:
...
+
+    {
+      "resourceFilters": [
+        {
+          "categories": [
+            "coll_tree"
+          ],
+          "rules": [
+            "+/collations/big5han",
+            "+/collations/gb2312han"
```
sffc, line 118: Does `+/collations/*` work?

I assume you did this because `+/collations` does not work?

sffc: If the user wants to include "all" collations, they should get them all, even if ICU starts suppressing even more of them in the future

markusicu: The default is to include everything, with this new exception. The filter that I provided here undoes the exception. I think it makes sense as is, including “If you do want to include **these** collation tailorings, ...” (emphasis for comment).

markusicu: I just verified that just `+/collations` works as well. I wouldn't replace what I have, but I could add that as an alternative, with an explanatory paragraph. Which way do you want to go?

markusicu: I just amended the docs to add a paragraph with alternate rules but without spelling out the whole filter file with each of those alternatives.

If we can't settle this before the ICU 73 code freeze, then we need to punt it to a post-73 doc fix.

sffc (dismissed):

sven-oly (approved):

PR #2376, opened 2023-03-16
ICU-10297 Add tests showing that display name bug in ticket no longer exists

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-10297

Add tests demonstrating that the display name bug in the ticket no longer exists.


richgillam (approved):
I don't know what they were getting before, but this all looks reasonable.

pedberg-icu:
> I don't know what they were getting before, but this all looks reasonable.

"they" is me (Reporter), but what was happening before (several releases ago) was that if we requested a zh_Hant dialect name and it did not have one for the desired item but zh did, we would get the zh version, instead of falling back to the zh_Hant standard version i.e. "language (region)". Now the fallback seems to be working correctly.

PR #2385, opened 2023-03-19
ICU-22319 Fix number range semanticallyEquivalent

sffc:
There was a bug that was sometimes causing semantically equivalent modifiers to not be recognized as semantically equivalent. It was checking for pointer equality instead of semantic equality.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22319

### DIFF
--- a/icu4c/source/i18n/number_modifiers.cpp
+++ b/icu4c/source/i18n/number_modifiers.cpp
@@ -60,6 +60,41 @@ Modifier::Parameters::Parameters(
     const ModifierStore* _obj, Signum _signum, StandardPlural::Form _plural)
         : obj(_obj), signum(_signum), plural(_plural) {}
 
+bool Modifier::semanticallyEquivalent(const Modifier& other) const {
+    Parameters paramsThis;
+    Parameters paramsOther;
+    getParameters(paramsThis);
+    other.getParameters(paramsOther);
+    if (paramsThis.obj == nullptr && paramsOther.obj == nullptr) {
+        return strictEquals(other);
+    } else if (paramsThis.obj == nullptr || paramsOther.obj == nullptr) {
+        return false;
+    }
+    for (size_t i=0; i<SIGNUM_COUNT; i++) {
+        auto signum = static_cast<Signum>(i);
+        for (size_t j=0; j<StandardPlural::COUNT; j++) {
+            auto plural = static_cast<StandardPlural::Form>(j);
+            auto mod1 = paramsThis.obj->getModifier(signum, plural);
+            auto mod2 = paramsOther.obj->getModifier(signum, plural);
+            if (mod1 == mod2) {
+                // Equal pointers
+                continue;
+            } else if (mod1 == nullptr || mod2 == nullptr) {
+                // One pointer is null but not the other
+                return false;
+            } else if (!mod1->strictEquals(*mod2)) {
+                // The modifiers are NOT equivalent
+                return false;
+            } else {
+                // The modifiers are equivalent
+                continue;
+            }
+        }
+    }
+    return true;
+}
+
+
 ModifierStore::~ModifierStore() = default;
 
 AdoptingSignumModifierStore::~AdoptingSignumModifierStore()  {
@@ -110,7 +145,7 @@ void ConstantAffixModifier::getParameters(Parameters& output) const {
     UPRV_UNREACHABLE_EXIT;
 }
 
-bool ConstantAffixModifier::semanticallyEquivalent(const Modifier& other) const {
+bool ConstantAffixModifier::strictEquals(const Modifier& other) const {
     auto* _other = dynamic_cast<const ConstantAffixModifier*>(&other);
     if (_other == nullptr) {
         return false;
@@ -197,13 +232,10 @@ void SimpleModifier::getParameters(Parameters& output) const {
     output = fParameters;
 }
 
-bool SimpleModifier::semanticallyEquivalent(const Modifier& other) const {
+bool SimpleModifier::strictEquals(const Modifier& other) const {
     auto* _other = dynamic_cast<const SimpleModifier*>(&other);
     if (_other == nullptr) {
         return false;
-    }
-    if (fParameters.obj != nullptr) {
-        return fParameters.obj == _other->fParameters.obj;
     }
     return fCompiledPattern == _other->fCompiledPattern
         && fField == _other->fField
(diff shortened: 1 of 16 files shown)


sffc:
@richgillam Assigned this to you to review since it is in number formatting code. No rush because I don't expect this to land in ICU 73 unless the team specifically wants it in there.

icu4c/source/i18n/number_modifiers.cpp
```diff
@@ -60,6 +60,41 @@ Modifier::Parameters::Parameters(
     const ModifierStore* _obj, Signum _signum, StandardPlural::Form _plural)
         : obj(_obj), signum(_signum), plural(_plural) {}
 
+bool Modifier::semanticallyEquivalent(const Modifier& other) const {
+    Parameters params0;
+    Parameters params1;
```
markusicu, line 65: suggestion: this function be more readable with more specific names of these variables, for example thisParams vs. otherParams

icu4c/source/i18n/number_modifiers.cpp
```diff
@@ -60,6 +60,41 @@ Modifier::Parameters::Parameters(
...
+        return strictEquals(other);
+    } else if (params0.obj == nullptr || params1.obj == nullptr) {
+        return false;
+    }
+    for (size_t i=0; i<SIGNUM_COUNT; i++) {
+        auto signum = static_cast<Signum>(i);
+        for (size_t j=0; j<StandardPlural::COUNT; j++) {
+            auto plural = static_cast<StandardPlural::Form>(j);
+            auto mod1 = params0.obj->getModifier(signum, plural);
+            auto mod2 = params1.obj->getModifier(signum, plural);
```
markusicu, line 78: similarly, consider using prefixes or suffixes like "this" and "other" rather than 1 and 2. especially when mod1 is from params0 and mod2 is from params1.....

icu4j/main/classes/core/src/com/ibm/icu/number/NumberRangeFormatterImpl.java
```diff
@@ -433,11 +435,48 @@ Modifier resolveModifierPlurals(Modifier first, Modifier second) {
...
         Modifier mod = firstParameters.obj.getModifier(firstParameters.signum, resultPlural);
         assert mod != null;
         return mod;
     }
 
+    /**
+     * Returns whether this Modifier is *semantically equivalent* to the other Modifier;
+     * in many cases, this is the same as equal, but parameters should be ignored.
+     *
+     * In ICU4C, this is a function on the Modifier class.
```
markusicu, line 447: Why is it not a function on the Modifier class in Java as well?

markusicu (commented):
With the new & improved semanticallyEquivalent(), would it be possible + make sense to add the assertions back using that, like you had tried initially?

----
Nice, thorough work, thank you!
I agree that this should go into ICU 74 not 73.

icu4j/main/classes/core/src/com/ibm/icu/number/NumberRangeFormatterImpl.java
markusicu: We are requiring Java 8 since ICU 72 :-)
If this is awkward on the interface, then maybe turn that into a base class?
Or leave it as is?

Just noticed that it's different from C++, and therefore requires changing more call sites.

icu4c/source/i18n/number_modifiers.cpp
richgillam: +1

icu4c/source/i18n/number_modifiers.cpp
```diff
@@ -110,7 +145,7 @@ void ConstantAffixModifier::getParameters(Parameters& output) const {
     UPRV_UNREACHABLE_EXIT;
 }
 
-bool ConstantAffixModifier::semanticallyEquivalent(const Modifier& other) const {
+bool ConstantAffixModifier::strictEquals(const Modifier& other) const {
```
richgillam, line 148: Just out of curiosity, why the name change?  And is there a "lenient" or "approximate" equals function?

(**UPDATED TO ADD:**  Oh, I see.  You still have a `semanticallyEquivalent()` method as well.  You commented at the declaration site about how they're different, but you might also want to say something there about who needs one and who needs the other and why.)

While I'm at it, what are your thoughts on using `operator==()` for this?  Is it already being used to do something else, or does it have other problems that make it impossible or inconvenient?

richgillam (dismissed):
I agree with Markus's comments and think you might need a little more documentation, but this looks good.

icu4c/source/i18n/number_modifiers.cpp
sffc: I err away from `operator==` because that doesn't work in Java and I think it's one of the things that causes bugs like this in the first place, since it's not obvious during a port whether `==` invokes a special impl.

richgillam: Yeah, I thought your reasoning would be something like that.  That's fine with me; I have no objection to what you're doing here.

icu4c/source/i18n/number_modifiers.cpp
sffc: There is no "this" and "other" because this is a static function. There is A and B, Left and Right, or 0 and 1. I chose 0 and 1.

sffc: Clarification: in Java it is a static function. In C++ it is a member function.

sffc:
I rebased this on main. A bunch of Java files moved around.

markusicu:
> I rebased this on main. A bunch of Java files moved around.

recte: *all* Java files have moved around ;-)

icu4j/main/classes/core/src/com/ibm/icu/number/NumberRangeFormatterImpl.java
markusicu: ping

sffc: ok I changed it to Java 8 `default`

icu4c/source/i18n/number_modifiers.cpp
sffc: Okay now that this is no longer a static function, I changed to This/Other in both C++ and Java

sffc:
@markusicu Once you approve I will squash.

markusicu (approved):

sffc:
Force-pushed to bump CI. When it passes please feel free to merge

PR #2404, opened 2023-03-29
ICU-22340 Fix it so that UNUM_NUMBERING_SYSTEM again always returns a RuleBasedNumberFormat.

richgillam:
This fixes a regression from https://github.com/unicode-org/icu/pull/2139

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22340

### DIFF
--- a/icu4c/source/i18n/unum.cpp
+++ b/icu4c/source/i18n/unum.cpp
@@ -60,7 +60,6 @@ unum_open(  UNumberFormatStyle    style,
     case UNUM_CURRENCY_ACCOUNTING:
     case UNUM_CASH_CURRENCY:
     case UNUM_CURRENCY_STANDARD:
-    case UNUM_NUMBERING_SYSTEM:
         retVal = NumberFormat::createInstance(Locale(locale), style, *status);
         break;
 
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/unum.cpp
```diff
@@ -113,6 +112,22 @@ unum_open(  UNumberFormatStyle    style,
...
+
+    case UNUM_NUMBERING_SYSTEM: {
+        // if the locale ID specifies a numbering system, go through NumberFormat::createInstance()
+        // to handle it properly (we have to specify UNUM_DEFAULT to get it to handle the numbering
+        // system, but we'll always get a RuleBasedNumberFormat back); otherwise, just go ahead and
+        // create a RuleBasedNumberFormat ourselves
+        UErrorCode err = U_ZERO_ERROR;
+        Locale localeObj(locale);
+        char keywordBuffer[20];
+        int32_t keywordLength = localeObj.getKeywordValue("numbers", keywordBuffer, 20, err);
```
markusicu, line 124: optional:
You could get false failures here from over-long keyword values.
In newer code, we generally use a CharStringByteSink and [this version of getKeywordValue()](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1Locale.html#aef599483560e19d4f0ac8a65e0a4ddd4).

sffc: Does this work?

```cpp
int32_t keywordLength = localeObj.getKeywordValue("numbers", nullptr, 0, localStatus);
// ignore localStatus: expected to be U_BUFFER_OVERFLOW_ERROR
```

richgillam: I like Shane's option better, since I don't need the actual value, and it seems to work.

icu4c/source/i18n/unum.cpp
```diff
@@ -113,6 +112,21 @@ unum_open(  UNumberFormatStyle    style,
...
         break;
+
+    case UNUM_NUMBERING_SYSTEM: {
+        // if the locale ID specifies a numbering system, go through NumberFormat::createInstance()
+        // to handle it properly (we have to specify UNUM_DEFAULT to get it to handle the numbering
+        // system, but we'll always get a RuleBasedNumberFormat back); otherwise, just go ahead and
+        // create a RuleBasedNumberFormat ourselves
+        UErrorCode localErr = U_ZERO_ERROR;
+        Locale localeObj(locale);
+        int32_t keywordLength = localeObj.getKeywordValue("numbers", nullptr, 0, localErr);
```
sffc, line 123: Thought: It's weird coming back from Rust that `localErr` has no `&` on it and yet it is being passed by reference. In Rust it is more explicit at the call site whether things are passed by reference or value.

sffc (approved):

icu4c/source/i18n/unum.cpp
richgillam: Yup.  That's C++ for you...

pedberg-icu (approved):

PR #2408, opened 2023-03-29
ICU-22341 commit checker: umbrella, explanations

srl295:
- support Umbrella ticket type (no commits)
- explain why commit policy was applied

##### Checklist

### DIFF
--- a/tools/commit-checker/check.py
+++ b/tools/commit-checker/check.py
@@ -24,10 +24,24 @@ from commit_metadata import CommitMetadata
 ICUCommit = namedtuple("ICUCommit", ["issue_id", "commit"])
 
 class CommitWanted(Enum):
+    """Defines the commit policy given a ticket"""
     REQUIRED = 1
     OPTIONAL = 2
     FORBIDDEN = 3
     ERROR = 4
+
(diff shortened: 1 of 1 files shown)


tools/commit-checker/check.py
```diff
@@ -486,7 +547,6 @@ def run_section(section, out) -> int:
         elif section == CLOSED_ILLEGAL_RESOLUTION:
             print("Tip: Fixed tickets should have resolution 'Fixed by Other Ticket' or 'Fixed'.", file=out)
```
AEApple, line 548: This doesn't seem right to me? There shouldn't be any commits against tickets with resolution 'Fixed by Other Ticker', so it's not clear to me why this is being recommended at the same time as the resolution 'Fixed'.

srl295: it's kind of an ICU tip, I'm wondering about just removing this tip entirely because the new explanations will be much more on point.

sffc (approved):

tools/commit-checker/check.py
AEApple: Up to you :)

PR #2006, opened 2022-02-25
ICU-21802 Fix error where strict date parsing could succeed on certai…

richgillam:
…n malformed input strings.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21802
- [ ] Required: The PR title must be prefixed with a JIRA Issue number. 
- [ ] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [ ] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [ ] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2409, opened 2023-03-30
ICU-22337 restore strict parsing length tolerance for non-abutting numeric date fields

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22337

ICU-21802 was filed about a valid issue with strict parsing for a formatter set to yyyyMMdd (i.e. with numeric fields that abut). The fix in [PR-2006](https://github.com/unicode-org/icu/pull/2006) did fix the problem in the ticket, but also affected strict parsing of non-abutting numeric date fields, causing them to require exact length match, though they have traditional allowed different lengths  even for strict parsing. This causes compatibility problems for clients

The fix here preserves the intended fix for ICU-21802, while restoring the traditional (pre-ICU-71) behavior for parsing non-abutting numeric date fields. In addition to fixing https://unicode-org.atlassian.net/browse/ICU-22337, this also fixes https://unicode-org.atlassian.net/browse/ICU-22259 and likely other tickets as well.

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -3818,7 +3818,7 @@ int32_t SimpleDateFormat::subParse(const UnicodeString& text, int32_t& start, ch
         src = &text;
     }
     parseInt(*src, number, pos, allowNegative,currentNumberFormat);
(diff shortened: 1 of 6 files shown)


richgillam (approved):
Looks great.

PR #2410, opened 2023-03-31
ICU-7747 add meson.build files

srl295:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-7747

see also #2249 

fyi @eli-schwartz @neheb

### DIFF
--- a/.github/workflows/meson_ci.yml
+++ b/.github/workflows/meson_ci.yml
@@ -7,9 +7,9 @@ name: Meson CI
 
 on:
   push:
-    branches:
-      - main
-      - 'maint/maint*'
+    # branches:
+    #   - main
+    #   - 'maint/maint*'
   pull_request:
     branches: '**'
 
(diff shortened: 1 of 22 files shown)


srl295:
- [ ] doesn't build without an icudt prebuild data file… not surprisingly

eli-schwartz:
Isn't this actually https://unicode-org.atlassian.net/browse/ICU-22343 ?

srl295:
> Isn't this actually https://unicode-org.atlassian.net/browse/ICU-22343 ?

no?

but it is meson instead of cmake so might need a different ticket

eli-schwartz:
Ironically I then made an even worse mistake trying to make this link cleaner: https://unicode-org.atlassian.net/plugins/servlet/mobile?originPath=%2Fbrowse%2FICU-22343#issue/ICU-21826 ?

srl295:
> Ironically I then made an even worse mistake trying to make this link cleaner: https://unicode-org.atlassian.net/plugins/servlet/mobile?originPath=%2Fbrowse%2FICU-22343#issue/ICU-21826

That takes me on a very wild and wonderful journey and ends up on the mobile site.

We can also just call it `ICU-21826` here => ICU-21826

and you're right.  Will fix.

neheb:
I'll just add this here: https://github.com/openwrt/packages/pull/20628

IIRC the normal build doesn't build in parallel either.

PR #2337, opened 2023-02-22
ICU-22220 update root collation from CLDR 43

markusicu:
Partial update only for the root collation.
See
- CLDR-15946 “Treat quote marks as equivalent when strength=UCOL_PRIMARY”
- https://github.com/unicode-org/cldr/pull/2691

The related changes to tailorings were already integrated in the earlier PR #2304.

This update is for the root collation,
which is handled by different tools than the locale data updates.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22220
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2411, opened 2023-04-06
ICU-22221 update root collation again from CLDR 43

markusicu:
Partial update only for the root collation.
Follow-up on PR #2337

See
- CLDR-15946 “Treat quote marks as equivalent when strength=UCOL_PRIMARY”
- second pull request https://github.com/unicode-org/cldr/pull/2833

This update is for the root collation,
which is handled by different tools than the locale data updates.

FYI @hsivonen

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22221


macchiati (approved):

markusicu:
@pedberg-icu I think you can create a "release" tag now and pull that into CLDR.

pedberg-icu:
> @pedberg-icu I think you can create a "release" tag now and pull that into CLDR.

Yes, the tag will have the release version but also a date (since this is not the final ICU release).

PR #2415, opened 2023-04-10
ICU-21964 use a single LICENSE file

srl295:
- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21964

c.f. https://github.com/unicode-org/cldr/pull/2849

### Why

- more standard filename, visible at the top level in GitHub
- reduce maintenance

### What's inside

- make the icu4c and icu4j LICENSE files symlinks
- Some URLs won't work until this is merged

### confirmed
- ICU4C make dist uses `git archive` which copies the symlink into a file
- ICU4J ant icu4jSrcJar  makes a `.jar` file including LICENSE as a file.
- Windows supports symlinks too, [see](https://stackoverflow.com/a/52097145/185799)


markusicu (changes requested):
The right ticket for this seems to be ICU-21964 which is assigned to @yumaoka.
Please change the commit message and PR title.
I set Yoshito as the PR assignee (=main reviewer).

LICENSE
markusicu, line 1: Just to be sure: This is a verbatim copy of either of the previous files, which should have been identical, right?

docs/userguide/icu/release.md
```diff
@@ -30,4 +30,4 @@ Changes in previous releases can also be found the main [ICU download page](http
 
 ## License Information
 
-The ICU projects (ICU4C and ICU4J) are hosted by the [Unicode Consortium](http://www.unicode.org/). The ICU binary and source files are distributed under the [UNICODE DATA FILES AND SOFTWARE LICENSE](http://www.unicode.org/copyright.html). The full copy of the license and third party software licenses are available in [LICENSE](https://github.com/unicode-org/icu/blob/main/icu4j/main/shared/licenses/LICENSE) file included in this package.
+The ICU projects (ICU4C and ICU4J) are hosted by the [Unicode Consortium](http://www.unicode.org/). The ICU binary and source files are distributed under the [UNICODE DATA FILES AND SOFTWARE LICENSE](http://www.unicode.org/copyright.html). The full copy of the license and third party software licenses are available in [LICENSE](https://github.com/unicode-org/icu/blob/main/LICENSE) file included in this package.
```
markusicu, line 33:
```suggestion
The ICU projects (ICU4C and ICU4J) are hosted by the [Unicode Consortium](https://www.unicode.org/). The ICU binary and source files are distributed under the [Unicode Data Files and Software License](https://www.unicode.org/license.txt). The full copy of the license and third party software licenses are available in the [LICENSE](https://github.com/unicode-org/icu/blob/main/LICENSE) file included in this package.
```

icu4j/maven-build/maven-icu4j/pom.xml
```diff
@@ -254,7 +254,7 @@
                   <configuration>
                       <executable>cp</executable>
                       <arguments>
-                        <argument>../../../icu4c/LICENSE</argument>
+                        <argument>../../main/shared/licenses/LICENSE</argument>
```
markusicu, line 257: why point to the symlink, not to the actual file in root?

markusicu:
PS: @srl295 thanks for doing this!

LICENSE
srl295: Yes

icu4j/maven-build/maven-icu4j/pom.xml
srl295: > why point to the symlink, not to the actual file in root?

Ooh! This one. 

Because you might be working from an icu4j sources jar instead of out of the whole git repo.

markusicu: Makes sense. Could you please add a comment saying so, to prevent the next person from asking, or from changing it?

markusicu (approved):
lgtm tnx!

yumaoka:
@srl We used to include LICENSE file also in library jars (e.g. icu4j.jar). Now, with this change, LICENSE file is included in a jar, but with the content `../../../../LICENSE`.

yumaoka:
@srl295 You're right. It looks this is a problem on system not supporting symlink. I tried both Windows and Ubuntu, and only Windows output had the problem.

yumaoka:
I'm also wondering if we need LICENSE file in each library jar file. We probably should check if we need to continue this.

srl295:
> I'm also wondering if we need LICENSE file in each library jar file. We probably should check if we need to continue this.

It's probably a good idea.   By the way common practice seems to be to put it in `META-INF/LICENSE` not in the root.

markusicu:
New ticket please with a problem description and maybe a proposed plan?

srl295:
@markusicu do we really need a new ticket or just to reopen this 74.1 ticket as still having trouble? I can do either

markusicu:
> @markusicu do we really need a new ticket or just to reopen this 74.1 ticket as still having trouble? I can do either

reopening is fine

tuffnatty:
Hello @srl295,

> * ICU4C make dist uses `git archive` which copies the symlink into a file

It may be true, but [the 74.2 src archive](https://github.com/unicode-org/icu/releases/download/release-74-2/icu4c-74_2-src.tgz) contains a dangling `icu/LICENSE` symlink pointing to nonexistant `../LICENSE`; build attempts fail at `make install`. The 74.1 src archive is OK, though; still, I've decided to report it here as the most related ticket.

srl295:
> Hello @srl295,
> 
> 
> 
> > * ICU4C make dist uses `git archive` which copies the symlink into a file
> 
> 
> 
> It may be true, but [the 74.2 src archive](https://github.com/unicode-org/icu/releases/download/release-74-2/icu4c-74_2-src.tgz) contains a dangling `icu/LICENSE` symlink pointing to nonexistant `../LICENSE`; build attempts fail at `make install`. The 74.1 src archive is OK, though; still, I've decided to report it here as the most related ticket.
> 
> 
> 
> P.S. Oh. I've found there's already a [ticket](https://unicode-org.atlassian.net/jira/software/c/projects/ICU/issues/ICU-22601). Please disregard.

Ticket is the right path. Thanks. !

srl295:
See #2749

PR #2422, opened 2023-04-11
ICU-22349 Change to use just -j value in make to "not limit the number of jobs that can run simultaneously"

FrankYFTang:
This should speed up our build process by specify a better -j value in make.

According to https://askubuntu.com/questions/1295713/make-use-of-j-option it is recommended to use twice as much as cores in a machine for the number of thread, which specified by the -j option of make during the build to better utilize the machine.

According to https://man7.org/linux/man-pages/man1/lscpu.1.html#:~:text=DESCRIPTION%20top,for%20easy%20readability%20by%20humans.  we can get the number of CPU from 'lscpu -p="CPU"'

MacOS is BSD base so we use "sysctl -a|grep "machdep.cpu.core_count"|cut -d':' -f2" to find out the number of CPU.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22349

### DIFF
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -185,7 +185,7 @@ jobs:
           cd icu4c/source;
           ICU_DATA_BUILDTOOL_OPTS=--include_uni_core_data ICU_DATA_FILTER_FILE=../../.github/lstm_for_th_my.json ./runConfigureICU --enable-debug --disable-release Linux -disable-layoutex;
           make clean;
-          make -j2 ICU4J_ROOT=../../../icu4j icu4j-data-install;
+          make -j ICU4J_ROOT=../../../icu4j icu4j-data-install;
           cd ../..
       - name: ICU4J
         run: |
@@ -217,7 +217,7 @@ jobs:
           cd icu4c/source;
           ICU_DATA_BUILDTOOL_OPTS=--include_uni_core_data ICU_DATA_FILTER_FILE=../../.github/adaboost.json CPPFLAGS=-DUCONFIG_USE_ML_PHRASE_BREAKING=1 ./runConfigureICU --enable-debug --disable-release Linux -disable-layoutex;
           make clean;
-          make -j2 ICU4J_ROOT=../../../icu4j icu4j-data-install;
+          make -j ICU4J_ROOT=../../../icu4j icu4j-data-install;
           cd ../..
       - name: ICU4J
         run: |
@@ -251,8 +251,7 @@ jobs:
           mkdir build;
           cd build;
           ../icu4c/source/runConfigureICU --enable-debug --disable-release Linux/gcc --prefix=$PREFIX --enable-tracing;
-          make -j2;
-          make -j2 check;
+          make -j check;
           ( cd ../icu4c/source/test/depstest && ./depstest.py ../../../../build/ );
           make install;
           PATH=$PREFIX/bin:$PATH make -C test/hdrtst check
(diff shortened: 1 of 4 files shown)


.github/workflows/icu_ci.yml
```diff
@@ -268,19 +270,34 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: ICU4C with clang
+      - name: Install doxygen
```
markusicu, line 273: FYI: Frank says that he split some large steps into smaller ones so that it's easier to figure out what fails (if something fails).

markusicu (commented):
nifty! :-)

srl295 (approved):
makes sense.  not sure if there are any special considerations for these VMs or not.

roubert:
Is there any reason for why it wouldn't be possible to instead use the `-l` flag and let `make` calculate the maximum number of concurrent jobs on its own? A simple `make -j -l90` would both be less complex and more likely to give good resouce utilization than counting available CPUs.

FrankYFTang:
I reads https://linux.die.net/man/1/make again
"
-j [jobs], --jobs[=jobs]
Specifies the number of jobs (commands) to run simultaneously. If there is more than one -j option, the last one is effective. If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously.

-l [load], --load-average[=load]
Specifies that no new jobs (commands) should be started if there are others jobs running and the load average is at least load (a floating-point number). With no argument, removes a previous load limit.
"

so... would it be even simplier if we just use "-j " (without a number after the -j nor a -l?)

echeran:
The original reason for limiting make commands to using only 2 cores is because the Github Actions VM Runners only have 2 cores: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

The main benefit right now from the change from `-j2` to `-j` might be to make things resiliently automatically use more cores if VM Runners ever increase their core count (who knows, maybe it could happen). Unless the "number of cores x2" trick works, I wouldn't expect performance to significantly improve due to the constraints of the VM Runner specs.

I would just check that the change doesn't _degrade_ performance (and I don't see why it would).

FrankYFTang:
> The original reason for limiting make commands to using only 2 cores is because the Github Actions VM Runners only have 2 cores

My understanding is -j n is to specify the number of jobs, not to specify the number of cores. 
If we use the same number of jobs as the core, what will happen is while each core read the .cpp file and hit the #include file, it will be blocked and start reading from the file, and the OS can schedule that core to pick up other thread to run other jobs, but there are no more jobs to be pick up. So all the core which blocked by IO will be just blocked there and idel until the IO complete. Specifying the number of thread = x 2 of the number of the core means for each core, while they got blocked by the IO, that core will have at least one more thread (job) to switch to, of course, it is also possible both jobs are all blocked by IO and still waiting for the network (in case of NFS) or disk to return the data and will seek for yet another thread to start but that chances is much less seems if we allocate x 2 cores of jobs some core are still busy parsing. 

I am not sure what will be the effect of -j or -l. But if we assume there are 2 cores always, then the right number after -j should be 4 not 2. Otherwise, while the parser read the many #include and keep reading those .h file, the core running the thread will be blocked and there are no thread the core can switch to and will be idel.

roubert:
> so... would it be even simplier if we just use "-j " (without a number after the -j nor a -l?)

I just thought that the reason for using `-j` with a number was that we wanted to put some kind of limit on it and therefore suggested using `-l` to set a limit instead.

FrankYFTang:
Here are the number for all the 

GHA CI icu_ci.yml build

Before: "make -j2":  18m 31s
just "make -j " 17m 40s

GHA CI Valgrind icu_valgrind.yml

Before  "make -j2":  30m 32s
just "make -j " 24m 45s

FrankYFTang:
PTAL

echeran (approved):
LGTM

markusicu (approved):
lgtm

PR #2431, opened 2023-04-17
ICU-22356 Revert ICU-21833 change for ures.h

xry111:
These usages of UChar are in inline functions for calling C API functions, so they should not be changed to char16_t or the code will fail to compile with `-DUCHAR_TYPE=`.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22356

### DIFF
--- a/icu4c/source/common/unicode/ures.h
+++ b/icu4c/source/common/unicode/ures.h
@@ -812,7 +812,7 @@ inline UnicodeString
 ures_getUnicodeString(const UResourceBundle *resB, UErrorCode* status) {
     UnicodeString result;
     int32_t len = 0;
-    const char16_t *r = ures_getString(resB, &len, status);
+    const UChar *r = ures_getString(resB, &len, status);
     if(U_SUCCESS(*status)) {
         result.setTo(true, r, len);
     } else {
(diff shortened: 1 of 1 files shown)


roubert:
The code here is explicitly C++ code, which is why I think the solution proposed in PR #2432 is more in line with the original intentions behind ICU-21833.

xry111:
> The code here is explicitly C++ code, which is why I think the solution proposed in PR #2432 is more in line with the original intentions behind [ICU-21833](https://unicode-org.atlassian.net/browse/ICU-21833).

IMO #2432 is an [align rule break](https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing) and leads to undefined behavior.  Or am I misunderstanding something here?

roubert:
> IMO #2432 is an [align rule break](https://en.cppreference.com/w/cpp/language/reinterpret_cast#Type_aliasing) and leads to undefined behavior.

To ensure that this doesn't become a problem, I updated that PR to instead use the ICU4C helper class `ConstChar16Ptr` to do the required typecasting in a safe way.

xry111:
Closing as superseded.

PR #2454, opened 2023-05-08
ICU-22384 Limit execution of performance tests

gnrunge:
Run the tests only on the main branch of the unicode-org/icu repository. This avoids diluting the performance charts with performance results from the maintenance branches. Also, the performance tests won't execute on forked directories anymore, on which they fail after execution anyway, thus using GitHub resources without purpose.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22384

### DIFF
--- a/.github/workflows/icu_merge_ci.yml
+++ b/.github/workflows/icu_merge_ci.yml
@@ -39,6 +39,8 @@ jobs:
 
   # Compile libraries used by all ICU4C performance tests.
   icu4c-store-perf-libs:
+    # Run performance tests only on the main branch of the ICU repository.
+    if: github.repository == 'unicode-org/icu' && github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -59,6 +61,8 @@ jobs:
 
   # Assorted ICU4C performance tests
   icu4c-performance-tests:
+    # Run performance tests only on the main branch of the ICU repository.
+    if: github.repository == 'unicode-org/icu' && github.ref == 'refs/heads/main'
     needs: icu4c-store-perf-libs
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


roubert (approved):
LGTM

FrankYFTang:
Can we use 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_runbranchesbranches-ignore

    branches:
      - 'main'

instead?

FrankYFTang:
Should we do
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

    if: github.event.pull_request.merged == true

gnrunge:
> Can we use
> 
> https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_runbranchesbranches-ignore
> 
> ```
> branches:
>   - 'main'
> ```
> 
> instead?

icu_merge_ci.yml has already 'main' among its branches enumeration to make sure that the actions run on main after each merge. The issue is that the performance tests currently are executed for the maintenance branches as well and also for the main branches of every forked directory, i.e. whenever a forked repository is synced with unicode-org/icu.

gnrunge:
> Should we do https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges
> 
> ```
> if: github.event.pull_request.merged == true
> ```

That doesn't add anything when GH Action run icu_merge_ci.yml on the main branch on unicode-org/icu given the ICU process. It may prevent execution of the performance tests on the main branch of the forked repositories upon synchronization (not sure about this).
However, the condition

```
github.repository == 'unicode-org/icu'
```

makes the intention of executing the job only on the base ICU repository much more obvious.

PR #2421, opened 2023-04-11
ICU-22349 Build cygwin with -j  to run multiple cores

FrankYFTang:
Trying -j in the make check to see would it speed up the build process in cygwin. See https://github.com/unicode-org/icu/pull/2422 for other changes.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22349
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [X] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2456, opened 2023-05-09
ICU-22349 Use -l with make -j to limit jobs based on load average.

roubert:
It has been proposed that make running too many parallel jobs recently has led to resource exhaustion in our CI, so that some kind of limit would be helpful to set.

The load average 2.5 limit choosen here is simply the limit used as an example in the make documentation, as we don't really have any way of picking an initial value that's certain to be better.

https://www.gnu.org/software/make/manual/html_node/Parallel.html

See #2421
See #2422

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22349

### DIFF
--- a/.ci-builds/.azure-exhaustive-tests.yml
+++ b/.ci-builds/.azure-exhaustive-tests.yml
@@ -62,7 +62,7 @@ jobs:
       lfs: true
       fetchDepth: 10
     - script: |
-        cd icu4c/source && ./runConfigureICU Linux && make -j2
+        cd icu4c/source && ./runConfigureICU Linux && make -j -l2.5
       displayName: 'Build'
     - script: |
         cd icu4c/source && make check-exhaustive
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -33,7 +33,7 @@ jobs:
       lfs: true
       fetchDepth: 10
     - script: |
-        cd icu4c/source && ./runConfigureICU Linux && make -j2 check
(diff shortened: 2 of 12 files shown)


roubert:
With these changes, both _gcc-10-stdlib14_ & _gcc-debug-build-and-test_, which recently have been repeatedly terminated before finishing, now pass.

As for the tests listed in ICU-22349 as targets for speed improvements, the current execution times with these changes are comparable (small enough that the differences might even be just noise) with the execution times for their last run at `HEAD`:

| Test | last | current |
| :--- | ---: | ---: |
| Cygwin GCC x86_64 Release | 34m | 36m |
| MSVC x64 ARM64 Release (VS 2019) + Distrelease | 23m | 27m |
| MSYS2 GCC x86_64 Release | 24m | 29m |

roubert:
This seems like it would be an improvement. What do you think?

robertbastian (approved):
If this is fixing existing breakages it looks fine, however it might be worth looking at the actual hardware we're running on instead of taking the 2.5 from the docs

robertbastian:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources for the `GHA CI` jobs, the Azure ones are custom.

robertbastian:
Fair enough

roubert:
@echeran What do you think?

echeran (approved):
LGTM. Agreed that this is an improvement for all of the above reasons that you stated (We keep the speed improvements from the original PR for ICU-22349 but we don't crash / thrash because have a controlled configuration of resource utilization)

PR #2458, opened 2023-05-11
ICU-22368 Reduce ~200K langInfo.res size by encode LSR into 32bits int.

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22368


FrankYFTang:
This change reduce the size of icudt73l.dat from 32206240 to 31999424 and reduced 206816 bytes

FrankYFTang:
Please review. thanks

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/localedistance/LocaleDistanceMapper.java
```diff
@@ -434,21 +437,90 @@ private static RbValue ofMapEntries(Map<String, String> map) {
...
+        return ret;
+    }
+    // This method is added only to support encodeToIntForResource()
+    // It only support [A-Z]{2}|001|143|419 and does not work for other cases.
+    // TODO(ftang) Remove after LSR.encodeToIntForResource is available to the tool.
+    static private int encodeRegionToInt(String region) {
+        assert region.length() >= 2;
+        assert region.length() <= 3;
+        // Do not have enough bits to store the all 1000 possible combination of \d{3}
+        // Only support what is needed in resource- "001", "143" and "419".
```
pedberg-icu, line 480: This is not the complete set. We also have for example 150 (in en_150). In RBNF data we used to have 003. Also I think it would be better to have at least an array of supported values, then just find the index of the correct element in that array (instead of the if-else sequence). And perhaps generate an error if you find a region code that is not in the list, because we could add a new numeric region at any time.

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -240,6 +241,82 @@ struct XLikelySubtagsData {
...
+    }
+    UnicodeString toRegion(int encoded) {
+        if (encoded == 0 || encoded == 1) {
+            return UNICODE_STRING_SIMPLE("");
+        }
+        encoded &= 0x00ffffff;
+        encoded /= 27 * 27 * 27;
+        encoded %= 27 * 27;
+        if (encoded < 27) {
+            switch (encoded) {
```
pedberg-icu, line 284: This way of decoding is fragile; it already does not support all numeric regions  codes used. I think LocaleDistanceMapper should write an array of supported numeric regions that this function can use for decoding.

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
```diff
@@ -89,4 +90,109 @@ public boolean equals(Object obj) {
...
+        encoded = (encoded >> 24) & 0x000000ff;
+        return UScript.getShortName(encoded);
+    }
+    private static String toRegion(int encoded) {
+        if (encoded == 0 || encoded == 1) return "";
+        encoded &= 0x00ffffff;
+        encoded /= 27 * 27 * 27;
+        encoded %= 27 * 27;
+        if (encoded < 27) {
+            switch (encoded) {
```
pedberg-icu, line 177: This way of decoding is fragile; it already does not support all numeric regions codes used. I think LocaleDistanceMapper should write an array of supported numeric regions that this function can use for decoding.

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
```diff
@@ -89,4 +90,109 @@ public boolean equals(Object obj) {
...
+    }
+    // This method is added only to support encodeToIntForResource()
+    // It only support [A-Z]{2}|001|143|419 and does not work for other cases.
+    private int encodeRegionToInt() {
+        assert region.length() >= 2;
+        assert region.length() <= 3;
+        // Do not have enough bits to store the all 1000 possible combination of \d{3}
+        // Only support what is needed in resource- "001", "143" and "419".
+        if (region.length() == 3) {
+            assert region.equals("001") || region.equals("143") || region.equals("419");
```
pedberg-icu, line 124: See my comments about this approach in LocaleDistanceMapper.java. It is too fragile and inflexible. I think LocaleDistanceMapper.java should write an array of supported numeric region codes that this method can use for encoding.

pedberg-icu (changes requested):
Needs a better way of handling numeric region codes, see comments (does not even support all currently used codes)

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/localedistance/LocaleDistanceMapper.java
FrankYFTang: well do. Changing the code now. Stay tune.

icu4c/source/common/loclikelysubtags.cpp
FrankYFTang: well do. Changing the code now. Stay tune.

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
FrankYFTang: well do. Changing the code now. Stay tune.

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
FrankYFTang: well do. Changing the code now. Stay tune.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/localedistance/LocaleDistanceMapper.java
FrankYFTang: DONE PTAL

icu4c/source/common/loclikelysubtags.cpp
FrankYFTang: DONE PTAL

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
FrankYFTang: DONE PTAL

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
FrankYFTang: DONE PTAL

FrankYFTang:
Peter - I modified the PR to output the M49 codes into the resources from the CLDR to ICU tool and use that during the decoding now.

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
pedberg-icu: Does not seem like this code got updated to use the m49 array?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/localedistance/LocaleDistanceMapper.java
```diff
@@ -102,19 +105,22 @@ public final class LocaleDistanceMapper {
...
     private static final RbPath MATCH_REGION_TO_PARTITIONS = RbPath.of("match", "regionToPartitions:bin");
     private static final RbPath MATCH_PARTITIONS = RbPath.of("match", "partitions");
-    private static final RbPath MATCH_PARADIGMS = RbPath.of("match", "paradigms");
+    private static final RbPath MATCH_PARADIGMNUM = RbPath.of("match", "paradigmnum:intvector");
     private static final RbPath MATCH_DISTANCES = RbPath.of("match", "distances:intvector");
 
     // To split locale specifications (e.g. "ja_Latn" or "en_*_$!enUS").
     private static final Splitter UNDERSCORE = Splitter.on('_');
 
+    private static final List<String> M49 = Arrays.asList("001", "143", "419");
```
pedberg-icu, line 122: So we do not have 150 in the data? Because we do have a locale for en_150.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/localedistance/LocaleDistanceMapper.java
```diff
@@ -434,21 +441,88 @@ private static RbValue ofMapEntries(Map<String, String> map) {
...
+        // Do not have enough bits to store the all 1000 possible combination of \d{3}
+        // Only support what is in M49.
+        if (region.length() == 3) {
+            int index = m49.indexOf(region);
+            assert index >= 0;
+            if (index < 0) {
+                throw new IllegalStateException(
+                    "Please add '" + region + "' to M49 in LocaleDistanceMapper.java");
+            }
+            return index;
```
pedberg-icu, line 492: This assumes that m49 has fewer that 27 entries; it certainly does now, but we might add a comment in the array def at the top, and/or a check here.

pedberg-icu:
> Peter - I modified the PR to output the M49 codes into the resources from the CLDR to ICU tool and use that during the decoding now.

Oh, I like this much better, thanks! But it looks like LSR.java is still not using the new M49 array...

icu4j/main/classes/core/src/com/ibm/icu/impl/locale/LSR.java
FrankYFTang: opps. you are right. need to change it.

FrankYFTang:
> Needs a better way of handling numeric region codes, see comments (does not even support all currently used codes)

It is not intend to support ALL used codes in everywhere, but only for those encoded in the value of the likelySubtag.xml

FrankYFTang:
> > Peter - I modified the PR to output the M49 codes into the resources from the CLDR to ICU tool and use that during the decoding now.
> 
> Oh, I like this much better, thanks! But it looks like LSR.java is still not using the new M49 array...

yes, sorry, I missed that. Will fix

pedberg-icu (approved):
Looks good now, thanks for this!

PR #2459, opened 2023-05-12
ICU-22390 Speed-up ICU4J Transliterator.<clinit>

gvictor:
It reduces the method runtime by approx. 60%.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22390

### DIFF
--- a/icu4j/main/classes/translit/src/com/ibm/icu/text/AnyTransliterator.java
+++ b/icu4j/main/classes/translit/src/com/ibm/icu/text/AnyTransliterator.java
@@ -53,6 +53,12 @@ class AnyTransliterator extends Transliterator {
     static final String LATIN_PIVOT = "-Latin;Latin-";
 
     /**
+     * Special code for handling width characters
+     */
+    private static final Transliterator WIDTH_FIX =
+            Transliterator.getInstance("[[:dt=Nar:][:dt=Wide:]] nfkd");
+
+    /**
      * Cache mapping UScriptCode values to Transliterator*.
      */
     private ConcurrentHashMap<Integer, Transliterator> cache;
@@ -66,11 +72,6 @@ class AnyTransliterator extends Transliterator {
      * The target script code.  Never USCRIPT_INVALID_CODE.
      */
     private int targetScript;
-
-    /**
-     * Special code for handling width characters
-     */
-    private Transliterator widthFix = Transliterator.getInstance("[[:dt=Nar:][:dt=Wide:]] nfkd");
 
     /**
      * Implements {@link Transliterator#handleTransliterate}.
@@ -176,7 +177,7 @@ class AnyTransliterator extends Transliterator {
             if (isWide(targetScript)) {
                 return null;
             } else {
-                return widthFix;
+                return WIDTH_FIX;
             }
         }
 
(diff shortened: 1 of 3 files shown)


gvictor:
Thanks, Mark. Do you know how I can accept the ticket?

macchiati:
They are reviewed each week by the ICU TC. (cc @markusicu )

markusicu:
We ran out of time today.
Before I take a look myself, would similar improvements help in C++ as well?

gvictor:
The major improvement is re-using the width_fix instance because width_fix instance is expensive. but I can't find it in ICU4C. The commit introducing widthFix seemed to be specific to ICU4J. https://github.com/unicode-org/icu/commit/d64ecece540a827cacac12c81af0057ac8bb546d
However, I am not familiar with the code base. 

The other improvement of replacing List with Set in TransliteratorRegistry seems to be appropriate in ICU4C. I can take a look as a follow-up CL in this bug if you are okay with this ICU4J change.

The 3rd one seems to be Java-specific, as String is first-class type in Java, but not in ICU4C.

icu4j/main/classes/translit/src/com/ibm/icu/text/AnyTransliterator.java
```diff
@@ -52,6 +52,12 @@ class AnyTransliterator extends Transliterator {
     static final String NULL_ID = "Null";
     static final String LATIN_PIVOT = "-Latin;Latin-";
 
+    /**
+     * Special code for handling width characters
+     */
+    private static final Transliterator WIDTH_FIX =
```
markusicu, line 58: optional: Is this even commonly used? If not, then it might be worth moving this constant into a nested class, in order to defer initialization until it is actually used.

If it is commonly used, then that's not worth it.

gvictor:
I have just accepted the ticket.

> Could I possibly interest you in taking over [ICU-6850](https://unicode-org.atlassian.net/browse/ICU-6850)? Looks like it could be relatively easy.

> We can't use STL collections in ICU4C, but we do have a hashmap class.

I will have a look. Do you mind reminding me why ICU4C can't use STL collections?

markusicu:
> Do you mind reminding me why ICU4C can't use STL collections?

Because
1. they throw exceptions; ICU code itself neither throws nor catches exceptions, and is not exception-safe (can leak memory if an exception goes through, leave objects in a state that can lead to crashes, etc.)
2. we allow users to set custom memory allocation (malloc/free) functions; we would have to use a custom, STL-compatible Allocator, but that wouldn't work if malloc returns NULL

For more on this see the User Guide Coding Guidelines.

macchiati (approved):

icu4j/main/classes/translit/src/com/ibm/icu/text/AnyTransliterator.java
gvictor: Given that it's always initialized in <clinit> even before this PR, Is it okay to submit this PR first?  I will take a look at this along with other ICU4C issues mentioned in this PR.

gvictor: If it's okay, could you help merge this PR? As I am not authorized to merge it

gvictor: https://github.com/unicode-org/icu/pull/2491 for moving it into the nested class.

gvictor:
> Could I possibly interest you in taking over [ICU-6850](https://unicode-org.atlassian.net/browse/ICU-6850)? Looks like it could be relatively easy.

I ported other changes to ICU4C, but I don't pick up ICU-6850 as I encountered some problems.

PR #2468, opened 2023-05-23
ICU-22379 Update ICU PersonNameFormatter to match the spec change req…

richgillam:
…uested by CLDR-16623

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22379

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
@@ -5,11 +5,7 @@ package com.ibm.icu.impl.personname;
 import static com.ibm.icu.util.UResourceBundle.ARRAY;
 import static com.ibm.icu.util.UResourceBundle.STRING;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Set;
+import java.util.*;
 
 import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.impl.ICUResourceBundle;
@@ -132,15 +128,22 @@ public class PersonNameFormatterImpl {
     public String formatToString(PersonName name) {
         // TODO: Should probably return a FormattedPersonName object
 
-        if (!nameScriptMatchesLocale(name, this.locale)) {
-            Locale nameLocale = getNameLocale(name);
-            PersonNameFormatterImpl nameLocaleFormatter = new PersonNameFormatterImpl(nameLocale, this.length,
+        Locale nameLocale = getNameLocale(name);
+        String nameScript = getNameScript(name);
+
+        if (!nameScriptMatchesLocale(nameScript, this.locale)) {
+            Locale newFormattingLocale;
+            if (formattingLocaleExists(nameLocale)) {
+                newFormattingLocale = nameLocale;
+            } else {
+                newFormattingLocale = newLocaleWithScript(null, nameScript, nameLocale.getCountry());
+            }
+            PersonNameFormatterImpl nameLocaleFormatter = new PersonNameFormatterImpl(newFormattingLocale, this.length,
                     this.usage, this.formality, this.displayOrder, this.capitalizeSurname);
             return nameLocaleFormatter.formatToString(name);
         }
 
         String result = null;
-        Locale nameLocale = getNameLocale(name);
 
         // choose the GN-first or SN-first pattern based on the name itself and use that to format it
         if (snFirstPatterns == null || nameIsGnFirst(name)) {
@@ -269,6 +272,67 @@ public class PersonNameFormatterImpl {
     }
 
     /**
+     * Internal function to figure out the name's script by examining its characters.
+     * @param name The name for which we need the script
+     * @return The four-letter script code for the name.
+     */
+    private String getNameScript(PersonName name) {
+        // Rather than exhaustively checking all the fields in the name, we just check the given-name
+        // and surname fields, giving preference to the script of the surname if they're different
+        // (we concatenate them into one string for simplicity).  The "name script" is the script
+        // of the first character we find whose script isn't "common".  If that script is one
+        // of the scripts used by the specified locale, we have a match.
+        String givenName = name.getFieldValue(PersonName.NameField.SURNAME, Collections.emptySet());
+        String surname = name.getFieldValue(PersonName.NameField.GIVEN, Collections.emptySet());
+        String nameText = ((surname != null) ? surname : "") + ((givenName != null) ? givenName : "");
+        int stringScript = UScript.UNKNOWN;
+        for (int i = 0; stringScript == UScript.UNKNOWN && i < nameText.length(); i++) {
+            int c = nameText.codePointAt(i);
+            int charScript = UScript.getScript(c);
+            if (charScript != UScript.COMMON && charScript != UScript.INHERITED && charScript != UScript.UNKNOWN) {
+                stringScript = charScript;
+            }
+        }
+        return UScript.getShortName(stringScript);
+    }
+
+    private Locale newLocaleWithScript(Locale oldLocale, String scriptCode, String regionCode) {
+        Locale workingLocale;
+        String localeScript;
+
+        // if we got the "unknown" script code, don't do anything with it-- just return the original locale
+        if (scriptCode.equals("Zzzz")) {
+            return oldLocale;
+        }
+
+        Locale.Builder builder = new Locale.Builder();
+        if (oldLocale != null) {
+            workingLocale = oldLocale;
+            builder.setLocale(oldLocale);
+            localeScript = ULocale.addLikelySubtags(ULocale.forLocale(oldLocale)).getScript();
+        } else {
+            ULocale tmpLocale = ULocale.addLikelySubtags(new ULocale("und_" + scriptCode));
+            builder.setLanguage(tmpLocale.getLanguage());
+            workingLocale = ULocale.addLikelySubtags(new ULocale(tmpLocale.getLanguage())).toLocale();
+            localeScript = workingLocale.getScript();
+
+            if (regionCode != null) {
+                builder.setRegion(regionCode);
+            }
+        }
+
+        // if the detected character script matches one of the default scripts for the name's locale,
+        // use the name locale's default script code in the locale ID we return (this converts a detected
+        // script of "Hani" to "Hans" for "zh", "Hant" for "zh_Hant", and "Jpan" for "ja")
+        if (!scriptCode.equals(localeScript) && nameScriptMatchesLocale(scriptCode, workingLocale)) {
+            scriptCode = localeScript;
+        }
+
+        builder.setScript(scriptCode);
+        return builder.build();
+    }
+
+    /**
      * Internal function to figure out the name's locale when the name doesn't specify it.
      * (Note that this code assumes that if the locale is specified, it includes a language
      * code.)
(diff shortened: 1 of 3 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -268,6 +279,30 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+        // of the first character we find whose script isn't "common".  If that script is one
+        // of the scripts used by the specified locale, we have a match.
+        String nameText = name.getFieldValue(PersonName.NameField.GIVEN, Collections.emptySet())
+                + name.getFieldValue(PersonName.NameField.SURNAME, Collections.emptySet());
+        int stringScript = UScript.UNKNOWN;
+        for (int i = 0; stringScript == UScript.UNKNOWN && i < nameText.length(); i++) {
+            char c = nameText.charAt(i);
+            int charScript = UScript.getScript(c);
+            if (charScript != UScript.COMMON && charScript != UScript.INHERITED) {
+                stringScript = charScript;
```
macchiati, line 300: This stops at the first "strong" character. Probably ok.
Might be a bit better to keep going just in case we hit a character whose script is compatible with the name locale or with the formatter locale. But we can discuss that later.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -276,65 +311,72 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+        if (specifiedNameLocale != null) {
+            builder.setLocale(specifiedNameLocale);
+            nameLanguage = specifiedNameLocale.getLanguage();
+        } else {
+            ULocale tmpLocale = ULocale.addLikelySubtags(new ULocale("und_" + nameScript));
+            nameLanguage = tmpLocale.getLanguage();
+            builder.setLanguage(nameLanguage);
+        }
+
+        // HACK to work around the fact that "Hani" isn't used in the IDs of any resource bundles
```
macchiati, line 328: There are 4 "special" scripts that we have to worry about: Jpan, Hani, Hans, Kore. I think we can lift the handling from the CLDR code.

macchiati (changes requested):

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -268,6 +279,30 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+        // Rather than exhaustively checking all the fields in the name, we just check the given-name
+        // and surname fields, giving preference to the script of the surname if they're different
+        // (we concatenate them into one string for simplicity).  The "name script" is the script
+        // of the first character we find whose script isn't "common".  If that script is one
+        // of the scripts used by the specified locale, we have a match.
+        String nameText = name.getFieldValue(PersonName.NameField.GIVEN, Collections.emptySet())
+                + name.getFieldValue(PersonName.NameField.SURNAME, Collections.emptySet());
+        int stringScript = UScript.UNKNOWN;
+        for (int i = 0; stringScript == UScript.UNKNOWN && i < nameText.length(); i++) {
+            char c = nameText.charAt(i);
```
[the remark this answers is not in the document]
richgillam: Yeah, you're right.  Fixed.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: At today's Person Names meeting, we agreed to punt on this request for now, just for performance and simplicity, on the ground that mixed script name input should be really rare in practice.  But we all acknowledged that this might come back in a future release if we turn out to be wrong about this.

We also discussed putting the surname before the given name and all agreed that made sense.  Amusingly, the doc comment at the top of the function already says we're doing that ("...we just check the given-name and surname fields, _giving preference to the script of the surname if they're different_..."), so my code didn't match my documentation.  I fixed the code.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: I just looked and didn't find anything.  Can you clue me in?  In the meantime, I've updated this to be table-driven and to take "Kore" into account.

richgillam: I was looking in the wrong place.  I'm thinking the `SpecialScripts` table in `GenerateMaximalLocales.java` might be what you're talking about?  But that looks like it does something a little different from what I'm trying to do here.  Can you expound more on what you're after here?  Is what I currently have good enough?  I'm running out of time here.

macchiati: It's not directly in the code, but here is code in the CheckPersonNames.java which is:

    public UnicodeSet getUnicodeSetForScript(String script) {
        switch (script) {
            case "Jpan":
                return JPAN;
            case "Kore":
                return KORE;
            case "Hant":
            case "Hans":
                return HANI;
            default:
                return new UnicodeSet("[\\p{sc=" + script + "}]");
        }
    }
    static final UnicodeSet HANI = new UnicodeSet("[\\p{sc=Hani}]").freeze();
    static final UnicodeSet KORE = new UnicodeSet("[\\p{sc=Hang}]").addAll(HANI).freeze();
    static final UnicodeSet JPAN =
            new UnicodeSet("[\\p{sc=Kana}\\p{sc=Hira}]").addAll(HANI).freeze();

This gets to the 'is compatible with' step, since you can get Jpan, Kore, Hant, and Hans out of maximize.

macchiati: Sorry, the "match" definition. "A few script values represent a set of scripts, such as Jpan = {Hani, Kana, Hira}. Two script codes are said to match when they are either identical, or one represents a set which contains the other, or they both represent sets which intersect. For example, Hani and Jpan match, because {Hani, Kana, Hira} contains Hani." We should probably update the spec to specifically address those 4 script values.

richgillam: Peter referred me to this same snippet, but I don't understand how it applies to this situation at all.  In my case, I've detected the script `Hani`from examining the characters and need to convert it to the default script for the relevant locale.  This code seems like it's doing the opposite.  And all I'm trying to do is fabricate something that'll let the "has resource data" check pass.

richgillam: The script-matching stuff already handles this situation, or so I believe. It's just the resource-lookup stuff I'm trying to get to do the right thing.

macchiati: Ok, great.

richgillam:
@macchiati , have I addressed your concerns, and is it okay to squash.  You have the PR blocked.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -276,65 +307,73 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+            builder.setLanguage(nameLanguage);
         }
-        return nameLocale;
+
+        // HACK to work around the fact that "Hani" isn't used in the IDs of any resource bundles-- map it to the
+        // default scripts for the relevant language codes
+        Map<String, String> SCRIPT_OVERRIDES = new HashMap<String, String>() {{
+            put("zh-Hani", "Hans");
+            put("ja-Hani", "Jpan");
+            put("ko-Hani", "Kore");
```
macchiati, line 329: Should also include ko-Hang, ja-Kana, ja-Hira. They will probably not come up, but just in case. Also, I haven't checked, but are you dealing with old-style ICU locales (like zh_Hani) or BCP47 style (like zh-Hani).

I think it might be better to match in the other direction for Hani, because the resource bundles will contain both zh_Hans and zh_Hant

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameFormatterTest.java
```diff
@@ -507,4 +507,35 @@ public void TestCapitalization() {
...
+
+    @Test
+    public void TestLocaleDerivation() {
+        // Test for https://unicode-org.atlassian.net/browse/ICU-22379, which implements the algorithm
+        // described in https://unicode-org.atlassian.net/browse/CLDR-16623.
+        executeTestCases(new NameAndTestCases[]{
+            // If we have a name that's tagged as Japanese, but contains Latin characters, and we're using
+            // a Japanese formatter, we actually use the English formatter to format it, but because the name is
+            // tagged as Japanese, we still use Japanese field order
+            new NameAndTestCases("given=Richard,surname=Gillam,locale=ja_AQ", new String[][]{
```
macchiati, line 519: Add some cases with ja_Hani and ja_Jpan, ko_Hang, zh_Hant, and zh_Hans, just to make sure those work.

macchiati (approved):

macchiati:
Just reviewed it again, and I think we're almost there.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: >Also, I haven't checked, but are you dealing with old-style ICU locales (like zh_Hani) or BCP47 style (like zh-Hani).

In this table, it's neither, really.  I need to look up based on both the language and the script, so I combine them into one key by concatenating them and putting a hyphen in between.  It looks like a BCP47 language tag, but it's really just an arbitrary string to facilitate lookup in a hash map.

>I think it might be better to match in the other direction for Hani, because the resource bundles will contain both zh_Hans and zh_Hant

I don't know what you mean by "match in the other direction."  If I get "Hani" as my script, it means I can't distinguish between "Hans" and "Hant", so I arbitrarily picked "Hant".  My thinking was that for the purposes of finding out if locale data for the locale ID exists, it doesn't matter, because we know both zh-Hans and zh-Hant exist.

icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameFormatterTest.java
richgillam: Added a few new ones for Japanese to get at this issue.  I didn't add any for Korean or Chinese, on the assumption that if Japanese works, they will too, but if you really want them, I'll try to figure something out.

macchiati:
Ah, - was an arbitrary separator, that explains it.

What I was concerned with is the two cases where the name has Hani
characters

   1. the name locale is zh_Hans or zh_Hant
   2. the formatter locale is zh_Hant or zh_Hans

We just want to make sure all 4 combinations of #1 and #2 don't cause the
formatter locale to switch.

richgillam:
> What I was concerned with is the two cases where the name has Hani characters 1. the name locale is zh_Hans or zh_Hant 2. the formatter locale is zh_Hant or zh_Hans We just want to make sure all 4 combinations of #1 and #2 don't cause the formatter locale to switch.

(sigh)

I figured out a way to do that, and it fails and that pretty much invalidates this whole algorithm.  I'm trying to come up with a new one...

macchiati:
This definitiely points out that the spec is still underspecified.

Here is an idea. We I think we can have a Multimap from the characters's script to each other script code that is compatible with it:

Hani => Hans, Hant, Hani, Kore, Japn
Hira => Japn
Kata => Japn
Hang => Kore

Let me try that out.

richgillam:
> Here is an idea. We I think we can have a Multimap from the characters's script to each other script code that is compatible with it:
> 
> Hani => Hans, Hant, Hani, Kore, Japn Hira => Japn Kata => Japn Hang => Kore

I've just been using `UScript.getCode()` to do that.

macchiati:
Here's what I mean, something like:

    *private* *static* *final* Multimap<String,String>
*CHARACTER_SCRIPT_TO_COMPATIBLE* = ImmutableMultimap.<String, String>
*builder*()

        .putAll("Hani", "Hant", "Hans", "Kore", "Japn")

        .putAll("Hira", "Japn")

        .putAll("Kana", "Japn")

        .putAll("Hang", "Kore")

    .build();

    /**

     * checks whether two short script names are compatible

     */

    *public* *boolean* isCompatible(String nameScript, String otherScript) {

        *if* (nameScript.equals(otherScript)) {

            *return* *true*;

        }

        Collection<String> otherCompatibles =
*CHARACTER_SCRIPT_TO_COMPATIBLE*.get(nameScript);

        *return* otherCompatibles.isEmpty() ? *false* : otherCompatibles
.contains(otherScript);

    }



    *public* *void* testCompatible() {

        String[][] tests = {

            {"一", "Hani", "Latn", "false"},

            {"一", "Hani", "Hani", "true"},

            {"一", "Hani", "Hant", "true"},

            {"一", "Hani", "Hans", "true"},

            {"一", "Hani", "Kore", "true"},

            {"一", "Hani", "Japn", "true"},


            {"あ", "Hira", "Latn", "false"},

            {"あ", "Hira", "Hira", "true"},

            {"あ", "Hira", "Japn", "true"},


            {"ア", "Kana", "Latn", "false"},

            {"ア", "Kana", "Kana", "true"},

            {"ア", "Kana", "Japn", "true"},


            {"가", "Hang", "Latn", "false"},

            {"가", "Hang", "Hang", "true"},

            {"가", "Hang", "Kore", "true"},


        };

        *for* (String[] test : tests) {

            *int* sourceCharacter = test[0].codePointAt(0);

            String expectedScript = test[1];

            String scriptToTest = test[2];

            *boolean* expectedResult = Boolean.*valueOf*(test[3]);

            *int* scriptCode = UScript.*getScript*(sourceCharacter);

            String shortName = UScript.*getShortName*(scriptCode);

            assertEquals("script of character", expectedScript, shortName);

            *boolean* actualResult= isCompatible(shortName, scriptToTest);

            assertEquals("scripts compatible?", expectedResult, actualResult
);

        }

    }

richgillam:
> Here's what I mean, something like:

I took a different tack that doesn't involve creating a whole new equivalence table.  Instead, it relies on `UScript.getCode()`, although I did have an add a hack to handle `Hans`/`Hant`/`Hani` properly.

I had to rewrite a bunch of other stuff to accommodate Chinese properly, but this does seem to produce the correct results.

I really, really hope this passes muster...

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -276,65 +334,47 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
-        int stringScript = UScript.COMMON;
-        for (int i = 0; stringScript == UScript.COMMON && i < nameText.length(); i++) {
-            char c = nameText.charAt(i);
-            stringScript = UScript.getScript(c);
-        }
+        int nameScript = UScript.getCodeFromName(nameScriptID);
 
         for (int localeScript : localeScripts) {
-            if (localeScript == stringScript) {
+            if (localeScript == nameScript || (localeScript == UScript.SIMPLIFIED_HAN && nameScript == UScript.HAN) || (localeScript == UScript.TRADITIONAL_HAN && nameScript == UScript.HAN)) {
```
macchiati, line 348: I don't see it handling Hani or Kana or Kata vs Jpan, etc. Is that handled somewhere else? It looks like the test cases work, but I'm a bit puzzled.

macchiati (commented):
Some questions

pedberg-icu (dismissed):

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -268,6 +271,61 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+        // and surname fields, giving preference to the script of the surname if they're different
+        // (we concatenate them into one string for simplicity).  The "name script" is the script
+        // of the first character we find whose script isn't "common".  If that script is one
+        // of the scripts used by the specified locale, we have a match.
+        String nameText = name.getFieldValue(PersonName.NameField.SURNAME, Collections.emptySet())
+                + name.getFieldValue(PersonName.NameField.GIVEN, Collections.emptySet());
+        int stringScript = UScript.UNKNOWN;
+        for (int i = 0; stringScript == UScript.UNKNOWN && i < nameText.length(); i++) {
+            int c = nameText.codePointAt(i);
+            int charScript = UScript.getScript(c);
```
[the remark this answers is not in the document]
macchiati: Those will have script=Unknown (see https://util.unicode.org/UnicodeJsps/character.jsp?a=D800 && https://util.unicode.org/UnicodeJsps/character.jsp?a=DC00) so they won't interfere with the processing.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: `Hana`/`Kata`/`Hani` vs `Jpan` and `Hang`/`Hani` vs `Kore` are dealt with using `UScript.getCode(Locale)`.  If the locale's language is `ja`, this returns an array containing `UScript.KATAKANA`, `UScript.HIRAGANA`, and `UScript.HAN`.  If the locale's language is `ko`, it returns an array containing `UScript.HANGUL` and `UScript.HAN`.  Otherwise (including for Chinese), it falls back on likely subtags.  I had to special-case Chinese because `UScript.getCode(Locale)` doesn't do anything special for it.  (Actually, that's not strictly true-- for `zh-Hant`, it'll return `UScript.HAN` and `UScript.BOPOMOFO`, but that doesn't really help me.)

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -268,6 +271,61 @@ private PersonNamePattern getBestPattern(PersonNamePattern[] patterns, PersonNam
...
+     * @param name The name for which we need the script
+     * @return The four-letter script code for the name.
+     */
+    private String getNameScript(PersonName name) {
+        // Rather than exhaustively checking all the fields in the name, we just check the given-name
+        // and surname fields, giving preference to the script of the surname if they're different
+        // (we concatenate them into one string for simplicity).  The "name script" is the script
+        // of the first character we find whose script isn't "common".  If that script is one
+        // of the scripts used by the specified locale, we have a match.
+        String nameText = name.getFieldValue(PersonName.NameField.SURNAME, Collections.emptySet())
```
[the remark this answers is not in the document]
richgillam: Very good point.  Fixed.  (The fix uncovered a few _other_ things I wasn't doing right-- I've fixed them too, I think.)

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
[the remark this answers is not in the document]
richgillam: I double-checked it in the actual code, and Mark's right-- unpaired surrogates return `UScript.UNKNOWN`.  It turns out I _wasn't_ handling that correctly, though-- I've added a fix.

macchiati (approved):

pedberg-icu (approved):

PR #2430, opened 2023-04-16
ICU-22357 Update gb18030 mappings for the -2022 version

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22357
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

Update ICU gb18030 converter mappings from gb18030-2005 to gb18030-2022 (36 rearranged mappings). Two sets of updated mappings were near the beginning or end of a range that was algorithmically mapped; in those cases the range was shortened, and the mappings removed from the range (including some that are not being changed) were converted to explicit mappings.

PR #2469, opened 2023-05-24
Revert "ICU-22357 Update gb18030 mappings for the -2022 version"

sffc:
Reverts unicode-org/icu#2430

I want to see if this fixes CI. If it does, we should figure out why.

CC @pedberg-icu @markusicu @richgillam @gvictor


markusicu:
> I want to see if this fixes CI. If it does, we should figure out why.

All tests pass, including the DataFilter test. Which is bad news :-/

(All pass except for "jira-ticket" and "single-commit".)

pedberg-icu:
@sffc This can be closed now, right?

sffc:
Superseded by https://github.com/unicode-org/icu/pull/2471

sffc:
I do find it quite curious how two things coincided: a PR landed in the uconv code, which very rarely changes, and also we start getting flaky tests that point to the uconv code. But, we seem to agree that this is a pure coincidence and that the root cause is actually the CI parallelism change.

PR #2487, opened 2023-06-01
ICU-22350 Set minimal workflow token permissions

pnacht:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22350

As described in the issue, this PR sets read-only tokens for all workflows. Only the deployment job in `jekyll-gh-pages.yml` workflow receives a few additional permissions it needs.

This protects the project from supply-chain attacks.

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -2,6 +2,10 @@ name: Publish icu4j.jar/utilities.jar to GH Maven
 on:
   release:
     types: [created]
+
+permissions:
+  contents: read
+
 jobs:
   publish:
     runs-on: ubuntu-latest
--- a/.github/workflows/cache_retain.yml
+++ b/.github/workflows/cache_retain.yml
@@ -26,6 +26,9 @@ on:
     # this cron schedule is set to run every 6 days to ensure retention
     - cron: '0 12 */6 * *'
 
+permissions:
+  contents: read
+
 jobs:
   retain-maven-cache:
     name: Run all tests with Maven
--- a/.github/workflows/icu_ci.yml
+++ b/.github/workflows/icu_ci.yml
@@ -13,6 +13,9 @@ on:
   pull_request:
     branches: '**'
 
+permissions:
+  contents: read
+
 jobs:
 
   # ICU4C docs build using doxygen..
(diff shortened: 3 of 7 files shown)


echeran (approved):
LGTM, thanks for doing this!

markusicu (approved):
lgtm2 tnx!

I will propose to the team that we move the ticket to "accepted" status :-)

mihnita (approved):

mihnita (approved):
Thank you!
M.

FrankYFTang (approved):
LGTM

markusicu:
> It looks as if this change broke all performance tests on the CI. Is that so?

Possible. We didn't look at those last week. They do write to a repo.

pnacht:
Woah, sorry about that!

I'm taking a look at the code, is the issue the [icu_merge_ci.yml](https://github.com/unicode-org/icu/blob/2e3e49fed507108933c8b8ec07be2116b0722ecb/.github/workflows/icu_merge_ci.yml)? Specifically here (and similar):

https://github.com/unicode-org/icu/blob/2e3e49fed507108933c8b8ec07be2116b0722ecb/.github/workflows/icu_merge_ci.yml#L102-L115

I'm looking at that Action's documentation and I'm having a bit of trouble understanding which permissions it requires... it seems to save the results to `icu4c/source/test/perf/results/${{ matrix.perf }}/output.txt`, but the repo doesn't have a `.../results` folder...

Let me know and I'll send a PR to fix this ASAP. (Can I re-use the ICU-22350 bug ID?)

.github/workflows/maven.yaml
```diff
@@ -2,6 +2,10 @@ name: Publish icu4j.jar/utilities.jar to GH Maven
 on:
   release:
     types: [created]
+
+permissions:
+  contents: read
```
srl295, line 7: this workflow needs to `mvn deploy` to github packages

srl295: in fact that's the only purpose of this workflow

srl295: so i'd just revert these two lines.

pnacht: Sorry! Fixed in https://github.com/unicode-org/icu/pull/2529.

PR #2507, opened 2023-06-20
ICU-22335 Remove `!!reverse`, `!!safe_reverse` and `!!safe_forward`

gz83:
`!!reverse`, `!!safe_reverse` and `!!safe_forward` will be deprecated in future versions, we should delete all related code in the code tree.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22335


gz83:
@aheninger 

PTAL

aheninger:
While the code to recognize and quietly ignore the deprecated rule types, !!reverse etc., continues to exist, the test cases containing them should also continue to exist.

Fully removing support for them, causing them to become rule parse errors, would first need TC discussion. Removal would be a breaking change for anyone who is using an older set of custom rules. Which should work fine as things are.

PR #2516, opened 2023-06-26
ICU-22425 Eliminate double map lookup for common case of present argument

JakeWharton:
In the uncommon case where the map lookup returns null, only then perform a second map lookup to determine whether it was an absent value or explicit null.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22425


yumaoka (approved):
LGTM

PR #2522, opened 2023-07-05
ICU-22433 uppercase decomposed greek to decomposed greek and precomposed greek to precomposed greek.

eggrobin:
Right now Greek uppercasing produces unnormalized text (it always introduces a precomposed dialytika, and always uses the decomposed eta tonos).

This change produces precomposed eta tonos from precomposed eta tonos, as was originally intended, and combining dialytika from combining accents on the preceding vowel; the test checks that NFD is mapped to NFD and NFC to NFC for all test strings.

The change also adds some tests for better coverage of the iota subscript behaviour.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22433


icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UCharacterCaseTest.java
```diff
@@ -837,7 +838,12 @@ else if (lowerTest.charAt(i)
     }
 
     private void assertGreekUpper(String s, String expected) {
-        assertEquals("toUpper/Greek(" + s + ')', expected, UCharacter.toUpperCase(GREEK_LOCALE_, s));
+        Normalizer2 nfc = Normalizer2.getNFCInstance();
```
macchiati, line 841: Great stuff.

If we want to get a more complete workout, we might refactor as follows:

1. Add assertGreekUpperTitle(String s, String expectedUpper, String expectedTitle), which calls assertGreekUpper(s, expectedUpper, s)
4. Add the code to test titlecase also. 
6. When it fails, you can use the actual title values to expand the test cases, checking the actual for validity.
8. I think that titlecase is also expected to roundtrip, so you could add that also.

But this goes beyond the original ticket, so you might just clone it off into a separate ticket for the future. So approving in case you want to do that.

macchiati (approved):

icu4j/main/tests/core/src/com/ibm/icu/dev/test/lang/UCharacterCaseTest.java
eggrobin: If I recall correctly, there is no Greek tailoring for titlecase (that is, accents are retained in titlecase and only removed in ALL CAPS), hence the lack of dedicated tests for that.

macchiati: Yup, now that I think about it. Never mind then!

PR #2523, opened 2023-07-06
ICU-22434 Not calling setFirstDayOfWeek(MONDAY) if the locale has fw

FrankYFTang:
The Calendar constructor already take care of the fw override. We should not set the first day of week for iso8601 to Monday if we have a fw keyword/type in the locale.

Also fix a bug which caused by an incorrect calendar= keyword extraction code which cannot handle more than one keyword/type, by replacing the correct one from Locale class.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22434

### DIFF
--- a/icu4c/source/test/intltest/caltest.cpp
+++ b/icu4c/source/test/intltest/caltest.cpp
@@ -185,7 +185,9 @@ void CalendarTest::runIndexedTest( int32_t index, UBool exec, const char* &name,
     TESTCASE_AUTO(TestActualLimitsOrdinalMonth);
     TESTCASE_AUTO(TestChineseCalendarMonthInSpecialYear);
     TESTCASE_AUTO(TestClearMonth);
-  
+
+    TESTCASE_AUTO(TestFWWithISO8601);
+
     TESTCASE_AUTO_END;
 }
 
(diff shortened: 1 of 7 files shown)


pedberg-icu (dismissed):

icu4c/source/test/intltest/caltest.cpp
```diff
@@ -5494,6 +5496,31 @@ void CalendarTest::TestChineseCalendarMonthInSpecialYear() {
...
+    for (int i = UCAL_SUNDAY; i <= UCAL_SATURDAY; i++) {
+        UErrorCode status = U_ZERO_ERROR;
+        const char* locale = locales[i];
+        LocalPointer<Calendar> cal(
+            Calendar::createInstance(
+                Locale(locale), status), status);
+        if (failure(status, "Constructor failed")) continue;
+        std::string msg("Calendar::createInstance(\"");
+        msg = msg + locale + "\")->getFirstDayOfWeek()";
+        assertEquals(msg.c_str(), i, cal->getFirstDayOfWeek());
```
pedberg-icu, line 5521: Hmm, do we need to cast cal->getFirstDayOfWeek() to int?

pedberg-icu (approved):

FrankYFTang:
@macchiati  Mark - could you take a look at the unit test and approve if you think that is the right expected outcome for these cases. 

Thanks.

macchiati (approved):

PR #2531, opened 2023-07-21
ICU-22435 Add C API for Locale 

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22435

### DIFF
--- a/icu4c/source/common/ulocale.cpp
+++ b/icu4c/source/common/ulocale.cpp
@@ -1,0 +1,96 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+//
+#include "unicode/errorcode.h"
+#include "unicode/stringpiece.h"
+#include "unicode/utypes.h"
+#include "unicode/ustring.h"
+#include "unicode/ulocale.h"
+#include "unicode/locid.h"
+
+#include "charstr.h"
+#include "cmemory.h"
+
+U_NAMESPACE_USE
+#define EXTERNAL(i) (reinterpret_cast<ULocale*>(i))
+#define CONST_INTERNAL(e) (reinterpret_cast<const icu::Locale*>(e))
+#define INTERNAL(e) (reinterpret_cast<icu::Locale*>(e))
+
+ULocale*
+ulocale_openForLocaleID(const char* localeID, int32_t length, UErrorCode* err) {
+    CharString str(length < 0 ? StringPiece(localeID) : StringPiece(localeID, length), *err);
+    if (U_FAILURE(*err)) return nullptr;
+    return EXTERNAL(icu::Locale::createFromName(str.data()).clone());
+}
+
+ULocale*
+ulocale_openForLanguageTag(const char* tag, int32_t length, UErrorCode* err) {
+  Locale l = icu::Locale::forLanguageTag(length < 0 ? StringPiece(tag) : StringPiece(tag, length), *err);
+  if (U_FAILURE(*err)) return nullptr;
+  return EXTERNAL(l.clone());
+}
+
+void
+ulocale_close(ULocale* locale) {
+    delete INTERNAL(locale);
+}
+
+#define IMPL_ULOCALE_STRING_GETTER(N1, N2) \
+const char* ulocale_get ## N1(const ULocale* locale) { \
+    if (locale == nullptr) return nullptr; \
+    return CONST_INTERNAL(locale)->get ## N2(); \
+}
+
+#define IMPL_ULOCALE_STRING_IDENTICAL_GETTER(N) IMPL_ULOCALE_STRING_GETTER(N, N)
+
+#define IMPL_ULOCALE_GET_KEYWORD_VALUE(N) \
+int32_t ulocale_get ##N ( \
+    const ULocale* locale, const char* keyword, int32_t keywordLength, \
+    char* valueBuffer, int32_t bufferCapacity, UErrorCode *err) { \
+    if (U_FAILURE(*err)) return 0; \
+    if (locale == nullptr) { \
+        *err = U_ILLEGAL_ARGUMENT_ERROR; \
+        return 0; \
+    } \
+    CheckedArrayByteSink sink(valueBuffer, bufferCapacity); \
+    CONST_INTERNAL(locale)->get ## N( \
+        keywordLength < 0 ? StringPiece(keyword) : StringPiece(keyword, keywordLength), \
+        sink, *err); \
+    if (U_FAILURE(*err)) return 0; \
+    if (sink.Overflowed()) { \
+       *err = U_BUFFER_OVERFLOW_ERROR; \
+       return sink.NumberOfBytesAppended()+1; \
+    } \
+    int32_t len = sink.NumberOfBytesWritten(); \
+    if (len < bufferCapacity) valueBuffer[len] = '\0'; \
+    return len; \
+}
+
+#define IMPL_ULOCALE_GET_KEYWORDS(N) \
+UEnumeration* ulocale_get ## N(const ULocale* locale, UErrorCode *err) { \
+    if (U_FAILURE(*err)) return nullptr; \
+    if (locale == nullptr) { \
+        *err = U_ILLEGAL_ARGUMENT_ERROR; \
+        return nullptr; \
+    } \
+    return uenum_openFromStringEnumeration( \
+        CONST_INTERNAL(locale)->create ## N(*err), err); \
+}
+
+IMPL_ULOCALE_STRING_IDENTICAL_GETTER(Language)
+IMPL_ULOCALE_STRING_IDENTICAL_GETTER(Script)
+IMPL_ULOCALE_STRING_GETTER(Region, Country)
+IMPL_ULOCALE_STRING_IDENTICAL_GETTER(Variant)
+IMPL_ULOCALE_STRING_GETTER(LocaleID, Name)
+IMPL_ULOCALE_STRING_IDENTICAL_GETTER(BaseName)
+IMPL_ULOCALE_GET_KEYWORD_VALUE(KeywordValue)
+IMPL_ULOCALE_GET_KEYWORD_VALUE(UnicodeKeywordValue)
+IMPL_ULOCALE_GET_KEYWORDS(Keywords)
+IMPL_ULOCALE_GET_KEYWORDS(UnicodeKeywords)
+
+bool ulocale_isBogus(const ULocale* locale) {
+    if (locale == nullptr) return false;
+    return CONST_INTERNAL(locale)->isBogus();
+}
+
+/*eof*/
--- a/icu4c/source/common/unicode/ulocale.h
+++ b/icu4c/source/common/unicode/ulocale.h
@@ -1,0 +1,229 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+#ifndef ULOCALE_H
+#define ULOCALE_H
+
+#include "unicode/localpointer.h"
+#include "unicode/uenum.h"
+#include "unicode/utypes.h"
+
+/**
+ * \file
+ * \brief C API: Locale ID functionality similar to C++ class Locale
+ */
+
+#ifndef U_HIDE_DRAFT_API
+/**
+ * Opaque C service object type for the locale API
+ * @draft ICU 74
+ */
+struct ULocale;
+
+/**
+ * C typedef for struct ULocale.
+ * @draft ICU 74
+ */
+typedef struct ULocale ULocale;
+
+/**
+ * Constructs an ULocale from the locale ID.
+ * The created ULocale should be destroyed by calling
+ * ulocale_close();
+ * @param localeID the locale, a const char * pointer (need not be terminated when
+ *               the length is non-negative)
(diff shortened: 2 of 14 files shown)


FrankYFTang:
ICU TC approved the API proposal on 2023-08-03 as implemented in this PR. However, please do NOT review this PR yet since I am still figuring out how to port C++/Java test to test these C interface.

FrankYFTang:
All necessary tests are added. Please start review. Thanks

icu4c/source/common/ulocale.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+U_NAMESPACE_USE
+#define EXTERNAL(i) ((ULocale*)(i))
+#define CONST_INTERNAL(e) ((const icu::Locale*)(e))
+#define INTERNAL(e) ((icu::Locale*)(e))
+
+ULocale*
+ulocale_openForLocaleID(const char* localeID, int32_t length) {
+    if (length < 0 || localeID[length] == '\0') {
+        return EXTERNAL(icu::Locale::createFromName(localeID).clone());
```
richgillam, line 21: Why do you need the `clone()` here?  And doesn't it mean the result of `createFromName()` leaks?  Or are the locale objects cached or something?

icu4c/source/common/ulocale.cpp
```diff
@@ -0,0 +1,107 @@
...
+    } else {
+        if (length < ULOC_FULLNAME_CAPACITY-1) {
+            char buf[ULOC_FULLNAME_CAPACITY];
+            uprv_memcpy(buf, localeID, length);
+            buf[length] = '\0';
+            return EXTERNAL(icu::Locale::createFromName(buf).clone());
```
richgillam, line 27: I think you could use a `CharString` here to simplify this a little...

icu4c/source/common/ulocale.cpp
```diff
@@ -0,0 +1,107 @@
...
+        // return a bogus one
+        Locale bogus;
+        bogus.setToBogus();
+        return EXTERNAL(bogus.clone());
+    }
+}
+
+ULocale*
+ulocale_openForLanguageTag(const char* tag, int32_t length, UErrorCode* err) {
+  if (U_FAILURE(*err)) return nullptr;
```
richgillam, line 38: Won't `Locale::forLanguageTag()` check `U_FAILURE()` for you?  Or is the issue the call to `clone()` at the end?

icu4c/source/common/ulocale.cpp
```diff
@@ -0,0 +1,107 @@
...
+  return EXTERNAL(icu::Locale::forLanguageTag(
+      length < 0 ? StringPiece(tag) : StringPiece(tag, length), *err).clone());
+}
+
+void
+ulocale_close(ULocale* locale) {
+    if (locale == nullptr) return;
+    delete INTERNAL(locale);
```
richgillam, line 46: Isn't the `delete` operator null-safe?

icu4c/source/common/unicode/ulocale.h
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * C typedef for struct ULocale.
+ * @draft ICU 74
+ */
+typedef struct ULocale ULocale;
+
+/**
+ * Constructs an ULocale from the locale ID.
+ * The created ULocale should be destoried by calling
```
richgillam, line 31: s/destoried/destroyed

icu4c/source/common/unicode/ulocale.h
```diff
@@ -0,0 +1,228 @@
...
+ * @return the locale.
+ *
+ * @draft ICU 74
+ */
+U_CAPI ULocale* U_EXPORT2
+ulocale_openForLocaleID(const char* localeID, int32_t length);
+
+/**
+ * Constructs an ULocale from the provided IETF BCP 47 language tag.
+ * The created ULocale should be destoried by calling
```
richgillam, line 46: s/destoried/destroyed

icu4c/source/common/unicode/ulocale.h
```diff
@@ -0,0 +1,228 @@
...
+ *                      keyword need to be null terminated.
+ * @param valueBuffer The buffer to receive the value.
+ * @param valueBufferCapacity The capacity of receiving valueBuffer.
+ * @param err the error code
+ * @draft ICU 74
+ */
+U_CAPI int32_t U_EXPORT2
+ulocale_getKeywordValue(
+    const ULocale* locale, const char* keyword, int32_t keywordLength,
+    char* valueBuffer, int32_t valueBufferCapacity, UErrorCode *err);
```
richgillam, line 185: I know we approved the API change, so I probably shouldn't bring this up, and I'm kind of ashamed that I didn't catch this during API review.  But the whole purpose of adding this API was to get an object with getters that would just return `const char*` so that the caller doesn't have to mess with allocating and filling a buffer.  The `getKeywordValue() ` and `getUnicodeKeywordValue()` methods don't do that, and it makes me wonder how they're any better than the C APIs we have for that capability now.  Is this the best that the C++ API gives us?

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -0,0 +1,321 @@
...
+    // variant code
+    {   "",     "",     "",     "",     "NY",   "",     "",   "",
+        ""},
+    // full name
+    {   "en_US",    "fr_FR",    "ca_ES",    "el_GR",    "no_NO_NY", "it",   "xx_YY",   "zh_Hans_CN",
+        "x-klingon_Latn_ZX.utf32be@special"},
+    // basename
+    {   "en_US",    "fr_FR",    "ca_ES",    "el_GR",    "no_NO_NY", "it",   "xx_YY",   "zh_Hans_CN",
+        "x-klingon_Latn_ZX.utf32be@special"},
+};
```
richgillam, line 44: I think this table would read a lot easier if you organized it the other way, where the individual fields were the columns and the individual test cases were the rows.  It also would mean you could add test cases to the table without having to change the dimension in line 25.

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -0,0 +1,321 @@
...
+        log_err_status(status, WHERE "ulocale_openForLanguageTag() failed");
+        return;
+    }
+    char buffer[ULOC_FULLNAME_CAPACITY];
+    ulocale_getUnicodeKeywordValue(l, "nu", -1, buffer, ULOC_FULLNAME_CAPACITY, &status);
+
+    if (U_FAILURE(status)) {
+        log_err_status(
+            status, WHERE "ulocale_getUnicodeKeywordValue(\"nu\") failed");
+    } else {
```
richgillam, line 286: You can compress this part of the test down a little with `assertSuccess()`.

richgillam (approved):
Looks good.  I have a few nits and one significant API-design issue it might be too late to do anything about.

icu4c/source/common/ulocale.cpp
FrankYFTang: The signature of createFromName is

static Locale icu::Locale::createFromName(const char * name) 
It return an object, not an return a pointer nor a reference to an object so I have to use clone() to return a poitner.

icu4c/source/common/unicode/ulocale.h
FrankYFTang: In C++ API, getKeywordValue()  and getUnicodeKeywordValue() both copy out the string and does not retain a copy inside the C++ object.  This Is the best that the C++ API gives us w/o changing the  C++ API ?

icu4c/source/common/ulocale.cpp
richgillam: That's what I was afraid of.  In that case, no objections from me.

icu4c/source/common/unicode/ulocale.h
richgillam: I was afraid of that, too.  If that's what C++ is doing, I agree-- it's not worth it to try to improve on that in C, and I don't think we want to mess with changing the C++ API at this point either.  I withdraw my objection.

FrankYFTang:
@richgillam I think I addressed all your comments PTAL

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -22,48 +22,40 @@ enum {
...
-        assertEquals("ulocale_getLanguage()", rawData[LANGUAGE][i], ulocale_getLanguage(l));
-        assertEquals("ulocale_getScript()", rawData[SCRIPT][i], ulocale_getScript(l));
-        assertEquals("ulocale_getRegion()", rawData[REGION][i], ulocale_getRegion(l));
-        assertEquals("ulocale_getVariant()", rawData[VAR][i], ulocale_getVariant(l));
-        assertEquals("ulocale_getLocaleID()", rawData[NAME][i], ulocale_getLocaleID(l));
-        assertEquals("ulocale_getBaseName()", rawData[BASENAME][i], ulocale_getBaseName(l));
+    for (int32_t i = 0; i < UPRV_LENGTHOF(rawData); i++)  {
+        UErrorCode status = U_ZERO_ERROR;
+        ULocale* l = ulocale_openForLocaleID(rawData[i][NAME], -1, &status);
+        assertSuccess(WHERE "ulocale_openForLocaleID()", &status);
```
richgillam, line 42: `assertSuccess()` returns true if the error code passed `U_SUCCESS`, and you probably want to check that before doing all the other tests, since if `ulocale_openForLocaleID()` fails, `l` is bad.

That is...

```
if (assertSuccess(WHERE "ulocale_openForLocaleID()", &status) {
    assertEquals(WHERE "ulocale_getLanguage()", rawData[i][LANGUAGE], ulocale_getLanguage(l));
    assertEquals(WHERE "ulocale_getScript()", rawData[I][SCRIPT], ulocale_getScript(l));
    // etc...
    ulocale_close(l);
}
```

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -22,48 +22,40 @@ enum {
...
     UErrorCode status = U_ZERO_ERROR;
-    ULocale* fromID = ulocale_openForLocaleID(localeID, -1);
+    ULocale* fromID = ulocale_openForLocaleID(localeID, -1, &status);
+    assertSuccess(WHERE "ulocale_openForLocaleID()", &status);
     ULocale* fromTag = ulocale_openForLanguageTag(tag, -1, &status);
-    if (U_FAILURE(status)) {
-        log_err_status(status, WHERE "ulocale_openForLanguageTag(\"%s\") failed",
-                       tag);
-    }
+    assertSuccess(WHERE "ulocale_openForLanguageTag()", &status);
```
richgillam, line 58: Same as above:

```
ULocale* fromID = ulocale_openForLocaleID(localeID, -1, &status);
if (assertSuccess(WHERE "ulocale_openForLocaleID()", &status)) {
    ULocale* fromTag = ulocale_openForLanguageTag(tag, -1, &status);
    if (assertSuccess(WHERE "ulocale_openForLanguageTag()", &status)) {
        assertEquals(tag, ulocale_getLocaleID(fromID), ulocale_getLocaleID(fromTag));
        ulocale_close(fromID);
    }
    ulocale_close(fromTag);
}
```

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -94,12 +86,12 @@ static void TestForLanguageTag() {
...
-    ULocale* l = ulocale_openForLocaleID("de@calendar=buddhist;collation=phonebook", -1);
     UErrorCode status = U_ZERO_ERROR;
+    ULocale* l = ulocale_openForLocaleID("de@calendar=buddhist;collation=phonebook", -1, &status);
+    assertSuccess(WHERE "ulocale_openForLocaleID()", &status);
     UEnumeration* en = ulocale_getKeywords(l, &status);
-    if (U_FAILURE(status)) {
-        log_err_status(status, WHERE "ulocale_getKeywords() failed");
-    } else {
+    assertSuccess(WHERE "ulocale_getKeywords()", &status);
+    if (U_SUCCESS(status)) {
```
richgillam, line 94: See comments above about checking the return value from `assertSuccess()`.

icu4c/source/test/cintltst/ulocaletst.c
```diff
@@ -123,28 +115,22 @@ static void TestGetKeywords() {
 }
 
 static void TestGetKeywordsEmpty() {
-    ULocale* l = ulocale_openForLocaleID("de", -1);
     UErrorCode status = U_ZERO_ERROR;
+    ULocale* l = ulocale_openForLocaleID("de", -1, &status);
+    assertSuccess(WHERE "ulocale_openForLocaleID()", &status);
```
richgillam, line 120: See comments above about checking the return value from `assertSuccess()`.

richgillam (commented):
`assertSuccess()` returns true if the error code passed `U_SUCCESS` and false if it passed `U_FAILURE`, so when you're using it, you don't need to do those checks yourself, and you want to be sure to check the return value in cases where the test depends on the previous call having succeeded.

I think there are a couple of other small mistakes in here.

But this looks a lot better!  Thanks a lot for the changes!

FrankYFTang:
@richgillam PTAL, I correct the assertSuccess based on your comments

richgillam (approved):
Looks great!  Thank you for the changes!

PR #2528, opened 2023-07-19
ICU-22304 Miscellanous PersonNameFormatter fixes; made ExhaustivePers…

richgillam:
…onNameFormatterTest into a real unit test.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22304
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2537, opened 2023-08-01
ICU-22362 Fix the name order derivation code in PersonNameFormatter t…

richgillam:
…o match the CLDR spec.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22362

### DIFF
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
@@ -98,11 +98,8 @@ public class PersonNameFormatterImpl {
     /**
      * THIS IS A DUMMY CONSTRUCTOR JUST FOR THE USE OF THE UNIT TESTS TO CHECK SOME OF THE INTERNAL IMPLEMENTATION!
      */
-    public PersonNameFormatterImpl(Locale locale, String[] patterns) {
+    public PersonNameFormatterImpl(Locale locale, String[] gnFirstPatterns, String[] snFirstPatterns, String[] gnFirstLocales, String[] snFirstLocales) {
         // first, set dummy values for the other fields
-        snFirstPatterns = null;
-        gnFirstLocales = null;
-        snFirstLocales = null;
         length = PersonNameFormatter.Length.MEDIUM;
         usage = PersonNameFormatter.Usage.REFERRING;
         formality = PersonNameFormatter.Formality.FORMAL;
@@ -114,10 +111,22 @@ public class PersonNameFormatterImpl {
         nativeSpaceReplacement = " ";
         formatterLocaleUsesSpaces = true;
 
-        // then, set values for the fields we actually care about
+        // then, set values for the fields we actually care about (all but gnFirstPatterns are optional)
         this.locale = locale;
-        gnFirstPatterns = PersonNamePattern.makePatterns(patterns, this);
-
+        this.gnFirstPatterns = PersonNamePattern.makePatterns(gnFirstPatterns, this);
+        this.snFirstPatterns = (snFirstPatterns != null) ? PersonNamePattern.makePatterns(snFirstPatterns, this) : null;
+        if (gnFirstLocales != null) {
+            this.gnFirstLocales = new HashSet<>();
+            Collections.addAll(this.gnFirstLocales, gnFirstLocales);
+        } else {
+            this.gnFirstLocales = null;
+        }
+        if (snFirstLocales != null) {
+            this.snFirstLocales = new HashSet<>();
+            Collections.addAll(this.snFirstLocales, snFirstLocales);
+        } else {
+            this.snFirstLocales = null;
+        }
     }
 
     @Override
@@ -193,6 +202,8 @@ public class PersonNameFormatterImpl {
 
     private final Set<String> LOCALES_THAT_DONT_USE_SPACES = new HashSet<>(Arrays.asList("ja", "zh", "yue", "km", "lo", "my"));
 
+    static final Set NON_DEFAULT_SCRIPTS = new HashSet<>(Arrays.asList("Hani", "Hira", "Kana"));
+
     /**
      * Returns the value of the resource, as a string array.
      * @param resource An ICUResourceBundle of type STRING or ARRAY.  If ARRAY, this function just returns it
@@ -223,23 +234,57 @@ public class PersonNameFormatterImpl {
             return false;
         }
 
-        String localeStr = getNameLocale(name).toString();
+        // Otherwise, search the gnFirstLocales and snFirstLocales for the locale's name.
+        // For our purposes, the "locale's name" is the locale the name itself gives us (if it
+        // has one), or the locale we guess for the name (if it doesn't).
+        Locale nameLocale = name.getNameLocale();
+        if (nameLocale == null) {
+            nameLocale = getNameLocale(name);
+        }
+
+        // this is a hack to deal with certain script codes that are valid, but not the default, for their locales--
+        // to make the parent-chain lookup work right, we need to replace any of those script codes (in the name's locale)
+        // with the appropriate default script for whatever language and region we have
+        ULocale nameULocale = ULocale.forLocale(nameLocale);
+        if (NON_DEFAULT_SCRIPTS.contains(nameULocale.getScript())) {
+            ULocale.Builder builder = new ULocale.Builder();
+            builder.setLocale(nameULocale);
+            builder.setScript(null);
+            nameULocale = ULocale.addLikelySubtags(builder.build());
+        }
+
+        // now search for the locale in the gnFirstLocales and snFirstLocales lists...
+        String localeStr = nameULocale.getName();
+        String origLocaleStr = localeStr;
+        String languageCode = nameULocale.getLanguage();
+
         do {
+            // first check if the locale is in one of those lists
             if (gnFirstLocales.contains(localeStr)) {
                 return true;
             } else if (snFirstLocales.contains(localeStr)) {
                 return false;
             }
 
-            int lastUnderbarPos = localeStr.lastIndexOf("_");
-            if (lastUnderbarPos >= 0) {
-                localeStr = localeStr.substring(0, lastUnderbarPos);
-            } else {
-                localeStr = "root";
+            // if not, try again with "und" in place of the language code (this lets us use "und_CN" to match
+            // all locales with a region code of "CN" and makes sure the last thing we try is always "und", which
+            // is required to be in gnFirstLocales or snFirstLocales)
+            String undStr = localeStr.replaceAll("^" + languageCode, "und");
+            if (gnFirstLocales.contains(undStr)) {
+                return true;
+            } else if (snFirstLocales.contains(undStr)) {
+                return false;
             }
-        } while (!localeStr.equals("root"));
 
-        // should never get here-- "root" should always be in one of the locales
+            // if we haven't found the locale ID yet, look up its parent locale ID and try again-- if getParentLocaleID()
+            // returns null (i.e., we have a locale ID, such as "zh_Hant", that inherits directly from "root"), try again
+            // with just the locale ID's language code (this fixes it so that "zh_Hant" matches "zh", even though "zh" isn't,
+            // strictly speaking, its parent locale)
+            String parentLocaleStr = ICUResourceBundle.getParentLocaleID(localeStr, origLocaleStr, ICUResourceBundle.OpenType.LOCALE_DEFAULT_ROOT);
+            localeStr = (parentLocaleStr != null) ? parentLocaleStr : languageCode;
+        } while (localeStr != null);
+
+        // should never get here ("und" should always be in gnFirstLocales or snFirstLocales), but if we do...
         return true;
     }
 
--- a/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/impl/ICUResourceBundle.java
@@ -1285,7 +1285,7 @@ public  class ICUResourceBundle extends UResourceBundle {
         return result;
     }
 
-    private static String getParentLocaleID(String name, String origName, OpenType openType) {
+    public static String getParentLocaleID(String name, String origName, OpenType openType) {
         // early out if the locale ID has a variant code or ends with _
         if (name.endsWith("_") || !ULocale.getVariant(name).isEmpty()) {
             int lastUnderbarPos = name.lastIndexOf('_');
--- a/icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java
+++ b/icu4j/main/classes/core/src/com/ibm/icu/text/PersonNameFormatter.java
@@ -319,8 +319,8 @@ public class PersonNameFormatter {
      * @deprecated This API is for unit testing only.
      */
     @Deprecated
-    public PersonNameFormatter(Locale locale, String[] patterns) {
-        this.impl = new PersonNameFormatterImpl(locale, patterns);
+    public PersonNameFormatter(Locale locale, String[] gnFirstPatterns, String[] snFirstPatterns, String[] gnFirstLocales, String[] snFirstLocales) {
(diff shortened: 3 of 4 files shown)


icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -215,32 +224,84 @@ private String[] asStringArray(ICUResourceBundle resource) {
...
+            // before looking up the parent locale, try again with "und" in place of the language code
+            String undStr = localeStr.replaceAll("^" + languageCode, "und");
+            if (localesToMatch.contains(undStr)) {
+                return undStr;
             }
-        } while (!localeStr.equals("root"));
 
-        // should never get here-- "root" should always be in one of the locales
-        return true;
+            String parentLocaleStr = ICUResourceBundle.getParentLocaleID(localeStr, origLocaleStr, ICUResourceBundle.OpenType.LOCALE_DEFAULT_ROOT);
```
richgillam, line 299: I don't like calling `ICUResourceBundle` like this, but I'm uncomfortable moving that logic into `ULocale`, where it probably belongs...

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -215,32 +224,84 @@ private String[] asStringArray(ICUResourceBundle resource) {
...
+    // name.getNameLocale() just returns the contents of the name's nameLocale field, but getNameLocale(name)
+    // guesses a locale for the name (using the stated locale as a hint, if it exists), filling in a script
+    // based on the characters used in the actual name text.
+    // localeMatchingLanguageCode is there for a few instances (notably, zh_Hant) where both localeMatchingDerivedLocale
+    // and localeMatchingStatedLocale fall back to "und" because we're doing a cross-script match and we never try
+    // just the language code by itself.
+    Locale derivedLocale = getNameLocale(name);
+    String localeMatchingDerivedLocale = closestMatchingLocaleID(derivedLocale, namedLocales);
+    String localeMatchingStatedLocale = closestMatchingLocaleID(name.getNameLocale(), namedLocales);
+    String localeMatchingLanguageCode = closestMatchingLocaleID(new Locale(derivedLocale.getLanguage()), namedLocales);
```
richgillam, line 249: This is infuriatingly complicated compared to the algorithm in the spec, but I couldn't come up with a simpler way to achieve the same result.  Suggestions are welcome.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
macchiati: That is really ugly... Didn't realize that the functionality was just in the RBs! @markusicu was this in case applications wanted their own parent relationship?

macchiati (commented):
Just did a first pass; now need to look at the complicated bit!

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: I think I may have actually added that function-- I know I added to it-- when I was making all the horizontal-inheritance fixes a while back.  And at that point, the only thing that needed it was ICUResourceBundle.  I'm still worried, though, that it's not quite ready to put into ULocale, at least if that would mean making it public.

One of the things that made my logic here complicated was that that method returns null if you change scripts moving up the parent chain, but I think we actually do want to do that here...

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -215,32 +224,84 @@ private String[] asStringArray(ICUResourceBundle resource) {
...
      * @return If true, use given-first order to format the name; if false, use surname-first order.
      */
-    private boolean nameIsGnFirst(PersonName name) {
-        // the name can declare its order-- check that first (it overrides any locale-based calculation)
-        if (name.getPreferredOrder() == PersonName.PreferredOrder.GIVEN_FIRST) {
-            return true;
-        } else if (name.getPreferredOrder() == PersonName.PreferredOrder.SURNAME_FIRST) {
-            return false;
+private boolean nameIsGnFirst(PersonName name) {
+    // the name can declare its order-- check that first (it overrides any locale-based calculation)
```
macchiati, line 228: Maybe put in some comments so we have the relationship to the spec?

```suggestion
// the name can declare its order-- check that first (it overrides any locale-based calculation)
// The numbered comments are from https://www.unicode.org/reports/tr35/tr35-personNames.html#derive-the-name-order
// 1. If the calling API requests sorting order, that is used.
// that is done outside of this method
```

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -215,32 +224,84 @@ private String[] asStringArray(ICUResourceBundle resource) {
...
+    if (name.getPreferredOrder() == PersonName.PreferredOrder.GIVEN_FIRST) {
+        return true;
+    } else if (name.getPreferredOrder() == PersonName.PreferredOrder.SURNAME_FIRST) {
+        return false;
+    }
+
```
macchiati, line 234:
```suggestion

// 3. Otherwise, use the nameOrderLocales elements to find the best match for the name locale, as follows.
```

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -215,32 +224,84 @@ private String[] asStringArray(ICUResourceBundle resource) {
...
+    Set<String> namedLocales = new HashSet<>();
```
macchiati, line 235: Having trouble relating this to the steps in https://www.unicode.org/reports/tr35/tr35-personNames.html#derive-the-name-order
For one thing, that loops over locales, whereas this doesn't (here).
So will return to this later

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
macchiati: What should happen is you skip from (say) en-Cyrl to root. So you only change scripts if the script is **not** the default for the language. And after root in any case you should fail (return null).

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -193,6 +202,8 @@ public boolean shouldCapitalizeSurname() {
 
     private final Set<String> LOCALES_THAT_DONT_USE_SPACES = new HashSet<>(Arrays.asList("ja", "zh", "yue", "km", "lo", "my"));
```
macchiati, line 203: I don't think these are needed anymore. Not a blocker for this PR, but lets talk if your code still uses them, because it shouldn't need to.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -223,23 +234,57 @@ private boolean nameIsGnFirst(PersonName name) {
...
 
-            int lastUnderbarPos = localeStr.lastIndexOf("_");
-            if (lastUnderbarPos >= 0) {
-                localeStr = localeStr.substring(0, lastUnderbarPos);
-            } else {
-                localeStr = "root";
+            // if not, try again with "und" in place of the language code (this lets us use "und_CN" to match
+            // all locales with a region code of "CN" and makes sure the last thing we try is always "und", which
+            // is required to be in gnFirstLocales or snFirstLocales)
+            String undStr = localeStr.replaceAll("^" + languageCode, "und");
```
macchiati, line 272: Huh, what is "^" + languageCode doing?

macchiati (approved):
I approved this, but we might want to talk about the two comments.

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: I have another ticket to clean up the foreignSpaceReplacement/nativeSpaceReplacement stuff, and I think that'll come out as part of that (I'll try to remember to double-check this).

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
richgillam: I'm calling (or am supposedly calling) an override that takes a regex, not a straight string, and I'm using the `^` to make sure I only match the language code at the beginning of the string (so I don't inadvertently mess up some other part of the locale ID that happens to match the language code).  If you can suggest a better way to do this that's still only one line long, I'm happy to change it.

Come to think of it, I guess I could do something like...

```
String undStr = "und" + localeStr.substring(languageCode.length());
```

...but I'm worried that wouldn't be as clear.  Are there better options?

pedberg-icu (approved):

macchiati:
I'd forgotten that replace all takes a regex.

Either way is fine

richgillam:
> I'd forgotten that replace all takes a regex.
> 
> Either way is fine

Okay, I think I'll leave it, then.  I just squashed my commits, which (I'm pretty sure) invalidates your approval.  Mind approving again?

richgillam:
> I just squashed my commits, which (I'm pretty sure) invalidates your approval. Mind approving again?

Actually, maybe they changed that-- it's still showing as approved.  Let's see what happens when the CI checks have finished...

macchiati:
Ok, lemme know if it needs a restamp

richgillam:
> Ok, lemme know if it needs a restamp

Looks like we're good!  Thank you!

FrankYFTang:
@richgillam this merge break the "Azure Pipelines / CI-Exhaustive-Main" again see.
https://github.com/unicode-org/icu/runs/15635250498
Compiling 512 source files to /home/vsts/work/1/s/icu4j/main/classes/core/out/bin
    [javac] /home/vsts/work/1/s/icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java:205: warning: [rawtypes] found raw type: Set
    [javac]     static final Set NON_DEFAULT_SCRIPTS = new HashSet<>(Arrays.asList("Hani", "Hira", "Kana"));
    [javac]                  ^
    [javac]   missing type arguments for generic class Set<E>
    [javac]   where E is a type-variable:
    [javac]     E extends Object declared in interface Set
    [javac] Note: Some input files use or override a deprecated API.
    [javac] Note: Recompile with -Xlint:deprecation for details.
    [javac] 1 warning

icu4j/main/classes/core/src/com/ibm/icu/impl/personname/PersonNameFormatterImpl.java
```diff
@@ -193,6 +202,8 @@ public boolean shouldCapitalizeSurname() {
 
     private final Set<String> LOCALES_THAT_DONT_USE_SPACES = new HashSet<>(Arrays.asList("ja", "zh", "yue", "km", "lo", "my"));
 
+    static final Set NON_DEFAULT_SCRIPTS = new HashSet<>(Arrays.asList("Hani", "Hira", "Kana"));
```
FrankYFTang, line 205: This line is problematic because you didn't import Set

FrankYFTang:
Please always run "/azp run CI-Exhaustive" for code in the PersonNameFormatter area so you can catch the error eariler.

richgillam:
Help me out here-- I don't even know what `/azp run CI-Exhaustive` is, what it does, why it's failing when the regular CI checks are passing, or why PersonNameFormatter should be special in this regard.

I thought I was okay, and now you're telling me my code won't build, and I don't understand why the normal CI checks don't catch that.

As far as actually fixing the problem is concerned, are you telling me the error is that I should have declared the variable as `Set<String>`?  That's easy enough to change, but I guess you're saying I'm going to have to create a whole new PR to do that now?

FrankYFTang:
>  I don't even know what /azp run CI-Exhaustive is
See the end of https://github.com/unicode-org/icu/blob/main/docs/userguide/dev/gitdev.md#requesting-an-exhaustive-test-run-on-a-pull-request-pr

The normal PR does not run these test since it take longer to run
It is run post merged
but could be run in the PR only if you put in "/azp run CI-Exhaustive" in the PR comment to triger it
Once that is put in, the following configure ation run 
https://github.com/unicode-org/icu/blob/main/.ci-builds/.azure-exhaustive-tests.yml

        echo "Building ICU4J" && cd icu4j && ant init && ant exhaustiveCheck

I think you can try it on your local build by running
```
 ant exhaustiveCheck
```

I am not sure how they hook up the PR comment to run .azure-exhaustive-tests.yml


>  why it's failing when the regular CI checks are passing, or 

Because the regular CI checks does not run exhaustiveCheck. For example, see
https://github.com/unicode-org/icu/blob/main/.ci-builds/.azure-pipelines.yml

> I thought I was okay, and now you're telling me my code won't build
I could be wrong. that could be just a warning. 

But what we know is that code is breaking post merge
```
ant check && ant localespiCheck
```
see also https://unicode-org.github.io/icu/processes/release/tasks/healthy-code.html#test-exhaustive-mode-c--j
see also https://icu.unicode.org/setup/java/ant
about
  exhaustiveCheck          Run the standard ICU4J test suite in exhaustive mode



to save time.
see https://github.com/unicode-org/icu/blob/main/icu4j/build.xml#L488

> why PersonNameFormatter should be special in this regard.
because in 
https://github.com/unicode-org/icu/pull/2528/files#diff-20b0715e99fbb5d581651582797ccab866e3008b498bbec5b3b189d12742c59bR37

your code ask it to only run that under exhaustiveCheck 
(that is your PR, right? I think you merge the following code. 
```
    @Before
    public void beforeMethod() {
        // Disable this test class except for exhaustive mode.
        // To enable exhaustive mode, pass the JVM argument "-DICU.exhaustive=10"
        org.junit.Assume.assumeTrue(getExhaustiveness() > 5);
    }
```

richgillam:
I understand all that, and I'll try to remember to run those tests in the future.  **BUT...**

This _particular_ problem was a _build_ error, not a test failure or a crash while running the tests, and it failed building PersonNameFormatter itself, not the exhaustive tests.  So my question is Why would the PersonNameFormatter fail to _build_ while trying to build the exhaustive tests, when it built just fine when doing a normal ICU build and running the regular unit tests?  Either I've misdiagnosed the problem or these builds are happening with different tools or build options, and that seems like a real problem.

Yeah, the fact that I'm passing the regular unit tests doesn't guarantee I'll pass the exhaustive tests, but the fact that I built without error doing the normal CI checks _should_ mean I can at least build when trying to build and run the exhaustive tests (unless the build error is in the exhaustive tests, and in this case it wasn't).

FrankYFTang:
> This _particular_ problem was a _build_ error,

I think it was a warning, not an error and I mistakenly think it was an error.

FrankYFTang:
see https://github.com/unicode-org/icu/actions/runs/5765026878/job/15630176054#step:5:107
this is on the regular build. The warning is also there. I think misunderstand that.
I believe the issue is probably Timeout on test. and While I try to figure out what happen I saw this warning and mistakenly think that is the cause. After you fix the warning, one of the test passed (without timeout) while the other still timeout, and I click rerun and it passed. So I think I was wrong to tell you your merged cause the error. It is just a new warning and the breakage is due to the flakiness of our testing and Timeout of tests and have no connection to your tests (beside the fact that you may add more test which may make the timeout happen more often)

richgillam:
>this is on the regular build. The warning is also there. I think misunderstand that.

Oh, okay.  That makes a lot more sense.  I should have caught the warning.  I'll try to be more vigilant next time.

>It is just a new warning and the breakage is due to the flakiness of our testing and Timeout of tests and have no connection to your tests (beside the fact that you may add more test which may make the timeout happen more often)

That's a relief!  By the way, my "exhaustive" test actually doesn't take very long to run, so it's probably not adding more flakiness here (but who knows?).

PR #2539, opened 2023-08-03
ICU-22450 Disable the type_info workaround on newer versions of libstdc++

glandium:
The workaround for http://llvm.org/bugs/show_bug.cgi?id=13364 is related to an ancient version of libstdc++ (4.6). On the flip side, with more recent versions of GCC, the workaround can cause problems under some conditions.

The _GLIBCXX_RELEASE macro was introduced in libstdc++ 7, checking for it not being defined to enable the workaround makes it disabled with libstdc++ >= 7, which don't need the workaround in the first place.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22450


yumaoka:
@glandium Can you sign CLA? Otherwise, we cannot look at your changes.

glandium:
I'm sorting this out through my employer. (an email has been sent to member-services at unicode.org about this. My work email is different than the one in the current commit, though, so I'll update that.

glandium:
The CLA should be sorted out.

yumaoka (approved):
LGTM

glandium:
Can this be merged?

PR #2538, opened 2023-08-03
ICU-20777 Merge likelysubtag implementation

FrankYFTang:
Remove likelySubtags.res (268528 bytes) from the data file and use the langInfo.res (only 76272 bytes) only instead. delegate the likelySubtags code to XLikelySubtags to handle and remove duplicate and unneeded code. This reduce 262+K of duplicate data (both likelySubtags.res and langInfo.res were build from CLDR likelySubtags.xml and contains duplicate information)

* Fix the algorithm to also consider macroregions. 
* Fix some incorrect expectation in the test code. 
* We need another PR to change the CLDR conversion tool to stop generating misc/likelySubtags.txt but prefer to put into a separate PR.
* Merge the loclikelysubtags dependency to resourcebundle since now they are interdependent. 
* One difference between the C++ and Java version is Java version use Region to find out Macroregion but since Region is in i18n directory instead of common, we avoid the dependency by reading the resource in loclikelysubtags.cpp instead. 
* Add data driven unit tests by using the  likelySubtags.txt file from 
https://github.com/unicode-org/cldr/blob/a9a7d97b74d7ae7a6245d19dc55b66a35ea7a8c7/common/testData/localeIdentifiers/likelySubtags.txt
  * per Mark suggestion, logKnownIssue if the Locale is  
         "und-((001|AQ|CC|SL|SS|ZM)|-Latn-.*"
* Data Driven test is too slow to run in C++ so we only run it under  -e mode but skip it in quick mode.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20777
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [X] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [X] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2543, opened 2023-08-07
ICU-22453 Fix non null terminated buffer issue.

FrankYFTang:
UniqueCharStrings was designed to ref to strings return by resource bundle which the strings will not be freed and is used in two places- inside locid.cpp and loclikelysubtags.cpp

But two months ago I misuse that in https://github.com/unicode-org/icu/pull/2458 to also store stack allocated UnicodeString that 1) may not be null terminated and 2) will be freed from the stack before the destruction of UniqueCharStrings. We discover the issue while work on https://github.com/unicode-org/icu/pull/2538

This PR has been changed several time because I found the fix is not safe nor efficient due to the fact that 
a. While the data came from the ResourceBundle, we should use the internal data without allocate unecessary memory.
b. While the data is generated from some computation, we need to keep the memory around until the destruction of the  UniqueCharStrings because the uhash internally use them and expect it to be null terminiated. 

After some face to face discussion with @sffc, we belive it should be address by changing UniqueCharStrings
1. Have a function add() which take a const chart16_t * , which UniqueCharStrings can safely assume it is always null-terminiated and the memory is guanantee around until the destruction of the UniqueCharStrings. We will use this version for all the data getting from the resource bundle. The caller will pass in the the getTerminatedBuffer() result from the UnicodeString. 
2. Have a new function addByValue() which take UnicodeString by value instead of const ref. In this version, we first try to call the getTerminatedBuffer() and do we already store it, if so, just return the stored value. Otherwise, clone a copy and store it and use the getTerminatedBuffer() of the newly clone() copy. We add a UVector to UniqueCharStrings to track these UnicodeString which need to be freed in the destruction. We use addByValue for the result of toLanguage(), toScript() and toRegion() calls.  We also make sure the UnicodeString created by these three functions were null-terminated in this PR.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22453

### DIFF
--- a/icu4c/source/common/loclikelysubtags.cpp
+++ b/icu4c/source/common/loclikelysubtags.cpp
@@ -241,9 +241,11 @@ private:
                 return false;
             }
             for (int i = 0; i < length; ++i) {
-                stringArray.getValue(i, value);  // returns true because i < length
-                rawIndexes[i] = strings.add(value.getUnicodeString(errorCode), errorCode);
-                if (U_FAILURE(errorCode)) { return false; }
+                if (stringArray.getValue(i, value)) {  // returns true because i < length
+                    int32_t strLength = 0;
+                    rawIndexes[i] = strings.add(value.getString(strLength, errorCode), errorCode);
+                    if (U_FAILURE(errorCode)) { return false; }
+                }
             }
         }
         return true;
@@ -261,10 +263,10 @@ private:
         lang[0] = 'a' + ((encoded % 27) - 1);
         lang[1] = 'a' + (((encoded / 27 ) % 27) - 1);
         if (encoded / (27 * 27) == 0) {
-            return UnicodeString(lang, 2);
+            return UnicodeString(lang, 2, US_INV);
         }
         lang[2] = 'a' + ((encoded / (27 * 27)) - 1);
-        return UnicodeString(lang, 3);
+        return UnicodeString(lang, 3, US_INV);
     }
     UnicodeString toScript(int encoded) {
         if (encoded == 0) {
@@ -278,7 +280,8 @@ private:
         if (script == nullptr) {
             return UNICODE_STRING_SIMPLE("");
         }
-        return UnicodeString(script, 4);
+        U_ASSERT(uprv_strlen(script) == 4);
+        return UnicodeString(script, 4, US_INV);
     }
     UnicodeString m49IndexToCode(const ResourceArray &m49Array, ResourceValue &value, int index, UErrorCode &errorCode) {
         if (U_FAILURE(errorCode)) {
@@ -306,7 +309,7 @@ private:
         char region[2];
         region[0] = 'A' + ((encoded % 27) - 1);
         region[1] = 'A' + (((encoded / 27) % 27) - 1);
-        return UnicodeString(region, 2);
+        return UnicodeString(region, 2, US_INV);
     }
 
     bool readLSREncodedStrings(const ResourceTable &table, const char* key, ResourceValue &value, const ResourceArray& m49Array,
@@ -321,9 +324,10 @@ private:
                 return false;
             }
             for (int i = 0; i < length; ++i) {
-                rawIndexes[i*3] = strings.add(toLanguage(vectors[i]), errorCode);
-                rawIndexes[i*3+1] = strings.add(toScript(vectors[i]), errorCode);
-                rawIndexes[i*3+2] = strings.add(toRegion(m49Array, value, vectors[i], errorCode), errorCode);
+                rawIndexes[i*3] = strings.addByValue(toLanguage(vectors[i]), errorCode);
+                rawIndexes[i*3+1] = strings.addByValue(toScript(vectors[i]), errorCode);
+                rawIndexes[i*3+2] = strings.addByValue(
+                    toRegion(m49Array, value, vectors[i], errorCode), errorCode);
                 if (U_FAILURE(errorCode)) { return false; }
             }
             length *= 3;
--- a/icu4c/source/common/uniquecharstr.h
+++ b/icu4c/source/common/uniquecharstr.h
@@ -10,6 +10,7 @@
 #include "charstr.h"
 #include "uassert.h"
 #include "uhash.h"
+#include "cmemory.h"
 
 U_NAMESPACE_BEGIN
 
@@ -47,22 +48,20 @@ public:
     }
 
     /**
-     * Adds a string and returns a unique number for it.
-     * The string's buffer contents must not change, nor move around in memory,
+     * Adds a NUL-terminated string and returns a unique number for it.
+     * The string must not change, nor move around in memory,
      * while this UniqueCharStrings is in use.
-     * The string contents must be NUL-terminated exactly at s.length().
      *
-     * Best used with read-only-alias UnicodeString objects that point to
-     * stable storage, such as strings returned by resource bundle functions.
+     * Best used with string data in a stable storage, such as strings returned
+     * by resource bundle functions.
      */
-    int32_t add(const UnicodeString &s, UErrorCode &errorCode) {
-        if (U_FAILURE(errorCode)) { return 0; }
+    int32_t add(const char16_t*p, UErrorCode &errorCode) {
+        if (U_FAILURE(errorCode)) { return -1; }
         if (isFrozen) {
             errorCode = U_NO_WRITE_PERMISSION;
-            return 0;
+            return -1;
         }
         // The string points into the resource bundle.
-        const char16_t *p = s.getBuffer();
         int32_t oldIndex = uhash_geti(&map, p);
         if (oldIndex != 0) {  // found duplicate
             return oldIndex;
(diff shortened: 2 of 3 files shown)


FrankYFTang:
Sorry, I gave up https://github.com/unicode-org/icu/pull/2541 that PR has something strange and will not pick up my forcepush so I get the diff and create this oen.

icu4c/source/test/intltest/uniquecharstrtst.cpp
```diff
@@ -0,0 +1,53 @@
...
+    UniqueCharStrings strings(status);
+    if (U_FAILURE(status)) {
+       errln("failed to construct UniqueCharStrings - %s", u_errorName(status));
+    }
+    UnicodeString key1("abcde", -1, US_INV);
+    UnicodeString key2(u"fgihjk", 2); // the key should be only "fg" since the
+                                      // length is 2.
+    UnicodeString key3("lmn", -1, US_INV);
+    int32_t index1 = strings.add(key1, status);
+    if (U_FAILURE(status)) {
```
richgillam, line 37: Come to think of it, you could also use `assertSuccess()` here to make things a little simpler...

richgillam (approved):
LGTM.

FrankYFTang:
I read the code again. I think my approach to fix this is wrong. Need to start over again.

FrankYFTang:
Please do NOT look at this PR until I completely fix this issue.

FrankYFTang:
I notice we have a foundamental memory ownership issue, not just null terminated. So I make a bigger change here. 
PTAL

icu4c/source/common/uniquecharstr.h
```diff
@@ -48,21 +50,41 @@ class UniqueCharStrings {
...
+     * If copy is false, then the string's buffer contents must not change,
+     * nor move around in memory, while this UniqueCharStrings is in use.
      * Best used with read-only-alias UnicodeString objects that point to
      * stable storage, such as strings returned by resource bundle functions.
+     * The string contents must be NUL-terminated exactly at s.length().
+     *
+     * If copy is true, a copy of the string will be stored and freed internally.
      */
-    int32_t add(const UnicodeString &s, UErrorCode &errorCode) {
+    int32_t add(const UnicodeString &s, bool copy, UErrorCode &errorCode) {
```
sffc, line 61: Nit: `bool` parameters are not great for readability

icu4c/source/common/uniquecharstr.h
```diff
@@ -48,21 +50,41 @@ class UniqueCharStrings {
 
     /**
      * Adds a string and returns a unique number for it.
-     * The string's buffer contents must not change, nor move around in memory,
-     * while this UniqueCharStrings is in use.
-     * The string contents must be NUL-terminated exactly at s.length().
-     *
+     * If copy is false, then the string's buffer contents must not change,
+     * nor move around in memory, while this UniqueCharStrings is in use.
```
sffc, line 54: Suggestion: `UnicodeString` itself has a constructor that points to other memory:

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeString.html#a28a13b2ef3a75b8a947bcd040a98e206

So, maybe just change this function to take the UnicodeString by value, and call sites that point to constant in-memory data can use the readonly-alias UnicodeString constructor.

icu4c/source/common/uniquecharstr.h
```diff
@@ -48,21 +50,41 @@ class UniqueCharStrings {
...
+            p = cloned->getTerminatedBuffer();
+            copied.adoptElement(cloned, errorCode);
+        } else {
+            p = s.getBuffer();
+            // The string contents must be NUL-terminated exactly at s.length().
+            if (p[s.length()] != '\0') {
+                errorCode = U_ILLEGAL_ARGUMENT_ERROR;
+                return 0;
+            }
+        }
```
sffc, line 87: Actually, why do you need to do any of this? `s` is only ever used as an argument to `strings->appendInvariantChars`. It's not clear why the string needs to be NUL terminated when passing to that function.

Maybe the call to `strings->append(0, errorCode);` a few lines down needs to be moved after the call to `strings->appendInvariantChars`.

sffc (changes requested):

icu4c/source/common/uniquecharstr.h
FrankYFTang: The issue is uhash_puti(&map, const_cast<char16_t *>(p)... )
the map is initialized with  uhash_init(&map, uhash_hashUChars, uhash_compareUChars, uhash_compareLong, &errorCode);
so p need to be processed by uhash_hashUChars and uhash_compareUChars which both expect "null-terminated UChar* string"

```
/**
 * Generate a hash code for a null-terminated UChar* string.  If the
 * string is not null-terminated do not use this function.  Use
 * together with uhash_compareUChars.
 * @param key The string (const UChar*) to hash.
 * @return A hash code for the key.
 */
U_CAPI int32_t U_EXPORT2
uhash_hashUChars(const UHashTok key);
...
/**
 * Comparator for null-terminated UChar* strings.  Use together with
 * uhash_hashUChars.
 * @param key1 The string for comparison
 * @param key2 The string for comparison
 * @return true if key1 and key2 are equal, return false otherwise.
 */
U_CAPI UBool U_EXPORT2
uhash_compareUChars(const UHashTok key1, const UHashTok key2);
```

icu4c/source/common/uniquecharstr.h
sffc, line 61: Why not just make the function be defined to take a NUL-terminated `const char16_t*` with no length parameter and let the callers sort this out?

FrankYFTang: that make sense.

FrankYFTang:
BTW, this PR is currently blocking https://github.com/unicode-org/icu/pull/2538 to work properly on Windows build.

sffc (commented):
This looks fine from a safety perspective, but I still don't understand why you need to have a terminated buffer in order to call `appendInvariantChars`. That function does not retain ownership over the string; it just reads from it and never uses it again. Therefore, it seems unnecessary to store the terminated strings in their own data store `ownedStrings`.

FrankYFTang:
I try the patch in https://github.com/unicode-org/icu/pull/2538 It does fix the windows problem. PTAL

richgillam:
I thought I understood this the first time I looked at it, but I don't now.

Before, this was all about fixing it so that UniqueCharStrings made sure the strings it contained were all null-terminated.  Now you're changing several classes in addition to UniqueCharStrings, which seems to have shifted things so that we're talking about a different solution to a different, but related, problem.

What I think I'm getting from the history here is that your original approach to fixing this was both unsafe and inefficient and that the only good solution was to move some of the intelligence back to the call site, which is the likely-subtags stuff.  Is the likely-subtags stuff the only client of this UniqueCharStrings class, and if not, why don't there need to be changes to its other clients?  How does this change interact with your other change for the likely-subtags logic?  Can you provide an updated description of this PR that explains what you're doing now and how you arrived here?

FrankYFTang:
> I thought I understood this the first time I looked at it, but I don't now.
> 
> Before, this was all about fixing it so that UniqueCharStrings made sure the strings it contained were all null-terminated. Now you're changing several classes in addition to UniqueCharStrings, which seems to have shifted things so that we're talking about a different solution to a different, but related, problem.
> 
> What I think I'm getting from the history here is that your original approach to fixing this was both unsafe and inefficient and that the only good solution was to move some of the intelligence back to the call site, which is the likely-subtags stuff. Is the likely-subtags stuff the only client of this UniqueCharStrings class, and if not, why don't there need to be changes to its other clients? How does this change interact with your other change for the likely-subtags logic? Can you provide an updated description of this PR that explains what you're doing now and how you arrived here?

sorry. see https://github.com/unicode-org/icu/pull/2543#issue-1840240948

icu4c/source/common/uniquecharstr.h
```diff
@@ -90,6 +113,7 @@ class UniqueCharStrings {
 private:
     UHashtable map;
     CharString *strings;
+    UVector keyStore;
```
sffc, line 116: Suggestion: use `class MemoryPool` from cmemory.h instead of `UVector`

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -257,13 +259,15 @@ struct XLikelySubtagsData {
         }
         encoded &= 0x00ffffff;
         encoded %= 27*27*27;
-        char lang[3];
+        char lang[4];
         lang[0] = 'a' + ((encoded % 27) - 1);
         lang[1] = 'a' + (((encoded / 27 ) % 27) - 1);
         if (encoded / (27 * 27) == 0) {
+            lang[2] = '\0';  // Null terminate
             return UnicodeString(lang, 2);
```
sffc, line 267: Issue (pre-existing): this is the slow UnicodeString constructor that pulls in the codepage converter.

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeString.html#aad4fc33d0aac19c45e73cdf7f844ef5d

You should either build up your UnicodeString directly or use the `US_INV` constructor:

https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeString.html#a613a04db62048c98e32112e5864716db

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -257,13 +259,15 @@ struct XLikelySubtagsData {
         }
         encoded &= 0x00ffffff;
         encoded %= 27*27*27;
-        char lang[3];
+        char lang[4];
```
sffc, line 262: Nit: You don't need the NUL termination in this function any more, so you can remove it from this PR to reduce the diff size.

sffc (commented):
I think this is the best architecture so far. Some minor comments

FrankYFTang:
@sffc - PTAL . I also tried it in https://github.com/unicode-org/icu/pull/2538 and that still fix the Windows problem

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -241,9 +241,11 @@ struct XLikelySubtagsData {
                 return false;
             }
             for (int i = 0; i < length; ++i) {
-                stringArray.getValue(i, value);  // returns true because i < length
-                rawIndexes[i] = strings.add(value.getUnicodeString(errorCode), errorCode);
-                if (U_FAILURE(errorCode)) { return false; }
+                if (stringArray.getValue(i, value)) {  // returns true because i < length
+                    int32_t strLength = 0;
```
sffc, line 245: question: why strLength = 0?

sffc (approved):

icu4c/source/common/loclikelysubtags.cpp
FrankYFTang: I copy the code from https://github.com/unicode-org/icu/blob/56850c9a42eab1a28c65e0c177accd9ea9272a5d/icu4c/source/common/resource.h#L142

PR #2545, opened 2023-08-09
ICU-22452 API for resolving preferred IANA zone ID from a zone ID.

yumaoka:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22452


icu4c/source/data/misc/timezoneTypes.txt
```diff
@@ -5,14 +5,47 @@ timezoneTypes:table(nofallback){
...
+            "Asia:Katmandu"{"Asia/Kathmandu"}
+            "Asia:Rangoon"{"Asia/Yangon"}
+            "Asia:Saigon"{"Asia/Ho_Chi_Minh"}
+            "Atlantic:Faeroe"{"Atlantic/Faroe"}
+            "Europe:Kiev"{"Europe/Kyiv"}
+            "Pacific:Enderbury"{"Pacific/Kanton"}
+            "Pacific:Ponape"{"Pacific/Pohnpei"}
+            "Pacific:Truk"{"Pacific/Chuuk"}
+        }
+    }
```
richgillam, line 48: I see in the code what this new table is doing, but it wasn't clear to me from the rest of this PR what all the other changes to this file are doing or how they relate to the new API you're adding.  Maybe this is obvious to everybody who's spent a lot of time in the time-zone stuff, but would you mind explaining for the benefit of people like me who haven't?

richgillam (dismissed):
You probably want to wait for review feedback from somebody who actually understands the time-zone stuff.  As for me, it all seems pretty straightforward, except for many of the changes in `timezoneTypes.txt` other than the new `ianaMap` resource.  I didn't know what all the other changes in that file were all about.

icu4c/source/data/misc/timezoneTypes.txt
richgillam: Whoops.  Didn't see the "Incomplete" label until now.  Maybe that explains my questions...

yumaoka: @richgillam 
>but it wasn't clear to me from the rest of this PR what all the other changes to this file are doing or how they relate to the new API you're adding.

To support the new data structure and produce a new table in ICU data files, we need to update cldr2icu tool. These changes found under tools were made for the data mapping. All other files in icu4c/icu4j are all related to 1) new data .txt 2) API and implementation 3) test cases.

Do you see something unrelated to this work?

richgillam: My question was specifically about `timeZoneTypes.txt` and the changes to resources in it other than the new `ianaMap` resource.  Are the other changes in there somehow necessary to make the new `getIanaID()` functions work, or were they just other changes to the time-zone data in CLDR that came "along for the ride" when you re-ran the CLDR-to-ICU tools?  If they're part of the implementation of `getIanaID()`, I couldn't figure out what they were doing.

yumaoka: OK. There are a few goals in CLDR-14453. First, sort out what set is appropriate for CLDR canonical zone IDs. And the second part is to provide mapping from CLDR canonical zone ID to IANA preferred zone ID. On CLDR side, we come up with new policy for a set of CLDR canonical zones. This was required to define IANA preferred zone ID mapping. This is the reason why you see a bunch of changes in timeZoneTypes.txt.

richgillam: That makes sense.  Thank you for the explanation!  I guess those changes either don't result in a caller-visible change in behavior, or CLDR has judged any changes in behavior to be beneficial?

icu4c/source/i18n/ucal.cpp
```diff
@@ -628,6 +628,27 @@ ucal_getCanonicalTimeZoneID(const char16_t* id, int32_t len,
     return reslen;
 }
 
+U_DRAFT int32_t U_EXPORT2
+ucal_getIanaTimeZoneID(const char16_t* id, int32_t len,
+                        char16_t* result, int32_t resultCapacity, UErrorCode* status)
+{
+    if (status == 0 || U_FAILURE(*status)) {
```
FrankYFTang, line 635: Why do we need to check  status == 0 here but not other methods? 
I only see ucal_getCanonicalTimeZoneID in this file perform such check.

icu4c/source/i18n/ucal.cpp
```diff
@@ -628,6 +628,27 @@ ucal_getCanonicalTimeZoneID(const char16_t* id, int32_t len,
...
+        return 0;
+    }
+    if (id == 0 || len == 0 || result == 0 || resultCapacity <= 0) {
+        *status = U_ILLEGAL_ARGUMENT_ERROR;
+        return 0;
+    }
+    int32_t reslen = 0;
```
FrankYFTang, line 642: I thnik we only need
```
    UnicodeString ianaID;
    return TimeZone::getIanaID(UnicodeString(id, len), ianaID, *status)
                  .extract(result, resultCapacity, *status);
```
Since both getIanaID and extract should check *status and return immediately anyway.

icu4c/source/i18n/timezone.cpp
```diff
@@ -1591,6 +1591,21 @@ TimeZone::getCanonicalID(const UnicodeString& id, UnicodeString& canonicalID, UB
...
 
+UnicodeString&
+TimeZone::getIanaID(const UnicodeString& id, UnicodeString& ianaID, UErrorCode& status)
+{
+    ianaID.remove();
+    if (U_FAILURE(status)) {
+        return ianaID;
+    }
+    if (id.compare(ConstChar16Ptr(UNKNOWN_ZONE_ID), UNKNOWN_ZONE_ID_LENGTH) == 0) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
```
FrankYFTang, line 1602: ZoneMeta::getIanaID will set the returned ianaID to bogus, should we also set ianaID to bogus here for this error?

icu4c/source/i18n/unicode/timezone.h
```diff
@@ -444,6 +444,31 @@ class U_I18N_API TimeZone : public UObject {
...
+     * These canonical time zone IDs in CLDR were based on very old version of the time zone database.
+     * In the IANA time zone database, some IDs were updated since then. This API returns a newer
+     * time zone ID. For example, CLDR defines "Asia/Calcutta" as the canonical time zone ID. This
+     * method returns "Asia/Kolkata" instead.
+     *
+     * @param id        The input time zone ID.
+     * @param ianaID    Receives the preferred time zone ID in the IANA time zone database.
+     * @param status    Receives the status.  When the given time zone ID is not a known system time zone
+     *                  ID, U_ILLEGAL_ARGUMENT_ERROR is set.
+     * @return  A reference to the result.
```
FrankYFTang, line 466: Should we also document that the return value could be bogus for error and "" if the input is "Etc/Unkown" ? or should both case return bogus string ?

icu4c/source/i18n/zonemeta.cpp
```diff
@@ -389,6 +390,35 @@ ZoneMeta::getCanonicalCLDRID(const TimeZone& tz) {
...
+    /* int32_t keyLen = */ tmpKey.extract(0, tmpKey.length(), keyBuf, sizeof(keyBuf), US_INV);
+
+    StackUResourceBundle r;
+    ures_openDirectFillIn(r.getAlias(), nullptr, gKeyTypeData, &tmpStatus);
+    ures_getByKey(r.getAlias(), gIanaMapTag, r.getAlias(), &tmpStatus);
+    ures_getByKey(r.getAlias(), gTimezoneTag, r.getAlias(), &tmpStatus);
+    int32_t tmpLen = 0;
+    const char16_t* tmpIana = ures_getStringByKey(r.getAlias(), keyBuf, &tmpLen, &tmpStatus);
+    if (U_SUCCESS(tmpStatus)) {
+        ianaID.setTo(tmpIana, tmpLen);
```
FrankYFTang, line 415: since tmpIana is returned by ures_getStringByKey that is "a pointer to a zero-terminated UChar array which lives in a memory mapped/DLL file." Should we use the  ianaID.setTo(true, tmpIana, tmpLen);
to save some memory?
https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1UnicodeString.html#a9a2f115fb0975766d9b8823d98622295

icu4c/source/i18n/zonemeta.cpp
```diff
@@ -389,6 +390,35 @@ ZoneMeta::getCanonicalCLDRID(const TimeZone& tz) {
...
+    } else {
+        ianaID.setTo(canonicalID, -1);
```
FrankYFTang, line 417: Since the canonicalID is returned by the version which "overload method returns a persistent const char16_t*, which is guaranteed to persist    * (a pointer to a resource)."  of getCanonicalCLDRID() 

I think we should also use ianaID.setTo(true, canonicalID, -1);

https://github.com/unicode-org/icu/blob/cc2ddc0d113aa4db6efa6b071eca172e4a72f6cd/icu4c/source/i18n/zonemeta.h#L50

to save internal memory allocatioun for ianaID

icu4c/source/test/cintltst/ccaltst.c
```diff
@@ -2826,4 +2827,45 @@ TestFWWithISO8601() {
...
+        {0,                     UNKNOWN},
+        {u"America/New_York",   u"America/New_York"},
+        {u"Asia/Calcutta",      u"Asia/Kolkata"},
+        {u"Europe/Kiev",        u"Europe/Kyiv"},
+        {u"Europe/Zaporozhye",  u"Europe/Kyiv"},
+        {u"Etc/GMT-1",          u"Etc/GMT-1"},
+        {u"Etc/GMT+20",         UNKNOWN},
+        {u"PST8PDT",            u"PST8PDT"},
+        {u"GMT-08:00",          UNKNOWN},
+        {0,                     0}
```
FrankYFTang, line 2849: we should aslo test the case which the id is UNKNOWN and make sure the return of that is correct.

icu4c/source/test/intltest/tztest.cpp
```diff
@@ -2623,4 +2625,37 @@ void TimeZoneTest::TestRawOffsetAndOffsetConsistency22041() {
...
+        {0,                     UNKNOWN},
+        {u"America/New_York",   u"America/New_York"},
+        {u"Asia/Calcutta",      u"Asia/Kolkata"},
+        {u"Europe/Kiev",        u"Europe/Kyiv"},
+        {u"Europe/Zaporozhye",  u"Europe/Kyiv"},
+        {u"Etc/GMT-1",          u"Etc/GMT-1"},
+        {u"Etc/GMT+20",         UNKNOWN},
+        {u"PST8PDT",            u"PST8PDT"},
+        {u"GMT-08:00",          UNKNOWN},
+        {0,                     0}
```
FrankYFTang, line 2645: as C test, we should also test the condition that id is UNKNOWN

icu4c/source/test/intltest/tztest.cpp
```diff
@@ -2623,4 +2625,37 @@ void TimeZoneTest::TestRawOffsetAndOffsetConsistency22041() {
...
+    };
+
+    for (int32_t i = 0; TESTDATA[i].expected != 0; i++) {
+        UErrorCode sts = U_ZERO_ERROR;
+        UnicodeString inputID(TESTDATA[i].id);
+        UnicodeString ianaID;
+
+        TimeZone::getIanaID(inputID, ianaID, sts);
```
FrankYFTang, line 2653: we need to test the return value of TimeZone::getIanaID and not assume the return value is always the same as ianaID from blackbox testing point of view. Maybe assert ianaID and the return value is always contain the same?

icu4c/source/test/intltest/tztest.cpp
```diff
@@ -2623,4 +2625,37 @@ void TimeZoneTest::TestRawOffsetAndOffsetConsistency22041() {
...
+        if (u_strcmp(TESTDATA[i].expected, UNKNOWN) == 0) {
+            assertEquals(inputID + " should fail", (int32_t)U_ILLEGAL_ARGUMENT_ERROR, sts);
```
FrankYFTang, line 2655: we should also assert the return value is set to bogus here

icu4c/source/i18n/ucal.cpp
yumaoka: There are other functions which also checks nullptr, such as _createTimeZone, ucal_getDefaultTimeZone, ucal_getHostTimeZone, and some others. Some of the does this differently. For example:

```
U_CAPI int32_t U_EXPORT2
ucal_getDefaultTimeZone(char16_t* result, int32_t resultCapacity, UErrorCode* ec) {
    int32_t len = 0;
    if (ec != nullptr && U_SUCCESS(*ec)) {
        TimeZone* zone = TimeZone::createDefault();
        if (zone == nullptr) {
            *ec = U_MEMORY_ALLOCATION_ERROR;
        } else {
            UnicodeString id;
            zone->getID(id);
            delete zone;
            len = id.extract(result, resultCapacity, *ec);
        }
    }
    return len;
}
```
I think ICU API should be consistent - maybe just not checking it and let the calling application crash. Anyway, we might discuss about our policy for this next week.

pedberg-icu: Agree about check status == 0, per coding guidelines we should not be doing that: https://unicode-org.github.io/icu/userguide/dev/codingguidelines.html#sample-function-with-error-checking

icu4c/source/i18n/timezone.cpp
pedberg-icu: I think that is in the category of a "nice to have" consistency fix, but maybe it can be fixed later, should not block this PR.

yumaoka: Thanks for pointing out. I'll make this change.

icu4c/source/i18n/ucal.cpp
yumaoka: Will check it again.

icu4c/source/i18n/unicode/timezone.h
yumaoka: I will add a note in the API reference, explaining about "Etc/Unknown".

icu4c/source/i18n/zonemeta.cpp
yumaoka: Thanks for the suggestion. Will do.

icu4c/source/test/cintltst/ccaltst.c
yumaoka: Hmm.. it looks I missed the case. Will do.

icu4c/source/test/intltest/tztest.cpp
yumaoka: Yes.

icu4j/main/classes/core/src/com/ibm/icu/impl/ZoneMeta.java
```diff
@@ -69,7 +69,7 @@ private static synchronized Set<String> getSystemZIDs() {
             systemZones = REF_SYSTEM_ZONES.get();
         }
         if (systemZones == null) {
-            Set<String> systemIDs = new TreeSet<String>();
+            Set<String> systemIDs = new TreeSet<>();
```
FrankYFTang, line 72: Why are these changes here? They are unrelated and unneeded for this new API, right?

icu4c/source/test/intltest/tztest.cpp
yumaoka: At this moment, it's not an API contact that bogus is set. Setting bogus is this implementation is for safety purpose. I'm OK add the check in the test case although.

icu4j/main/classes/core/src/com/ibm/icu/impl/ZoneMeta.java
yumaoka: By the standard Eclipse project setting. When you touch the file, it automatically optimize this to Java 7+ style.
We used to control coding style through Eclipse project file settings, but I know more people not using the env these days.

icu4j/main/classes/core/src/com/ibm/icu/impl/ZoneMeta.java
```diff
@@ -419,6 +419,38 @@ private static String findCLDRCanonicalID(String tzid) {
...
+     * is not known, this method returns null.
+     *
+     * @param tzid  An input zone ID.
+     * @return  A primary IANA zone ID equivalent to the input zone ID.
+     */
+    public static String getIanaID(String tzid) {
+        String ianaID = null;
+        // First, get CLDR canonical ID
+        String canonicalID = getCanonicalCLDRID(tzid);
+        if (canonicalID != null) {
```
FrankYFTang, line 433: how about we write write it in a way which has less condictional block in the function as:
```
if (canonicalID == null) return null;
try {
     UResourceBundle keyTypeData = UResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME,
                        "keyTypeData", ICUResourceBundle.ICU_DATA_CLASS_LOADER);
     UResourceBundle ianaMap = keyTypeData.get("ianaMap");
     UResourceBundle ianaTzMap = ianaMap.get("timezone");
     try {
         return ianaTzMap.getString(canonicalID.replace('/', ':'));
     } catch (MissingResourceException e) {
         // No IANA zone ID mapping. In this case, ianaId set by getCanonicalCLDRID()
         // is also a primary IANA id.
        return canonicalID;
     }
 } catch (MissingResourceException e) {
     // When CLDR canonical ID is not found for the input tzid.\
     return null;
 }
```

icu4c/source/i18n/ucal.cpp
pedberg-icu: And if id == 0 || len == 0  then UnicodeString(id, len) will produce an empty string which will presumably cause TimeZone::getIanaID to fail with an error.

So yes, Frank's suggestion  probably simplifies things. But... the current implementation still works fine. Could be simplified in a future PR.

icu4c/source/i18n/unicode/timezone.h
pedberg-icu: Since this has a UErrorCode parameter, not sure we want to make setting return to bogus part of the API contract.

icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
```diff
@@ -1166,6 +1166,38 @@ public static String getCanonicalID(String id, boolean[] isSystemID) {
...
+     *
+     * @param id    The input time zone ID.
+     * @return  The preferred time zone ID in the IANA time zone database, or {@link TimeZone#UNKNOWN_ZONE_ID}
+     * if the input ID is not a system ID.
+     * @see #getCanonicalID(String)
+     * @draft ICU 74
+     */
+    public static String getIanaID(String id) {
+        String ianaId = TimeZone.UNKNOWN_ZONE_ID;
+        if (id != null && id.length() != 0) {
```
FrankYFTang, line 1190: We have a 3 level nested if block here. Could we reduce that by 
```
        if (id == null || id.length() == 0 || id.equals(TimeZone.UNKNOWN_ZONE_ID)) {
            return ianaId;
        }
        String tmpIanaId = ZoneMeta.getIanaID(id);
        if (tmpIanaId == null) {
            return ianaId;
        }
        return tmpIanaId;
```

icu4j/main/tests/core/src/com/ibm/icu/dev/test/timezone/TimeZoneTest.java
```diff
@@ -2390,6 +2390,29 @@ public void TestRawOffsetAndOffsetConsistency22041() {
...
+                {null,                  UNKNOWN},
+                {"America/New_York",    "America/New_York"},
+                {"Asia/Calcutta",       "Asia/Kolkata"},
+                {"Europe/Kiev",         "Europe/Kyiv"},
+                {"Europe/Zaporozhye",   "Europe/Kyiv"},
+                {"Etc/GMT-1",           "Etc/GMT-1"},
+                {"Etc/GMT+20",          UNKNOWN},
+                {"PST8PDT",             "PST8PDT"},
+                {"GMT-08:00",           UNKNOWN},
+        };
```
FrankYFTang, line 2409: we should also check the case that the input is NKNOWN

icu4c/source/i18n/zonemeta.cpp
pedberg-icu: Good idea

pedberg-icu (approved):
Frank has a lot of good suggestions for improvements in the ICU4C code. However, I think they can be done in  a follow-up PR, they would not change the basic operation of the code (except in some cases setting strings to bogus, which should not be part of the API contract for APIs that have a UErrorCode param. So I am approving this as is to unblock things...

pedberg-icu:
> ....So I am approving this as is to unblock things...

Oh. But it seems the maven data did not get updated along with the ICU4J data jars, so the maven tests are failing. The maven data should be updated, see process in https://unicode-org.github.io/icu/processes/cldr-icu.html Process section 12

icu4c/source/data/misc/timezoneTypes.txt
yumaoka: >I guess those changes either don't result in a caller-visible change in behavior, or CLDR has judged any changes in behavior to be beneficial?

Let me use an actual example. America/Thunder_Bay and America/Toronto. Previously, these zones are defined differently.
```
Zone America/Toronto	-5:17:32 -	LMT	1895
			-5:00	Canada	E%sT	1919
			-5:00	Toronto	E%sT	1942 Feb  9  2:00s
			-5:00	Canada	E%sT	1946
			-5:00	Toronto	E%sT	1974
			-5:00	Canada	E%sT
Zone America/Thunder_Bay -5:57:00 -	LMT	1895
			-6:00	-	CST	1910
			-5:00	-	EST	1942
			-5:00	Canada	E%sT	1970
			-5:00	Toronto	E%sT	1973
			-5:00	-	EST	1974
			-5:00	Canada	E%sT
```
Above definition is found in 2014g northamerica file. Both were in zone.tab, and some implementations show these two zones in the zone selection list. CLDR treated both of them as canonical zones. Since then, time zone database maintainer decided to make America/Thunder_Bay as "Link" of America/Toronto in backward file as below.
```
Link	America/Toronto		America/Thunder_Bay
```
The maintainer pointed out the difference before 1974 were not coming from reliable source, and concluded not worth keeping America/Thunder_Bay from America/Toronto. So he decided to move America/Thunder_Bay to backward file and use the rule same with America/Toronto.

ICU still support both America/Toronto and America/Thunder_Bay to create a time zone instance. Previously, getCanonicalID("America/Thunder_Bay") returns "America/Thunder_Bay, but with this change, it will be "America/Toronto".

In CLDR, there are two ways to define display names. One is through meta zone. Toronto and New_York share same zone display name like - "Eastern Time". In CLDR, we maintain the display name associated with meta zone America_Eastern, and maintain historic mapping between each canonical zone to meta zone. Another way to define display name is to assign names directly to canonical zone ID. The latter override the meta zone based display names.

So, deprecating America/Thunder_Bay as canonical zone and make it an alias of America/Toronto may change display names if there are zone based display name overrides (not by metazone and mapping), then display names will be changed. But assigning display names directly with zone ID is rare exception in CLDR at this moment. So practically, it should have no impacts.

Only possible change is - location based display name. In CLDR/ICU, date format patterm VVV shows exemplar city name. To produce a display name for VVV, CLDR maintains exemplar city translation. For example, "サンダーベイ" for America/Thunder_Bay in Japanese. To resolve the name, ICU gets a zone's canonical ID first, then look up the time zone exemplar city data. Previously it was America/Thunder_Bay, but it will be changed to America/Toronto. This is the actual impact.

From the time zone database maintainer's point of view, America/Thuder_Bay is only kept for backward compatibility purpose (therefore, it's defined in backward file). This means, time zone database consumer should not include America/Thunder_Bay in the zone selection list. For example, you can no longer select Thunder Bay as the system time zone in the current version of Ubuntu. VVV for America/Thunder_Bay produces "トロント" (Toronto) instead of "サンダーベイ" might not be great, but user is not supposed to select America/Thunder_Bay as system time zone any more.

icu4c/source/i18n/zonemeta.cpp
yumaoka: will make this change.

icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
yumaoka: OK

yumaoka:
> > ....So I am approving this as is to unblock things...
> 
> Oh. But it seems the maven data did not get updated along with the ICU4J data jars, so the maven tests are failing. The maven data should be updated, see process in https://unicode-org.github.io/icu/processes/cldr-icu.html Process section 12

I'm looking at maven one. I think it's related to data packaging, because ant build works.

icu4c/source/test/intltest/tztest.cpp
yumaoka: done

icu4j/main/tests/core/src/com/ibm/icu/dev/test/timezone/TimeZoneTest.java
yumaoka: done

yumaoka:
@pedberg-icu @FrankYFTang I think I fixed issues and updated code based on review feedback. Please review it again. Thanks.

icu4c/source/i18n/ucal.cpp
yumaoka: @pedberg-icu Thanks for the link. I see a bunch of other code checking status == 0 or status != 0 along with U_FAILURE/U_SUCCESS. We may look at these code later.

icu4c/source/data/misc/timezoneTypes.txt
richgillam: I see.  Thanks for the long explanation!  This makes sense, and I'm personally comfortable with what you're describing.

richgillam (approved):
LGTM.  Thanks for helping me understand what you're doing!

pedberg-icu:
The copyright scan issue is fixed by https://github.com/unicode-org/icu/pull/2550, probably need to rebase to pick that up when it is merged. Working on that now.

pedberg-icu:
> The copyright scan issue is fixed by #2550, probably need to rebase to pick that up when it is merged. Working on that now.

OK, I just merged #2550, so you can rebase this to pick up the missing copyright headers tha are blocking the tests for this.

yumaoka:
> > The copyright scan issue is fixed by #2550, probably need to rebase to pick that up when it is merged. Working on that now.
> 
> @yumaoka OK, I just merged #2550, so you can rebase this to pick up the missing copyright headers that are blocking the tests for this.

Thanks @pedberg-icu. I rebased and squashed commits. Let's see if all CI checks passing.

pedberg-icu (approved):

yumaoka:
@FrankYFTang I think I updated the code as you suggested. I think @pedberg-icu is waiting for this, so I will go ahead and merge. If something is still missing, then I can work on another PR. I'm going to merge this now.

icu4c/source/i18n/unicode/timezone.h
yumaoka: done

icu4c/source/i18n/zonemeta.cpp
yumaoka: done

icu4c/source/test/cintltst/ccaltst.c
yumaoka: done

icu4c/source/test/intltest/tztest.cpp
yumaoka: done

icu4c/source/test/intltest/tztest.cpp
yumaoka: done. setToBogus, and test case also checks it.

icu4j/main/classes/core/src/com/ibm/icu/impl/ZoneMeta.java
yumaoka: done

icu4j/main/classes/core/src/com/ibm/icu/util/TimeZone.java
yumaoka: done

icu4c/source/i18n/ucal.cpp
yumaoka: Updated the C wrapper skip nullptr checking and directly calls TimeZone::getIanaID(). So sts==0 check is gone.

pedberg-icu:
@yumaoka Many thanks for seeing this through!

PR #2549, opened 2023-08-10
ICU-22364 Modify ulocimp_getRegionForSupplementalData() to ignore the…

richgillam:
… subdivision code, rather than requiring it to

be "zzzz".

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22364

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -1375,15 +1375,19 @@ ulocimp_getRegionForSupplementalData(const char *localeID, UBool inferRegion,
 
     // First check for rg keyword value
     int32_t rgLen = uloc_getKeywordValue(localeID, "rg", rgBuf, ULOC_RG_BUFLEN, &rgStatus);
-    if (U_FAILURE(rgStatus) || rgLen != 6) {
+    if (U_FAILURE(rgStatus) || rgLen < 3 || rgLen > 7) {
         rgLen = 0;
     } else {
         // rgBuf guaranteed to be zero terminated here, with text len 6
-        char *rgPtr = rgBuf;
-        for (; *rgPtr!= 0; rgPtr++) {
-            *rgPtr = uprv_toupper(*rgPtr);
+        // chop off the subdivision code (which will generally be "zzzz" anyway)
+        if (uprv_isASCIILetter(rgBuf[0])) {
+            rgLen = 2;
+            rgBuf[0] = uprv_toupper(rgBuf[0]);
+            rgBuf[1] = uprv_toupper(rgBuf[1]);
+        } else {
+            // assume three-digit region code
+            rgLen = 3;
         }
-        rgLen = (uprv_strcmp(rgBuf+2, "ZZZZ") == 0)? 2: 0;
     }
 
     if (rgLen == 0) {
--- a/icu4c/source/test/cintltst/ccaltst.c
+++ b/icu4c/source/test/cintltst/ccaltst.c
@@ -101,7 +101,10 @@ static const UCalGetTypeTest ucalGetTypeTests[] = {
     { "fr_CH",                   UCAL_DEFAULT,   "gregorian" },
     { "fr_SA",                   UCAL_DEFAULT,   "islamic-umalqura" },
     { "fr_CH@rg=sazzzz",         UCAL_DEFAULT,   "islamic-umalqura" },
+    { "fr_CH@rg=sa14",           UCAL_DEFAULT,   "islamic-umalqura" },
     { "fr_CH@calendar=japanese;rg=sazzzz", UCAL_DEFAULT, "japanese" },
+    { "fr_CH@rg=twcyi",          UCAL_DEFAULT,   "gregorian" }, // test for ICU-22364
+    { "fr_CH@rg=ugw",            UCAL_DEFAULT,   "gregorian" }, // test for ICU-22364
     { "fr_TH@rg=SA",             UCAL_DEFAULT,   "buddhist"  }, /* ignore malformed rg tag */
     { "th@rg=SA",                UCAL_DEFAULT,   "buddhist"  }, /* ignore malformed rg tag */
     { "",                        UCAL_GREGORIAN, "gregorian" },
@@ -1613,7 +1616,7 @@ void TestGregorianChange() {
 }
 
 static void TestGetKeywordValuesForLocale() {
-#define PREFERRED_SIZE 16
+#define PREFERRED_SIZE 26
 #define MAX_NUMBER_OF_KEYWORDS 5
     const char *PREFERRED[PREFERRED_SIZE][MAX_NUMBER_OF_KEYWORDS+1] = {
             { "root",        "gregorian", NULL, NULL, NULL, NULL },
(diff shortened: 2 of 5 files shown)


richgillam:
There is no unit test; this is an internal API, and I'm not sure this change will have an effect on any existing behavior.

I have some ideas about how to do https://unicode-org.atlassian.net/browse/ICU-22449 that might involve this, and I'm thinking my fix for that ticket can include a unit-test case that would test this behavior as well.  Alternatively (or in addition), I could update any existing unit tests that use the @rg subtag to include variants that use something other than "zzzz" as the subdivision code...

icu4c/source/common/loclikely.cpp
```diff
@@ -1379,11 +1379,10 @@ ulocimp_getRegionForSupplementalData(const char *localeID, UBool inferRegion,
...
-        char *rgPtr = rgBuf;
-        for (; *rgPtr!= 0; rgPtr++) {
-            *rgPtr = uprv_toupper(*rgPtr);
-        }
-        rgLen = (uprv_strcmp(rgBuf+2, "ZZZZ") == 0)? 2: 0;
+        // chop off the subdivision code (which will generally be "zzzz" anyway)
+        rgLen = 2;
+        rgBuf[0] = uprv_toupper(rgBuf[0]);
+        rgBuf[1] = uprv_toupper(rgBuf[1]);
     }
```
pedberg-icu, line 1386: And it does look like 3-digit regions are not allowed here (just confirming): https://www.unicode.org/reports/tr35/#RegionOverride

pedberg-icu (approved):

FrankYFTang:
> There is no unit test; this is an internal API, and I'm not sure this change will have an effect on any existing behavior.

Then why we need to change this function here? A change without a test to trigger the execution to validate the change of before/after behavior seems a unnecessary change to me. How about test it by calling ucal_getKeywordValuesForLocale ? 
If you call with locale = "zh-CN-u-rg-twzzzz" vs "zh-CN-u-rg-twcyi" I think they will give you different answer before your fix but the same after your fix.

richgillam:
>Then why we need to change this function here?

I was doing Apple-specific work where it made a difference (actually, I wrote a unit test where I forgot and used "xxxx" instead and then got all confused when it failed), and it prompted a generic question about why we're requiring "zzzz" here when we're not doing anything with the value in the first place.

It does look like there are some existing unit tests where I can add a line to verify this change, so I'll do that.  My original plan was to add a test case that covers this to whatever I end up doing for https://unicode-org.atlassian.net/browse/ICU-22449.  I'll probably still do that too.

macchiati:
It is basically to allow for the fact that 2 letters alone would be illegal
bcp47.

richgillam:
>How about test it by calling ucal_getKeywordValuesForLocale ?

I actually ended up doing, and it worked well.  Thanks for the suggestion.

icu4c/source/common/loclikely.cpp
```diff
@@ -1379,11 +1379,10 @@ ulocimp_getRegionForSupplementalData(const char *localeID, UBool inferRegion,
         rgLen = 0;
```
FrankYFTang, line 1379: I think we have more problem here. Because the if / else mandate the rgLen to be 6.
But the definition of https://unicode.org/reports/tr35/tr35.html#RegionOverride
said 
"The value is a [unicode_subdivision_id](https://unicode.org/reports/tr35/tr35.html#unicode_subdivision_id) of type “unknown” or “regular”; this consists of a [unicode_region_subtag](https://unicode.org/reports/tr35/tr35.html#unicode_region_subtag) for a regular region (not a macroregion), suffixed either by “zzzz” (case is not significant) to designate the region as a whole, or by a unicode_subdivision_suffix to provide more specificity."

And then
unicode_subdivision_id	= unicode_region_subtag unicode_subdivision_suffix ;
unicode_subdivision_suffix  = alphanum{1,4} ;
 
so in theory unicode_region_subtag could be 2 - 3  and unicode_subdivision_suffix could be 1 - 4 so a valid rgLen should be 3 - 7, not just 6.

For example, if we look at https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml

the following are valid rg value but rgLen != 6
at1
aeaj
afbal

Should we change the if statement to

```
if (U_FAILURE(rgStatus) || rgLen < 3 || rgLen > 7) {
```
or 
```
if (U_FAILURE(rgStatus) || rgLen < 3 || rgLen > 6) {
```
if we assume there are no 3 digits unicode_region_subtag ?

icu4c/source/test/cintltst/ccaltst.c
```diff
@@ -100,6 +100,7 @@ static const UCalGetTypeTest ucalGetTypeTests[] = {
     { "fr_CH",                   UCAL_DEFAULT,   "gregorian" },
     { "fr_SA",                   UCAL_DEFAULT,   "islamic-umalqura" },
     { "fr_CH@rg=sazzzz",         UCAL_DEFAULT,   "islamic-umalqura" },
+    { "fr_CH@rg=sawxyz",         UCAL_DEFAULT,   "islamic-umalqura" }, /* ignore bogus subdivision tag (ICU-22364) */
```
FrankYFTang, line 103: sawxyz is not listed in https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml

how about 
```
{ "fr_CH@rg=sa14",         UCAL_DEFAULT,   "islamic-umalqura" },
```
instead and also try
rg as "twcyi" (listed in https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml )
rg as "ugw"

icu4c/source/test/cintltst/ccaltst.c
```diff
@@ -1631,8 +1632,12 @@ static void TestGetKeywordValuesForLocale() {
             { "zh_TW",       "gregorian", "roc", "chinese", NULL, NULL },
             { "ar_IR",       "persian", "gregorian", "islamic", "islamic-civil", "islamic-tbla" },
             { "th@rg=SAZZZZ", "islamic-umalqura", "gregorian", "islamic", "islamic-rgsa", NULL },
+        
+            // test for ICU-22364
+            { "zh_CN@rg=TWzzzz", "gregorian", "roc", "chinese", NULL, NULL },
+            { "zh_CN@rg=TWxxxx", "gregorian", "roc", "chinese", NULL, NULL },
```
FrankYFTang, line 1638: TWxxxx is not listed in https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml

macchiati:
Good catches. Minimum is 3, max is currently 6. (It is possible to reach 7
in the future with eg 666zzzz, but only if ISO reuses 2 letter codes)

macchiati:
With xxxx the test might be that a syntactic but illegal code works.

FrankYFTang:
Since you are fixing this. I think we should at least add the test to make sure valid rg type work: "-u-rg-ugw"  "-u-rg-sa14" "-u-rg-twcyi" 
(Look at https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml and pick some valid unicode_subdivision_id  which length 3, 4, 5, 6 that  unicode_subdivision_suffix = 
alpha{1} (such as "ara")
alpha{2} (such as "aeaj" )
alpha{3} (such as "afbal")
alpha{4} (such as "thzzzz" )
digit{1} (such as "at1")
digit{2} (such as "ad02" )
digit{3} (such as "cz711")
digit{4}  (I cannot find any valid unicode_subdivision_id with unicode_subdivision_suffix = digit{4}  from https://github.com/unicode-org/cldr/blob/main/common/supplemental/subdivisions.xml  so maybe just ignore that)

To test, you can always use "TW" as the region tag and with the above as rg type. If rg is not in effect, then the rg will not override theTW and the result will have "gregorian", "roc", "chinese" as the calendar but if the rg is in effect, you should not see "roc" in the list. (unless. you have rg prefix with "tw")

icu4c/source/common/loclikely.cpp
richgillam: >unicode_subdivision_suffix = alphanum{1,4} ;

Ugh.  You're right.  I'll fix this.

I notice in subdivisions.xml, there are some five-letter subdivision codes, but TR35 says one to four letters.  The first two letters are always the same as the region code, so I assume the "subdivision code" is just the characters after the region code?  That is, I'd say `en_GB@rg=USca`, not `en_GB@rg=USusca`?  Or, to take a five-letter example, `en_GB@rg=AUnsw` and not `en_GB@AUaunsw`?

icu4c/source/test/cintltst/ccaltst.c
richgillam: Okay, I'll change this and add the others you're suggesting.

richgillam:
> Since you are fixing this. I think we should at least add the test to make sure valid rg type work: "-u-rg-ugw" "-u-rg-sa14" "-u-rg-twcyi"

It seems like overkill to be that comprehensive, but okay, I'll do it.

macchiati:
Here are all the currently valid subdivisions:
https://github.com/unicode-org/cldr/blob/main/common/validity/subdivision.xml

(The code does include the first two letters of the region code. Currently
there are no 3-digit, but that could change in the future. So best would be:

([A-Za-z]{2}|[0-9]{3})[A-Za-z0-9]{1,4}

Note: ISO only allows 3 letters after the region. That's why we chose zzzz
as the 'unknown', so that we could never collide with a valid ISO code. I
was not clear in a previous comment. So any CLDR-derived codes would be 4
letters.

macchiati (approved):

icu4c/source/common/loclikely.cpp
macchiati: Yes, en_GB@rg=USca (or equivalently en-GB-u-rg-usca)
It is possible in the future to have 3-digit regions, so 7 would be safer as long as you are touching the code.

richgillam: That means actually parsing the string, which is something we've avoided up until now.  I was kind of hoping we could avoid going there until we really knew we needed it, but I've gone back, added a test, and done a really cheap and hacky version of paring the string to get the region code.  Let me know if I need to go to greater lengths.

icu4c/source/common/loclikely.cpp
macchiati: It isn't currently. And I hope we never face it. 

But there is provision in BCP47 to require use of 3-digit codes if ISO reuses 2-letter codes. We'd have to change the spec at that point, but I wouldn't want ICU to blow up.

macchiati (dismissed):

FrankYFTang (approved):

FrankYFTang:
It will be nice if we can also fix Java (could be done in a later PR) see https://github.com/unicode-org/icu/blob/27181e36a6bed372482e5a11c25009fa4dd971f1/icu4j/main/classes/core/src/com/ibm/icu/util/ULocale.java#L981

richgillam:
> It will be nice if we can also fix Java (could be done in a later PR)

Crud.  Didn't realize this was an issue on the Java side too (and I should have).  This was pretty easy to port over, so I'm just adding a new commit here with the Java code.

pedberg-icu (approved):

FrankYFTang (approved):
Thanks for adding the Java changes. LGTM

PR #2555, opened 2023-08-18
ICU-22325 integrate release 44 alpha0 to ICU

DraganBesevic:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22325

Commit 1 is the binary part, commit 2 is source code and data.

Integrated CLDR release-44-alpha0 plus the changes from the following PRs:
* https://github.com/unicode-org/cldr/pull/3187 CLDR-16986 Restore Persian calendar time formats for ckb_IR
* https://github.com/unicode-org/cldr/pull/3188 CLDR-16888 Regenerate unitPreferencesTest.txt
* https://github.com/unicode-org/cldr/pull/3192 CLDR-16995 Add space between variable $IPA_CONSONANT and combining mark
* https://github.com/unicode-org/cldr/pull/3195 CLDR-17000 oc_ES, promote some latn miscPatterns items to contributed
* https://github.com/unicode-org/cldr/pull/3196 CLDR-17001 en, add display name for locale vmw Makhuwa
* https://github.com/unicode-org/cldr/pull/3199 CLDR-17003 fix the currency formats unit pattern
* https://github.com/unicode-org/cldr/pull/3202 CLDR-17002 lang/script display names from ↑↑↑ -> code value if likely real name, did not have ↑↑↑

Fixed ICUCurrencyDataDisplayInfoProvider::getPluralName to properly handle the fallback setting: if fallback is false it should not prevent inheriting name data across the different plural and display names for a particular currency in a particular locale.


icu4c/source/data/curr/af.txt
```diff
@@ -689,754 +689,42 @@ af{
...
     Currencies%variant{
         RUB{"₽"}
-        TRY{"TL"}
     }
     CurrencyPlurals{
         AED{
             one{"VAE-dirham"}
             other{"VAE-dirham"}
         }
-        AFN{
```
macchiati, line 796: I was surprised that these disappeared, is that a cldr issue,erhaps

pedberg-icu: @macchiati It is because the currency plural names can inherit from the display name if they are the same (non-other -> other -> displayName). Dragan just fixed one bug in ICU4J about this.

DraganBesevic: The fix is in ICUCurrencyDataDisplayInfoProvider::getPluralName

markusicu (dismissed):
Thanks for splitting this PR into two commits for binary+rest!
The non-binary commit is still almost too much for reviewing on GitHub (in my browser). In future PRs maybe split even further? (By data "tree", by data vs. code, or something.)

Also, maybe we should allow these kinds of PRs to be merged with multiple commits, to avoid having to squash and needing re-approval?

I glanced over a bunch of the .txt/.cpp/.java changes and they look plausible.

It looks like the size of our data might get reduced this time by removing what is now deemed redundant data. That would be nice!

DraganBesevic:
> Oh, and note that there is a merge conflict in icudata.jar.

Oh yes, we just rebased and merged the resolved conflict

pedberg-icu:
> > Oh yes, we just rebased and merged the resolved conflict
> 
> Great!
> 
> You still have only two commits, but now they are name "part 2" and "part 3" :-(
> 
> I think it would be good to either rename one or both of the commits (and we turn on allowing multiple commits), or else squash them into one.

When we rebased to main, rebuild the data jars and pushed, something very strange happened. The first commit seems to have disappeared and all of the binary files seem to have reverted to their state prior to this PR. Will need to work on fixing this. But rebasing also brought in the Unicode 15.1 stuff and there may be strange interactions there. We may need to redo starting with the latest CLDR data (but keeping all of the ICU-side fixes so far) in which case it would be alpha1

pedberg-icu:
> When we rebased to main, rebuild the data jars and pushed, something very strange happened. The first commit seems to have disappeared and all of the binary files seem to have reverted to their state prior to this PR. Will need to work on fixing this. But rebasing also brought in the Unicode 15.1 stuff and there may be strange interactions there. We may need to redo starting with the latest CLDR data (but keeping all of the ICU-side fixes so far) in which case it would be alpha1

This is replaced by https://github.com/unicode-org/icu/pull/2559, so closing this.

PR #2557, opened 2023-08-19
ICU-22313 Various fixes for duration formatting:

richgillam:
- Changed the C interface so that calling unum_open(UNUM_DURATION) produces a formatter that uses a MeaureFormat under the covers instead of a RuleBasedNumberFornmat.
- Changed the C++ and Java interfaces to that the URBNF_DURATION ruleset is marked deprecated.
- Fixed a bug in RuleBasedNumberFormat in both Java and C++ that caused the existing duration-formatting rules to produce bogus results when used on a non-integral value.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22313

### DIFF
--- a/icu4c/source/i18n/nfsubs.cpp
+++ b/icu4c/source/i18n/nfsubs.cpp
@@ -103,7 +103,25 @@ public:
     }
 
     virtual double transformNumber(double number) const override {
-        if (getRuleSet()) {
+        bool doFloor = getRuleSet() != nullptr;
+        if (!doFloor) {
+            // This is a HACK that partially addresses ICU-22313.  The original code wanted us to do
+            // floor() on the result if we were passing it to another rule set, but not if we were passing
+            // it to a DecimalFormat.  But the DurationRules rule set has multiplier substitutions where
+            // we DO want to do the floor() operation.  What we REALLY want is to do floor() any time
+            // the owning rule also has a ModulusSubsitution, but we don't have access to that information
+            // here, so instead we're doing a floor() any time the DecimalFormat has maxFracDigits equal to
+            // 0.  This seems to work with our existing rule sets, but could be a problem in the future,
+            // but the "real" fix for DurationRules isn't worth doing, since we're deprecating DurationRules
+            // anyway.  This is enough to keep it from being egregiously wrong, without obvious side
+            // effects.     --rtg 8/16/23
+            const DecimalFormat* decimalFormat = getNumberFormat();
+            if (decimalFormat == nullptr || decimalFormat->getMaximumFractionDigits() == 0) {
+                doFloor = true;
+            }
+        }
+        
+        if (doFloor) {
             return uprv_floor(number / divisor);
         } else {
             return number / divisor;
--- a/icu4c/source/i18n/unum.cpp
+++ b/icu4c/source/i18n/unum.cpp
@@ -28,17 +28,19 @@
 #include "unicode/dcfmtsym.h"
 #include "unicode/curramt.h"
 #include "unicode/localpointer.h"
+#include "unicode/measfmt.h"
 #include "unicode/udisplaycontext.h"
 #include "uassert.h"
 #include "cpputils.h"
 #include "cstring.h"
+#include "putilimp.h"
 
 
 U_NAMESPACE_USE
 
 
 U_CAPI UNumberFormat* U_EXPORT2
-unum_open(  UNumberFormatStyle    style,  
+unum_open(  UNumberFormatStyle    style,
             const    char16_t*    pattern,
             int32_t            patternLength,
             const    char*     locale,
@@ -671,6 +673,7 @@ unum_getTextAttribute(const UNumberFormat*  fmt,
 
     const NumberFormat* nf = reinterpret_cast<const NumberFormat*>(fmt);
     const DecimalFormat* df = dynamic_cast<const DecimalFormat*>(nf);
+    const RuleBasedNumberFormat* rbnf = nullptr;    // cast is below for performance
     if (df != nullptr) {
         switch(tag) {
         case UNUM_POSITIVE_PREFIX:
@@ -701,8 +704,7 @@ unum_getTextAttribute(const UNumberFormat*  fmt,
             *status = U_UNSUPPORTED_ERROR;
             return -1;
         }
-    } else {
-        const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
+    } else  if ((rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf)) != nullptr) {
         U_ASSERT(rbnf != nullptr);
         if (tag == UNUM_DEFAULT_RULESET) {
             res = rbnf->getDefaultRuleSetName();
@@ -716,6 +718,9 @@ unum_getTextAttribute(const UNumberFormat*  fmt,
             *status = U_UNSUPPORTED_ERROR;
             return -1;
         }
+    } else {
+        *status = U_UNSUPPORTED_ERROR;
+        return -1;
     }
 
     return res.extract(result, resultLength, *status);
@@ -794,15 +799,16 @@ unum_toPattern(    const    UNumberFormat*          fmt,
 
     const NumberFormat* nf = reinterpret_cast<const NumberFormat*>(fmt);
     const DecimalFormat* df = dynamic_cast<const DecimalFormat*>(nf);
+    const RuleBasedNumberFormat* rbnf = nullptr;    // cast is below for performance
     if (df != nullptr) {
       if(isPatternLocalized)
         df->toLocalizedPattern(pat);
       else
         df->toPattern(pat);
+    } else if ((rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf)) != nullptr) {
+        pat = rbnf->getRules();
     } else {
-      const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
-      U_ASSERT(rbnf != nullptr);
-      pat = rbnf->getRules();
+        // leave `pat` empty
     }
     return pat.extract(result, resultLength, *status);
 }
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
@@ -560,8 +560,9 @@ public class RuleBasedNumberFormat extends NumberFormat {
 
     /**
      * Selector code that tells the constructor to create a duration formatter
-     * @stable ICU 2.0
+     * @deprecated ICU 74 Use MeasureFormat instead.
      */
+    @Deprecated
     public static final int DURATION = 3;
 
     /**
--- a/icu4c/source/i18n/unicode/rbnf.h
+++ b/icu4c/source/i18n/unicode/rbnf.h
@@ -64,18 +64,20 @@ enum URBNFRuleSetTag {
      * @stable ICU 2.2
      */
     URBNF_ORDINAL,
+#ifndef U_HIDE_DEPRECATED_API
     /**
      * Requests predefined ruleset for formatting a value as a duration in hours, minutes, and seconds.
-     * @stable ICU 2.2
+     * @deprecated ICU 74 Use MeasureFormat instead.
      */
     URBNF_DURATION,
+#endif // U_HIDE_DERECATED_API
     /**
      * Requests predefined ruleset for various non-place-value numbering systems.
      * WARNING: The same resource contains rule sets for a variety of different numbering systems.
      * You need to call setDefaultRuleSet() on the formatter to choose the actual numbering system.
      * @stable ICU 2.2
      */
-    URBNF_NUMBERING_SYSTEM,
+    URBNF_NUMBERING_SYSTEM = 3,
 #ifndef U_HIDE_DEPRECATED_API
     /**
      * One more than the highest normal URBNFRuleSetTag value.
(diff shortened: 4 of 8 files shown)


richgillam:
I'm not sure this is really the way we want to go, but it seems to work (more or less), so I'm putting it up for further discussion.  My own thoughts:
- The actual fix for `RuleBasedNumberFormat` might be sufficient to address the issue (without messing with the rules themselves), but it's pretty hacky, and I'm not sure if I'm increasing risk that something more important will break by doing what I'm doing.
- I got `unum_open()` with `UNUM_DURATION` to use `MeasureFormat` under the covers, but the code is kind of complicated and ugly, and I wonder if it's really worth it.
- More importantly, `MeasureFormat` doesn't support parsing, so if (God forbid) somebody uses `unum_open(UNUM_DURATION)` for parsing, they'll fail, and we don't even seem to have a mechanism for reporting the failure.

All of that said, I have an open ticket to create a real `DurationFormatter` that could be used inside a `NumberFormatter`, and that would be a cleaner solution to the problem.  I just don't think I can get that in for ICU 74.

icu4c/source/i18n/unum.cpp
```diff
@@ -28,17 +28,83 @@
...
 #include "cpputils.h"
 #include "cstring.h"
+#include "putilimp.h"
 
 
 U_NAMESPACE_USE
 
 
+// to support UNUM_DURATION
+class MeasureFormatWrapper : public NumberFormat {
```
sffc, line 43: Suggestion: `duration` should be in the name somewhere, either `MeasureFormatDurationWrapper` or `MeasureFormatWrapper::formatDuration` (probably the latter)

icu4c/source/i18n/unum.cpp
```diff
@@ -28,17 +28,83 @@
...
+    }
+    
+    NumberFormat *clone() const override {
+        UErrorCode ignoreErr = U_ZERO_ERROR;
+        return new MeasureFormatWrapper(locale, ignoreErr);
+    }
+    
+    UnicodeString &format(double number, UnicodeString &appendTo, FieldPosition &pos) const override {
+        UErrorCode err = U_ZERO_ERROR;
+        if (number < 60) {
```
sffc, line 70: Question: is it documented somewhere that the input is expected to be in seconds, and not minutes, milliseconds, etc?

sffc (commented):
Seems like a reasonable way to use a scalpel to get the behavior that you want. I like that most of these changes live in unum.cpp which is already kind-of a mess.

icu4c/source/i18n/nfsubs.cpp
```diff
@@ -103,7 +103,25 @@ class MultiplierSubstitution : public NFSubstitution {
...
+            // but the "real" fix for DurationRules isn't worth doing, since we're deprecating DurationRules
+            // anyway.  This is enough to keep it from being egregiously wrong, without obvious side
+            // effects.     --rtg 8/16/23
+            const DecimalFormat* decimalFormat = getNumberFormat();
+            if (decimalFormat == nullptr || decimalFormat->getMaximumFractionDigits() == 0) {
+                doFloor = true;
+            }
+        }
+        
+        if (doFloor) {
```
grhoten, line 124: For discussion, here are the original rules for English.
```
%with-words:
0: 0 seconds; 1 second; =0= seconds;
60/60: <%%min<[, >>];
3600/60: <%%hr<[, >>>];
%%min:
0: 0 minutes; 1 minute; =0= minutes;
%%hr:
0: 0 hours; 1 hour; =0= hours;
%in-numerals:
0: =0= sec.;
60: =%%min-sec=;
3600: =%%hr-min-sec=;
%%min-sec:
0: :=00=;
60/60: <0<>>;
%%hr-min-sec:
0: :=00=;
60/60: <00<>>;
3600/60: <#,##0<:>>>;
%duration:
0: =%in-numerals=;
%%lenient-parse:
& ':' = '.' = ' ' = '-';
```
As I recall, the reason why this issue happens is because all of the other RBNF rules won't trigger this issue. Here's why.

1. Those rules have an `x.x` rule. They will either do the math to use the words, or they will format the invalid fraction rule with `=#,##0.#=` in the scenario like ordinals. Any results from fractional ordinals are considered undefined because it's bad input. So we don't test it for accuracy beyond showing that something is formatted.
2. When a number is formatted, it's typically using an integer rule like the following `=#,##0=` for really large values instead of a floating point. Usually this is hit well beyond the precision of the mantissa of a little over 9 quadrillion. For the vetted RBNF languages, this rule is typically hit around 1 quintillion - 1. A maximum signed 64-bit integer is around 9.22 quintillion.

There hasn't been a tested scenario where you're formatting a floating point number instead of words, **and** applying a transformation on it without an x.x rule. So we haven't used `=00=` in a non-integer situation before in the other rules.

So the reported rounding is bad, but none of the **supported** rules really hit it. It's yet another reason why RBNF isn't suited well for durations.

grhoten (approved):

icu4c/source/i18n/unum.cpp
grhoten: As I recall, RBNF durations are poorly documented. The existing rules always assumed that it was in seconds, and it was only translated into English. When you scale up to beyond English, RBNF durations are too painful to attempt translation.  You have to copy a lot of math around.  For backwards compatibility reasons, I recommend not breaking the English assumptions about how the rules worked.

For reference, it's actually better to use a list format, and to specify what units are in the list.  RBNF isn't well suited to change the base unit. It's a similar translation pattern with miles, feet and inches.

grhoten:
FYI it's possible that these changes will break the scenario where someone uses unum_setTextAttribute with `%with-words` in English. I think that's the only thing that these changes are breaking by switching to a duration format.

richgillam:
> FYI it's possible that these changes will break the scenario where someone uses unum_setTextAttribute with `%with-words` in English. I think that's the only thing that these changes are breaking by switching to a duration format.

I didn't realize that.  Do you think that's a strong enough argument not to make the other changes here and just deprecate everything and leave it all broken?  Or should we not worry about that?

icu4c/source/i18n/unum.cpp
```diff
@@ -716,6 +782,9 @@ unum_getTextAttribute(const UNumberFormat*  fmt,
             *status = U_UNSUPPORTED_ERROR;
             return -1;
         }
+    } else {
+        *status = U_UNSUPPORTED_ERROR;
+        return -1;
     }
```
grhoten, line 788: This is a good robustness change. Please keep this change.

grhoten:
> > FYI it's possible that these changes will break the scenario where someone uses unum_setTextAttribute with `%with-words` in English. I think that's the only thing that these changes are breaking by switching to a duration format.
> 
> I didn't realize that. Do you think that's a strong enough argument not to make the other changes here and just deprecate everything and leave it all broken? Or should we not worry about that?

The transformNumber change seems reasonable. I definitely want the deprecation changes in. The class cast checking is a good robustness improvement.

It's hard to say if breaking something to improve something that was already broken is good or bad. It's still going to be an incomplete fix. As was mentioned above, what if you want to mix units together differently. If there is already C API to do it the right way, then perhaps it's better to just leave the current RBNF duration implementation broken. Though if that's the case, then I recommend improving the documentation to say what the better alternative is.

grhoten:
After consulting with other team members, it's probably better to leave the current broken duration implementation there with this number fix, but the documentation of the deprecated API should suggest users to go use the MeasureFormat API instead.

richgillam:
> After consulting with other team members, it's probably better to leave the current broken duration implementation there with this number fix, but the documentation of the deprecated API should suggest users to go use the MeasureFormat API instead.

Just so we're clear, you're saying this whole PR should go in, but with improved documentation to steer people away from using it?  Or you're saying that I should back out the `MultiplierSubstitution` change and keep the rest of it (and add the documentation, etc.)?

grhoten:
I'm saying the following:

- Keep the deprecation changes.
- Add more documentation about the deprecation, and encourage using the MeasureFormat API.
- Keep the MultiplierSubstitution change. It may be helpful later.
- Keep the type safety changes in unum_.
- Back out the `MeasureFormatWrapper` switch and keep calling the broken duration type in RBNF.

richgillam:
I've been thinking about this off an on since the last time I posted here, and I'd like to suggest a resolution and get people's opinion:

I propose that I take George's advice: keep the MultiplierSubstitution fix, deprecate all of the RBNF-duration stuff, and keep the type-safety change in `unum_open()`, but back out the wrapper that let me use `MeasureFormat` to back the `UNUM_DURATION` constant and keep using RBNF for that.  (And add a little more commentary to the deprecation warnings, steering people to use `MeasureFormat`.)

In addition, I propose to file a new ticket that would be based on https://unicode-org.atlassian.net/browse/ICU-21634, asking for `UNUM_DURATION` to be replaced by something new that uses the new `DurationFormatter` as its implementation, so we're living with RBNF here for now, but with a plan for fixing that.

As for the long-term fix, I think we've got a few options:

1. When we have a new `DurationFormatter`, just give it a C wrapper API and deprecate `UNUM_DURATION`.
2. When we have a new `DurationFormatter`, plumb things through so that a subset of its functionality is available through NumberFormatter (something like `"measure-unit/duration-minutes-and-seconds unit-width-numeric"`), which would make the functionality available in C via the existing `unum_openForSkeletonAndLocale()` API, and deprecate `UNUM_DURATION`.
3. Same as 2, but plumb things through so that `UNUM_DURATION` uses `NumberFormatter` under the covers.

My recommendation would be that we do option 2, or both options 1 and 2.  3, like the `MeasureFormat` thing I've got in place now, introduces a backward-compatibility issue for anyone who's using `unum_setTextAttribute()` to pick a rule set.

@grhoten , @sffc , @markusicu , does this sound good to you?

macchiati:
Sounds good to me. As for the options, I have no strong preference, beyond
deprecating UNUM_DURATION relatively soon (or replacing its guts)

sffc:
For the list of options, please add for completeness:

4. Implement DurationFormat in ICU4X and use that C API

richgillam:
I filed a new ticket for the unfinished work: https://unicode-org.atlassian.net/browse/ICU-22487

richgillam:
> For the list of options, please add for completeness:
> 
> 4. Implement DurationFormat in ICU4X and use that C API

Consider it added.  But I have a bunch of questions and concerns about how this would work.  Expect a follow-up email in the coming days.

grhoten (approved):

icu4c/source/i18n/unum.cpp
```diff
@@ -110,8 +112,8 @@ unum_open(  UNumberFormatStyle    style,
         break;
 
     case UNUM_DURATION:
-        retVal = new RuleBasedNumberFormat(URBNF_DURATION, Locale(locale), *status);
-        break;
+         retVal = new RuleBasedNumberFormat(URBNF_DURATION, Locale(locale), *status);
+         break;
```
markusicu, line 116:
```suggestion
        retVal = new RuleBasedNumberFormat(URBNF_DURATION, Locale(locale), *status);
        break;
```
back to indent 8 not 9

icu4c/source/i18n/unum.cpp
```diff
@@ -701,8 +704,7 @@ unum_getTextAttribute(const UNumberFormat*  fmt,
             *status = U_UNSUPPORTED_ERROR;
             return -1;
         }
-    } else {
-        const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
+    } else  if (rbnf != nullptr) {
```
markusicu, line 707: nit: consider doing
```suggestion
    } else  if ((rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf)) != nullptr) {
```
so that we need not spend the small amount of dynamic_cast time in the common case of it being a simple DecimalFormat

icu4c/source/i18n/unum.cpp
```diff
@@ -794,15 +799,16 @@ unum_toPattern(    const    UNumberFormat*          fmt,
...
+    const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
     if (df != nullptr) {
       if(isPatternLocalized)
         df->toLocalizedPattern(pat);
       else
         df->toPattern(pat);
-    } else {
-      const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
-      U_ASSERT(rbnf != nullptr);
+    } else if (rbnf != nullptr) {
```
markusicu, line 808: nit / ditto

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
```diff
@@ -560,7 +560,7 @@ public class RuleBasedNumberFormat extends NumberFormat {
 
     /**
      * Selector code that tells the constructor to create a duration formatter
-     * @stable ICU 2.0
+     * @deprecated ICU 74 Use MeasureFormat instead.
      */
     public static final int DURATION = 3;
```
markusicu, line 565:
```suggestion
     * @deprecated ICU 74 Use MeasureFormat instead.
     */
    @Deprecated
    public static final int DURATION = 3;
```

markusicu (commented):
- mostly lgtm
- mostly optional suggestions, except please do add the Java `@Deprecated` *annotation*.
- You probably anyway need to rebase so that this goes on top of PR #2601 which disables the mysteriously failing Java test functions.

icu4c/source/i18n/unum.cpp
richgillam: Not crazy about this, but I'll do it...

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
richgillam: Sorry about that.  Done.

markusicu (approved):
tnx lgtm

PR #2558, opened 2023-08-19
ICU-20777 Remove the generation of likelySubtags.res

FrankYFTang:
The intend of this PR is to remove the generation and checked in files of icu/icu4c/source/data/misc/likelySubtags.txt and the generated likelySubtags.res because now all the code read from icu/icu4c/source/data/misc/langInfo.txt instead now.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-20777


FrankYFTang:
@pedberg-icu I am not familar with CLDR-to-ICU tool and somehow I have problem to figure out how to run and test these changes.

richgillam (commented):
I'm a little confused.  On the ICU side, there's `likelySubtags.txt` and `likelySubtags.res`, both of which you're deleting, but there should also be some sort of tool generating `likelySubtags.res` from `likelySubtags.txt`, and entries for both in both the makefile and the Python tool that generates the makefile, and I'm not seeing any of that in this PR.  Or am I wrong, and the LDML-to-ICU tool is actually also creating the binary file?

I'm guessing there's also more to the LDML-to-ICU side of things than what you're deleting here, but Peter's going to have to speak to that, as he's WAY more familiar with it than I am.

FrankYFTang:
About .txt => .res
> some sort of tool generating likelySubtags.res from likelySubtags.txt ... and I'm not seeing any of that in this PR
that is part of generate_misc in https://github.com/unicode-org/icu/blob/main/icu4c/source/data/BUILDRULES.py#L397C5-L397C18

Notice the code
```
    input_files = [InFile(filename) for filename in io.glob("misc/*.txt")]
```
so... by removing misc/likelySubtags.txt in the directory, it will be removed from input_files

>I'm guessing there's also more to the LDML-to-ICU side of things
About LDML-to-ICU
so my understanding is this. 
A. For LDML-to-ICU process, we have a tool to take one input and generate two output.
The input is 
https://github.com/unicode-org/cldr/blob/main/common/supplemental/likelySubtags.xml

and the two output are

https://github.com/unicode-org/icu/blob/main/icu4c/source/data/misc/likelySubtags.txt
and
https://github.com/unicode-org/icu/blob/main/icu4c/source/data/misc/langInfo.txt

the information stored in langInfo.txt contains the same information stored in likelySubtags.txt (and maybe more) so we really do not need to have bot of them. 

Note: Mark also change another part of the CLDR to ICU conversion to generate a test file, which also called likelySubtags.txt but that is a test file not a resource file, that one is located in 
https://github.com/unicode-org/cldr/blob/main/common/testData/localeIdentifiers/likelySubtags.txt#L4

B. In the building of ICU data, the ICU tool compile the misc/likelySubtags.txt and misc/langInfo.txt into likelySubtags.res and langInfo.res and these res files are also checked into icu4j 
as https://github.com/unicode-org/icu/blob/main/icu4j/maven-build/maven-icu4j-datafiles/src/main/resources/com/ibm/icu/impl/data/icudt74b/likelySubtags.res and jar file

C. In https://github.com/unicode-org/icu/pull/2561 I remove all the ICU4C and ICU4J code which read the likelySubtags.res and use the equivlant information by reading only langInfo.res but I didn't remove any files

D. In this PR, what I like to do is
1. change the CLDR to ICU tool that read likelySubtags.xml to only generate icu4c/source/data/misc/langInfo.txt but not generate  icu4c/source/data/misc/likelySubtags.txt
(I think the change https://github.com/unicode-org/icu/pull/2558/files#diff-bb57ed3c0ce5d90f10e4c03649e414feb36c13c9fed1c09ded359694eb82aba5 will do so (but I am not 100% sure)
2. remove the  icu4c/source/data/misc/likelySubtags.txt
3. remove the icu4j/maven-build/maven-icu4j-datafiles/src/main/resources/com/ibm/icu/impl/data/icudt74b/likelySubtags.res
4. change the generated jar file to not include likelySubtags.res

richgillam:
>>some sort of tool generating likelySubtags.res from likelySubtags.txt ... and I'm not seeing any of that in this PR

>that is part of generate_misc in https://github.com/unicode-org/icu/blob/main/icu4c/source/data/BUILDRULES.py#L397C5-L397C18

Ah, so it is.  And you're right-- it's checking for that file implicitly, so if it's not there, it doesn't do any work, and the tool being run is just `genrb`, so there's no special tool that needs to be removed either.  Thank you!

richgillam (approved):

pedberg-icu:
> @pedberg-icu I am not familar with CLDR-to-ICU tool and somehow I have problem to figure out how to run and test these changes.

See https://unicode-org.github.io/icu/processes/cldr-icu.html

pedberg-icu:
This looks good to me but... data/misc/likelySubtags.txt got updated as part of the CLDR release-44-alpha1 integration, so even though this PR deletes it, github thinks there is a conflict (sigh). Once you rebase I can approve this.

richgillam (approved):

PR #2561, opened 2023-08-22
ICU-22466 Fix incorrect memory read while the locale is bogus

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22466

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -483,7 +483,16 @@ _uloc_addLikelySubtags(const char* localeID,
         if(U_FAILURE(*err)) {
             goto error;
         }
-        icu::LSR lsr = likelySubtags->makeMaximizedLsrFrom(icu::Locale::createFromName(localeID), true, *err);
+        // We need to keep l on the stack because lsr may point into internal
+        // memory of l.
+        icu::Locale l = icu::Locale::createFromName(localeID);
+        if (l.isBogus()) {
+            goto error;
+        }
+        icu::LSR lsr = likelySubtags->makeMaximizedLsrFrom(l, true, *err);
+        if(U_FAILURE(*err)) {
+            goto error;
+        }
         const char* language = lsr.language;
         if (uprv_strcmp(language, "und") == 0) {
             language = "";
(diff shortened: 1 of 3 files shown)


roubert (approved):
LGTM

icu4c/source/common/loclikely.cpp
```diff
@@ -483,7 +483,16 @@ _uloc_addLikelySubtags(const char* localeID,
         if(U_FAILURE(*err)) {
             goto error;
         }
-        icu::LSR lsr = likelySubtags->makeMaximizedLsrFrom(icu::Locale::createFromName(localeID), true, *err);
+        // We need to keep l on the stack because lsr may point into internal
+        // memory of l.
```
richgillam, line 487: Is declaring `l` as a local variable sufficient to do that?  I don't think you're guaranteed the object will persist until the end of the function if you don't reference it again-- I think the compiler is free to destruct `l` earlier, although I don't know if any of them actually do that.

richgillam (approved):

icu4c/source/common/loclikely.cpp
sffc: Yes, because `l` will exist until the `}` of the enclosing block, whereas previously it existed only until the `;` on the line it was used as a transient variable.

FrankYFTang: the lsr.language, lsr.region and lsr.script will reference to the memory returned by the getLanguage(), getScript()  and getCountry() of the Locale passed into makeMaximizedLsrFrom. So if we do not keep it as a local on the stack, those value became dangling pointers after the call to makeMaximizedLsrFrom and while we use it in the uprv_strcmp (line 488/497), or uprv_strlen(line 493-506/502-506) it became illegal read. But if we keep it on the stack, they are not illegal read until line 506/515 but we no longer need them at that point.

richgillam: Yeah, you guys are right.  I know that in some languages, the compiler is free to deallocate objects _before_ the closing brace if they know they're not used, but if that were true in C++, all kinds of common C++ idioms would break.  So never mind.  :-)

PR #2566, opened 2023-08-25
ICU-22471 Bring ICU4J PersonNameFormatter up to date

richgillam:
- Bring in the latest version of the CLDR PersonNameFormatter test data.
- Fix PersonNameConsistencyTest to recognize the new “givenFirst” and “surnameFirst” keywords used in the test data files.
- Fix tokenization code in InitialModifier to use a BreakIterator.  Add unit test.
- Add support for the “retain” modifier.  Add unit test.
- Remove the “log known failure” logic for all of the PersonNameConsistencyTest data files that pass now.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22471

### DIFF
--- a/icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameConsistencyTest.java
+++ b/icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameConsistencyTest.java
@@ -14,6 +14,7 @@ import com.ibm.icu.dev.test.TestUtil;
 import com.ibm.icu.text.PersonName;
 import com.ibm.icu.text.PersonNameFormatter;
 import com.ibm.icu.text.SimplePersonName;
+import com.ibm.icu.dev.test.rbbi.RBBITstUtils;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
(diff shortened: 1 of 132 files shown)


icu4j/main/tests/core/src/com/ibm/icu/dev/test/format/PersonNameConsistencyTest.java
```diff
@@ -30,16 +30,10 @@ public class PersonNameConsistencyTest extends TestFmwk {
...
 
     static private Collection<String> FILENAMES_TO_SKIP_FOR_17028 =
-        Arrays.asList("az.txt", "be.txt", "bs.txt", "fi.txt", "gu.txt", "hr.txt", "is.txt",
-                      "km.txt", "ky.txt", "lo.txt", "mk.txt", "ml.txt", "mn.txt", "mr.txt",
-                      "pa.txt", "pt.txt", "pt_PT.txt", "sw.txt", "tr.txt", "zu.txt",
-                      "yue_Hans.txt", "fa.txt","ka.txt", "zh_Hant_HK.txt", "zh_Hant.txt",
-                      "bn.txt", "zh.txt", "nl.txt", "to.txt", "uk.txt", "my.txt",
-                      "bg.txt", "tk.txt", "ps.txt", "ko.txt", "ms.txt", "ne.txt",
-                      "gd.txt", "lv.txt" );
+        Arrays.asList("yue_Hans.txt", "to.txt", "gl.txt", "ie.txt" );
```
FrankYFTang, line 36: you still have 19 errors in my.txt
per https://github.com/unicode-org/icu/actions/runs/5980715785/job/16227389539?pr=2566#step:7:222

FrankYFTang: It seems you no longer faile on ie.txt

richgillam: That's very weird.  There are many CI checks that build ICU4J and run its unit tests, and they're all passing except for this one.  I also tried it again locally-- with both Ant and Maven, in case that made a difference-- and it worked fine.  I can't think of a reason why that one CI check would have failures like that unless there's some kind of inconsistency in the ICU locale data it's using.  I'm re-running the test in case it's a transient problem, but if it fails again, I really don't know what to try next.

richgillam: Oh, okay, now I understand what the problem is (with Myanmar, anyway).

The discrepancies are coming in the initial-generation logic.  For the first failure, on line 123, for example, the expected result is expecting the surname field to be broken up into four separate "words", each of which is then converted to an initial.  Several of those "words" are not present in the actual result.

This build uses the LSTM break iterator for Myanmar instead of the regular dictionary-based break iterator, and they're segmenting the surname field of this name (and many other names in the test) differently, leading to the test failures.

I don't think have (or want) a whole different set of expected-results files for builds that use the LSTM break iterator.  I could theoretically fix it so that the PersonNameFormatter always uses the old ICU break iterator, but I don't think there's a way to do that, and I also don't think it's a good idea.

If there's a reasonable way to tell at run time when we're using the LSTM break iterator, I could use that to determine when to skip this test.  And I think that's probably the right answer here.  Otherwise, I don't think there's much I can do here except to skip this locale.

richgillam: >It seems you no longer fail on ie.txt

The test still fails; I just wasn't _running_ it.  Looks like I screwed up and forgot to include `ie.txt` in `catalog.txt`.  I've fixed that now, and it'll go in whenever I come up with a way to deal with the `my.txt` problem.

richgillam: And there _is_ a way to tell at run time whether the LSTM stuff is being used.  Posting a fix now.  I hope this works...

richgillam:
@macchiati @pedberg Can I get your eyes on this PR sometime soon?

pedberg-icu (approved):
Looks good! Thanks.

FrankYFTang (approved):
LGTM

PR #2567, opened 2023-08-29
ICU-22475 Fix double free in Locale under OOM

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22475

### DIFF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -2092,12 +2092,13 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             int32_t oldTagLength = tagLen;
             if (tagLen < newTagLength) {
                 uprv_free(tagBuf);
-                tagBuf = (char*)uprv_malloc(newTagLength + 1);
+                // Change t->buf after the free and before return to avoid the second double free in
+                // the destructor of t when t is out of scope.
+                t->buf = tagBuf = (char*)uprv_malloc(newTagLength + 1);
                 if (tagBuf == nullptr) {
                     *status = U_MEMORY_ALLOCATION_ERROR;
                     return nullptr;
                 }
-                t->buf = tagBuf;
                 tagLen = newTagLength;
             }
             parsedLenDelta = checkLegacyLen - replacementLen;
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -2508,6 +2508,10 @@ static void TestCanonicalization21749StackUseAfterScope(void)
                 input, u_errorName(status));
         return;
     }
+
+    // ICU-22475 test that we don't free an internal buffer twice.
+    status = U_ZERO_ERROR;
+    uloc_canonicalize("ti-defaultgR-lS-z-UK-0P", buffer, UPRV_LENGTHOF(buffer), &status);
 }
 
 static void TestDisplayKeywords(void)


icu4c/source/test/cintltst/cloctst.c
```diff
@@ -2508,6 +2508,9 @@ static void TestCanonicalization21749StackUseAfterScope(void)
                 input, u_errorName(status));
         return;
     }
+
+    status = U_ZERO_ERROR;
+    uloc_canonicalize("ti-defaultgR-lS-z-UK-0P", buffer, 256, &status);
```
roubert, line 2513: Please use `sizeof buffer` instead of hardcoding 256.

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2093,11 +2093,13 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             if (tagLen < newTagLength) {
                 uprv_free(tagBuf);
                 tagBuf = (char*)uprv_malloc(newTagLength + 1);
+                t->buf = tagBuf; // Change t->buf after the free and before
+                                 // return to avoid the second double free in
+                                 // the destructor.
```
roubert, line 2098: Do you by _the destructor_ mean `ultag_close()`? If so, I think it'd be clearer if you wrote that instead. (If not, what do you mean then?)

roubert (changes requested):

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2093,11 +2093,13 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             if (tagLen < newTagLength) {
                 uprv_free(tagBuf);
```
markusicu, line 2094: It seems like this here is really the problematic line: Earlier on line 2065 `t` adopted tagBuf but t->buf is not cleared after this uprv_free(), so when the new tagBuf allocation fails, t->buf still has the pointer to the freed old buffer. The most natural fix seems to be to simply change the malloc line to start with `t->buf = tagBuf = [malloc...]` and not have a separate pointer-copy line. No comment needed.

**However**, I think a much more robust fix would be to change tagBuf from a raw pointer to a CharString, and to cloneData() its contents into t->buf only just before returning a successful `t`. That would also simplify appending string parts and NUL termination handling.

Also, I haven't looked at what's inside a ULanguageTag. An even bigger and potentially better fix might be to turn it into an internal C++ class LanguageTag that has a CharString buf field, and casting it to an opaque, C-style ULanguageTag at the C API boundaries. *This* would probably be fodder for a separate ticket.

FrankYFTang: I agree w/ you changing tagBuf to CharString could be a more robust fix but since the logic here w/ tagBuf is very complicate in this function, a rewrite for such fix have more protential to introduce even more bugs than a simple local fix. I think we should focus on security fix on minimum risk to create more security bugs than focus on the local issue at hand. What you suggest is basically a rewrite that is a good task for a different reason/ticket to reduce the complexity and improve the robustness our code but it should not be a blood stopping fix for a security ticket.

FrankYFTang: File ticket https://unicode-org.atlassian.net/browse/ICU-22476 for the CharString rewrite later so we will not forget about it.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -2508,6 +2508,9 @@ static void TestCanonicalization21749StackUseAfterScope(void)
                 input, u_errorName(status));
         return;
     }
+
+    status = U_ZERO_ERROR;
+    uloc_canonicalize("ti-defaultgR-lS-z-UK-0P", buffer, UPRV_LENGTHOF(buffer), &status);
```
markusicu, line 2513:
```suggestion

    // ICU-22475 test that we don't free an internal buffer twice.
    status = U_ZERO_ERROR;
    uloc_canonicalize("ti-defaultgR-lS-z-UK-0P", buffer, UPRV_LENGTHOF(buffer), &status);
```
... to help explain why this is here and why we don't check for the results of the function call.

markusicu (commented):
lgtm tnx, please squash again

waiting to see if Fredrik is happy

icu4c/source/common/uloc_tag.cpp
```diff
@@ -2092,12 +2092,13 @@ ultag_parse(const char* tag, int32_t tagLen, int32_t* parsedLen, UErrorCode* sta
             int32_t oldTagLength = tagLen;
             if (tagLen < newTagLength) {
                 uprv_free(tagBuf);
-                tagBuf = (char*)uprv_malloc(newTagLength + 1);
+                // Change t->buf after the free and before return to avoid the second double free in
+                // the destructor of t when t is out of scope.
```
roubert, line 2096: I really think it'd be clearer to write `ultag_close()` instead of _the destructor_, if that's what you mean. It took me unnecessarily long time to figure out that this probably was what you meant.

FrankYFTang: the function signature of ultag_close() is
```
void ultag_close(ULanguageTag* langtag);
```
and the only code which ever call ultag_close is by the destructor of LocalULanguageTagPointer
There are no code in ultag_parse call ultag_close directly so if you said it is free by ultag_close it is very difficult for the reader to understand WHERE is such call connect to THIS (ultag_parse) function. My updated comments said "second double free in the destructor of t " will bring the reader to look into the type of t and therefore look into
the desctructor of LocalULanguageTagPointer and then see it call the close function which is ultag_close. If my comments only mention ultag_close it is harder for the reader to understand why that ultag_close will be called.

markusicu (approved):
lgtm

PR #2572, opened 2023-08-30
ICU-22455 Implemented algorithm in CLDR-16981 to preserve regional un…

richgillam:
…it overrides when they don't conflict

with the ms subtag.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22455

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/units/UnitPreferences.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/units/UnitPreferences.java
@@ -3,15 +3,12 @@
 package com.ibm.icu.impl.units;
 
 import java.math.BigDecimal;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
 
 import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.impl.ICUResourceBundle;
 import com.ibm.icu.impl.UResource;
+import com.ibm.icu.util.MeasureUnit;
 import com.ibm.icu.util.ULocale;
 import com.ibm.icu.util.UResourceBundle;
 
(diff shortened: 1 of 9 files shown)


younies (approved):

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/UnitPreferences.java
```diff
@@ -92,36 +89,39 @@ public UnitPreference[] getPreferencesFor(String category, String usage, ULocale
...
+                for (UnitPreference unitPref : result) {
+                    MeasureUnitImpl measureUnit = MeasureUnitImpl.forIdentifier(unitPref.getUnit());
+                    List<SingleUnitImpl> singleUnits = new ArrayList<>(measureUnit.getSingleUnits());
+                    for (SingleUnitImpl singleUnit : singleUnits) {
+                        String systems = rates.extractSystems(singleUnit);
+                        if (!systems.contains("metric_adjacent")) {
+                            if (!systems.contains(localeSystem)) {
+                                unitsMatchSystem = false;
+                            }
+                        }
```
sffc, line 116: Shouldn't we allow `metric_adjacent` only when the `localeSystem` is `metric`? Because if we have a unit preference with `metric_adjacent` units but `imperial` was requested, then we should override with the imperial units, right?

richgillam: Well, the big problem I was having was with time units, which are all labeled `metric-adjacent`.  From looking at the other units, I didn't see any where `metric-adjacent` was being used in contradistinction to `ussystem` or `uksystem`, so this approach seemed safe.  An alternative approach might be to only look at the _first_ unit of a compound unit instead of all of them, or to explicitly avoid checking time units.  I didn't consider the first option, fearing I'd missing things that way, and the second option seemed too complicated (I couldn't see an easy way to tell whether something was a time unit).

sffc: Okay, fair enough. I think this is code we can continue to tweak as more cases come up.

macchiati:
My fault for not providing a derivation for compound units. I created a
subtask for documenting that in
https://unicode-org.atlassian.net/browse/CLDR-17072

Please let me know your thoughts on that. Note that in practice, we don't
have a lot of compound units in the preferences right now, so we could get
by in ICU in this release with a simpler mechanism to get the same result.

richgillam:
>Please let me know your thoughts on that.

`mile-per-second` will give you an empty intersection with the logic you have in that ticket right now.  (In fairness, my logic will barf on that too.)

>My fault for not providing a derivation for compound units.

Are you okay with going forward with what I have right now for ICU 74 and then iterating on it in future releases?

macchiati:
I think it should probably work for the range of units we have in
preferences.

I think some of the unit's systems need to be cleaned up in order to
make it work well for compound units for preferences, and you definitely
hit one of them!

I made a quick dump of the systems for units at
https://docs.google.com/spreadsheets/d/1wmlNKyQUyS-7oKGRoYown0LASozOvtoPx4qj9AvhBIg/edit?usp=sharing,
where I organize by units that share the same set of systems: it's
illustrative.

In particular, the following should all have should all have {ussystem,
uksystem, jpsystem, astronomical} added to them.

second, minute, hour, day, week, month, year, decade, century; arc-second,
arc-minute, degree


Or maybe just have some special system for "any", meaning compatible with
any system.

richgillam:
> I made a quick dump of the systems for units at https://docs.google.com/spreadsheets/d/1wmlNKyQUyS-7oKGRoYown0LASozOvtoPx4qj9AvhBIg/edit?usp=sharing, where I organize by units that share the same set of systems: it's illustrative.

It does look a little haphazard...

>In particular, the following should all have should all have {ussystem, uksystem, jpsystem, astronomical} added to them.
>
>second, minute, hour, day, week, month, year, decade, century; arc-second, arc-minute, degree

I like the idea of adding "any" for units like these, although it'll require a code change on the ICU side.

macchiati:
I like the 'any' option too, but to avoid ICU changes for this release, am making a subtask to just add the extra systems.

richgillam:
> I like the 'any' option too, but to avoid ICU changes for this release, am making a subtask to just add the extra systems.

That sounds fine to me.  You might want to go ahead and file a ticket to clean up the ICU code after the CLDR data has been updated, but I'm fine with waiting to do that until somebody implements the "any" keyword on the CLDR side, since I think the ICU code will be okay until then (it might have some unnecessary logic in it with your next change, but I don't think any of that will hurt anything).

PR #2574, opened 2023-08-31
ICU-22479 Improve collator_compare_fuzzer

FrankYFTang:
1. Add seed corpus based on exemplarCharacters in data in cldr/exemplars/main/*.xml
2. Use random Locale and strength to repalce US/TERTIARY

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22479


icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
```diff
@@ -10,23 +11,36 @@
...
+    icu::Collator::IDENTICAL
+};
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   UErrorCode status = U_ZERO_ERROR;
 
   if (size < 2)
     return 0;
 
+  uint16_t rnd = *(reinterpret_cast<const uint16_t *>(data));
```
markusicu, line 28: This is likely an unaligned read. Please use memcpy to copy the first two bytes into rnd.

icu4c/source/test/fuzzer/collator_compare_fuzzer_seed_corpus.txt
markusicu, line 1: The extension is .txt but GitHub thinks it's binary.
If it's ASCII-compatible text (e.g., UTF-8) then please fix the file properties.
If it's binary, then please use an extenion other than .txt.
If it's UTF-16, then .txt should be ok. I assume that GitHub is just not set up for UTF-16.

markusicu, line 1: The code changes don't read this new file. Did you mean to add it in a different PR?

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
markusicu, line 28: You are now using the first two data bytes for the locale, the strength, and the first char16_t of the compbuff1 string.

Wouldn't it improve the fuzzer if these used separate parts of the data? For example,
- locale = bytes 0..1
- strength = byte 2
- compbuff1 = first half of remaining bytes
- compbuff2 = second half of remaining bytes

It seems easy to do so: On line 25 require at least 5 bytes, and when you read the locale value and the strength, increase the data pointer and decrease the size.

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
```diff
@@ -10,23 +11,36 @@
...
+
   std::unique_ptr<char16_t> compbuff1(new char16_t[size/4]);
   std::memcpy(compbuff1.get(), data, (size/4)*2);
   data = data + size/2;
```
markusicu, line 32: Managing data and size would be simpler if you used an icu::StringPiece. Initialize it early from the inputs. You update both pointer and length with remove_prefix(n) and then get the new values via data() and length(). (In C++17 we could use a std::string_view.)

icu4c/source/test/fuzzer/collator_compare_fuzzer_seed_corpus.txt
FrankYFTang: See https://unicode-org.github.io/icu/userguide/dev/fuzzer_targets.html#fuzzer-seed-corpus
it must be `<fuzzer_target>_seed_corpus.txt` the file name is mandated by the framework

FrankYFTang: See https://unicode-org.github.io/icu/userguide/dev/fuzzer_targets.html#fuzzer-seed-corpus
it must be `<fuzzer_target>_seed_corpus.txt` the file name is mandated by the framework and my understanding is the file will be automatically seeded by the driver in the framework

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
FrankYFTang: I follow the example from 
https://github.com/unicode-org/icu/blob/main/icu4c/source/test/fuzzer/break_iterator_fuzzer.cpp#L27
https://github.com/unicode-org/icu/blob/main/icu4c/source/test/fuzzer/number_format_fuzzer.cpp#L24

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
FrankYFTang: That is in the pre-existing code and I didn't change it.

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
markusicu: Well, for a few years it seemed like people didn't care as much because Intel CPUs were lenient. More recently I have seen people chasing alignment issues again with sanitizers and claiming undefined behavior.

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
markusicu: The existing code at least looked correct, using separate parts of the fuzzer input for the various pieces. The new code introduces an overlap between pieces, and fixing that will complicate tracking data+size. StringPiece will help.

icu4c/source/test/fuzzer/collator_compare_fuzzer_seed_corpus.txt
markusicu: Ok. But is it text, and can be marked as such, or is it binary as GitHub currently shows?

icu4c/source/test/fuzzer/collator_compare_fuzzer.cpp
FrankYFTang: ok, I shift the data by two bytes now. We really do not need to have separate bytes for locale and strength because random 16 bits number is much greater than the permutation of strength and number of locale. Just a rnd16 %5 and rnd16/5 is good enough.

FrankYFTang:
PTAL

markusicu (commented):
lgtm please squash

icu4c/source/test/fuzzer/collator_compare_fuzzer_seed_corpus.txt
markusicu: ping

FrankYFTang: All I know about the requirement is stated in

https://unicode-org.github.io/icu/userguide/dev/fuzzer_targets.html#fuzzer-seed-corpus

"Any fuzzer seed data for a fuzzer target goes into a file with name <fuzzer_target>_seed_corpus.txt. In many cases the input parameter of the ICU API under test is of type UnicodeString, in case of which the seed data should be in UTF-16 format. As an example,see [collator_rulebased_fuzzer_seed_corpus.txt](https://github.com/unicode-org/icu/blob/main/icu4c/source/test/fuzzer/collator_rulebased_fuzzer_seed_corpus.txt)."

I do not know how to answer your question since any kind of binary combiation in theory could be feed into fuzzer in order to attack the API for correct AND incorrect input usage and the seed corpus is defined as the seeding data for that same format. They are just any kind of bytes.

markusicu: ok, binary it is

markusicu (approved):

PR #2575, opened 2023-08-31
ICU-22477 Fix the performance tests after the maven folder changes

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22477


mihnita:
These are pretty hard to test.

But I tested:
- (some) by hand (not all the possible matrix combinations) locally
- temporarily disabled the condition that the repo is 'unicode-org/icu' and triggered them by hand
  (`if: github.repository == 'unicode-org/icu' && github.ref == 'refs/heads/main'`)
  That still fails, but with "No commit information is found in payload", likely because it was
manually triggered, not from a commit.

So I am as confident as I can be, but we will not know until after the merge.

Maybe something to improve once the dust settles...

FrankYFTang (approved):

PR #2581, opened 2023-09-05
ICU-22324 Mavenization, resolve some of the release tasks

mihnita:
##### Checklist

### DIFF
--- a/icu4j/pom.xml
+++ b/icu4j/pom.xml
@@ -46,6 +46,7 @@
     <icu4j.api.doc.copyright.footer>&lt;font size=-1&gt;Copyright © 2016 Unicode, Inc. and others.&lt;/font&gt;</icu4j.api.doc.copyright.footer>
     <!-- Needs to be overriden in sub-modules -->
     <icu4j.api.doc.root.dir>${project.basedir}</icu4j.api.doc.root.dir>
+    <icu4j.api.report.version>74</icu4j.api.report.version>
 
     <!-- Used for the .jar Manifest files -->
 
@@ -170,15 +171,17 @@
             <taglets>
               <taglet>
                 <tagletClass>com.ibm.icu.dev.tool.docs.ICUTaglet</tagletClass>
-                <!-- We can use `tagletArtifact` instead of `tagletpath`,
-                     but then we must `mvn install` the `tools_build` artifact. -->
-                <tagletpath>${icu4j.api.doc.root.dir}/tools/build/target/classes</tagletpath>
               </taglet>
               <taglet>
                 <tagletClass>ch.arrenbrecht.jcite.JCiteTaglet</tagletClass>
               </taglet>
             </taglets>
             <tagletArtifacts>
+              <tagletArtifact>
+                <groupId>com.ibm.icu</groupId>
+                <artifactId>tools_build</artifactId>
+                <version>${project.version}</version>
+              </tagletArtifact>
               <tagletArtifact>
                 <groupId>org.mcraig</groupId>
                 <artifactId>jcite</artifactId>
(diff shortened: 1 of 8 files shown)


mihnita:
The icu4j/stylesheet8.css change adds a bit of padding around the sample code:

Before:
image

After:
image

icu4j/pom.xml
```diff
@@ -328,6 +331,59 @@
...
+                  ```
+                  Error while creating javadoc report:
+                  Exit code: 1
+                  Command line was: ...
+                  Refer to the generated Javadoc files in '...' dir.
+                  ```
+                But in most cases these are not related to what we do here, which is API reporting.
+                We ignore them so that we can progess with what we really care.
+              -->
+              <failOnError>false</failOnError>
```
markusicu, line 366: @yumaoka sound ok?

markusicu (commented):
Looks plausible to me. I would like to give @yumaoka a chance to weigh in as well.

echeran (approved):
LGTM

icu4j/pom.xml
yumaoka: I prefer to let it fail. `gatherapi` is a part of release process producing official API signature for a release. And it will be used for future API change report.

I always thought the check is done when any API changes are made. I'm not sure `gatherapi` is triggered during regular javadoc generation. But for all downstream release process, we need to fix issues reported by gatherapi - e.g. lacking docs for default no-args constructor, etc.

icu4j/releases_tools/api_reports.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+# Copyright (C) 2023 and later: Unicode, Inc. and others.
+# License & terms of use: http://www.unicode.org/copyright.html
+
+export MAVEN_ARGS='--no-transfer-progress'
+
+export artifact_version='74.0.1-SNAPSHOT'
```
yumaoka, line 7: I assume the hardcoded version here is a temporary thing. `74.0.1-SNAPSHOT` is also referred by line 39 in this shell script. Ideally, we want to pass version to this shell script, instead of embedding it. If it is too much, then I'd prefer to define it in one place.

yumaoka:
> The icu4j/stylesheet8.css change adds a bit of padding around the sample code:
> 
> Before: image
> 
> After: image

This stylesheet came from JDK 8. We used the same style with JDK at that time. But I'm fine to make some adjustments like this.

icu4j/releases_tools/api_reports.sh
mihnita: Everything in "release_tools" is work in progress.
Just making sure I have things in one place and don't get lost, instead of just having them on my machine.
Will fine-tune them once all the ant tasks have a maven equivalent.

mihnita: TLDR: this works as expected.

The way "gatherapi" works is that `com.ibm.icu.dev.tool.docs.GatherAPIData` is a docket.
It completely replaces the standard `javadoc`
It is invoked for each class and method, but instead of collecting the comments and generating javadoc, it only adds the method names to the API report.

https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/doclet/overview.html

So the failure is because the `maven-javadoc-plugin` does not detect the generated `javadoc`
And it does not detect it because it does not exist, as expected (it is not generated).

This didn't fail before because the ant script invoked the javadoc generation with the custom doclet,
but didn't check that the javadoc was generated, the way maven does.

The real javadoc is generated in another maven profile, which uses the standard javadoc doclet.

icu4j/pom.xml
mihnita: TLDR: this works as expected.

The way "gatherapi" works is that com.ibm.icu.dev.tool.docs.GatherAPIData is a docket.
It completely replaces the standard javadoc doclet.
It is invoked for each class and method, but instead of collecting the comments and generating javadoc, it only adds the method names to the API report.

See https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/doclet/overview.html

So the failure is because the maven-javadoc-plugin does not detect the generated javadoc
And it does not detect it because it does not exist, as expected (it is not generated).

This didn't fail before because the ant script invoked the javadoc generation with the custom doclet,
but didn't check that the javadoc was generated, the way maven does.

The real javadoc is generated in another maven profile, which uses the standard javadoc doclet.

markusicu:
@yumaoka could you please take another look today? We should get this in.

markusicu:
35 failing CI checks -- all C++ ??!

icu4j/pom.xml
```diff
@@ -328,6 +331,61 @@
...
+                The way "gatherapi" works is that com.ibm.icu.dev.tool.docs.GatherAPIData is a doclet.
+                It completely replaces the standard javadoc doclet.
+                It is invoked for each class and method, but instead of collecting the comments
+                and generating html, it only adds the method names to the API report.
+
+                See https://docs.oracle.com/javase/8/docs/technotes/guides/javadoc/doclet/overview.html
+
+                maven-javadoc-plugin fails because it does not detect the generated javadoc
+                And it does not detect it because it does not exist, as expected (it is not generated).
+                So we suppress that faulure, bacuse it works as intended.
```
markusicu, line 366:
```suggestion
                maven-javadoc-plugin fails because it does not detect the generated javadoc.
                And it does not detect it because the javadoc does not exist, as expected (it is not generated).
                So we suppress that failure, because "gatherapi" works as intended.
```

markusicu (commented):
lgtm except for typos, ambiguous "it" references, and... lots of CI check failures

markusicu (approved):
thanks -- approving, in case the approval survives beyond squashing

echeran (approved):
RSLGTM

markusicu (approved):

PR #2584, opened 2023-09-06
ICU-22479 Add fuzzer for Calendar API

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22479

### DIFF
--- a/icu4c/source/test/fuzzer/calendar_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/calendar_fuzzer.cpp
@@ -1,0 +1,132 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+// Fuzzer for ICU Calendar.
+
+#include <cstring>
+
+#include "fuzzer_utils.h"
+
+#include "unicode/calendar.h"
+#include "unicode/localebuilder.h"
+#include "unicode/locid.h"
+
+icu::TimeZone* CreateRandomTimeZone(uint16_t rnd) {
+    icu::Locale und("und");
+    UErrorCode status = U_ZERO_ERROR;
+    std::unique_ptr<icu::StringEnumeration> enumeration(
+        icu::TimeZone::createEnumeration(status));
+    if (U_SUCCESS(status)) {
+        int32_t count = enumeration->count(status);
+        if (U_SUCCESS(status)) {
+            int32_t i = rnd % count;
+            const icu::UnicodeString* id = nullptr;
+            do {
+              id = enumeration->snext(status);
+            } while (U_SUCCESS(status) && --i > 0);
+            if (U_SUCCESS(status)) {
+                return icu::TimeZone::createTimeZone(*id);
+            }
+        }
+    }
+    return icu::TimeZone::getGMT()->clone();
+}
+const char* GetRandomCalendarType(uint8_t rnd) {
+    icu::Locale und("und");
+    UErrorCode status = U_ZERO_ERROR;
+    std::unique_ptr<icu::StringEnumeration> enumeration(
+        icu::Calendar::getKeywordValuesForLocale("calendar", und, false, status));
+    const char* type = "";
+    if (U_SUCCESS(status)) {
+        int32_t count = enumeration->count(status);
+        if (U_SUCCESS(status)) {
+            int32_t i = rnd % count;
+            do {
+              type = enumeration->next(nullptr, status);
+            } while (U_SUCCESS(status) && --i > 0);
+        }
+    }
+    type = uloc_toUnicodeLocaleType("ca", type);
+    return type;
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+    uint16_t rnd;
+    if (size < 2*sizeof(rnd) + 1) return 0;
+    icu::StringPiece fuzzData(reinterpret_cast<const char *>(data), size);
+    // Byte 0 and 1 randomly select a TimeZone
+    std::memcpy(&rnd, fuzzData.data(), sizeof(rnd));
+    fuzzData.remove_prefix(sizeof(rnd));
+    std::unique_ptr<icu::TimeZone> timeZone(CreateRandomTimeZone(rnd));
+
+    // Byte 1 and 2 randomly select a Locale
+    std::memcpy(&rnd, fuzzData.data(), sizeof(rnd));
+    fuzzData.remove_prefix(sizeof(rnd));
+    icu::Locale locale = GetRandomLocale(rnd);
+
+    // Byte 4 randomly select a Calendar type
+    const char* type = GetRandomCalendarType(*fuzzData.data());
+    fuzzData.remove_prefix(1);
+
+    UErrorCode status = U_ZERO_ERROR;
+    icu::LocaleBuilder bld;
+    bld.setLocale(locale);
+    bld.setUnicodeLocaleKeyword("ca", type);
+    locale = bld.build(status);
+    if (U_FAILURE(status)) return 0;
+    std::unique_ptr<icu::Calendar> cal(
+        icu::Calendar::createInstance(*timeZone, locale, status));
+    if (U_FAILURE(status)) return 0;
+    cal->clear();
+
+    int32_t amount;
+    double time;
+    while (fuzzData.length() > 2 + static_cast<int32_t>(sizeof(time))) {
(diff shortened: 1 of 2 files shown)


icu4c/source/test/fuzzer/calendar_fuzzer.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+    if (size < 2) return 0;
+    uint16_t rnd16;
+    std::memcpy(&rnd16, data, sizeof(rnd16));
+    data += 2;
+    size -= 2;
+    icu::Locale locale = GetRandomLocale(rnd16);
+    const char* type = GetRandomCalendarType(rnd16);
+    std::unique_ptr<icu::TimeZone> timeZone(CreateRandomTimeZone(rnd16));
```
richgillam, line 61: Doesn't this mean you're using the same random number for all three of these values?  Is this really what you want to do?  Maybe it's okay because it's random, but it seems like it'd always yoke together certain combinations of locale, calendar type, and time zone instead of giving you truly random combinations.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
```diff
@@ -0,0 +1,125 @@
...
+    cal->clear();
+
+    while (size > 1 + sizeof(double)) {
+        uint8_t rnd8 = *data++;
+        uint32_t rnd32;
+        double date;
+        std::memcpy(&date, data, sizeof(date));
+        std::memcpy(&rnd32, data, sizeof(rnd32));
+        UCalendarDateFields field = static_cast<UCalendarDateFields>(rnd32 % UCAL_FIELD_COUNT);
+        int32_t srnd32 = (rnd32 % 2 == 0 ? -1 : 1) * rnd32;
```
richgillam, line 80: If I'm reading this correctly, it's saying that `srnd32` is the same as `rnd32` if `rnd32` is odd, and the negative of `rnd32` if `rnd32` is even.  So you're trying the sign to the ones bit.  I'm still not quite sure how this is being used, but that also seems to tie these two properties together in ways that won't give you true randomness.

In fact, I see `field` is derived from `rnd32` as well, so you're restricting yourself to certain combinations of field and increment value in the tests below.  Again, maybe this is okay, but it's not completely random.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+        // Shift right 0 - 23 bits to distribute the value more evenly on
+        // different scale.
+        rnd32 >>= (rnd32 % 23);
+        srnd32 >>= (srnd32 % 23);
```
richgillam, line 85: You're also tying how far you shift to the original value.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+        data += sizeof(double);;
+        size -= (1 + sizeof(double));
```
richgillam, line 88: The 1 is because of the read on line 74, I presume, and the `sizeof(double)` in both expressions is because of what you're doing in ll. 77-78, but that's not immediately obvious.  You might want to either adjust `size` and `data` closer to where you're reading from the data buffer or add a comment.

richgillam (approved):
I'm not sure I'm the best person to review this, but I did look at it, and it generally looks okay, except that there are several spots where you're using the same random number from the data array for different purposes, tying certain input values and operations together in ways that you might not want to be doing.  I don't know if this is okay or not, so I'm approving, but it's something you might want to think about.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
markusicu: I share Rich's hunch. I believe that the fuzzing framework uses some kind of guided-random generator that automatically finds interesting combinations of inputs. For that to work well, it seems like a distinct/disjoint input byte sequence should be used for each algorithm input.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
markusicu: I like wrapping the input (data, size) pair into a StringPiece and advancing them together via remove_prefix.
If we were on C++17, then we could use a `std::basic_string_view<uint8_t>` with fewer char/uint8_t casts. Can the fuzzers be compiled with C++17?
If we were on C++20, then we could use a std::span but it doesn't have a handy remove_prefix so we would have to use last() or subspan() I guess.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: "the same random number for all three of these values?"
Not exactly. The random number for each are % by the number of combinations for each category therefore are not the same. 

consider the following, if we have X locales, Y calendars and Z timezones.

The locale to be pick is R % X
The calendar to be pick is R % Y
The timezone to be pick is R % Z

As long as X x Y x Z << 2^16 we should hit any combination of locale, calendar, timezone randomly. 
Of course, if X Y and Z share common denomiator then some combination will not be hit. 

We currently have 831 Locales, 18 calendars, and 637 timezones

I can change these lines to 
```
    icu::Locale locale = GetRandomLocale(rnd16 & 0x003ff);  
    const char* type = GetRandomCalendarType( (rnd16 >> 10) & 0x001f ); 
    std::unique_ptr<icu::TimeZone> timeZone(CreateRandomTimeZone((rnd16 >> 15) & 0x003ff );
```
to make sure they pass in muturally exclusive bits from the rnd16.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: I try to make the srnd32 not randomly distributed evenly to the 32bit space, but more evenly in different scale. If I do not do that, the chances of hitting a random very small value is much smaller than hitting a random very large value, for example, there are 10 single digits number can be hit, and 90 double digitis and 900 three ditits number to be hit if we generate randomly without adjustment. Therefore, the code will test 10 times more on the range of 3 digits number than 2 digits number, and 90 times more on  the range of 3 ditis number than a single digit number. I try to shift the bit also based on the random info so the chances of hitting 3 digits random number is closer to hit a 1 digit random number.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
richgillam: My concern is that those values are still mathematically related and that you might miss combinations as a result, but maybe it's not really a problem in practice.  I like the alternative you suggest better, but I also wonder-- why not have a simple wrapper class around the random number list you can just read random numbers from, like a stream or a generator?  Then you don't have to worry about updating `data` and `size` every time to use a random number, because you're calling a separate class that does that for you.  I think this is basically what Markus is suggesting below, too.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
richgillam: That's fine, and I agree with you, but a) you're always shifting `rnd32` and `srnd32` the same distance (which maybe isn't really a problem), and b) the random number that determines the actual shift distance is the same random number you're using for other purposes in this test, again meaning two values that should be totally independent of each other have a predictable mathematical relationship.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: ok, I address all these now.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: ok, not shifting now.

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: done

icu4c/source/test/fuzzer/calendar_fuzzer.cpp
FrankYFTang: ok, I just simplify it and avoid complicate logic.

richgillam (approved):
Yeah, I think this looks better.  Thanks for the changes!

PR #2586, opened 2023-09-07
ICU-22169 Fixed ures_getFunctionalEquivalent() to check the resource …

richgillam:
…bundles' %%Parent resources when ascending

the parent tree.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22169

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -2974,6 +2974,31 @@ static UBool isLocaleInList(UEnumeration *locEnum, const char *locToSearch, UErr
     return false;
 }
 
+static void getParentForFunctionalEquivalent(const char*      localeID,
+                                             UResourceBundle* res,
+                                             UResourceBundle* bund1,
+                                             char*            parent,
+                                             int32_t          parentCapacity) {
+    // Get parent.
+    // First check for a parent from %%Parent resource (Note that in resource trees
+    // such as collation, data may have different parents than in parentLocales).
+    UErrorCode subStatus = U_ZERO_ERROR;
+    parent[0] = '\0';
+    if (res != NULL) {
+        ures_getByKey(res, "%%Parent", bund1, &subStatus);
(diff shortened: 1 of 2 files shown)


richgillam:
@pedberg-icu , I think you especially want to look at this, since this is basically your code.  I thought I had a simpler way to do this, but it didn't work-- what you're doing in Apple ICU seems to be the only way to do this-- so I refined your fix a little and posted it here.

icu4c/source/common/uresbund.cpp
```diff
@@ -3145,10 +3174,33 @@ ures_getFunctionalEquivalent(char *result, int32_t resultCapacity,
...
+        UBool haveFound = false;
+        // At least for collations which may be aliased, we need to use the VALID locale
+        // as the parent instead of just truncating, as long as the VALID locale is not
+        // root and has a different language than the parent. Use of the VALID locale
+        // here is similar to the procedure used at the end of the previous do-while loop
+        // for all resource types.
+        if (res != NULL && uprv_strcmp(resName, "collations") == 0) {
+            subStatus = U_ZERO_ERROR;
+            const char *validLoc = ures_getLocaleByType(res, ULOC_VALID_LOCALE, &subStatus);
+            if (U_SUCCESS(subStatus) && validLoc != NULL && validLoc[0] != 0 && uprv_strcmp(validLoc, "root") != 0) {
```
pedberg-icu, line 3186: The check for validLoc != "root" may be unnecessary in open-source ICU (not entirely sure), but it certainly does not hurt anything. Apple ICU retains use of "root" as a name where open-source ICU may use empty string...

pedberg-icu (approved):

icu4c/source/common/uresbund.cpp
richgillam: Thanks; if it doesn't hurt, I'm going to leave it.

PR #2591, opened 2023-09-09
ICU-22478 Fix log spew in PersonNameConsistencyTest.

richgillam:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22478

### DIFF
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/format/PersonNameConsistencyTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/format/PersonNameConsistencyTest.java
@@ -28,10 +28,16 @@ import junitparams.Parameters;
  */
 @RunWith(JUnitParamsRunner.class)
 public class PersonNameConsistencyTest extends TestFmwk {
+    /**
+     * Change this to true to cause the tests that would normally be skipped to also run (without
+     * causing the test suite to fail).
+     */
+    private static boolean RUN_ALL_TESTS = false;
+
     private static final String DATA_PATH = TestUtil.DATA_PATH + "cldr/personNameTest/";
 
     static private Collection<String> FILENAMES_TO_SKIP =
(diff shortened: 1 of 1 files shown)


richgillam:
This change cleans up the log spew we were getting from the PersonNameConsistencyTest due to the fact that we were calling `System.out.println()` all over the place.  Now we don't send anything to the log unless a test file actually has an error, and we don't get any log output for tests we're supposed to skip.

I changed most of the `println()`s to `logln()`s and added a mechanism for still running the skipped tests, but that uncovered https://unicode-org.atlassian.net/browse/ICU-22492: `logln()` doesn't actually do anything; all of its guts are commented out.  So anyone trying to debug a test failure will either have to restore the `println()`s or change `logln()`.  But it seems to me that our logging in the ICU4J unit tests is seriously broken, probably has been for a long time, and should be fixed.

I also discovered that we had a bug in this test where tests that we aren't skipping, but that throw an exception, we're showing up as failures.  We had two of these `gd.txt` and `lv.txt`, both for the same reason: missing modifier names.  `gd.txt` has rules that expect `vocative`, and `lv.txt` has rules that expect `genitive`.  I suspect we had both under the "log known issue" pile before, so I added them back, but this is production CLDR data we're not handling correctly, so do I really need to (groan) file a new API-change request and add those two modifier keywords to `FieldModifier` and `FieldModifierImpl`?  Can we defer that to ICU 75? Is there an existing ticket, or do I need to create one?  @macchiati, can you shed light on this?

For that matter, what about the other locales we're currently skipping?  Are there outstanding tickets for those?  Do we know why they're failing?  And what's ICU-17028?  This seems like an awfully low ticket number to have anything to do with person names, and I don't seem to have access to it for some reason.

pedberg-icu (approved):

FrankYFTang (approved):
LGTM

PR #2595, opened 2023-09-12
ICU-22489 Clarify the default setting of Collator

FrankYFTang:
Make the document clear that for a Collator object normally created the default setting is not a specific constant value but depends on the locale, and therefore the value store in the resource as the data for that locale.

See the background information about the historical misunderstanding in 
* **ECMA402** (since 2012 see https://402.ecma-international.org/1.0/#sec-10.1.1.1 the ignorePuncation therefore was design to default to "false" always, instead of undefined or locale dependent . Issues in https://github.com/tc39/ecma402/issues/832 fix in https://github.com/tc39/ecma402/pull/833 )
* **V8** (since 2018-08-08 by gsathya@ - see line 508-511 in https://source.chromium.org/chromium/_/chromium/v8/v8.git/+/363fe1eb66371e55971cbde815369182189cf19f:src/objects/js-collator.cc;dlc=18a0a9746f3147950eaca355ca4a5d5d08e15552 )
* **Mozilla/JSC** - see  https://github.com/tc39/ecma402/issues/832#issuecomment-1712470101 )

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22489

### DIFF
--- a/icu4c/source/i18n/unicode/ucol.h
+++ b/icu4c/source/i18n/unicode/ucol.h
@@ -251,42 +251,52 @@ typedef enum {
       */
      UCOL_FRENCH_COLLATION, 
      /** Attribute for handling variable elements.
-      * Acceptable values are UCOL_NON_IGNORABLE (default)
-      * which treats all the codepoints with non-ignorable 
+      * Acceptable values are UCOL_NON_IGNORABLE
+      * which treats all the codepoints with non-ignorable
       * primary weights in the same way,
-      * and UCOL_SHIFTED which causes codepoints with primary 
+      * and UCOL_SHIFTED which causes codepoints with primary
       * weights that are equal or below the variable top value
-      * to be ignored on primary level and moved to the quaternary 
-      * level.
+      * to be ignored on primary level and moved to the quaternary
+      * level. The default setting in a Collator object depends on the
+      * locale data loaded from the resources. For most locales, the
+      * default is UCOL_NON_IGNORABLE, but for others, such as "th",
+      * the default could be UCOL_SHIFTED.
       * @stable ICU 2.0
       */
-     UCOL_ALTERNATE_HANDLING, 
+     UCOL_ALTERNATE_HANDLING,
      /** Controls the ordering of upper and lower case letters.
-      * Acceptable values are UCOL_OFF (default), which orders
+      * Acceptable values are UCOL_OFF, which orders
       * upper and lower case letters in accordance to their tertiary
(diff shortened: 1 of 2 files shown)


richgillam (dismissed):
I can't speak to the accuracy of the information here (I know diddly about the collation implementation at this point), but the documentation itself is clear and seems to make sense.

FrankYFTang:
* Example of UCOL_ALTERNATE_HANDLING default to UCOL_SHIFTED - "th"
https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/th.txt#L9
* Example of UCOL_CASE_FIRST default to UCOL_UPPER - "da" (only for "standard" not for "search") and "mt"
  * https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/da.txt#L16 
  * https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/mt.txt#L8
* Example of UCOL_NORMALIZATION_MODE default to UCOL_ON - "hi", "vi", "bn" (and many more):
  * https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/hi.txt#L8
  * https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/vi.txt#L8
  * https://github.com/unicode-org/icu/blob/main/icu4c/source/data/coll/bn.txt#L8

markusicu:
Looks ok for C/C++.
What about Java?
- https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/RuleBasedCollator.html#setAlternateHandlingShifted-boolean-
  - “The default value for this mode is false, ...”
- https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/RuleBasedCollator.html#setUpperCaseFirst-boolean-
- ...

I guess the User Guide is mostly out of scope for the moment because we lost important formatting in the migration from Google Sites and apparently didn't notice:
https://unicode-org.github.io/icu/userguide/collation/customization/#default-options
“Default options are **in emphasis**.”
... but in Markdown you can't emphasize `code` so we lost that formatting ...

We *could* add now that the default options may be overridden by the rules in the locale data.

FrankYFTang:
> What about Java?

Thanks for pointing out that, I now change Java too. PTAL

markusicu (commented):
lgtm pse squash

markusicu (approved):

macchiati (approved):

PR #2597, opened 2023-09-14
ICU-22324 Include all unit tests, move tests needing multiple components

echeran:
Trying to import ICU4J into a new IntelliJ project surfaced the issue that our Maven build isn't running all of the tests. Some test files have names not fitting the default convention (`Test....java` or `...Test.java`).

Re-enabling those tests then surfaced 2 tests in `core` that depend on `currdata`. Those 2 test files were moved to `common_tests` on the assumption that `common_tests` is where tests depending on multiple components/modules are being placed.

##### Checklist


mihnita (approved):

icu4j/pom.xml
```diff
@@ -93,6 +93,18 @@
...
+          <configuration>
+            <includes>
+              <include>**/Test*.java</include>
+              <include>**/*Test.java</include>
+              <include>**/*Tests.java</include>
+              <include>**/*TestCase.java</include>
+              <!-- Some test files start with "IntlTest...". Many don't fit
+               the Surefire test file name default patterns. To be safe, use
+               a catchall pattern. -->
+              <include>**/*.java</include>
```
markusicu, line 105: Do we need the first four includes when we have the last one as a "catchall"?

echeran: No, they're redundant. I copied them because those are the plugin's defaults for the includes. Should I remove them?

markusicu: I think that the unexplained redundancy is confusing. I suggest you do one of
- remove the subset includes
- comment out the subset includes

Either way, please add a comment saying why we don't just use the default (maybe with a link to the default, or with an example).

echeran: Done.

PR #2601, opened 2023-09-15
ICU-22325 Disable tests as workaround for failures following CI changes

echeran:
I hope this workaround unblocks PRs from failing Java tests. 

The issue for the underlying problem is https://unicode-org.atlassian.net/browse/ICU-22505

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22325
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [X] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [ ] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2603, opened 2023-09-16
ICU-22324 Allow single Maven cache creation and multiple read-only usage

echeran:
We still have instances of CI jobs needing to be rerun because the build process dependencies are not being cached. That gives us errors like "...dependencies could not be resolved: Failed to read artifact descriptor..." or "Could not transfer artifact... Connection timed out (Read failed)", as seen today: https://github.com/unicode-org/icu/pull/2601#issuecomment-1721499440

This PR:
* chooses the job with the most build dependencies to create the artifact cache, and the other jobs have a read-only view of that cache.
* changes the Maven target for building from `install` to `verify` so that no artifacts built from the current snapshot of the ICU codebase are stored in the cache, while still ensuring `integration-tests` is run
* gets rid of redundant CI jobs for ICU4J as a partial cleanup

##### Checklist

### DIFF
--- a/.github/workflows/cache_retain.yml
+++ b/.github/workflows/cache_retain.yml
@@ -26,6 +26,9 @@ on:
     # this cron schedule is set to run every 6 days to ensure retention
     - cron: '0 12 */6 * *'
 
+env:
+  SHARED_MVN_ARGS: '--show-version --no-transfer-progress'
+
 permissions:
   contents: read
 
@@ -44,8 +47,15 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: '11'
-          cache: maven
+      - name: Restore read-only cache of local Maven repository
+        uses: actions/cache/restore@v3
+        id: cache
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
       - name: Run Maven unit & integration tests
         run: |
-          cd icu4j/maven-build;
-          mvn --batch-mode verify
+          cd icu4j;
+          mvn ${SHARED_MVN_ARGS} verify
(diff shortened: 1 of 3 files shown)


markusicu (dismissed):
lgtm

mihnita (commented):
1. removing  icu4j-test-maven and icu4j-verify-maven: LGTM

2. replacing install with verify: LGTM

3. the custom cache part, I am still not convinced :-(
Not so much the idea, but not convinced by the implementation.

I wonder if there is a way to "inject" that `lookup-only` in the existing `cache: maven`

There might also be a way to define "blocks" of yaml content and refer to them.
The `cache: maven` is defined and implemented somewhere, after all.

Last, I am not sure `lookup-only` is the right flag to use.
See here: https://github.com/actions/cache/pull/1041/files
> `lookup-only` - Skip downloading cache. Only check if cache entry exists.

So it is not every task will download and use the cache, but don't upload (read only),
and only one will upload.
It is really: only use cache for one single task, not for the others.

markusicu:
icu4j-mvn-init failure:

Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.8:go-offline (default-cli): Goal requires a project to execute but there is no POM in this directory (/home/runner/work/icu/icu). Please verify you invoked Maven from the correct directory.

echeran:
Tests are passing, need a stamp now.

Note: in addition to the PR description at the top about what this PR does: this PR also cleans up the console output by removing artifact download progress printouts. We do this via the CLI args to Maven. Since the version of Maven currently on the Github Actions Linux VM runner images is 3.8.8, but the special `MAVEN_ARGS` environment variable is only supported in Maven 3.9.0 onwards for setting default Maven CLI args, we made our own environment variable called `SHARED_MVN_ARGS` and used it throughout.

Note: there is an apparent bug in the `maven-dependency-plugin` as used in the `icu4j-mvn-init-cache` job to download all of the project's artifact dependencies. It complains about not being able to download `jdk.tools:jdk.tools:1.8`, even though that is a `system` scope artifact to be found on the filesystem. So the workaround is to exclude the only profile that triggers the resolving (downloading) of that single artifact.

PR #2607, opened 2023-09-19
ICU-22508 Reject NaN as input to Calendar::setTimeInMillis

FrankYFTang:
Fix DateFormat::format issue, format internally call Calendar:: setTime and setTimeInMillis  and if the date is NaN it will cause problem. The two if check will be false because nan is not a number and compare with a constant will always be false. 

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22508


richgillam (approved):
Good catch!

PR #2608, opened 2023-09-19
ICU-22509 Limit the dateStyle and timeStyle in fuzzer to only valid enum

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22509

### DIFF
--- a/icu4c/source/test/fuzzer/date_format_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/date_format_fuzzer.cpp
@@ -12,20 +12,40 @@
 
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
     uint16_t rnd;
+    uint8_t rnd2;
     UDate date;
-    icu::DateFormat::EStyle dateStyle;
-    icu::DateFormat::EStyle timeStyle;
-    if (size < sizeof(rnd) + sizeof(date) + sizeof(dateStyle) + sizeof(timeStyle)) return 0;
+    icu::DateFormat::EStyle styles[] = {
+        icu::DateFormat::EStyle::kNone,
+        icu::DateFormat::EStyle::kFull,
+        icu::DateFormat::EStyle::kLong,
+        icu::DateFormat::EStyle::kMedium,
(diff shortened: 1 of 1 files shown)


icu4c/source/test/fuzzer/date_format_fuzzer.cpp
```diff
@@ -12,20 +12,40 @@
...
+        icu::DateFormat::EStyle::kShort,
+        icu::DateFormat::EStyle::kDateOffset,
+        icu::DateFormat::EStyle::kDateTime,
+        icu::DateFormat::EStyle::kDateTimeOffset,
+        icu::DateFormat::EStyle::kRelative,
+        icu::DateFormat::EStyle::kFullRelative,
+        icu::DateFormat::EStyle::kLongRelative,
+        icu::DateFormat::EStyle::kMediumRelative,
+        icu::DateFormat::EStyle::kShortRelative,
+    };
```
richgillam, line 31: It seems bad to duplicate this whole enum here like this; then we have to be careful to keep this in sync with the "real" one.  There's no way to just include the actual header?

richgillam (approved):
One concern, but generally looks fine.

icu4c/source/test/fuzzer/date_format_fuzzer.cpp
FrankYFTang: take a look at https://github.com/unicode-org/icu/blob/d8659b476d4703b18bb9ea040798c1e62ff3329e/icu4c/source/i18n/unicode/datefmt.h#L166

richgillam: Oh, I get it.  You *are* including the regular header.  I misunderstood what you were doing here.  Sorry about that.

FrankYFTang:
I add a new PR in https://github.com/unicode-org/icu/pull/2612 which take both the new code in this PR and the way I test before this PR to test both the valid and invlide enum and fix the bug by adding a validatino internally to prevent infinity loop when the style is not valid.

richgillam:
Looks great!  Approved.

—Rich

PR #2616, opened 2023-09-21
ICU-22463 Fix the conversion from gasoline-equivalent units to kilograms-per-meter-squared-per-second

younies:
Correcting the conversion from gasoline-equivalent units to kilograms-per-meter-squared-per-second is essential. The issue doesn't lie with the Java segment, as Java code employs a more appropriate method for handling unit constants.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22463

### DIFF
--- a/icu4c/source/i18n/units_converter.cpp
+++ b/icu4c/source/i18n/units_converter.cpp
@@ -372,11 +372,11 @@ void U_I18N_API addSingleFactorConstant(StringPiece baseStr, int32_t power, Sign
         factor.constantExponents[CONSTANT_FT2M] += 3 * power * signum;
     } else if (baseStr == "in3_to_m3") {
         factor.constantExponents[CONSTANT_FT2M] += 3 * power * signum;
-        factor.factorDen *= 12 * 12 * 12;
+        factor.factorDen *= std::pow(12 * 12 * 12, power * signum);
     } else if (baseStr == "gal_to_m3") {
-        factor.factorNum *= 231;
         factor.constantExponents[CONSTANT_FT2M] += 3 * power * signum;
-        factor.factorDen *= 12 * 12 * 12;
+        factor.factorNum *= std::pow(231, power * signum);
+        factor.factorDen *= std::pow(12 * 12 * 12, power * signum);
     } else if (baseStr == "gal_imp_to_m3") {
         factor.constantExponents[CONSTANT_GAL_IMP2M3] += power * signum;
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/units_converter.cpp
```diff
@@ -372,11 +372,20 @@ void U_I18N_API addSingleFactorConstant(StringPiece baseStr, int32_t power, Sign
         factor.constantExponents[CONSTANT_FT2M] += 3 * power * signum;
     } else if (baseStr == "in3_to_m3") {
         factor.constantExponents[CONSTANT_FT2M] += 3 * power * signum;
-        factor.factorDen *= 12 * 12 * 12;
+        if (signum == Signum::NEGATIVE) {
+            factor.factorNum *= 12 * 12 * 12;
+        } else {
+            factor.factorDen *= 12 * 12 * 12;
+        }
```
sffc, line 379: Do we also need to consider `power` here?

younies: Excellent, yes.

We are doing it in Java but not here.

we should add a test case to test this case for the future.

markusicu: I see that Younies changed the implementation to consider power, but I don't see a new “test case to ensure this scenario is accounted for in the future.”

younies: This test need to be added to the CLDR: `unitsTest.txt` which should be in another PR: one in CLDR first and then one to ICU.

we need to add something like:

`square-gasoline-equivalent` to `square-kilograms-per-square-meter-p4-per-second`

Or something more simpler:

`square-gallon` to `p6-meter`.

or both.

markusicu: If you have an open CLDR ticket for adding that test case, then that should be good. I don't care if it's a ticket just for this, or an item in some other open ticket. Please reply with the CLDR ticket number.

We probably don't need a separate ICU ticket, assuming the next CLDR-to-ICU integration picks up the test data.

younies: sure, here is it

https://unicode-org.atlassian.net/browse/CLDR-17114

markusicu (approved):
lgtm
@sffc ok?

sffc (approved):
More tests would be nice but the logic looks correct to me now so let's merge it

PR #2617, opened 2023-09-21
ICU-22498 Write directly to sink (instead of into legacy buffers)

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22498

### DIFF
--- a/icu4c/source/common/uloc_tag.cpp
+++ b/icu4c/source/common/uloc_tag.cpp
@@ -1888,11 +1888,8 @@ static void
 _appendPrivateuseToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool strict, UBool hadPosix, UErrorCode* status) {
     (void)hadPosix;
     char buf[ULOC_FULLNAME_CAPACITY];
-    char tmpAppend[ULOC_FULLNAME_CAPACITY];
     UErrorCode tmpStatus = U_ZERO_ERROR;
     int32_t len, i;
-    int32_t reslen = 0;
-    int32_t capacity = sizeof tmpAppend;
 
     if (U_FAILURE(*status)) {
         return;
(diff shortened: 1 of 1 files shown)


markusicu (approved):
lgtm tnx

icu4c/source/common/uloc_tag.cpp
```diff
@@ -1945,37 +1942,18 @@ _appendPrivateuseToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool
...
-                            if (reslen < capacity) {
-                                uprv_memcpy(tmpAppend + reslen, PRIVUSE_VARIANT_PREFIX, uprv_min(len, capacity - reslen));
-                                reslen += uprv_min(len, capacity - reslen);
-                            }
-
-                            if (reslen < capacity) {
-                                tmpAppend[reslen++] = SEP;
-                            }
-
+                            sink.Append(PRIVATEUSE_KEY, UPRV_LENGTHOF(PRIVATEUSE_KEY) - 1);
```
FrankYFTang, line 1948: why "- 1" here?

FrankYFTang, line 1948: Is  uprv_strlen(PRIVATEUSE_KEY) better?

icu4c/source/common/uloc_tag.cpp
```diff
@@ -1945,37 +1942,18 @@ _appendPrivateuseToLanguageTag(const char* localeID, icu::ByteSink& sink, UBool
...
+                            sink.Append("-", 1);
+                            sink.Append(PRIVUSE_VARIANT_PREFIX, UPRV_LENGTHOF(PRIVUSE_VARIANT_PREFIX) - 1);
+                            sink.Append("-", 1);
```
FrankYFTang, line 1951: Is  uprv_strlen(PRIVUSE_VARIANT_PREFIX) better?

icu4c/source/common/uloc_tag.cpp
roubert: I would find it a bit odd to call a function to dynamically calculate the length of a static string constant that has a length known at compile time.

icu4c/source/common/uloc_tag.cpp
roubert: I would find it a bit odd to call a function to dynamically calculate the length of a static string constant that has a length known at compile time.

PR #2620, opened 2023-09-21
ICU-22325 BRS#13: urename.h for ICU 74

sven-oly:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22325


icu4c/source/common/unicode/urename.h
```diff
@@ -512,9 +513,6 @@
 #define ubrk_setText U_ICU_ENTRY_POINT_RENAME(ubrk_setText)
 #define ubrk_setUText U_ICU_ENTRY_POINT_RENAME(ubrk_setUText)
 #define ubrk_swap U_ICU_ENTRY_POINT_RENAME(ubrk_swap)
-#define ucache_compareKeys U_ICU_ENTRY_POINT_RENAME(ucache_compareKeys)
-#define ucache_deleteKey U_ICU_ENTRY_POINT_RENAME(ucache_deleteKey)
-#define ucache_hashKeys U_ICU_ENTRY_POINT_RENAME(ucache_hashKeys)
```
markusicu, line 517: FYI @pedberg we looked at these, and they are defined inside U_NAMESPACE_BEGIN so they should be part of the general C++ renaming via the renamed "icu" namespace. They have been inside the namespace for 9 years, so we suspect that you just retained these out of an abundance of caution. ICU 74 should not need these entries.

markusicu (approved):

icu4c/source/common/unicode/urename.h
pedberg-icu: @markusicu Sounds good, thanks

PR #2624, opened 2023-09-22
ICU-22324 Mavenization, minimize the places that need a version

mihnita:
This should make the version update easier:
 - run maven to update `<version>` in all projects (`mvn versions:set` or `mvn release:update-versions`, tbd)
 - `icu.major.version` in the root pom
 - three variables in `releases_tools/api_reports.sh`

Can also grep / search for `"Version update!"`

##### Checklist

### DIFF
--- a/icu4j/pom.xml
+++ b/icu4j/pom.xml
@@ -45,21 +45,24 @@
     <proj-title>International Components for Unicode for Java</proj-title>
     <proj.displayname>${project.artifactId}</proj.displayname>
 
+    <!-- Version update! -->
+    <icu.major.version>74</icu.major.version>
+
     <!-- Variables used to generate the javadoc -->
-    <api.doc.version>74 Release Candidate</api.doc.version>
+    <!-- Version update! -->
+    <api.doc.version>${icu.major.version} Release Candidate</api.doc.version>
     <icu4j.api.doc.window.title>ICU4J ${api.doc.version}</icu4j.api.doc.window.title>
     <icu4j.api.doc.header>ICU4J ${api.doc.version}</icu4j.api.doc.header>
     <icu4j.api.doc.copyright.footer>&lt;font size=-1&gt;Copyright © 2016 Unicode, Inc. and others.&lt;/font&gt;</icu4j.api.doc.copyright.footer>
+
     <!-- Needs to be overriden in sub-modules -->
     <icu4j.api.doc.root.dir>${project.basedir}</icu4j.api.doc.root.dir>
-    <icu4j.api.report.version>74</icu4j.api.report.version>
+    <icu4j.api.report.version>${icu.major.version}</icu4j.api.report.version>
 
     <!-- Used for the .jar Manifest files -->
-
-    <mf.Specification-Version>74</mf.Specification-Version>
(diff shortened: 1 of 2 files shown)


mihnita:
* Also made the `releases_tools/api_reports.sh` script friendlier
(checks that the default JDK is 8, and checks the output, and tells you where it is)
* Forced the compiler to go "all JDK 8" (added --release, and workaround a compiler bug)

icu4j/pom.xml
```diff
@@ -517,6 +531,31 @@
          <module>tools/build</module>
       </modules>
     </profile>
+
+    <!-- Starting with JDK 9 the Java compiler supports the -release version.
```
markusicu, line 535: `-release` as you write here, or `--release` as you write in the PR description?

icu4j/pom.xml
```diff
@@ -517,6 +531,31 @@
...
+      Unlinke source (checking the language features) and target (the version of the class generated),
+      this option also checks that newer APIs are not used.
+        https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html
+      Would be nice if the maven-compiler-plugin would filder this out below JDK 9, but it does not.
+      So we must use an auto-activated profile for this option.
+    -->
```
markusicu, line 541:
```suggestion
    <!-- Starting with JDK 9, the Java compiler supports the -release version.
      Unlike source (checking the language features) and target (the version of the classes generated),
      this option also checks that newer APIs are not used.
        https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html
      Would be nice if the maven-compiler-plugin would filter this out below JDK 9, but it does not.
      So we must use an auto-activated profile for this option.
    -->
```

icu4j/pom.xml
```diff
@@ -45,21 +45,24 @@
     <proj-title>International Components for Unicode for Java</proj-title>
     <proj.displayname>${project.artifactId}</proj.displayname>
 
+    <!-- Version update! -->
+    <icu.major.version>74</icu.major.version>
+
     <!-- Variables used to generate the javadoc -->
-    <api.doc.version>74 Release Candidate</api.doc.version>
+    <!-- Version update! -->
+    <api.doc.version>${icu.major.version} Release Candidate</api.doc.version>
```
markusicu, line 53: nice: using placeholders to reduce search-and-replace is great!

icu4j/pom.xml
mihnita: Double, but `--` is not allowed in XML comments :-(

mihnita: I tried to find ways to escape, but I can't find anything.

I can do `\-\-`, and it will show that way. People will think that it's weird, but they will get it.

Or I can "cheat with Unicode" (put a ZWJ between the dashes, or use the minus sign (`U+2212`)).
But then one can't copy-paste it and have it work in command line.
True, it does not work now either...

mihnita: I went with U+2212

echeran (approved):
LGTM. Let's get this in first, and then subsequently we can add user-facing docs in a followup PR.

markusicu (approved):

PR #2625, opened 2023-09-22
ICU-21289 Switch to using CharString for calling uloc_getKeywordValue()

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21289

### DIFF
--- a/icu4c/source/common/ucurr.cpp
+++ b/icu4c/source/common/ucurr.cpp
@@ -20,6 +20,7 @@
 #include "unicode/usetiter.h"
 #include "unicode/utf16.h"
 #include "ustr_imp.h"
+#include "bytesinkutil.h"
 #include "charstr.h"
 #include "cmemory.h"
 #include "cstring.h"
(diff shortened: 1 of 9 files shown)


icu4c/source/common/ucurr.cpp
```diff
@@ -520,14 +521,18 @@ ucurr_forLocale(const char* locale,
...
+        ulocimp_getKeywordValue(locale, "currency", sink, &localStatus);
+    }
+    int32_t resLen = currency.length();
+
+    if (U_SUCCESS(localStatus) && resLen == 3 && uprv_isInvariantString(currency.data(), resLen)) {
         if (resLen < buffCapacity) {
-            T_CString_toUpperCase(currency);
-            u_charsToUChars(currency, buff, resLen);
+            T_CString_toUpperCase(currency.data());
+            u_charsToUChars(currency.data(), buff, resLen);
```
markusicu, line 535: Consider using currency.appendInvariantChars(buff, resLen, ec);

markusicu (commented):
thanks, good changes!

consider timing vs. ICU 74:
- code freeze is in less than a week
- i assume that you are planning further changes like this
- if you get this in now for 74, then changes after the code freeze need to go into 75 under a new ticket split off from this one

icu4c/source/common/ucurr.cpp
roubert: Done.

roubert:
> * code freeze is in less than a week

Yes, so I'm just doing some clean-up before the freeze.

> * i assume that you are planning further changes like this

Definitely.

> * if you get this in now for 74, then changes after the code freeze need to go into 75 under a new ticket split off from this one

As already so many times before.

icu4c/source/common/ucurr.cpp
roubert: You read this backwards, right? The existing code copies from `currency` to `buff`. Calling `appendInvariantChars()` instead would copy in the opposite direction.

markusicu: oops! sorry!

markusicu:
> > * if you get this in now for 74, then changes after the code freeze need to go into 75 under a new ticket split off from this one
> 
> As already so many times before.

Possible option: Create a new ticket specifically for the 74 changes (using this PR title for the ticket title), change this PR to use that, and punt the existing ticket to 75.

roubert:
> We wouldn't need to split an open-ended ticket into two open-ended tickets.

But the nature of this work is that it's open-ended, there's no natural or logical boundary for what'll go into this and into the next release, it'll be two open-ended tickets regardless of what we do.

markusicu:
> > We wouldn't need to split an open-ended ticket into two open-ended tickets.
> 
> But the nature of this work is that it's open-ended, there's no natural or logical boundary for what'll go into this and into the next release, it'll be two open-ended tickets regardless of what we do.

We can do it that way, but because we have just two days or so left for ICU 74, work going into this release is not open-ended. Just thinking that we could carve off a specific ticket for the 74 work.

roubert:
> We can do it that way, but because we have just two days or so left for ICU 74, work going into this release is not open-ended. Just thinking that we could carve off a specific ticket for the 74 work.

That's what I mean by open-ended: The end is defined by when we run out of time, not at any particular set of changes that could be enumerated.

markusicu (approved):

PR #2633, opened 2023-09-26
ICU-22324 Mavenization, building the CLDR utilities

mihnita:
##### Checklist

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -7,7 +7,7 @@ on:
     # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
 
 env:
-  MAVEN_ARGS: '--show-version --no-transfer-progress'
+  SHARED_MVN_ARGS: '--show-version --no-transfer-progress'
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


icu4j/tools/utilities-for-cldr/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+  <profiles>
+    <profile>
+      <id>cldr_tools</id>
+      <build>
+        <plugins>
+          <plugin>
+            <!-- Only deploy when the `cldr_tools` profile is active,
+              otherwise it will also be deployed to Maven Central.
+              We can also try to specify `altDeploymentRepository` here.
+            -->
```
srl295, line 65: Is there a reason it can't be deployed to maven central? Then we might be able to have CLDR pull from there also.  Might reconsider this.

srl295 (commented):
generally LGTM

icu4j/tools/utilities-for-cldr/pom.xml
mihnita: Until now we only published it to the GitHub repo.
So I didn't want to shake the boat close to a release.

---

I think that using the Nexus snapshot repo would be nice.
It would make it possible to use without the annoying requirement that everyone configures they own access.
(GitHub token, settings.xml, https://cldr.unicode.org/development/maven)

But it is a bit more work to setup the publishing here (setup the Nexus user / password as repo secrets and using them)

mihnita: I thought a bit more about this.
In order to publish to Nexus we need to make sure we never deploy this for non-SNAPSHOT versions.
Otherwise it goes to the public maven repo (no snapshot), and we would be forced to bump the version.
So I would not do this now, I would plan it a bit better.

mihnita:
The change bringing back the SNAPSHOT in the version was merged.

Tried to run the action three times, successfully:
* Run 1: https://github.com/mihnita/icu/actions/runs/6328923716
* Run 2: https://github.com/mihnita/icu/actions/runs/6328950591
* Run 3: https://github.com/mihnita/icu/actions/runs/6329064351
* And the (non-conflicting) published packages are visible here: https://github.com/mihnita/icu/packages/1954637 (not sure if they are visible to you, but they are named `utilities-for-cldr-74.1-20230927.164410-1.jar`, `utilities-for-cldr-74.1-20230927.164709-2.jar`, and `utilities-for-cldr-74.1-20230927.165929-3.jar`. Also including the -sources.jar

---

TO USE:
- use version `74-1-SNAPSHOT`
- `UOption` moved from `com.ibm.icu.dev.tool`  to `com.ibm.icu.dev.tool.shared`, so a small code change is needed.

---

TLDR: all ready for submission

Thank you,
M.

echeran (approved):
LGTM

.github/workflows/maven.yaml
```diff
@@ -25,28 +25,22 @@ jobs:
...
-            -Dpackaging=jar \
-            -Dfile=icu4j/release_cldr/utilities.jar \
-            -Dsources=icu4j/release_cldr/utilities-src.jar \
-            -DrepositoryId=github \
-            -Durl=https://maven.pkg.github.com/${GITHUB_REPOSITORY}
+          echo Github Ref ${GITHUB_REF} @ ${GITHUB_SHA};
+          cd icu4j
+          mvn deploy ${SHARED_MVN_ARGS} \
+            -pl main/icu4j \
+            -DaltDeploymentRepository=github::https://maven.pkg.github.com/${GITHUB_REPOSITORY} \
```
srl295, line 38: @mihnita I think `-Dversion` might be needed here still

mihnita: Not needed, the .jar is now a "real maven artifact" and already has a version

PR #2473, opened 2023-05-26
ICU-22149 Be more graceful with missing lang bundle data

sffc:
I have this fully covered now in ICU4C. I wrote the code for ICU4J but I haven't tested it there.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22149
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #2635, opened 2023-09-27
ICU-22149 Be more graceful with missing lang bundle data

sffc:
Replaces https://github.com/unicode-org/icu/pull/2473

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22149

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -65,7 +65,8 @@ jobs:
         ICU_DATA_FILTER_FILE=../../.ci-builds/data-filter.json ./runConfigureICU Linux && \
         make -j2 tests && \
         \[ ! -d data/out/build/icudt66l/translit \] && \
-        (cd test/intltest && LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./intltest translit/TransliteratorTest/TestBasicTransliteratorEvenWithoutData)
+        (cd test/intltest && LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./intltest translit/TransliteratorTest/TestBasicTransliteratorEvenWithoutData) && \
+        (cd test/cintltst && LANG=C LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./cintltst /tsutil/cloctst/TestEnglishExemplarCharacters /tsutil/cldrtest/TestCoverage)
       displayName: 'Build with Data Filter'
       env:
         CC: clang
--- a/icu4c/source/i18n/ulocdata.cpp
+++ b/icu4c/source/i18n/ulocdata.cpp
@@ -68,11 +68,17 @@ ulocdata_open(const char *localeID, UErrorCode *status)
 
    uld->noSubstitute = false;
    uld->bundle = ures_open(nullptr, localeID, status);
-   uld->langBundle = ures_open(U_ICUDATA_LANG, localeID, status);
 
    if (U_FAILURE(*status)) {
       uprv_free(uld);
       return nullptr;
+   }
+
+   // ICU-22149: not all functions require lang data, so fail gracefully if it is not present
+   UErrorCode oldStatus = *status;
+   uld->langBundle = ures_open(U_ICUDATA_LANG, localeID, status);
+   if (*status == U_MISSING_RESOURCE_ERROR) {
+      *status = oldStatus;
    }
 
    return uld;
@@ -288,6 +294,11 @@ ulocdata_getLocaleDisplayPattern(ULocaleData *uld,
     if (U_FAILURE(*status))
         return 0;
 
+    if (uld->langBundle == nullptr) {
+        *status = U_MISSING_RESOURCE_ERROR;
+        return 0;
+    }
+
     patternBundle = ures_getByKey(uld->langBundle, "localeDisplayPattern", nullptr, &localStatus);
 
     if ( (localStatus == U_USING_DEFAULT_WARNING) && uld->noSubstitute ) {
@@ -339,6 +350,11 @@ ulocdata_getLocaleSeparator(ULocaleData *uld,
 
     if (U_FAILURE(*status))
         return 0;
+
+    if (uld->langBundle == nullptr) {
+        *status = U_MISSING_RESOURCE_ERROR;
+        return 0;
+    }
 
     separatorBundle = ures_getByKey(uld->langBundle, "localeDisplayPattern", nullptr, &localStatus);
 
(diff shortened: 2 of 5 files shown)


icu4c/source/i18n/ulocdata.cpp
```diff
@@ -288,6 +291,11 @@ ulocdata_getLocaleDisplayPattern(ULocaleData *uld,
     if (U_FAILURE(*status))
         return 0;
 
+    if (uld->langBundle == nullptr) {
+        *status = U_INVALID_STATE_ERROR;
```
markusicu, line 295: I think it would be more appropriate to set a U_MISSING_RESOURCE_ERROR, assuming that that's why the ures_open() call failed.

It would be even cleaner to store that ures_open() call's UErrorCode in a new field in "uld" but maybe that's overkill.

icu4c/source/i18n/ulocdata.cpp
```diff
@@ -340,6 +348,11 @@ ulocdata_getLocaleSeparator(ULocaleData *uld,
...
```
markusicu, line 352: ditto

icu4j/main/core/src/main/java/com/ibm/icu/util/LocaleData.java
```diff
@@ -225,7 +225,12 @@ public UnicodeSet getExemplarSet(int options, int extype) {
     public static final LocaleData getInstance(ULocale locale) {
         LocaleData ld = new LocaleData();
         ld.bundle = (ICUResourceBundle)UResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME, locale);
-        ld.langBundle = (ICUResourceBundle)UResourceBundle.getBundleInstance(ICUData.ICU_LANG_BASE_NAME, locale);
+        try {
+            ld.langBundle = (ICUResourceBundle)UResourceBundle.getBundleInstance(ICUData.ICU_LANG_BASE_NAME, locale);
+        } catch(Exception e) {
+            // Expected error case: ICU-22149
+            ld.langBundle = null;
```
markusicu, line 232: nit/optional: the constructor should have left it at null (Java always initializes fields) so we should not need to set it to null here

markusicu, line 232: optional: The "cleaner" C++ solution of storing the UErrorCode, if you decide to do so, would have a parallel in Java in the form of storing the Exception in a new field.

icu4j/main/core/src/main/java/com/ibm/icu/util/LocaleData.java
```diff
@@ -421,6 +426,9 @@ public static final PaperSize getPaperSize(ULocale locale){
      * @stable ICU 4.2
      */
     public String getLocaleDisplayPattern() {
+        if (langBundle == null) {
+            throw new ICUException("Language data missing for getLocaleDisplayPattern");
```
markusicu, line 430: UResourceBundle.getBundleInstance() would have thrown a MissingResourceException, right? As in C++, I think throwing that would be more appropriate. Or even storing the real exception but maybe that's overkill?

icu4j/main/core/src/main/java/com/ibm/icu/util/LocaleData.java
```diff
@@ -432,6 +440,9 @@ public String getLocaleDisplayPattern() {
      * @stable ICU 4.2
      */
     public String getLocaleSeparator() {
+        if (langBundle == null) {
+            throw new ICUException("Language data missing for getLocaleDisplayPattern");
```
markusicu, line 444: ditto

icu4c/source/i18n/ulocdata.cpp
```diff
@@ -68,13 +68,16 @@ ulocdata_open(const char *localeID, UErrorCode *status)
...
-   uld->langBundle = ures_open(U_ICUDATA_LANG, localeID, status);
 
    if (U_FAILURE(*status)) {
       uprv_free(uld);
       return nullptr;
    }
 
+   // ICU-22149: not all functions require lang data, so fail gracefully if it is not present
+   UErrorCode localStatus = U_ZERO_ERROR;
+   uld->langBundle = ures_open(U_ICUDATA_LANG, localeID, &localStatus);
```
richgillam, line 79: You don't need to check the actual error code here?  And we know that an error will mean `uld->langBundle` will sill be null?

richgillam (approved):
I don't really have any comments beyond the ones Markus already made.  When he's good with this, I'm good with it.

icu4c/source/i18n/ulocdata.cpp
markusicu: The error should _normally_ be missing-resource, but it could sometimes be memory-allocation or maybe something about the localeID.

Factory methods _should_ all return nullptr if they return with a failure code.

markusicu:
Team discussion: Ok to get this one into 74 past today's code freeze if it can be ready by next Wednesday oct04.

.ci-builds/.azure-pipelines.yml
```diff
@@ -65,7 +65,8 @@ jobs:
         ICU_DATA_FILTER_FILE=../../.ci-builds/data-filter.json ./runConfigureICU Linux && \
         make -j2 tests && \
         \[ ! -d data/out/build/icudt66l/translit \] && \
-        (cd test/intltest && LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./intltest translit/TransliteratorTest/TestBasicTransliteratorEvenWithoutData)
+        (cd test/intltest && LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./intltest translit/TransliteratorTest/TestBasicTransliteratorEvenWithoutData) && \
+        (cd icu4c/source/test/cintltst && LANG=C LD_LIBRARY_PATH=../../lib:../../tools/ctestfw ./cintltst /tsutil/cloctst/TestEnglishExemplarCharacters /tsutil/cldrtest/TestCoverage)
```
sffc, line 69: Need to fix something about this test path

```
/home/vsts/work/_temp/491c83e3-02f3-4b1e-9afb-4f051d1ab2e7.sh: line 6: cd: icu4c/source/test/cintltst: No such file or directory
```

icu4c/source/i18n/ulocdata.cpp
sffc: I made it catch only U_MISSING_RESOURCE_ERROR and then we know we can return U_MISSING_RESOURCE_ERROR because it's the only way to get to the state of a null langBundle.

markusicu (approved):
lgtm tnx!

sffc:
Feel free to click the merge button once it comes back green.

PR #2640, opened 2023-09-28
ICU-22521 Avoid crash by return U_INTERNAL_PROGRAM_ERROR instead of  "%  0"

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22521

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -1927,6 +1927,10 @@ void Calendar::roll(UCalendarDateFields field, int32_t amount, UErrorCode& statu
 
             // Now roll between start and (limit - 1).
             int32_t gap = limit - start;
+            if (gap == 0) {
+                status =  U_INTERNAL_PROGRAM_ERROR;
+                return;
+            }
             int32_t day_of_month = (internalGet(UCAL_DAY_OF_MONTH) + amount*7 -
                 start) % gap;
             if (day_of_month < 0) day_of_month += gap;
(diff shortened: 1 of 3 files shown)


FrankYFTang:
I think we have two possible way to fix the crash
 if gap (or the vairable to be % at) is 0, 
 1. skip the % step, OR
 2. set the status to U_INTERNAL_PROGRAM_ERROR and return.
 
 I am currently taking approach 1. but if @richgillam think we should take approach 2 I can change the PR to do that too.

icu4c/source/i18n/calendar.cpp
```diff
@@ -1927,6 +1927,10 @@ void Calendar::roll(UCalendarDateFields field, int32_t amount, UErrorCode& statu
 
             // Now roll between start and (limit - 1).
             int32_t gap = limit - start;
+            if (gap == 0) {
+                status =  U_INTERNAL_PROGRAM_ERROR;
+                return;
+            }
```
richgillam, line 1933: Well, this'll keep it from crashing, at least...

But no idea how we're actually getting into this state?  Is it worth trying to figure that out, or do you think it's rare enough not to worry that much about?

richgillam (approved):

icu4c/source/i18n/calendar.cpp
FrankYFTang: The algorithm is very complex here and the case cause this issue that I can find to reproduce is in many many years ago ( about ~1120000 years ago )

PR #2642, opened 2023-09-28
ICU-22324 Mavenization, publish the root pom to Maven

mihnita:
The fix seems more complicated than it is.

But there are basically 2 small changes that account for all the touched files:
1. Publishing was disabled in root, was inherited by all modules, and turned on in the 3 artifacts we care about (`icu4j`, `charset`, and `localespi`. But because now we want to publish the root artifact we need to disable it in all artifacts, except the 3. This accounts for all `the maven-deploy-plugin` ... `<skip>` changes
2. I removed `main/pom.xml` and skipped one level, so that we don't have to also publish the `main` artifact, just the root. This accounts for all the `<relativePath>../../pom.xml</relativePath>` changes in `<parent>`, and the `main/foo` in the root module.

What's left are the small but more interesting changes to review.

##### Checklist


mihnita:
Tested by publishing in my own fork and tested that cldr/tools works (after pointing it to my fork)

icu4j/pom.xml
```diff
@@ -36,8 +53,8 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
 
-    <maven_central_repo_url>https://oss.sonatype.org</maven_central_repo_url>
-    <maven-central-releases-repo-url>${maven-central-repo-url}/service/local/staging/deploy/maven2</maven-central-releases-repo-url>
+    <!-- <maven-central-repo-url>https://oss.sonatype.org</maven-central-repo-url> -->
+    <maven-central-repo-url>http://mnita.mtv.corp.google.com:8081/nexus</maven-central-repo-url>
```
markusicu, line 57: please fix...

echeran (dismissed):
LGTM, thanks!

icu4j/pom.xml
mihnita: Ouch!
Thanks, will do.

echeran (approved):

.github/workflows/maven.yaml
```diff
@@ -7,7 +7,7 @@ on:
     # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
 
 env:
-  SHARED_MVN_ARGS: '--show-version --no-transfer-progress'
+  SHARED_MVN_ARGS: '--show-version'
```
srl295, line 10: thanks

srl295 (approved):

pedberg-icu:
I will probably merge this as soon as I see that it is done (if somebody else does not get to it first...).

mihnita:
> I will probably merge this as soon as I see that it is done (if somebody else does not get to it first...).

Thank you!
I had an eye on it, but you were faster.
M

PR #2644, opened 2023-10-02
ICU-22527 Add -Wambiguous-reversed-operator test code

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22527

### DIFF
--- a/icu4c/source/test/intltest/measfmttest.cpp
+++ b/icu4c/source/test/intltest/measfmttest.cpp
@@ -6240,6 +6240,10 @@ void MeasureFormatTest::TestMeasureEquality() {
     };
     static const char *const names[] = { "1 liter", "another liter", "2 liters", "1 gram" };
 
+    // Verify that C++20 -Wambiguous-reversed-operator isn't triggered.
+    assertTrue("Equal", measures[0] == measures[1]);
+    assertTrue("Not Equal", measures[2] != measures[3]);
+
     for (int32_t i = 0; i < UPRV_LENGTHOF(measures); ++i) {
         for (int32_t j = 0; j < UPRV_LENGTHOF(measures); ++j) {
             const Measure &a = measures[i];


markusicu:
Whoops, it's a race... #2645

icu4c/source/i18n/unicode/measure.h
```diff
@@ -89,6 +89,16 @@ class U_I18N_API Measure: public UObject {
      */
     bool operator==(const UObject& other) const;
 
+#ifndef U_HIDE_DRAFT_API
```
markusicu, line 92: I should have thought of this... done now in my PR.

icu4c/source/test/intltest/measfmttest.cpp
```diff
@@ -5341,6 +5341,17 @@ void MeasureFormatTest::TestBadArg() {
 
 void MeasureFormatTest::TestEquality() {
     UErrorCode status = U_ZERO_ERROR;
+
+    // First verify that the equality operators of class Measure work.
```
markusicu, line 5345: I like my test a little better.

roubert: What I find problematic with your test is that it compiles (and runs) just fine, without even the smallest little warning, also with `-std=c++20`, so it does in no way help verifying that the fix actually fixes what it's supposed to.

markusicu: I exercise both operator== and operator!=, just like your test, but with more combinations.
How does your test do better? Is it because you have a `Measure` on the right hand side?

roubert: > Is it because you have a `Measure` on the right hand side?

Yes, that's the important part.

markusicu: Ok. Ideas:
- You could rebase your PR and add your test as well.
- You or I could improve my test code so that it uses both UObject and Measure on the right hand side.

markusicu (approved):
lgtm tnx

icu4c/source/test/intltest/measfmttest.cpp
roubert: Done.

PR #2645, opened 2023-10-02
ICU-22527 add Measure::operator!=()

markusicu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22527

### DIFF
--- a/icu4c/source/i18n/unicode/measure.h
+++ b/icu4c/source/i18n/unicode/measure.h
@@ -89,6 +89,16 @@ class U_I18N_API Measure: public UObject {
      */
     bool operator==(const UObject& other) const;
 
+#ifndef U_HIDE_DRAFT_API
+    /**
+     * Inequality operator.  Returns true if this object is not equal to the other object.
+     * @param other the object to compare with
+     * @return true if the objects are not equal
+     * @draft ICU 74
+     */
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/unicode/measure.h
```diff
@@ -89,6 +89,14 @@ class U_I18N_API Measure: public UObject {
      */
     bool operator==(const UObject& other) const;
 
+    /**
+     * Inequality operator.  Returns true if this object is not equal to the other object.
+     * @param other the object to compare with
+     * @return true if the objects are not equal
+     * @draft ICU 74
+     */
+    inline bool operator!=(const UObject& other) const { return !operator==(other); }
```
eggrobin, line 98: I would tend to write that `!(*this == other)`, but I see that this is consistent with the rest of this file, so LGTM.

eggrobin (approved):

eggrobin (approved):

icu4c/source/i18n/unicode/measure.h
markusicu: It's more explicit this way. We are not consistent in ICU, but we have used this in a number of classes.

PR #2653, opened 2023-10-03
ICU-22528 Improve date formatting performance

mohd-akram:
This improves date formatting performance to be 1.5-2x as fast.

<details>
<summary>Script to generate pattern char lookup table</summary>

```js
const chars = [
  0x47, 0x79, 0x4D, 0x64, 0x6B, 0x48, 0x6D, 0x73, 0x53, 0x45,
  0x44, 0x46, 0x77, 0x57, 0x61, 0x68, 0x4B, 0x7A, 0x59, 0x65,
  0x75, 0x67, 0x41, 0x5A, 0x76, 0x63, 0x4c, 0x51, 0x71, 0x56,
  0x55, 0x4F, 0x58, 0x78, 0x72, 0x62, 0x42,
  0x3a,
];
process.stdout.write(' '.repeat(7));
for (let c = 0; c < 128; c++) {
  if (c % 16 == 0) process.stdout.write('\n'+' '.repeat(7));
  process.stdout.write(String(chars.indexOf(c)).padStart(3)+',');
}
process.stdout.write('\n');
```
</details>

<details>
<summary>Benchmarking program</summary>

```cpp
#include <chrono>
#include <iostream>
#define U_SHOW_CPLUSPLUS_API 1
#include "unicode/smpdtfmt.h"
#include "unicode/ustream.h"

using namespace icu;

int main(int argc, char *argv[])
{
	UErrorCode success = U_ZERO_ERROR;
	auto format = argc > 2 ? argv[2] : "hh:mm:ss a OOOO";
	SimpleDateFormat df(UnicodeString(format), success);
	UnicodeString myString;
	for (int32_t i = 0; i < 1; ++i) {
		myString.remove();
		auto x = df.format(1695464381915, myString);
	}
	int n = argc > 1 ? atoi(argv[1]) : 10000;
	auto start = std::chrono::high_resolution_clock::now();
	for (int32_t i = 0; i < n; ++i) {
		myString.remove();
		auto x = df.format(1695464381915, myString);
	}
	std::cout << myString << "\n";
	auto finish = std::chrono::high_resolution_clock::now();
	auto ms = std::chrono::duration_cast<std::chrono::microseconds>(finish-start).count()/1000.0;
	std::cout << ms << "ms\n";
}
```
</details>

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22528

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -1532,11 +1532,11 @@ void Calendar::computeFields(UErrorCode &ec)
     U_ASSERT(fFields[UCAL_DST_OFFSET] <= getMaximum(UCAL_DST_OFFSET));
 }
 
-uint8_t Calendar::julianDayToDayOfWeek(double julian)
+uint8_t Calendar::julianDayToDayOfWeek(int32_t julian)
 {
     // If julian is negative, then julian%7 will be negative, so we adjust
     // accordingly.  We add 1 because Julian day 0 is Monday.
-    int8_t dayOfWeek = (int8_t) uprv_fmod(julian + 1, 7);
+    int8_t dayOfWeek = (int8_t) ((julian + 1) % 7);
 
     uint8_t result = (uint8_t)(dayOfWeek + ((dayOfWeek < 0) ? (7+UCAL_SUNDAY ) : UCAL_SUNDAY));
     return result;
--- a/icu4c/source/i18n/datefmt.cpp
+++ b/icu4c/source/i18n/datefmt.cpp
@@ -28,6 +28,7 @@
 #include "unicode/smpdtfmt.h"
 #include "unicode/dtptngen.h"
 #include "unicode/udisplaycontext.h"
+#include "unicode/gregocal.h"
 #include "reldtfmt.h"
 #include "sharedobject.h"
 #include "unifiedcache.h"
@@ -277,15 +278,25 @@ DateFormat::format(Calendar& /* unused cal */,
 UnicodeString&
 DateFormat::format(UDate date, UnicodeString& appendTo, FieldPosition& fieldPosition) const {
     if (fCalendar != nullptr) {
-        // Use a clone of our calendar instance
-        Calendar* calClone = fCalendar->clone();
-        if (calClone != nullptr) {
-            UErrorCode ec = U_ZERO_ERROR;
-            calClone->setTime(date, ec);
+        UErrorCode ec = U_ZERO_ERROR;
+        auto calType = fCalendar->getType();
+        // Avoid a heap allocation and corresponding free for the common case
+        if (uprv_strcmp(calType, "gregorian") == 0) {
+            GregorianCalendar cal(*static_cast<GregorianCalendar*>(fCalendar));
+            cal.setTime(date, ec);
             if (U_SUCCESS(ec)) {
-                format(*calClone, appendTo, fieldPosition);
+                format(cal, appendTo, fieldPosition);
             }
-            delete calClone;
+        } else {
+            // Use a clone of our calendar instance
+            Calendar *calClone = fCalendar->clone();
+            if (calClone != nullptr) {
+                calClone->setTime(date, ec);
+                if (U_SUCCESS(ec)) {
+                    format(*calClone, appendTo, fieldPosition);
+                }
+                delete calClone;
+            }
         }
     }
     return appendTo;
@@ -297,13 +308,24 @@ UnicodeString&
 DateFormat::format(UDate date, UnicodeString& appendTo, FieldPositionIterator* posIter,
                    UErrorCode& status) const {
     if (fCalendar != nullptr) {
-        Calendar* calClone = fCalendar->clone();
-        if (calClone != nullptr) {
-            calClone->setTime(date, status);
-            if (U_SUCCESS(status)) {
-               format(*calClone, appendTo, posIter, status);
+        UErrorCode ec = U_ZERO_ERROR;
+        auto calType = fCalendar->getType();
+        // Avoid a heap allocation and corresponding free for the common case
+        if (uprv_strcmp(calType, "gregorian") == 0) {
+            GregorianCalendar cal(*static_cast<GregorianCalendar*>(fCalendar));
+            cal.setTime(date, ec);
+            if (U_SUCCESS(ec)) {
+                format(cal, appendTo, posIter, status);
             }
-            delete calClone;
+        } else {
+            Calendar* calClone = fCalendar->clone();
+            if (calClone != nullptr) {
+                calClone->setTime(date, status);
+                if (U_SUCCESS(status)) {
+                    format(*calClone, appendTo, posIter, status);
+                }
+                delete calClone;
+            }
         }
     }
     return appendTo;
--- a/icu4c/source/i18n/dtfmtsym.cpp
+++ b/icu4c/source/i18n/dtfmtsym.cpp
@@ -86,6 +86,33 @@ static const char16_t gPatternChars[] = {
     0
 };
 
+/**
+ * Map of each ASCII character to its corresponding index in the table above if
+ * it is a pattern character and -1 otherwise.
+ */
+static const int8_t gLookupPatternChars[] = {
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    //
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    //       !   "   #   $   %   &   '   (   )   *   +   ,   -   .   /
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+#if UDAT_HAS_PATTERN_CHAR_FOR_TIME_SEPARATOR
+    //   0   1   2   3   4   5   6   7   8   9   :   ;   <   =   >   ?
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, 37, -1, -1, -1, -1, -1,
+#else
+    //   0   1   2   3   4   5   6   7   8   9   :   ;   <   =   >   ?
+        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+#endif
+    //   @   A   B   C   D   E   F   G   H   I   J   K   L   M   N   O
+        -1, 22, 36, -1, 10,  9, 11,  0,  5, -1, -1, 16, 26,  2, -1, 31,
+    //   P   Q   R   S   T   U   V   W   X   Y   Z   [   \   ]   ^   _
+        -1, 27, -1,  8, -1, 30, 29, 13, 32, 18, 23, -1, -1, -1, -1, -1,
+    //   `   a   b   c   d   e   f   g   h   i   j   k   l   m   n   o
+        -1, 14, 35, 25,  3, 19, -1, 21, 15, -1, -1,  4, -1,  6, -1, -1,
+    //   p   q   r   s   t   u   v   w   x   y   z   {   |   }   ~
+        -1, 28, 34,  7, -1, 20, 24, 12, 33,  1, 17, -1, -1, -1, -1, -1
+};
+
 //------------------------------------------------------
 // Strings of last resort.  These are only used if we have no resource
 // files.  They aren't designed for actual use, just for backup.
@@ -1391,12 +1418,11 @@ DateFormatSymbols::getPatternUChars()
 
 UDateFormatField U_EXPORT2
 DateFormatSymbols::getPatternCharIndex(char16_t c) {
-    const char16_t *p = u_strchr(gPatternChars, c);
-    if (p == nullptr) {
+    if (c >= UPRV_LENGTHOF(gLookupPatternChars)) {
         return UDAT_FIELD_COUNT;
-    } else {
-        return static_cast<UDateFormatField>(p - gPatternChars);
     }
+    const auto idx = gLookupPatternChars[c];
+    return idx == -1 ? UDAT_FIELD_COUNT : static_cast<UDateFormatField>(idx);
 }
 
 static const uint64_t kNumericFieldsAlways =
--- a/icu4c/source/i18n/gregoimp.cpp
+++ b/icu4c/source/i18n/gregoimp.cpp
@@ -32,6 +32,15 @@ int64_t ClockMath::floorDivide(int64_t numerator, int64_t denominator) {
         numerator / denominator : ((numerator + 1) / denominator) - 1;
 }
 
+int32_t ClockMath::floorDivide(int32_t numerator, int32_t denominator,
+                          int32_t* remainder) {
+    auto quotient = floorDivide(numerator, denominator);
+    if (remainder != nullptr) {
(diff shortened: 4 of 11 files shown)


icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -86,6 +86,29 @@ static const char16_t gPatternChars[] = {
...
+    //   @   A   B   C   D   E   F   G   H   I   J   K   L   M   N   O
+        -1, 22, 36, -1, 10,  9, 11,  0,  5, -1, -1, 16, 26,  2, -1, 31,
+    //   P   Q   R   S   T   U   V   W   X   Y   Z   [   \   ]   ^   _
+        -1, 27, -1,  8, -1, 30, 29, 13, 32, 18, 23, -1, -1, -1, -1, -1,
+    //   `   a   b   c   d   e   f   g   h   i   j   k   l   m   n   o
+        -1, 14, 35, 25,  3, 19, -1, 21, 15, -1, -1,  4, -1,  6, -1, -1,
+    //   p   q   r   s   t   u   v   w   x   y   z   {   |   }   ~
+        -1, 28, 34,  7, -1, 20, 24, 12, 33,  1, 17, -1, -1, -1, -1, -1
+};
+
```
richgillam, line 111: I'm worried, again, about the possibility of somebody in the future modifying `gPatternChars` and not knowing they also have to modify this table.  Again, that doesn't seem terribly likely (probably not likely at all), but it worries me a little.

My first response was to ask if you could generate this table on the fly the first time it's used for something, but the more I think about it, the more I think that's probably overkill.  I'd be happy with a comment above this declaration explaining what it is and how it was derived.

icu4c/source/i18n/gregoimp.cpp
```diff
@@ -130,17 +130,15 @@ void Grego::dayToFields(double day, int32_t& year, int32_t& month,
...
+    dow = ((int32_t) day + 1) % 7;
+    dow += UCAL_SUNDAY + 7*(dow < 0);
 
     // Common Julian/Gregorian calculation
-    int32_t correction = 0;
-    int32_t march1 = isLeap ? 60 : 59; // zero-based DOY for March 1
-    if (doy >= march1) {
-        correction = isLeap ? 1 : 2;
-    }
+    int32_t march1 = 59 + isLeap; // zero-based DOY for March 1
```
richgillam, line 137: Not crazy about using `isLeap` as an integer this way; I've seen too many cases where somebody used a value other than 1 for `true`, and I also don't think it's as clear.  Is there really a performance advantage to what you're doing here?

icu4c/source/i18n/gregoimp.cpp
```diff
@@ -130,17 +130,15 @@ void Grego::dayToFields(double day, int32_t& year, int32_t& month,
     UBool isLeap = isLeapYear(year);
     
     // Gregorian day zero is a Monday.
-    dow = (int32_t) uprv_fmod(day + 1, 7);
-    dow += (dow < 0) ? (UCAL_SUNDAY + 7) : UCAL_SUNDAY;
+    dow = ((int32_t) day + 1) % 7;
```
richgillam, line 133: You changed the signature of `julianDayToDayOfWeek()` to reflect the fact that everybody was passing in an `int32_t` anyway, and it simplified the code inside more-- you didn't need all the casts.  It seems like you could do the same thing here.  I didn't do a comprehensive search of all the call sites, but I think I only found one that was using a `double`, and you could do the cast there.  Or am I missing something?

icu4c/source/i18n/gregoimp.cpp
```diff
@@ -130,17 +130,15 @@ void Grego::dayToFields(double day, int32_t& year, int32_t& month,
...
-    int32_t march1 = isLeap ? 60 : 59; // zero-based DOY for March 1
-    if (doy >= march1) {
-        correction = isLeap ? 1 : 2;
-    }
+    int32_t march1 = 59 + isLeap; // zero-based DOY for March 1
+    int32_t correction = doy >= march1 ? 2 - isLeap : 0;
+
     month = (12 * (doy + correction) + 6) / 367; // zero-based month
-    dom = doy - DAYS_BEFORE[month + (isLeap ? 12 : 0)] + 1; // one-based DOM
+    dom = doy - DAYS_BEFORE[month + isLeap*12] + 1; // one-based DOM
```
richgillam, line 141: Please put spaces around the `*`.

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -1074,7 +1074,8 @@ SimpleDateFormat::_format(Calendar& cal, UnicodeString& appendTo,
     UDisplayContext capitalizationContext = getContext(UDISPCTX_TYPE_CAPITALIZATION, status);
 
     // loop through the pattern string character by character
-    for (int32_t i = 0; i < fPattern.length() && U_SUCCESS(status); ++i) {
+    int32_t len = fPattern.length();
```
richgillam, line 1077: Maybe `patternLength` instead of just `len`?

icu4c/source/i18n/smpdtfmt.cpp
```diff
@@ -1480,6 +1481,9 @@ SimpleDateFormat::subFormat(UnicodeString &appendTo,
                             Calendar& cal,
                             UErrorCode& status) const
 {
+    static const int32_t maxIntCount = 10;
+    static const UnicodeString hebr("hebr", 4, US_INV);
```
richgillam, line 1485: Making this declaration `static const` is an improvement, but I wonder if we can get rid of it altogether.  I only see it used in one spot, where we have ` if (fDateOverride.compare(hebr)==0 && value>HEBREW_CAL_CUR_MILLENIUM_START_YEAR ...)`, and I think you could just change this to `if (fDateOverride == u"hebr" && ...)` and avoid creating a `UnicodeString` at all.  (In addition, using `compare()` instead of `operator==()` to test for equality seems questionable.)

icu4c/source/i18n/unicode/formattednumber.h
```diff
@@ -190,6 +191,9 @@ class U_I18N_API FormattedNumber : public UMemory, public FormattedValue {
 
     // To give C API access to internals
     friend struct impl::UFormattedNumberImpl;
+
+    // To give access to the data pointer for non-heap allocation
+    friend class icu::SimpleDateFormat;
```
richgillam, line 196: I'd kind of like @sffc to weigh in on whether he's okay with adding these `friend` declarations to `SimpleNumberFormatter` and `FormattedNumber`.

richgillam (commented):
Generally looks great, but I do have a few small things...

icu4c/source/i18n/gregoimp.cpp
mohd-akram: Okay, apparently `UBool` is actually `int8_t`. That typedef should probably be changed to `bool` to get stronger guarantees based on the C++ spec about the allowed values, as this is a possibly hidden source of bugs. Given the current situation, I'll revert the changes assuming true is 1.

icu4c/source/i18n/dtfmtsym.cpp
mohd-akram: This would be much cleaner with designated initializers but C++ does not support them (the compiler accepts them but I believe `-Werror` is enabled so it fails in the CI). I've found the pattern characters are mentioned in many tables like this one (I got this template from `SimpleDateFormat::getLevelFromChar`). Alternatively, `constexpr` could be used. I'll settle for the comment for now, and just out of curiosity I'll see how feasible it would be to generate it.

icu4c/source/i18n/datefmt.cpp
```diff
@@ -277,15 +278,25 @@ DateFormat::format(Calendar& /* unused cal */,
...
+            }
+        } else {
+            // Use a clone of our calendar instance
+            Calendar *calClone = fCalendar->clone();
+            if (calClone != nullptr) {
+                calClone->setTime(date, ec);
+                if (U_SUCCESS(ec)) {
+                    format(*calClone, appendTo, fieldPosition);
+                }
+                delete calClone;
```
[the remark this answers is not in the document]
mohd-akram: I did it this way in case another common calendar type would be added as well (and that point each if condition could just be a macro generation). But I suppose that bridge can be crossed if/when we get to it.

icu4c/source/i18n/gregoimp.cpp
mohd-akram: I was under the impression that `Grego` was a public API so I refrained from changing the declaration, and there are also a couple other functions in there taking a double which should probably be changed as well. It seems it is not actually a public API so I'll change this too.

icu4c/source/i18n/smpdtfmt.cpp
mohd-akram: This doesn't seem to work:

```c++
smpdtfmt.cpp:1565:27: warning: result of comparison against a string literal is unspecified (use an explicit string comparison function instead) [-Wstring-compare]
        if (fDateOverride == u"hebr" && value>HEBREW_CAL_CUR_MILLENIUM_START_YEAR && value<HEBREW_CAL_CUR_MILLENIUM_END_YEAR) {
                          ^  ~~~~~~~
smpdtfmt.cpp:1565:27: error: invalid operands to binary expression ('const UnicodeString' and 'const char16_t[5]')
        if (fDateOverride == u"hebr" && value>HEBREW_CAL_CUR_MILLENIUM_START_YEAR && value<HEBREW_CAL_CUR_MILLENIUM_END_YEAR) {
            ~~~~~~~~~~~~~ ^  ~~~~~~~
../common/unicode/stringpiece.h:326:1: note: candidate function not viable: no known conversion from 'const UnicodeString' to 'const StringPiece' for 1st argument
operator==(const StringPiece& x, const StringPiece& y);
^
../common/unicode/unistr.h:3978:16: note: candidate function not viable: no known conversion from 'const char16_t[5]' to 'const UnicodeString' for 1st argument
UnicodeString::operator== (const UnicodeString& text) const
```

icu4c/source/i18n/calendar.cpp
```diff
@@ -1532,13 +1532,13 @@ void Calendar::computeFields(UErrorCode &ec)
...
-uint8_t Calendar::julianDayToDayOfWeek(double julian)
+uint8_t Calendar::julianDayToDayOfWeek(int32_t julian)
 {
     // If julian is negative, then julian%7 will be negative, so we adjust
     // accordingly.  We add 1 because Julian day 0 is Monday.
-    int8_t dayOfWeek = (int8_t) uprv_fmod(julian + 1, 7);
+    int8_t dayOfWeek = (int8_t) ((julian + 1) % 7);
 
-    uint8_t result = (uint8_t)(dayOfWeek + ((dayOfWeek < 0) ? (7+UCAL_SUNDAY ) : UCAL_SUNDAY));
+    uint8_t result = (uint8_t)(dayOfWeek + (7*(dayOfWeek < 0)+UCAL_SUNDAY));
```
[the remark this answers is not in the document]
mohd-akram: I've reverted this change with the other bool changes. I'm actually less sure about this one than the other optimizations since here we still do a comparison and a multiplication. This simplifies the calendar changes to removing `fmod` (which had the biggest performance impact) and the unnecessary casts.

icu4c/source/i18n/smpdtfmt.cpp
mohd-akram: Thanks, fixed.

icu4c/source/i18n/calendar.cpp
```diff
@@ -1532,13 +1532,13 @@ void Calendar::computeFields(UErrorCode &ec)
     U_ASSERT(fFields[UCAL_DST_OFFSET] <= getMaximum(UCAL_DST_OFFSET));
 }
 
-uint8_t Calendar::julianDayToDayOfWeek(double julian)
+uint8_t Calendar::julianDayToDayOfWeek(int32_t julian)
```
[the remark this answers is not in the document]
mohd-akram: I wish I knew :D

icu4c/source/i18n/unicode/formattednumber.h
sffc: Fine with me

icu4c/source/i18n/datefmt.cpp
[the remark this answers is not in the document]
richgillam: >I get an error with the suggestion above

Ugh.  No default constructor.  Okay, I'm fine with leaving this the way you have it.  If we find things getting more complicated in the future, we may need to search anew for a way to avoid the duplicated code.

icu4c/source/i18n/calendar.cpp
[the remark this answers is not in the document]
richgillam: That sounds good.

icu4c/source/i18n/dtfmtsym.cpp
richgillam: A simple comment is fine with me.  The more I think about it, the more I think generating the table on the fly isn't worth the trouble, considering the remote likelihood of its contents changing.

icu4c/source/i18n/gregoimp.cpp
richgillam: As I remember it, there were compatibility issues with changing the definition of `UBool` to just be `bool`, but we've started using `bool` internally in new code (I think it's just the API where we're stuck with it).

icu4c/source/i18n/smpdtfmt.cpp
richgillam: I guess that means you have to do `fDateOverride == UnicodeString(u"hebr")`, which is kind of ugly.  You could do that or stick with the static declaration.  Since it's only used in the one spot, I'm inclined to say you should put the declaration inline, but it's up to you.  Regardless, I'd do `UnicodeString(u"hebr")` instead of `UnicodeString("hebr", 4, US_INV)`.  The second version is old and does a conversion-- it dates to long before Unicode string literals were supported in C++ (or in the version of C++ ICU was using, which was always way behind the current one).  I'm pretty sure the constructor that takes a `const UChar*` just wraps the `UnicodeString` object around that storage, with no copying or conversion, and should be fairly fast.

richgillam (dismissed):
I think I had maybe one more suggestion for you, but I'm happy with this the way it is.  Squash your commits (and optionally make the other change), and I'll approve the final version.

icu4c/source/i18n/smpdtfmt.cpp
mohd-akram: Looking at the code it does seem to call `doAppend`:

```cpp
UnicodeString::UnicodeString(const char16_t *text) {
  fUnion.fFields.fLengthAndFlags = kShortString;
  doAppend(text, 0, -1);
}
```

I'll switch to the new style but keep it as `static` to prevent the construction overhead.

richgillam (approved):
I just realized I should have asked you about Java.  We usually try to keep our C++ and Java code as similar as possible.  But these all seem like pretty C++-specific changes, so maybe you don't need to worry about that.

mohd-akram:
Thanks @richgillam for the review!

richgillam:
Do you have merge privileges, or do you need me to merge for you?

mohd-akram:
If you could merge it that would be great as I don't have merge privileges.

richgillam:
Done!  Thank you for the fix!

FrankYFTang:
we recently got some fuzzer detected overflow from these new code :(

PR #2657, opened 2023-10-04
ICU-22324 Remove Ant build's `shared` directory

echeran:
The `icu4j/main/shared` directory appears to be a vestige of the Ant build. Therefore, let's remove it.

##### Checklist


markusicu:
sgtm but I defer to @yumaoka. In particular, will we need the Eclipse settings _somewhere_ for things like the "clean up import statements" task?

mihnita:
> sgtm but I defer to @yumaoka. In particular, will we need the Eclipse settings _somewhere_ for things like the "clean up import statements" task?

I created https://unicode-org.atlassian.net/browse/ICU-22530
Please assign to me.

---

Over the week-end I compared the old settings, and they mostly matched across projects, but not 100%.
We can always go back and get them from history.
The one in `shared` (removed by this PR) is probably not the best to take as a model anyway.
The one in `core` (already deleted) would probably be a better start.

And also reviewed for the latest Eclipse, maybe a way to also have the same for IntelliJ.

mihnita (approved):

yumaoka (approved):
LGTM. Yes, we don't need these files used for ant and ant on Eclipse configuration.

PR #2663, opened 2023-10-06
ICU-22536 Fix ICUServiceThreadTest flakiness

gvictor:
Sometimes getVisibleIDs() method returns a null reference which might happend because of inaccurate concurrent access. This change attempts to fix this ICUServiceThreadTest flakiness.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22536

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/ICUService.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/ICUService.java
@@ -592,13 +592,17 @@ public class ICUService extends ICUNotifier {
                         Factory f = lIter.previous();
                         f.updateVisibleIDs(mutableMap);
                     }
-                    this.idcache = Collections.unmodifiableMap(mutableMap);
+                    // Capture the return value in a local variable.
+                    // Avoids returning an idcache value changed by another thread (could be null after clearCaches()).
+                    Map<String, Factory> result = Collections.unmodifiableMap(mutableMap);
+                    this.idcache = result;
+                    return result;
                 } finally {
                     factoryLock.releaseRead();
                 }
             }
+            return idcache;
         }
-        return idcache;
     }
     private Map<String, Factory> idcache;
 


icu4j/main/core/src/main/java/com/ibm/icu/impl/ICUService.java
```diff
@@ -592,13 +592,15 @@ private Map<String, Factory> getVisibleIDMap() {
                         Factory f = lIter.previous();
                         f.updateVisibleIDs(mutableMap);
                     }
-                    this.idcache = Collections.unmodifiableMap(mutableMap);
+                    Map<String, Factory> result = Collections.unmodifiableMap(mutableMap);
```
yumaoka, line 595: @gvictor Can we resolve this problem by simply changing `private Map<String, Factory> idcache;` to `private volatile Map<String, Factory> idcache;`?

gvictor: I think not. If I understand the ICUServiceThreadTest correctly, the problem is that `clearCaches()` runs after `this.idcache` is set and before `return  idcache`. Making volatile will likely make the problem more discoverable, but doesn't fix it, as the CPU cache value of `this.idcache` is invalidated by `clearCaches()` during that window. The fix here is to store the newly created cache instance in a local stack reference which can never be null.

yumaoka: Thanks for your explanation. It would be nice to add a comment why it makes difference. For meanwhile, I approve this.

yumaoka (approved):

icu4j/main/core/src/main/java/com/ibm/icu/impl/ICUService.java
FrankYFTang: so your fix is return the "result" in that case instead of this.idcache because idcache could be set to null by other thread before the return but the local result will never be a null regardless how other thread setting idcache, right?

gvictor: Frank, it seems to resolve the flaky tests in our internal CI. If you are satisfied with it, can you help merge it? As I don't have the right to submit this.

markusicu, line 595:
```suggestion
                    // Capture the return value in a local variable.
                    // Avoids returning an idcache value changed by another thread (could be null after clearCaches()).
                    Map<String, Factory> result = Collections.unmodifiableMap(mutableMap);
```

markusicu (commented):
@gvictor please
1. accept my suggested comment -- or something like it, trying to satisfy Yoshito's and Frank's questions
2. rebase main -- which hopefully gets the CI unstuck
3. squash to one commit

markusicu:
> @gvictor please
> 
> 1. accept my suggested comment -- or something like it, trying to satisfy Yoshito's and Frank's questions
> 2. rebase main -- which hopefully gets the CI unstuck
> 3. squash to one commit

ping @gvictor waiting for these...
We are going to make a branch for ICU 75 today. If you miss that, then you could redirect this PR to that branch and still make 75.

icu4j/main/core/src/main/java/com/ibm/icu/impl/ICUService.java
gvictor: Done now.

gvictor: running CI now.

markusicu (approved):

PR #2664, opened 2023-10-06
ICU-22537 unbreak build for external break engine / UCONFIG_NO_SERVICE

srl295:
fix broken config guards

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22537

### DIFF
--- a/icu4c/source/common/unicode/rbbi.h
+++ b/icu4c/source/common/unicode/rbbi.h
@@ -45,7 +45,6 @@ class  UStack;
 
 
 #ifndef U_HIDE_DRAFT_API
-#if !UCONFIG_NO_SERVICE
 /**
  * The ExternalBreakEngine class define an abstract interface for the host environment
  * to provide a low level facility to break text for unicode text in script that the text boundary
@@ -104,7 +103,6 @@ class ExternalBreakEngine : public UObject {
                                int32_t* foundBreaks, int32_t foundBreaksCapacity,
                                UErrorCode& status) const = 0;
 };
-#endif  /* UCONFIG_NO_SERVICE */
 #endif  /* U_HIDE_DRAFT_API */
 
 
@@ -802,7 +800,6 @@ private:
 #endif  /* U_HIDE_INTERNAL_API */
 
 #ifndef U_HIDE_DRAFT_API
-#if !UCONFIG_NO_SERVICE
     /**
      * Register a new external break engine. The external break engine will be adopted.
      * Because ICU may choose to cache break engine internally, this must
@@ -814,7 +811,6 @@ private:
      */
     static void U_EXPORT2 registerExternalBreakEngine(
                   ExternalBreakEngine* toAdopt, UErrorCode& status);
-#endif /* UCONFIG_NO_SERVICE */
 #endif  /* U_HIDE_DRAFT_API */
 
 };


FrankYFTang:
Why do we have a empty "icu4c/source/common/rbbicst.pl"

icu4c/source/common/brkeng.cpp
```diff
@@ -301,6 +302,7 @@ ICULanguageBreakFactory::loadDictionaryMatcherFor(UScriptCode script) {
 }
 
 
+#if !UCONFIG_NO_SERVICE
```
markusicu, line 305: The new ExternalBreakEngine uses a simpler "registration" model than the older "service" code. Why does it need to be under the same config switch?

icu4c/source/common/brkeng.h
```diff
@@ -286,13 +288,15 @@ class ICULanguageBreakFactory : public LanguageBreakFactory {
   */
   virtual const LanguageBreakEngine *getEngineFor(UChar32 c, const char* locale) override;
 
+#if !UCONFIG_NO_SERVICE
```
markusicu, line 291: In public header files / class definitions, I don't think we can conditionally hide a virtual method -- that would make the vtable incompatible.
But I guess it's ok in an internal header file, since with such a switch one would build a different library anyway?

markusicu (commented):
I thought that we had CI checks for all of the UCONFIG_... testing?!

icu4c/source/common/brkeng.h
srl295: i guess https://github.com/unicode-org/icu/blob/maint/maint-74/icu4c/source/common/unicode/rbbi.h#L805-L815 is a static.

icu4c/source/common/brkeng.cpp
srl295: because https://github.com/unicode-org/icu/blob/maint/maint-74/icu4c/source/common/unicode/rbbi.h#L805-L815 is under a config switch

srl295:
> Why do we have a empty "icu4c/source/common/rbbicst.pl"

bug

srl295:
Probably best if someone can take this over

icu4c/source/common/brkeng.cpp
FrankYFTang: I think we should remove the !UCONFIG_NO_SERVICE around the  https://github.com/unicode-org/icu/blob/maint/maint-74/icu4c/source/common/unicode/rbbi.h#L805-L815  instead

srl295:
by the way, Node.js builds with this branch.

icu4c/source/common/brkeng.cpp
srl295: Take a look, how do you like this now?

FrankYFTang (approved):

FrankYFTang:
Thanks for working on this.

srl295:
Is this the right place to merge, into the maint branch?

srl295:
ticket's not accepted yet

markusicu:
> Is this the right place to merge, into the maint branch?

yes

markusicu (approved):

srl295:
I confirmed that with this update, Node.js main builds.  @targos @ryzokuken

PR #2669, opened 2023-10-11
ICU-22541 Fix MacOS 14 default timezone issue

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22541

### DIFF
--- a/icu4c/source/common/putil.cpp
+++ b/icu4c/source/common/putil.cpp
@@ -1175,6 +1175,21 @@ uprv_tzname(int n)
         if (ret != nullptr && uprv_strcmp(TZDEFAULT, gTimeZoneBuffer) != 0) {
             int32_t tzZoneInfoTailLen = uprv_strlen(TZZONEINFOTAIL);
             const char *tzZoneInfoTailPtr = uprv_strstr(gTimeZoneBuffer, TZZONEINFOTAIL);
+            // MacOS14 has the realpath as something like
+            // /usr/share/zoneinfo.default/Australia/Melbourne
+            // which will not have "/zoneinfo/" in the path.
+            // Therefore if we fail, we fall back to read the link which is
+            // /var/db/timezone/zoneinfo/Australia/Melbourne
+            // We also fall back to reading the link if the realpath leads to something like
+            // /usr/share/zoneinfo/posixrules
+            if (tzZoneInfoTailPtr == nullptr ||
+                    uprv_strcmp(tzZoneInfoTailPtr + tzZoneInfoTailLen, "posixrules") == 0) {
+                ssize_t size = readlink(TZDEFAULT, gTimeZoneBuffer, sizeof(gTimeZoneBuffer)-1);
+                if (size > 0) {
+                    gTimeZoneBuffer[size] = 0;
+                    tzZoneInfoTailPtr = uprv_strstr(gTimeZoneBuffer, TZZONEINFOTAIL);
+                }
+            }
             if (tzZoneInfoTailPtr != nullptr) {
                 tzZoneInfoTailPtr += tzZoneInfoTailLen;
                 skipZoneIDPrefix(&tzZoneInfoTailPtr);
--- a/icu4c/source/test/depstest/dependencies.txt
+++ b/icu4c/source/test/depstest/dependencies.txt
@@ -22,7 +22,7 @@ system_symbols:
     exp_and_tanhf
     stdlib_qsort
     system_locale
-    stdio_input stdio_output file_io realpath_function dir_io mmap_functions dlfcn
+    stdio_input stdio_output file_io dir_io mmap_functions dlfcn
     # C++
     cplusplus iostream
     std_mutex
@@ -110,11 +110,8 @@ group: file_io
     # Additional symbols in an optimized build.
     __xstat
 
-group: realpath_function
-    realpath  # putil.cpp uprv_tzname() calls this in a hack to get the time zone name
-
 group: dir_io
-    opendir closedir readdir  # for a hack to get the time zone name
+    opendir closedir readdir readlink realpath # for a hack to get the time zone name
 
 group: mmap_functions  # for memory-mapped data loading
     mmap munmap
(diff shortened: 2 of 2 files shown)


FrankYFTang:
@Robbos @syg

icu4c/source/common/putil.cpp
```diff
@@ -1175,6 +1175,18 @@ uprv_tzname(int n)
         if (ret != nullptr && uprv_strcmp(TZDEFAULT, gTimeZoneBuffer) != 0) {
             int32_t tzZoneInfoTailLen = uprv_strlen(TZZONEINFOTAIL);
             const char *tzZoneInfoTailPtr = uprv_strstr(gTimeZoneBuffer, TZZONEINFOTAIL);
+            // MacOS14 has the realpath as something like
+            // /usr/share/zoneinfo.default/Australia/Melbourne
+            // which will not have "/zoneinfo/" in the path.
+            // Therefore if we fail, we fall back to read the link which is
+            // /var/db/timezone/zoneinfo/Australia/Melbourne
+            if (tzZoneInfoTailPtr == nullptr) {
+                ssize_t size = readlink(TZDEFAULT, gTimeZoneBuffer, sizeof(gTimeZoneBuffer)-1);
```
markusicu, line 1184: In other words, if the new code doesn't give us a ".../zoneinfo/..." string, then we try the old code from before PR #2323. Makes sense to me.

Alternatively, I suppose we could have looked for "zoneinfo.default" as an alternate "tail". No idea if that would be better or worse.

Apple people: Any concerns?

markusicu (approved):
lgtm

@richgillam @pedberg-icu et al. ok?

srl295:
Sigh, if only tzdb would have agreed (I asked) to include an actual ID in the file.  Instead, we deal with randomness such as `../posixrules`

srl295 (approved):

icu4c/source/common/putil.cpp
srl295: I think the fallback seems reasonable.

syg:
This might not be the right place to discuss: this won't fix the Ubuntu 18.04 breakage we saw with Chrome, which was due to /etc/localtime -> /usr/share/zoneinfo/America/New_York -> /usr/share/zoneinfo/posixrules. (To compare, my Linux box (Google's corp Linux distro) has /usr/share/zoneinfo/posixrules -> /usr/share/zoneinfo/America/New_York instead.)

Are you planning to fix that breakage for 74 as well?

srl295:
> This might not be the right place to discuss: this won't fix the Ubuntu 18.04 breakage we saw with Chrome, 

it seems like it is the right place to discuss since the ticket claims to be about that specific breakage

markusicu:
> does this fix the Ubuntu 18.04 breakage we saw with Chrome, which was due to /etc/localtime -> /usr/share/zoneinfo/America/New_York -> /usr/share/zoneinfo/posixrules?

I think not, because we should see tzZoneInfoTailPtr != nullptr.
We should be able to fix that by changing the check to
```
if (tzZoneInfoTailPtr == nullptr ||
        uprv_strcmp(tzZoneInfoTailPtr + tzZoneInfoTailLen, "posixrules") == 0) {
```

icu4c/source/common/putil.cpp
```diff
@@ -1175,6 +1175,18 @@ uprv_tzname(int n)
         if (ret != nullptr && uprv_strcmp(TZDEFAULT, gTimeZoneBuffer) != 0) {
             int32_t tzZoneInfoTailLen = uprv_strlen(TZZONEINFOTAIL);
             const char *tzZoneInfoTailPtr = uprv_strstr(gTimeZoneBuffer, TZZONEINFOTAIL);
+            // MacOS14 has the realpath as something like
+            // /usr/share/zoneinfo.default/Australia/Melbourne
+            // which will not have "/zoneinfo/" in the path.
+            // Therefore if we fail, we fall back to read the link which is
+            // /var/db/timezone/zoneinfo/Australia/Melbourne
+            if (tzZoneInfoTailPtr == nullptr) {
```
markusicu, line 1183:
```suggestion
            // MacOS14 has the realpath as something like
            // /usr/share/zoneinfo.default/Australia/Melbourne
            // which will not have "/zoneinfo/" in the path.
            // Therefore if we fail, we fall back to read the link which is
            // /var/db/timezone/zoneinfo/Australia/Melbourne
            // We also fall back to reading the link if the realpath leads to something like
            // /usr/share/zoneinfo/posixrules
            if (tzZoneInfoTailPtr == nullptr ||
                    uprv_strcmp(tzZoneInfoTailPtr + tzZoneInfoTailLen, "posixrules") == 0) {
```

richgillam (dismissed):
I think this looks okay, but I'm interested in what Debbie Goldsmith has to say in the email discussion that accompanied this PR.  From talking to Peter, it sounds possible that we're doing something weird in the way we write the time zone files to the disk, and that therefore the issue might not be with ICU so much as with our own time-zone-updating code.

This is all stuff I personally know diddly-squat about, but with Peter leaving, I guess I'll have to learn.  But I think this is probably okay, at least for now.

markusicu:
Test failures. Please change icu4c/source/test/depstest/dependencies.txt:
- line 25: remove realpath_function
- lines 113..115: remove the realpath_function group
- line 117: add `readlink realpath` after readdir

icu4c/source/common/putil.cpp
markusicu: You applied this change... and then you reverted it in your force-push: https://github.com/unicode-org/icu/compare/unicode-org:7d58692..unicode-org:dad7f7c
Why?

richgillam:
FWIW, I added commentary to the JIRA ticket explaining the underlying cause and recommended solution for this issue.

richgillam (approved):
I can't speak to Markus's comments, but Frank's logic here looks correct to me.  Based on my investigation of the code (see my comments in the JIRA ticket), I think that algorithm will work perfectly well as a long-term solution.

markusicu (changes requested):
Please un-revert the "posixrules" change that you accepted yesterday and reverted today.

FrankYFTang:
opps.. I commit it on github, and then forget to pull into my local branch, and then I added the dependecy stuff and force push, causing the revert. Sorry.

markusicu (approved):
thanks!

PR #2670, opened 2023-10-12
ICU-22505 Ensure default TZ remains unchanged by each test

echeran:
Summary:
1. We extended the test framework base class so that we can automatically check that every test restores the default time zone to what it was when it began (both JDK and ICU TZ)
2. @mihnita checked, and the behavior of JDK `TimeZone.getDefault()` always checks the latest value, but ICU `TimeZone.getDefault()` gets whatever is cached.
3. All of the tests had already been properly restoring the default time zone. So what we think is that somewhere in the main (aka runtime) code is setting the ICU TimeZone, and it's throwing off our assumptions that the test code starts off with a proper and undisturbed state.
4. To counteract the problem, we are setting the ICU default TimeZone before every test to be the same default JDK TimeZone that is used for our testing purposes. Even though it might mask the source of the underlying problem, perhaps in our main code, it gets our tests working and enabled again.
5. This buys us time for ICU 75 to do a proper investigation and fix for whatever the source of the underlying problem is.
6. While we were at it, because we could, we also added checks to ensure that tests aren't changing the default locale (both ICU and JDK). It turns out that a couple didn't restore the default locale, so we fixed that, even though it wasn't causing test failures yet.

As @mihnita mentioned in the ICU-TC meeting on Thursday, the key to reproducing the problem was when he realized that there are 2 different systems for setting the default time zone on Linux, and Debian and Ubuntu differ on which one they prefer in their desktop UIs. He wrote:

> This is the "official" way if you search for changing time zone on Debian:
> `sudo timedatectl set-timezone "America/Los_Angeles"`
> Does nothing for Java. Seems to be in sync with the changes you do in UI.
> 
> ---
> The Ubuntu way (also works on Debian), and affects the Java behavior:
> `sudo dpkg-reconfigure tzdata`

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22505

### DIFF
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberParserTest.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberParserTest.java
@@ -2,10 +2,8 @@
 // License & terms of use: http://www.unicode.org/copyright.html
 package com.ibm.icu.dev.test.number;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-
+import com.ibm.icu.dev.test.CoreTestFmwk;
+import org.junit.Assert;
 import org.junit.Test;
 
 import com.ibm.icu.impl.StringSegment;
@@ -30,7 +28,7 @@ import com.ibm.icu.util.ULocale;
  * @author sffc
  *
  */
-public class NumberParserTest {
+public class NumberParserTest extends CoreTestFmwk {
     @Test
     public void testBasic() {
         Object[][] cases = new Object[][] {
@@ -199,13 +197,13 @@ public class NumberParserTest {
 
         ParsedNumber resultObject = new ParsedNumber();
         parser.parse("epäluku", false, resultObject);
-        assertTrue(resultObject.success());
-        assertEquals(Double.NaN, resultObject.getNumber().doubleValue(), 0.0);
+        assertTrue("Parsing locale fi spellout number", resultObject.success());
+        Assert.assertEquals(Double.NaN, resultObject.getNumber().doubleValue(), 0.0);
 
         resultObject = new ParsedNumber();
         parser.parse("1,2e3", false, resultObject);
-        assertTrue(resultObject.success());
-        assertEquals(1200.0, resultObject.getNumber().doubleValue(), 0.0);
+        assertTrue("Parsing formatted compact decimal", resultObject.success());
+        Assert.assertEquals(1200.0, resultObject.getNumber().doubleValue(), 0.0);
     }
 
     @Test
@@ -219,9 +217,9 @@ public class NumberParserTest {
         series.addMatcher(IgnorablesMatcher.getInstance(0));
         series.freeze();
 
-        assertFalse(series.smokeTest(new StringSegment("x", false)));
-        assertFalse(series.smokeTest(new StringSegment("-", false)));
-        assertTrue(series.smokeTest(new StringSegment("+", false)));
+        Assert.assertFalse(series.smokeTest(new StringSegment("x", false)));
+        Assert.assertFalse(series.smokeTest(new StringSegment("-", false)));
+        Assert.assertTrue(series.smokeTest(new StringSegment("+", false)));
 
         Object[][] cases = new Object[][] {
                 { "", 0, true },
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/CoreTestFmwk.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/CoreTestFmwk.java
@@ -1,0 +1,85 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+package com.ibm.icu.dev.test;
+
+import com.ibm.icu.util.ULocale;
+import org.junit.Rule;
+import org.junit.rules.TestName;
+
+/**
+ * A base class for testing within the cross-component tests in the `common_tests` component.
+ *
+ * <p>This subclass of TestFmwk allows the usage of functionality from the `core` component. TestFmwk
+ * is not able to depend on `core` since it is located in `framework`, while `core` tests already
+ * use TestFmwk, which requires that `core` depends on `framework`. We cannot have a cycle in our
+ * dependencies.
+ *
+ * <p>With the allowance for this test class to use functionality from components, we can do things
+ * like assert whether tests are properly resetting the ICU and JDK time zones to their original
+ * values at the end of their execution.
+ */
+public class CoreTestFmwk extends TestFmwk {
+
+  private com.ibm.icu.util.TimeZone testStartDefaultIcuTz;
+
+  private java.util.TimeZone testStartDefaultJdkTz;
+
+  private com.ibm.icu.util.ULocale testStartDefaultULocale;
+
+  private java.util.Locale testStartDefaultLocale;
+
+  @Rule
+  public TestName name = new TestName();
+
+  @Override
+  public void localTestInitialize() {
+    super.localTestInitialize();
+
+    // Just like TestFmwk initializes JDK TimeZone and Locale before every test,
+    // do the same for ICU TimeZone and ULocale.
+    ULocale.setDefault(ULocale.forLocale(defaultLocale));
+    com.ibm.icu.util.TimeZone.setDefault(
+        com.ibm.icu.util.TimeZone.getTimeZone(defaultTimeZone.getID()));
+
+    // Save starting timezones
+    testStartDefaultIcuTz = com.ibm.icu.util.TimeZone.getDefault();
+    testStartDefaultJdkTz = java.util.TimeZone.getDefault();
+
+    // Save starting locales
+    testStartDefaultULocale = com.ibm.icu.util.ULocale.getDefault();
+    testStartDefaultLocale = java.util.Locale.getDefault();
+  }
+
+  @Override
+  public void localTestTeardown() {
+    String testMethodName = name.getMethodName();
+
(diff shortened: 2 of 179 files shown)


icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberParserTest.java
```diff
@@ -199,13 +197,13 @@ public void testLocaleFi() {
 
         ParsedNumber resultObject = new ParsedNumber();
         parser.parse("epäluku", false, resultObject);
-        assertTrue(resultObject.success());
-        assertEquals(Double.NaN, resultObject.getNumber().doubleValue(), 0.0);
+        assertTrue("Parsing locale fi spellout number", resultObject.success());
+        Assert.assertEquals(Double.NaN, resultObject.getNumber().doubleValue(), 0.0);
```
mihnita, line 201: Inconsistent style: `assertTrue` but `Assert.assertEquals`.
Either everything is static import, or nothing is.

---

This should maybe be a bigger cleanup, because `TestFmwk` defines some asserts (for example `assertTrue`).
So when you see `assertTrue(...)` it is hard to tell if it is the `TestFmwk` one or the `JUnit` one. (why should we even have one in `TestFmwk`)

mihnita (commented):
Reviewed. Thanks.

mihnita:
It looks like this PR only adds the detection, but does not fix anything
(I don't see test classes updated to restore a "damaged" default timezone).

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/CoreTestFmwk.java
```diff
@@ -0,0 +1,64 @@
...
+ * This subclass of TestFmwk allows the usage of functionality from the `core` component. TestFmwk
+ * is not able to depend on `core` since it is located in `framework`, while `core` tests already
+ * use TestFmwk, which requires that `core` depends on `framework`. We cannot have a cycle in our
+ * dependencies.
+ *
+ * With the allowance for this test class to use functionality from components, we can do things
+ * like assert whether tests are properly resetting the ICU and JDK time zones to their original
+ * values at the end of their execution.
+ */
+public class CoreTestFmwk extends TestFmwk {
```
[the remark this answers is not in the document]
echeran: Done.

mihnita (dismissed):
Thank you very much!
Sorry for the delay, it was a lot bigger than I thought :-)

echeran:
@markusicu @yumaoka @roubert @hugovdm I'll wait a little bit before squashing so you can have a chance to review commit by commit, if you want. Despite the number of files touched in this PR, I don't think much of it is interesting beyond `CoreTestFmwk.java`, maybe `TestFmwk.java`, and this commit https://github.com/unicode-org/icu/pull/2670/commits/7f8cf220e29b78a1f65ea840f12836b3bae71e65 that re-enables tests that were previously failing due to the mismatch in the default JDK timezone.

Summary:
1. We extended the test framework base class so that we can automatically check that every test restores the default time zone to what it was when it began (both JDK and ICU TZ)
2. @mihnita checked, and the behavior of JDK `TimeZone.getDefault()` always checks the latest value, but ICU `TimeZone.getDefault()` gets whatever is cached.
3. All of the tests had already been properly restoring the default time zone. So what we think is that somewhere in the main (aka runtime) code is setting the ICU TimeZone, and it's throwing off our assumptions that the test code starts off with a proper and undisturbed state.
4. To counteract the problem, we are setting the ICU default TimeZone before every test to be the same default JDK TimeZone that is used for our testing purposes. Even though it might mask the source of the underlying problem, perhaps in our main code, it gets our tests working and enabled again.
5. This buys us time for ICU 75 to do a proper investigation and fix for whatever the source of the underlying problem is.
6. While we were at it, because we could, we also added checks to ensure that tests aren't changing the default locale (both ICU and JDK). It turns out that a couple didn't restore the default locale, so we fixed that, even though it wasn't causing test failures yet.

markusicu:
> Summary:

Please hoist this summary up to the PR description for easier discovery later.

> 5. This buys us time for ICU 75 to do a proper investigation and fix for whatever the source of the underlying problem is.

Ticket?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/CoreTestFmwk.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * This subclass of TestFmwk allows the usage of functionality from the `core` component. TestFmwk
+ * is not able to depend on `core` since it is located in `framework`, while `core` tests already
+ * use TestFmwk, which requires that `core` depends on `framework`. We cannot have a cycle in our
+ * dependencies.
+ *
+ * With the allowance for this test class to use functionality from components, we can do things
+ * like assert whether tests are properly resetting the ICU and JDK time zones to their original
+ * values at the end of their execution.
+ */
```
markusicu, line 21: nit
```suggestion
/**
 * A base class for testing within the cross-component tests in the `common_tests` component.
 *
 * <p>This subclass of TestFmwk allows the usage of functionality from the `core` component. TestFmwk
 * is not able to depend on `core` since it is located in `framework`, while `core` tests already
 * use TestFmwk, which requires that `core` depends on `framework`. We cannot have a cycle in our
 * dependencies.
 *
 * <p>With the allowance for this test class to use functionality from components, we can do things
 * like assert whether tests are properly resetting the ICU and JDK time zones to their original
 * values at the end of their execution.
 */
```

icu4j/main/framework/src/test/java/com/ibm/icu/dev/test/TestFmwk.java
```diff
@@ -86,6 +86,9 @@ public final void testInitialize() {
      * Because JUnit does not guarantee the order of multiple Before
      * methods, TestFmwk implementation class should override this
      * method, instead of annotating Before.
+     *
+     * Any implementation of the override should be sure to call
+     * super.localTestInitialize().
```
markusicu, line 91: nit
```suggestion
     *
     * <p>Any implementation of the override should be sure to call
     * super.localTestInitialize().
```

icu4j/main/framework/src/test/java/com/ibm/icu/dev/test/TestFmwk.java
```diff
@@ -94,6 +97,9 @@ protected void localTestInitialize() {
      * This method is called at the beginning of {@link #testTeardown()}.
      * TestFmwk implementation class should override this method, instead
      * of annotating After.
+     *
+     * Any implementation of the override should be sure to call
+     * super.localTestTeardown().
```
markusicu, line 102: nit
```suggestion
     *
     * <p>Any implementation of the override should be sure to call
     * super.localTestTeardown().
```

markusicu (commented):
@echeran @mihnita praise: very nice piece of work! thank you!

roubert:
I'm sorry to say, but I just fetched the echeran/ICU-22505 branch into my own repository and ran `mvn test` and it still failed in _common_tests_, just like it does at HEAD. The default timezone on my machine is `Europe/Zurich` and it's running current Debian _testing_.

markusicu:
> I'm sorry to say, but I just fetched the echeran/[ICU-22505](https://unicode-org.atlassian.net/browse/ICU-22505) branch into my own repository and ran `mvn test` and it still failed in _common_tests_, just like it does at HEAD. The default timezone on my machine is `Europe/Zurich` and it's running current Debian _testing_.

This is good... for locally reproducing! Can you load it into an IDE and debug and tell us which test fails why?

hugovdm:
Tests are also passing on my machine on commit 0af450be189. :+1: 
(I did need an "mvn clean", interestingly - due to my previous build from my own branch.)

hugovdm (commented):
Tests are also passing on my machine on commit 0af450be189. :+1: 
(I did need an "mvn clean", interestingly - due to my previous build from my own branch.)

echeran:
> Tests are also passing on my machine on commit [0af450b](https://github.com/unicode-org/icu/commit/0af450be189edc3aefbe5838e64d4758c9b47e7b). 👍 (I did need an "mvn clean", interestingly - due to my previous build from my own branch.)

Great to hear, and thanks for the info on needing to use `mvn clean`. @mihnita and I were eventually able to reproduce the problem and then see that the PR fixes it locally, and since we are using the same setup as @roubert, your info might help us diagnose what the discrepancy is.

echeran:
> Please hoist this summary up to the PR description for easier discovery later.

Done.

> Ticket?

https://unicode-org.atlassian.net/browse/ICU-22550

roubert:
> This is good... for locally reproducing! Can you load it into an IDE and debug and tell us which test fails why?

I've now taken a closer look at what exactly fails in _common_tests_ for me and it seems like it's unrelated to timezones and this PR, the failure is in `TestDataDrivenJDK` and I can reproduce it both at HEAD and at this PR like this:

```
$ mvn test -Dtest='com.ibm.icu.dev.test.format.NumberFormatDataDrivenTest#TestDataDrivenJDK' -Dsurefire.failIfNoSpecifiedTests=false
```

If I delete that test case then `mvn test` finishes successfully on my machine, both at HEAD and at this PR.

markusicu:
> I've now taken a closer look at what exactly fails in _common_tests_ for me and it seems like it's unrelated to timezones and this PR, the failure is in `TestDataDrivenJDK` and I can reproduce it both at HEAD and at this PR like this:
> ...
> If I delete that test case then `mvn test` finishes successfully on my machine, both at HEAD and at this PR.

Good to know!
This test is known to be sensitive to changes in JDK behavior, and depends on using or avoiding certain kinds of JDKs.
It sounds like this PR is good to go then.
Thanks!

markusicu (approved):
lgtm pse squash

PR #2672, opened 2023-10-13
ICU-22548 Reduce the loop limit to fail faster

FrankYFTang:
Reduce the error return from about 100s to 30s to avoid fuzzer 60s timeout error.
Also add a include file to fix uint8_t build breakage.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22548

### DIFF
--- a/icu4c/source/i18n/collationbuilder.cpp
+++ b/icu4c/source/i18n/collationbuilder.cpp
@@ -1119,9 +1119,9 @@ CollationBuilder::addWithClosure(const UnicodeString &nfdPrefix, const UnicodeSt
 // Please let us know if you have a reasonable use case that needed
 // for a practical Collation rule that needs to increase this limit.
 // This value is needed for compiling a rule with eight Hangul syllables such as
-// "&a=b쫊쫊쫊쫊쫊쫊쫊쫊" without error, which should be more than realistic
+// "&a=b쫊쫊쫊쫊쫊쫊쫊" without error, which should be more than realistic
 // usage.
-static constexpr int32_t kClosureLoopLimit = 6560;
+static constexpr int32_t kClosureLoopLimit = 3000;
 
 uint32_t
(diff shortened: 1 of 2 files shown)


icu4c/source/test/fuzzer/locale_util.h
```diff
@@ -5,6 +5,7 @@
 #define I18N_ICU_FUZZ_LOCALE_UTIL_H_
 
 #include <string>
+#include "unicode/ptypes.h"
```
markusicu, line 8: Please just include unicode/utypes.h (not ptypes.h) as usual. It includes ptypes.h indirectly.

icu4c/source/i18n/collationbuilder.cpp
```diff
@@ -1119,9 +1119,9 @@ CollationBuilder::addWithClosure(const UnicodeString &nfdPrefix, const UnicodeSt
 // Please let us know if you have a reasonable use case that needed
 // for a practical Collation rule that needs to increase this limit.
 // This value is needed for compiling a rule with eight Hangul syllables such as
-// "&a=b쫊쫊쫊쫊쫊쫊쫊쫊" without error, which should be more than realistic
+// "&a=b쫊쫊쫊쫊쫊쫊쫊" without error, which should be more than realistic
 // usage.
-static constexpr int32_t kClosureLoopLimit = 6560;
+static constexpr int32_t kClosureLoopLimit = 3000;
```
markusicu, line 1124: I assume that this is still enough to cover the earlier unit test for the internal builder overrun?

FrankYFTang: yes, I tried that.

markusicu (approved):

FrankYFTang:
somehow this merge cause https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63289

FrankYFTang:
The build problem in fuzzer is because somehow there are own build script which need to add an include path . See https://github.com/google/oss-fuzz/pull/11132

PR #2673, opened 2023-10-17
ICU-22511 Fix "vi" locale collator hang in comparison

FrankYFTang:
Avoid infinity loop by terminating and returning error if we loop the same for too many times.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22511

### DIFF
--- a/icu4c/source/i18n/collationcompare.cpp
+++ b/icu4c/source/i18n/collationcompare.cpp
@@ -41,6 +41,8 @@ CollationCompare::compareUpToQuaternary(CollationIterator &left, CollationIterat
     }
     UBool anyVariable = false;
 
+    uint32_t lastPrimary = 0;
+    uint32_t samePrimaryCount = 0;
     // Fetch CEs, compare primaries, store secondary & tertiary weights.
     for(;;) {
         // We fetch CEs until we get a non-ignorable primary or reach the end.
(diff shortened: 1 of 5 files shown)


FrankYFTang:
@markusicu friendly ping. this bug was filed Nov 2, 2023 by fuzzer and and the PR is ready for review for a week.

macchiati:
I think both are out of office for Thanksgiving week.

macchiati:
I took a quick look at the code, and I am concerned. It appears to return as equal whenever three times in a row, a primary is equal. But then we would get the clearly incorrect:

aaah = aaagh

Each pass through the main loop should be resetting either the right or left CEs, or both. Because getting a ce advances an internal pointer, that should always terminate. So something fishy is going on.

I suggest that you augment your printout to provide more readable information, since lines like the following are too hard to follow:
ce from left 0xa4000500
ce from left 0x2a00000005009c00
...

I suggest when you print, break down the left and right ce values in to primary, secondary, tertiary, and put tabs between the fields  so that this can be loaded into a spreadsheet.
Eg something like:
...
Lp 2a0000 Ls 0500 Lt 0042
...

I can't remember, but you make also be able to fetch a code point offset from the ce; if so it would help to include that in the printout.

FrankYFTang:
sorry, I assume you look at the Java code first. My java change is indeed wrong. I should throw there not return equal.

macchiati:
Thanks. I'm still a bit worried about the change, because of the aaah case, so I'd appreciate a printout for the test case along the lines I suggested.

richgillam:
I looked at this and didn't understand what it was doing or why.  I think I'll just defer to Mark here.

FrankYFTang:
@markusicu ping

FrankYFTang:
ping @markusicu . this is more urgent than other PR because it is now able to hang v8 with very simple script (see the bug for details)

FrankYFTang:
Somehow the bug is fixed between 
https://github.com/unicode-org/icu/compare/788b89321454f3eeb20782032c8444aabba42417...699fb1dbc4cfbae6f78ff0b28570f44a20a7b149

per fuzzer. Obsolete this PR

FrankYFTang:
reopen 
ok, the fuzzer is verify the fixed just because we increase the number of supported locale and change the testing locale to a different one. If we try the tests in this PR it is still broken.

FrankYFTang:
The fix is not good enough.

PR #2680, opened 2023-10-25
ICU-22513 Return error if days is too large in IslamicUmalquraCalendar

FrankYFTang:
The computation in IslamicUmalquraCalendar is slower than others, if the days is more than 10,000 years from the epoch, we just return error to avoid hang.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22513

### DIFF
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/CalendarRegressionTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/CalendarRegressionTest.java
@@ -2710,5 +2710,17 @@ public class CalendarRegressionTest extends com.ibm.icu.dev.test.TestFmwk {
                     Calendar.getInstance(Locale.forLanguageTag(localeIds[i])).getFirstDayOfWeek());
         }
     }
+
+    @Test
+    public void TestIslamicUmalquraCalendarSlow() { // ICU-22513
+        Locale loc = new Locale("th@calendar=islamic-umalqura");
+        Calendar cal = Calendar.getInstance(loc);
+        cal.clear();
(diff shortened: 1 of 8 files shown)


icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/CalendarRegressionTest.java
```diff
@@ -2710,5 +2710,17 @@ public void TestRespectUExtensionFw() { // ICU-22226
...
     }
+
+    @Test
+    public void TestIslamicUmalquraCalendarSlow() { // ICU-22513
+        Locale loc = new Locale("th@calendar=islamic-umalqura");
+        Calendar cal = Calendar.getInstance(loc);
+        cal.clear();
+        cal.add(Calendar.YEAR, 1229080905);
+        cal.roll(Calendar.WEEK_OF_MONTH, 1499050699);
+        cal.fieldDifference(new Date(0), Calendar.YEAR_WOY);
```
richgillam, line 2721: Maybe I'm missing something, but shouldn't there be some kind of "expect exception to be thrown" logic in this test?

richgillam (approved):
Makes sense to me, but one question about the Java unit test.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/CalendarRegressionTest.java
FrankYFTang: Java use long (64 bits) for those value so it will not throw exception with that setting. The test just make sure it will not be infinity loop or time out.

PR #2692, opened 2023-10-28
ICU-22545 Fix addLikelySubtags for pseudo Locales

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22545

### DIFF
--- a/icu4c/source/common/loclikelysubtags.cpp
+++ b/icu4c/source/common/loclikelysubtags.cpp
@@ -561,12 +561,21 @@ LSR XLikelySubtags::makeMaximizedLsr(const char *language, const char *script, c
     if (region[0] == 'X' && (c1 = region[1]) != 0 && region[2] == 0) {
         switch (c1) {
         case 'A':
+            if (returnInputIfUnmatch) {
+                return LSR(language, script, region, LSR::EXPLICIT_LSR);
+            }
             return LSR(PSEUDO_ACCENTS_PREFIX, language, script, region,
                        LSR::EXPLICIT_LSR, errorCode);
         case 'B':
+            if (returnInputIfUnmatch) {
+                return LSR(language, script, region, LSR::EXPLICIT_LSR);
+            }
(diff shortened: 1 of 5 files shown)


richgillam:
I looked this over, and it looks okay, but I don't really understand the nature of the problem of how the solution solves it.  Would you mind adding a little bit of explanation here for ignorant people such as myself?

FrankYFTang:
> I looked this over, and it looks okay, but I don't really understand the nature of the problem of how the solution solves it. Would you mind adding a little bit of explanation here for ignorant people such as myself?

until ICU 74, the XLikelySubtags is only used as part of the LocaleMatcher implementation Markus put down as part of 
https://github.com/unicode-org/icu/pull/714
and
https://github.com/unicode-org/icu/pull/409

That code handle pseudo locales in LocaleMatcher. 
In the development cycle of ICU 74, I replace the code inside Locale with this implementation to remove the need for redundant data and consolidate the implementation into one. While I doing that, I added a boolean flag  returnInputIfUnmatch to the preexisting function, while the value is off, the function behave as how it did for the old use case of LocaleMatcher, and the new code pass in that value as true, to tweak the return behavior to conform with what the Locale class ask for.

However, I didn't realize the existing speical handling of Pseudo Locales in the code code since there are no unit test written for them

I am not sure why Markus add those special logic for the use of LocaleMatcher. All I did here is to use that flag to make sure while it is called from the Locale usage (not LocaleMatcher usage) it skip the special processing as what the bug reporter expected.

richgillam:
> I am not sure why Markus add those special logic for the use of LocaleMatcher. All I did here is to use that flag to make sure while it is called from the Locale usage (not LocaleMatcher usage) it skip the special processing as what the bug reporter expected.

Thanks; that helps.  I didn't know what `XA`, `XB`, and `XC` were.  But it sounds like the salient point is that for your purposes, you don't care either and are skipping the code that treats them as special.  That sounds good to me.

richgillam (approved):

PR #2697, opened 2023-11-09
ICU-22564 Java API for the host env to register a low level break eng…

FrankYFTang:
…ine to break CJ + Southeastern Asia script

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22564

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
@@ -40,6 +40,7 @@ import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.lang.UProperty;
 import com.ibm.icu.lang.UScript;
 import com.ibm.icu.util.CodePointTrie;
+import com.ibm.icu.util.ULocale;
 
 /**
  * Rule Based Break Iterator
@@ -727,7 +728,7 @@ public class RuleBasedBreakIterator extends BreakIterator {
         // We have a dictionary character.
         // Does an already instantiated break engine handle it?
         for (LanguageBreakEngine candidate : fBreakEngines) {
-            if (candidate.handles(c)) {
+            if (candidate.handles(c, getRequestedLocale())) {
                 return candidate;
             }
         }
@@ -737,7 +738,7 @@ public class RuleBasedBreakIterator extends BreakIterator {
             // Check the global list, another break iterator may have instantiated the
             // desired engine.
             for (LanguageBreakEngine candidate : gAllBreakEngines) {
-                if (candidate.handles(c)) {
+                if (candidate.handles(c, getRequestedLocale())) {
                     fBreakEngines.add(candidate);
                     return candidate;
                 }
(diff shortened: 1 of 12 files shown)


FrankYFTang:
@richgillam Somehow my FakeYueBreakEngine break the PersonNameConsistencyTest on yue.txt and yue_Hans.txt . I am suprise that changes to BreakIterator will impact the outcome of PersonName. Could you explain how does PersonName depends on BreakIterator?

richgillam (approved):
LGTM

richgillam:
> @richgillam Somehow my FakeYueBreakEngine break the PersonNameConsistencyTest on yue.txt and yue_Hans.txt . I am suprise that changes to BreakIterator will impact the outcome of PersonName. Could you explain how does PersonName depends on BreakIterator?

PersonName is using BreakIterator for a few things, as I remember it:
- Some field modifiers capitalize each individual word in the field, using BreakIterator to find the word boundaries.
- I think the initial-formation stuff uses BreakIterator to break a field into words so that it can convert each of them into an initial.
- I could be wrong, but I think the space-replacement stuff may also use BreakIterator to figure out where to insert or substitute spaces.

The first two don't seem relevant to Yue, but the third one might be (if it's real-- I don't remember for sure now).  Some of the foreign-name support might be triggering the capitalization/initial logic even though it's not strictly relevant (e.g., formatting Yue with an English person name formatter).  If you can tell me more about which tests are failing, what they expect, and what you're getting instead, I can help you track the problem down (if you need help, of course).

richgillam (commented):
Is that what we agreed to last week?  I thought I came away from that discussion thinking we were adding a _third_ method to say whether the break engine supported a locale, not turning the `isFor()` method into that method by getting rid of the `c` parameter.  Do we no longer need an `isFor()` function that tests a character to see if it can begin a run of characters to be broken?

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1081,6 +1082,39 @@ private static int CISetIndex32(CharacterIterator ci, int index) {
         return ci.getIndex();
     }
 
+    /**
+     * Register a new external break engine. The external break engine will be adopted.
+     * Because ICU may choose to cache break engine internally, this must
+     * be called at application startup, prior to any calls to
+     * object methods of RuleBasedBreakIterator to avoid undefined behavior.
```
macchiati, line 1089:
```suggestion
     * Register a new external break engine. The external break engine will be adopted.
     * Because ICU may choose to cache break engines internally, 
     * to avoid undefined behavior this must
     * be called at application startup, prior to any calls to
     * object methods of RuleBasedBreakIterator.
```

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1885,7 +1919,42 @@ void dumpCache() {
...
+         */
+        public boolean handles(int c);
 
+        /**
+         * <p>Divide up a range of text handled by this break engine.</p>
+         *
+         * @param text A CharacterIterator representing the text
+         * @param rangeStart The start of the range of known characters
+         * @param rangeEnd The end of the range of known characters
+         * @param foundBreaks Output of a list of Integer to denote break positions.
```
macchiati, line 1949: This should specify whether the contents of the List must be cleared before calling, or whether the fillBreaks is required to clear internally.

macchiati (commented):
Looks good to me, but someone more familiar with the break iterator api should review.

markusicu:
AFAIK @FrankYFTang has not yet addressed the feedback on his proposal from last year, and hasn't yet sent an updated proposal.

PR #2705, opened 2023-11-17
ICU-22549 Add UnicodeSet fuzzer

FrankYFTang:
##### Checklist


richgillam (approved):
This seems fine, but I wonder a bit whether it'd be worth it to create a `UnicodeSet` and then do multiple iterations of calling the various API functions with different randomly-generated strings.  That is, put ll. 26-55 inside a loop, move ll. 16-19 into that loop, and iterate through maybe 5 times.  I guess I'm curious if there's any state retained that might cause trouble if you call, say, `applyPattern()` a few times in a row with different patterns.

Totally optional, but an idea that I thought was worth sharing...

PR #2706, opened 2023-11-18
ICU-22549 Add fuzzer for NumberFormatter

FrankYFTang:
##### Checklist


richgillam (approved):
This looks okay to me, I think, but I'd like @sffc or somebody else more familiar with `NumberFormatter` to weigh in on whether there are additional APIs on this class you should be fuzzing.

sffc (commented):
Suggestion: take 4 bytes of the input and make them into a floating point number to be formatted. Parsing the skeleton is good, but we can also cover data loading. However, doing so will make the fuzzer a bit slower.

richgillam (approved):
I like it.  Hopefully Shane does too.

sffc (commented):
Looks right; it's possible I missed something in all the pointer math

PR #2707, opened 2023-11-18
ICU-22575 Change AvailableFormatsSink to allow locales to inherit ava…

richgillam:
…ilableFormats items from the root locale.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22575


richgillam:
**NOTE:** I think this is the same problem we've been getting reports from various Web users about.  A few quick comments:

- This is the fix I made on the Apple side.  There may be a better, cleaner way to do this, although I thought about it a little and this seemed right.
- If there are new test cases I should add that would correspond better to the reports were getting from the browser people, I'll add them.  Can somebody suggest some?
- I have no idea why we were doing `!isRoot` before; there didn't seem to be any bad side effects from taking it out.
- I haven't yet vetted the test changes in this PR; I vetted some of them on the Apple side and confirmed that the new results were the right ones, but I need to go back and do that.
- I haven't yet ported this change to Java.  I'll do that before I check in (but feel free to remind me).
- I just realized to put this change up against the `main` branch, but that it's also needed against the `maint-74` branch.  I can cherry-pick over there and post another PR, but I'd like some validation of this approach first.

Anyway, sorry this is a little half-baked, but I ran out of time and wanted this up before the Thanksgiving break (I'm off all next week) to get feedback and reassure people that the problem was being worked on.

FrankYFTang:
There should be a Java counter part, right? Could we change the Java one in this PR together?

richgillam:
> There should be a Java counter part, right? Could we change the Java one in this PR together?

Of course.  I just ran out of time before the break to do that, and I kind of wanted some feedback on what I'm doing and whether it's a good idea or there's a better way to do it.  If people are behind this (or don't care), I'll try to put up a new commit that extends the fix to the ICU4J side.

FrankYFTang:
please fix the failing tests

richgillam:
Okay, I think this is finally ready to be reviewed.  I've fixed the build failures, verified that the changed unit test results make sense, added a new unit test that specifically replicates the problem described in the Mozilla bug report, and ported all of those things to Java.  @markusicu @FrankYFTang @pedberg-icu , PTAL.

macchiati (commented):
I looked this over, and it all looks good. 

One could also write a little tool that generates patterns from skeletons for different languages, where the skeletons are unions of those of the root and child locales, and captures that in a file. Then run it with and without the code change, and spot-check the differences.

But I'm maybe a bit paranoid..., so glad to approve if you don't have the time to do that.

pedberg-icu (approved):
The code change and the expected data changes in tests all seem good to me. The expected data changes seem to be places where the CLDR data had ↑↑↑ assuming inheritance from root, so making the ICU code align with that seems good.

richgillam:
Thank you for your reviews, @pedberg-icu and @macchiati !  I just squashed my commits-- would somebody mind re-approving?

FrankYFTang (approved):
LGTM. Thanks for working on this.

PR #2709, opened 2023-11-21
ICU-22549 Add RuleBasedBreakIterator fuzzer

FrankYFTang:
##### Checklist

### DIFF
--- a/icu4c/source/test/fuzzer/rule_based_break_iterator_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/rule_based_break_iterator_fuzzer.cpp
@@ -1,0 +1,26 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+#include <cstring>
+
+#include "fuzzer_utils.h"
+#include "unicode/localpointer.h"
+#include "unicode/locid.h"
+#include "unicode/rbbi.h"
+
+IcuEnvironment* env = new IcuEnvironment();
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+  UErrorCode status = U_ZERO_ERROR;
+
+  size_t unistr_size = size/2;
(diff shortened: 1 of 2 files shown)


icu4c/source/test/fuzzer/rule_based_break_iterator_fuzzer.cpp
```diff
@@ -0,0 +1,26 @@
...
+  UErrorCode status = U_ZERO_ERROR;
+
+  size_t unistr_size = size/2;
+  std::unique_ptr<char16_t[]> fuzzbuff(new char16_t[unistr_size]);
+  std::memcpy(fuzzbuff.get(), data, unistr_size * 2);
+  icu::UnicodeString fuzzstr(false, fuzzbuff.get(), unistr_size);
+
+  UParseError parse;
+  icu::LocalPointer<icu::RuleBasedBreakIterator> brk(
+      new icu::RuleBasedBreakIterator(fuzzstr, parse, status));
```
richgillam, line 23: This is almost always going to give you an error code saying the rule string is invalid.  Maybe that's fine, but I wonder if we need a more devious test that takes advantage of knowing what the syntax of RBBI looks like.

Of course, that'd be a lot of work.  I think this is probably fine for now, but we should think about smarter/more evil fuzz testing here for some time in the future.

richgillam (approved):
This feels like the bare minimum, but should be fine.

icu4c/source/test/fuzzer/rule_based_break_iterator_fuzzer.cpp
FrankYFTang: I think the purpuse for fuzzing is the data is power by AI and try to generate input that is "unexpected" by the programmer to make sure during that condition the code won't break but return error. The smart of the fuzzing is based on the fuzzer and I will leave that smart to the AI to do it. The AI might be able to see all the branching in our code and figure out how to generate better fuzzing data to attack those (if not, then it is not a good AI)

PR #2712, opened 2023-11-22
ICU-22522 Update the Azure CI from Clang 14 to Clang 16

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22522

ALLOW_MANY_COMMITS=true

### DIFF
--- a/.ci-builds/.azure-pipelines.yml
+++ b/.ci-builds/.azure-pipelines.yml
@@ -72,7 +72,7 @@ jobs:
         CC: clang
         CXX: clang++
 #-------------------------------------------------------------------------
-- job: ICU4C_Clang_Cpp14_Debug_Ubuntu_2204
+- job: ICU4C_Clang_Cpp17_Debug_Ubuntu_2204
   displayName: 'C: Linux Clang C++17 Debug (Ubuntu 22.04)'
   timeoutInMinutes: 30
   pool:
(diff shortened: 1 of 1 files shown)


.ci-builds/.azure-pipelines.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         CC: clang
         CXX: clang++
 #-------------------------------------------------------------------------
-- job: ICU4C_Clang_Cpp14_Debug_Ubuntu_2204
+- job: ICU4C_Clang_Cpp16_Debug_Ubuntu_2204
```
jefgen, line 75:
```suggestion
- job: ICU4C_Clang_Cpp17_Debug_Ubuntu_2204
```

I think this was originally intended to be the C++ version, rather than the Clang version...

Looking back in the history, it seems that it got out of sync when it was updated to C++17 here: https://github.com/unicode-org/icu/commit/edc62cfcc16b8d9dcf4e202b8de97c6fca8f4a25

So I think it might be better to change this back to match, i.e. Cpp17 instead of Cpp16.

roubert: Done.

jefgen (approved):
lgtm, thanks for working on this.

PR #2724, opened 2023-12-05
ICU-22593 Add missing layout/ prefix for Layout Engine header files.

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22593


icu4c/source/test/letest/testdata.cpp
```diff
@@ -16,8 +16,8 @@
  *   generated by: gendata.cpp
  */
```
markusicu, line 17: Why does GitHub show this with red background?

markusicu (approved):
Is there no make/build file where we can remove a `-I` now?

roubert:
> Is there no make/build file where we can remove a `-I` now?

Not directly in the way I think you're thinking, but there are two makesfiles with `-I` references to the ICU internal `layout/` subdirectory (which isn't actually used anymore now when these header files come from the icu-le-hb library) that would make some sense to remove in this PR so I've now done that.

markusicu (approved):
tnx!

PR #2731, opened 2023-12-07
ICU-22581 Fix RBBI Leakage

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22581

### DIFF
--- a/icu4c/source/common/rbbiscan.cpp
+++ b/icu4c/source/common/rbbiscan.cpp
@@ -289,6 +289,9 @@ UBool RBBIRuleScanner::doParseActions(int32_t action)
 
             // Terminate expression, leaves expression parse tree rooted in TOS node.
             fixOpStack(RBBINode::precStart);
+            if (U_FAILURE(*fRB->fStatus)) {
+                break;
+            }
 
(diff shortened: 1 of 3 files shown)


aheninger:
The fix looks believable.
Adding a test case that fails without the fix would be good.
(Failure can just be a leak detected when running with Asan.)
It looks like any duplicate $Variable definition in the rules would do it, so just construct a BI from a small fake rule string, something like `$foo=[abc]; $foo=[xyz]; $foo; ` and see what happens.

FrankYFTang:
> $foo=[abc]; $foo=[xyz]; $foo;

looks like your test case surfface multiple leakage issues . My current change fix on of them but there are another leakage in RBBIRuleScanner::scanSet() /usr/local/google/home/ftang/icu/icu/icu4c/source/common/rbbiscan.cpp:1263:13
not reported by the fuzzer bug also need to be fixed.

FrankYFTang:
@aheninger Add test as requested. Please take another look

aheninger (approved):

PR #2732, opened 2023-12-08
ICU-22584 Fix RBBI rule builder stack overflow.

aheninger:
The problem was found by fuzz testing.

A rule consisting of a long literal string produces a large, unbalanced parse tree, one node per string element. Deleting the tree was recursive, once per node, resulting in deep recursion.

This PR changes node deletion to use an iterative (non-recursive) approach.

This change only affects rule building. There is no change to the RBBI run time using pre-built rules.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22584

### DIFF
--- a/icu4c/source/common/rbbinode.cpp
+++ b/icu4c/source/common/rbbinode.cpp
@@ -123,19 +123,64 @@ RBBINode::~RBBINode() {
         break;
 
     default:
-        delete        fLeftChild;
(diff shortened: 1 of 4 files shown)


FrankYFTang (approved):
Thanks for working on this. It looks good. Indeed fix the issue

PR #2736, opened 2023-12-09
ICU-22585 Fix infinity loop while unicode set contains single surrogate

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22585


icu4c/source/common/rbbiscan.cpp
```diff
@@ -1269,6 +1269,7 @@ void RBBIRuleScanner::scanSet() {
 
         n = pushNewNode(RBBINode::setRef);
         if (U_FAILURE(*fRB->fStatus)) {
+            delete uset;
```
aheninger, line 1272: Optional: change uset to be a LocalPointer<UnicodeSet>, and get rid of all of the explicit deletes of it on error conditions within this function. This would be cleaner and less error prone, but is a more intrusive change.

aheninger (dismissed):

richgillam (dismissed):
I like Andy's suggestion, though...

FrankYFTang:
PTAL

PR #2747, opened 2023-12-15
ICU-22609 Fix crash when udat_open() is passed a bogus locale name.

dbaron:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22609

-----

This fixes a crash reported in [Chromium bug 1491726](https://bugs.chromium.org/p/chromium/issues/detail?id=1491726).  As described in that bug, the underlying problem is that when `udat_open()` is passed a bad locale name, the code ends up constructing a `Locale` object and performing operations on that object without first calling its `isBogus` method, which leads to a crash.  This change simplifies the code in a way that I believe is equivalent for non-bogus locale values, and fixes the crash.

-----

I was unable to figure out from the [contributing documentation](https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md) how to get an account to file a JIRA issue for this problem.  It links to [this contact page](https://icu.unicode.org/contacts) which links to a few mailing lists that all seem inappropriate and to [this page on the bug tracking system](https://icu.unicode.org/bugs) that mentions the need for an account but doesn't say how to create one.  If I actually go to [JIRA](https://unicode-org.atlassian.net/jira/software/c/projects/ICU/issues/) it says "Something's gone wrong",  "Looks like you've been signed out. Try logging in again.", "Log in", but clicking "Log in" just returns to the same page with the same error message, which doesn't appear to provide a mechanism for account creation.

### DIFF
--- a/icu4c/source/i18n/udat.cpp
+++ b/icu4c/source/i18n/udat.cpp
@@ -150,26 +150,17 @@ udat_open(UDateFormatStyle  timeStyle,
         return (UDateFormat*)fmt;
       } // else fall through.
     }
+    Locale localeObj(locale);
+    if (localeObj.isBogus()) {
+        localeObj = Locale::getDefault();
+    }
     if(timeStyle != UDAT_PATTERN) {
-        if(locale == 0) {
-            fmt = DateFormat::createDateTimeInstance((DateFormat::EStyle)dateStyle,
-                (DateFormat::EStyle)timeStyle);
-        }
-        else {
-            fmt = DateFormat::createDateTimeInstance((DateFormat::EStyle)dateStyle,
-                (DateFormat::EStyle)timeStyle,
-                Locale(locale));
-        }
+        fmt = DateFormat::createDateTimeInstance((DateFormat::EStyle)dateStyle, (DateFormat::EStyle)timeStyle, localeObj);
     }
     else {
         UnicodeString pat((UBool)(patternLength == -1), pattern, patternLength);
 
-        if(locale == 0) {
-            fmt = new SimpleDateFormat(pat, *status);
-        }
-        else {
-            fmt = new SimpleDateFormat(pat, Locale(locale), *status);
-        }
+        fmt = new SimpleDateFormat(pat, localeObj, *status);
     }
 
     if(fmt == nullptr) {
--- a/icu4c/source/test/cintltst/cdattst.c
+++ b/icu4c/source/test/cintltst/cdattst.c
@@ -51,6 +51,7 @@ static void TestNarrowQuarters(void);
 static void TestExtraneousCharacters(void);
 static void TestParseTooStrict(void);
 static void TestHourCycle(void);
+static void TestLocaleNameCrash(void);
 
 void addDateForTest(TestNode** root);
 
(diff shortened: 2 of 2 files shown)


dbaron:
Well, [the contributing documentation](https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md) says "If your employer has already signed a CLA, then when you open your first Pull Request, an automated comment will appear that contains a link you can follow to declare your affiliation with this employer." but I don't see that link in the above automated comment.

FrankYFTang:
@markusicu @sffc I sent a private message about dbaron's CLA status

FrankYFTang:
I do not think the proposed fix is "good enough". For example, the following case will still crash I believe:
```
DateFormat::createDateTimeInstance(DateFormat::kNone, DateFormat::kMedium, Locale("notalanguage")
```

dbaron:
It's not clear to me what the expectation is for users of `Locale` to check `isBogus`:  should the caller that constructs the `Locale` check it, the API that takes a `Locale` as a parameter, or both?  (I guess I see some of both in the existing code.)

FrankYFTang:
I put down a counter proposal fix in https://github.com/unicode-org/icu/pull/2750
we also need to make sure the C++ API will not crash

srl295:
> Well, [the contributing documentation](https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md) says "If your employer has already signed a CLA, then when you open your first Pull Request, an automated comment will appear that contains a link you can follow to declare your affiliation with this employer." but I don't see that link in the above automated comment.

@dbaron  it looks like there are some misleading aspects about how the CLA is displayed. As a result of this we're going to file a ticket upstream with the CLA assistant, and look at if we should update our wording for clarity.

Meanwhile, please click the 'Login to sign…' button. it won't actually sign until you click the Agree button. After logging in, it will as you for your affiliation, choose **My employer has already signed the CLA.** if that is appropriate.

As to the JIRA login, I'm not sure what's happening there. What do you get when you try to login to https://id.atlassian.com ?

FrankYFTang:
obsoleted by https://github.com/unicode-org/icu/pull/2750

srl295:
> > Well, [the contributing documentation](https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md) says "If your employer has already signed a CLA, then when you open your first Pull Request, an automated comment will appear that contains a link you can follow to declare your affiliation with this employer." but I don't see that link in the above automated comment.
> 
> @dbaron it looks like there are some misleading aspects about how the CLA is displayed. As a result of this we're going to file a ticket upstream with the CLA assistant, and look at if we should update our wording for clarity.
> 
> Meanwhile, please click the 'Login to sign…' button. it won't actually sign until you click the Agree button. After logging in, it will as you for your affiliation, choose **My employer has already signed the CLA.** if that is appropriate.
> 
> As to the JIRA login, I'm not sure what's happening there. What do you get when you try to login to https://id.atlassian.com ?

I've filed https://github.com/cla-assistant/cla-assistant/pull/1034 to improve the wording in the CLA assistant, please comment on that PR if you have opinions.

PR #2748, opened 2023-12-15
ICU-22549 Add DateFormatSymbols fuzzer

FrankYFTang:
##### Checklist

### DIFF
--- a/icu4c/source/test/fuzzer/dtfmtsym_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/dtfmtsym_fuzzer.cpp
@@ -1,0 +1,75 @@
+// © 2023 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+// Fuzzer for ICU DateFormatSymbols.
+
+#include <cstring>
+
+#include "fuzzer_utils.h"
+
+#include "unicode/dtfmtsym.h"
+#include "unicode/locid.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
+    uint16_t rnd;
+    icu::DateFormatSymbols::DtContextType context;
+    icu::DateFormatSymbols::DtWidthType width;
+    if (size < sizeof(rnd) + sizeof(context) + sizeof(width)) return 0;
+    icu::StringPiece fuzzData(reinterpret_cast<const char *>(data), size);
+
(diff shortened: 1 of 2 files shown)


icu4c/source/test/fuzzer/dtfmtsym_fuzzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    std::memcpy(&rnd, fuzzData.data(), sizeof(rnd));
+    icu::Locale locale = GetRandomLocale(rnd);
+    fuzzData.remove_prefix(sizeof(rnd));
+
+    std::memcpy(&context, fuzzData.data(), sizeof(context));
+    fuzzData.remove_prefix(sizeof(context));
+    icu::DateFormatSymbols::DtContextType context_mod =
+        static_cast<icu::DateFormatSymbols::DtContextType>(
+            context % icu::DateFormatSymbols::DtContextType::DT_CONTEXT_COUNT);
```
richgillam, line 28: With both the context and the width, do you want to do the "mod" thing (or _only_ do the "mod" thing)?  It seems again that it'd be useful also to check that passing a completely illegal enum value (with, of course, you can do in C++ with casting) doesn't cause anything to barf.

icu4c/source/test/fuzzer/dtfmtsym_fuzzer.cpp
```diff
@@ -0,0 +1,75 @@
...
+    icu::UnicodeString output;
+    dfs->getTimeSeparatorString(output);
+    dfs->getYearNames(count, context, width);
+    dfs->getYearNames(count, context_mod, width_mod);
+    dfs->getZodiacNames(count, context, width);
+    dfs->getZodiacNames(count, context_mod, width_mod);
+    dfs->getLeapMonthPatterns(count);
+    int32_t count2;
+    dfs->getZoneStrings(count, count2);
+    dfs->getLocalPatternChars(output);
```
richgillam, line 72: Unless I'm mistaken, a `DateFormatSymbols` object is mutable, so you might also want to check what happens if you actually set the symbols themselves to random garbage values.  Of course, the symbols object just behaves as a struct, so to really test that, you'd also have to use the bogus symbols object to format a date.  (This shouldn't be a problem unless you've got really long strings, unassigned characters, illegal byte combinations, or possibly characters with unusual semantics, but it seems worth checking for all those things...)

richgillam (approved):
I'm willing to let this go in as-is, but I have a few things you might want to think about...

PR #2750, opened 2023-12-15
ICU-22609 Fix nulldef w/ bogus locale in DateFormat::create*

FrankYFTang:
Copy the test from https://github.com/unicode-org/icu/pull/2747
This is my counter proposal for #2747 because the crash could still happen via C++ interface.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22609

### DIFF
--- a/icu4c/source/i18n/smpdtfmt.cpp
+++ b/icu4c/source/i18n/smpdtfmt.cpp
@@ -728,16 +728,17 @@ void SimpleDateFormat::construct(EStyle timeStyle,
         if (locale.getKeywordValue("rg", nullptr, 0, dummyErr1) > 0 || locale.getKeywordValue("hours", nullptr, 0, dummyErr2) > 0) {
             hasRgOrHcSubtag = true;
         }
-            
+
         const char* baseLocID = locale.getBaseName();
-        if (baseLocID[0]!=0 && uprv_strcmp(baseLocID,"und")!=0) {
+        if (baseLocID != nullptr && uprv_strcmp(baseLocID,"und")!=0) {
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/datefmt.cpp
```diff
@@ -519,9 +519,15 @@ DateFormat::createInstanceForSkeleton(
 //----------------------------------------------------------------------
 
 DateFormat* U_EXPORT2
-DateFormat::create(EStyle timeStyle, EStyle dateStyle, const Locale& locale)
+DateFormat::create(EStyle timeStyle, EStyle dateStyle, const Locale& inLocale)
 {
     UErrorCode status = U_ZERO_ERROR;
+    Locale locale(inLocale);
+    // If the locale is bogus, fallback to use default locale first.
```
markusicu, line 526: It looks like create() is internal/private.
Wouldn't it be cleaner to add a UErrorCode parameter, and fail here with an illegal-argument-error?

FrankYFTang: I found a better fix, PTAL

markusicu: I found the answer to my own question: The public createXyZInstance() functions all don't take UErrorCode either, so we couldn't propagate any error :-(

Looking at your new fix now.

markusicu (approved):
seems reasonable

FrankYFTang:
Sorry, I refine it. PTAL

markusicu (approved):

PR #2751, opened 2023-12-16
ICU-22549 Improve fuzzer to test more locale

FrankYFTang:
We found bogus locale cause crash in DateFormat so here we enhance the fuzzer to also test locale name which are not return by the available locale list.

##### Checklist

### DIFF
--- a/icu4c/source/test/fuzzer/break_iterator_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/break_iterator_fuzzer.cpp
@@ -66,6 +66,26 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   for (int32_t p = bi->first(); p != icu::BreakIterator::DONE; p = bi->next()) {}
 
   utext_close(fuzzstr);
+
+  icu::Locale locale2(reinterpret_cast<const char*>(data));
+  switch (rnd8 % 5) {
+    case 0:
+      bi.reset(icu::BreakIterator::createWordInstance(locale2, status));
+      break;
+    case 1:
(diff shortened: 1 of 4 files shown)


richgillam (approved):
The changes themselves look okay, but the description says you're testing locale names that aren't in the available-locales list, and I don't see anything here that's doing that.

FrankYFTang:
> The changes themselves look okay, but the description says you're testing locale names that aren't in the available-locales list, and I don't see anything here that's doing that.

Notice "icu::Locale locale2(reinterpret_cast<const char*>(data));" 
In these test, we used to only pick a known locale from a locale list that the system return to test
Now, we add additional test to create the locale with random data then test the constructor. In some (actually many) cases, this locale will be a bogus locale. But we need to make sure passing bogus locale to the code will not cause crash (which fuzzer recently find some)

richgillam:
> Notice "icu::Locale locale2(reinterpret_cast<const char*>(data));" In these test, we used to only pick a known locale from a locale list that the system return to test Now, we add additional test to create the locale with random data then test the constructor. In some (actually many) cases, this locale will be a bogus locale. But we need to make sure passing bogus locale to the code will not cause crash (which fuzzer recently find some)

Oh, I saw the `locale2` stuff, but missed the line where it was actually being created.  My bad.

FrankYFTang:
The landing of this PR looks cause the fuzzer to break too frequently and considered as a "build failure" in 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65237&q=label%3AProj-icu&can=1&sort=-id. Let's revert this PR first to see would that fix the fuzzer build failure problem first.

PR #2753, opened 2023-12-19
ICU-22615 Test TimeZoneNames API will not assert with non ASCII.

FrankYFTang:
Add tests and avoid assertion when the ID has non-ASCII

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22615

### DIFF
--- a/icu4c/source/i18n/tznames_impl.cpp
+++ b/icu4c/source/i18n/tznames_impl.cpp
@@ -34,6 +34,7 @@
 #include "ureslocs.h"
 #include "zonemeta.h"
 #include "ucln_in.h"
+#include "uinvchar.h"
 #include "uvector.h"
 #include "olsontz.h"
 
@@ -2280,6 +2281,10 @@ TZDBTimeZoneNames::getMetaZoneNames(const UnicodeString& mzID, UErrorCode& statu
         return nullptr;
     }
     mzIDKey[mzID.length()] = 0;
+    if (!uprv_isInvariantUString(mzIDKey, mzID.length())) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return nullptr;
+    }
 
     static UMutex gTZDBNamesMapLock;
     umtx_lock(&gTZDBNamesMapLock);
--- a/icu4c/source/test/intltest/tzfmttst.cpp
+++ b/icu4c/source/test/intltest/tzfmttst.cpp
@@ -89,6 +89,7 @@ TimeZoneFormatTest::runIndexedTest( int32_t index, UBool exec, const char* &name
         TESTCASE(9, TestCentralTime);
         TESTCASE(10, TestBogusLocale);
         TESTCASE(11, Test22614GetMetaZoneNamesNotCrash);
+        TESTCASE(12, Test22615NonASCIIID);
     default: name = ""; break;
     }
 }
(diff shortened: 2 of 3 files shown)


icu4c/source/i18n/tznames_impl.cpp
```diff
@@ -2275,6 +2276,10 @@ TZDBTimeZoneNames::getMetaZoneNames(const UnicodeString& mzID, UErrorCode& statu
         return nullptr;
     }
     mzIDKey[mzID.length()] = 0;
+    if (!uprv_isInvariantUString(mzIDKey, mzID.length())) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return nullptr;
+    }
```
richgillam, line 2282: The ticket says there are "many places" in the API that assume the time zone ID is all ASCII but don't check for it, but this is the only substantive change in the PR.  Do all of the "many places" vector through here, or did you only fix one instance of the problem?

And is returning `U_ILLEGAL_ARGUMENT_ERROR` the right thing to do here?  The unit test appears to be calling a different API that doesn't return an error code.  What's going on between the functions the unit test is calling and this logic, and why is that the right thing to be doing?

My uninformed take is that you might be better off just converting the input time zone ID into an invariant-characters string and then using that value as the time zone ID.  If you get an invalid ID from doing that conversion, that's fine-- now it's just an invalid ID and you can handle that the way everything else handles an invalid ID.

icu4c/source/test/intltest/tzfmttst.cpp
```diff
@@ -1308,6 +1309,30 @@ TimeZoneFormatTest::TestFormatCustomZone() {
...
+    output.remove();
+    tzdb->getMetaZoneID(id, 0, output);
+    output.remove();
+    tzdb->getMetaZoneDisplayName(id, UTZNM_EXEMPLAR_LOCATION, output);
+    output.remove();
+    tzdb->getTimeZoneDisplayName(id, UTZNM_SHORT_DAYLIGHT, output);
+    output.remove();
+    tzdb->getExemplarLocationName(id, output);
+    output.remove();
+    tzdb->getDisplayName(id, UTZNM_LONG_GENERIC, 0, output);
```
richgillam, line 1333: I'm a little uncomfortable with a unit test that calls a bunch of stuff and doesn't check the results.  While "it doesn't crash" is a reasonable success criterion, it seems like you should be taking some kind of stand as to what the code _should_ do when passed an invalid value like this.

In particular, you're calling a bunch of functions that don't return an error code, so you can't just check the error code.  If they can't signal an error and you pass then an invalid ID, what should they do?  I can't tell that from this unit test.

richgillam (commented):
Finally had a chance to look at this and had a number of questions about what you're trying to do here.

icu4c/source/i18n/tznames_impl.cpp
FrankYFTang: getMetaZoneNames is the internal function that many public surfaced API called into. That is why if we check there it fix many functions.

icu4c/source/test/intltest/tzfmttst.cpp
FrankYFTang: I check the value now

icu4c/source/i18n/tznames_impl.cpp
FrankYFTang: the problem is in TZDBTimeZoneNames but that is an internal class. I add the unit test which eventually call into TZDBTimeZoneNames

FrankYFTang:
PTAL

richgillam (approved):
Looks good.  Thanks for the changes and explanations.

PR #2757, opened 2023-12-21
ICU-22576 Remove now obsolete pre-C11 backward compatibility code

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22576

### DIFF
--- a/icu4c/source/test/cintltst/ccapitst.c
+++ b/icu4c/source/test/cintltst/ccapitst.c
@@ -15,6 +15,7 @@
 ******************************************************************************
 */
 
+#include <stdalign.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -36,12 +37,6 @@
 #define NUM_CODEPAGE 1
 #define MAX_FILE_LEN 1024*20
 #define UCS_FILE_NAME_SIZE 512
-
-#if __STDC_VERSION__ < 201112
-# define alignof(type) offsetof (struct { char c; type member; }, member)
-#elif __STDC_VERSION__ < 202311
-# include <stdalign.h>
-#endif
 
 /*returns an action other than the one provided*/
 #if !UCONFIG_NO_LEGACY_CONVERSION


roubert:
@jefgen Would you please be so kind and help me take a look at this puzzling MSVC compilation error?

As far as I can tell from the documentation, the code in question should be just fine the way it is, also for MSVC:

https://learn.microsoft.com/en-us/cpp/c-language/alignment-c

But instead the result is this:

```
icu4c\source\test\cintltst\ccapitst.c(1838,72): Error C2275: 'UConverter': expected an expression instead of a type
            if( actualSizes[idx] <= (bufferSizes[j] - (int32_t)alignof(UConverter)) &&
                                                                       ^
```

Do you have any idea what might be going on here?

markusicu:
> @jefgen Would you please be so kind and help me take a look at this puzzling MSVC compilation error?

@rp9-next @jalaj-microsoft FYI

markusicu:
> As far as I can tell from the documentation, the code in question should be just fine the way it is, also for MSVC:
> 
> https://learn.microsoft.com/en-us/cpp/c-language/alignment-c

Do we need to explicitly set the C version?
Do we need to abandon an old compiler?

> Requirements
> Compile with /std:c11.
> Windows SDK 10.0.20348.0 (version 2104) or later.

roubert:
> Do we need to explicitly set the C version?

As it turns out, yes, we do. Apparently the MSVC configuration for ICU4C doesn't specify the language standard versions to use, at all, but instead just relies on the default (or a commandline override) to be at least the version required by ICU4C (and this has apparently happened to be true until now, with the unwelcome but unnoticed side-effect of our MSVC CI accepting C++14 for quite some time now even though ICU4C was only to require C++11).

So I've uploaded PR #2767 to explicitly set C11 & C++17 in the MSVC configuration, which will solve the mystery in this PR here, but unfortunately create a new mystery of one particular unit test that for no discernable reason fails in two particular MSVC configurations when compiled as C++17.

roubert:
> So I've uploaded PR #2767 to explicitly set C11 & C++17 in the MSVC configuration,

With that merged this PR now builds fine on every sufficiently new version of MSVC, but shows that _C11 Standard library features_ weren't introduced until VS 2019, so we'll have to drop older versions from our CI in order to actually use things from the C11 standard library (such as `alignof`) in our code:

https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170#c-standard-library-features-1

So I've now uploaded PR #2790 to remove VS 2015 and VS 2017.

roubert:
> So I've now uploaded PR #2790 to remove VS 2015 and VS 2017.

With also that merged this PR now finally builds fine with all tests passing on every configuration in our CI, as expected. Ready for review.

cjchapman (approved):
👍

markusicu (approved):

PR #2766, opened 2024-01-03
ICU-22520 Move all localeID parsing logic into new ulocimp_getSubtags()

roubert:
The logic for parsing a localeID string into its constituent subtags is currently repeated over and over again in each one of the `uloc_get*()` functions, so that calling all these functions one after the other in order to get all the subtags does the parsing all over again from the beginning for each function call.

In order to avoid having to do this parsing over and over again, a lot of code instead has its own copy of the parsing logic in order to call the underlying `ulocimp_get*()` functions directly for lower runtime cost at the price of increased code complexity and repetition.

This new `ulocimp_getSubtags()` function, which writes natively to `icu::ByteSink` and has a convenience wrapper to write to `icu::CharString`, removes the repeated code from the `uloc_get*()` functions and makes it possible to update all code that calls the `ulocimp_get*()` functions.

##### Checklist

ALLOW_MANY_COMMITS=true


markusicu:
This sounds like a good idea for this code, operating on locale ID *strings*.
However, before I review it in detail, I wonder if this is what we want to or should do long-term.

It is good to consolidate subtags parsers, but effectively you are creating a full locale ID / language tag parser that does most of the same stuff as the C++ Locale constructor and .forLanguageTag().
Redundant code is not good. You are reducing it, but still.

So my question is, would we be better off making the C++ Locale constructor/factory parser good and efficient, and then change this sort-of-C-code to construct C++ Locale objects and use their normal functions to fetch the subtags?
This would let us focus on one implementation, and should make the calling code more similar to the Java code.

If we did that, we might need to add more StringPiece/string_view/ByteSink APIs to the C++ Locale. We could start by adding `@internal` functions; when we are happy with the new code, we could propose the new functions to be truly public. It might also give us another motivation to improve the structure and parser of class Locale (which we have wanted to do for a long time).

And maybe your new code could be repurposed as the new parser implementation in class Locale.

Adding @richgillam @macchiati @yumaoka @FrankYFTang to the discussion.

roubert:
> However, before I review it in detail, I wonder if this is what we want to or should do long-term.

The way I see it, the changes in this PR are one further step towards making this code base possible to refactor, so it's a valuable change regardless of where we eventually want to end up in the future.

icu4c/source/common/uloc.cpp
```diff
@@ -1192,23 +1194,23 @@ ulocimp_getLanguage(const char *localeID,
         }
     }
 
-    if(pEnd!=nullptr) {
-        *pEnd=localeID;
+    if (sink != nullptr && !result.isEmpty()) {
+        sink->Append(result.data(), result.length());
```
markusicu, line 1198: nit: Append(result.toStringPiece())  ?

icu4c/source/common/uloc.cpp
```diff
@@ -1147,17 +1147,19 @@ uloc_getCurrentLanguageID(const char* oldID){
...
  */
-CharString U_EXPORT2
-ulocimp_getLanguage(const char *localeID,
-                    const char **pEnd,
-                    UErrorCode &status) {
+static void
+_getLanguage(const char* localeID,
+             ByteSink* sink,
+             const char** pEnd,
+             UErrorCode& status) {
```
markusicu, line 1161: optional: Consider *returning* the end pointer rather than writing it to an output argument.

icu4c/source/common/uloc.cpp
roubert: I'll soon replace it with writing directly to the sink instead.

icu4c/source/common/uloc.cpp
roubert: I'll keep that in mind, thanks!

icu4c/source/common/uloc.cpp
```diff
@@ -1312,6 +1317,140 @@ _getVariant(const char *localeID,
...
+    }
+
+    if (variant == nullptr) { return; }
+
+    CharString tmp;
+    if (_hasBCP47Extension(origLocaleID)) {
+        CharStringByteSink sink(&tmp);
+        ulocimp_forLanguageTag(origLocaleID, -1, sink, nullptr, &status);
+        if (U_SUCCESS(status) && !tmp.isEmpty()) {
+            // This recursive call sets its pEnd to where any variants begin.
```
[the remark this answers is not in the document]
markusicu: On this item I do want to put my foot down. This is too obscure and surprising even for keeping it only temporarily. Please do add some kind of parameter for stopping after the region code / before the variants.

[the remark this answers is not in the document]
roubert: OK, I've now replaced the use of `ulocimp_forLanguageTag()` in `uloc_getVariant()` with a little more straightforward (and much less wasteful) code that instead of converting the entire language tag just checks for the `-t-`, `-u-` or `-x-` marker that indicates the start of a BCP-47 extension.

icu4c/source/common/uloc.cpp
```diff
@@ -1103,6 +1103,14 @@ ulocimp_setKeywordValue(const char* keywords,
  */
 #define _isTerminator(a)  ((a==0)||(a=='.')||(a=='@'))
 
+inline bool _isBCP47Extension(const char* p) {
+    return p[0] == '-' &&
```
markusicu, line 1109: Should this be lenient with `-` vs `_`?
I *think* we usually are... but may be wrong.

icu4c/source/common/uloc.cpp
```diff
@@ -1103,6 +1103,14 @@ ulocimp_setKeywordValue(const char* keywords,
...
+           (p[1] == 't' || p[1] == 'T' ||
+            p[1] == 'u' || p[1] == 'U' ||
+            p[1] == 'x' || p[1] == 'X') &&
+           p[2] == '-';
```
markusicu, line 1113: ditto?

icu4c/source/common/uloc.cpp
```diff
@@ -1557,32 +1557,18 @@ uloc_openKeywords(const char* localeID,
         tmpLocaleID=localeID;
     }
 
-    /* Skip the language */
-    ulocimp_getLanguage(tmpLocaleID, &tmpLocaleID, *status);
+    ulocimp_getSubtags(
```
markusicu, line 1559: Why does the old and new code here bother to parse/identify the subtags at all, when eventually it just looks for `@` and parses the keywords from there?

icu4c/source/common/uloc.cpp
```diff
@@ -1312,6 +1317,140 @@ _getVariant(const char *localeID,
...
+        U_ASSERT(end != nullptr);
+        if (end != begin) {
+            hasRegion = true;
+            localeID = end;
+            if (region != nullptr) { region->Append(tmp.data(), tmp.length()); }
+            if (pEnd != nullptr) { *pEnd = localeID; }
+        }
+    }
+
+    if (variant == nullptr) { return; }
```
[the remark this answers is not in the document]
markusicu: I still don't like this asymmetric code. It looks like a bug, and someone else will stumble over it again.

Please check for pEnd here, like before, and if call sites want to stop between region & variant even though they want pEnd but not the variant, then please add a parameter for that.

markusicu (commented):
praise: more good cleanup!

richgillam (approved):
I didn't give this quite as close a reading as maybe I should have, but since it seems like Markus has given it a very close reading, I don't feel so bad about that.

I really like what you're doing here; I think this simplifies and prettifies things quite a bit.

I wonder if we need a follow-on ticket.  It looks like you changed all the call sites that used the stuff in `ulocimp.h`, but I think there are some other places that use the public APIs that could also be simplified by using your new stuff in `ulocimp.h`.  (I think I may have written some of those.)  I think it might be worth it to check through the code for any such spots and either fix them or file tickets to go back and do that later.

richgillam:
Actually, maybe I shouldn't have given an approval while you're still waiting for approval from Markus.  I think you should wait for him to sign off on this before checking in.

markusicu (approved):
thank you!!🎉

roubert:
> It looks like you changed all the call sites that used the stuff in `ulocimp.h`, but I think there are some other places that use the public APIs that could also be simplified by using your new stuff in `ulocimp.h`.

That work has already been going on for a long time now and it is still going on, see eg. new PR #2812.

icu4c/source/common/uloc.cpp
```diff
@@ -1743,11 +1882,28 @@ uloc_getLanguage(const char*    localeID,
...
+        return length;
     }
 
-    return ulocimp_getLanguage(localeID, nullptr, *err).extract(language, languageCapacity, *err);
+    if (sink.Overflowed()) {
+        *err = U_BUFFER_OVERFLOW_ERROR;
+        return length;
+    }
+
+    return u_terminateChars(language, languageCapacity, length, err);
```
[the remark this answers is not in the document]
roubert: > I still think that it for the time being is a good idea to simply continue copy-pasting this recipe until all the locale memory management has been updated, for at that point in time we'll have a much clearer view of exactly what would be the most useful design for such a helper function.

PR #2863

PR #2767, opened 2024-01-03
ICU-21107 Specify language standard versions C11 & C++17 also for MSVC

roubert:
There were until now no versions specified at all, relying on the default (or commandline overrides) to be sufficiently recent.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21107

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/extra/uconv/uconv.vcxproj
+++ b/icu4c/source/extra/uconv/uconv.vcxproj
@@ -34,9 +34,11 @@
       <TypeLibraryName>$(OutDir)\uconv.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <AdditionalIncludeDirectories>..\..\..\include;..\..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>UCONVMSG_LINK;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <PrecompiledHeaderOutputFile>$(OutDir)\uconv.pch</PrecompiledHeaderOutputFile>
--- a/icu4c/source/stubdata/stubdata.vcxproj
+++ b/icu4c/source/stubdata/stubdata.vcxproj
@@ -35,9 +35,11 @@
       <TypeLibraryName>$(OutDir)\icudt.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>STUBDATA_BUILD;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
       <PrecompiledHeaderOutputFile>$(OutDir)/icudt.pch</PrecompiledHeaderOutputFile>
--- a/icu4c/source/test/iotest/iotest.vcxproj
+++ b/icu4c/source/test/iotest/iotest.vcxproj
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)\iotest.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\..\..\include;..\..\tools\ctestfw;..\..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>U_ATTRIBUTE_DEPRECATED=;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
       <PrecompiledHeaderOutputFile>$(OutDir)\iotest.pch</PrecompiledHeaderOutputFile>
--- a/icu4c/source/tools/ctestfw/ctestfw.vcxproj
+++ b/icu4c/source/tools/ctestfw/ctestfw.vcxproj
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)/icutest.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\..\..\include;..\..\common;..\toolutil;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>T_CTEST_IMPLEMENTATION;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
       <PrecompiledHeaderOutputFile>$(OutDir)/icutest.pch</PrecompiledHeaderOutputFile>
--- a/icu4c/source/tools/gencmn/gencmn.vcxproj
+++ b/icu4c/source/tools/gencmn/gencmn.vcxproj
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)/gencmn.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <AdditionalIncludeDirectories>..\..\common;..\toolutil;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PrecompiledHeaderOutputFile>$(OutDir)/gencmn.pch</PrecompiledHeaderOutputFile>
       <AssemblerListingLocation>$(OutDir)/</AssemblerListingLocation>
--- a/icu4c/source/tools/gendict/gendict.vcxproj
+++ b/icu4c/source/tools/gendict/gendict.vcxproj
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)\gendict.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <AdditionalIncludeDirectories>..\..\common;..\toolutil;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PrecompiledHeaderOutputFile>$(OutDir)\gendict.pch</PrecompiledHeaderOutputFile>
       <AssemblerListingLocation>$(OutDir)/</AssemblerListingLocation>
--- a/icu4c/source/common/common.vcxproj
+++ b/icu4c/source/common/common.vcxproj
@@ -34,6 +34,8 @@
       <TypeLibraryName>$(OutDir)\icuuc.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <PreprocessorDefinitions>U_ATTRIBUTE_DEPRECATED=;U_COMMON_IMPLEMENTATION;U_PLATFORM_USES_ONLY_WIN32_API=1;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <DisableLanguageExtensions>false</DisableLanguageExtensions>
       <WarningLevel>Level3</WarningLevel>
--- a/icu4c/source/common/common_uwp.vcxproj
+++ b/icu4c/source/common/common_uwp.vcxproj
@@ -72,6 +72,8 @@
       <TypeLibraryName>$(OutDir)\icuuc.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\..\include;..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <!-- U_DISABLE_RENAMING -->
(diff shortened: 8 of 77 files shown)


roubert:
@jefgen @rp9-next @jalaj-microsoft @cjchapman

_Calling everyone I believe to know more about the MSVC compiler than I do … help needed … help needed …_

The changes in these `.vcxproj` files in this PR here seem to work as expected and correctly switch the MSVC compiler to use C11 & C++17 respectively for building all of ICU4C, including a previously existing warning about `alignof` disappearing as expected with the switch to C11, and everything still building and all tests still passing everywhere, with this one single mysterious exception:

The unit test `format/DataDrivenCalendarTest/TestCalendarOperations` now suddenly fails in two specific configurations, **CI (C: MSVC 32-bit Debug (VS 2019))** and **CI (C: MSVC x64 Debug (VS 2022))**, perfectly repeatable, even though it runs just fine everywhere else, including all other MSVC configurations.

Does anyone of you have a clue what's going on here? Would anyone of you be able to recreate the environment of those CI jobs and take a look in a debugger what's actually going on there?

cjchapman:
I defer to @jefgen @rp9-next @jalaj-microsoft on this, as I'm sure they know more about MSVC than I do, and I don't have a setup where I can reproduce the environment.

roubert:
What I've done to rule out the obvious potential sources of problems in `calendar.cpp` is to resolve the compiler warning (fix merged in commit 539e8f41a35fec30fba4cfb3a65ea4db67978f85) and also locally removed the `UPRV_NO_SANITIZE_UNDEFINED` annotations and verified that the sanitizer really doesn't find any other undefined behaviour in this code than out-of-range enum values intentionally passed by unit tests to verify that the code is able to deal with this (which were the reason for those annotations being added in the first place).

So that's where this mystery stands now, no compiler warnings, no sanitizer errors, works fine almost everywhere, and just like @cjchapman I don't have a setup where I can reproduce the environment.

markusicu:
@mihnita has a personal Windows machine I think.

markusicu:
Oh, and @eggrobin has a Windows machine.

icu4c/source/extra/uconv/uconv.vcxproj
```diff
@@ -34,9 +34,11 @@
       <TypeLibraryName>$(OutDir)\uconv.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 41: Why does this change? We don't normally want non-standard compiler features.

icu4c/source/stubdata/stubdata.vcxproj
```diff
@@ -35,9 +35,11 @@
       <TypeLibraryName>$(OutDir)\icudt.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>STUBDATA_BUILD;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 42: ditto

icu4c/source/test/iotest/iotest.vcxproj
```diff
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)\iotest.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\..\..\include;..\..\tools\ctestfw;..\..\common;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>U_ATTRIBUTE_DEPRECATED=;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 39: ditto

icu4c/source/tools/ctestfw/ctestfw.vcxproj
```diff
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)/icutest.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <AdditionalIncludeDirectories>..\..\..\include;..\..\common;..\toolutil;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
       <PreprocessorDefinitions>T_CTEST_IMPLEMENTATION;%(PreprocessorDefinitions)</PreprocessorDefinitions>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 39: ditto

icu4c/source/tools/gencmn/gencmn.vcxproj
```diff
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)/gencmn.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 39: ditto

icu4c/source/tools/gendict/gendict.vcxproj
```diff
@@ -32,9 +32,11 @@
       <TypeLibraryName>$(OutDir)\gendict.tlb</TypeLibraryName>
     </Midl>
     <ClCompile>
+      <LanguageStandard>stdcpp17</LanguageStandard>
+      <LanguageStandard_C>stdc11</LanguageStandard_C>
       <WarningLevel>Level3</WarningLevel>
       <CompileAs>Default</CompileAs>
-      <DisableLanguageExtensions>true</DisableLanguageExtensions>
+      <DisableLanguageExtensions>false</DisableLanguageExtensions>
```
markusicu, line 39: another, and more in further files...

icu4c/source/extra/uconv/uconv.vcxproj
roubert: Those changes are required to be able to select a non-default language standard version (see the compile errors from the first CI run of this PR, before changing those flags):

https://developercommunity.visualstudio.com/t/za-stdc11-leads-to-command-line-error-d8016-za-and/1398258

mihnita:
There is no need to update all of these files.
The common settings are in  `.props` files in the `allinone` folder.

I don't know exactly how these are used, so after spending a bit of time on it, this is the summary:
1. `Build.Windows.ProjectConfiguration.props`:  used by 69 `.vcxproj` files
2. `Build.Windows.UWP.ProjectConfiguration.props` : used by the 3 `_uwp.vcxproj` files
    * `source/common/common_uwp.vcxproj`
    * `source/data/makedata_uwp.vcxproj`
    * `source/i18n/i18n_uwp.vcxproj`
3. There are 7 projects that seem to not use shared properties:
    * `source/samples/legacy/legacy.vcxproj`
    * `source/test/letest/cletest.vcxproj`
    * `source/test/letest/gendata.vcxproj`
    * `source/test/letest/letest.vcxproj`
    * `source/test/perf/normperf/dtfmtrtperf.vcxproj`
    * `source/tools/icuswap/icuswap.vcxproj`
    * `source/tools/tzcode/icuzdump.vcxproj`

So I think that you should just change `Build.Windows.ProjectConfiguration.props` & `Build.Windows.UWP.ProjectConfiguration.props`, and that will cover 72 (69 + 3) projects
And the extra 7 projects directly.

---

You see that in my digging I find 69 + 3 + 7 = 79 files, vs 76 files in this PR.
(76 .vcxproje files, and 1 cpp file, source/i18n/calendar.cpp)

The difference seem to be these 3 files, missing in the PR:
* source/data/makedata.vcxproj
* source/data/makedata_uwp.vcxproj
* source/samples/uresb/resources.vcxproj

mihnita:
See https://learn.microsoft.com/en-us/cpp/build/create-reusable-property-configurations

roubert:
> So I think that you should just change `Build.Windows.ProjectConfiguration.props` & `Build.Windows.UWP.ProjectConfiguration.props`, and that will cover 72 (69 + 3) projects And the extra 7 projects directly.

From the point-of-view of someone like myself who isn't familiar with these configuration files it sure is tedious to have the language standard versions repeated in every single file, but it at least is logical and easy to understand. Instead having them specified in only one single file would also be logical, easy to understand and much easier to maintain. But having them specified in 9 particularly selected files doesn't really make much sense at all and seems difficult to work with, inviting a mistake to be made.

Therefore I propose to first proceed with repeating this in every single file as it currently is in this PR. Then, once this has been merged and works, it would be very nice if someone who's familiar with these configuration files (and has a setup where changes to them can be tested and verified), like maybe yourself, could spend the effort to refactor them so that the language standard versions get specified in only one single file.

roubert:
> Today (Monday, Jan 8) I work from home, and I will have access to my Windows machine. So after the `MessageFormat 2` meeting (ending ~11:00 am MTV time) I can help, review, or take over, whatever you think it's best.

If you have any clue why that `TestCalendarOperations` test fails in those two specific configurations, **CI (C: MSVC 32-bit Debug (VS 2019))** and **CI (C: MSVC x64 Debug (VS 2022))**, or if you have the possibility to reproduce those environments and take a look in a debugger what's actually happening, then that would be of great help.

mihnita:
> Therefore I propose to first proceed with repeating this in every single file as it currently is in this PR.

But we already have shared preferences.
And they are used for everything, including the main libraries we release.
So that is the clean solution, in use, and already correct.

Why should be "break" 72 projects that do the right thing already just to be consistent with 7 "broken" ones?

We are currently inconsistent already (because the 70+ projects already use the shared properties).
By doing why I suggested we remain as inconsistent as we are today.
With the PR as is we actively make things worse.

+100% to fix the 7 components, after this PR.

mihnita:
Sorry, in the office now.
But in the evening (once I'm home again) I'll look into `TestCalendarOperations`

mihnita:
I've spent a few hours setting up the environment, applying the C11/C++17 changes.
And I can reproduce the failures in VS 2022, the x64 Debug case.

BUT!

I have no clue how to debug that.
It is all data driven (in `icu4c\source\test\testdata\calendar.txt`), but I am not familiar how that works.
Looks like it's loaded by `dadrcal.cpp`, but I don't know how to run one test only.
So it is very time consuming running everything to get to the failing part to debug.

But somehow I suspect the test suite, not the library itself.
Why?
Look at some of the reports:
```
FAIL: [case 29] - , :  
    from EXTENDED_YEAR=4638,MONTH=3,DATE=30,IS_LEAP_MONTH=0: roll [[[ MONTH=15 ]]]
    >>> EXTENDED_YEAR=4638,MONTH=4,DATE=30,IS_LEAP_MONTH=0
    Differences: MONTH=5 not 4, DATE=29 not 30, ', status: U_USING_DEFAULT_WARNING
```
There are MONTH differences, but it is 3 != 4, not 4 and 5. And there are no DATE differences.

And that's the case for almost all failures:
```
FAIL: [case 18] - , :  
    from EXTENDED_YEAR=4640,MONTH=0,DATE=15,IS_LEAP_MONTH=0: roll [[[ MONTH=-1 ]]]
    >>> EXTENDED_YEAR=4640,MONTH=11,DATE=15,IS_LEAP_MONTH=0
    Differences: MONTH=0 not 11, EXTENDED_YEAR=4641 not 4640, ', status: U_USING_DEFAULT_WARNING
```
It complains about year and month, but only the months are different.

---
Sorry, I need to stop here, it's almost 2 am.

mihnita:
The first failing case One of the failing cases (in `calendar.txt`):
```
{ // normal
     "en_US@calendar=chinese",         
     "EXTENDED_YEAR=4642,MONTH=2,DATE=15,IS_LEAP_MONTH=0",  // ch  
     "roll",
     "MONTH=11", // month + 11
     "EXTENDED_YEAR=4642,MONTH=1,DATE=15,IS_LEAP_MONTH=0",  // ch  
}
```

The failure reported is:
```
 FAIL: [case 16] - , :
    from EXTENDED_YEAR=4642,MONTH=2,DATE=15,IS_LEAP_MONTH=0: roll [[[ MONTH=11 ]]]
    >>> EXTENDED_YEAR=4642,MONTH=1,DATE=15,IS_LEAP_MONTH=0
    Differences: MONTH=0 not 1, ', status: U_USING_DEFAULT_WARNING
```
Looks like "from" is the input, and ">>>" is the expected, not the actual result.
The actual result is not displayed, only the delta?

But how can a month of 2 + roll month = 11 result in a month of 1 (or, erroneously 0), without changing the year?

I'm lost. And I really need to stop :-)

roubert:
> But how can a month of 2 + roll month = 11 result in a month of 1 (or, erroneously 0), without changing the year?
> 
> I'm lost.

Mysterious indeed!

The way I see it, there are only a few main suspects in `DataDrivenCalendarTest::testOps()` at this point in time, in the order they appear in the code:

1. `fromCalendar`, set by `fromSet.setOnCalendar()`.
2. `toCalendar`, set by `toCalendar->roll()`. ← _prime suspect_
3. `toSet.matches()`, returns false here, what input does it get?
4. `diffSet.diffFrom()`, does this diff accurately reflect the input?

If I were with your debugger, I'd set a breakpoint on each of those calls, run the first failing test case and then for each of those calls verify their inputs and outputs. Then I'd expect at least one of them to have either unexpected inputs or outputs.

I guess that one of them has expected input but unexpected output (as there's more code that could possibly go wrong below than above in the call stack), in which case I'd set a new breakpoint there instead and drill down, rinse and repeat.

eggrobin:
I looked at it under the debugger in both Debug and Release, and in Debug on the first failing test case `hasLeapMonthBetweenWinterSolstices` is true, whereas it is false in Release, which explains the failure.

Now to figure out who is messing with `hasLeapMonthBetweenWinterSolstices`…

eggrobin:
I tried tracing `dynamic_cast<ChineseCalendar*>(fromCalendar)->hasLeapMonthBetweenWinterSolstices`. This is false initially.

After
https://github.com/unicode-org/icu/blob/ab72ab1d4a3c3f9beeb7d92b0c7817ca93dfdb04/icu4c/source/test/intltest/dadrcal.cpp#L214
it is true in Release, and false in Debug.
And then after
https://github.com/unicode-org/icu/blob/ab72ab1d4a3c3f9beeb7d92b0c7817ca93dfdb04/icu4c/source/test/intltest/dadrcal.cpp#L224
it is false in Release, and true in Debug.

???

eggrobin:
Looking just at the setOnCalendar, in both Debug and Release, we go through the following thrice: https://github.com/unicode-org/icu/blob/ab72ab1d4a3c3f9beeb7d92b0c7817ca93dfdb04/icu4c/source/i18n/chnsecal.cpp#L705

But in Debug the pair (`firstMoon`, `lastMoon`) is
1. (19 733, 20 058)
2. (19 349, 19 704)
3. (19 733, 20 058)

and in Release it is 
1. (19 733, 20 058)
2. (19 733, 20 058)
3. (19 349, 19 704)


The stack of the last two has handleComputeMonthStart from
https://github.com/unicode-org/icu/blob/ab72ab1d4a3c3f9beeb7d92b0c7817ca93dfdb04/icu4j/main/core/src/main/java/com/ibm/icu/util/Calendar.java#L6168-L6171
so we are depending on the evaluation order of those operands.

Changing that function to
```C++
    int32_t const nextYearStart =
        handleComputeMonthStart(eyear + 1, 0, false);
    int32_t const thisYearStart =
        handleComputeMonthStart(eyear, 0, false);
    return nextYearStart - thisYearStart;
```

fixes the issue.

roubert:
> […] so we are depending on the evaluation order of those operands.

This is an absolutely marvellous find! `ChineseCalendar::handleComputeMonthStart()` is marked `const` but internally typecasts the `this` pointer and modifies the object upon which it is called, so that the order in which these calls are made actually matters. Thank you!

roubert:
With the patch that @eggrobin came up with this now all works and is ready for review. It'd give me a warm fuzzy feeling if someone from Microsoft could be so kind and review and approve.

rp9-next (approved):
Thanks @eggrobin for finding this rare evaluation order mismatch!
Apologies, I have been OOF sick, couldn't help on this earlier.

roubert:
> With the patch that @eggrobin came up with this now all works and is ready for review.

The underlying problem, which that patch circumvented, has now been fixed by commit c5160765d40cf18751f4b3407e1cc7e8800c1cbe.

roubert:
> +100% to fix the 7 components, after this PR.

I think it'd be very nice if you'd do that.

PR #2770, opened 2024-01-03
ICU-22549 Revert "ICU-22549 Improve fuzzer to test more locale"

FrankYFTang:
This reverts commit 11d1148e5680fe62456659af85b37f690612ba73.

##### Checklist

### DIFF
--- a/icu4c/source/test/fuzzer/break_iterator_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/break_iterator_fuzzer.cpp
@@ -66,26 +66,6 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   for (int32_t p = bi->first(); p != icu::BreakIterator::DONE; p = bi->next()) {}
 
   utext_close(fuzzstr);
-
-  icu::Locale locale2(reinterpret_cast<const char*>(data));
-  switch (rnd8 % 5) {
-    case 0:
-      bi.reset(icu::BreakIterator::createWordInstance(locale2, status));
-      break;
-    case 1:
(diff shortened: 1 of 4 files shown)


richgillam (approved):
Really interested to see if this fixes the problem you're seeing...

FrankYFTang:
we got "build failure" on fuzzer . Revert to stop the bleeding first.

FrankYFTang:
The revert did fix that "build failure" as automatically find out by the oss-fuzz project. 

see  https://oss-fuzz-build-logs.storage.googleapis.com/index.html#icu 
notice the build of "1/2/2024, 11:56:04 PM  (Pacific Standard Time)" is bad which is based on git commit 1a60a038e14f0c56f50052c03fe76c4933cda339

then the build of "1/3/2024, 10:44:22 PM (Pacific Standard Time)" is good which is based on git commit 
539e8f41a35fec30fba4cfb3a65ea4db67978f85

and the diff between these two commits are only
52ff51e82c74adc52629b411d8cf6718c45e2210 ICU-22532 Compiler warning: conversion from 'double' to 'int32_t'.

and
539e8f41a35fec30fba4cfb3a65ea4db67978f85 ICU-22549 Revert "Improve fuzzer to test more locale"


I think what happen is the reverted part expose too many crash at once and make the system think it is a bad build. I will crate separated smaller impact PR of what got reverted and land into the fuzzer one by one so the system will not consider the PR itself is bad.

FrankYFTang:
I believe I found the bug of my PR. I incorrectly pass the fuzzer data, which is not guarantee to be null terminated into the Locale constructor which expect null terminated const char* . That is not an API issue but my bad in fuzzer code.

PR #2779, opened 2024-01-09
ICU-22623 Null terminate CharString extract result

FrankYFTang:
The CharString::extract method is not guaranteed to null terminate the string."If the string fits into dest but cannot be NUL-terminated (length()==capacity), then the error code is set to U_STRING_NOT_TERMINATED_WARNING." But the operations later assume the string is null terminated. So we should set extraction length one byte shorter and null terminate if the subStatus is U_STRING_NOT_TERMINATED_WARNING to avoid Stack-buffer-overflow read.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22623

### DIFF
--- a/icu4c/source/common/uresbund.cpp
+++ b/icu4c/source/common/uresbund.cpp
@@ -3097,8 +3097,18 @@ ures_getFunctionalEquivalent(char *result, int32_t resultCapacity,
     ures_initStackObject(&bund1);
     ures_initStackObject(&bund2);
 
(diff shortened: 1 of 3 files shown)


markusicu:
Looks plausible, but Fredrik's PR #2780 looks better.

FrankYFTang:
yield to https://github.com/unicode-org/icu/pull/2780

FrankYFTang:
> Looks plausible, but Fredrik's PR #2780 looks better.

Agree

PR #2782, opened 2024-01-11
ICU-22626 Fix leakage when 2 '=' in PluralRules

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22626

### DIFF
--- a/icu4c/source/i18n/plurrule.cpp
+++ b/icu4c/source/i18n/plurrule.cpp
@@ -656,6 +656,11 @@ PluralRuleParser::parse(const UnicodeString& ruleData, PluralRules *prules, UErr
         case tEqual:
             {
                 U_ASSERT(curAndConstraint != nullptr);
+                if (curAndConstraint->rangeList != nullptr) {
+                    // Already get a '='.
+                    status = U_UNEXPECTED_TOKEN;
+                    break;
+                }
                 LocalPointer<UVector32> newRangeList(new UVector32(status), status);
                 if (U_FAILURE(status)) {
                     break;
--- a/icu4c/source/test/intltest/plurults.cpp
+++ b/icu4c/source/test/intltest/plurults.cpp
@@ -69,6 +69,7 @@ void PluralRulesTest::runIndexedTest( int32_t index, UBool exec, const char* &na
     TESTCASE_AUTO(testFixedDecimal);
     TESTCASE_AUTO(testSelectTrailingZeros);
     TESTCASE_AUTO(testLocaleExtension);
+    TESTCASE_AUTO(testDoubleEqualSign);
     TESTCASE_AUTO_END;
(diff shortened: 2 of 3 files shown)


icu4c/source/i18n/plurrule.cpp
```diff
@@ -656,6 +656,10 @@ PluralRuleParser::parse(const UnicodeString& ruleData, PluralRules *prules, UErr
         case tEqual:
             {
                 U_ASSERT(curAndConstraint != nullptr);
```
roubert, line 658: It's illogical to have first this assertion, which indicates that this is something that shouldn't be possible, and then right after that have some code for handling it if it happens anyway. If this is something that actually can happen (as the fuzzer result suggests) then the assertion should be deleted (as it then isn't actually true).

If I follow the code correctly, this `PluralRuleParser::parse()` function is called from `PluralRules::createRules()` which is public API, meaning that a user can pass in any random string and that makes me think that all these assertions in this function are highly suspicious.

I would recommend you to widen the scope of ICU-22626 from just patching this one particular issue found by the fuzzer to investigating all the assertions in this function and replace all of those that can actually happen with proper error handling code.

icu4c/source/i18n/plurrule.cpp
```diff
@@ -905,6 +909,10 @@ AndConstraint::AndConstraint(const AndConstraint& other) {
         if (U_FAILURE(fInternalStatus)) {
             return;
         }
+        if (this->rangeList != nullptr) {
```
roubert, line 912: This doesn't make any sense to me. This code here is in the constructor for `class AndConstraint` and I can't see any code before this line that assigns anything to `this->rangeList` so the code in this if-statement will never be executed. Is that so, or what am I missing?

roubert (changes requested):

icu4c/source/i18n/plurrule.cpp
FrankYFTang: The assertion is on 
"curAndConstraint != nullptr", the if return is on 
"curAndConstraint->rangeList != nullptr" NOT "curAndConstraint != nullptr"
two different things.

icu4c/source/i18n/plurrule.cpp
FrankYFTang: you are right, I didn't realize it is in the constructor code.

FrankYFTang:
PTAL

icu4c/source/i18n/plurrule.cpp
roubert: Oops, now when I read it again it looks fine, sorry for the noise.

roubert (approved):
LGTM

PR #2785, opened 2024-01-15
ICU-22520 Switch to using CharString for calling uloc_setKeywordValue()

roubert:
##### Checklist

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -537,35 +537,26 @@ locale_getKeywordsStart(const char *localeID) {
 }
 
 /**
- * @param buf buffer of size [ULOC_KEYWORD_BUFFER_LEN]
  * @param keywordName incoming name to be canonicalized
  * @param status return status (keyword too long)
- * @return length of the keyword name
+ * @return the keyword name
  */
-static int32_t locale_canonKeywordName(char *buf, const char *keywordName, UErrorCode *status)
+static CharString locale_canonKeywordName(const char *keywordName, UErrorCode *status)
 {
-  int32_t keywordNameLen = 0;
+  CharString result;
 
   for (; *keywordName != 0; keywordName++) {
     if (!UPRV_ISALPHANUM(*keywordName)) {
       *status = U_ILLEGAL_ARGUMENT_ERROR; /* malformed keyword name */
-      return 0;
+      return result;
     }
-    if (keywordNameLen < ULOC_KEYWORD_BUFFER_LEN - 1) {
-      buf[keywordNameLen++] = uprv_tolower(*keywordName);
-    } else {
-      /* keyword name too long for internal buffer */
-      *status = U_INTERNAL_PROGRAM_ERROR;
-      return 0;
-    }
+    result.append(uprv_tolower(*keywordName), *status);
   }
-  if (keywordNameLen == 0) {
+  if (result.isEmpty()) {
     *status = U_ILLEGAL_ARGUMENT_ERROR; /* empty keyword name */
-    return 0;
   }
-  buf[keywordNameLen] = 0; /* terminate */
 
-  return keywordNameLen;
+  return result;
 }
 
 typedef struct {
@@ -739,8 +730,6 @@ ulocimp_getKeywordValue(const char* localeID,
 {
     const char* startSearchHere = nullptr;
     const char* nextSeparator = nullptr;
-    char keywordNameBuffer[ULOC_KEYWORD_BUFFER_LEN];
-    char localeKeywordNameBuffer[ULOC_KEYWORD_BUFFER_LEN];
 
     if(status && U_SUCCESS(*status) && localeID) {
       CharString tempBuffer;
@@ -751,7 +740,7 @@ ulocimp_getKeywordValue(const char* localeID,
         return;
       }
 
-      locale_canonKeywordName(keywordNameBuffer, keywordName, status);
+      CharString canonKeywordName = locale_canonKeywordName(keywordName, status);
       if(U_FAILURE(*status)) {
         return;
       }
@@ -773,7 +762,6 @@ ulocimp_getKeywordValue(const char* localeID,
       /* find the first keyword */
       while(startSearchHere) {
           const char* keyValueTail;
-          int32_t keyValueLen;
 
           startSearchHere++; /* skip @ or ; */
           nextSeparator = uprv_strchr(startSearchHere, '=');
@@ -795,25 +783,21 @@ ulocimp_getKeywordValue(const char* localeID,
               *status = U_ILLEGAL_ARGUMENT_ERROR; /* empty keyword name in passed-in locale */
               return;
           }
-          keyValueLen = 0;
+          CharString localeKeywordName;
           while (startSearchHere < keyValueTail) {
             if (!UPRV_ISALPHANUM(*startSearchHere)) {
               *status = U_ILLEGAL_ARGUMENT_ERROR; /* malformed keyword name */
               return;
             }
-            if (keyValueLen < ULOC_KEYWORD_BUFFER_LEN - 1) {
-              localeKeywordNameBuffer[keyValueLen++] = uprv_tolower(*startSearchHere++);
-            } else {
-              /* keyword name too long for internal buffer */
-              *status = U_INTERNAL_PROGRAM_ERROR;
-              return;
-            }
+            localeKeywordName.append(uprv_tolower(*startSearchHere++), *status);
           }
-          localeKeywordNameBuffer[keyValueLen] = 0; /* terminate */
+          if (U_FAILURE(*status)) {
+              return;
+          }
 
           startSearchHere = uprv_strchr(nextSeparator, ';');
 
-          if(uprv_strcmp(keywordNameBuffer, localeKeywordNameBuffer) == 0) {
+          if (canonKeywordName == localeKeywordName) {
                /* current entry matches the keyword. */
              nextSeparator++; /* skip '=' */
               /* First strip leading & trailing spaces (TC decided to tolerate these) */
@@ -909,12 +893,7 @@ ulocimp_setKeywordValue(const char* keywords,
                         UErrorCode* status)
 {
     /* TODO: sorting. removal. */
-    int32_t keywordNameLen;
-    int32_t keywordValueLen;
     int32_t needLen = 0;
-    char keywordNameBuffer[ULOC_KEYWORD_BUFFER_LEN];
-    char keywordValueBuffer[ULOC_KEYWORDS_CAPACITY+1];
-    char localeKeywordNameBuffer[ULOC_KEYWORD_BUFFER_LEN];
     int32_t rc;
     const char* nextSeparator = nullptr;
     const char* nextEqualsign = nullptr;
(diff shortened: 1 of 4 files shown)


roubert:
It'd be very helpful if you could find time to review this sometime soon.

icu4c/source/common/uloc.cpp
```diff
@@ -848,20 +848,59 @@ uloc_setKeywordValue(const char* keywordName,
...
+
+    char* keywords = const_cast<char*>(locale_getKeywordsStart(buffer));
+    int32_t baseLen = keywords == nullptr ? bufLen : keywords - buffer;
+    CheckedArrayByteSink sink(keywords == nullptr ? buffer + bufLen : keywords,
+                              bufferCapacity - baseLen - 1);
+    int32_t reslen = ulocimp_setKeywordValue(
+            keywords, keywordName, keywordValue, sink, status);
+
+    if (U_FAILURE(*status)) {
+        return reslen > 0 ? reslen + baseLen : reslen;
```
markusicu, line 860: Should this not be always `reslen + baseLen`?
Especially when there is an overflow, we should report how much space is needed.

icu4c/source/common/uloc.cpp
```diff
@@ -848,20 +848,59 @@ uloc_setKeywordValue(const char* keywordName,
...
+    }
+
+    // See the documentation for this function, it's guaranteed to never
+    // overflow the buffer but instead abort with BUFFER_OVERFLOW_ERROR.
+    U_ASSERT(!sink.Overflowed());
```
markusicu, line 865: Why this assertion?
Yes, we don't want to write beyond the capacity, but rather than asserting that the result cannot be too long, we should check for that and return with an overflow error and a length of `baseLen + sink.NumberOfBytesAppended()` right?

icu4c/source/common/uloc.cpp
```diff
@@ -848,20 +848,59 @@ uloc_setKeywordValue(const char* keywordName,
...
+    U_ASSERT(reslen >= 0);
```
markusicu, line 866: Why this assertion? The new function returns 0 if the buffer starts without keywords and keywordValueLen == 0. Is that wrong?

icu4c/source/common/uloc.cpp
```diff
@@ -906,34 +939,33 @@ uloc_setKeywordValue(const char* keywordName,
     }
     keywordValueBuffer[keywordValueLen] = 0; /* terminate */
 
-    startSearchHere = (char*)locale_getKeywordsStart(buffer);
-    if(startSearchHere == nullptr || (startSearchHere[1]==0)) {
+    if (keywords == nullptr || keywords[1] == '\0') {
         if(keywordValueLen == 0) { /* no keywords = nothing to remove */
             U_ASSERT(*status != U_STRING_NOT_TERMINATED_WARNING);
-            return bufLen;
+            return 0;
```
markusicu, line 935: Should this return 1 if keywords != nullptr because there is something (a separator character?) in keywords[0]?

icu4c/source/common/uloc.cpp
```diff
@@ -848,20 +848,59 @@ uloc_setKeywordValue(const char* keywordName,
...
+    if(bufferCapacity<bufLen) {
+        /* The capacity is less than the length?! Is this NUL terminated? */
+        *status = U_ILLEGAL_ARGUMENT_ERROR;
+        return 0;
+    }
+
+    char* keywords = const_cast<char*>(locale_getKeywordsStart(buffer));
+    int32_t baseLen = keywords == nullptr ? bufLen : keywords - buffer;
+    CheckedArrayByteSink sink(keywords == nullptr ? buffer + bufLen : keywords,
+                              bufferCapacity - baseLen - 1);
```
markusicu, line 855: Please precede this with a comment like
```
// Remove -1 from the capacity so that this function can guarantee NUL termination.
```

icu4c/source/i18n/ucal.cpp
```diff
@@ -167,21 +168,15 @@ ucal_open(  const char16_t*  zoneID,
...
-      }
-      uprv_strcpy(localeBuf, locale);
-      uloc_setKeywordValue("calendar", "gregorian", localeBuf, ULOC_LOCALE_IDENTIFIER_CAPACITY, status);
+      CharString localeBuf(locale, *status);
+      ulocimp_setKeywordValue("calendar", "gregorian", localeBuf, status);
       if (U_FAILURE(*status)) {
           return nullptr;
       }
-      return (UCalendar*)Calendar::createInstance(zone.orphan(), Locale(localeBuf), *status);
+      return (UCalendar*)Calendar::createInstance(zone.orphan(), Locale(localeBuf.data()), *status);
```
markusicu, line 179: Since we are eventually constructing a C++ Locale, wouldn't it be easier to do that first and then call its .setKeywordValue()?

icu4c/source/common/uloc.cpp
roubert: We know that `ulocimp_setKeywordValue()` must promise to never cause an overflow so this assertion is there to verify that this promise is being kept (and give a nice clear error message in case there's a bug that causes an overflow that should never happen).

icu4c/source/common/uloc.cpp
roubert: Yes, we know that this is what's supposed to happen, this assertion is there to verify that this really is what happens (for `reslen + baseLen` on the following line would be really nasty to debug if there was some bug that caused `reslen` to be negative).

icu4c/source/common/uloc.cpp
roubert: Done.

icu4c/source/common/uloc.cpp
roubert: It's not quite that simple, for historical reasons the return value here has a dual role, if it's positive it's a length, otherwise it's an error code. I've added a source code comment about that now.

icu4c/source/common/uloc.cpp
roubert: I think it should be the way it is, the zero return value here is the status code for _do nothing_.

icu4c/source/i18n/ucal.cpp
roubert: That sounds like a good idea, but let's make one change at a time, there's a nearly endless amount of improvements that could be done in this code base and we have bigger fish to fry.

icu4c/source/common/uloc.cpp
markusicu: I still don't understand. This function here is public API, and the [docs](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/uloc_8h.html#aee57778b4fdc2233f278bb6359205d4e) don't say anything about negative return values. The return value is "the length needed for the buffer".

icu4c/source/common/uloc.cpp
markusicu: Sorry, I don't follow. If the new function writes too much to the sink, then the CheckedArrayByteSink here will report Overflowed() but will not have written past the end of the array. When this happens, we need to set a buffer-overflow-error (if there isn't a failure UErrorCode already) and return with the length of the full output. What am I missing?

icu4c/source/common/uloc.cpp
markusicu: Is that documented somewhere? When this code was part of the API function, it used to return bufLen for the length of the result. Where does "less-or-equal-zero is a status code" come from?

icu4c/source/common/uloc.cpp
roubert: I don't know but I guess that whoever wrote that documentation expected the reader to guess that the return value only represented a length if it was a value that could mean a length. If you look at the current implementation, the one that I intend to replace with this PR, it begins like this:

```c++
    if(U_FAILURE(*status)) {
        return -1;
    }
    if (*status == U_STRING_NOT_TERMINATED_WARNING) {
        *status = U_ZERO_ERROR;
    }
    if (keywordName == nullptr || keywordName[0] == 0 || bufferCapacity <= 1) {
        *status = U_ILLEGAL_ARGUMENT_ERROR; 
        return 0; 
    }
```

Documented or not, that's the behaviour that this PR intends to preserve.

icu4c/source/common/uloc.cpp
roubert: This, on the other hand, is documented:

> NOTE: Unlike almost every other ICU function which takes a buffer, this function will NOT truncate the output text, and will not update the buffer with unterminated text setting a status of U_STRING_NOT_TERMINATED_WARNING. If a BUFFER_OVERFLOW_ERROR is received, it means a terminated version of the updated locale ID would not fit in the buffer, and the original buffer is untouched. This is done to prevent incorrect or possibly even malformed locales from being generated and used.

In order to preserve that behaviour, nothing at all gets written to the sink if there isn't room to write everything that needs to be written (and therefore no overflow can happen).

icu4c/source/common/uloc.cpp
roubert: I don't think that's documented anywhere, it's just preserving existing behaviour (cf. my other comment with the code snippet).

icu4c/source/common/uloc.cpp
markusicu: I think I see what you are trying to do here.
It looks like you are trying to match the code returning a variety of values when there is a failure. However, the return value is undefined when there is a failure other than a buffer overflow.

So let's not try to go out of our way to preserve returning -1 in one case and 0 in another case, because that makes for weird conditions that are hard to understand/explain/document.

If the function succeeds, the return value needs to be the length of the output string.
If the function fails with an overflow, the return value needs to be the length of the output string that it wants to write.
If the function fails for other reasons, do something simple and understandable.

icu4c/source/common/uloc.cpp
markusicu: I see, so the sink does not see anything which means it can't have Overflowed().

icu4c/source/common/uloc.cpp
```diff
@@ -848,20 +848,61 @@ uloc_setKeywordValue(const char* keywordName,
...
+    int32_t reslen = ulocimp_setKeywordValue(
+            keywords, keywordName, keywordValue, sink, status);
+
+    if (U_FAILURE(*status)) {
+        // A positive return value is a length, otherwise it's an error code.
+        return reslen > 0 ? reslen + baseLen : reslen;
+    }
+
+    // See the documentation for this function, it's guaranteed to never
+    // overflow the buffer but instead abort with BUFFER_OVERFLOW_ERROR.
```
markusicu, line 866: Please add:
```
// In this case, nothing has been written to the sink, so it cannot have Overflowed().
```

roubert: Done.

icu4c/source/common/uloc.cpp
roubert: I don't like changing behaviour and implementation at the same time. It's much easier to backtrack in case anything goes wrong if those changes are done separately from each other and one never knows who might depend on any undocumented behaviour. But if you like, I can do a once-over on all these functions once the memory management refactoring is done and make them all return the same simple return value (zero, preferrably) in case of error.

markusicu: I don't like leaving the confusing condition and comment, but I will give in here to move forward.
I also don't think we need a global cleanup for "don't care" return values.
I would like you to come back to *just this pair of functions* in another PR and simplify the return value handling.

markusicu (approved):

PR #2792, opened 2024-01-18
ICU-22637 Rewrite custom timezone parser

FrankYFTang:
The old code use NumberFormat to parse and have the following problem
1. It accept non ASCII digits when the default locale is in a locale which use native non ASCII digits
2. It accept timezone offset in Scientific notation such as GMT-4E23 or GMT-2E03 which should be rejected
3. It may cause integer overflow in case 3 such as GMT-4E58

This PR remove the NumberFormat::parse dependency

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22637

### DIFF
--- a/icu4c/source/i18n/timezone.cpp
+++ b/icu4c/source/i18n/timezone.cpp
@@ -42,7 +42,9 @@
 #include "unicode/utypes.h"
 #include "unicode/ustring.h"
 #include "uassert.h"
+#include "uinvchar.h"
 #include "ustr_imp.h"
+#include "util.h"
 
 #ifdef U_DEBUG_TZ
 # include <stdio.h>
@@ -75,7 +77,6 @@ static char gStrBuf[256];
 #include "unicode/gregocal.h"
 #include "unicode/ures.h"
 #include "unicode/tzfmt.h"
-#include "unicode/numfmt.h"
 #include "gregoimp.h"
 #include "uresimp.h" // struct UResourceBundle
 #include "olsontz.h"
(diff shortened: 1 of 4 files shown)


yumaoka:
Can you do the same for Java version?

icu4c/source/i18n/timezone.cpp
```diff
@@ -1369,123 +1370,102 @@ TimeZone::getCustomID(const UnicodeString& id, UnicodeString& normalized, UError
...
+    min = 0;
+    sec = 0;
+
+    if (id[GMT_ID_LENGTH] == MINUS /*'-'*/) {
+        sign = -1;
+    } else if (id[GMT_ID_LENGTH] != PLUS /*'+'*/) {
+        return false;
+    }
+
+    // Support the following cases
```
yumaoka, line 1395: You should move the comment for patterns without `:` into if block.

yumaoka (approved):
LGTM. One minor comment

icu4c/source/i18n/timezone.cpp
FrankYFTang: done

FrankYFTang:
PTAL Java added

richgillam (approved):
Looks great.

icu4c/source/test/intltest/tztest.cpp
```diff
@@ -1171,36 +1171,56 @@ void TimeZoneTest::TestCustomParse()
...
+        {u"GMT-9E00",        kUnparseable}, // ICU-22637
+        {u"GMT-0X0F",        kUnparseable}, // ICU-22637
+        {u"GMT-0x0F",        kUnparseable}, // ICU-22637
+        {u"GMT-0x12",        kUnparseable}, // ICU-22637
+        {u"GMT-B111",        kUnparseable}, // ICU-22637
+        {u"GMT-b111",        kUnparseable}, // ICU-22637
+        {u"GMT-0b11",        kUnparseable}, // ICU-22637
+        {u"GMT-๑๒",          kUnparseable}, // ICU-22637
+        {u"GMT-๑๒:๓๔",       kUnparseable}, // ICU-22637
+        {u"GMT+๑๒:๓๔:๕๖",    kUnparseable}, // ICU-22637
```
sffc, line 1223: Why are we rejecting non-ASCII digits? We had been accepting them and they could reasonably appear in a localized string.

FrankYFTang: yes, you are correct. This is my bad. I have a misunderstanding of the prior behavior. I should change that back.

FrankYFTang: Please review my fix in https://github.com/unicode-org/icu/pull/2803

PR #2793, opened 2024-01-18
ICU-22636 Return U_BRK_RULE_SYNTAX when status number is too large

FrankYFTang:
To avoid overflow the parsing, we now return U_BRK_RULE_SYNTAX if the status value is in the rule is too large and overflow the int32_t

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22636

### DIFF
--- a/icu4c/source/common/rbbiscan.cpp
+++ b/icu4c/source/common/rbbiscan.cpp
@@ -530,7 +530,13 @@ UBool RBBIRuleScanner::doParseActions(int32_t action)
             n = fNodeStack[fNodeStackPtr];
             uint32_t v = u_charDigitValue(fC.fChar);
             U_ASSERT(v < 10);
-            n->fVal = n->fVal*10 + v;
+            int64_t updated = static_cast<int64_t>(n->fVal)*10 + v;
+            // Avoid overflow n->fVal
+            if (updated > INT32_MAX) {
+                error(U_BRK_RULE_SYNTAX);
+                break;
+            }
+            n->fVal = static_cast<int32_t>(updated);
             break;
         }
 
--- a/icu4c/source/test/intltest/rbbitst.cpp
+++ b/icu4c/source/test/intltest/rbbitst.cpp
@@ -149,6 +149,7 @@ void RBBITest::runIndexedTest( int32_t index, UBool exec, const char* &name, cha
     TESTCASE_AUTO(TestBug22584);
     TESTCASE_AUTO(TestBug22585);
     TESTCASE_AUTO(TestBug22602);
+    TESTCASE_AUTO(TestBug22636);
 
 #if U_ENABLE_TRACING
     TESTCASE_AUTO(TestTraceCreateCharacter);
@@ -5887,6 +5888,19 @@ void RBBITest::TestBug22602() {
     UParseError pe {};
     UErrorCode ec {U_ZERO_ERROR};
     RuleBasedBreakIterator bi(rule, pe, ec);
+}
+
+void RBBITest::TestBug22636() {
+    UParseError pe {};
+    UErrorCode ec {U_ZERO_ERROR};
+    RuleBasedBreakIterator bi(u"A{77777777777777};", pe, ec);
+    assertEquals(WHERE, ec, U_BRK_RULE_SYNTAX);
+    ec = U_ZERO_ERROR;
+    RuleBasedBreakIterator bi2(u"A{2147483648};", pe, ec);
+    assertEquals(WHERE, ec, U_BRK_RULE_SYNTAX);
+    ec = U_ZERO_ERROR;
+    RuleBasedBreakIterator bi3(u"A{2147483647};", pe, ec);
+    assertEquals(WHERE, ec, U_ZERO_ERROR);
 }
 
 void RBBITest::TestBug22584() {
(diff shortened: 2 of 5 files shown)


richgillam (dismissed):
Been a long time since I've looked at this code, so I may not know what I'm talking about, but this looks okay to me.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -5889,6 +5890,13 @@ void RBBITest::TestBug22602() {
     RuleBasedBreakIterator bi(rule, pe, ec);
 }
 
+void RBBITest::TestBug22636() {
+    UParseError pe {};
+    UErrorCode ec {U_ZERO_ERROR};
+    RuleBasedBreakIterator bi(u"A{77777777777777}", pe, ec);
```
aheninger, line 5896: It would be good to check that 2147483647 is accepted and that 2147483648 fails, just to verify that the check is doing the right thing on the boundary.

aheninger, line 5896: This test rule is missing its terminating semicolon, which causes the break iterator constructor to always return U_BRK_RULE_SYNTAX, no matter what the value of the integer tag.
```RuleBasedBreakIterator bi(u"A{2147483648};", pe, ec);```

**Review by aheninger — CHANGES_REQUESTED** [review]

File: icu4c/source/common/rbbiscan.cpp [hunk 2]
```diff
@@ -530,6 +530,11 @@ UBool RBBIRuleScanner::doParseActions(int32_t action)
             n = fNodeStack[fNodeStackPtr];
             uint32_t v = u_charDigitValue(fC.fChar);
             U_ASSERT(v < 10);
+            // Avoid overflow n->fVal
+            if (n->fVal > static_cast<int32_t>((INT32_MAX - v) / 10)) {
```
aheninger, line 534: It might be clearer to compute the value in a temporary of type int64_t. The 32 bit overflow check inside the loop is then much simpler
```temp > INT32_MAX```

File: icu4c/source/test/intltest/rbbitst.cpp (hunk 1 above)
Comment (FrankYFTang) [reply icu4c/source/test/intltest/rbbitst.cpp]: DONE. also add Java fix

File: icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java [hunk 3]
```diff
@@ -1002,6 +1002,28 @@ public void TestBug22602() {
...
     }
+    @Test
+    public void TestBug22636() {
+        try {
+            RuleBasedBreakIterator bi = new RuleBasedBreakIterator("A{77777777777777};");
+        }
+        catch (IllegalArgumentException e) {
+            // expected exception with a long string followed by a ';'.
+        }
+        catch(StackOverflowError e) {
```
aheninger, line 1013: I don't think we need to check for StackOverflowError here.
Looks like it got here from a copy/paste from TestBug22602, which was throwing an unexpected StackOverflow, but that was never the situation with these new test cases for int overflow problems.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java
```diff
@@ -1002,6 +1002,28 @@ public void TestBug22602() {
...
+            fail("TestBug22636: Unexpected exception while new RuleBasedBreakIterator() with a large status value inside {}: " + e);
+        }
+        try {
+            RuleBasedBreakIterator bi2 = new RuleBasedBreakIterator("A{2147483648};");
+        }
+        catch (IllegalArgumentException e) {
+            // expected exception with a long string followed by a ';'.
+        }
+        catch(StackOverflowError e) {
```
aheninger, line 1022: Same as above.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java
```diff
@@ -1002,6 +1002,28 @@ public void TestBug22602() {
             fail("TestBug22602: Unexpected exception while new RuleBasedBreakIterator() with a long string followed by a ';': " + e);
         }
     }
+    @Test
+    public void TestBug22636() {
+        try {
+            RuleBasedBreakIterator bi = new RuleBasedBreakIterator("A{77777777777777};");
+        }
+        catch (IllegalArgumentException e) {
+            // expected exception with a long string followed by a ';'.
```
aheninger, line 1011: Incorrect comment, pertains to Bug22602, not this one.
Say something about numeric status value overflowing an int.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java
```diff
@@ -1002,6 +1002,28 @@ public void TestBug22602() {
...
+        }
+        catch(StackOverflowError e) {
+            fail("TestBug22636: Unexpected exception while new RuleBasedBreakIterator() with a large status value inside {}: " + e);
+        }
+        try {
+            RuleBasedBreakIterator bi2 = new RuleBasedBreakIterator("A{2147483648};");
+        }
+        catch (IllegalArgumentException e) {
+            // expected exception with a long string followed by a ';'.
```
aheninger, line 1020: Same as above.

aheninger (changes requested):

PR #2797, opened 2024-01-18
ICU-22639 Clone the calendar so we don't mess with the real one

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22639

### DIFF
--- a/icu4c/source/i18n/chnsecal.cpp
+++ b/icu4c/source/i18n/chnsecal.cpp
@@ -325,8 +325,6 @@ const UFieldResolutionTable* ChineseCalendar::getFieldResolutionTable() const {
  * @stable ICU 2.8
  */
 int32_t ChineseCalendar::handleComputeMonthStart(int32_t eyear, int32_t month, UBool useMonth) const {
-    ChineseCalendar *nonConstThis = (ChineseCalendar*)this; // cast away const
-
     // If the month is out of range, adjust it into range, and
     // modify the extended year value accordingly.
     if (month < 0 || month > 11) {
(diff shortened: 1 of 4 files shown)


icu4c/source/i18n/chnsecal.cpp
```diff
@@ -325,8 +325,6 @@ const UFieldResolutionTable* ChineseCalendar::getFieldResolutionTable() const {
  * @stable ICU 2.8
  */
 int32_t ChineseCalendar::handleComputeMonthStart(int32_t eyear, int32_t month, UBool useMonth) const {
-    ChineseCalendar *nonConstThis = (ChineseCalendar*)this; // cast away const
```
eggrobin, line 328: I am most happy to see one of these go away.
Twenty more to go.

eggrobin (approved):

icu4c/source/test/intltest/caltest.cpp
```diff
@@ -5628,6 +5630,35 @@ void CalendarTest::TestFirstDayOfWeek() {
...
+        status);
+    if (failure(status, "Calendar::createInstance")) return;
+
+    // This test case is a friend of ChineseCalendar and may access internals.
+    const ChineseCalendar& chinese =
+        *dynamic_cast<ChineseCalendar*>(calendar.getAlias());
+
+    // The initial value of hasLeapMonthBetweenWinterSolstices should be false.
+    assertFalse("hasLeapMonthBetweenWinterSolstices [#1]",
+                chinese.hasLeapMonthBetweenWinterSolstices);
```
eggrobin, line 5652: Is this assertion going to be broken by the passage of time?

icu4c/source/i18n/chnsecal.cpp
roubert: As far as I can see, this is the only one that tries to save, modify and then restore the internal state (please shout if you can find anywhere else where that is being done). While it would be nice to get rid of the rest too, I wouldn't be all too concerned.

eggrobin:
I am surprised that the tests pass, given that the following passes on my machine:
```C++
        LocalPointer<Calendar> calendar(
            Calendar::createInstance(Locale("en_US@calendar=chinese"), status),
            status);
        calendar->set(4642, 2, 15);
        assertEquals("month", calendar->get(UCAL_MONTH, status), 2);
        calendar->roll(UCAL_MONTH, 11, status);
        assertEquals("rolled month", 0, calendar->get(UCAL_MONTH, status));
```
and given that we have this data-driven test:
https://github.com/unicode-org/icu/blob/981057143ceb68bcf6dc6c81a3c279645f98d6a9/icu4c/source/test/testdata/calendar.txt#L295-L301

Note that 1, not 0, is expected. What am I missing?

icu4c/source/test/intltest/caltest.cpp
roubert: The `ChineseCalendar` constructor is supposed to always set this to `false`, regardless of when the constructor is called, so no, this should not be affected by the passage of time.

eggrobin: Ah right, this is only going to be set when the calendar is completed.

roubert:
> What am I missing?

Maybe you're not thinking about the fact that all existing code and tests rely on the ordering of all calls to `handleComputeMonthStart()` being such that `hasLeapMonthBetweenWinterSolstices` never happens to be left with the wrong value and then this PR won't change any behaviour, which is why all existing code and tests continue to work as if nothing happened also after this PR.

eggrobin:
Ah, I would have expected the
```C++
        assertEquals("month", calendar->get(UCAL_MONTH, status), 2);
```
in my test to leave `hasLeapMonthBetweenWinterSolstices` with the right value, since it calls `complete`.

But there are calls to handleComputeMonthStart in there. Apparently _those_ are badly ordered.

eggrobin:
> Apparently _those_ are badly ordered.

Hm, no, patching in this PR, that test still passes, the rolled month is still 0. I remain confused.

roubert:
> I remain confused.

The difference between the data driven test and your local test code is that you initialize the `Calendar` object differently, particularly you're setting totally different year fields, `UCAL_YEAR` vs. `UCAL_EXTENDED_YEAR`.

I guess you'll become less confused if you update your local test code to this, which should make your `Calendar` object identical to the one created in the data driven test:

```
    calendar->set(UCAL_EXTENDED_YEAR, 4642);
    calendar->set(UCAL_MONTH, 2);
    calendar->set(UCAL_DAY_OF_MONTH, 2);
    calendar->set(UCAL_IS_LEAP_MONTH, 0);
```

eggrobin:
> […] you're setting totally different year fields, UCAL_YEAR vs. UCAL_EXTENDED_YEAR.
>
> I guess you'll become less confused if you update your local test code to this, which should make your `Calendar` object identical to the one created in the data driven test:
>
> ```C++
>    calendar->set(UCAL_EXTENDED_YEAR, 4642);
>    calendar->set(UCAL_MONTH, 2);
>    calendar->set(UCAL_DAY_OF_MONTH, 15);
>    calendar->set(UCAL_IS_LEAP_MONTH, 0);
>```

I am indeed less confused now, thanks.

PR #2800, opened 2024-01-19
ICU-22641 Change ref to pointer so we will not def nullptr

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22641


FrankYFTang:
I just cannot find a better way to fix this. If you know one please let me know.

FrankYFTang:
or maybe we should just remove nfd and nfcImpl from the class and make them file internal static pointer and init once it on the construction time?

icu4c/source/common/caniter.cpp
```diff
@@ -64,6 +65,25 @@ U_NAMESPACE_BEGIN
 
 UOBJECT_DEFINE_RTTI_IMPLEMENTATION(CanonicalIterator)
 
+
+static icu::UInitOnce gNormalizerInitOnce {};
+const static Normalizer2 *gNfd = nullptr;  // Singleton. Do not delete
+const static Normalizer2Impl *gNfcImpl = nullptr;  // Singleton. Do not delete
```
markusicu, line 71: Redundant. Normalizer2::getNFCInstance() already uses init-once singletons.
https://github.com/unicode-org/icu/blob/main/icu4c/source/common/normalizer2.cpp#L263

markusicu (changes requested):
Way overkill. AFAICT the only failure mode is a memory-allocation-error, and if we get a UBSan failure because of that, then let's just turn the const references into const pointers but not go further.

markusicu (approved):

PR #2801, opened 2024-01-22
ICU-22532 Move the definition of _POSIX_C_SOURCE to the Makefile

roubert:
This is already set in `Makefile.in` and therefore results in a macro redefined warning if also defined in the source file. It seems that setting this in the Makefile was how it was originally intended do be done (but then it was just never updated there).

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22532


FrankYFTang (approved):
thanks for working on this

PR #2802, opened 2024-01-22
ICU-22520 Bugfix: Use macro parameter name instead of variable name

roubert:
This was added by commit a998c7cac4d53415e8e699717fc641d6617c14b3 and must surely be a mistake. The old code that was replaced used the macro parameter name `id` but the new code replacing it instead (without any explanation) used the variable name `localeID` which at that point in time was the variable name that all invocations of the macro used as the argument (so the two names happened to be interchangeable in this context).

##### Checklist


markusicu (approved):
right, a macro should very rarely use an identifier from outside its own context

PR #2806, opened 2024-01-23
ICU-22633 Fix integer overflow inside Calendar code

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22633

### DIFF
--- a/icu4c/source/i18n/cecal.cpp
+++ b/icu4c/source/i18n/cecal.cpp
@@ -76,10 +76,27 @@ CECalendar::operator=(const CECalendar& right)
 // Calendar framework
 //-------------------------------------------------------------------------
 
-int32_t
+int64_t
 CECalendar::handleComputeMonthStart(int32_t eyear,int32_t emonth, UBool /*useMonth*/) const
 {
-    return ceToJD(eyear, emonth, 0, getJDEpochOffset());
+    int64_t year64 = eyear;
+    // handle month > 12, < 0 (e.g. from add/set)
+    if ( emonth >= 0 ) {
+        year64 += emonth/13;
+        emonth %= 13;
+    } else {
+        ++emonth;
+        year64 += emonth/13 - 1;
+        emonth = emonth%13 + 12;
+    }
+
+    return (
+        getJDEpochOffset()                    // difference from Julian epoch to 1,1,1
+        + 365LL * year64                      // number of days from years
+        + ClockMath::floorDivide(year64, 4LL) // extra day of leap year
+        + 30 * emonth                         // number of days from months (months are 0-based)
+        - 1                                   // number of days for present month (1 based)
+        );
 }
 
 int32_t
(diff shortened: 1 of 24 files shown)


icu4c/source/i18n/hebrwcal.cpp
```diff
@@ -423,7 +423,15 @@ int32_t HebrewCalendar::startOfYear(int32_t year, UErrorCode &status)
...
-        CalendarCache::put(&gCache, year, day, status);
+        if (day > INT32_MAX || day < INT32_MIN) {
+            status = U_ILLEGAL_ARGUMENT_ERROR;
+            return 0;
+        }
+        CalendarCache::put(&gCache, year, static_cast<int32_t>(day), status);
+    }
+    if (day > INT32_MAX || day < INT32_MIN) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return 0;
```
richgillam, line 434: You have this same check twice in the same function, inside and outside in `if` block.  Can you consolidate?  And if not, especially if you've got this same check in a bunch of other places, is it worth making this a macro or an inline function for readability?

richgillam (commented):
Generally looks good, but I'm worried there may be compatibility issues.  Can you explain how you've guarded against that?

icu4c/source/i18n/cecal.cpp
```diff
@@ -76,7 +76,7 @@ CECalendar::operator=(const CECalendar& right)
 // Calendar framework
 //-------------------------------------------------------------------------
 
-int32_t
+int64_t
```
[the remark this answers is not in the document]
FrankYFTang: > Are these all completely internal functions, or are some of them API or part of the subclassing interface?
yes, I only change handleComputeMonthStart is marked as @internal so it is not part of the API. 

> can't you still get integer overflow with int64_t?
Since the input are in type int32_t the computation result of these two int32_t should be still < the limitation of int64_t

> Are you doing anything to prevent that?
Let me think about this.

icu4c/source/i18n/hebrwcal.cpp
FrankYFTang: ok, since I check that before I put into cache I think this one is not necessary

icu4c/source/i18n/islamcal.cpp
```diff
@@ -935,9 +952,17 @@ void IslamicUmalquraCalendar::handleComputeFields(int32_t julianDay, UErrorCode
     }
 
     dayOfMonth = (days - monthStart(year, month)) + 1;
+    if (dayOfMonth > INT32_MAX || dayOfMonth < INT32_MIN) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return;
+    }
```
richgillam, line 958: I see three instances in this PR of this pattern, where you're checking some integer value to make sure it's between `INT32_MIN` and `INT32_MAX` and setting the status code to `U_ILLEGAL_ARGUMENT_ERROR` and returning if it's not.  There's also two similar-looking spots at the top of the PR where you're checking `uprv_add32_overflow()` and returning `U_ILLEGAL_ARGUMENT_ERROR`.  It kind of feels like there might be a way to factor these out into a utility function: `FAIL_IF_OVERFLOW(dayOfMonth)` or something like that.  Just to improve readability.  But I'm not going to hold up the PR over that.

richgillam (dismissed):
LGTM

FrankYFTang:
PTAL

icu4c/source/i18n/islamcal.cpp
FrankYFTang: > FAIL_IF_OVERFLOW(dayOfMonth) 

that will be hard because the variable name of the status is not always the same and the return type of the function are also not the same (some are void and some are int32_t)

richgillam: You're probably right, and it's definitely not a hill I want to die on.

richgillam (approved):
Nice job!  I'm glad you did this.

PR #2810, opened 2024-01-30
ICU-22651 Refactor U_DEFINE_LOCAL_OPEN_POINTER into a template

roubert:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22651


aheninger (approved):
This all looks good to me.

icu4c/source/common/unicode/localpointer.h
```diff
@@ -548,46 +548,52 @@ class LocalArray : public LocalPointerBase<T> {
...
-            p1.swap(p2); \
-        } \
-        void adoptInstead(Type *p) { \
-            if (ptr != nullptr) { closeFunction(ptr); } \
-            ptr=p; \
-        } \
-        operator std::unique_ptr<Type, decltype(&closeFunction)> () && { \
-            return std::unique_ptr<Type, decltype(&closeFunction)>(LocalPointerBase<Type>::orphan(), closeFunction); \
-        } \
+    using LocalPointerClassName = LocalOpenPointer<Type, closeFunction>
```
markusicu, line 551: nice!

icu4c/source/common/unicode/localpointer.h
```diff
@@ -548,46 +548,52 @@ class LocalArray : public LocalPointerBase<T> {
...
+
+#ifndef U_IN_DOXYGEN
+template <typename Type, auto closeFunction>
+class LocalOpenPointer : public LocalPointerBase<Type> {
```
markusicu, line 555: problem: This is now a newly, publicly defined class. As such, it should be `@internal` -- or `@draft ICU 75` with a proposal.

icu4c/source/tools/pkgdata/pkgdata.cpp
```diff
@@ -60,11 +60,9 @@ U_CDECL_BEGIN
 U_CDECL_END
 
 #if U_HAVE_POPEN
-
-using icu::LocalPointerBase;
-
+U_NAMESPACE_BEGIN
```
markusicu, line 63: question: Should the macro qualify the right-side type with `icu::` so that users need not invoke the macro in the icu namespace?

markusicu (commented):
sorry for being late to the party!

icu4c/source/common/unicode/localpointer.h
markusicu: I didn't appreciate the doxygen guard, but it's still confusing to someone reading the header file.
It would be good to add an `@internal` comment.

icu4c/source/tools/pkgdata/pkgdata.cpp
markusicu: Users could use the macro for their own types, and then it would be weird to have to define those in the icu namespace. This was more interesting before C++11 added unique_ptr, but is illustrated by this tools code.

icu4c/source/common/unicode/localpointer.h
markusicu: Maybe more interestingly, if you are in an IDE and look at the class, it would help to have a doc string that says it's internal.

roubert: Does such a thing as an IDE that expands the macro and then resolves the generated `using`-statement and then displays the docstring for the underlying class template really exist? (If it does, then sure, then it certainly would be meaningful to add an `@internal` annotation here.)

icu4c/source/tools/pkgdata/pkgdata.cpp
roubert: Well, I don't think that `U_DEFINE_LOCAL_OPEN_POINTER()` is something that should be called from non-ICU code, but regardless of what I think that's an issue beyond this PR and ticket.

icu4c/source/common/unicode/localpointer.h
roubert: To ensure that anyone poking around in the header file will understand that this class template is internal, I propose wrapping it in a nested namespace named `internal` as the most effective solution, that should be impossible to miss in any context: PR #2815

markusicu: ok that looks acceptable tnx

roubert: > (If it does, then sure, then it certainly would be meaningful to add an `@internal` annotation here.)

Now when Andy has shown that it really does: PR #2816

PR #2811, opened 2024-02-06
ICU-22583 BRS 75rc CLDR 45-alpha0 to ICU main

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22583

ALLOW_MANY_COMMITS=true

First integration of CLDR 45 - in this case at alpha0 - to ICU main for ICU 75rc. This is in 4 commits mainly for ease of review (and these should be kept separate for merging):
1. Binary data, and binary-like source data (does not need review)
2. Data generated or copied from CLDR (can be spot checked)
3. Source and test code changes
4. A follow-up fix to pick up the new unitPrefixes data from CLDR

The main CLDR changes:
- Collation type "reformed" was deprecated and translations removed
- Metazones Macquarie and Mexico_Northwest were removed (along with translations); several IANA timezones were removed (along with translations). Metazone Greenland was added, with translation only in English.
- Latvian collation was updated.
- For CW and SX, new currency XCG becomes valid 2024-03-31, with the old currency ANG renaming valid until 2024-06-30. Translations for XCG are added in the most relevant locales.
- Display name updates in English for `ace`, in French for `njo`
- Double spaces are fixed in various patterns (CLDR 44.1 fix)
- Update to RBNF data for id
- CLDR also adds a mechanism to specify the timezone for currency transitions, and applies this to transitions after 2022-01-01 (though not yet for the ANG/XCG transition). This is handled just in the cldr-icu converter, no code updates needed in ICU.
- CLDR adds conversion data for (wind) speed unit "Beaufort", using new table-based mapping mechanism (can use for this and other scales). Full support is needed per ICU-22655, here we just do the minimum to get ICU to build and tests to pass.
- CLDR adds new unitPrefixes data. This was not correctly picked up in the first 3 commits, so commit 4 makes the necessary corrections and picks up the data.


markusicu (approved):
(partial rs)lgtm tnx

PR #2813, opened 2024-02-06
ICU-22520 Use a ByteSink append buffer instead of a local CharString

roubert:
These functions that eventually write their output to a `ByteSink` need a small temporary buffer for processing the subtag they're about to write and currently use a local `CharString` object to provide this buffer, which then gets written to the `ByteSink` and discarded.

This intermediate step is unnecessary as a `ByteSink` can provide an append buffer which can be used instead, eliminating the need to allocate a local temporary buffer and to copy the data around.

This approach also makes it natural to split the processing into two steps, first calculating the length of the subtag, then processing it, which makes it possible to return early when no output is requested.

##### Checklist

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -1172,7 +1172,7 @@ _getLanguage(const char* localeID,
              const char** pEnd,
              UErrorCode& status) {
     U_ASSERT(pEnd != nullptr);
-    CharString result;
+    *pEnd = localeID;
 
     if (uprv_stricmp(localeID, "root") == 0) {
         localeID += 4;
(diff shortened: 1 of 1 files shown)


icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
-    } else {
-        result.clear();
+    *pEnd = localeID + len;
+    if (sink == nullptr) { return; }
+
+    char scratch[ULOC_COUNTRY_CAPACITY];
+    int32_t capacity = 0;
+    char* buffer = sink->GetAppendBuffer(
+            ULOC_COUNTRY_CAPACITY,
+            ULOC_COUNTRY_CAPACITY,
```
markusicu, line 1286: rather than capacity=ULOC_COUNTRY_CAPACITY, we should use min(len, 4) to maximize use of tight ByteSink buffers; with a comment that we need 4 for NUL-terminating a 3-letter code for lookup of a 2-letter code

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
-        *pEnd = localeID + idLen;
-        if (idLen >= 1) {
-            result.append((char)uprv_toupper(*(localeID++)), status);
-        }
-        for (i = 1; i < idLen; i++) {
-            result.append((char)uprv_tolower(*(localeID++)), status);
-        }
+    int32_t len = 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len]) &&
+            uprv_isASCIILetter(localeID[len])) {
```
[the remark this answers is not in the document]
roubert: Done.

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
+
+    constexpr int32_t MINLEN = 2;
+    constexpr int32_t MAXLEN = ULOC_COUNTRY_CAPACITY - 1;  // Minus NUL.
 
-    /* copy the country as far as possible and count its length */
-    while(!_isTerminator(localeID[idLen]) && !_isIDSeparator(localeID[idLen])) {
-        result.append((char)uprv_toupper(localeID[idLen]), status);
-        idLen++;
+    int32_t len = 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len])) {
```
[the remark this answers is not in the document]
roubert: Done.

icu4c/source/common/uloc.cpp
roubert: Done.

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,132 @@ _getLanguage(const char* localeID,
...
 
+    constexpr int32_t MAXLEN = ULOC_LANG_CAPACITY - 1;  // Minus NUL.
+
     /* if it starts with i- or x- then copy that prefix */
-    if(_isIDPrefix(localeID)) {
-        result.append((char)uprv_tolower(*localeID), status);
-        result.append('-', status);
-        localeID+=2;
+    int32_t len = _isIDPrefix(localeID) ? 2 : 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len])) {
```
markusicu, line 1191:
```suggestion
    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len]) && len <= MAXLEN) {
```

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,132 @@ _getLanguage(const char* localeID,
...
-    /* copy the language as far as possible and count its length */
-    while(!_isTerminator(*localeID) && !_isIDSeparator(*localeID)) {
-        result.append((char)uprv_tolower(*localeID), status);
-        localeID++;
+    char scratch[MAXLEN];
+    int32_t capacity = 0;
+    int32_t min_capacity = uprv_min(len, 4);  // Minimum 3 letters plus NUL.
+    char* buffer = sink->GetAppendBuffer(
+            min_capacity, MAXLEN, scratch, UPRV_LENGTHOF(scratch), &capacity);
+    U_ASSERT(capacity >= min_capacity);
```
markusicu, line 1207:
```suggestion
    int32_t minCapacity = uprv_min(len, 4);  // Minimum 3 letters plus NUL.
    char* buffer = sink->GetAppendBuffer(
            minCapacity, minCapacity, scratch, UPRV_LENGTHOF(scratch), &capacity);
    U_ASSERT(capacity >= minCapacity);
```

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,132 @@ _getLanguage(const char* localeID,
...
+    char scratch[ULOC_COUNTRY_CAPACITY];
+    int32_t capacity = 0;
+    int32_t min_capacity = uprv_min(len, 4);  // Minimum 3 letters plus NUL.
+    char* buffer = sink->GetAppendBuffer(
+            min_capacity,
+            ULOC_COUNTRY_CAPACITY,
+            scratch,
+            UPRV_LENGTHOF(scratch),
+            &capacity);
+    U_ASSERT(capacity >= min_capacity);
```
markusicu, line 1294:
```suggestion
    int32_t minCapacity = uprv_min(len, 4);  // Minimum 3 letters plus NUL.
    char* buffer = sink->GetAppendBuffer(
            minCapacity,
            minCapacity,
            scratch,
            UPRV_LENGTHOF(scratch),
            &capacity);
    U_ASSERT(capacity >= minCapacity);
```

icu4c/source/common/uloc.cpp
roubert: Why would this be better?

icu4c/source/common/uloc.cpp
roubert: Why would this be better?

icu4c/source/common/uloc.cpp
markusicu: 1. lowerCamelCase as usual in ICU
2. don't need to ask for more space than we plan to write when the sink does have to reallocate

roubert: 1. The variable named `min_capacity` is being passed to the parameter named `min_capacity` in `GetAppendBuffer()`, there's a nice consistency in that, that can't be important to change.
2. I don't think I understand the changes you propose here at all. Let's take a concrete example:

The first entry in `LANGUAGES_3` is `"aar"` which should be shortened to `"aa"`. With the change you requested before, the value of `min_capacity` will be `uprv_min(len, 4)` which will be 3 which is one less than needed in order to store `"aar"` plus a NUL terminator in the buffer. I had previously assumed that this would work because the value of `desired_capacity_hint` was larger and the scratch buffer was large enough to accomodate that larger value. But when I now read the implementation of `CheckedArrayByteSink::GetAppendBuffer()` I see that it ignores the value of `desired_capacity_hint` and will return a buffer of size 3 if that's all that's available in the underlying array, which means that the NUL terminator written later will then write out of bounds. That's not good.

Did you maybe mean that the buffer here should be allocated with `min_capacity` set to `uprv_max(len, 4)` so that it allocates at least enough space to do the three-letter lookup but not any more than that unless needed?

That would make sense to me, but in that case the corresponding changes in `_getRegion()` don't make any sense at all, the hardcoded size in the first iteration of this PR is the smallest buffer that could be used.

So should I change the code here to use `uprv_max(len, 4)` and revert the changes in `_getRegion()`, or am I missing something?

markusicu: 1. The StringPiece and ByteSink classes came from Google with its different conventions. In code _using_ them we normally use ICU conventions.
2. Snafu -- need max, not min. Sorry!
3. If we don't need a max because the buffer has a fixed length, then that's fine. Note that I didn't comment on using LENGTH for scripts; and I didn't look at the numeric values of the other constants.

roubert: OK, I've now fixed up the code in these functions to always allocate the minimal buffer size needed, but not more than that for language tags shorter than the longest possible language tag.

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
-        int32_t i;
-        *pEnd = localeID + idLen;
-        if (idLen >= 1) {
-            result.append((char)uprv_toupper(*(localeID++)), status);
-        }
-        for (i = 1; i < idLen; i++) {
-            result.append((char)uprv_tolower(*(localeID++)), status);
-        }
+    int32_t len = 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len]) &&
```
markusicu, line 1240: optional: we can probably simply check for letters, since terminators and separators are not letters:
```
while (uprv_isASCIILetter(localeID[len])) {
```

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
+    constexpr int32_t MAXLEN = ULOC_LANG_CAPACITY - 1;  // Minus NUL.
+
     /* if it starts with i- or x- then copy that prefix */
-    if(_isIDPrefix(localeID)) {
-        result.append((char)uprv_tolower(*localeID), status);
-        result.append('-', status);
-        localeID+=2;
+    int32_t len = _isIDPrefix(localeID) ? 2 : 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len])) {
+        if (len == MAXLEN) {
```
markusicu, line 1192: optional / not a new problem in this PR: we apparently allow a length of 1, and allow characters other than letters...

icu4c/source/common/uloc.cpp
```diff
@@ -1184,104 +1184,128 @@ _getLanguage(const char* localeID,
...
+    constexpr int32_t MINLEN = 2;
+    constexpr int32_t MAXLEN = ULOC_COUNTRY_CAPACITY - 1;  // Minus NUL.
 
-    /* copy the country as far as possible and count its length */
-    while(!_isTerminator(localeID[idLen]) && !_isIDSeparator(localeID[idLen])) {
-        result.append((char)uprv_toupper(localeID[idLen]), status);
-        idLen++;
+    int32_t len = 0;
+    while (!_isTerminator(localeID[len]) && !_isIDSeparator(localeID[len])) {
+        if (len == MAXLEN) { return; }
```
markusicu, line 1275: optional / also not a new problem: we seem to allow things other than letters and digits in region subtags

markusicu (approved):

icu4c/source/common/uloc.cpp
markusicu: thank you for doing what i meant not what i said!! :-)

roubert:
Let's leave the _not a new problem in this PR_ issues the way they are, one step at a time is the only possible way forward here.

PR #2817, opened 2024-02-07
ICU-22520 Refactor function macros into inline functions

roubert:
This is to facilitate further refactoring of the locale code.

##### Checklist

### DIFF
--- a/icu4c/source/common/loclikely.cpp
+++ b/icu4c/source/common/loclikely.cpp
@@ -199,24 +199,27 @@ error:
     goto exit;
 }
 
-#define CHECK_TRAILING_VARIANT_SIZE(trailing, trailingLength) UPRV_BLOCK_MACRO_BEGIN { \
-    int32_t count = 0; \
-    int32_t i; \
-    for (i = 0; i < trailingLength; i++) { \
-        if (trailing[i] == '-' || trailing[i] == '_') { \
-            count = 0; \
-            if (count > 8) { \
-                goto error; \
-            } \
-        } else if (trailing[i] == '@') { \
-            break; \
-        } else if (count > 8) { \
-            goto error; \
-        } else { \
-            count++; \
-        } \
-    } \
-} UPRV_BLOCK_MACRO_END
+namespace {
+inline bool CHECK_TRAILING_VARIANT_SIZE(const char* trailing, int32_t trailingLength) {
+    int32_t count = 0;
+    int32_t i;
+    for (i = 0; i < trailingLength; i++) {
+        if (trailing[i] == '-' || trailing[i] == '_') {
+            count = 0;
+            if (count > 8) {
+                return false;
+            }
+        } else if (trailing[i] == '@') {
+            break;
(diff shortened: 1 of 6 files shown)


icu4c/source/common/loclikely.cpp
```diff
@@ -199,24 +199,27 @@ parseTagString(
...
+inline bool CHECK_TRAILING_VARIANT_SIZE(const char* trailing, int32_t trailingLength) {
+    int32_t count = 0;
+    int32_t i;
+    for (i = 0; i < trailingLength; i++) {
+        if (trailing[i] == '-' || trailing[i] == '_') {
+            count = 0;
+            if (count > 8) {
+                return false;
+            }
+        } else if (trailing[i] == '@') {
```
markusicu, line 212: just FYI: I see that we already have fairly low-level code that won't work on EBCDIC machines...

icu4c/source/common/loclikely.cpp
```diff
@@ -266,7 +269,8 @@ _uloc_addLikelySubtags(const char* localeID,
     trailing = &localeID[trailingIndex];
     trailingLength = (int32_t)uprv_strlen(trailing);
 
-    CHECK_TRAILING_VARIANT_SIZE(trailing, trailingLength);
+    if (!CHECK_TRAILING_VARIANT_SIZE(trailing, trailingLength)) goto error;
```
markusicu, line 272:
```suggestion
    if (!CHECK_TRAILING_VARIANT_SIZE(trailing, trailingLength)) { goto error; }
```
... and can we do better now than `goto error`?

icu4c/source/common/loclikely.cpp
```diff
@@ -362,7 +366,7 @@ _uloc_minimizeSubtags(const char* localeID,
...
```
markusicu, line 369:
```suggestion
    if (!CHECK_TRAILING_VARIANT_SIZE(trailing, trailingLength)) { goto error; }
```

icu4c/source/common/loclikely.cpp
roubert: I'll be coming for `goto` next, that's just easier to do if getting this out of macros first.

markusicu: ok, but braces please?

roubert: Is spending an hour on running the CI after applying that change really a meaningful use of time for something that's about to be deleted anyway?

roubert: Applying those changes will trigger a CI run, which takes about an hour. It then leaves the PR with commits that'll need to be squashed before the PR can be merged. Squashing them will trigger a CI run, which takes about an hour again. Is that really a meaningful use of time for something that's about to be replaced anyway?

markusicu: amending the commit would trigger only one round, not two...

markusicu (approved):

PR #2821, opened 2024-02-08
ICU-22661 Limit the size of variants in Locale

FrankYFTang:
Limit the size of acceptable variants in Locale to avoid long running time of 
resource fallback look up when trying all different combinations of many variants.

I first try to limit to the max size of 10 variants but some test break, so I change it to 20 now.

Also ensure the Locale::getVariant will not return invalid address while the locale is bogus.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22661

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -1324,14 +1324,30 @@ _getVariant(const char* localeID,
             char prev,
             ByteSink* sink,
             const char** pEnd,
-            bool needSeparator) {
-    bool hasVariant = false;
+            bool needSeparator,
+            UErrorCode& status) {
+    if (U_FAILURE(status)) return;
     if (pEnd != nullptr) { *pEnd = localeID; }
 
+    // Reasonable upper limit for variants
+    // There are no strict limitation of the syntax of variant in the legacy
+    // locale format. If the locale is constructed from unicode_locale_id
+    // as defined in UTS35, then we know each unicode_variant_subtag
+    // could have max length of 8 ((alphanum{5,8} | digit alphanum{3})
+    // 179 would allow 20 unicode_variant_subtag with sep in the
+    // unicode_locale_id
+    // 8*20 + 1*(20-1) = 179
+    constexpr int32_t MAX_VARIANTS_LENGTH = 179;
+
     /* get one or more variant tags and separate them with '_' */
-    if(_isIDSeparator(prev)) {
+    int32_t index = 0;
+    if (_isIDSeparator(prev)) {
         /* get a variant string after a '-' or '_' */
-        while(!_isTerminator(*localeID)) {
+        for (index=0; !_isTerminator(localeID[index]); index++) {
+            if (index >= MAX_VARIANTS_LENGTH) { // same as length > MAX_VARIANTS_LENGTH
+                status = U_ILLEGAL_ARGUMENT_ERROR;
+                return;
+            }
             if (needSeparator) {
                 if (sink != nullptr) {
                     sink->Append("_", 1);
@@ -1339,26 +1355,28 @@ _getVariant(const char* localeID,
                 needSeparator = false;
             }
             if (sink != nullptr) {
-                char c = (char)uprv_toupper(*localeID);
+                char c = (char)uprv_toupper(localeID[index]);
                 if (c == '-') c = '_';
                 sink->Append(&c, 1);
             }
-            hasVariant = true;
-            localeID++;
         }
-        if (pEnd != nullptr) { *pEnd = localeID; }
+        if (pEnd != nullptr) { *pEnd = localeID+index; }
     }
 
     /* if there is no variant tag after a '-' or '_' then look for '@' */
-    if(!hasVariant) {
-        if(prev=='@') {
+    if (index == 0) {
+        if (prev=='@') {
             /* keep localeID */
         } else if((localeID=locale_getKeywordsStart(localeID))!=nullptr) {
             ++localeID; /* point after the '@' */
         } else {
             return;
         }
-        while(!_isTerminator(*localeID)) {
+        for(; !_isTerminator(localeID[index]); index++) {
+            if (index >= MAX_VARIANTS_LENGTH) { // same as length > MAX_VARIANTS_LENGTH
+                status = U_ILLEGAL_ARGUMENT_ERROR;
+                return;
+            }
             if (needSeparator) {
                 if (sink != nullptr) {
                     sink->Append("_", 1);
@@ -1366,13 +1384,12 @@ _getVariant(const char* localeID,
                 needSeparator = false;
             }
             if (sink != nullptr) {
-                char c = (char)uprv_toupper(*localeID);
+                char c = (char)uprv_toupper(localeID[index]);
                 if (c == '-' || c == ',') c = '_';
                 sink->Append(&c, 1);
             }
-            localeID++;
         }
-        if (pEnd != nullptr) { *pEnd = localeID; }
+        if (pEnd != nullptr) { *pEnd = localeID + index; }
     }
 }
 
@@ -1545,7 +1562,8 @@ ulocimp_getSubtags(
         }
         const char* begin = localeID + 1;
         const char* end = nullptr;
-        _getVariant(begin, *localeID, variant, &end, false);
+        _getVariant(begin, *localeID, variant, &end, false, status);
+        if (U_FAILURE(status)) { return; }
         U_ASSERT(end != nullptr);
         if (end != begin && pEnd != nullptr) { *pEnd = end; }
     }
@@ -1842,7 +1860,8 @@ _canonicalize(const char* localeID,
             }
 
             CharStringByteSink s(&tag);
-            _getVariant(tmpLocaleID+1, '@', &s, nullptr, !variant.isEmpty());
+            _getVariant(tmpLocaleID+1, '@', &s, nullptr, !variant.isEmpty(), err);
+            if (U_FAILURE(err)) { return; }
         }
 
         /* Look up the ID in the canonicalization map */
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/LocaleIDParser.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/LocaleIDParser.java
@@ -365,6 +365,15 @@ public final class LocaleIDParser {
         }
     }
 
+    // There are no strict limitation of the syntax of variant in the legacy
+    // locale format. If the locale is constructed from unicode_locale_id
+    // as defined in UTS35, then we know each unicode_variant_subtag
+    // could have max length of 8 ((alphanum{5,8} | digit alphanum{3})
+    // 179 would allow 20 unicode_variant_subtag with sep in the
+    // unicode_locale_id
+    // 8*20 + 1*(20-1) = 179
+    private static final int MAX_VARIANTS_LENGTH = 179;
+
     /**
      * Advance index past variant, and accumulate normalized variant in buffer.  This ignores
      * the codepage information from POSIX ids.  Index must be immediately after the country
@@ -432,10 +441,12 @@ public final class LocaleIDParser {
                     c = UNDERSCORE;
                 }
                 append(c);
+                if (buffer.length() - oldBlen > MAX_VARIANTS_LENGTH) {
+                    throw new IllegalArgumentException("variants is too long");
+                }
             }
         }
         --index; // unget
-
         return oldBlen;
     }
 
--- a/icu4c/source/common/unicode/locid.h
+++ b/icu4c/source/common/unicode/locid.h
@@ -1211,7 +1211,7 @@ Locale::getScript() const
 inline const char *
 Locale::getVariant() const
 {
-    return &baseName[variantBegin];
+    return fIsBogus ? "" : &baseName[variantBegin];
 }
 
 inline const char *
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -511,6 +511,11 @@ enum AllowedHourFormat{
 void
 DateTimePatternGenerator::initData(const Locale& locale, UErrorCode &status, UBool skipStdPatterns) {
     //const char *baseLangName = locale.getBaseName(); // unused
+    if (U_FAILURE(status)) { return; }
+    if (locale.isBogus()) {
+        status = U_ILLEGAL_ARGUMENT_ERROR;
+        return;
+    }
 
     skipMatcher = nullptr;
     fAvailableFormatKeyHash=nullptr;
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -224,6 +224,7 @@ void addLocaleTest(TestNode** root)
     TESTCASE(TestBasicGetters);
     TESTCASE(TestNullDefault);
     TESTCASE(TestPrefixes);
+    TESTCASE(TestVariantLengthLimit);
     TESTCASE(TestSimpleResourceInfo);
     TESTCASE(TestDisplayNames);
     TESTCASE(TestGetDisplayScriptPreFlighting21160);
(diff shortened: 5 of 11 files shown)


macchiati:
Another way to attack this, that would be much faster, is not try resource fallback for every one of the variants in a locale.

Instead, do a lazy build of all the variant combinations that actually occur in the resource bundles, eg

0: {posix}
1: {posix valencia}
2: {foo}

When doing resource fallback, cycle through these (in the example, 0..2) and only try the ones that are contained within the set of variants. That will typically be none for oddball variants, and (very) occasionally one or two of the combinations.

roubert:
> Another way to attack this, that would be much faster, is not try resource fallback for every one of the variants in a locale.

This seems like a very good idea to me.

markusicu:
> Another way to attack this, that would be much faster, is not try resource fallback for every one of the variants in a locale.
> 
> Instead, do a lazy build of all the variant combinations that actually occur in the resource bundles, eg

That sounds good, but runs into old complications with enumerating the set of available locales / bundles.

Enumerating is relatively easy when they are in a single file or place, but we let users provide data in one or more .dat files and/or bunches of single files in the file system (with different OS APIs) and/or linked in with one or another type of TOC... and tangled up with service registration caches. In C++ and Java.

This would be really good to figure out, but shouldn't block other progress.

markusicu:
> I first try to limit to the max size of 10 variants but some test break, so I change it to 20 now.

If we think that 10 is already plenty, then it should be ok to reduce the number of variants in unit tests.

roubert:
> This would be really good to figure out, but shouldn't block other progress.

OK, if putting limits on the length and number of variants is the best path forward, let's figure out how to best do that.

First of all, I'd like to have a unit test for this that fails if language tags with too long or too many variants are accepted, so that we'll see immediately if some future change accidentally breaks the checking for length and number.

Then I find `Locale::initBaseName()` to be a rather odd place for this logic, far too late in the processing. It should at least be moved up the chain to `Locale::init()` where `variantBegin` is set. But I wonder if it wouldn't be even better to move the logic all the way up to `_getVariant()` in `uloc.cpp` so that the same limits are enforced for all of ICU4C and not just for code that uses `class Locale`.

FrankYFTang:
> First of all, I'd like to have a unit test for this that fails if language tags with too long or too many variants are accepted, so that we'll see immediately if some future change accidentally breaks the checking for length and number.


see testLongVariants

FrankYFTang:
PTAL

icu4c/source/common/uloc.cpp
```diff
@@ -1381,6 +1385,14 @@ _getVariant(const char* localeID,
         }
         if (pEnd != nullptr) { *pEnd = localeID; }
     }
+    int32_t variantsLength = localeID - start;
+    // Reasonable upper limit for variants
+    static constexpr int32_t MAX_NUMBER_OF_VARIANTS = 20;
+    static constexpr int32_t MAX_VARIANT_LENGTH = 8;
+    static constexpr int32_t MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS;
```
roubert, line 1392:
```suggestion
    constexpr int32_t MAX_NUMBER_OF_VARIANTS = 20;
    constexpr int32_t MAX_VARIANT_LENGTH = 8;
    constexpr int32_t MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS;
```

icu4c/source/common/uloc.cpp
```diff
@@ -1546,7 +1558,7 @@ ulocimp_getSubtags(
         }
         const char* begin = localeID + 1;
         const char* end = nullptr;
-        _getVariant(begin, *localeID, variant, &end, false);
+        _getVariant(begin, *localeID, variant, &end, false, status);
```
roubert, line 1561:
```suggestion
        _getVariant(begin, *localeID, variant, &end, false, status);
        if (U_FAILURE(status)) { return; }
```

icu4c/source/common/uloc.cpp
```diff
@@ -1844,7 +1856,7 @@ _canonicalize(const char* localeID,
             }
 
             CharStringByteSink s(&tag);
-            _getVariant(tmpLocaleID+1, '@', &s, nullptr, !variant.isEmpty());
+            _getVariant(tmpLocaleID+1, '@', &s, nullptr, !variant.isEmpty(), err);
```
roubert, line 1859: Wouldn't you want to return here in case there was an error? (If not, why not?)

icu4c/source/common/uloc.cpp
```diff
@@ -1381,6 +1385,14 @@ _getVariant(const char* localeID,
         }
         if (pEnd != nullptr) { *pEnd = localeID; }
     }
+    int32_t variantsLength = localeID - start;
```
roubert, line 1388: If there is a very large amount of variants, that now are about to be rejected, the code will have spent a lot of effort repeatedly calling `sink->Append()` to copy all those characters before reaching this point. As the purpose of this PR is to avoid long running time, you might want to consider adding a preprocessing step at the beginning of this function, a simpler loop that just counts the length and returns early with an error if the variants are too long, before starting to copy anything (see `_getLanguage()` for an analogous such loop in one of the other of these functions).

roubert:
> > First of all, I'd like to have a unit test for this that fails if language tags with too long or too many variants are accepted, so that we'll see immediately if some future change accidentally breaks the checking for length and number.
> 
> see testLongVariants

The problem with that test is that it doesn't fail if a language tags with overlong variants gets accepted, it just runs for a long time and then passes anyway. That's not helpful to verify that the logic for rejecting overlong variants works. I'd like to have a test that fails if the detection of overlong variants fails.

Such a test could be very simple:

1. Verify that a language tag with the maximum length of variants is accepted.
2. Verify that a language tag with variants one character longer than the maximum is rejected.

FrankYFTang:
> > > First of all, I'd like to have a unit test for this that fails if language tags with too long or too many variants are accepted, so that we'll see immediately if some future change accidentally breaks the checking for length and number.
> > 
> > 
> > see testLongVariants
> 
> The problem with that test is that it doesn't fail if a language tags with overlong variants gets accepted, it just runs for a long time and then passes anyway. That's not helpful to verify that the logic for rejecting overlong variants works. I'd like to have a test that fails if the detection of overlong variants fails.
> 
> Such a test could be very simple:
> 
> 1. Verify that a language tag with the maximum length of variants is accepted.
> 2. Verify that a language tag with variants one character longer than the maximum is rejected.

changed . Added those tests. Also add Java test which will show stack overflow if not including the fix. PTAL

roubert:
While I think the updated test is much better as it now will fail if the rejection of overlong variants fails, I still don't understand why you don't want to have a simple test that just tests this directly. If the rejection of overlong variants fails, your updated test runs for an entire minute on my workstation before it eventually fails. Isn't our CI already slow enough as it is?

Please take a look at the unit tests that I just added, which test the C API and C++ API for locale ID parsing directly and verify that parsing of variants of the maximum allowed length succeeds and that parsing of variants one character longer than that fails. The code is perfectly trivial and runs in a fraction of a second.

What advantage is it that you see in instead going through creating `DateTimePatternGenerator` objects and having loops that build very large strings? I don't understand it.

FrankYFTang:
> What advantage is it that you see in instead going through creating `DateTimePatternGenerator` objects and having loops that build very large strings? I don't understand it.

that is what the test case in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65761 trigger the problem. I would like to see this fix indeed fix that particular case

FrankYFTang:
> Please take a look at the unit tests that I just added, which test the C API and C++ API for locale ID parsing directly and verify that parsing of variants of the maximum allowed length succeeds and that parsing of variants one character longer than that fails.

I think what you added are good, and we should include them. But that test does not verify the hang in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65761 no longer hang.

roubert:
> But that test does not verify the hang in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65761 no longer hang.

I still don't get it. The reason that `DateTimePatternGenerator::createInstance()` runs for a very long time (it doesn't actually hang, if not terminated by a timeout it'd eventually finish) is not because of anything in `DateTimePatternGenerator` itself but in the underlying resource loading, which is used also in many other places around ICU4C.

If you want to test that `ures_getFunctionalEquivalent()`, the resource loading function that currently runs for a very long time when given a locale ID with many variants, no longer will try loading resources for too many variants, then I think the natural thing would be a unit test for that particular function, not for any randomly selected user facing function that eventually calls it somewhere within its implementation.

Please take a look at the unit test that I just added, which does just that. The code is perfectly trivial and runs in a fraction of a second.

What is it that you think that such a unit test fails to cover?

As for the loops that you propose to use to build very large strings, what is the advantage of having those? The `testLongVariants()` test case currently in this PR will _triple_ the total running time of the entire `intltest` test suite if the code that's supposed to abort when there are too many variants fails to do that, and that really doesn't seem like a reasonable increase in running time to me.

FrankYFTang:
> I still don't get it. The reason that `DateTimePatternGenerator::createInstance()` runs for a very long time (it doesn't actually hang, if not terminated by a timeout it'd eventually finish) is not because of anything in `DateTimePatternGenerator` itself but in the underlying resource loading, which is used also in many other places around ICU4C.

The only way we first figure out the reason is that run for a very long time is AFTER we debug the test I added, right? So... how could we ever know what "The reason that ... is becaue" ? I want to added test to ensure in the future, if anything else change, it will not introduce any other bug to cause such regression for any other reason. How can we ensure such particular breakge will stay fixed withoug any regression in the next 20 years without adding such a test?  It is true that the current fix does fix the problem, but there are no other test to ensure in the future no body will introudce new bugs to cause the same bug unless I add such a test, right?

FrankYFTang:
> As for the loops that you propose to use to build very large strings, what is the advantage of having those?

The advantage is it mimic the EXACT failure condition what the fuzzer report in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65761

FrankYFTang:
> I think it'd be better to keep unit tests as unit tests and fuzzer tests as fuzzer tests.

ok, I will then remove my origional unit test. that make sense.

FrankYFTang:
@roubert  PTAL. I removed my origional test that you have issue with and port the test you proposed to Java

icu4j/main/core/src/main/java/com/ibm/icu/impl/LocaleIDParser.java
```diff
@@ -435,7 +435,9 @@ private int parseVariant() {
             }
         }
         --index; // unget
-
+        if (buffer.length() - oldBlen > 180) {
```
roubert, line 438: Instead of iterating through the entire (possibly very long) string, it'd be more conducive towards the goal of this PR of avoiding long running time to instead exit immediately as soon as it can be known that the list of variants is too long, just like you do in the C++ implementation.

icu4c/source/common/uloc.cpp
```diff
@@ -1324,13 +1324,31 @@ _getVariant(const char* localeID,
...
+    constexpr int32_t MAX_VARIANT_LENGTH = 8;
+    constexpr int32_t MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS;
+
     /* get one or more variant tags and separate them with '_' */
     if(_isIDSeparator(prev)) {
         /* get a variant string after a '-' or '_' */
+        while(!_isTerminator(*localeID)) {
+            localeID++;
+        }
+        int32_t variantsLength = localeID - start;
```
roubert, line 1345: The code would become simpler if you instead of pointer aritmetic just used a for-loop that increased an `int32_t` index variable by one upon each iteration, then you wouldn't have to store the original start of the string and calculate the length anew in each iteration.

roubert (changes requested):

roubert:
> and port the test you proposed to Java

You'd want to add a Java unit test also for `ULocale`, wouldn't you?

icu4c/source/common/uloc.cpp
```diff
@@ -1324,13 +1324,31 @@ _getVariant(const char* localeID,
...
+            UErrorCode& status) {
+    if (U_FAILURE(status)) return;
+    const char* start = localeID;
     bool hasVariant = false;
     if (pEnd != nullptr) { *pEnd = localeID; }
 
+    // Reasonable upper limit for variants
+    constexpr int32_t MAX_NUMBER_OF_VARIANTS = 20;
+    constexpr int32_t MAX_VARIANT_LENGTH = 8;
+    constexpr int32_t MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS;
```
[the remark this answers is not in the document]
FrankYFTang: This is just an estimation since we do not know the exact legnth of each variant. If each variant value is only length of three, we could fit 45 variants here so I do not think we need it to be exact since neither  MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS 
nor
MAX_VARIANTS_LENGTH = (MAX_VARIANT_LENGTH + 1) * MAX_NUMBER_OF_VARIANTS - 1
could guarantee there are at most 20 variants in this.
and due to the acceptable grammar of variant, if all the variant are lenght of 8 there will be no syntax valid variants string be exact 180 long anyway.

FrankYFTang:
PTAL

FrankYFTang:
@roubert  PTAL

icu4c/source/common/uloc.cpp
```diff
@@ -1324,55 +1324,75 @@ _getVariant(const char* localeID,
...
-            localeID++;
+            if (index > MAX_VARIANTS_LENGTH) {
+                status = U_ILLEGAL_ARGUMENT_ERROR;
+                return;
+            }
         }
-        if (pEnd != nullptr) { *pEnd = localeID; }
+        if (pEnd != nullptr) { *pEnd = localeID + index; }
+    }
+    if (index > MAX_VARIANTS_LENGTH) {
```
roubert, line 1394: It seems to me that this could never happen. Are there any circumstances under which this code could actually be reached? If not, it won't ever be used and can be deleted.

icu4c/source/common/uloc.cpp
```diff
@@ -1324,55 +1324,75 @@ _getVariant(const char* localeID,
...
                 needSeparator = false;
             }
             if (sink != nullptr) {
-                char c = (char)uprv_toupper(*localeID);
+                char c = (char)uprv_toupper(localeID[index]);
                 if (c == '-' || c == ',') c = '_';
                 sink->Append(&c, 1);
             }
-            localeID++;
+            if (index > MAX_VARIANTS_LENGTH) {
```
roubert, line 1387: This if-statement can be the first statement inside of the loop (it's a bit odd to first do some unnecessary work and then afterwards exit with an error).

icu4c/source/common/uloc.cpp
```diff
@@ -1324,55 +1324,75 @@ _getVariant(const char* localeID,
...
             }
             if (sink != nullptr) {
-                char c = (char)uprv_toupper(*localeID);
+                char c = (char)uprv_toupper(localeID[index]);
                 if (c == '-') c = '_';
                 sink->Append(&c, 1);
             }
-            hasVariant = true;
-            localeID++;
+            if (index > MAX_VARIANTS_LENGTH) {
```
roubert, line 1358: This if-statement can be the first statement inside of the loop (it's a bit odd to first do some unnecessary work and then afterwards exit with an error).

icu4j/main/core/src/main/java/com/ibm/icu/impl/LocaleIDParser.java
```diff
@@ -432,10 +441,15 @@ private int parseVariant() {
...
                 }
                 append(c);
+                if (buffer.length() - oldBlen > MAX_VARIANTS_LENGTH) {
+                    throw new IllegalArgumentException("variants is too long");
+                }
             }
         }
         --index; // unget
-
+        if (buffer.length() - oldBlen > MAX_VARIANTS_LENGTH) {
```
roubert, line 450: It seems to me that this could never happen. Are there any circumstances under which this code could actually be reached? If not, it won't ever be used and can be deleted.

icu4c/source/common/uloc.cpp
FrankYFTang: yes, you are right. It is hard for me to see this is will not happen with a small screen but after carefully looking up/down that is surely the case and I now remove this.

icu4c/source/common/uloc.cpp
FrankYFTang: moved

icu4c/source/common/uloc.cpp
FrankYFTang: moved

icu4c/source/common/uloc.cpp
FrankYFTang: This happen when  _isTerminator(localeID[MAX_VARIANTS_LENGTH+1]) is true.
Inside the loop, in the previous iteration, index == MAX_VARIANTS_LENGTH so it does not error return by the "if (index > MAX_VARIANTS_LENGTH)" check, then the loop go to the next iteration by index++ which make index is MAX_VARIANTS_LENGTH +1 first, then it call _isTerminator(localeID[index]) , but that return true cause the loop terminate. at this point index is MAX_VARIANTS_LENGTH +1 which is > MAX_VARIANTS_LENGTH.

icu4j/main/core/src/main/java/com/ibm/icu/impl/LocaleIDParser.java
FrankYFTang: yes this is correct. I removed it.

roubert (approved):
LGTM

PR #2823, opened 2024-02-12
ICU-22520 Use C++ function signatures for internal C++ functions

roubert:
Some of this code was originally written as C code and some of this code was originally written as C++ code but made to resemble the then already existing code that had once been C code. Changing it all to normal C++ now will make it easier and safer to work with going forward.

- Use unnamed namespace instead of static.
- Use reference instead of non-nullable pointer.
- Use bool instead of UBool.
- Use constexpr for static data.
- Use U_EXPORT instead of U_CAPI or U_CFUNC.
- Use the default calling convention instead of U_EXPORT2.

##### Checklist

### DIFF
--- a/icu4c/source/common/localebuilder.cpp
+++ b/icu4c/source/common/localebuilder.cpp
@@ -10,16 +10,14 @@
 #include "unicode/localebuilder.h"
 #include "unicode/locid.h"
 
-U_NAMESPACE_BEGIN
-
 namespace {
+
 inline bool UPRV_ISDIGIT(char c) { return c >= '0' && c <= '9'; }
 inline bool UPRV_ISALPHANUM(char c) { return uprv_isASCIILetter(c) || UPRV_ISDIGIT(c); }
-}  // namespace
 
 constexpr const char* kAttributeKey = "attribute";
 
-static bool _isExtensionSubtags(char key, const char* s, int32_t len) {
+bool _isExtensionSubtags(char key, const char* s, int32_t len) {
     switch (uprv_tolower(key)) {
         case 'u':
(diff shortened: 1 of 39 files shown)


markusicu (commented):
Changes lgtm but I wonder what this looks like when you regenerate urename.h -- without `extern "C"` will we get mangled C++ entry point names there? That then would make urename.h depend on how the compiler+version does the name mangling.

roubert:
> Changes lgtm but I wonder what this looks like when you regenerate urename.h -- without `extern "C"` will we get mangled C++ entry point names there?

I've now verified that running genren after the changes in this PR doesn't cause any changes to the generated `urename.h` file (note that some of these functions already had C++ linkage, even overloads, so it's to be expected that nothing should change).

markusicu:
> I've now verified that running genren after the changes in this PR doesn't cause any changes to the generated `urename.h` file (note that some of these functions already had C++ linkage, even overloads, so it's to be expected that nothing should change).

Thanks for checking! I guess the ones that get exported at all (to be used elsewhere in ICU) all are, and should be, in the "icu" namespace, and are versioned via that.

markusicu (approved):

roubert:
> I guess the ones that get exported at all (to be used elsewhere in ICU) all are, and should be, in the "icu" namespace, and are versioned via that.

There are also a lot of freestanding C++ functions outside of any namespace, like the `ulocimp_*()` functions, but they all end up correctly in `urename.h`. (As an idea for future improvements, one might consider moving all such functions into a namespace instead of keeping that old C style name prefix.)

PR #2828, opened 2024-02-15
ICU-22556 Prefer cc and c++ compilers

jwillikers:
When building icu4c, it defaults to clang instead of gcc when the default compiler, cc / c++, is a symlink to gcc / g++. This is not the expected behavior when building C and C++ code. It appears that this behavior was put in place originally for supporting C++11, which hopefully is no longer such a concern. This PR adjusts the configure.ac for icu4c to prefer the cc and c++ compilers first.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22556

ALLOW_MANY_COMMITS=true

### DIFF
--- a/docs/devsetup/cpp/index.md
+++ b/docs/devsetup/cpp/index.md
@@ -70,11 +70,11 @@ For example:
         --disable-release Linux --prefix=/home/*your_user_name*/icu/mine/inst >
         config.out 2>&1`
     *   build: make -j5 check > out.txt 2>&1
-    *   Be sure to test with gcc and g++ too! `CC=gcc CXX=g++
-        CXXFLAGS="-DU_USING_ICU_NAMESPACE=0"
+    *   Be sure to test with both gcc/g++ and clang/clang++!
+        `CXXFLAGS="-DU_USING_ICU_NAMESPACE=0"
         CPPFLAGS="-DU_NO_DEFAULT_INCLUDE_UTF_HEADERS=1"
         ../../src/icu4c/source/runConfigureICU --enable-debug --disable-release
-        Linux`
+        Linux/clang`
 *   `~/icu/mine/icu4c/**nm_utf8**`
     *   not-using-namespace and default-hardcoded-UTF-8
     *   setup: mkdir+cd to here, then something like
--- a/icu4c/source/configure.ac
+++ b/icu4c/source/configure.ac
@@ -147,10 +147,8 @@ AC_SUBST(ENABLE_RELEASE)
 : ${CXXFLAGS=""}
 
 # Checks for compilers
-AC_PROG_CC([clang gcc cc c99 c89 xlc_r xlc cl.exe icc])
-# Make sure that we try clang++ first, which provides C++17 support.
-# The g++ compiler is less likely to support C++17.
-AC_PROG_CXX([clang++ g++ c++ gpp xlC_r xlC aCC CC cxx cc++ cl.exe icc FCC KCC RCC])
+AC_PROG_CC([cc clang gcc c99 c89 xlc_r xlc cl.exe icc])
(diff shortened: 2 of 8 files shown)


icu4c/source/configure.ac
```diff
@@ -147,10 +147,10 @@ AC_SUBST(ENABLE_RELEASE)
 : ${CXXFLAGS=""}
 
 # Checks for compilers
-AC_PROG_CC([clang gcc cc c99 c89 xlc_r xlc cl.exe icc])
+AC_PROG_CC([cc clang gcc c99 c89 xlc_r xlc cl.exe icc])
 # Make sure that we try clang++ first, which provides C++17 support.
 # The g++ compiler is less likely to support C++17.
```
roubert, line 152: You'll need to also update this comment.

jwillikers: Hmmm, should I remove the comment entirely?

roubert: I think that's a good decision, that comment originates from a time when GCC didn't have sufficient C++11 support and it's no longer relevant.

roubert:
Oops, it seems as we both force pushed at almost exactly the same time, so that I overwrote your changes. Sorry about that. But please take a look at my changes to the workflow configuration, using `/clang` is only necessary where `CC` and `CXX` aren't explicitly set.

jwillikers:
Got it. Sorry about that force-push, I'll add further changes in separate commits to avoid that.

docs/devsetup/cpp/index.md
```diff
@@ -70,7 +70,7 @@ For example:
         --disable-release Linux --prefix=/home/*your_user_name*/icu/mine/inst >
         config.out 2>&1`
     *   build: make -j5 check > out.txt 2>&1
-    *   Be sure to test with gcc and g++ too! `CC=gcc CXX=g++
+    *   Be sure to test with both gcc/g++ and clang/clang++! `CC=clang CXX=clang++
```
roubert, line 73:
```suggestion
    *   Be sure to test with both gcc/g++ and clang/clang++! `CC=clang CXX=clang++`
```

roubert: Maybe suggest using `runConfigureICU Linux/clang` instead?

roubert (changes requested):

docs/devsetup/cpp/index.md
roubert: Better, but the quoting is still off, now there's no initial ` before the code block. Please fix that and then squash all the manual changes together into one single commit (after which then follows the commit that I made with the autogenerated changes).

jwillikers: Done.

roubert: Looks good, but please squash all manual changes, including the one's that I've made, into a single commit, the only thing that's advantageous to keep separate are the autogenerated changes.

jwillikers: Okay, I think you must be talking about the commit `ICU-22556 Update configure files from configure.ac using autoreconf.` that should be separate, right? If that's the case, the branch should be good now.

roubert (dismissed):
I think this looks good now, so I now approve this PR, so that it can be merged once the ICU Technical Committee has accepted the ticket.

jwillikers:
> I think this looks good now, so I now approve this PR, so that it can be merged once the ICU Technical Committee has accepted the ticket.

Thanks for the review!

markusicu:
@jwillikers could you please rebase this PR? It has a merge conflict now.

We did discuss and approve this change in the ICU-TC meeting today.

jwillikers:
> @jwillikers could you please rebase this PR? It has a merge conflict now.
> 
> We did discuss and approve this change in the ICU-TC meeting today.

@markusicu Done.

roubert (approved):
LGTM

PR #2834, opened 2024-02-20
ICU-22671 format_date uses c++ nullptr instead of 0 for udat_open/DateFormat::c…

devnexen:
…reate


echeran:
Looks reasonable. Would you mind re-adding the ICU PR checklist into the PR description and fill it out?

For clarity purposes, I think rewording the title from "... uses ..." to "... should use " would be better.

roubert:
You'll need to rebase this on top of current main (past commit 3c82e6857c0bc0bbe93c0cdeaf03be3844dcad96) for the presubmit tests to pass.

roubert (approved):
LGTM

PR #2845, opened 2024-02-27
ICU-22532 Increase Github Actions CI parallelism from 2 to 4

echeran:
Github has increased the number of cores in the machines available for Github Actions from 2 to 4:
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22532

### DIFF
--- a/.github/workflows/icu4c.yml
+++ b/.github/workflows/icu4c.yml
@@ -61,7 +61,7 @@ jobs:
           mkdir build;
           cd build;
           ../icu4c/source/runConfigureICU --enable-debug --disable-release Linux/gcc --prefix=$PREFIX --enable-tracing;
-          make -j -l2.5 check;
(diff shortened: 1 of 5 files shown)


markusicu (approved):

roubert:
The previous load average limit of 2.5 was chosen rather arbitrarily in commit e1e1c5feaf67652d9f124b23749a430d3278fe7b so any larger value is probably as good as any other larger value, but when done without any comment about why, it does seem a bit odd to me to increase the load average limit by 80% in response to the number of available cores being increased by 100% (so I feel compelled to write this comment for the benefit of anyone who in the future might try to figure out why the particular limit of 4.5 was chosen).

PR #2853, opened 2024-02-28
ICU-22679 Optimize calendar code for edge cases

FrankYFTang:
Optimize the calendar code to avoid very slow edge case found by fuzzer
1. Use local variable to remember result to avoid call the same function twice with the same value.
2. Remove function of private usage from header and make them file private and inline.
3. Merge utility function which only called once into the calling function.

This is to address https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67009

##### Checklist

### DIFF
--- a/icu4c/source/i18n/islamcal.cpp
+++ b/icu4c/source/i18n/islamcal.cpp
@@ -192,15 +192,6 @@ static const int UMALQURA_MONTHLENGTH[] = {
                             0x06AA,           0x0AD6,           0x055D,           0x029D
 };
 
-int32_t getUmalqura_MonthLength(int32_t y, int32_t m) {
-    int32_t mask = (int32_t) (0x01 << (11 - m));    // set mask for bit corresponding to month
-    if((UMALQURA_MONTHLENGTH[y] & mask) == 0 )
-        return 29;
-    else
-        return 30;
-
-}
-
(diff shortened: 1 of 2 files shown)


icu4c/source/i18n/islamcal.cpp
```diff
@@ -877,7 +867,8 @@ int32_t IslamicUmalquraCalendar::handleGetMonthLength(int32_t extendedYear, int3
         }
         return length;
     }
-    return getUmalqura_MonthLength(extendedYear - UMALQURA_YEAR_START, month);
+    int32_t mask = (int32_t) (0x01 << (11 - month));    // set mask for bit corresponding to month
+    return 29 + (((UMALQURA_MONTHLENGTH[extendedYear - UMALQURA_YEAR_START] & mask) != 0) ? 1 : 0);
```
richgillam, line 871: I think this would probably be a little clearer as

```
return ((UMALQURA_MONTHLENGTH[extendedYear - UMALQURA_YEAR_START] & mask) != 0) ? 30 : 29;
```

richgillam (approved):
Changes look reasonable.  There was one thing I didn't quite understand.

richgillam (approved):
Better.  Thank you!

PR #2860, opened 2024-03-02
ICU-22612 usage of C++11 implies the existence of stdint.h, so remove useless checks

vtorri:
ICU-22612

ALLOW_MANY_COMMITS=true


richgillam (approved):
Nice!  Thank you!

PR #2862, opened 2024-03-04
ICU-22520 Use operator* instead of calling std::optional::value()

roubert:
There's a subtle difference between these two ways of accessing the value of an optional and that is that the `value()` method can throw an exception if there isn't any value, but `operator*` won't do that (it's just undefined behavior if there isn't any value).

ICU4C code never tries to access any optional value without first checking that it exists, but the ability of the `value()` method to throw an exception in case there wasn't any such check first is the reason why `std::exception` symbols previously could show up in debug builds.

This reverts the changes that were made to `dependencies.txt` by commit dc70b5a056b618c014c71e8bfd45f3dd9145e9fe.

##### Checklist


markusicu (approved):
thanks!

PR #2863, opened 2024-03-04
ICU-22520 Refactor CheckedArrayByteSink & u_terminateChars into helper

roubert:
The repeated sequence of allocating a `CheckedArrayByteSink`, calling some function that writes into this, then checking for overflow and returning through `u_terminateChars()` can all be moved into a single shared helper function.

##### Checklist

### DIFF
--- a/icu4c/source/common/bytesinkutil.h
+++ b/icu4c/source/common/bytesinkutil.h
@@ -7,11 +7,14 @@
 #ifndef BYTESINKUTIL_H
 #define BYTESINKUTIL_H
 
+#include <type_traits>
+
 #include "unicode/utypes.h"
 #include "unicode/bytestream.h"
 #include "unicode/edits.h"
 #include "cmemory.h"
 #include "uassert.h"
+#include "ustr_imp.h"
 
 U_NAMESPACE_BEGIN
 
@@ -56,6 +59,40 @@ public:
     static UBool appendUnchanged(const uint8_t *s, const uint8_t *limit,
                                  ByteSink &sink, uint32_t options, Edits *edits,
                                  UErrorCode &errorCode);
+
+    /**
+     * Calls a lambda that writes to a ByteSink with a CheckedArrayByteSink
+     * and then returns through u_terminateChars(), in order to implement
+     * the classic ICU4C C API writing to a fix sized buffer on top of a
+     * contemporary C++ API.
+     *
+     * @param buffer receiving buffer
+     * @param capacity capacity of receiving buffer
+     * @param lambda that gets called with the sink as an argument
+     * @param status set to U_BUFFER_OVERFLOW_ERROR on overflow
+     * @return number of bytes written, or needed (in case of overflow)
+     * @internal
+     */
+    template <typename F,
+              typename = std::enable_if_t<
+                  std::is_invocable_r_v<void, F, ByteSink&, UErrorCode&>>>
+    static int32_t viaByteSinkToTerminatedChars(char* buffer, int32_t capacity,
+                                                F&& lambda,
+                                                UErrorCode& status) {
+        if (U_FAILURE(status)) { return 0; }
+        CheckedArrayByteSink sink(buffer, capacity);
+        lambda(sink, status);
+        if (U_FAILURE(status)) { return 0; }
+
+        int32_t reslen = sink.NumberOfBytesAppended();
+
+        if (sink.Overflowed()) {
+            status = U_BUFFER_OVERFLOW_ERROR;
+            return reslen;
+        }
+
+        return u_terminateChars(buffer, capacity, reslen, &status);
+    }
 
 private:
     static void appendNonEmptyUnchanged(const uint8_t *s, int32_t length,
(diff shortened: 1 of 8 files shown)


icu4c/source/common/bytesinkutil.h
```diff
@@ -83,6 +86,42 @@ class U_COMMON_API CharStringByteSink : public ByteSink {
...
+ * @param status set to U_BUFFER_OVERFLOW_ERROR on overflow
+ * @param lambda that gets called with the sink as an argument
+ * @return number of bytes written, or needed (in case of overflow)
+ * @internal
+ */
+template <
+    typename F,
+    typename = std::enable_if_t<std::is_invocable_r_v<void, F, ByteSink&>>>
+int32_t
+u_withCheckedArrayByteSinkAndTerminateChars(
```
markusicu, line 105: The `u_` prefix is strange for a C++ function in the C++ "icu" namespace. We use `u_` or `uservice_` for public C functions, and `uprv_` for internal C functions.

Why not make this a static function in class ByteSinkUtil as usual?

Also, the CheckedArrayByteSink is an implementation detail. How about
```
    ByteSinkUtil::viaByteSinkToTerminatedChars(lambda, buffer, capacity, errorCode)
```
?

icu4c/source/common/bytesinkutil.h
```diff
@@ -83,6 +86,42 @@ class U_COMMON_API CharStringByteSink : public ByteSink {
...
+        char* buffer,
+        int32_t capacity,
+        UErrorCode& status,
```
markusicu, line 108: It's also unusual to have the UErrorCode not as the last argument. Is this required for the template to work?

If possible, it's nicest to have
- config
- input
- output
- UErrorCode

icu4c/source/common/bytesinkutil.h
```diff
@@ -83,6 +86,42 @@ class U_COMMON_API CharStringByteSink : public ByteSink {
...
+        F&& lambda) {
+    if (U_FAILURE(status)) { return 0; }
+    CheckedArrayByteSink sink(buffer, capacity);
+    lambda(sink);
```
markusicu, line 112: Wouldn't it be cleaner for F to take two arguments -- both the sink and the errorCode -- rather than have the lambda definition grab the errorCode from the context?

I realize that the lambda definition already grabs other arguments from the context, but it seems extra confusing for the errorCode to be used in this wrapper and also have it bypass the wrapper for getting to the actual function inside F. Conceivably, the two UErrorCode's could be different.

icu4c/source/common/bytesinkutil.h
roubert: Done.

icu4c/source/common/bytesinkutil.h
roubert: Done.

icu4c/source/common/bytesinkutil.h
roubert: Done.

icu4c/source/common/bytesinkutil.h
```diff
@@ -57,6 +60,40 @@ class U_COMMON_API ByteSinkUtil {
...
+    /**
+     * Calls a lambda that writes to a ByteSink with a CheckedArrayByteSink
+     * and then returns through u_terminateChars(), in order to implement
+     * the classic ICU4C C API writing to a fix sized buffer on top of a
+     * contemporary C++ API.
+     *
+     * @param buffer receiving buffer
+     * @param capacity capacity of receiving buffer
+     * @param status set to U_BUFFER_OVERFLOW_ERROR on overflow
+     * @param lambda that gets called with the sink as an argument
```
markusicu, line 72:
```suggestion
     * @param lambda that gets called with the sink as an argument
     * @param status set to U_BUFFER_OVERFLOW_ERROR on overflow
```

icu4c/source/common/ulocbuilder.cpp
```diff
@@ -134,17 +134,12 @@ int32_t ulocbld_buildLanguageTag(ULocaleBuilder* builder,
...
-        *err = U_BUFFER_OVERFLOW_ERROR;
-    } else {
-        u_terminateChars(buffer, bufferCapacity, reslen, err);
-    }
-    return reslen;
+    return icu::ByteSinkUtil::viaByteSinkToTerminatedChars(
+        buffer, bufferCapacity,
+        [&](icu::ByteSink& sink, UErrorCode& status) {
+            l.toLanguageTag(sink, status);
+        },
```
markusicu, line 141: question: can this be simplified to just
```suggestion
        l.toLanguageTag(sink, status),
```
?

roubert: Eh, no, how would that work!?

markusicu: it's already a function that takes a sink and a status and returns void

markusicu (approved):

icu4c/source/common/ulocbuilder.cpp
roubert: Well, sure, it is, but the code you suggested would call the function and then pass its void return value as the lambda argument to `viaByteSinkToTerminatedChars()`, and besides that, it's a member function of a class so in order to call it the caller needs to somehow also be given some reference to the object upon which the function should be called (to use as the `this` pointer), which here is the reference `l` which gets passed to `viaByteSinkToTerminatedChars()` through the lambda capture.

PR #2866, opened 2024-03-05
ICU-22520 Refactor CharString & CharStringByteSink into helper

roubert:
The repeated sequence of allocating a `CharString` and `CharStringByteSink`, before calling some function that writes into this, can be moved into a single shared helper function which then is used to give all `ulocimp.h` functions that write to `ByteSink` an overload that instead returns a `CharString`, to make call sites look like perfectly normal C++ code.

##### Checklist


markusicu (approved):
nice!

PR #2872, opened 2024-03-07
ICU-22655 Implement "special" conversion for speed-beaufort, part 2 icu4c

pedberg-icu:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22655

This is the ICU4C counterpart to[ #2847](https://github.com/unicode-org/icu/pull/2847), the ICU4J implementation.

Both implement unit conversion for speed-beaufort using the new CLDR `special` attribute for the `convertUnit` element; this attribute specifies special conversions (such as table-based) between a base unit (e.g. meters/sec in this case) and a scale in which a given scale value represents a range of base unit values (instead of the mapping being represented by a factor and optional offset). Conversions from base to scale map any unit on the range to the specific scale value; conversions from the scale value to base use a value in the middle of the range. The only special scale currently supported is "beaufort".

Both the J and C implementations in these PRs are preliminary, mainly intended to enable CLDR to add tests including test data which will then be imported to ICU (and thus needs to work in ICU), per https://unicode-org.atlassian.net/browse/CLDR-17420. But I am not proud of this implementation, and have filed https://unicode-org.atlassian.net/browse/ICU-22683 for an improved implementation.

### DIFF
--- a/icu4c/source/i18n/units_converter.cpp
+++ b/icu4c/source/i18n/units_converter.cpp
@@ -270,49 +270,96 @@ UBool checkSimpleUnit(const MeasureUnitImpl &unit, UErrorCode &status) {
     return true;
 }
 
+// Map the MeasureUnitImpl for a simpleUnit to a SingleUnitImpl, then use that
+// SingleUnitImpl's simpleUnitID to get the corresponding ConversionRateInfo;
+// from that we get the specialMappingName (which may be empty if the simple unit
+// converts to base using factor + offset instelad of a special mapping).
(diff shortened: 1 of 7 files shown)


richgillam (approved):
Looks good.

FrankYFTang:
The landing of this PR broke "C: MSYS2 GCC x86_64 Release " build see 

https://dev.azure.com/ms/icu/_build/results?buildId=552112&view=logs&j=d69c23bb-4a3f-572d-3cb6-52938b11d573&t=2b5ffab7-f018-50b9-e6ec-1a8ee74e5693&l=34

PR #2880, opened 2024-03-12
ICU-22621 Apply straightforward Clang-Tidy fixes

roubert:
ALLOW_MANY_COMMITS=true

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22621

### DIFF
--- a/icu4c/source/common/unicode/simpleformatter.h
+++ b/icu4c/source/common/unicode/simpleformatter.h
@@ -25,10 +25,8 @@
 U_NAMESPACE_BEGIN
 
 // Forward declaration:
-namespace number {
-namespace impl {
+namespace number::impl {
 class SimpleModifier;
-}
 }
 
 /**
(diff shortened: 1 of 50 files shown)


eggrobin (commented):
I am somewhat skeptical about the braced-init-lists on multi-argument constructors on old classes, as these constructors were added when there was no decision as to whether they should be explicit or not. It seems to me that nowadays we might make something like the (UBool, ConstChar16Ptr, int32_t) constructor explicit, since it is not immediately obvious, at least to this reviewer, how `{true, r, len}` makes up a string.

The {}, the {x}, and the `emplace`s all look good.

eggrobin (approved):

eggrobin:
> That change was approved by the ICU TC

Some reservations about this one were noted in ICU-22621; and with the full diff at hand, I think in cases like `{true, r, len}` I would be more comfortable if others weighed in on the readability. Again, no objections to `return {}` and `return {x}`.

PR #2881, opened 2024-03-12
ICU-22687 Fix genren.pl not to skip C++ functions on Mac

richgillam:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22687

### DIFF
--- a/icu4c/source/tools/genren/genren.pl
+++ b/icu4c/source/tools/genren/genren.pl
@@ -154,9 +154,17 @@ for(;@ARGV; shift(@ARGV)) {
             chop $qtype;
         } elsif ($mode =~ /Mach-O/) {
             ($address, $type, $_) = split(/ /);
-            if(/^_(.*)$/) {
+            if (/^(.*)\(/) {
+            	# on Mac, C++ functions and methods are NOT prefixed with an underscore,
+            	# but do contain their parameter lists (in patentheses)-- remove
+            	# the parameter list
+            	$_ = $1;
+            } elsif(/^_(.*)$/) {
+            	# C function names (and maybe also C++ functions on Linux?) are all
+            	# prefixed with an underscore-- remove it
                 $_ = $1;
             } else {
+            	# skip symbols in any other format
                 next;
             }
         } else {


icu4c/source/tools/genren/genren.pl
```diff
@@ -154,7 +154,9 @@
             chop $qtype;
         } elsif ($mode =~ /Mach-O/) {
             ($address, $type, $_) = split(/ /);
-            if(/^_(.*)$/) {
+            if (/^(.*)\(/) {
+            	$_ = $1;
+            } elsif(/^_(.*)$/) {
```
markusicu, line 159: I think a comment would be nice. I think this does
// Remove a leading underscore.

If you have any idea for why this is done here, maybe you could add that... maybe even an example?

icu4c/source/tools/genren/genren.pl
```diff
@@ -154,7 +154,9 @@
             chop $qtype;
         } elsif ($mode =~ /Mach-O/) {
             ($address, $type, $_) = split(/ /);
-            if(/^_(.*)$/) {
+            if (/^(.*)\(/) {
```
markusicu, line 157: Like below, a comment would be nice for what we do here and why. Maybe
// On macOS, C++ function symbols include their full signature, with (arguments in parens). Remove those.

pedberg-icu (approved):
Thanks for investigating and fixing this! Should have been done long ago 🙃

icu4c/source/tools/genren/genren.pl
richgillam: Good idea.  Done.

As for why it does this, I don't know-- I can confirm that the `_` is there in the names of all the actual C functions in the `nm` output, even on Mac, but I don't know why it does that.

markusicu: Thanks for adding the comments, but please swap them :-)

richgillam: Dang it!  I caught that mistake, but then forgot to do a "git add".  Hopefully fixed now, and I also went ahead and (optimistically?) squashed my commits.

markusicu (approved):
lgtm tnx!

PR #2891, opened 2024-03-14
ICU-22679 Refactor Islamic Calendar and Optmize starting condition of loop

FrankYFTang:
Refactor different rules into implementation of private interface Algorithm.
Since IslamicCalendar is public in Java (not in C++), we cannot put them into different subclass of Calendar and have to let them use the same class and object since the caller UNFORTUNALLY can call setCalculationType and setCivil to change the object to different rule. With this refactoring, we group the formula for the same rule into the same Algorithm 
implementaiton and now we only do the if/switch check in the constructor or during the call of setCalculationType and setCivil only. The calculation operation is then just delegate the work to the assigned concrete Algorithm.

Improve the efficency while the month is very large. Apply the same optimization in PR 2853 which estimate starting point of year iteration based on an inverse calculation.

##### Checklist

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
@@ -11,6 +11,7 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.util.Date;
 import java.util.Locale;
+import java.util.function.IntConsumer;
 
 import com.ibm.icu.impl.CalendarAstronomer;
 import com.ibm.icu.impl.CalendarCache;
@@ -171,13 +172,14 @@ public class IslamicCalendar extends Calendar {
     private static final long HIJRA_MILLIS = -42521587200000L;    // 7/16/622 AD 00:00
 
     /**
-     * Friday EPOC
+     * Friday EPOCH
      */
-    private static final long CIVIL_EPOC = 1948440; // CE 622 July 16 Friday (Julian calendar) / CE 622 July 19 (Gregorian calendar)
+    private static final long CIVIL_EPOCH = 1948440; // CE 622 July 16 Friday (Julian calendar) / CE 622 July 19 (Gregorian calendar)
+                                                     //
     /**
-     * Thursday EPOC
+     * Thursday EPOCH
      */
-    private static final long ASTRONOMICAL_EPOC = 1948439; // CE 622 July 15 Thursday (Julian calendar)
+    private static final long ASTRONOMICAL_EPOCH = 1948439; // CE 622 July 15 Thursday (Julian calendar)
 
     //-------------------------------------------------------------------------
     // Constructors...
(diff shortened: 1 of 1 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -582,28 +579,88 @@ private final static boolean civilLeapYear(int year)
         return (14 + 11 * year) % 30 < 11;
     }
 
+    /**
+     * The yearStart function for Islamic Civil calendar.
+     */
+    private long yearStartIslamicCivil(int year) {
```
richgillam, line 585: The naming here sounds a little weird to my ears.  Maybe `yearStartForIslamicCivil()`?  (Same for the other two functions.)

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -582,28 +579,88 @@ private final static boolean civilLeapYear(int year)
...
+            default:
+                return 0;
         }
-        return ys;
     }
 
+    /**
+     * The monthStart function for Islamic Civil calendar.
+     */
+    private long monthStartIslamicCivil(int year, int month) {
```
richgillam, line 631: Same naming issue as with the year functions.  How about `monthStartForIslamicCivil()`, etc.?

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -732,6 +780,33 @@ static final double moonAge(long time)
     // Calendar framework
     //----------------------------------------------------------------------
 
+    /**
+     * The handleGetMonthLength function for Islamic Civil calendar.
+     */
+    private int handleGetMonthLengthIslamicCivil(int extendedYear, int month) {
```
richgillam, line 786: Same naming issue again: `handleGetMonthLengthForIslamicCivil()`, etc.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -732,6 +780,33 @@ static final double moonAge(long time)
...
+        return ((month+1) % 2) +
+            ((month == DHU_AL_HIJJAH && civilLeapYear(extendedYear)) ? 30 : 29);
```
richgillam, line 788: I find this expression a little hard to read.  I think it might be worth being a little more verbose here, something like:

```
int monthLength = 29;
if (month % 2 == 0) {
    ++monthLength;
}
if (month == DHU_AL_HIJJAH && civilLeapYear(extendedYear)) {
    ++monthLength;
}
return monthLength;
```

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -732,6 +780,33 @@ static final double moonAge(long time)
...
+
+    /**
+     * The handleGetMonthLength function for Islamic Umalqura calendar.
+     */
+    private int handleGetMonthLengthIslamicUmalqura(int extendedYear, int month) {
+        if (extendedYear < UMALQURA_YEAR_START || extendedYear > UMALQURA_YEAR_END) {
+            return handleGetMonthLengthIslamicCivil(extendedYear, month);
+        }
+        return ((UMALQURA_MONTHLENGTH[extendedYear - UMALQURA_YEAR_START] & // calculate year offset into bit map array
+              (0x01 << (11 - month))) == 0 ) ? 29 : 30; // set mask for bit corresponding to month
```
richgillam, line 807: This logic might also be worth exploding out into its constituent pieces, more or less the way it was before this refactoring.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -741,54 +816,64 @@ static final double moonAge(long time)
...
+            default:
+                return 0;
+        }
+    }
 
-        int length;
+    /**
+     * The handleGetYearLength function for Islamic Civil calendar.
+     */
+    private int handleGetYearLengthIslamicCivil(int extendedYear) {
```
richgillam, line 835: Naming again: `handleGetYearLengthForIslamicCivil()`, etc.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -826,6 +911,88 @@ protected int handleGetExtendedYear() {
...
+        internalSet(MONTH, month);
+        internalSet(ORDINAL_MONTH, month);
+        internalSet(DAY_OF_MONTH, dayOfMonth);
+        internalSet(DAY_OF_YEAR, dayOfYear);
+    }
+
+    /**
+     * handleComputeFields function for Islamic Civil calendar.
+     */
+    private void handleComputeFieldsIslamicCivil(long days) {
```
richgillam, line 934: Naming again: `handleComputeFieldsForIslamicCivil()`, etc.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -826,6 +911,88 @@ protected int handleGetExtendedYear() {
...
+        // Use the civil calendar approximation, which is just arithmetic
+        int year  = (int)Math.floor( (30 * days + 10646) / 10631.0 );
```
richgillam, line 936: What are the magic numbers here?

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -844,84 +1011,20 @@ protected int handleGetExtendedYear() {
...
-                        }
-                        break;
-                    }
-                }
-                year = y;
-                month = m;
-            }
+        switch (cType) {
+            case ISLAMIC_CIVIL:
+                handleComputeFieldsIslamicCivil(julianDay - CIVIL_EPOC);
```
richgillam, line 1016: If these symbols are internal and you can change them, the word is spelled "epoch".  `CIVIL_EPOCH` and `ASTRONOMICAL_EPOCH`.

richgillam (commented):
If I'm reading this correctly, this is all refactoring, with no substantive changes to the logic (which is why you don't need to change the unit tests).  I think this refactoring is a huge win-- it makes the code a LOT more readable.

I did have a few stylistic nits, but generally this looks great.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: ok, I decide to refactor everything into different algorithm by adding an internal private interface

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: totally changed now

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: totally changed now.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: done

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: The same magic number as my mechanical refactoring, my guess is not better than yours. 
I do not want to change the outcome. Just want to reduce run time if (cType == ) or switch (cType) ... check here  after the refactoring

FrankYFTang:
Please do not look at it yet. Have some issue of serialization.

FrankYFTang:
> If I'm reading this correctly, this is all refactoring, with no substantive changes to the logic (which is why you don't need to change the unit tests). I think this refactoring is a huge win-- it makes the code a LOT more readable.
> 
> I did have a few stylistic nits, but generally this looks great.

yes, it is refactoring plus an optimization in https://github.com/unicode-org/icu/pull/2853

that part is a logic change. that code is in UmalquraAlgorithm.compute() now
The origional code start the loop of linear estimation from
int y =UMALQURA_YEAR_START-1 and iterate year by year

If the year is huge, then the loop while (d > 0) loop will take a very long time with many iteration. 
so we adjust the starting y by using the formular which is the inverse function of what in the startYear before we run the linear esitmation in the iteration. This will make the algorithm much faster.

richgillam:
Are you ready for me to look at this again?  If not, let me know when you are...

FrankYFTang:
OK, document added and the serialization issue is fix. PTAL

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -319,6 +320,294 @@ public IslamicCalendar(int year, int month, int date, int hour,
...
+         */
+        public long monthStart(int year, int month);
+
+        /**
+         * Return the length (in days) of the given month.
+         *
+         * @param extendedYear  The hijri year
+         * @param month The hijri month, 0-based
+         */
+        public int monthLength(int extendedYear, int month);
```
richgillam, line 365: How's the "extended year" on this function different from the "year" on `monthStart()`?

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -319,6 +320,294 @@ public IslamicCalendar(int year, int month, int date, int hour,
...
+        public int yearLength(int extendedYear);
+
+        /**
+         * Compute the year, month, dayOfMonth, and dayOfYear of the given julian days
+         * and current time. Return an array of 4 integer which contains the resolved
+         * year, month, dayOfMonth and dayOfYear.
+         * @param julianDays
+         * @param current the time in millisecond.
+         */
+        public int[] compute(long julianDays, long current);
```
richgillam, line 381: I'm not super bananas on having this function return an array of `int`, but I think I agree it's not really worth the trouble to define another class just to use as the return value of this function.

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -319,6 +320,294 @@ public IslamicCalendar(int year, int month, int date, int hour,
...
+    private static Algorithm CIVIL_ALGORITHM;
+    private static Algorithm TBLA_ALGORITHM;
+    private static Algorithm UMALQURA_ALGORITHM;
+
+    static {
+        ISLAMIC_ALGORITHM = new IslamicAlgorithm();
+        CIVIL_ALGORITHM = new CivilAlgorithm();
+        TBLA_ALGORITHM = new TBLAAlgorithm();
+        UMALQURA_ALGORITHM = new UmalquraAlgorithm();
+    };
```
richgillam, line 609: Just putting these initializers on the member definitions above doesn't work?

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
richgillam: Fair enough...

richgillam:
> > If I'm reading this correctly, this is all refactoring, with no substantive changes to the logic (which is why you don't need to change the unit tests). I think this refactoring is a huge win-- it makes the code a LOT more readable.
> > I did have a few stylistic nits, but generally this looks great.
> 
> yes, it is refactoring plus an optimization in #2853
> 
> that part is a logic change. that code is in UmalquraAlgorithm.compute() now The origional code start the loop of linear estimation from int y =UMALQURA_YEAR_START-1 and iterate year by year
> 
> If the year is huge, then the loop while (d > 0) loop will take a very long time with many iteration. so we adjust the starting y by using the formular which is the inverse function of what in the startYear before we run the linear esitmation in the iteration. This will make the algorithm much faster.

I'll buy that.

richgillam (approved):
This looks really good!  (I'm trusting you a bit on the math...)

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: no, you cannot do that in java
https://www.baeldung.com/java-static-instance-initializer-blocks

icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
FrankYFTang: ok, I change to passing 4 IntConsumer to consume the calculated result.

richgillam (approved):
Yeah, the `IntConsumer` stuff looks a little better-- thank you!

PR #2894, opened 2024-03-15
ICU-22611 Fixed the RBNF MultiplierSubstitution to only perform floor…

richgillam:
…() on the value being formatted when the

substitution is using a DecimalFormat and its owning rule also has a modulus substitution.  Took out a redundant call to floor().  Added appropriate unit tests.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22611

### DIFF
--- a/icu4c/source/i18n/nfrule.h
+++ b/icu4c/source/i18n/nfrule.h
@@ -66,6 +66,8 @@ public:
     char16_t getDecimalPoint() const { return decimalPoint; }
 
     int64_t getDivisor() const;
+    
+    bool hasModulusSubstitution() const;
 
     void doFormat(int64_t number, UnicodeString& toAppendTo, int32_t pos, int32_t recursionCount, UErrorCode& status) const;
     void doFormat(double  number, UnicodeString& toAppendTo, int32_t pos, int32_t recursionCount, UErrorCode& status) const;
@@ -117,6 +119,9 @@ private:
 
     NFRule(const NFRule &other); // forbid copying of this class
     NFRule &operator=(const NFRule &other); // forbid copying of this class
+    
+    // TODO: temporary hack to allow MultiplierSubstitution to get to formatter's rounding mode
+    friend class MultiplierSubstitution;
 };
 
 U_NAMESPACE_END
--- a/icu4c/source/i18n/nfsubs.cpp
+++ b/icu4c/source/i18n/nfsubs.cpp
@@ -73,6 +73,7 @@ SameValueSubstitution::~SameValueSubstitution() {}
 
 class MultiplierSubstitution : public NFSubstitution {
     int64_t divisor;
+    const NFRule* owningRule;
 
 public:
     MultiplierSubstitution(int32_t _pos,
@@ -80,7 +81,7 @@ public:
         const NFRuleSet* _ruleSet,
         const UnicodeString& description,
         UErrorCode& status)
-        : NFSubstitution(_pos, _ruleSet, description, status), divisor(rule->getDivisor())
+        : NFSubstitution(_pos, _ruleSet, description, status), divisor(rule->getDivisor()), owningRule(rule)
     {
         if (divisor == 0) {
             status = U_PARSE_ERROR;
@@ -103,25 +104,22 @@ public:
     }
 
     virtual double transformNumber(double number) const override {
-        bool doFloor = getRuleSet() != nullptr;
-        if (!doFloor) {
-            // This is a HACK that partially addresses ICU-22313.  The original code wanted us to do
-            // floor() on the result if we were passing it to another rule set, but not if we were passing
-            // it to a DecimalFormat.  But the DurationRules rule set has multiplier substitutions where
-            // we DO want to do the floor() operation.  What we REALLY want is to do floor() any time
-            // the owning rule also has a ModulusSubsitution, but we don't have access to that information
-            // here, so instead we're doing a floor() any time the DecimalFormat has maxFracDigits equal to
-            // 0.  This seems to work with our existing rule sets, but could be a problem in the future,
-            // but the "real" fix for DurationRules isn't worth doing, since we're deprecating DurationRules
-            // anyway.  This is enough to keep it from being egregiously wrong, without obvious side
-            // effects.     --rtg 8/16/23
-            const DecimalFormat* decimalFormat = getNumberFormat();
-            if (decimalFormat == nullptr || decimalFormat->getMaximumFractionDigits() == 0) {
-                doFloor = true;
-            }
-        }
-        
-        if (doFloor) {
+        // Most of the time, when a number is handled by an NFSubstitution, we do a floor() on it, but
+        // if a substitution uses a DecimalFormat to format the number instead of a ruleset, we generally
+        // don't want to do a floor()-- we want to keep the value intact so that the DecimalFormat can
+        // either include the fractional part or round properly.  The big exception to this is here in
+        // MultiplierSubstitution.  If the rule includes two substitutions, the MultiplierSubstitution
+        // (which is handling the larger part of the number) really _does_ want to do a floor(), because
+        // the ModulusSubstitution (which is handling the smaller part of the number) will take
+        // care of the fractional part.  (Consider something like `1/12: <0< feet >0.0> inches;`.)
+        // But if there is no ModulusSubstitution, we're shortening the number in some way-- the "larger part"
+        // of the number is the only part we're keeping.  Even if the DecimalFormat doesn't include the
+        // fractional part in its output, we still want it to round.  (Consider something like `1/1000: <0<K;`.)
+        // (TODO: The kRoundFloor thing is a kludge to preserve the previous floor-always behavior.  What we
+        // probably really want to do is just set the rounding mode on the DecimalFormat to match the rounding
+        // mode on the RuleBasedNumberFormat and then pass the number to it whole and let it do its own rounding.
+        // But before making that change, we'd have to make sure it didn't have undesirable side effects.)
+        if (getRuleSet() != nullptr || owningRule->hasModulusSubstitution() || owningRule->formatter->getRoundingMode() == NumberFormat::kRoundFloor) {
             return uprv_floor(number / divisor);
         } else {
             return number / divisor;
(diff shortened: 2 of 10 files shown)


grhoten (dismissed):
Hmm... I disagree with this repurpose of RBNF. The CompactDecimalFormat should be used in such situations, and this usage with RBNF is not expected. I'd prefer not to support this use case.

Here's why. RBNF is meant to handle things like "1 million 1". It's not meant to round to "1 million" by default.  That remainder is supposed to be handled with other rules, like `>>`. Such rounding should really be handled with UNUM_SIGNIFICANT_DIGITS_USED or other kinds of rounding attributes before getting into these rules.

grhoten:
I guess from my perspective, there's 2 ways to round depending on the context.  If it's a unit, then you will want to round like compact decimal format. If it's in reference to the magnitude of the number, then this is the wrong behavior.  For example, we have this syntax in German.

```
%spellout-numbering-year:
-x: minus >>;
x.x: =0.0=;
0: =%spellout-numbering=;
1100/100: <<­hundert[­>>];
2000: =%spellout-numbering=;
```

This style appears a lot in multiple spellout-numbering-year rules. We really do want floor in such situations. There should be several languages that already expect floor instead of rounding for year pronunciation.  Other languages include: ar, en, fr, he, ru and sv. I don't want to break those.

icu4c/source/i18n/nfsubs.cpp
```diff
@@ -103,25 +104,30 @@ class MultiplierSubstitution : public NFSubstitution {
...
-            // anyway.  This is enough to keep it from being egregiously wrong, without obvious side
-            // effects.     --rtg 8/16/23
-            const DecimalFormat* decimalFormat = getNumberFormat();
-            if (decimalFormat == nullptr || decimalFormat->getMaximumFractionDigits() == 0) {
-                doFloor = true;
-            }
-        }
-        
-        if (doFloor) {
+//        bool doFloor = getRuleSet() != nullptr;
```
sffc, line 107: Should you delete this code instead of commenting it out?

icu4c/source/i18n/nfsubs.cpp
```diff
@@ -73,14 +73,15 @@ SameValueSubstitution::~SameValueSubstitution() {}
...
+    const NFRule* owningRule;
 
 public:
     MultiplierSubstitution(int32_t _pos,
         const NFRule *rule,
         const NFRuleSet* _ruleSet,
         const UnicodeString& description,
         UErrorCode& status)
-        : NFSubstitution(_pos, _ruleSet, description, status), divisor(rule->getDivisor())
+        : NFSubstitution(_pos, _ruleSet, description, status), divisor(rule->getDivisor()), owningRule(rule)
```
sffc, line 84: Just making sure, but now that you are saving the pointer to `NFRule*`, can you verify that it does indeed always outlive the `MultiplierSubstitution`? (in C++ this requires manual verification; in Rust it is automatic.) Also, please add docs to this function about the lifetime of the `NFRule*`.

icu4c/source/i18n/nfrule.h
```diff
@@ -66,6 +66,8 @@ class NFRule : public UMemory {
     char16_t getDecimalPoint() const { return decimalPoint; }
 
     int64_t getDivisor() const;
+    
+    bool hasModulusSubstitution() const;
```
sffc, line 70: Please add docs explaining what a "modulus substitution" is and how it relates to flooring the number

richgillam:
> I guess from my perspective, there's 2 ways to round depending on the context. If it's a unit, then you will want to round like compact decimal format. If it's in reference to the magnitude of the number, then this is the wrong behavior. For example, we have this syntax in German.
> 
> ```
> %spellout-numbering-year:
> -x: minus >>;
> x.x: =0.0=;
> 0: =%spellout-numbering=;
> 1100/100: <<­hundert[­>>];
> 2000: =%spellout-numbering=;
> ```
> 
> This style appears a lot in multiple spellout-numbering-year rules. We really do want floor in such situations. There should be several languages that already expect floor instead of rounding for year pronunciation. Other languages include: ar, en, fr, he, ru and sv. I don't want to break those.

Do we currently have test cases that cover this situation?  If not, can you suggest some I should add?

If I'm grokking that ruleset correctly, I don't believe it'll be affected by the change I'm proposing.  The substitution that uses the `DecimalFormat` is a same-value substitution, not a multiplier substitution, so it's unaffected.  The multiplier substitution that _leads_ to that same-value substitution is likewise unaffected because it's calling into another RBNF ruleset (well, actually it's calling into the _same_ ruleset), and not using a `DecimalFormat` to format its value.  In other words, the extra indirection there is isolating you from my change.  If we didn't have that extra indirection, we'd potentially have a problem (at least in say the "<<hundert;" rule that omits the modulus substitution), but it seems like the extra indirection is integral to what that whole ruleset is doing.

I'll poke around and see if I see anything that looks more questionable, and/or I'll try to add unit tests to cover this kind of situation, if we don't already have them...

richgillam:
>This style appears a lot in multiple spellout-numbering-year rules. We really do want floor in such situations. There should be several languages that already expect floor instead of rounding for year pronunciation.

@grhoten I added the following to `IntlTestRBNF::TestGermanSpellout()` in an attempt to test the issue you have with rules in the style you cite above:

```
        static const char* testDataYear[][2] = {
            { "101", "ein\\u00adhundert\\u00adeins" },
            { "900", "neun\\u00adhundert" },
            { "1,001", "ein\\u00adtausend\\u00adeins" },
            { "1,100", "elf\\u00adhundert" },
            { "1,101", "elf\\u00adhundert\\u00adeins" },
            { "1,234", "zw\\u00f6lf\\u00adhundert\\u00advier\\u00adund\\u00addrei\\u00dfig" },
            { "2,001", "zwei\\u00adtausend\\u00adeins" },
            { "10,001", "zehn\\u00adtausend\\u00adeins" },
            { "-100", "minus ein\\u00adhundert" },
            { "12.34", "12,3" },
            { nullptr, nullptr }
        };

        status = U_ZERO_ERROR;
        formatter->setDefaultRuleSet("%spellout-numbering-year", status);
        if (U_SUCCESS(status)) {
            logln("testing year rules");
            doTest(formatter, testDataYear, false);
        }
```

It works just fine, in both C++ and Java.  And as I explained, I tried to be careful not to break things like this.  So again, what's so bad about this change?  What am I breaking?

icu4c/source/i18n/nfsubs.cpp
richgillam: Oops.  You're absolutely right.  Removed.

icu4c/source/i18n/nfsubs.cpp
richgillam: The rule owns the substitution; this is just a backpointer.  When the rule is deleted, it'll take the substitution down with it.  See `NFRule::~NFRule()`.

icu4c/source/i18n/nfrule.h
richgillam: Done.  Please let me know what you think.

grhoten:
> @grhoten I added the following to `IntlTestRBNF::TestGermanSpellout()` in an attempt to test the issue you have with rules

Perhaps this is something for discussion in a larger group, like the ICU-TC. The changes here are trying to turn something like the value 1900 into 1 of 2 possible values. This is counterintuitive, and emulation of CompactDecimalFormat in RBNF is undesirable. 

The current behavior will turn 1900 into "one thousand". That is desirable, and I believe you're trying to keep that behavior.
This changed behavior will turn 1900 into "2 thousand". That is undesirable and confusing due to different rounding behavior.

The context is different, and I believe that CompactDecimalFormat is supposed to be used as approximation by default. While RBNF is meant to be precise by default, and that the other existing rounding options need to be used to apply it to RBNF to make it emulate CompactDecimalFormat.

richgillam:
> The current behavior will turn 1900 into "one thousand". That is desirable, and I believe you're trying to keep that behavior. This changed behavior will turn 1900 into "2 thousand". That is undesirable and confusing due to different rounding behavior.

Just to be clear, I'm not changing things so that 1900 goes from "one thousand" to "two thousand"-- it still comes out as "one thousand".  I *am* changing things so that 1900 goes from "1 thousand" to "2 thousand".

I agree that this is inconsistent and potentially confusing; I'm not sure I agree that it's _undesirable_.  The whole motivation for this change is that somebody out there in the field was depending on this behavior and I broke it.  That is, before my earlier change to address durations, we were _already_ rounding instead of flooring in this situation.  I'm trying to solve that problem without re-breaking duration formatting.

It's true that one shouldn't use RBNF to format _either_ compact numbers _or_ durations, but for a long time we didn't have better answers, and we still don't have a better answer for durations in C.  I think this compromise, while kludgey, is defensible and doesn't really have that much of a downside.

>Perhaps this is something for discussion in a larger group, like the ICU-TC.

I'll put it on the agenda for tomorrow morning's meeting.

grhoten:
I still recommend using CompactDecimalFormat instead of the rules from this ticket. I also recommend being able to get the behavior prior to this pull request by setting the rounding behavior as necessary on RuleBasedNumberFormat.

richgillam:
>As per the TC discussion, this oddity in formatting differences should be documented in the API, and some of this older behavior be brought back.

I've added documentation to explain what's going on.  PTAL.

richgillam:
> I still recommend using CompactDecimalFormat instead of the rules from this ticket.

I agree.

>I also recommend being able to get the behavior prior to this pull request by setting the rounding and significant digits behavior as necessary on RuleBasedNumberFormat.

I added a hack to use the formatter's rounding mode to get the previous flooring behavior.  It's ugly, but I'm afraid of making the "appropriate" change for fear that it'll have more wide-ranging side effects.  PTAL and let me know if this hack is okay.

icu4c/source/i18n/unicode/rbnf.h
```diff
@@ -436,7 +436,16 @@ enum URBNFRuleSetTag {
...
+ *     <td>Divide the number by the rule's divisor, perform floor() on the quotient,
+ *         and format the resulting value.<br>
+ *         If there is a DecimalFormat pattern between the &lt; characters and the
+ *         rule does NOT also contain a &gt;&gt; substitution, we DON'T perform
+ *         floor() on the quotient-- the quotient is passed through to the DecimalFormat
+ *         intact.  That is, for the value 1,900:<br>
+ *         - "1/1000: &lt;&lt; thousand;" will produce "one thousand"<br>
+ *         - "1/1000: &lt;0&lt; thousand;" will produce "2 thousand" (NOT "1 thousand")<br>
+ *         - "1/1000: &lt;0&lt; seconds &gt;0&gt; milliseconds;" will produce "1 second 900 milliseconds"
+ *     </td>
```
grhoten, line 448: This is helpful.

grhoten (approved):

FrankYFTang:
This merge break EnvTest see
https://github.com/unicode-org/icu/actions/runs/8401177375/job/23009269867

to reproduce run the intltest with LC_ALL set to af_ZA , ar_KW.utf8 , be_BY@latin etc
Got the following
```
TestDFRounding {
      FAIL: check failed for -1,440, expected -1.4K but got -1,4K
   
      } ERRORS (1) in TestDFRounding
```

FrankYFTang:
see my additional comments in https://unicode-org.atlassian.net/browse/ICU-22611 I think you could fix it by using the verson of RuleBasedNumberFormat constructo which take an explicit Locale parameter so it will not format the digits and separator depending on the default locale, which may not result as "1.4 K" but rather different digits or decimal point char

PR #2895, opened 2024-03-15
ICU-22643 Fixed it so that clients can make UChar be uint16_t by chan…

richgillam:
…ging line 370 in umachine.h to "#if 0"

and still successfully build ICU4C.

##### Checklist


icu4c/source/common/unicode/umachine.h
```diff
@@ -377,9 +377,17 @@ typedef int8_t UBool;
 #endif
 
 #if defined(U_COMBINED_IMPLEMENTATION) || defined(U_COMMON_IMPLEMENTATION) || \
-        defined(U_I18N_IMPLEMENTATION) || defined(U_IO_IMPLEMENTATION)
+        defined(U_I18N_IMPLEMENTATION) || defined(U_IO_IMPLEMENTATION) || \
+        defined(U_TOOLUTIL_IMPLEMENTATION)
     // Inside the ICU library code, never configurable.
     typedef char16_t UChar;
+#elif defined(T_CTEST_IMPLEMENTATION)
+    // internally to ctestfw, we want to use char16_t in C++ and uint_16 in C
```
markusicu, line 385:
```suggestion
    // internally to ctestfw, we want to use char16_t in C++ and uint16_t in C
```

markusicu, line 385: Why do we need to default to uint16_t in C? We are on C11 now, and char16_t is part of the standard.
https://en.cppreference.com/w/c/string/multibyte/char16_t

icu4c/source/extra/uconv/uconv.cpp
```diff
@@ -20,6 +20,10 @@
  * See source code repository history for changes.
  */
 
+#ifndef UCHAR_TYPE
+#define UCHAR_TYPE char16_t
+#endif
```
markusicu, line 25: umachine.h says
```
    // The intltest Makefile #defines UCHAR_TYPE=char16_t,
    // so we only #define it to uint16_t if it is undefined so far.
```
Should we define this in more makefiles, so that we don't need these definitions in the source files?
Or do you think it's more robust to do this in the source files, so that it works out of the box in any build system?

markusicu (dismissed):

icu4c/source/extra/uconv/uconv.cpp
aheninger: Both choices seem unfortunate. Moving the def to makefiles would require separate defs  for Windows and regular make. But having all the repeated defs in the various source files also feels wrong.

I have no strong opinion. Leave it as proposed, and hope that some year we can remove all support for UChar being anything but char16_t, everywhere?

icu4c/source/common/unicode/umachine.h
richgillam: Oops.  Will fix.

richgillam: That wasn't true when I originally made this fix in Apple ICU, but that was before we changed minimum C versions.  Let me try it again now...

icu4c/source/extra/uconv/uconv.cpp
richgillam: Having that snippet in all those source files feels really bad to me, but I originally fixed this in Apple ICU under time pressure, so I went with the easiest thing I could think of.

I'd forgotten Windows has all different makefiles, and having to change them scares me, since I don't have any way to test it.

What I was originally thinking of was to add something analogous to `U_COMMON_IMPLEMENTATION` for the unit-test files and then just checking for that in `umachine.h`.  Although I guess that still amounts to messing with the makefiles.  Let me see if I can get that to work...

>...and hope that some year we can remove all support for UChar being anything but char16_t, everywhere?

Well, we're here because I inadvertently committed the unchanged OSICU code (with `#if 1`), and it caused a whole mess of build failures here at Apple.  So we have numerous projects here that still depend on `UChar` being `uint16_t`.  I should probably start working on people to change that dependency, but I don't know how quickly I can get that to happen.

icu4c/source/common/unicode/umachine.h
richgillam: Well, **THIS** is bad...

I just figured out the hard way that Apple platforms don't actually support C11.  Or maybe they do partially, but they don't have the `<uchar.h>` header, where `char16_t` is defined.  From looking at the existing bug reports about this, I'm not getting the impression anyone's going to fix it on Apple platforms any time soon.  So I either need to keep this code as is or simplify it for other platforms and come up with a workaround for Apple platforms...

richgillam:
Okay, I think I've got this (more or less) straightened out.  PTAL.

By the way, I wonder if we should consider having a CI check (or some less-frequent automated test?) that changes that line to `#if 0` and verifies that everything still builds on all our platforms?

icu4c/source/common/unicode/umachine.h
```diff
@@ -52,6 +52,16 @@
...
 #include <stddef.h>
 
+#if U_CPLUSPLUS_VERSION == 0
+#if !U_PLATFORM_IS_DARWIN_BASED
+#include <uchar.h>
+#else
+// The <uchar.h> file does not exist in Apple SDKs (as of 3/15/24),
+// so we have to define char16_t ourselves
+typedef uint16_t char16_t;
+#endif // !U_PLATFORM_IS_DARWIN_BASED
```
markusicu, line 62: praise: I like this solution to Apple compilers having this gap in C11 support.

icu4c/source/common/unicode/umachine.h
```diff
@@ -52,6 +52,16 @@
...
+#endif // U_CPLUSPLUS_VERSION == 0
+
```
markusicu, line 64: optional: It feels like this might fit better in ptypes.h, sometime after it includes platform.h.
Looking at the various header files,
- platform.h tells us about the platform
- ptypes.h levels out platform differences -- recently most of it was removed since we rely on C++17 and C11 (ICU-22612)
- umachine.h defines ICU-specific types on top of the previous headers

But we can move this snippet later if you like.

icu4c/source/common/unicode/umachine.h
```diff
@@ -377,15 +387,14 @@ typedef int8_t UBool;
...
+        defined(U_TOOLUTIL_IMPLEMENTATION) || defined(T_CTEST_IMPLEMENTATION)
     // Inside the ICU library code, never configurable.
     typedef char16_t UChar;
 #elif defined(UCHAR_TYPE)
     typedef UCHAR_TYPE UChar;
-#elif U_CPLUSPLUS_VERSION != 0
-    typedef char16_t UChar;  // C++
 #else
-    typedef uint16_t UChar;  // C
+    typedef char16_t UChar;  // C++
```
markusicu, line 397:
```suggestion
    typedef char16_t UChar;
```

markusicu:
> I wonder if we should consider having a CI check (or some less-frequent automated test?) that changes that line to `#if 0` and verifies that everything still builds on all our platforms?

Good idea.
We should be able to configure ICU4C such that UCHAR_TYPE=uint16_t is set on the command line, probably in the CPPFLAGS.

richgillam:
> Good idea. We should be able to configure ICU4C such that UCHAR_TYPE=uint16_t is set on the command line, probably in the CPPFLAGS.

Would setting `UCHAR_TYPE` to `uint16_t` on the command line have the same effect as changing that line in the source code to `#if 0`?  They're supposed to, but I was a little worried those two mechanisms wouldn't do exactly the same thing and might expose different bugs.

markusicu:
> > Good idea. We should be able to configure ICU4C such that UCHAR_TYPE=uint16_t is set on the command line, probably in the CPPFLAGS.
> 
> Would setting `UCHAR_TYPE` to `uint16_t` on the command line have the same effect as changing that line in the source code to `#if 0`? They're supposed to, but I was a little worried those two mechanisms wouldn't do exactly the same thing and might expose different bugs.

It should work just about the same, and as well as what the makefiles do, because when you change it to `#if 0` we do `define UCHAR_TYPE uint16_t` unless it's already defined, so if you predefine it like that it's the same as the manual edit.

icu4c/source/common/unicode/umachine.h
richgillam: Thank you!

icu4c/source/common/unicode/umachine.h
richgillam: Nah, I went ahead and moved it.  I think you're right.

icu4c/source/common/unicode/umachine.h
richgillam: Oops.  Fixed.

richgillam:
>We should be able to configure ICU4C such that UCHAR_TYPE=uint16_t is set on the command line, probably in the CPPFLAGS.

Okay, so we have two choices here: I can try to add the CI check as part of this PR, but somebody's going to have to walk me through doing that, as I've never messed without CI checks and don't know anything about how to work with them.  Or we can file a separate ticket and somebody (ideally not me) can do that later.  My preference is for option 2.

roubert:
I don't think that setting `UCHAR_TYPE` in the makefile was a good idea to begin with, as it goes wholly against the idea that this is something that's supposed to be user configurable (and will generate a flood of warnings when the user sets `UCHAR_TYPE`), and requiring that to now be spread to more makefiles also introduces the new problem that this then becomes a magic incantation that everyone will need to know in order to be able to correctly create a new subdirectory with a new makefile.

So I've looked at what could be done here instead and tried to keep the core idea from this PR (a typedef for `char16_t` when building C on Mac), but use that in what I think would be a more robust way. I've prepared PR #2913 with that solution now, please take a look at that and say what you think (and please test it yourself to verify that it really solves your original problem).

Long term, I believe that the `UCHAR_TYPE` feature is getting obsolete, but for this upcoming release the changes in my PR would make it work as intended also on Mac while simplifying the surrounding logic and getting it all a step closer to eventually being deleted.

richgillam:
@roubert has a better solution to this problem in https://github.com/unicode-org/icu/pull/2913, so I'm closing this PR in favor of that one.

richgillam:
> So I've looked at what could be done here instead and tried to keep the core idea from this PR (a typedef for `char16_t` when building C on Mac), but use that in what I think would be a more robust way. I've prepared PR #2913 with that solution now, please take a look at that and say what you think (and please test it yourself to verify that it really solves your original problem).

The new fix looks much better than what I have here, and it seems to build correctly whichever way I have that #if set in `umachine.h`.  I gave you a ship-it and closed this PR.

> Long term, I believe that the `UCHAR_TYPE` feature is getting obsolete, but for this upcoming release the changes in my PR would make it work as intended also on Mac while simplifying the surrounding logic and getting it all a step closer to eventually being deleted.

Everybody keeps saying this, but I'm not sure it's true.  The whole reason I made this change was that when I last integrated open-source ICU into Apple ICU, I took the change of `UChar` to `char16_t`, and broke the builds of a bunch of client projects that really expected `UChar` to be `uint16_t`.  I can try to get everybody to change, but I don't know how much luck I'll have.  Apple might have to live with `UChar` being `uint16_t` forever.

roubert:
> Everybody keeps saying this, but I'm not sure it's true. The whole reason I made this change was that when I last integrated open-source ICU into Apple ICU, I took the change of `UChar` to `char16_t`, and broke the builds of a bunch of client projects that really expected `UChar` to be `uint16_t`. I can try to get everybody to change, but I don't know how much luck I'll have. Apple might have to live with `UChar` being `uint16_t` forever.

Sure, it's clear that we'll have to keep using `uint16_t` in the Apple world for the foreseeable future, but the feature that I believe to be obsolete (and therefore possible to remove soon) is the ability to at compile time set `UCHAR_TYPE` willy-nilly.

roubert:
> I can try to add the CI check as part of this PR,

While I think that it would be very valuable to have a CI that builds ICU4C exactly like it's being done inside of Apple (using the Apple toolchain, running on macOS, &c., &c.), that'd be beyond the scope of this particular ticket and for now the simple build test that we already have is sufficient to verify that the `UCHAR_TYPE` feature continues to work and I just propose adding all public header files to it: PR #2917

richgillam:
> > I can try to add the CI check as part of this PR,
> 
> While I think that it would be very valuable to have a CI that builds ICU4C exactly like it's being done inside of Apple (using the Apple toolchain, running on macOS, &c., &c.), that'd be beyond the scope of this particular ticket and for now the simple build test that we already have is sufficient to verify that the `UCHAR_TYPE` feature continues to work and I just propose adding all public header files to it: PR #2917

That's fine with me.

PR #2896, opened 2024-03-15
ICU-22679 Remove duplicate code inside Islamic calendar

FrankYFTang:
##### Checklist

### DIFF
--- a/icu4c/source/i18n/islamcal.cpp
+++ b/icu4c/source/i18n/islamcal.cpp
@@ -717,7 +717,7 @@ IslamicCivilCalendar* IslamicCivilCalendar::clone() const {
 * from the Hijri epoch, origin 0.
 */
 int64_t IslamicCivilCalendar::yearStart(int32_t year) const{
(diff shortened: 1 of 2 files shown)


richgillam (commented):
This looks okay, but how do the changes in this PR relate to the changes in https://github.com/unicode-org/icu/pull/2891, which appears to be for the same ticket?  It looks like they do very similar things, and to the same code.  Is this meant to be applied after 2891 lands?  Why keep it separate instead of just making the changes to that PR?

FrankYFTang:
> This looks okay, but how do the changes in this PR relate to the changes in #2891, which appears to be for the same ticket? It looks like they do very similar things, and to the same code. Is this meant to be applied after 2891 lands? Why keep it separate instead of just making the changes to that PR?

PR 2891 is Java. This is C++

richgillam (approved):

richgillam:
> PR 2891 is Java. This is C++

Duh!  Okay, then, approved!

That said, this looks fairly different from the Java implementation.  Can they be made more parallel, or is that not worth the trouble?

PR #2905, opened 2024-03-19
ICU-22698 Clean up CalendarAstronomer

FrankYFTang:
1. Remove dead code inside CalendarAstronomer that was not used.
The following methhods are not used and unneeded for the Chinese and Islamic calendar usage and therefore removed:
getJulianCentury(), getGreenwichSidereal(), getLocalSidereal(), lstToUT(), eclipticToHorizon(), getSunPosition(), getSunRiseSet(), getMoonPhase(), getMoonRiseSet(), 
We also removed 6 double member data from CalendarAstronomer since they were only used by removed functions and/or never with complete implementation (for example, lat/long in this class). This save us 48 bytes per CalendarAstronomer object.

3. Since all the current usage call setTime then get a value, the cache inside CalendarAstronomer is not used across different call. Change to allocate the object on stack instead of using mutex to share a singlton since it is cheap to create a new one and there are no benefit to share across different calls. This reduce unnecessary multi thread concerns.
4. Remove many unneeded member data inside CalendarAstronomer that is needed for methods that no code use. 

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22698

### DIFF
--- a/icu4c/source/i18n/astro.cpp
+++ b/icu4c/source/i18n/astro.cpp
@@ -242,7 +242,7 @@ inline static  double normPI(double angle)  {
  * @deprecated ICU 2.4. This class may be removed or modified.
  */
 CalendarAstronomer::CalendarAstronomer():
-  fTime(Calendar::getNow()), fLongitude(0.0), fLatitude(0.0), fGmtOffset(0.0), moonPosition(0,0), moonPositionSet(false) {
+  fTime(Calendar::getNow()), moonPosition(0,0), moonPositionSet(false) {
   clearCache();
 }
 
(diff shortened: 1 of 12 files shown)


richgillam (approved):
I think this looks great.  The one thing that's worrying me is Why are all those dead functions there to begin with?  Are they things we used to use but don't need anymore, or are they things we put in expecting somebody to need them at some point?  If we decide to add more calendar types in the future, will any of them need the calculations that you're removing?

I guess if they do, they can put them back, but it'd be bad if they had to reimplement them from scratch when somebody had already done that.  I guess we could include a comment in the code recommending that somebody go to the git history to resurrect those functions if they need them, or something like that...

FrankYFTang:
> I think this looks great. The one thing that's worrying me is Why are all those dead functions there to begin with? Are they things we used to use but don't need anymore, or are they things we put in expecting somebody to need them at some point? If we decide to add more calendar types in the future, will any of them need the calculations that you're removing?
> 
> I guess if they do, they can put them back, but it'd be bad if they had to reimplement them from scratch when somebody had already done that. I guess we could include a comment in the code recommending that somebody go to the git history to resurrect those functions if they need them, or something like that...

The were landed 25 years ago (by John Fitzpatrick committed on Feb 9, 2000  ) in https://github.com/unicode-org/icu/commit/4857053b2e8b75cb908e899a5ee17e3e14053dac#diff-5293cf77eecdd25fd9ca38901069f9dd3a77f421fec6bcd10b46c03b883ba2ee

and then port to C++ 21 years ago (by srl295 committed on Nov 24, 2003 ) in https://github.com/unicode-org/icu/commit/24f0431f7834313e8cb138b73afc50d0ad29fa51

both with the following comments

```
 * <strong>WARNING:</strong> This class is very early in its development, and
 * it is highly likely that its API will change to some degree in the future.
 * At the moment, it basically does just enough to support {@link IslamicCalendar}
 * and a Chinese calendar class that is still in development.
```
It seems the author for Java is Laura Werner and the C++ are Laura Werner and Alan Liu.

Also both have comments with
```
    /* These aren't used yet, but they'll be needed for sunset calculations
     * and equatorial to horizon coordinate conversions
     */
    private double fLongitude = 0.0;
    private double fLatitude  = 0.0;
    private long   fGmtOffset = 0;
```

It seems no major developement / improvement on this class the last 21/25 years. From my understanding a complete correct Islamic calendar implementation require the calculation of sunset but we do not have that in the last 21/25 years and I do not believe anyone intend to improve it at the near future. Also, there are no real unit test code testing them. (there are unit test code which log the out but not verification of the calculated value.) I believe even if we eventualy want to add the feature, the developer has to write it again and cannot just assume a deadcode without test will calculate the correct output for them.

FrankYFTang:
ok, I did a little bit more reading of Morrison and figure what these deleted function were needed:

getSunPosition(), eclipticToHorizon(), getSunPosition(), getSunRiseSet() are needed when supporting the Hebrew and Islamic day system which the day starting at sun set (Morrison 197-203)

getMoonPhase(), getMoonRiseSet(), getJulianCentury(), getGreenwichSidereal(), getLocalSidereal(), lstToUT() are needed to implement a "Islamic calendar ... approximat at best" (Morrison 206). Also, "The calculation of such an astronomical Islamc calendar ... is quite intricate and not generally accepted" (Morrison 90).

I assume that is why these functions were never used to implement a calendar, because even if we did, that calendar is "not generally accepted" (Morrison 90) anyway.

                                         Works Cited
Morrison, Philip, et al. "Calendrical Calculations: The Millennium Edition." American Scientist 90.1 (2002): 80.

foolip:
It turns out this code was used in Chromium, I've filed https://crbug.com/419726206.

PR #2906, opened 2024-03-19
ICU-22633 Fix hang on large negative day in hebrew calendar

FrankYFTang:
Check error status and return error in the loop

Left over bug fix for ICU-22633 bug fix for Hebrew Calendar. 
Found by fuzzer in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=67463

It is not really a infinity loop but a very slow loop because we should return the error inside the loop right after it happen.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22633


richgillam (approved):
LGTM

PR #2908, opened 2024-03-19
ICU-22691 Delete abandoned ICU4C sample code, fix what remains

roubert:
ALLOW_MANY_COMMITS=true

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22691


icu4c/source/samples/cal/cal.c
```diff
@@ -1,766 +0,0 @@
...
-    puts("Usage: icucal [OPTIONS] [[MONTH] YEAR]");
-    puts("");
-    puts("Options:");
-    puts("  -h, --help        Print this message and exit.");
-    puts("  -v, --version     Print the version number of cal and exit.");
-    puts("  -l, --long        Use long names.");
-    puts("");
-    puts("Arguments (optional):");
-    puts("  MONTH             An integer (1-12) indicating the month to display");
-    puts("  YEAR              An integer indicating the year to display");
```
markusicu, line 215: @srl295 this one is fairly elaborate and interesting. And most of this C API hasn't changed much, right? Worth something?

icu4c/source/samples/datefmt/README.TXT
```diff
@@ -1,107 +0,0 @@
...
-
-  #Date/Time/Number Formatting Support
-  9:30am - 10:30am
-  Alan Liu
-
-  Topics:
-  1. What is the date/time support in ICU?
-  2. What is the timezone support in ICU?
-  3. What kind of formatting and parsing support is available in ICU, i.e.
-  NumberFormat, DateFormat, MessageFormat?
```
markusicu, line 24: @srl295 this also looks elaborate and interesting. Worth something?

icu4c/source/samples/legacy/README
```diff
@@ -1,44 +0,0 @@
-Copyright (C) 2016 and later: Unicode, Inc. and others.
-License & terms of use: http://www.unicode.org/copyright.html
-
-Copyright (c) 2002, International Business Machines Corporation and others. All Rights Reserved.
-This example demonstrates running an instance of  ICU 1.8.1. together with a current version of ICU. It only tests u_getVersion and several collation APIs. 
-
-Generally, one should be able to simultaneously use one or more versions of ICU 2.0 or higher and one version of ICU 1.8.1 or lower. 
-
-What is it all about: Let's say you have a 10 Tb database indexed using ICU 1.8.1. sortkeys. New ICU comes out, with neat new features you would like to use, but also with new sortkeys and you don't care to reindex your 10 Tb database. What to do then??? You can use ICU 1.8.1. in one of your compilation units and current version in all the others. So, you can use old collation until you decide to reindex.
```
markusicu, line 9: This is actually useful because we, and some of our users, still rely on this functionality. It's why we have urename.h.

This kind of sample does require two versions of ICU to build against. One of them need not be 1.8.1.

icu4c/source/samples/uciter8/uciter8.c
```diff
@@ -1,364 +0,0 @@
...
-*   encoding:   UTF-8
-*   tab size:   8 (not used)
-*   indentation:4
-*
-*   created on: 2003jan10
-*   created by: Markus W. Scherer
-*
-*   This file contains sample code that illustrates reading
-*   8-bit Unicode text leniently, accepting a mix of UTF-8 and CESU-8
-*   and also accepting single surrogates.
```
markusicu, line 24: OMG I did this because some user needed it... good riddance!

FrankYFTang (approved):

markusicu (changes requested):
Let's see what others think for a few hours...

sffc (commented):
As I said via email:

> It seems like the right move for clients would be to more deeply integrate these samples into docs or tutorials, since ICU tends to lack useful sample code in general. But, if no one is stepping up to do that (I can't remember the last time we've given a priority to docs), maybe we don't need obsolete code that doesn't build sitting around.

roubert:
> Let's see what others think for a few hours...

We've been waiting to see what others think about the sample code for nearly four months now, I think that ought to be sufficient time. If anyone actually cared, they would long since have voiced their opinion by now.

icu4c/source/samples/cal/cal.c
markusicu: Let's please keep this one for now.

icu4c/source/samples/datefmt/README.TXT
markusicu: Let's please keep this one for now.

icu4c/source/samples/legacy/README
markusicu: Let's please keep this one for now.

srl295:
looking..

icu4c/source/common/unicode/ucnv.h
```diff
@@ -344,8 +344,6 @@ ucnv_compareNames(const char *name1, const char *name2);
  * other than its an alias starting with the letters "cp". Please do not
  * associate any meaning to these aliases.</p>
  *
- * \snippet samples/ucnv/convsamp.cpp ucnv_open
- *
```
srl295, line 348: if the sample worked, would it be helpful to have a _compiling_ sample here in the API docs? That was the purpose of it

icu4c/source/samples/legacy/README
srl295: should be easy enough to get this behavior using actions.  In fact, the 'old' version of ICU could come from `libicu-dev` or such from a package manager if building on ubuntu-latest.

icu4c/source/samples/datefmt/README.TXT
srl295: makes sense to me, as long as it builds

roubert:
> i have copies.

It's all stored in a revision control system, if anyone in the future comes to the conclusion that they want some of the deleted sample code after all, they could resurrect it from the revision history.

markusicu (approved):
Ok. Yes, I should have responded months ago, and yes, we can always find the deleted code in ICU 74.

icu4c/source/samples/translit/answers/main_3.cpp
```diff
@@ -9,7 +9,7 @@
  ***********************************************************************/
 
 #include "unicode/translit.h"
-#include "unicode/rbt.h"
+#include "rbt.h"
```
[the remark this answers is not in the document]
roubert: > Ideally sample code should not need internal header files...

Indeed, but let's make it compile at all before making it correct.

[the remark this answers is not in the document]
srl295: perhaps that was taken private..

srl295:
> > i have copies.
> 
> It's all stored in a revision control system, if anyone in the future comes to the conclusion that they want some of the deleted sample code after all, they could resurrect it from the revision history.

Certainly.

PR #2913, opened 2024-03-20
ICU-22643 Make UCHAR_TYPE work as intended with C++11, C11 & macOS.

roubert:
Since ICU4C requires C++11 and C11 the `char16_t` data type can be used everywhere, as long as the `uchar.h` header file gets included when building as C (where it isn't a keyword as in C++), but this doesn't work on macOS which for unknown reasons lacks the `uchar.h` header file and therefore needs a workaround to define `char16_t`.

For any ICU4C internal code the `char16_t` data type should always be used and `UCHAR_TYPE` ignored, which this PR solves by setting `U_ALL_IMPLEMENTATION` in its own build system.

There shouldn't really be any reason to ever use the `UCHAR_TYPE` feature on any platform that doesn't use the autoconf build system so I think it'd be fine to leave this without adding the `U_ALL_IMPLEMENTATION` define to any other build systems (which would be something that I couldn't test myself).

##### Checklist

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -711,9 +711,9 @@
     /*
      * Notes:
      * C++11 and C11 require support for UTF-16 literals
-     * TODO: Fix for plain C. Doesn't work on Mac.
+     * Doesn't work on Mac C11 (see workaround in ptypes.h).
      */
-#   if U_CPLUSPLUS_VERSION >= 11 || (defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L)
+#   if defined(__cplusplus) || !U_PLATFORM_IS_DARWIN_BASED
 #       define U_HAVE_CHAR16_T 1
 #   else
 #       define U_HAVE_CHAR16_T 0
--- a/icu4c/source/common/unicode/ptypes.h
+++ b/icu4c/source/common/unicode/ptypes.h
@@ -51,4 +51,16 @@
 
 #include <stdint.h>
 
+// C++11 and C11 both specify that the data type char16_t should exist, C++11
+// as a keyword and C11 as a typedef in the uchar.h header file, but not all
+// implementations (looking at you, Apple, spring 2024) actually do this, so
+// ICU4C must detect and deal with that.
+#if !defined(__cplusplus) && !defined(U_IN_DOXYGEN)
+#   if U_HAVE_CHAR16_T
+#       include <uchar.h>
+#   else
+        typedef uint16_t char16_t;
+#   endif
+#endif
+
 #endif /* _PTYPES_H */
--- a/icu4c/source/common/unicode/umachine.h
+++ b/icu4c/source/common/unicode/umachine.h
@@ -370,22 +370,14 @@ typedef int8_t UBool;
 #if 1
     // #if 1 is normal. UChar defaults to char16_t in C++.
     // For configuration testing of UChar=uint16_t temporarily change this to #if 0.
-    // The intltest Makefile #defines UCHAR_TYPE=char16_t,
-    // so we only #define it to uint16_t if it is undefined so far.
-#elif !defined(UCHAR_TYPE)
+#else
 #   define UCHAR_TYPE uint16_t
 #endif
 
-#if defined(U_COMBINED_IMPLEMENTATION) || defined(U_COMMON_IMPLEMENTATION) || \
-        defined(U_I18N_IMPLEMENTATION) || defined(U_IO_IMPLEMENTATION)
-    // Inside the ICU library code, never configurable.
+#if defined(U_ALL_IMPLEMENTATION) || !defined(UCHAR_TYPE)
     typedef char16_t UChar;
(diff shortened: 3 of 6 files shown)


markusicu (commented):
@roubert thanks this looks like an improved version of PR #2895
@richgillam ok?

richgillam (approved):
Gotta admit, this is much simpler and nicer-looking than my attempt to fix this problem, and it does seem to work-- I changed "#if 1" to "#if 0" on line 370 of umachine.h, and everything still built successfully and passed its unit tests.  I didn't go the extra step to make sure that when I do that, `UChar` in the public headers was actually `uint16_t`, but that seems to be what `uchar_type_build_test` is doing...

I'll close my PR.

markusicu (approved):

moubctez:
This breaks the build on NetBSD where _uchar.h_ is not available. From what I've checked so far, _uchar.h_ might only be available on platforms with glibc. This patch fixes the problem for me:

```
--- common/unicode/ptypes.h.orig	2024-04-16 21:08:53.000000000 +0000
+++ common/unicode/ptypes.h
@@ -56,7 +56,7 @@
 // implementations (looking at you, Apple, spring 2024) actually do this, so
 // ICU4C must detect and deal with that.
 #if !defined(__cplusplus) && !defined(U_IN_DOXYGEN)
-#   if U_HAVE_CHAR16_T
+#   if U_HAVE_CHAR16_T && U_PLATFORM_IS_LINUX_BASED
 #       include <uchar.h>
 #   else
         typedef uint16_t char16_t;
```

roubert:
> This breaks the build on NetBSD where _uchar.h_ is not available.

The C11 standard specifies that `uchar.h` must be available. Are you certain that you're building in C11 mode? ICU4C requires C11 since release 75.

moubctez:
> The C11 standard specifies that `uchar.h` must be available. Are you certain that you're building in C11 mode? ICU4C requires C11 since release 75.

Yes, ICU builds with std=c11.

roubert:
> Yes, ICU builds with std=c11.

I'm sorry, but I'm unsure what you mean here. Do you mean that ICU now builds successfully for you using `-std=c11`, or do you mean that it still fails to build for you even though you've verified that you're using `-std=c11`?

If it's the latter, please try compiling this trivial test case, just this single line in a file:

```
#include <uchar.h>
```

That should be possible to compile using a command line like this:

```
$ cc -std=c11 -c c11.c
```

If that doesn't work for you, you have a problem with the C11 support in your compiler toolchain, which you'll need to file a bug report about (if it hasn't been reported already). And if you tell exactly what compiler and standard library it is that you're using, we could add a workaround for that to ICU4C while waiting for that bug to get fixed.

You can also add `-DUCHAR_TYPE=uint16_t` to your configuration yourself if you need a workaround that works immediately.

moubctez:
No, neither ICU nor the trivial test case build with `-std=c11`.

This explains the issue https://www.gnu.org/software/gnulib/manual/html_node/uchar_002eh.html

richgillam:
`<uchar.h>` isn't there in Apple builds either-- apparently our compiler/stdlib isn't 100% C11-compliant.  I thought I had special logic in Apple's fork of ICU that typedefs `char16_t` to be `uint16_t`, but I can't find that now.  I think it might just be that I typdefed `UChar` to always be `uint16_t` if `<uchar.h>` wasn't there.  But I think we need to account in the open-source project for the fact that some major platforms that purport to support C11 still don't have a `<uchar.h>` file.

roubert:
> I thought I had special logic in Apple's fork of ICU that typedefs `char16_t` to be `uint16_t`,

As of release 75 there's an exception for Apple built in (see `platform.h`).

> But I think we need to account in the open-source project for the fact that some major platforms that purport to support C11 still don't have a <uchar.h> file.

It's still possible for anyone to add `-DUCHAR_TYPE=uint16_t` to the configuration if needed.

richgillam:
>As of release 75 there's an exception for Apple built in (see platform.h).

Ah, so there is.  I thought I'd cherry-picked this fix into the Apple ICU, but apparently I haven't.  Apologies.

markusicu:
> > But I think we need to account in the open-source project for the fact that some major platforms that purport to support C11 still don't have a <uchar.h> file.
> 
> It's still possible for anyone to add `-DUCHAR_TYPE=uint16_t` to the configuration if needed.

ICU should build out of the box if possible, and if it doesn't hurt other platforms.
I don't think it's reasonable to tell everyone one a variety of Linuxes/Unixes that they have to configure with a weird override.
I am also disappointed that apparently many build systems that purport to support C11 don't do so for an important and trivial part of the spec.
So... I would say that we are open to pull requests getting ICU to work without uchar.h on platforms that don't have it.
My preference is for preprocessor checks rather than autoconf magic (because not everyone uses autoconf to build ICU).

roubert:
> So... I would say that we are open to pull requests getting ICU to work without uchar.h on platforms that don't have it.

Of course, it makes perfect sense to add workarounds for all known and commonly used platform / toolchain combinations that need it, just like we already do for Apple.

PR #2914, opened 2024-03-20
ICU-22701 Bugfix: Make test independent of the default locale

roubert:
The original intention behind this test case was to use the root locale, but ures`_getFunctionalEquivalent()` is implemented by calling `ures_open()` which sets `URES_OPEN_LOCALE_DEFAULT_ROOT` which will cause the default locale to be loaded before the root locale.

To avoid that, pick a locale other than the root locale for the test.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22701


richgillam (approved):
LOKTM

FrankYFTang (approved):

PR #2915, opened 2024-03-21
ICU-22679 Clean up Calendar code.

FrankYFTang:
1. Remove redudant implementation of default system century by using macro
2. Fold long if / else block if one block return.

##### Checklist

### DIFF
--- a/icu4c/source/i18n/coptccal.cpp
+++ b/icu4c/source/i18n/coptccal.cpp
@@ -11,6 +11,7 @@
 
 #if !UCONFIG_NO_FORMATTING
 
+#include "gregoimp.h"
 #include "umutex.h"
 #include "coptccal.h"
 #include "cecal.h"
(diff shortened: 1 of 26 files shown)


icu4c/source/i18n/coptccal.cpp
```diff
@@ -62,22 +63,19 @@ CopticCalendar::handleGetExtendedYear(UErrorCode& status)
...
+    int32_t era = internalGet(UCAL_ERA, CE);
+    if (era == BCE) {
+        return 1 - internalGet(UCAL_YEAR, 1); // Convert to extended year
     }
-    return eyear;
+    if (era == CE){
+        return internalGet(UCAL_YEAR, 1); // Default to year 1
+    }
+    status = U_ILLEGAL_ARGUMENT_ERROR;
+    return 0;
```
richgillam, line 78: I just have to editorialize for a minute here (I'm not really asking you to change anything).

For years, I worked on a team that was very dogmatic about having one return statement per function and regularly got slapped around for having more than one return, and so seeing a bunch of code in here where you're doing the exact opposite and preferring mid-function returns over else clauses just makes me twitch uncontrollably.

My own preference is somewhere in the middle-- I like if statements to be balanced most of the time.  I think it improves readability and helps protect against early returns accidentally bypassing needed cleanup code.  But I also think always requiring a single return can really mess up readability in a lot of functions for more complex logic.  

To my knowledge ICU doesn't have a set of style guidelines that mandate anything in particular.  You're the principal maintainer of these files now, so I guess you should do it however you find it the most comfortable.  But I flinch a little at changes like this that are all about coding style.

richgillam (approved):
I love the way you're eliminating tons of redundant code here, and (despite my earlier comment) I'm actually fairly happy with all the early returns and the way they reduce indentation.

There's a part of me that wonders a bit whether a macro is the best tool for this particular job, though.  It makes the code look better without actually changing anything under the covers.  I'm wondering if it's possible to move the common code up into the base class in some way.  Or can't you do that without breaking API compatibility?

I'm fine with this going in as-is, but I did wonder about that...

icu4c/source/i18n/coptccal.cpp
FrankYFTang: > having one return statement per function

That will work great if we ALSO ensure if all functions are < 20 lines (including comment line), one TAB = 2 chars, and break longer function to several smaller functions. With that two constraits together, I will also prefer one return statement per function. The problem in ICU code right now is we have TAB=4 chars and we currently have a lot of 
```
if (U_FAILURE(status)) {
   return;
}
// 10 lines of code
if (U_FAILURE(status)) {
   return;
}
// another 3 pages of code
return
```
if we maintain the "one return statement per function" style, then it will become

```
if (U_SUCCESS(status)) {
  // 10 lines of code
  if (U_SUCECESS(status)) {
      // another 3 pages of code
  }
}
return
```
and the if block is keep indenting and moving right and greater than 80 col per line. (The fact we use TAB=4 chars make the block moving left x2 than TAB=2)

FrankYFTang:
> I'm wondering if it's possible to move the common code up into the base class in some way.

I am researching how could we refactor that to use template to address that. But for later

FrankYFTang:
I somehow landed without aware I broke the header test. Somehow the system let me merge the PR even the test fail. I am fixing it now. sorry

FrankYFTang:
fix in https://github.com/unicode-org/icu/pull/2918 sorry

icu4c/source/i18n/coptccal.cpp
richgillam: Yeah, I understand.  I just keep flashing back to the beatings I got on a previous team when I did that kind of thing.  :-)

PR #2917, opened 2024-03-21
ICU-22643 Verify that UCHAR_TYPE works for all public header files.

roubert:
This'll cause `make tests` to fail for any developer who accidentally breaks the `UCHAR_TYPE` feature in any one of these header files.

A future improvement could possibly be to generate this file dynamically at build time, but new public header files with C API aren't added very often and dynamic generation has its own set of maintenance costs too.

##### Checklist


markusicu (approved):
Thanks.

Note that we have some other compilation tests that dynamically generate code to test with each header file in a set of folders, for example https://github.com/unicode-org/icu/blob/main/icu4c/source/test/hdrtst/testtagsguards.sh -- that kind of test should also work for UCHAR_TYPE, by writing temp .cpp files with the define and one of the includes.

richgillam (approved):
LOKTM

PR #2918, opened 2024-03-21
ICU-22679 Fix broken header test

FrankYFTang:
This fix the breakage of U_HIDE_INTERNAL_API test on source/i18n/unicode/gregocal.h
caused by the landing of PR 2915

##### Checklist

### DIFF
--- a/icu4c/source/i18n/unicode/calendar.h
+++ b/icu4c/source/i18n/unicode/calendar.h
@@ -2564,7 +2564,6 @@ Calendar::internalSet(UCalendarDateFields field, int32_t value)
     fIsSet[field]     = true; // Remove later
 }
 
-#ifndef U_HIDE_INTERNAL_API
 /**
  * Macro for the class to declare it override
  * haveDefaultCentury, defaultCenturyStart, and
@@ -2576,6 +2575,7 @@ Calendar::internalSet(UCalendarDateFields field, int32_t value)
     virtual UDate defaultCenturyStart() const override; \
     virtual int32_t defaultCenturyStartYear() const override;
 
+#ifndef U_HIDE_INTERNAL_API
 inline int32_t  Calendar::weekNumber(int32_t dayOfPeriod, int32_t dayOfWeek)
 {
   return weekNumber(dayOfPeriod, dayOfPeriod, dayOfWeek);


icu4c/source/i18n/unicode/calendar.h
```diff
@@ -2564,7 +2564,6 @@ Calendar::internalSet(UCalendarDateFields field, int32_t value)
     fIsSet[field]     = true; // Remove later
 }
 
-#ifndef U_HIDE_INTERNAL_API
```
markusicu, line 2567: optional: you might be able to use `#ifndef U_FORCE_HIDE_INTERNAL_API` for the macro, similar to how we use `U_FORCE_HIDE_DRAFT_API` etc. elsewhere ([code search](https://github.com/search?q=repo%3Aunicode-org%2Ficu%20U_FORCE_HIDE_&type=code)) to make the header test happy without breaking normal use of the non-FORCE config

markusicu (approved):
rslgtm

icu4c/source/i18n/unicode/calendar.h
FrankYFTang: I am confused (and still is) why we have the following inside [i18n/unicode/gregocal.h](https://github.com/unicode-org/icu/commit/fbc1f33e7e8afeb92ed7498fca1a1686d8b6b9c4#diff-ded31ae20bc831e7bdac6226b31f22422cf1b5eb5908d08c450d1619561a6155)
```

    /**
     * @return true if this calendar has the notion of a default century
     * @internal 
     */
    virtual UBool haveDefaultCentury() const override;

    /**
     * @return the start of the default century
     * @internal
     */
    virtual UDate defaultCenturyStart() const override;
    /**
     * @return the beginning year of the default century
     * @internal 
     */
    virtual int32_t defaultCenturyStartYear() const override;
```
which was marked by @internal but NOT wrap between 

#ifndef U_HIDE_INTERNAL_API and  #endif

markusicu: Because, if you compile with U_HIDE_INTERNAL_API, the vtable becomes different from the real one exported from the DLL, and then the program would crash. So we have some limitations of where we can use U_HIDE_xxx_API.

We might be able to use U_FORCE_HIDE_INTERNAL_API.

FrankYFTang:
somehow my PR got stuck in the CI-ICU4C and CI-ICU4J here
The two pipeline won't see my branch and won't run it. 
https://dev.azure.com/ms/icu/_build?definitionId=360

https://dev.azure.com/ms/icu/_build?definitionId=631

PR #2921, opened 2024-03-22
ICU-22534 Update version number to 75.1

rp9-next:
##### Note: 
- Maven [pom.xml](https://github.com/unicode-org/icu/blob/6eb43b164806efdbe0d55fc0db9c60a8cbcb971f/icu4j/pom.xml#L11) files are not touched because those should be changed only during GA.
GA Change: `<version>75.0.1-SNAPSHOT</version>` to `<version>75.1</version>`
- [`icu4j/releases_tools/shared.sh`](https://github.com/unicode-org/icu/blob/main/icu4j/releases_tools/shared.sh) seems to be accurate for RC
- Changing [`<api.doc.version>`](https://github.com/unicode-org/icu/blob/6eb43b164806efdbe0d55fc0db9c60a8cbcb971f/icu4j/pom.xml#L71)  to RC is deprecated.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22534


markusicu (approved):
I don't remember if we wanted to move to 75.1-SNAPSHOT but I realize that if we did that, then CLDR wouldn't auto-update to the rc snapshots so maybe this is right.

markusicu:
@rp9-next please merge -- or are you waiting for additional eyes on this?

PR #2931, opened 2024-03-27
ICU-22261 Add UCONFIG_NO_MF2 flag that can be used to disable MessageFormat 2 functionality

catamorphism:
This is a followup to #2579 that adds a compile-time flag to disable the MessageFormat 2 functionality.
 

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22261 


catamorphism:
I don't think I have the ability to request reviewers, but paging @markusicu @echeran 

I'm also not sure if I should file a separate bug for this. (I just used the same bug number as in #2579 for now.)

richgillam (approved):
Didn't actually test to make sure this works, but it _looks_ good.

Are there MF2 tests in `cintltst`?  If so, they should probably have this guard around them as well.

catamorphism:
> Are there MF2 tests in `cintltst`?

No, only in `intltest` (which I assumed was the right place, but I hope someone will tell me if it's not :) )

richgillam:
> > Are there MF2 tests in `cintltst`?
> 
> No, only in `intltest` (which I assumed was the right place, but I hope someone will tell me if it's not :) )

I thought you also had C API, and it'd probably be good to do at least a superficial set of tests to make sure your C wrappers do call through appropriately to the C++ implementation.  But I definitely wouldn't ask you to do that _now_, and since you don't have anything there now, I think you're good to go.  :-)

catamorphism:
> I thought you also had C API, and it'd probably be good to do at least a superficial set of tests to make sure your C wrappers do call through appropriately to the C++ implementation. But I definitely wouldn't ask you to do that _now_, and since you don't have anything there now, I think you're good to go. :-)

Ah, no -- no C API yet.

icu4c/source/test/intltest/messageformat2test_utils.h
```diff
@@ -4,15 +4,17 @@
 #ifndef _TESTMESSAGEFORMAT2_UTILS
 #define _TESTMESSAGEFORMAT2_UTILS
 
+#if !UCONFIG_NO_FORMATTING
```
markusicu, line 7: Please make sure that before any of these config switches are checked, at least one of the public header files has been included so that the compiler has seen uconfig.h.

If in doubt, include unicode/utypes.h then check the config switches then include the rest of the headers.

catamorphism: Fixed in [d744f2c](https://github.com/unicode-org/icu/pull/2931/commits/d744f2cd1ed135048b620ce10f2c2e31ae58ec3b)

markusicu (approved):
lgtm please squash

PR #2932, opened 2024-03-27
ICU-9972 Fix Chinese/Dangi Calendar getActualMaximum(UCAL_DAY_OF_YEAR)

FrankYFTang:
Also fix ICU-12620, which is mark duplicate of ICU-9972 just now, and ICU-22258.

Separate the new year and winter solstice cache since the calculated value for these two calendar are mostly but not always the same due to slightly different observation timeZone.

Remove the epochYear and zoneAstroCalc from the member data and instead return them from a getStting() method with the two caches since all four of them are constant per subclass of ChineseCalendar and do not need to be different per object.

The known issues in the TestLimit is caused by both Calendar get/put the value from the same cache while the calculated result depends on the timeZone zoneAstroCalc.

We do not have this issue in Java because both the new year and winter solstice cache in ICU4J 's ChineseCalendar are not global but per object cache. So the value in these cache is always agree with to the result of the same zoneAstro.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-9972

### DIFF
--- a/icu4c/source/i18n/chnsecal.cpp
+++ b/icu4c/source/i18n/chnsecal.cpp
@@ -24,7 +24,7 @@
 #include "umutex.h"
 #include <float.h>
 #include "gregoimp.h" // Math
-#include "astro.h" // CalendarAstronomer
+#include "astro.h" // CalendarAstronomer and CalendarCache
 #include "unicode/simpletz.h"
 #include "uhash.h"
 #include "ucln_in.h"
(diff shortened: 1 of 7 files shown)


richgillam (approved):
I had a little trouble following this, but yeah, I think this makes sense.

markusicu:
@FrankYFTang approved & tests pass; ready to merge before we create the release branch for ICU 75?

PR #2933, opened 2024-03-28
ICU-22534 Regenerate urename.h for the ICU 75 release candidate.

richgillam:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22534


icu4c/source/common/unicode/urename.h
```diff
@@ -118,6 +118,7 @@
 #define allowedHourFormatsCleanup U_ICU_ENTRY_POINT_RENAME(allowedHourFormatsCleanup)
 #define cmemory_cleanup U_ICU_ENTRY_POINT_RENAME(cmemory_cleanup)
 #define dayPeriodRulesCleanup U_ICU_ENTRY_POINT_RENAME(dayPeriodRulesCleanup)
+#define decltype U_ICU_ENTRY_POINT_RENAME(decltype)
```
richgillam, line 121: This is the only thing in here that looked weird to me.  My best guess is that it's coming from ICU-22651 or ICU-21107, both of which were submitted by @roubert .  Is this okay and what we want?

richgillam: Well, it seems to be causing build errors all over the place, so I'm taking it out, but I'm also wondering whether the fact that it showed up in `urename.h` at all points to a problem somewhere in the code...

markusicu: Something weird in how the Perl script parses some C++ function signature. We definitely can't have "decltype" to get a version suffix. Follow-up ticket for 76 please?

markusicu (approved):

icu4c/source/common/unicode/urename.h
richgillam: Yeah.  I'll try to figure out what's going on when I go back in there to regenerate `urename.h` again for the final release.

PR #2956, opened 2024-04-10
ICU-22261 Improved identification of the MSVC compiler and library

roubert:
It's not uncommon for code for Windows to use the `_MSC_VER` preprocessor macro to identify that it's being compiled for Windows so it's also not uncommon for compilers other than the real MSVC compiler to also set this to be able to compile such code.

It's also not possible to use `_MSC_VER` to determine whether the C++ standard library implementation used is the Microsoft STL.

Clang will however refuse to instantiate a template with a forward declared type, so the code that currently does this needs to be moved to after the type has been properly defined, which in turn makes MSVC warn that those templates aren't instantiated, so those warnings need to be disabled, but then the disabling of warning C4661 doesn't work any longer (for some unknown reason) but this can be resolved by properly deleting the non-existent operators instead of disabling the warning.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22261

### DIFF
--- a/icu4c/source/common/unicode/localpointer.h
+++ b/icu4c/source/common/unicode/localpointer.h
@@ -162,11 +162,11 @@ protected:
     T *ptr;
 private:
     // No comparison operators with other LocalPointerBases.
-    bool operator==(const LocalPointerBase<T> &other);
-    bool operator!=(const LocalPointerBase<T> &other);
+    bool operator==(const LocalPointerBase<T> &other) = delete;
+    bool operator!=(const LocalPointerBase<T> &other) = delete;
     // No ownership sharing: No copy constructor, no assignment operator.
-    LocalPointerBase(const LocalPointerBase<T> &other);
-    void operator=(const LocalPointerBase<T> &other);
+    LocalPointerBase(const LocalPointerBase<T> &other) = delete;
+    void operator=(const LocalPointerBase<T> &other) = delete;
 };
 
 /**
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -207,6 +207,17 @@
 #endif
 
 /**
+ * \def U_REAL_MSVC
+ * Defined if the compiler is the real MSVC compiler (and not something like
+ * Clang setting _MSC_VER in order to compile Windows code that requires it).
+ * Otherwise undefined.
+ * @internal
+ */
+#if (defined(_MSC_VER) && !(defined(__clang__) && __clang__)) || defined(U_IN_DOXYGEN)
+#   define U_REAL_MSVC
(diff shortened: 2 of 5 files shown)


roubert:
@catamorphism, could you please review this, it'll make MF2 build for Windows both with MSVC and Clang.

icu4c/source/common/unicode/localpointer.h
```diff
@@ -162,11 +162,11 @@ class LocalPointerBase {
     T *ptr;
 private:
     // No comparison operators with other LocalPointerBases.
-    bool operator==(const LocalPointerBase<T> &other);
-    bool operator!=(const LocalPointerBase<T> &other);
+    bool operator==(const LocalPointerBase<T> &other) = delete;
```
catamorphism, line 165: I think there's other code (not related to MF2) that also includes pragmas to disable warning 4661, for the same reason, that could be removed now that these operators are deleted. Could be a different PR, though.

catamorphism (approved):
LGTM, thanks for fixing this!

icu4c/source/common/unicode/localpointer.h
markusicu: ... and fixes for other code should be under a different ticket. Let's finish up ICU 75.

markusicu (approved):
thanks!

icu4c/source/common/unicode/localpointer.h
roubert: Yes, those pragmas will be possible to delete now, but between the release candidate and the release I'd like to do only necessary changes.

icu4c/source/common/unicode/platform.h
```diff
@@ -206,6 +206,17 @@
...
 #endif
 
+/**
+ * \def U_REAL_MSVC
+ * Defined if the compiler is the real MSVC compiler (and not something like
+ * Clang setting _MSC_VER in order to compile Windows code that requires it).
+ * Otherwise undefined.
+ * @internal
+ */
+#if (defined(_MSC_VER) && !(defined(__clang__) && __clang__)) || defined(U_IN_DOXYGEN)
```
mihnita, line 216: Sorry, I had some days off last week.

Shouldn't we use `_LIBCPP_VERSION` instead of `__clang__`, as recommended by our colleague working in this area? \
This seems to be more of a difference in libraries, not in the compiler.

roubert: `_LIBCPP_VERSION` identifies LLVM's libc++, which often is used together with the Clang compiler but isn't really tied to any compiler. The expression here attempts to identify the compiler, regardless of which implementation of the standard library is being used.

PR #2961, opened 2024-04-12
ICU-22726 Add new CLDR 45 currency code XCG also to gCurrencyList

roubert:
@pedberg-icu This seems to have been forgotten in PR #2811.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22726


markusicu (approved):
Thanks!!
@roubert @pedberg-icu @yumaoka do we need a similar change in Java?

roubert:
> Thanks!! @roubert @pedberg-icu @yumaoka do we need a similar change in Java?

I can't find any such hardcoded list of currencies in ICU4J.

markusicu:
> > do we need a similar change in Java?
> 
> Using GitHub repo search for the adjacent XDR:
> 
> * Another C++ list: https://github.com/unicode-org/icu/blob/dbb3fe957a52921aabf433d77ae87c33de328801/icu4c/source/i18n/measunit.cpp#L403
> * In Java, a mapping from letter code to numeric code: https://github.com/unicode-org/icu/blob/dbb3fe957a52921aabf433d77ae87c33de328801/tools/currency/src/com/ibm/icu/dev/tool/currency/NumericCodeData.java#L394
> 
> Do we have instructions for how to add a currency code? I found https://github.com/unicode-org/icu/blob/main/tools/currency/readme.txt but it refers to "ant", and it's not clear that the tool would update all of the relevant files.
> 
> Adding @dra8an FYI

@roubert could you please add XCG to the two additional hardcoded lists that I found? Or else note why it doesn't belong there?

@dra8an @DraganBesevic @pedberg-icu please take a look at the questions here.

markusicu:
> > > do we need a similar change in Java?
> > 
> > Using GitHub repo search for the adjacent XDR:
> > 
> > * Another C++ list: https://github.com/unicode-org/icu/blob/dbb3fe957a52921aabf433d77ae87c33de328801/icu4c/source/i18n/measunit.cpp#L403
> > * In Java, a mapping from letter code to numeric code: https://github.com/unicode-org/icu/blob/dbb3fe957a52921aabf433d77ae87c33de328801/tools/currency/src/com/ibm/icu/dev/tool/currency/NumericCodeData.java#L394
> > 
> > Do we have instructions for how to add a currency code? I found https://github.com/unicode-org/icu/blob/main/tools/currency/readme.txt but it refers to "ant", and it's not clear that the tool would update all of the relevant files.
> > Adding @dra8an FYI
> 
> @roubert could you please add XCG to the two additional hardcoded lists that I found? Or else note why it doesn't belong there?

Actually, looking at the list in measunit.cpp:
- It's a list of measurement units, not just currencies. We may not want to insert a currency code manually, in case the array indexes are meaningful and need to be kept in sync with some other data that indexes into here.
- Is there a tool or process, or docs, for how this list gets updated?
- @younies @sffc @richgillam @pedberg-icu

And about NumericCodeData.java:
- This is a hardcoded mapping between currency letter codes and numeric codes.
- There is no numeric code for XCG in CLDR 45 supplementalData.xml -- FYI @macchiati @srl295
- (And XCG is not yet listed on https://en.wikipedia.org/wiki/ISO_4217, and https://en.wikipedia.org/wiki/Caribbean_guilder does not show its numeric code.)
- We should probably not manually add a mapping in ICU4J if CLDR doesn't know about it -- and can't, if we don't know what the numeric code is.

> @dra8an @DraganBesevic @pedberg-icu please take a look at the questions here.

So... maybe we just claim that this PR is good enough for ICU 75 and merge it? @richgillam

richgillam:
@DraganBesevic wanted to ask @pedberg-icu , but if we can't raise him, we both say go for it.

macchiati:
I believe that XCG is not valid until some time in 2025, so is not an issue
for 45/75

srl295:
XCG is 532 https://www.six-group.com/dam/download/financial-information/data-center/iso-currrency/amendments/dl-currency-iso-amendment-176.pdf

PR #2966, opened 2024-04-15
ICU-21172 fix install race

rossburton:
The generic recursive target calls target-local so also adding it to the dependency list results in races due to install-local being executed twice in parallel.  For example, install-manx can fail if the two install processes race and one process tries to chown a file that the other process has just deleted.

Also install-manx should be a phony target, and for clarity use `$^` instead of `$?` in the install command.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21172

### DIFF
--- a/icu4c/source/Makefile.in
+++ b/icu4c/source/Makefile.in
@@ -77,7 +77,7 @@ EXTRA_DATA =
 
 ## List of phony targets
 .PHONY : all all-local all-recursive install install-local install-udata install-udata-files install-udata-dlls		\
-install-recursive clean clean-local clean-recursive distclean		\
+install-recursive install-manx clean clean-local clean-recursive distclean		\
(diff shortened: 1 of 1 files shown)


rossburton:
Regarding the CLA, Arm has a corporate CLA signed.

srl295 (approved):
Looks great, thanks.

srl295:
tested locally LGTM,. why is CI stuck?

markusicu:
> tested locally LGTM,. why is CI stuck?

I was thinking maybe there is a problem between an old commit and newer changes to the CI system, and tried to "squash" this commit via the UI, but then I got a permission-denied error message.

@rossburton would you mind rebasing your commit to trigger a fresh CI round?
@echeran FYI

PR #2970, opened 2024-04-18
ICU-22721 Prevent inconsistent ICU4J Maven deploys via CI

echeran:
Update the CI workflow that deploys Maven artifacts to Github for ICU4J so that:

* We don't publish the `icu4j` artifact to Github for release (non-snapshot) versions. This preserves Maven Central as the single source of truth for `icu4j` artifacts for release versions
  - Note: we keep the current behavior for snapshot versions, which is to publish to Github
*  We automatically prevent any manual attempts to publish a release version artifact

Trial runs in my personal fork:
- succeeding workflow run of the PR branch, triggered manually: https://github.com/echeran/icu/actions/runs/8730613164/job/23954716091
  * the condition execution of steps within the job shows that the workflow output variables are being captured as expected
- a previous iteration that failed when the version number looked like a release version (`1`), triggered manually: https://github.com/echeran/icu/actions/runs/8730537899/job/23954536566
  -   the automatic check is activated to prevent a manual deploy of a release version artifact

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22721

### DIFF
--- a/.github/workflows/maven.yaml
+++ b/.github/workflows/maven.yaml
@@ -17,6 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       packages: write
+    outputs:
+      version: ${{ steps.mvn-proj-version.outputs.version }}
+      version-type: ${{ steps.mvn-proj-version-type.outputs.version-type }}
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
         with:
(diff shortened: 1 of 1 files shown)


pedberg-icu (dismissed):

mihnita (commented):
Nitpicking on words.

* "manual" => my understanding (from the email thread) is that this can also run automatically, in some cases. So it is not always "manual"?
* "deploys" vs "deployments"

Thank you!
M.

Mihai

echeran:
> Nitpicking on words.
> 
> * "manual" => my understanding (from the email thread) is that this can also run automatically, in some cases. So it is not always "manual"?
> * "deploys" vs "deployments"
> 
> Thank you! M.
> 
> Mihai

I made a few wording changes, including the "deploy" -> "deployment" change.

I kept "manual" because that is used in a step that is an additional early exit check to provide a more explanatory warning to users of what is & isn't an appropriate usage of artifact deployment in relation to versions. This extra check explicitly addresses the scenario that triggered the whole email thread in the first place. If not for the above reasons, then yes, that particular step is not entirely necessary from a technical perspective because this PR down below in later steps already prevents publishing of `icu4j` for snapshot releases.

.github/workflows/maven.yaml
```diff
@@ -25,11 +28,32 @@ jobs:
...
+        id: mvn-proj-version
+        run: |
+          cd icu4j
+          version="`mvn help:evaluate -Dexpression=project.version -q -DforceStdout`"
+          echo "version=$version"  # debug/info logging for our own purposes, before sending to Github
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+      - name: Assess Maven version release/snapshot
+        id: mvn-proj-version-type
+        run: |
+          if echo "${{ steps.mvn-proj-version.outputs.version }}" | grep -E "\-|SNAPSHOT"
```
markusicu, line 41: Is that pipe symbol on purpose? It's a snapshot if the version contains either a dash or "SNAPSHOT"?

echeran: Yeah. I thought `SNAPSHOT` or `-SNAPSHOT` should be enough, but then I thought about some of the nonstandard version numbers used by other artifacts that still use a `-` separator (ex: [`icu4j-for-cldr`](https://github.com/unicode-org/icu/packages/411079)). Those versions aren't "proper" semantic versions, and that artifact isn't a user-facing artifact. So I felt that they should also be caught as not a release => snapshot.

markusicu: sgtm -- Could you please add a comment that explains this? Maybe *something like*
```
# Set version-type=snapshot for tags that contain SNAPSHOT,
# as well as for unicode-org-internal artifacts with a dash like `icu4j-for-cldr`.
```

echeran: Done.

markusicu (approved):

PR #2973, opened 2024-04-18
ICU-22747 MF2: Validate digit size for :number and :integer; validate options for :datetime

catamorphism:
In MF2, functions can be passed named option values. The spec defines two cases where built-in functions should error out when provided an unexpected value for one of its options:

* For the built-in `:number` and `:integer` functions: "If the value of a digit size option does not evaluate as a non-negative integer, or if the value exceeds any implementation-defined upper limit or any option-specific lower limit, a Bad Option Error is emitted." [1](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#digit-size-options)
* For the built-in `:datetime` function:  "The :datetime function can use either the appropriate style options or can use a collection of field options (but not both) to control the formatted output.

  If both are specified, a Bad Option error MUST be emitted and a fallback value used as the resolved value of the expression." [2](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#options-3)

This PR implements both of these checks for ICU4C.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22747


aphillips:
> Previously, it wasn't an error to write a message like::
>```
> .local $foo = {1 :integer minimumIntegerDigits=foo} {{$foo}}
>```
> This is an error according to the spec, because the minimumIntegerDigits option to :integer must have a value that's a non-negative integer.

It is not an error to write a message like that, from the point of view that this is a "valid" message.

I carefully searched the formatting and registry sections and double-checked errors. It is not currently _required_ to be an error (as an alternative, an implementation might ignore the option, although I'm dubious that this is a Good Idea). The registry part of the spec says what the value space for digit options is, but not what to do if the value provided doesn't match it. (It also fails to clearly state that what to do is implementation defined). It is permitted that implementations emit errors during function resolution (in which case, this will be an Invalid Expression error). When this is addressed (and I think it should be) I think it should be addressed in default registry, not the core spec.

mihnita:
In general the philosophy in how errors are handled in JS / Python and other other can be quite different from something like ICU.
If JS throws at runtime on some website, people complain, I fix my JS, and it's fixed in minutes.

If I do throw in ICU (C++ or Java) on something like a watch, it means a firmware update that has to be explicitly pushed to millions of devices.
So libraries like ICU have to be a lot more tolerant.

If the spec explicitly REQUIRES such mistakes to trigger errors, then I would strongly push to change the spec.

catamorphism:
@aphillips @mihnita When I say "the spec requires it", I mean in the sense that the tests are part of the spec.

I think I raised the question before of whether the test should be required to be an error, and Eemeli's comment [here](https://github.com/unicode-org/message-format-wg/pull/740) said yes, which is what I'm going off. But I'm happy to close the PR (for now) if there needs to be more discussion on the spec side.

aphillips:
@catamorphism Thanks for this.

@eemeli's comment and what the tests do might be the "right thing to do", but, as noted, they are not what the spec says. We need to be very careful about creating conformance tests. They need to match what the spec actually requires. I think it is valid for an implementation (ICU4C is one) to emit an error for this and render the fallback expression. But these are not syntax or data model errors.

catamorphism:
I'm converting this PR to draft since it's an ongoing topic of discussion in the spec working group.

ZL91 (approved):

ZL91 (approved):

catamorphism:
https://github.com/unicode-org/icu/pull/3423 adds some option validation, so I think it makes sense to land that PR first and then see what is still necessary in this PR. I'm making this a draft.

catamorphism:
Obsoleted by https://github.com/unicode-org/icu/pull/3536

PR #2974, opened 2024-04-18
ICU-22724 ICU BRS 76: front-load update version to 76.0.1

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22724


sffc:
From https://unicode-org.github.io/icu/processes/release/tasks/versions.html

Did you do this?

> SINCE ICU 68
> In [tools/cldr/cldr-to-icu/build-icu-data.xml](https://github.com/unicode-org/icu/blob/main/tools/cldr/cldr-to-icu/build-icu-data.xml) edit <property name="icuVersion" value="67.1.0.0"/> and <property name="icuDataVersion" value="67.1.0.0"/>.

And this?

> [icu4j/main/core/src/main/java/com/ibm/icu/util/VersionInfo.java](https://github.com/unicode-org/icu/blob/main/icu4j/main/core/src/main/java/com/ibm/icu/util/VersionInfo.java)
>
> There is a static block starting at line 501 (as of 54.1) in the source file.
>
> Update the ICU_VERSION value, where the first three arguments represent the major, minor, and patch versions of a semantic version.

And these?

> [icu4j/main/shared/build/common.properties](https://github.com/unicode-org/icu/blob/main/icu4j/main/shared/build/common.properties)
>
> icu4j/build.properties (For API change report and release target)

The rest of the changes look correct.

sffc (approved):

sffc:
Verified that the files are in the branch, I just didn't find them the first time amongst the 4000 other files.

mihnita:
Thank you very much for the quick review, 
and for being thorough 
M

PR #2987, opened 2024-04-30
ICU-21757 Remove ICU dependencies on CollectionUtilities

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21757

### DIFF
--- a/icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/AlphabeticIndexTest.java
+++ b/icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/AlphabeticIndexTest.java
@@ -22,7 +22,6 @@ import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 
 import com.ibm.icu.dev.test.TestFmwk;
-import com.ibm.icu.dev.util.CollectionUtilities;
 import com.ibm.icu.impl.ICUDebug;
 import com.ibm.icu.impl.Row;
 import com.ibm.icu.impl.Row.R4;
@@ -384,7 +383,8 @@ public class AlphabeticIndexTest extends TestFmwk {
                         indexCharacters2.addLabels((UnicodeSet)test[i]);
                     }
                 }
-                List<Bucket<Double>> buckets = CollectionUtilities.addAll(alphabeticIndex.iterator(), new ArrayList<Bucket<Double>>());
+                List<Bucket<Double>> buckets = new ArrayList<>(alphabeticIndex.getBucketCount());
+                alphabeticIndex.iterator().forEachRemaining(buckets::add);
                 logln(buckets.toString());
             }
             assertEquals(LabelType.OVERFLOW + "\t" + printList, 1, counter.get(LabelType.OVERFLOW));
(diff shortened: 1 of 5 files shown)


icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/AlphabeticIndexTest.java
```diff
@@ -374,7 +373,7 @@ public void TestInflow() {
             String printList = Arrays.asList(test).toString();
             assertEquals(LabelType.UNDERFLOW + "\t" + printList, 1, counter.get(LabelType.UNDERFLOW));
             assertEquals(LabelType.INFLOW + "\t" + printList, expected, counter.get(LabelType.INFLOW));
-            if (expected != counter.get(LabelType.INFLOW)) {
+            if (expected == counter.get(LabelType.INFLOW)) {
```
markusicu, line 376: I think this wants to remain as is: When the preceding assertion fails, then this code prints stuff "for debugging". It might effectively not work anymore since JUnit assertEquals() throws an exception, but we should at least keep the intent intact.

markusicu (commented):
otherwise lgtm

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/AlphabeticIndexTest.java
mihnita: Sorry, my bad.
I used it to force a printout.
Did it with the old code, new code, and compared that it is the same.
Fixed.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/PluralRulesTest.java
```diff
@@ -1523,7 +1542,8 @@ private void generateLOCALE_SNAPSHOT() {
             System.out.println("\n        // " + set);
             for (PluralRules rule : rules) {
                 Set<ULocale> locales = data.get(rule);
-                System.out.print("        \"" + CollectionUtilities.join(locales, ","));
+                String toShow = locales.stream().map(ULocale::toString).collect(Collectors.joining(","));
```
markusicu, line 1545: Just FYI -- I wonder how many people have to write this kind of work around because the various Java 8 joiner APIs want CharSequences instead of Objects...

markusicu (approved):

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/PluralRulesTest.java
mihnita: I don't know.
I thought about that.
My guess is that if it is too easy then people will pass objects without noticing and will get junk (because many objects don't have a friendly `toString`). In fact even the `Locale.toString()` is not that friendly (`zh_TW_#Hant` :-)

Probably a nice signature would be to take a type the collection on `T`, separators, prefix / suffix (optional) AND a method / lambda that maps from T to `String`.

Should be a few lines of code.
And then you can do something like `String.join(locales, Locale::toLanguageTag, ", ")`

PR #3012, opened 2024-05-16
ICU-22763 MF2: Handle time zones correctly in :datetime

catamorphism:
Previously, the time zone components of date/time literals were ignored. In order to store the time zone properly for re-formatting, this change also changes the representation of `Formattable` date values to use a `GregorianCalendar` rather than a `UDate`. 

This is a public API change and a design doc can be found in the "ICU 76 API proposal status" document (and was also emailed to the icu-design list on April 29.)

In the TC discussion on May 9, there was some uncertainty about whether to use `GregorianCalendar` or `Calendar` as the return type of `message2::Formattable::getDate()`. I'm leaving it as `GregorianCalendar` for now but can change it if necessary.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22763

### DIFF
--- a/icu4c/source/i18n/messageformat2_formattable.cpp
+++ b/icu4c/source/i18n/messageformat2_formattable.cpp
@@ -11,6 +11,8 @@
 
 #include "unicode/messageformat2_formattable.h"
 #include "unicode/smpdtfmt.h"
+#include "messageformat2_allocation.h"
+#include "messageformat2_function_registry_internal.h"
 #include "messageformat2_macros.h"
 
 #include "limits.h"
@@ -37,7 +39,6 @@ namespace message2 {
 
     Formattable::Formattable(const Formattable& other) {
         contents = other.contents;
-        holdsDate = other.holdsDate;
     }
 
     Formattable Formattable::forDecimal(std::string_view number, UErrorCode &status) {
@@ -55,7 +56,7 @@ namespace message2 {
 
     UFormattableType Formattable::getType() const {
         if (std::holds_alternative<double>(contents)) {
-            return holdsDate ? UFMT_DATE : UFMT_DOUBLE;
+            return UFMT_DOUBLE;
         }
         if (std::holds_alternative<int64_t>(contents)) {
             return UFMT_INT64;
@@ -75,6 +76,9 @@ namespace message2 {
                 return UFMT_INT64;
             }
             }
+        }
+        if (isDate()) {
+            return UFMT_DATE;
         }
         if (std::holds_alternative<const FormattableObject*>(contents)) {
             return UFMT_OBJECT;
@@ -225,14 +229,6 @@ namespace message2 {
         return df.orphan();
     }
 
-    void formatDateWithDefaults(const Locale& locale, UDate date, UnicodeString& result, UErrorCode& errorCode) {
-        CHECK_ERROR(errorCode);
-
-        LocalPointer<DateFormat> df(defaultDateTimeInstance(locale, errorCode));
-        CHECK_ERROR(errorCode);
-        df->format(date, result, 0, errorCode);
-    }
-
     // Called when output is required and the contents are an unevaluated `Formattable`;
     // formats the source `Formattable` to a string with defaults, if it can be
     // formatted with a default formatter
@@ -261,9 +257,9 @@ namespace message2 {
         switch (type) {
         case UFMT_DATE: {
             UnicodeString result;
-            UDate d = toFormat.getDate(status);
+            const DateInfo* dateInfo = toFormat.getDate(status);
             U_ASSERT(U_SUCCESS(status));
-            formatDateWithDefaults(locale, d, result, status);
+            formatDateWithDefaults(locale, *dateInfo, result, status);
             return FormattedPlaceholder(input, FormattedValue(std::move(result)));
         }
         case UFMT_DOUBLE: {
--- a/icu4c/source/i18n/messageformat2_function_registry.cpp
+++ b/icu4c/source/i18n/messageformat2_function_registry.cpp
@@ -17,6 +17,7 @@
 #include "unicode/messageformat2_data_model_names.h"
 #include "unicode/messageformat2_function_registry.h"
 #include "unicode/normalizer2.h"
+#include "unicode/simpletz.h"
 #include "unicode/smpdtfmt.h"
 #include "charstr.h"
 #include "double-conversion.h"
@@ -24,7 +25,9 @@
 #include "messageformat2_function_registry_internal.h"
 #include "messageformat2_macros.h"
 #include "hash.h"
+#include "mutex.h"
 #include "number_types.h"
+#include "ucln_in.h"
 #include "uvector.h" // U_ASSERT
 
 // The C99 standard suggested that C++ implementations not define PRId64 etc. constants
@@ -1007,6 +1010,213 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
     return result;
 }
 
+// DateFormat parsers that are shared across threads
+static DateFormat* dateParser = nullptr;
+static DateFormat* dateTimeParser = nullptr;
+static DateFormat* dateTimeUTCParser = nullptr;
+static DateFormat* dateTimeZoneParser = nullptr;
+static icu::UInitOnce gMF2DateParsersInitOnce {};
+
+// Clean up shared DateFormat objects
+static UBool mf2_date_parsers_cleanup() {
+    if (dateParser != nullptr) {
+        delete dateParser;
+        dateParser = nullptr;
+    }
+    if (dateTimeParser != nullptr) {
+        delete dateTimeParser;
+        dateTimeParser = nullptr;
+    }
+    if (dateTimeUTCParser != nullptr) {
+        delete dateTimeUTCParser;
+        dateTimeUTCParser = nullptr;
+    }
+    if (dateTimeZoneParser != nullptr) {
+        delete dateTimeZoneParser;
+        dateTimeZoneParser = nullptr;
+    }
+    return true;
+}
+
+// Initialize DateFormat objects used for parsing date literals
+static void initDateParsersOnce(UErrorCode& errorCode) {
+    U_ASSERT(dateParser == nullptr);
+    U_ASSERT(dateTimeParser == nullptr);
+    U_ASSERT(dateTimeUTCParser == nullptr);
+    U_ASSERT(dateTimeZoneParser == nullptr);
+
+    // Handles ISO 8601 date
+    dateParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd"), errorCode);
+    // Handles ISO 8601 datetime without time zone
+    dateTimeParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:ss"), errorCode);
+    // Handles ISO 8601 datetime with 'Z' to denote UTC
+    dateTimeUTCParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:ssZ"), errorCode);
+    // Handles ISO 8601 datetime with timezone offset; 'zzzz' denotes timezone offset
+    dateTimeZoneParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:sszzzz"), errorCode);
+
+    if (!dateParser || !dateTimeParser || !dateTimeUTCParser || !dateTimeZoneParser) {
+        errorCode = U_MEMORY_ALLOCATION_ERROR;
+        mf2_date_parsers_cleanup();
+        return;
+    }
+    ucln_i18n_registerCleanup(UCLN_I18N_MF2_DATE_PARSERS, mf2_date_parsers_cleanup);
+}
+
+// Lazily initialize DateFormat objects used for parsing date literals
+static void initDateParsers(UErrorCode& errorCode) {
+    CHECK_ERROR(errorCode);
+
+    umtx_initOnce(gMF2DateParsersInitOnce, &initDateParsersOnce, errorCode);
+}
+
+// From https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#date-and-time-operands :
+// "A date/time literal value is a non-empty string consisting of an ISO 8601 date, or
+// an ISO 8601 datetime optionally followed by a timezone offset."
+UDate StandardFunctions::DateTime::tryPatterns(const UnicodeString& sourceStr,
+                                               UErrorCode& errorCode) const {
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    // Handle ISO 8601 datetime (tryTimeZonePatterns() handles the case
+    // where a timezone offset follows)
+    if (sourceStr.length() > 10) {
+        return dateTimeParser->parse(sourceStr, errorCode);
+    }
+    // Handle ISO 8601 date
+    return dateParser->parse(sourceStr, errorCode);
+}
+
+// See comment on tryPatterns() for spec reference
+UDate StandardFunctions::DateTime::tryTimeZonePatterns(const UnicodeString& sourceStr,
+                                                       UErrorCode& errorCode) const {
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    int32_t len = sourceStr.length();
+    if (len > 0 && sourceStr[len] == 'Z') {
+        return dateTimeUTCParser->parse(sourceStr, errorCode);
+    }
+    return dateTimeZoneParser->parse(sourceStr, errorCode);
+}
+
+static TimeZone* createTimeZone(const DateInfo& dateInfo, UErrorCode& errorCode) {
+    NULL_ON_ERROR(errorCode);
+
+    TimeZone* tz;
+    if (dateInfo.zoneId.isEmpty()) {
+        // Floating time value -- use default time zone
+        tz = TimeZone::createDefault();
+    } else {
+        tz = TimeZone::createTimeZone(dateInfo.zoneId);
+    }
+    if (tz == nullptr) {
+        errorCode = U_MEMORY_ALLOCATION_ERROR;
+    }
+    return tz;
+}
+
+// Returns true iff `sourceStr` ends in an offset like +03:30 or -06:00
+// (This function is just used to determine whether to call tryPatterns()
+// or tryTimeZonePatterns(); tryTimeZonePatterns() checks fully that the
+// string matches the expected format)
+static bool hasTzOffset(const UnicodeString& sourceStr) {
+    int32_t len = sourceStr.length();
+
+    if (len <= 6) {
+        return false;
+    }
+    return ((sourceStr[len - 6] == PLUS || sourceStr[len - 6] == HYPHEN)
+            && sourceStr[len - 3] == COLON);
+}
+
+// Note: `calendar` option to :datetime not implemented yet;
+// Gregorian calendar is assumed
+DateInfo StandardFunctions::DateTime::createDateInfoFromString(const UnicodeString& sourceStr,
+                                                               UErrorCode& errorCode) const {
+    if (U_FAILURE(errorCode)) {
+        return {};
+    }
+
+    UDate absoluteDate;
+
+    // Check if the string has a time zone part
+    int32_t indexOfZ = sourceStr.indexOf('Z');
+    int32_t indexOfPlus = sourceStr.lastIndexOf('+');
+    int32_t indexOfMinus = sourceStr.lastIndexOf('-');
+    int32_t indexOfSign = indexOfPlus > -1 ? indexOfPlus : indexOfMinus;
+    bool isTzOffset = hasTzOffset(sourceStr);
+    bool isGMT = indexOfZ > 0;
+    UnicodeString offsetPart;
+    bool hasTimeZone = isTzOffset || isGMT;
+
+    if (!hasTimeZone) {
+        // No time zone; parse the date and time
+        absoluteDate = tryPatterns(sourceStr, errorCode);
+        if (U_FAILURE(errorCode)) {
+            return {};
+        }
+    } else {
+        // Try to split into time zone and non-time-zone parts
+        UnicodeString dateTimePart;
+        if (isGMT) {
+            dateTimePart = sourceStr.tempSubString(0, indexOfZ);
+        } else {
+            dateTimePart = sourceStr.tempSubString(0, indexOfSign);
+        }
+        // Parse the date from the date/time part
+        tryPatterns(dateTimePart, errorCode);
+        // Failure -- can't parse this string
+        if (U_FAILURE(errorCode)) {
+            return {};
+        }
+        // Success -- now parse the time zone part
+        if (isGMT) {
+            dateTimePart += UnicodeString("GMT");
+            absoluteDate = tryTimeZonePatterns(dateTimePart, errorCode);
+            if (U_FAILURE(errorCode)) {
+                return {};
+            }
+        } else {
+            // Try to parse time zone in offset format: [+-]nn:nn
+            absoluteDate = tryTimeZonePatterns(sourceStr, errorCode);
+            if (U_FAILURE(errorCode)) {
+                return {};
+            }
+            offsetPart = sourceStr.tempSubString(indexOfSign, sourceStr.length());
+        }
+    }
+
+    // If the time zone was provided, get its canonical ID,
+    // in order to return it in the DateInfo
+    UnicodeString canonicalID;
+    if (hasTimeZone) {
+        UnicodeString tzID("GMT");
+        if (!isGMT) {
+            tzID += offsetPart;
+        }
+        TimeZone::getCanonicalID(tzID, canonicalID, errorCode);
+        if (U_FAILURE(errorCode)) {
+            return {};
+        }
+    }
+
+    return { absoluteDate, canonicalID };
+}
+
+void formatDateWithDefaults(const Locale& locale,
+                            const DateInfo& dateInfo,
+                            UnicodeString& result,
+                            UErrorCode& errorCode) {
+    CHECK_ERROR(errorCode);
+
+    LocalPointer<DateFormat> df(defaultDateTimeInstance(locale, errorCode));
+    CHECK_ERROR(errorCode);
+
+    df->adoptTimeZone(createTimeZone(dateInfo, errorCode));
+    CHECK_ERROR(errorCode);
+    df->format(dateInfo.date, result, nullptr, errorCode);
+}
+
 FormattedPlaceholder StandardFunctions::DateTime::format(FormattedPlaceholder&& toFormat,
                                                    FunctionOptions&& opts,
                                                    UErrorCode& errorCode) const {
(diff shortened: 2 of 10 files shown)


catamorphism:
cc @echeran @mihnita

markusicu:
@FrankYFTang setting you as the main reviewer because Mihai is ooo for a while soon. Note that the API change proposal is still under discussion.

FrankYFTang:
Here is my view
1. Currently, the Message Format proposal only accept string in ISO8601, without the support of timeZone Extension and Calendar extension as stated in RFC 9557 https://datatracker.ietf.org/doc/html/rfc9557
2. therefore, the only additional information it need to and should be allowed to get back is for the case of the string has UTC offset, and ISO8601 only allow it to be in minutes precision. 
"
5.3.4.1	 Time shift between local time scale and UTC
The time shift between the local time scale and UTC can be expressed in hours and minutes, or hours only
"
3. therefore, the requirement is to allow the caller to get back this information only, but not open up a can of worm that more than that. and we should also allow the interface to distinguish the case that the string has Z (UTC) or no Z.

4. Currently, in ICU, we can get the zone offset from Calendar.get(UCAL_ZONE_OFFSET) and the DST offset from Calendar.get(UCAL_DST_OFFSET) and the UTC offset is the sum of that in millseconds  
So I think we should define a new type for input and return, which has
1. a UDate
2. int32_t rawOffsetGMT - represent the zone offset + dst offset in millisecond
and this timezone (2 above) can be used ot create a TimeZone by

icu::SimpleTimeZone::SimpleTimeZone	(	int32_t 	rawOffsetGMT,
const UnicodeString & 	ID 
)

FrankYFTang:
Maybe something like
```
struct {
  UDate date;
  int32_t rawOffsetGMT;
} DateAndGMTOffset
```
?

aphillips:
Note that MF2 will not stay with only supporting vanilla 8601 strings. Real time zone support is important. An offset is insufficient.

FrankYFTang:
> Note that MF2 will not stay with only supporting vanilla 8601 strings. Real time zone support is important. An offset is insufficient.

that is what I asked this morning and the answer I got back is only vanilla 8601 strings. if MF2 is going to support RFC 9557 https://datatracker.ietf.org/doc/html/rfc9557 then we will need to consider not just timezone extension in RFC9557 but also calendar extension in RFC 9557. Therefore, it will be a super bad idea to use GregorianCalendar since the calendar could be in a different calendar system. 

We should consider the following cases

1. the string is with a Z
2. the string is with a UTC offset 
(both 1 and 2 are just vanilla 8601 strings)
3. the string is with timeZone extension as in RFC 9557
4. the string is with calendar extnesion as in RFC 9557
5. any combination of 1,2,3,4 

For 4, we need a string to indicate the real time zone name but not more than that
For 5, we need a string to indicate the calendar name but not more than tat

how about
```
struct {
  UDate date;
  int32_t rawOffsetGMT;
  const char* zoneName; // "Z" if UTC, nullptr if timeZone offset, other if from RFC9557
  const char* calendar; // nullptr if not specified
} DateAndGMTOffset
```

catamorphism:
> that is what I asked this morning and the answer I got back is only vanilla 8601 strings.

I apologize for giving the wrong answer -- I was answering based on the existing spec. @aphillips is more of an authority than me about how the spec will evolve in the future.

aphillips:
(personal response, chair hat _off_)

The offset is irrelevant. The only thing an offset is useful for is adjusting the date value (the incremental time value--Temporal calls this an `Instant`). Once the incremental time value is computed, you can safely forget the offset. (In most situations, I would argue that the value should be normalized to UTC and the implementation **should** forget the offset) The time zone, however, is used in expression of the date value later. The time zone can also be changed or removed ("floating" the value).

The use of non-Gregorian calendars in date/time serializations is more problematic: there's not a lot of implementation experience, particularly with non-binary multi-era calendars. It's usually better (or at least more common) to use a proleptic Gregorian calendar (i.e. 8601) on the wire (especially for incremental time values!) and convert using calendar rules for processing/display. Perhaps this will evolve and mature further? My experience here is limited.

There also needs to be a concept of a [floating time](https://w3c.github.io/timezone/#floating-times) value (one not tied to specific instants on the timeline). This could use `null` for the time zone. Perhaps:

```
struct {
  UDate date; // what is a `UDate` anyway? Is it millis since the Unix epoch?
  const char* zoneName; // IANA tz name; "UTC" if UTC; nullptr if value is floating
  const char* calendarName; // nullptr if not specified; proleptic Gregorian (8601) calendar is default
} DateInfo
```

In any case, this was a significant gap in the MF2 tech preview. It's important for us to add time zone management to MF2 before LDML46 and the ICU4C implementation should be designed/adapted with that in mind.

catamorphism:
Thanks for the feedback, @aphillips and @FrankYFTang . I'll take a look at these options and update the design doc accordingly.

My first thought, though, is why not something like:

```
struct {
  UDate date;
  SimpleTimeZone tz; // Captures offset and time zone name
  const char* calendarName; // nullptr if not specified
} DateInfo
```

Why not re-use the existing `SimpleTimeZone` class?

catamorphism:
@FrankYFTang I've made changes to match the suggestions; can you take another look? I've also updated the design proposal in the "ICU 76 API proposal status" doc.

icu4c/source/i18n/unicode/messageformat2_formattable.h
```diff
@@ -64,6 +65,50 @@ namespace message2 {
...
+        UnicodeString zoneName;
+        /**
+         * Calendar name. Note: non-Gregorian calendars not yet implemented.
+         * String is empty if calendar not specified. Proleptic Gregorian calendar
+         * is default.
+         *
+         * @internal ICU 76 technology preview
+         * @deprecated This API is for technology preview only.
+         */
+        UnicodeString calendarName;
```
FrankYFTang, line 99: I do not understand why you need createTimeZone() here since all the information you need can be derived from zoneName itself. I will prefer we not adding  more function if it is really needed.

icu4c/source/i18n/unicode/messageformat2_formattable.h
```diff
@@ -64,6 +65,50 @@ namespace message2 {
...
+         * @deprecated This API is for technology preview only.
+         */
+        UDate date;
+        /**
+         * IANA time zone name; "UTC" if UTC; empty string if value is floating
+         *
+         * @internal ICU 76 technology preview
+         * @deprecated This API is for technology preview only.
+         */
+        UnicodeString zoneName;
```
FrankYFTang, line 90: should this be const?

icu4c/source/i18n/unicode/messageformat2_formattable.h
FrankYFTang, line 99: should this be const?

catamorphism: I included it because to call `DateFormat::format()` to eventually format the date, a `TimeZone` object has to be constructed (to pass to `DateFormat`'s `adoptTimeZone()` or `setTimeZone()`) method. But it doesn't have to be a method of the `DateInfo` struct (changed in [1c68d51](https://github.com/unicode-org/icu/pull/3012/commits/1c68d511902932ccccaef4f5eb75019e9f6d7d80)).

icu4c/source/i18n/unicode/messageformat2_formattable.h
catamorphism: `DateInfo` is movable (because it can appear inline in `Formattable`, which is movable), so the members can't be const. I could change `DateInfo` to a class and make the members private, if you think that would be better.

icu4c/source/i18n/unicode/messageformat2_formattable.h
catamorphism: See above

icu4c/source/i18n/unicode/messageformat2_formattable.h
```diff
@@ -64,6 +65,40 @@ namespace message2 {
...
+        /**
+         * Calendar name. Note: non-Gregorian calendars not yet implemented.
+         * String is empty if calendar not specified. Proleptic Gregorian calendar
+         * is default.
+         *
+         * @internal ICU 76 technology preview
+         * @deprecated This API is for technology preview only.
+         */
+        UnicodeString calendarName;
```
srl295, line 99: What is the calendarName supposed to map to? Is it the 'old style name' (`gregorian`) matching https://github.com/unicode-org/icu/blob/cf7ff1b0a508c216c57b61badfa44681cf0978fc/icu4c/source/i18n/unicode/calendar.h#L1297 or a bcp47 `-u-ca` type (`gregory`)? 

If it's from the locale, wouldn't that be visible from the *locale* of the message format itself?

srl295 (commented):
Generally looking GTM, some comments

srl295:
needs squash

icu4c/source/i18n/unicode/messageformat2_formattable.h
aphillips: The spec currently says that it's [this](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier), but it's not fully defined in v45 (the calendar option is reserved for future standardization). Probably we want to support whatever JS Temporal does or at least be compatible with that.

srl295: It should probably link there, as a 'calendar type' in ICU4C is one of the old types generally.

catamorphism: @srl295 Added a comment in [82f9ddc](https://github.com/unicode-org/icu/pull/3012/commits/82f9ddc591347bc3805230dbca8c9cea25bd4191) - let me know if that looks OK.

catamorphism:
@FrankYFTang Are you OK with the current state of [the design doc](https://docs.google.com/document/d/1xERVccTYsptzjfbjcj6HDtoKVF_mEKmslPsOiQzzaFg/edit?pli=1#heading=h.qg9wb813o6jc) for this PR? The JIRA ticket shows as "Accepted", but I've made changes to the design doc since your last comment.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,181 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        d = tryPattern("YYYY-MM-dd'T'HH:mm:ssZ", sourceStr, errorCode);
+    }
+    return d;
+}
+
+static TimeZone* createTimeZone(const DateInfo& dateInfo, UErrorCode& errorCode) {
+    NULL_ON_ERROR(errorCode);
+
+    TimeZone* tz;
+    if (dateInfo.zoneName.length() == 0) {
```
FrankYFTang, line 945: if (dateInfo.zoneName.isEmpty()) {

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,181 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        tz = TimeZone::createTimeZone(dateInfo.zoneName);
+    }
+    if (tz == nullptr) {
+        errorCode = U_MEMORY_ALLOCATION_ERROR;
+    }
+    return tz;
+}
+
+// Parse a string like [+-]nn:nn to a time zone
+SimpleTimeZone* createTimeZonePart(const UnicodeString& offsetStr, UErrorCode& errorCode) {
```
FrankYFTang, line 958: I cannot find where this function is called.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,181 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        return 0;
+    }
+    return dateParser->parse(sourceStr, errorCode);
+}
+
+static UDate tryPatterns(const UnicodeString& sourceStr, UErrorCode& errorCode) {
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    UDate d = tryPattern("YYYY-MM-dd'T'HH:mm:ss", sourceStr, errorCode);
```
FrankYFTang, line 921: instead of trying first one then the next one, could you make decision of which one to call by the length of the sourceStr? Isn't that true for the tryPattern("YYYY-MM-dd'T'HH:mm:ss" ) to return error the length of sourceStr must be in a certain length ? same question with the "YYYY-MM-dd".

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,181 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        d = tryPattern("YYYY-MM-dd", sourceStr, errorCode);
+    }
+    return d;
+}
+
+static UDate tryTimeZonePatterns(const UnicodeString& sourceStr, UErrorCode& errorCode) {
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    UDate d = tryPattern("YYYY-MM-dd'T'HH:mm:sszzzz", sourceStr, errorCode);
```
FrankYFTang, line 933: instead of trying first one then the next one, could you make decision of which one to call based on the length and  last character in sourceStr? Why are we trying theset two different pattern? Where in the spec specify it take both zzzz and Z ? (how about other pattern?)

icu4c/source/i18n/messageformat2_function_registry.cpp
FrankYFTang: also, why are we trying "YYYY-MM-dd'T'HH:mm:sszzzz" ? and "YYYY-MM-dd"? which spec are you based on to try these two patterns? (and these two patterns only?)

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,181 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+                            const DateInfo& dateInfo,
+                            UnicodeString& result,
+                            UErrorCode& errorCode) {
+    CHECK_ERROR(errorCode);
+
+    LocalPointer<DateFormat> df(defaultDateTimeInstance(locale, errorCode));
+    CHECK_ERROR(errorCode);
+
+    // Non-Gregorian calendars not supported yet
+    U_ASSERT(dateInfo.calendarName.length() == 0);
```
FrankYFTang, line 1078: U_ASSERT(dateInfo.calendarName.isEmpty());

icu4c/source/i18n/messageformat2_formattable.cpp
```diff
@@ -259,9 +255,9 @@ namespace message2 {
         switch (type) {
         case UFMT_DATE: {
             UnicodeString result;
-            UDate d = toFormat.getDate(status);
+            const DateInfo* dateInfo = toFormat.getDate(status);
             U_ASSERT(U_SUCCESS(status));
-            formatDateWithDefaults(locale, d, result, status);
+            formatDateWithDefaults(locale, *dateInfo, result, status);
```
FrankYFTang, line 260: dateInfo could be nullptr here, you cannot just deref here. Also, status could be error

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Removed (forgot to remove it while refactoring).

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Good idea, done!

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Done

icu4c/source/i18n/messageformat2_formattable.cpp
catamorphism: There's an assertion right before, and `getDate()` returns a non-null pointer if it succeeds, but I changed the code and added a comment to make this clearer.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,149 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+    return tryPattern("YYYY-MM-dd", sourceStr, errorCode);
+}
+
+// See comment on tryPatterns() for spec reference
+static UDate tryTimeZonePatterns(const UnicodeString& sourceStr, UErrorCode& errorCode) {
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    // 'zzzz' to handle the case where the timezone offset follows the ISO 8601 datetime
+    if (sourceStr.length() > 25) {
```
FrankYFTang, line 939: Could you explain where the 25 came from? How do you conclude it need to > 25? and 25 is enough?

https://unicode.org/reports/tr35/tr35-dates.html#Contents show

zzzz | Pacific Daylight Time | The long specific non-location format. Where that is unavailable, falls back to the long localized GMT format ("OOOO").
-- | -- | --

and

OOOO | GMT-08:00 | The long localized GMT format.
-- | -- | --

catamorphism:
In [197caef](https://github.com/unicode-org/icu/pull/3012/commits/197caefa5926efa5cae1df5b082fc9b92c49be0e), I added some comments explaining the meaning of the fields in the `DateInfo` struct. These can be checked against the code in messageformat2_formattable.cpp that formats an existing `DateInfo` struct, and that parses a date literal into a `DateInfo` -- the two locations where the `DateFormat::adoptTimeZone()` method is called.

catamorphism:
I also removed the `calendarName` field in [9800253](https://github.com/unicode-org/icu/pull/3012/commits/9800253a9d21071b1abf70d7256c82e36430af1a). On further thought, it doesn't seem necessary; a `DateInfo` represents a parsed date literal, and date literal strings don't include a calendar name. In the future, `calendar` will be an option on the `:datetime` formatter. Then it will be handled the same way as all other options, and the return value from the formatter will represent the previously passed-in calendar by default. Therefore, I don't think we need it in the `DateInfo` struct, which is meant to represent the parsed version of a date literal string.

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: I suppose it doesn't really need to check the length at all, just whether the last character is 'Z'. Changed that in [c63c6b5](https://github.com/unicode-org/icu/pull/3012/commits/c63c6b52832c7251b93636ff33ca54d8161b271c).

catamorphism:
This needs to be rebased because of #3050 moving where the test data is, but I'll do that at the very end in order to avoid losing review comments.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,145 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
     return result;
 }
 
+static UDate tryPattern(const char* pat, const UnicodeString& sourceStr, UErrorCode& errorCode) {
+    LocalPointer<DateFormat> dateParser(new SimpleDateFormat(UnicodeString(pat), errorCode));
+    if (U_FAILURE(errorCode)) {
+        return 0;
+    }
+    return dateParser->parse(sourceStr, errorCode);
```
FrankYFTang, line 914: 1. dateParser->parse() is a const method, which mean it could be shared across thread, right?
2. there are only total 4 calls into tryPattern() in this method

tryPattern("YYYY-MM-dd'T'HH:mm:ss"
tryPattern("YYYY-MM-dd"
tryPattern("YYYY-MM-dd'T'HH:mm:ssZ"
tryPattern("YYYY-MM-dd'T'HH:mm:sszzzz"

So there are total 4 possible pat , right?
Currently, the creation of the DateFormat is inside the tryPattern()
Everytime the tryPattern() got called, the code need to create a SimpleDateFormat and later destroy it
but there are only total 4 possible SimpleDateFormat needed to be created and the usage are thread safe (see my point 1 above)
So... should we have an lazy initialization routine which create these four SimpleDateFormat and cache it globally, and reuse them later. The creation need to be protected to be init once and therefore mutex to prevent different threads create them concurrently, but then we can reuse them across thread and destroy upon lib clean up. 

In this way, we do not need to repeatly create and destroy the SimpleDateFormat

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +906,145 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+    return tz;
+}
+
+static bool hasTzOffset(const UnicodeString& sourceStr) {
+    int32_t len = sourceStr.length();
+
+    if (len <= 6) {
+        return false;
+    }
+    return ((sourceStr[len - 6] == PLUS || sourceStr[len - 6] == HYPHEN)
```
FrankYFTang, line 969: Please add a comment in the beginning of this function to explain what is the expected input value and the output

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Yes, that seems reasonable. I'll try it.

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Done in [4a7da29](https://github.com/unicode-org/icu/pull/3012/commits/4a7da29c58c703ab5fe6a08c3357f3da98559dd4)

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: OK, done in [8d4da42](https://github.com/unicode-org/icu/pull/3012/commits/8d4da423992d90520b3e52043f8f317dbb2c3993). I'm not sure of how to unit-test this (i.e. testing that the DateFormats really are initialized only once), but I'll look for examples in the test suite.

catamorphism:
By the way, I'll be on vacation from now until September 9, so I won't see further comments until I return.

icu4c/source/i18n/messageformat2_function_registry_internal.h
```diff
@@ -57,6 +57,15 @@ namespace message2 {
 
             DateTimeType type;
             DateTimeFactory(DateTimeType t) : type(t) {}
+
```
FrankYFTang, line 60: so.. you store 4 pointer per DateTimeFactory object. But in messageformat2_function_registry.cpp you create multiple DateTimeFactory object, each object will crate the same four DateFormat. But there are no differences between them from the multiple DateTimeFactory . Could you use InitOnce and cleanup code tostore that four DateFormat to be globally shared ?

catamorphism: Done in [da9307c](https://github.com/unicode-org/icu/pull/3012/commits/da9307cc1b85da5ce0a39e98f7f5ea468f624216)

After implementing that change and running under asan, I noticed a use-after-free error. This required me to change some code in tznames.cpp that dealt with reference counting on `TimeZoneNamesDelegate` objects. The `deleteTimeZoneNamesCacheEntry()` function in that file was assuming there are no other references to the cache entry. But the new global `DateFormat` object holds a reference to a cache entry, and the new cleanup function I added was getting called after the `timeZoneNames_cleanup()` function and hence after the cache was already deleted. So I had to change both `deleteTimeZoneNamesCacheEntry()` and `~TimeZoneNamesDelegate()` to check the ref counts. I made this a separate patch (for ease of review), [1b8e43f](https://github.com/unicode-org/icu/pull/3012/commits/1b8e43f91e681f1dabbbdb5a2ca1e294959523df).

catamorphism:
@FrankYFTang ping?

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +908,195 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+
+    // Handles ISO 8601 date
+    dateParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd"), errorCode);
+    // Handles ISO 8601 datetime without time zone
+    dateTimeParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:ss"), errorCode);
+    // Handles ISO 8601 datetime with 'Z' to denote UTC
+    dateTimeUTCParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:ssZ"), errorCode);
+    // Handles ISO 8601 datetime with timezone offset; 'zzzz' denotes timezone offset
+    dateTimeZoneParser = new SimpleDateFormat(UnicodeString("YYYY-MM-dd'T'HH:mm:sszzzz"), errorCode);
+    ucln_i18n_registerCleanup(UCLN_I18N_MF2_DATE_PARSERS, mf2_date_parsers_cleanup);
```
FrankYFTang, line 954: what will happen when we have not enough memory and any of the new cause out of memory and return nullptr? Should we check these four pointers against nullptr and set the out of memory error code to errorCode?

catamorphism: Fixed in [0879820](https://github.com/unicode-org/icu/pull/3012/commits/087982055246942fccab09adbe8a15e3676ad51a)

catamorphism:
@FrankYFTang Ping?

FrankYFTang (approved):
LGTM. Thanks for working on this

FrankYFTang (changes requested):
sorry. please address the issues

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Fixed in [dca746a](https://github.com/unicode-org/icu/pull/3012/commits/dca746ad0842f8398f12baaadd32bed39b4220b8)

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +908,195 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        // Parse the date from the date/time part
+        tryPatterns(dateTimePart, errorCode);
+        // Failure -- can't parse this string
+        if (U_FAILURE(errorCode)) {
+            return {};
+        }
+        // Success -- now parse the time zone part
+        if (isGMT) {
+            dateTimePart += UnicodeString("GMT");
+            absoluteDate = tryTimeZonePatterns(dateTimePart, errorCode);
```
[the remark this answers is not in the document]
catamorphism: Fixed in [5e5866b](https://github.com/unicode-org/icu/pull/3012/commits/5e5866ba21259953733067af10aa53233a51c19b)

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +908,195 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+        } else {
+            // Try to parse time zone in offset format: [+-]nn:nn
+            absoluteDate = tryTimeZonePatterns(sourceStr, errorCode);
+            offsetPart = sourceStr.tempSubString(indexOfSign, sourceStr.length());
```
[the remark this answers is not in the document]
catamorphism: Fixed in [5e5866b](https://github.com/unicode-org/icu/pull/3012/commits/5e5866ba21259953733067af10aa53233a51c19b)

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -905,6 +908,195 @@ Formatter* StandardFunctions::DateTimeFactory::createFormatter(const Locale& loc
...
+
+    // If the time zone was provided, get its canonical ID,
+    // in order to return it in the DateInfo
+    UnicodeString canonicalID;
+    if (hasTimeZone) {
+        UnicodeString tzID("GMT");
+        if (!isGMT) {
+            tzID += offsetPart;
+        }
+        TimeZone::getCanonicalID(tzID, canonicalID, errorCode);
```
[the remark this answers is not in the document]
catamorphism: Fixed in [5e5866b](https://github.com/unicode-org/icu/pull/3012/commits/5e5866ba21259953733067af10aa53233a51c19b)

catamorphism:
@FrankYFTang I've fixed the remaining issues and rebased.

FrankYFTang (approved):
Please make sure ICU TC agree w/ the API change inside i18n/unicode directory before you land thi scode

catamorphism:
> Please make sure ICU TC agree w/ the API change inside i18n/unicode directory before you land thi scode

Thanks! The change was already approved in a previous meeting (~~2024-08-08 if I'm not mistaken~~ 2024-09-12 according to the ICU 76 API proposal status doc; and the JIRA issue is Accepted rather than Design.)

PR #3017, opened 2024-05-30
ICU-22642 Avoid spending too much time inside CanonicalIterator

FrankYFTang:
Fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65247 

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22642

### DIFF
--- a/icu4c/source/common/caniter.cpp
+++ b/icu4c/source/common/caniter.cpp
@@ -363,6 +363,9 @@ UnicodeString* CanonicalIterator::getEquivalents(const UnicodeString &segment, i
     char16_t USeg[256];
     int32_t segLen = segment.extract(USeg, 256, status);
     getEquivalents2(&basic, USeg, segLen, status);
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
 
     // now get all the permutations
     // add only the ones that are canonically equivalent
@@ -466,6 +469,9 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
             Hashtable remainder(status);
             remainder.setValueDeleter(uprv_deleteUObject);
             if (extract(&remainder, cp2, segment, segLen, i, status) == nullptr) {
+                if (U_FAILURE(status)) {
+                    return nullptr;
+                }
                 continue;
             }
 
@@ -489,6 +495,13 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
                 //if (PROGRESS) printf("Adding: %s\n", UToS(Tr(*toAdd)));
 
                 ne = remainder.nextElement(el);
+            }
+            // ICU-22642 Guards against strings that have so many permutations
+            // that they would otherwise hang the function.
+            constexpr int32_t kResultLimit = 4096;
+            if (fillinResult->count() > kResultLimit) {
+                status = U_UNSUPPORTED_ERROR;
+                return nullptr;
             }
         }
     }
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/CanonicalIterator.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/CanonicalIterator.java
@@ -172,7 +172,7 @@ public final class CanonicalIterator {
         permute(source, skipZeros, output, 0);
     }
 
-    private static int PERMUTE_DEPTH_LIMIT = 8;
+    private static final int PERMUTE_DEPTH_LIMIT = 8;
     /**
      * Simple implementation of permutation.
      * <br><b>Warning: The strings are not guaranteed to be in any particular order.</b>
@@ -308,6 +308,7 @@ public final class CanonicalIterator {
     }
 
 
+    static private final int RESULT_LIMIT = 4096;
     private Set<String> getEquivalents2(String segment) {
 
         Set<String> result = new HashSet<String>();
@@ -341,6 +342,12 @@ public final class CanonicalIterator {
                 for (String item : remainder) {
                     result.add(prefix + item);
                 }
+            }
+
+            // ICU-22642 Guards against strings that have so many permutations
+            // that they would otherwise hang the function.
+            if (result.size() > RESULT_LIMIT) {
+                throw new UnsupportedOperationException("Too many permutations");
             }
         }
         return result;


richgillam (dismissed):
Seems reasonable to me.  I assume you're depending on Markus to help with a test case?

FrankYFTang:
> Seems reasonable to me. I assume you're depending on Markus to help with a test case?

The test case is inside https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=65247

FrankYFTang:
I added a test to show the problem, without the fix, the constructor will run for a very long time. (several minutes)

icu4c/source/common/caniter.cpp
```diff
@@ -490,6 +493,15 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
 
                 ne = remainder.nextElement(el);
             }
+            // ICU-22642 To avoid security attack by an useless string which
+            // have too many permutations to hang the function, limit the
```
roubert, line 497: I assume you've mistyped here, this guards against strings that have so many permutations that they would otherwise hang the function and there's no such thing as having too many permutations to hang the function, right?

icu4c/source/common/caniter.cpp
```diff
@@ -490,6 +493,15 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
...
+            // return size and return error with U_UNSUPPORTED_ERROR if exceed
+            // the limit.
+            constexpr int32_t kResultLimit = 4096;
+            if (fillinResult->count() > kResultLimit) {
+                status = U_UNSUPPORTED_ERROR;
```
roubert, line 502: Don't we normally return `U_ILLEGAL_ARGUMENT_ERROR` in cases like this, or is there somthing about this kind of input that would make that status inappropriate here?

icu4c/source/test/intltest/canittst.cpp
```diff
@@ -332,4 +333,15 @@ void CanonicalIteratorTest::TestAPI() {
   }
 }
 
+void CanonicalIteratorTest::TestICU22642() {
```
roubert, line 336: Without the bugfix this test case will just continue running until it eventually exhausts all resources of the machine it's running on. That's not useful as a unit test.

roubert (changes requested):

icu4c/source/test/intltest/canittst.cpp
FrankYFTang: but we want to ensure there are no regression if someone change the code later and reintroudce the problem.

icu4c/source/common/caniter.cpp
FrankYFTang: we already use when fix 
https://github.com/unicode-org/icu/commit/757d2cd90a434b35596d04fc89421dd0add33952

icu4c/source/test/intltest/canittst.cpp
roubert: Yes, it's valuable to have a unit test, but the test you have written here now isn't a useful unit test. A useful unit test is something that fails with an error message when some unexpected change causes the thing being tested to no longer behave as expected, not something that continues running until all resources are exhausted and then crashes.

If you can't think of a way of writing a useful unit test for this, we'll simply have to rely on the fuzz test that found the problem this time to find the problem again if it reappears in the future.

FrankYFTang: The broken behavior is it hang, how other way we can test it won't hang?

richgillam: I see both of your points.  I worry that if somebody checks in a change that causes this problem to happen again, what'll happen is that none of the CI checks will finish.  It'll be a while before we know they're stuck, and we won't know what to do for a while and will probably think something's wrong with the CI stuff.

What you might be able to do in this test is spawn two threads-- one that triggers the problem and another one that waits some prescribed amount of time for the first one to finish and that kills it and reports an error if it doesn't.

roubert: > What you might be able to do in this test is spawn two threads-- one that triggers the problem and another one that waits some prescribed amount of time for the first one to finish and that kills it and reports an error if it doesn't.

I don't think this is a very good idea, for two reasons:

1. Introducing threading increases the complexity of the testing code significantly, introducing new potential sources of failures unrelated to the code actually being tested and also making it harder to run this in a debugger to see what's going on when something fails.
2. The speed of execution varies wildly between different execution environments and it'd be nigh impossible to pick a reasonable amount of time that isn't either far too long when run locally on a fast workstation or far too short when run on our slow CI pipelines.

FrankYFTang: >and that kills it and reports an error if it doesn't.
My code currently does not "kills it"

@richgillam  Would it be ok for me to listen to @roubert by just removing the test and let the fuzzer do it's job to verify?

richgillam: Yeah, I'm fine with that.

richgillam (approved):

icu4c/source/test/intltest/canittst.cpp
FrankYFTang: ok, I remove the test per @roubert . PTAL

icu4c/source/common/caniter.cpp
```diff
@@ -466,6 +466,9 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
             Hashtable remainder(status);
             remainder.setValueDeleter(uprv_deleteUObject);
             if (extract(&remainder, cp2, segment, segLen, i, status) == nullptr) {
+                if (U_FAILURE(status)) {
```
roubert, line 469: Wouldn't it be a good idea to also add one of these to the call to `getEquivalents2()` in `getEquivalents()` so that it aborts immediately upon error there too?

icu4c/source/common/caniter.cpp
```diff
@@ -490,6 +493,13 @@ Hashtable *CanonicalIterator::getEquivalents2(Hashtable *fillinResult, const cha
 
                 ne = remainder.nextElement(el);
             }
+            // ICU-22642 Guards against strings that have too many permutations
```
roubert, line 496: Either:

> Guards against strings that have so many permutations that they would otherwise hang the function.


Or:

> Guards against strings that have too many permutations and would otherwise hang the function.

icu4j/main/core/src/main/java/com/ibm/icu/text/CanonicalIterator.java
```diff
@@ -308,6 +308,7 @@ private String[] getEquivalents(String segment) {
     }
 
 
+    static private int PERMUTE_LIMIT = 4096;
```
roubert, line 311: Wouldn't it be more consistent to name this `RESULT_LIMIT` instead?

roubert, line 311: Shouldn't this (and `PERMUTE_DEPTH_LIMIT` too) be `final`?

roubert (changes requested):

FrankYFTang:
PTAL

richgillam (approved):
Still LGTM.

roubert (approved):
LGTM

PR #2926, opened 2024-03-26
ICU-21812 move Age/Block/Script to a separate trie

markusicu:
Move Age/Block/Script to a separate trie; because
- The properties vectors in uprops.icu/uprops.h are full, and making them longer is expensive.
- We are about to overflow the bits for the Age property.
- These properties seem correlated fairly well with each other but not with other properties.

Also
- Move some bit-set-setting code from the emojipropsbuilder to toolutil so that it can be shared.
- Some code cleanup, such as in the corepropsbuilder distinguish end vs. pvecEnd.
- Making a uprops.icu major-version change allows us to move the Script+Script_Extensions bits back together into a contiguous bit set.

Problem: Pulling these three properties out makes uprops.icu significantly larger. Based on Unicode 15.1 data:

Original:
```
trie size in bytes:                    46328
size in bytes of additional props trie:65584
number of additional props vectors:     2499
number of 32-bit words per vector:         3
number of 16-bit scriptExtensions:       298
data size:                            142560
```
(The data size includes 64 bytes for the file header and the indexes array.)

Pulling Age/Block/Script into a separate trie (fast, 32-bit code point trie):
```
trie size in bytes:                    46328
size in bytes of additional props trie:54760
number of additional props vectors:      673
number of 32-bit words per vector:         3
size in bytes of ABS trie:             86748
number of 16-bit scriptExtensions:       298
data size:                            196572
```

Pulling them out into separate tries:
- Age: fast 8-bit
- Block: small 16-bit indexed by code point/16
- Script/scx: fast 16-bit
```
trie size in bytes:                    46328
size in bytes of additional props trie:54760
number of additional props vectors:      673
number of 32-bit words per vector:         3
size in bytes of Age trie:             21596
size in bytes of Block trie:            7600
size in bytes of Script trie:          34420
number of 16-bit scriptExtensions:       298
data size:                            173440
```

Same, but small tries for Age & Script:
```
trie size in bytes:                    46328
size in bytes of additional props trie:54760
number of additional props vectors:      673
number of 32-bit words per vector:         3
size in bytes of Age trie:             17128
size in bytes of Block trie:            7600
size in bytes of Script trie:          25984
number of 16-bit scriptExtensions:       298
data size:                            160536
```

Pulling only Block out into a separate trie:
```
trie size in bytes:                    46328
size in bytes of additional props trie:62752
number of additional props vectors:     2026
number of 32-bit words per vector:         3
size in bytes of Block trie:            7600
[...]

PR #3022, opened 2024-06-02
ICU-22785 move Block bits from propsvec0 to new trie

markusicu:
The uprops.icu properties vectors (3*32 bits for various properties) are full. Making them longer would significantly increase the data size. I tried moving Age+Block+Script out (https://github.com/unicode-org/icu/pull/2926) but several variations also increased the data size.

The one thing that looks like it will slightly decrease the total data size is to move the Block property into its own, “small” code point trie. This is useful, because it frees up 9 properties vector bits for other properties that we want to add, such as Indic_Conjunct_Break (ICU-22503).

Data size before:
```
* uprops.icu stats *
trie size in bytes:                    47608
size in bytes of additional props trie:67392
number of additional props vectors:     2582
number of 32-bit words per vector:         3
number of 16-bit scriptExtensions:       314
data size:                            146676
```

Data size after:
```
* uprops.icu stats *
trie size in bytes:                    47608
size in bytes of additional props trie:64624
number of additional props vectors:     2095
number of 32-bit words per vector:         3
number of 16-bit scriptExtensions:       314
size in bytes of Block trie:            7752
data size:                            145816
```

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22785


echeran (approved):
LGTM

PR #3013, opened 2024-05-20
ICU-21633 Fix ClangCL cross-compilation on Windows

StefanStojanovic:
This PR is a WIP to fix issues with cross-compilation. The current approach is focused on fixing an issue I encountered on Windows with ClangCL (MSVC had no problems).

A quick backstory to better understand my background and motivation. I'm working on supporting Node.js on Windows. There is an ongoing effort to enable Node.js to be compiled with ClangCL. While working on it I noticed an issue when creating `.obj` file in ICU because of the `*pCPU = IMAGE_FILE_MACHINE_UNKNOWN;` line, and the code I made fixes that by adding a new option. After some discussion with @srl295, I was made aware of this being a broader problem that should be fixed for multiple compilers and platforms, not only ClangCL/Windows.

The code I have is a starting point and should be built upon. Please keep in mind that my knowledge and experience with other platforms is limited and I'd appreciate any help I can get. I'm also very new to the ICU, so feel free to correct any mistakes I may start making as soon as you see them.

Refs: https://github.com/nodejs/node/pull/53003

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21633
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #3023, opened 2024-06-03
ICU-22787 Fix ClangCL compilation on Windows

StefanStojanovic:
I'm working on supporting Node.js on Windows. There is an ongoing effort to enable it to be compiled with ClangCL. While working on it I noticed an issue when creating `.obj` file in ICU because of the `*pCPU = IMAGE_FILE_MACHINE_UNKNOWN;` line, and the code I made fixes that by adding a new option. Although this is a broader problem that should be fixed for multiple compilers and platforms, my change is aimed toward ClangCL and Windows, but can be used as a good foundation to work on fixes for other platforms and compilers and make corrections in future iterations.

This issue is currently our main blocker for enabling ClangCL compilation for Node.js and it'd be greatly appreciated if it could be reviewed soon.

Refs: https://github.com/nodejs/node/pull/53003

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22787

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -235,7 +235,7 @@
 /**
  * \def U_PLATFORM_USES_ONLY_WIN32_API
  * Defines whether the platform uses only the Win32 API.
- * Set to 1 for Windows/MSVC and MinGW but not Cygwin.
+ * Set to 1 for Windows/MSVC, ClangCL and MinGW but not Cygwin.
  * @internal
  */
 #ifdef U_PLATFORM_USES_ONLY_WIN32_API
@@ -250,7 +250,7 @@
 /**
  * \def U_PLATFORM_HAS_WIN32_API
  * Defines whether the Win32 API is available on the platform.
- * Set to 1 for Windows/MSVC, MinGW and Cygwin.
+ * Set to 1 for Windows/MSVC, ClangCL, MinGW and Cygwin.
  * @internal
  */
 #ifdef U_PLATFORM_HAS_WIN32_API
--- a/icu4c/source/tools/genccode/genccode.c
+++ b/icu4c/source/tools/genccode/genccode.c
@@ -70,6 +70,7 @@ enum {
 #ifdef CAN_GENERATE_OBJECTS
   kOptObject,
   kOptMatchArch,
+  kOptCpuArch,
   kOptSkipDllExport,
 #endif
   kOptFilename,
@@ -86,6 +87,7 @@ static UOption options[]={
 #ifdef CAN_GENERATE_OBJECTS
 /*6*/UOPTION_DEF("object", 'o', UOPT_NO_ARG),
      UOPTION_DEF("match-arch", 'm', UOPT_REQUIRES_ARG),
+     UOPTION_DEF("cpu-arch", 'c', UOPT_REQUIRES_ARG),
      UOPTION_DEF("skip-dll-export", '\0', UOPT_NO_ARG),
 #endif
      UOPTION_DEF("filename", 'f', UOPT_REQUIRES_ARG),
@@ -131,6 +133,8 @@ main(int argc, char* argv[]) {
             "\t-o or --object      write a .obj file instead of .c\n"
             "\t-m or --match-arch file.o  match the architecture (CPU, 32/64 bits) of the specified .o\n"
             "\t                    ELF format defaults to i386. Windows defaults to the native platform.\n"
+            "\t-c or --cpu-arch    Specify a CPU architecture for which to write a .obj file for ClangCL on Windows\n"
+            "\t                    Valid values for this opton are x64, x86 and arm64.\n"
             "\t--skip-dll-export   Don't export the ICU data entry point symbol (for use when statically linking)\n");
 #endif
         fprintf(stderr,
@@ -193,9 +197,17 @@ main(int argc, char* argv[]) {
                 break;
 #ifdef CAN_GENERATE_OBJECTS
             case CALL_WRITEOBJECT:
+                if(options[kOptCpuArch].doesOccur) {
+                    if (!checkCpuArchitecture(options[kOptCpuArch].value)) {
+                        fprintf(stderr,
+                            "CPU architecture \"%s\" is unknown.\n", options[kOptCpuArch].value);
+                        return -1;
+                    }
+                }
                 writeObjectCode(filename, options[kOptDestDir].value,
                                 options[kOptEntryPoint].doesOccur ? options[kOptEntryPoint].value : NULL,
                                 options[kOptMatchArch].doesOccur ? options[kOptMatchArch].value : NULL,
+                                options[kOptCpuArch].doesOccur ? options[kOptCpuArch].value : NULL,
                                 options[kOptFilename].doesOccur ? options[kOptFilename].value : NULL,
                                 NULL,
                                 0,
--- a/icu4c/source/tools/toolutil/pkg_genc.cpp
+++ b/icu4c/source/tools/toolutil/pkg_genc.cpp
@@ -16,6 +16,9 @@
 #   define NOMCX
 #include <windows.h>
 #include <time.h>
+#   if defined(__clang__)
+#       include <exception>
+#   endif
 #   ifdef __GNUC__
 #       define WINDOWS_WITH_GNUC
 #   endif
@@ -292,6 +295,11 @@ checkAssemblyHeaderName(const char* optAssembly) {
     }
 
     return false;
+}
+
+U_CAPI UBool U_EXPORT2
+checkCpuArchitecture(const char* optCpuArch) {
+    return strcmp(optCpuArch, "x64") == 0 || strcmp(optCpuArch, "x86") == 0 || strcmp(optCpuArch, "arm64") == 0;
 }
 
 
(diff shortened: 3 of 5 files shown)


srl295:
#3013

icu4c/source/tools/toolutil/pkg_genc.cpp
```diff
@@ -847,7 +852,24 @@ getArchitecture(uint16_t *pCPU, uint16_t *pBits, UBool *pIsBigEndian, const char
...
+        // this would potentially be problematic when cross-compiling as this code
+        // would most likely be ran on host machine to generate the .obj file for
+        // the target architecture.
+#       if defined(__clang__)
+            if (strcmp(optCpuArch, "x64") == 0) {
+                *pCPU = IMAGE_FILE_MACHINE_AMD64;
+            } else if (strcmp(optCpuArch, "arm64") == 0) {
+                *pCPU = IMAGE_FILE_MACHINE_ARM64;
+            } else {
+                // This should never happen.
```
roubert, line 867: In that case, I'd say that it'd be easier to debug if the process terminated with an error message if that which should never happen ever happened anyway.

icu4c/source/tools/genccode/genccode.c
```diff
@@ -131,6 +133,7 @@ main(int argc, char* argv[]) {
             "\t-o or --object      write a .obj file instead of .c\n"
             "\t-m or --match-arch file.o  match the architecture (CPU, 32/64 bits) of the specified .o\n"
             "\t                    ELF format defaults to i386. Windows defaults to the native platform.\n"
+            "\t-c or --cpu-arch    Specify a CPU architecture for which to write a .obj file for ClangCL on Windows\n"
```
roubert, line 136: Wouldn't it be easier to use this tool correctly if this commandline flag was `#ifdef`'ed out altogether on platforms where it isn't implemented?

icu4c/source/tools/toolutil/pkg_genc.cpp
```diff
@@ -847,7 +852,24 @@ getArchitecture(uint16_t *pCPU, uint16_t *pBits, UBool *pIsBigEndian, const char
 #   if defined(_M_IX86)
         *pCPU = IMAGE_FILE_MACHINE_I386;
 #   else
-        *pCPU = IMAGE_FILE_MACHINE_UNKNOWN;
+        // Linker for ClangCL doesn't handle IMAGE_FILE_MACHINE_UNKNOWN the same as
```
roubert, line 855: If I read the existing `U_PLATFORM_HAS_WIN32_API` block correctly it's all about Microsoft's `link.exe`, so it seems odd to now mix in another linker in that. Wouldn't this become easier to read if Microsoft's and LLVM's linkers got each their separate `#if` block?

icu4c/source/tools/toolutil/pkg_genc.cpp
```diff
@@ -847,7 +852,24 @@ getArchitecture(uint16_t *pCPU, uint16_t *pBits, UBool *pIsBigEndian, const char
...
+        // linker for MSVC. Because of this optCpuArch is used to define the CPU
+        // architecture in that case. While _M_AMD64 and _M_ARM64 could be used,
+        // this would potentially be problematic when cross-compiling as this code
+        // would most likely be ran on host machine to generate the .obj file for
+        // the target architecture.
+#       if defined(__clang__)
+            if (strcmp(optCpuArch, "x64") == 0) {
+                *pCPU = IMAGE_FILE_MACHINE_AMD64;
```
roubert, line 863: Are 32-bit builds not supported? It's been a long time since I last had any reason to do one, but I wasn't aware that they're no longer supported.

roubert (changes requested):

icu4c/source/tools/toolutil/pkg_genc.cpp
StefanStojanovic: Makes sense. I've changed it.

icu4c/source/tools/genccode/genccode.c
StefanStojanovic: For now, I've only `#ifdef`ed out the parts in `genccode.c`, as I didn't want to make `writeObjectCode` signature different depending on which compiler is used on Windows. I just pass it `NULL` if Clang isn't used. If you think I should make the signature `#ifdef`ed too, let me know.

icu4c/source/tools/toolutil/pkg_genc.cpp
StefanStojanovic: In `platform.h` for `U_PLATFORM_HAS_WIN32_API` it says: `Defines whether the Win32 API is available on the platform.`. This is true for ClangCL, so I've added a new constant `U_CLANG_CL` to distinguish when using Clang on Windows. After that, I changed to using that constant inside the mentioned block.

> Wouldn't this become easier to read if Microsoft's and LLVM's linkers got each their separate #if block

It would, but I do not know how to write linker-specific code (I'm not saying it is impossible). What can be done from my perspective is to detect the compiler (MSVC or Clang) and then infer a linker (link.exe or lld-link.exe), which is more or less what we have now after these changes.

icu4c/source/tools/toolutil/pkg_genc.cpp
StefanStojanovic: Makes sense. I've added a `strcmp(optCpuArch, "x86") == 0` branch in case someone needs it since it is still supported.

Above this code there is:

```
#   if defined(_M_IX86)
        *pCPU = IMAGE_FILE_MACHINE_I386;
```

but that would only apply if `genccode` was compiled targeting x86, which is not necessarily the case.

StefanStojanovic:
Thanks for the review @roubert, I tried to address all your comments in a new commit.

I see 2 checks failed: `jira-ticket` and `single-commit`. I'll squash commits and keep only the original commit message once all is ready to land, but I kept 2 commits now for easier reviewing.

markusicu:
> I see 2 checks failed: `jira-ticket` and `single-commit`. I'll squash commits and keep only the original commit message once all is ready to land, but I kept 2 commits now for easier reviewing.

Please wait with squashing until after feedback is resolved.

StefanStojanovic:
> Please wait with squashing until after feedback is resolved.

Will do.

By the way, I see that in `CI-ICU4C` all `MSVC` builds failed. I'm not sure why that happened, but I built it locally before pushing (`allinone.sln` -> Build Solution) and it worked.

icu4c/source/tools/genccode/genccode.c
srl295: I think it would be better to leave it a generic argument, even if it is only currently valid for ClangCL.  It may be needed in other situations.

StefanStojanovic:
Hey everyone, I wanted to check if there is anything else for me to do after my initial changes? If you think it is better to leave the new option I can every my latest changes in `genccode.c`. Other than that, are the other changes good? Regards.

roubert:
> I see that in `CI-ICU4C` all `MSVC` builds failed. I'm not sure why that happened,

The CI is currently broken, you'll just have to wait for it to get fixed (and then those tests should pass again).

roubert:
> The CI is currently broken,

It has now been fixed and the tests pass.

roubert (changes requested):

icu4c/source/common/unicode/platform.h
```diff
@@ -217,6 +217,16 @@
...
 #endif
 
+/**
+ * \def U_CLANG_CL
+ * Defined if the compiler is Clang compatible with MSVC (Clang on Windows).
+ * Otherwise undefined.
+ * @internal
+ */
+#if (defined(_MSC_VER) && defined(__clang__) && __clang__)
+#   define U_CLANG_CL
```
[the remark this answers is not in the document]
StefanStojanovic: I agree, I'll remove it.

icu4c/source/tools/toolutil/pkg_genc.cpp
StefanStojanovic: I added it based on your previous comment:

> In that case, I'd say that it'd be easier to debug if the process terminated with an error message if that which should never happen ever happened anyway.

Although I didn't think about it before you mentioned it, this log makes sense to me. For example, imagine someone makes a typo and passes `-c X64` or `-c amr64`, this log would help them fix it much quicker.

roubert: > For example, imagine someone makes a typo and passes `-c X64` or `-c amr64`, this log would help them fix it much quicker.

Wait a second, your original comment here was _"This should never happen."_ but the user making a typo on the commandline is the total opposite of that, that's something that certainly is going to happen, sooner or later.

But verifying that the commandline arguments are correct should be done already in `main()` in `genccode.c`, in the same way that the arguments to `-a` are being verified there by calling the `checkAssemblyHeaderName()` helper function.

roubert (changes requested):

icu4c/source/tools/toolutil/pkg_genc.cpp
StefanStojanovic: I understand the confusion. My comment was misleading. Sorry for that. I checked the code now and saw the `checkAssemblyHeaderName` helper, I'll add that type of check to the `-c` option as well. Thanks for pointing that out. Once I add that the else branch should really _**never happen**_, and then the change you suggested with `std::terminate()` will be correct.

StefanStojanovic:
@roubert I pushed the changes in a new commit. I addressed your comments and made changes based on [this comment](https://github.com/unicode-org/icu/pull/3023#discussion_r1631634392). Please let me know if there is something else I should change. Regards.

roubert (commented):
OK, this all looks reasonable to me now, so I'll leave it to @rp9-next (@microsoft) to take this review from here (and verify that what this does really is what it should be doing, and that this is the preferred way of doing that, and so on).

rp9-next (approved):

rp9-next:
@roubert , merging is blocked by a change requested check which I believe is outdated.. 
Could you dismiss it ?

@StefanStojanovic, please squash all the commits to a single commit for merging (you could use the[ squash branch tool](https://us-central1-dev-infra-273822.cloudfunctions.net/unicode-github-bot/info/unicode-org/icu/3023) as well)

PR #3025, opened 2024-06-04
ICU-22785 uprops.icu: coalesce scx+sc bits

markusicu:
More uprops.icu property vectors cleanup.
- Taking advantage of the uprops.icu major format version change to move the Script_Extensions and Script bit fields so that they are contiguous. No more functions to split & merge the bits into and from discontiguous fields.
- With that, moving the East_Asian_Width bit field to avoid awkward gaps.

The second commit moves the CodePointTrie bit setter functions from the emojipropsbuilder.cpp into the toolutil library, and adds a getCPTrieSize() function. This should help with future experimentation. (Copied from the experimental https://github.com/unicode-org/icu/pull/2926.)

---
I also experimented with moving the Age bits into yet another new trie (small, 8-bit CodePointTrie). (Very easy with that second commit.)

Good: Frees up another 8 bits in properties vector word 0.
Bad: Slightly increases the uprops.icu data size by some 4.5% compared to the current state, or 3.9% compared with when the Block bits were still in properties vector word 0.

With the current Unicode 16 alpha data, we would go from
```
trie size in bytes:                    47608
size in bytes of additional props trie:64624
number of additional props vectors:     2095
number of 32-bit words per vector:         3
number of 16-bit scriptExtensions:       314
size in bytes of Block trie:            7752
data size:                            145816
```

to
```
trie size in bytes:                    47608
size in bytes of additional props trie:60792
number of additional props vectors:     1502
number of 32-bit words per vector:         3
number of 16-bit scriptExtensions:       314
size in bytes of Block trie:            7752
size in bytes of Age trie:             17488
data size:                            152356
```

Since we don't need the additional bits yet, there is no immediate benefit to counterbalance the slight size increase, so I just added a comment.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22785

ALLOW_MANY_COMMITS=true


markusicu:
Dear reviewers: Fairly simple change that blocks the next round of Unicode 16 integration...

richgillam (approved):
LOKTM.

markusicu:
Thanks @richgillam !
@echeran if you have additional feedback, I will be happy to follow up with another PR.

PR #3026, opened 2024-06-05
ICU-22707 bug fix hst=V: hst=NA for Kirat Rai

markusicu:
ICU does not store separate data for the Hangul_Syllable_Type property. This property has been a fully redundant subset of the Grapheme_Cluster_Break, and is therefore trivially derived at runtime.

However, starting with Unicode 16, there are a few characters (initially: some Kirat Rai vowel signs) that have nothing to do with Hangul but have GCB=V in order to make grapheme clusters work for their script.

This PR adds some spot checks and a simple derivation fix.

See the "Segmentation Issues" in https://www.unicode.org/versions/beta-16.0.0.html, and https://www.unicode.org/reports/tr29/tr29-44.html#Modifications

FYI: I thought about testing the ICU implementation against ppucd.txt, but that file does not contain hst data either: We don't currently parse HangulSyllableType.txt at all.

##### Checklist

### DIFF
--- a/icu4c/source/common/uprops.cpp
+++ b/icu4c/source/common/uprops.cpp
@@ -590,7 +590,11 @@ static int32_t scriptGetMaxValue(const IntProperty &/*prop*/, UProperty /*which*
 
 /*
  * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
- * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
+ * Hangul_Syllable_Type is redundant with a subset of Grapheme_Cluster_Break.
+ *
+ * Starting with Unicode 16, there is an exception:
+ * Some Kirat Rai vowels are given GCB=V for proper grapheme clustering, but
+ * they are of course not related to Hangul syllables.
  */
 static const UHangulSyllableType gcbToHst[]={
     U_HST_NOT_APPLICABLE,   /* U_GCB_OTHER */
@@ -610,6 +614,11 @@ static const UHangulSyllableType gcbToHst[]={
 };
 
 static int32_t getHangulSyllableType(const IntProperty &/*prop*/, UChar32 c, UProperty /*which*/) {
+    // Ignore supplementary code points: They all have HST=NA.
+    // This is a simple way to handle the GCB!=hst cases since Unicode 16 (Kirat Rai vowels).
+    if(c>0xffff) {
+        return U_HST_NOT_APPLICABLE;
+    }
     /* see comments on gcbToHst[] above */
     int32_t gcb=(int32_t)(u_getUnicodeProperties(c, 2)&UPROPS_GCB_MASK)>>UPROPS_GCB_SHIFT;
     if(gcb<UPRV_LENGTHOF(gcbToHst)) {
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/UCharacterProperty.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/UCharacterProperty.java
@@ -648,7 +648,11 @@ public final class UCharacterProperty
 
     /*
      * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
-     * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
+     * Hangul_Syllable_Type is redundant with a subset of Grapheme_Cluster_Break.
+     *
+     * Starting with Unicode 16, there is an exception:
+     * Some Kirat Rai vowels are given GCB=V for proper grapheme clustering, but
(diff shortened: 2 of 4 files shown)


eggrobin (approved):
> FYI: I thought about testing the ICU implementation against ppucd.txt, but that file does not contain hst data either: We don't currently parse HangulSyllableType.txt at all.

I am very uneasy about the lack of systematic test here; but that may be nontrivial on the ICU side.

On the other hand, we could trivially add an invariant on the unicodetools side that `\p{GCB=V}&[\u0000-\uFFFF] = \p{HST=V}`, with a comment referring to the ICU ticket and this pull request.
@markusicu, do you want to do that, or should I do it?

markusicu:
> > FYI: I thought about testing the ICU implementation against ppucd.txt, but that file does not contain hst data either: We don't currently parse HangulSyllableType.txt at all.
> 
> I am very uneasy about the lack of systematic test here; but that may be nontrivial on the ICU side.

We could parse the HST data, and add it to ppucd.txt, just for the purpose of testing.
Right now I want to focus on integrating Unicode 16 beta into ICU...

> On the other hand, we could trivially add an invariant on the unicodetools side that ` \p{HST=V} = \p{GCB=V}&[\u0000-\uFFFF]`, with a comment referring to the ICU ticket and this pull request. @markusicu, do you want to do that, or should I do it?

Happy to let you do it. Should be quick & easy. I created a reminder earlier today:
- https://github.com/unicode-org/unicodetools/issues/848

eggrobin:
> We could parse the HST data, and add it to ppucd.txt, just for the purpose of testing.

We could, but that is still a bit of work, and not necessarily worth our time. But the invariant is dead easy, so unicode-org/unicodetools#850.

icu4c/source/common/uprops.cpp
```diff
@@ -591,6 +591,10 @@ static int32_t scriptGetMaxValue(const IntProperty &/*prop*/, UProperty /*which*
 /*
  * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
  * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
+ *
+ * Starting with Unicode 16, this is not quite true:
+ * Some Kirat Rai vowels are given GCB=V for proper grapheme clustering, but
+ * they are of course not related to Hangul syllables.
```
srl295, line 597: Suggest striking fully redundant if it's no longer true. Don't just append to the comment.

icu4j/main/core/src/main/java/com/ibm/icu/impl/UCharacterProperty.java
```diff
@@ -649,6 +649,10 @@ public int getType(int c) {
     /*
      * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
      * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
```
srl295, line 651: Same

icu4c/source/common/uprops.cpp
markusicu: ok done

eggrobin (approved):

icu4c/source/common/uprops.cpp
```diff
@@ -590,7 +590,11 @@ static int32_t scriptGetMaxValue(const IntProperty &/*prop*/, UProperty /*which*
 
 /*
  * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
- * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
+ * Hangul_Syllable_Type is redundant with a subset of Grapheme_Cluster_Break.
```
richgillam, line 593: I don't think just deleting "fully" does the trick here.  I think you either need to say something like "Hangul_Syllable_Type is mostly redundant with a set of Grapheme_Cluster_Break, but starting in Unicode 16, there is an exception..." or "HST used to be fully redundant with a set of GCB, but starting in Unicode 16, this is no longer fully true..."

icu4j/main/core/src/main/java/com/ibm/icu/impl/UCharacterProperty.java
```diff
@@ -648,7 +648,11 @@ public int getType(int c) {
 
     /*
      * Map some of the Grapheme Cluster Break values to Hangul Syllable Types.
-     * Hangul_Syllable_Type is fully redundant with a subset of Grapheme_Cluster_Break.
+     * Hangul_Syllable_Type is redundant with a subset of Grapheme_Cluster_Break.
```
richgillam, line 651: Same comment as above.

richgillam (commented):
LOKTM, except maybe for the clarity of the comments (although it looks like I'm posting this comment too late...)

PR #3027, opened 2024-06-05
ICU-21810 Fix error checking of ucurr.cpp part 1

FrankYFTang:
Change code inside ucurr.cpp to use LocalPointer  to prepare the later changes for correct memory change

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21810

### DIFF
--- a/icu4c/source/common/ucurr.cpp
+++ b/icu4c/source/common/ucurr.cpp
@@ -306,10 +306,9 @@ _findMetaData(const char16_t* currency, UErrorCode& ec) {
     // move out of the root locale file later; if it does, update this
     // code.]
     UResourceBundle* currencyData = ures_openDirect(U_ICUDATA_CURR, CURRENCY_DATA, &ec);
-    UResourceBundle* currencyMeta = ures_getByKey(currencyData, CURRENCY_META, currencyData, &ec);
+    LocalUResourceBundlePointer currencyMeta(ures_getByKey(currencyData, CURRENCY_META, currencyData, &ec));
 
     if (U_FAILURE(ec)) {
-        ures_close(currencyMeta);
         // Config/build error; return hard-coded defaults
         return LAST_RESORT_DATA;
     }
@@ -317,32 +316,25 @@ _findMetaData(const char16_t* currency, UErrorCode& ec) {
     // Look up our currency, or if that's not available, then DEFAULT
     char buf[ISO_CURRENCY_CODE_LENGTH+1];
     UErrorCode ec2 = U_ZERO_ERROR; // local error code: soft failure
-    UResourceBundle* rb = ures_getByKey(currencyMeta, myUCharsToChars(buf, currency), nullptr, &ec2);
+    LocalUResourceBundlePointer rb(ures_getByKey(currencyMeta.getAlias(), myUCharsToChars(buf, currency), nullptr, &ec2));
       if (U_FAILURE(ec2)) {
-        ures_close(rb);
-        rb = ures_getByKey(currencyMeta,DEFAULT_META, nullptr, &ec);
+        rb.adoptInstead(ures_getByKey(currencyMeta.getAlias(),DEFAULT_META, nullptr, &ec));
         if (U_FAILURE(ec)) {
-            ures_close(currencyMeta);
-            ures_close(rb);
             // Config/build error; return hard-coded defaults
             return LAST_RESORT_DATA;
         }
     }
 
     int32_t len;
-    const int32_t *data = ures_getIntVector(rb, &len, &ec);
+    const int32_t *data = ures_getIntVector(rb.getAlias(), &len, &ec);
     if (U_FAILURE(ec) || len != 4) {
         // Config/build error; return hard-coded defaults
         if (U_SUCCESS(ec)) {
             ec = U_INVALID_FORMAT_ERROR;
         }
-        ures_close(currencyMeta);
-        ures_close(rb);
         return LAST_RESORT_DATA;
     }
 
-    ures_close(currencyMeta);
-    ures_close(rb);
     return data;
 }
 
@@ -565,14 +557,14 @@ ucurr_forLocale(const char* locale,
         localStatus = U_ZERO_ERROR;
         UResourceBundle *rb = ures_openDirect(U_ICUDATA_CURR, CURRENCY_DATA, &localStatus);
         UResourceBundle *cm = ures_getByKey(rb, CURRENCY_MAP, rb, &localStatus);
-        UResourceBundle *countryArray = ures_getByKey(rb, id.data(), cm, &localStatus);
+        LocalUResourceBundlePointer countryArray(ures_getByKey(rb, id.data(), cm, &localStatus));
         // https://unicode-org.atlassian.net/browse/ICU-21997
         // Prefer to use currencies that are legal tender.
         if (U_SUCCESS(localStatus)) {
-            int32_t arrayLength = ures_getSize(countryArray);
+            int32_t arrayLength = ures_getSize(countryArray.getAlias());
             for (int32_t i = 0; i < arrayLength; ++i) {
                 LocalUResourceBundlePointer currencyReq(
-                    ures_getByIndex(countryArray, i, nullptr, &localStatus));
+                    ures_getByIndex(countryArray.getAlias(), i, nullptr, &localStatus));
                 // The currency is legal tender if it is *not* marked with tender{"false"}.
                 UErrorCode tenderStatus = localStatus;
                 const char16_t *tender =
@@ -592,7 +584,6 @@ ucurr_forLocale(const char* locale,
                 localStatus = U_MISSING_RESOURCE_ERROR;
             }
         }
-        ures_close(countryArray);
     }
 
     if ((U_FAILURE(localStatus)) && strchr(id.data(), '_') != nullptr) {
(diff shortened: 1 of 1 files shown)


icu4c/source/common/ucurr.cpp
```diff
@@ -954,14 +938,18 @@ toUpperCase(const char16_t* source, int32_t len, const char* locale) {
 
     ec = U_ZERO_ERROR;
     dest = (char16_t*)uprv_malloc(sizeof(char16_t) * MAX(destLen, len));
+    if (dest == nullptr) {
+        return dest;
```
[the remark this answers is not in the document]
FrankYFTang: set U_MEMORY_ALLOCATION_ERROR to what? nothing else return here. ec is local variable before/after my change here.

icu4c/source/common/ucurr.cpp
```diff
@@ -1009,113 +1002,86 @@ collectCurrencyNames(const char* locale,
...
             // Add currency long name.
-            s = ures_getStringByIndex(names, UCURR_LONG_NAME, &len, &ec2);
-            (*currencyNames)[*total_currency_name_count].IsoCode = iso;
+            s = ures_getStringByIndex(names.getAlias(), UCURR_LONG_NAME, &len, &ec2);
             char16_t* upperName = toUpperCase(s, len, locale);
-            (*currencyNames)[*total_currency_name_count].currencyName = upperName;
-            (*currencyNames)[*total_currency_name_count].flag = NEED_TO_BE_DELETED;
-            (*currencyNames)[(*total_currency_name_count)++].currencyNameLen = len;
+
+            if (upperName != nullptr) {
```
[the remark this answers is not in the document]
FrankYFTang: ok, I see. my origional intend is to fix problem, not to migrate the code. But I can split that into two PR if you think that is better.

icu4c/source/common/ucurr.cpp
```diff
@@ -1486,15 +1458,19 @@ getCacheEntry(const char* locale, UErrorCode& ec) {
...
-            uprv_strcpy(cacheEntry->locale, locale);
-            cacheEntry->currencyNames = currencyNames;
-            cacheEntry->totalCurrencyNameCount = total_currency_name_count;
-            cacheEntry->currencySymbols = currencySymbols;
-            cacheEntry->totalCurrencySymbolCount = total_currency_symbol_count;
-            cacheEntry->refCount = 2; // one for cache, one for reference
-            currentCacheEntryIndex = (currentCacheEntryIndex + 1) % CURRENCY_NAME_CACHE_NUM;
-            ucln_common_registerCleanup(UCLN_COMMON_CURRENCY, currency_cleanup);
+            if (cacheEntry != nullptr) {
+                currCache[currentCacheEntryIndex] = cacheEntry;
```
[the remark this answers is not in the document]
FrankYFTang: ok

icu4c/source/common/ucurr.cpp
```diff
@@ -2291,27 +2258,23 @@ ucurr_openISOCurrencies(uint32_t currType, UErrorCode *pErrorCode) {
 U_CAPI int32_t U_EXPORT2
 ucurr_countCurrencies(const char* locale, 
                  UDate date, 
-                 UErrorCode* ec)
-{
+                 UErrorCode* ec) {
```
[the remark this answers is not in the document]
FrankYFTang: The current format make it hard to read now. Would you accept a PR only change the format later?

icu4c/source/common/ucurr.cpp
```diff
@@ -976,28 +964,33 @@ collectCurrencyNames(const char* locale,
...
+        return;
 
-    if(currencyNames == nullptr || currencySymbols == nullptr) {
-      ec = U_MEMORY_ALLOCATION_ERROR;
     }
 
-    if (U_FAILURE(ec)) return;
+    *currencySymbols = (CurrencyNameStruct*)uprv_malloc(sizeof(CurrencyNameStruct) * (*total_currency_symbol_count));
+    if (*currencySymbols == nullptr) {
+        uprv_free(*currencyNames);
```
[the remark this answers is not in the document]
FrankYFTang: that will be a behavior change that you ask me not to do in this PR so I take it out and will put that into a later PR.

icu4c/source/common/ucurr.cpp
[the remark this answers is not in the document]
FrankYFTang: That will be a behavior change which you ask me not to do in this PR so I won't do here.

icu4c/source/common/ucurr.cpp
```diff
@@ -805,21 +796,19 @@ ucurr_getPluralName(const char16_t* currency,
     rb = ures_getByKey(rb, CURRENCYPLURALS, rb, &ec2);
```
roubert, line 796: What about this one, is there any reason for why it shouldn't use `LocalUResourceBundlePointer` too?

icu4c/source/common/ucurr.cpp
```diff
@@ -805,21 +796,19 @@ ucurr_getPluralName(const char16_t* currency,
...
-        s = ures_getStringByKeyWithFallback(rb, "other", len, &ec2);     
+        s = ures_getStringByKeyWithFallback(curr.getAlias(), "other", len, &ec2);     
         if (U_FAILURE(ec2)) {
-            ures_close(rb);
             // fall back to long name in Currencies
             return ucurr_getName(currency, locale, UCURR_LONG_NAME, 
                                  isChoiceFormat, len, ec);
         }
     }
-    ures_close(rb);
```
roubert, line 822: Is it really correct to now never close `rb`?

icu4c/source/common/ucurr.cpp
```diff
@@ -957,7 +941,7 @@ toUpperCase(const char16_t* source, int32_t len, const char* locale) {
     u_strToUpper(dest, destLen, source, len, locale, &ec);
     if (U_FAILURE(ec)) {
         u_memcpy(dest, source, len);
-    } 
+    }
```
roubert, line 944: Unrelated formatting change.

icu4c/source/common/ucurr.cpp
```diff
@@ -1009,113 +993,80 @@ collectCurrencyNames(const char* locale,
...
-            (*currencyNames)[*total_currency_name_count].flag = NEED_TO_BE_DELETED;
-            (*currencyNames)[(*total_currency_name_count)++].currencyNameLen = len;
+
+            (*currencyNames)[(*total_currency_name_count)++] = {iso, upperName, len, NEED_TO_BE_DELETED};
 
             // put (iso, 3, and iso) in to array
             // Add currency ISO code.
-            (*currencySymbols)[*total_currency_symbol_count].IsoCode = iso;
-            (*currencySymbols)[*total_currency_symbol_count].currencyName = (char16_t*)uprv_malloc(sizeof(char16_t)*3);
+            upperName = (char16_t*)uprv_malloc(sizeof(char16_t)*3);
```
roubert, line 1036: There's nothing to gain by reusing the `upperName` variable, instead allocating a new one, like eg. `char16_t* isoUChars` will make it clearer to the reader what's going on.

icu4c/source/common/ucurr.cpp
```diff
@@ -2295,7 +2234,7 @@ ucurr_countCurrencies(const char* locale,
 {
     int32_t currCount = 0;
 
-    if (ec != nullptr && U_SUCCESS(*ec)) 
+    if (ec != nullptr && U_SUCCESS(*ec))
```
roubert, line 2237: Unrelated formatting change.

icu4c/source/common/ucurr.cpp
```diff
@@ -2320,41 +2259,38 @@ ucurr_countCurrencies(const char* locale,
...
+                    LocalUResourceBundlePointer toRes(ures_getByKey(currencyRes.getAlias(), "to", nullptr, &localStatus));
+                    const int32_t *toArray = ures_getIntVector(toRes.getAlias(), &toLength, &localStatus);
 
                     currDate64 = (int64_t)toArray[0] << 32;
                     currDate64 |= ((int64_t)toArray[1] & (int64_t)INT64_C(0x00000000FFFFFFFF));
                     UDate toDate = (UDate)currDate64;
 
-                    if ((fromDate <= date) && (date < toDate))
-                    {
+                    if ((fromDate <= date) && (date < toDate)) {
```
roubert, line 2291: Unrelated formatting change.

Just to be clear about why I think this matters: When I'm debugging ICU and trying to find where and when some behaviour changed and therefore look through the blame log, if it there on a line of code says that the last change was made by a commit with a description like _"Fix error checking …"_ then that's a change I'll have to take a closer look at and if I then see that the change that commit did to this particular line actually wasn't anything but some unrelated formatting change then that's wasted time, but if a change was made by a commit with a description like _"Fix formatting …"_ then I immediately know that I can skip that one, saving time and effort.

roubert (changes requested):

icu4c/source/common/ucurr.cpp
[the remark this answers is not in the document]
roubert: > The current format make it hard to read now.

I wholeheartedly agree, much of the ICU source code is so inconsistently formatted as to make it unnecessarily hard to work with.

> Would you accept a PR only change the format later?

I'd be delighted.

icu4c/source/common/ucurr.cpp
```diff
@@ -2703,19 +2609,18 @@ ucurr_getNumericCode(const char16_t* currency) {
     if (currency && u_strlen(currency) == ISO_CURRENCY_CODE_LENGTH) {
         UErrorCode status = U_ZERO_ERROR;
 
-        UResourceBundle *bundle = ures_openDirect(nullptr, "currencyNumericCodes", &status);
-        ures_getByKey(bundle, "codeMap", bundle, &status);
+        LocalUResourceBundlePointer bundle(ures_openDirect(nullptr, "currencyNumericCodes", &status));
```
roubert, line 2612: This is different from what you're doing in `_findMetaData()` above, there the result from `ures_openDirect()` is stored in a plain pointer upon which `ures_close()` never gets called while the result from the subsequent call to `ures_getByKey()` is stored in a `LocalUResourceBundlePointer`, the opposite of what's being done here. Why are these different, is that intentional, is that correct?

icu4c/source/common/ucurr.cpp
FrankYFTang: I talk with Markus about this. ures_getByKey or ures_getByKeyWithFallback  take the third parameter as an optional fill in, in this case, the third parameter is rb which mean it will put the result into rb without allocate additional memory. so the line of LocalUResourceBundlePointer curr will put that pointer into a LocalUResourceBundlePointer.

icu4c/source/common/ucurr.cpp
FrankYFTang: destrutor of curr will close it

icu4c/source/common/ucurr.cpp
FrankYFTang: I think we should remove those junk whenever we see it. I do not mind to revert the { change, but this is an extra space or tab.

richgillam (approved):
LOKTM

roubert (approved):
LGTM

markusicu:
PS: The internal ures_... functions are another part of ICU that would benefit from modernization. In particular, the fillIn parameters were useful for avoiding some memory allocations, but are hard to get right, and hard to look at. If we did this with modern C++, we would return C++ objects (at least structs with destructors) by value, and move operators...

PR #3030, opened 2024-06-06
ICU-22310 Revert now obsolete autoconf "pld hack"

roubert:
This hack was originally added for ICU-7972 by commit 63e1b55feaa6effb0aa4478baa0f9ea312194a3b.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22310

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/configure
+++ b/icu4c/source/configure
@@ -1708,52 +1708,6 @@ fi
 
 } # ac_fn_c_try_cpp
 
-# ac_fn_cxx_try_link LINENO
-# -------------------------
-# Try to link conftest.$ac_ext, and return whether this succeeded.
-ac_fn_cxx_try_link ()
-{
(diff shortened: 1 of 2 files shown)


roubert:
I can't figure out what purpose this originally served but I also can't find any changes in the results of running the configure script after I've removed it, so I guess that whatever it once did was something that's no longer needed at least since autoconf 2.69 so would you please be so kind and take a look and make sure that it really doesn't still do something useful that I just can't see?

srl295:
> I can't figure out what purpose this originally served but I also can't find any changes in the results of running the configure script after I've removed it, so I guess that whatever it once did was something that's no longer needed at least since autoconf 2.69 so would you please be so kind and take a look and make sure that it really doesn't still do something useful that I just can't see?

As it said, it seems to be something that broke with "newer" autoconf but may have been fixed since. 

Pld is a Linux distribution; they had figured out the hack.

PR #3035, opened 2024-06-12
ICU-22798 Avoid stack overflow by return error.

FrankYFTang:
Change the nested level limit from 0x7fff to 0x0fff since the fuzzer found in some machine 6400 level deep already cause stack overflow.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22798

### DIFF
--- a/icu4c/source/common/messagepattern.cpp
+++ b/icu4c/source/common/messagepattern.cpp
@@ -437,7 +437,7 @@ MessagePattern::parseMessage(int32_t index, int32_t msgStartLength,
     if(U_FAILURE(errorCode)) {
         return 0;
     }
-    if(nestingLevel>Part::MAX_VALUE) {
+    if(nestingLevel>Part::MAX_NESTED_LEVELS) {
         errorCode=U_INDEX_OUTOFBOUNDS_ERROR;
         return 0;
     }
--- a/icu4c/source/common/unicode/messagepattern.h
+++ b/icu4c/source/common/unicode/messagepattern.h
@@ -821,6 +821,7 @@ public:
 
         static const int32_t MAX_LENGTH=0xffff;
         static const int32_t MAX_VALUE=0x7fff;
+        static const int32_t MAX_NESTED_LEVELS=0x03ff;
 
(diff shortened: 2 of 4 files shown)


richgillam (dismissed):
LGTM

icu4c/source/common/messagepattern.cpp
```diff
@@ -437,7 +437,7 @@ MessagePattern::parseMessage(int32_t index, int32_t msgStartLength,
     if(U_FAILURE(errorCode)) {
         return 0;
     }
-    if(nestingLevel>Part::MAX_VALUE) {
+    if(nestingLevel>Part::MAX_NESTED_LEVELS) {
```
rp9-next, line 440: The test `TestICU22798` is failing on Windows. After local repo, I found the value of `nestingLevel` to be zero here.
In the stack trace, this value is coming from [`MessagePattern::parse`](https://github.com/unicode-org/icu/blob/aefbea616201f69cd800744791baa3ebc4e258f1/icu4c/source/common/messagepattern.cpp#L270).

We might need to calculate `nestingLevel` (perhaps in the preParse stage) or here if it's passed as zero.

FrankYFTang:
MSVC build all fail on the new test I add. Maybe the stack is even shallower, trying to reduce the limit to see would that make it pass

FrankYFTang:
so I test to use different limit on MSVC in another testing PR
it turn out limit to 0x0fff (4095) or 0x07ff 2047 still cause very deep stack for MSVC and crash, but limit to 0x1ff (511) work. Trying the limit of 0x3ff (2047) on another PR now.

icu4c/source/common/messagepattern.cpp
FrankYFTang: need to change the limit to 0x03ff, 0x0fff is too large for MSVC.

FrankYFTang:
PTAL

rp9-next (approved):

richgillam (commented):
Belated ship-it.  The additional drop in the level is fine with me.

PR #3045, opened 2024-06-20
ICU-22716 Set timeout limit for uregex_match_fuzzer

FrankYFTang:
To avoid unnecessary timeout.
Needed to fix https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=69750

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22716


FrankYFTang:
on my local machine, it took 40 sec to timeout with limit 3000. I think that is a reasonable set up for 60 sec time out on the fuzzer first.

aheninger:
Optional: it might make sense to cut the limit back to something runs for only a few seconds - say 300 or so. It would give the fuzzer more time to explore additional cases, and spend less time pushing exponential time expressions to longer and longer runs, which seems unlikely to uncover additional failures.

aheninger (approved):
Optional: it might make sense to cut the limit back to something runs for only a few seconds - say 300 or so. It would give the fuzzer more time to explore additional cases, and spend less time pushing exponential time expressions to longer and longer runs, which seems unlikely to uncover additional failures.

PR #3028, opened 2024-06-05
ICU-22707 Unicode 16 beta jun04

markusicu:
- new short aliases ID_Status, ID_Type
- Unicode 16 beta data as of 2024-jun-04, including
  - https://github.com/unicode-org/cldr/pull/3783

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22707
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [x] API docs and/or User Guide docs changed or added, if applicable

ALLOW_MANY_COMMITS=true

PR #3049, opened 2024-06-26
ICU-22503 Add support for property Indic_Conjunct_Break

echeran:
This PR is for adding support for property `InCB` (`Indic_Conjunct_Break`).

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22503


echeran:
I rebased this PR on top of #3028 . Therefore, this PR:

1. should not be merged until #3028 is merged
2. should only introduce changes for InCB, and should not appear to introduce any changes for other properties after this branch is updated from the latest on `main` after #3028 is merged

richgillam (dismissed):
LOKTM.  Do you also have to make changes on the Java side?

echeran:
Yes, I need the change in Java, since we need a constant (effectively an enum) to represent the new property so that people can provide that vlaue into `UCharacter.getPropertValueEnum()` or `UCharacter.getPropertyName()`.

Although I'm puzzled why that addition of a single line `int` constant is causing the Maven build to just break altogether.

richgillam:
> Yes, I need the change in Java, since we need a constant (effectively an enum) to represent the new property so that people can provide that vlaue into `UCharacter.getPropertValueEnum()` or `UCharacter.getPropertyName()`.
> 
> Although I'm puzzled why that addition of a single line `int` constant is causing the Maven build to just break altogether.

Oops!  My fault.  You do have the Java change in there too.  I missed that.  But it's just `INDIC_CONJUNCT_BREAK`.  Don't you also need constants for the individual properties?

echeran:
> Don't you also need constants for the individual properties?

Yes indeed! Thanks for catching that.

richgillam (dismissed):
Lookin' good!

echeran:
> is causing the Maven build

The Unicode version constant is not being initialized from the Java resource files. I didn't proceed all the way through in the data update instructions in `changes.txt` so I'll be going through that further.

tools/unicode/py/preparseucd.py
```diff
@@ -624,6 +624,17 @@ def ParsePropertyAliases(in_file):
...
+  # https://www.unicode.org/reports/tr44/tr44-33.html#Indic_Conjunct_Break
+  name = "InCB"
+  prop = ("Enumerated", ["InCB", "Indic_Conjunct_Break"], set(), {})
+  _properties[name] = prop
+  _properties[NormPropName(name)] = prop
+  SetDefaultValue(name, "None")
+  AddEnumeratedValue(prop, "None")
+  AddEnumeratedValue(prop, "Consonant")
+  AddEnumeratedValue(prop, "Extend")
+  AddEnumeratedValue(prop, "Linker")
```
markusicu, line 637: Please try to remove these lines and see if the script still works. This property is defined in the regular UCD files (PropertyAliases.txt & PropertyValueAliases.txt), so they should be parsed just fine by this function and the following function.

We should only need to manually define a property here if it is ICU-specific (gcm, lccc, ...) or if it's outside of the UCD proper (like Identifier_Type from UTS 39).

tools/unicode/py/preparseucd.py
```diff
@@ -2012,7 +2023,7 @@ def PrintNameStats():
 # Sample line to match:
 #    U_EA_AMBIGUOUS,
 _prop_and_value_re = re.compile(
-    " *(U_(BPT|DT|EA|GCB|HST|ID_STATUS|ID_TYPE|INPC|INSC|LB|JG|JT|NT|SB|VO|WB)_([0-9A-Z_]+))")
+    " *(U_(BPT|DT|EA|GCB|HST|ID_STATUS|ID_TYPE|INCB|INPC|INSC|LB|JG|JT|NT|SB|VO|WB)_([0-9A-Z_]+))")
```
markusicu, line 2026: FYI: We will still need this line change in any case, so don't revert this whole file :-)

markusicu (commented):
This is missing the Java versions of the updated pnames.icu & uprops.icu files, but since the major version number has not changed in this PR, I would expect uprops.icu to still load fine, except that all characters would look like InCB=None.

tools/unicode/py/preparseucd.py
markusicu: Tried, and we don't need these lines :-)

markusicu (approved):

markusicu:
FYI --
> > Don't you also need constants for the individual properties?
> 
> Yes indeed! Thanks for catching that.

It was also missing the Java port of the uprops.h / uprops.cpp changes, to UCharacterProperty.java. The class initializer error indicated that an int property had been added (and the _LIMIT constant incremented) without adding an entry to the array of IntProps which defines how to fetch the values for each enumerated property from the data structure.

PR #3058, opened 2024-07-01
ICU-22813 Rise the size of the buffers used for the command strings at pkgdata

clopez:
The tool pkgdata uses snprintf() to build the strings of the commands that will execute later during the install process. But the maximum size of this buffers is not enough when there is a long path.

This has caused issues on some CI systems that use very long paths, causing the install process to produce a wrong result.

The maximum path on Linux is 4096 (defined as PATH_MAX at <linux/limits.h>) So the size of SMALL_BUFFER_MAX_SIZE should be 4096 to avoid errors related to truncated paths.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22813


markusicu (approved):
lgtm tnx!
Longer-term, we should rewrite this kind of code to print into a std::string, which we can use here since it's not part of the ICU runtime libraries.

PR #3063, opened 2024-07-22
ICU-22834 MF2: make tests compliant with schema and update spec tests

catamorphism:
This PR changes the code in the ICU4C and ICU4J tests that reads data-driven tests in JSON format to follow [the test schema in the conformance repo](https://github.com/unicode-org/conformance/blob/main/schema/message_fmt2/testgen_schema.json), as well as updating the spec tests to the version in [the spec repo](https://github.com/unicode-org/message-format-wg/tree/main/spec).

Due to the changed spec tests, some non-test code changes were necessary to get the tests to pass:
* ICU4C: Fixed number literal parsing in `Number::format` (this wasn't being done according to spec before, and previously there were no tests to expose that)
* ICU4C: Fixed parsing of `.input` declarations, which had assumed the annotation can't be a reserved annotation
* ICU4C: Fixed bug where `.i` was a parse error rather than an unsupported keyword
* ICU4C: Make parsing of text and escape characters consistent with spec
* ICU4C: Allow markup with space after the initial '{'
* ICU4C: Check for duplicate variant errors (which were recently added to the spec; ICU4J already checked for these errors)
* ICU4J: Allow trailing whitespace after complex messages (recent spec change)
* ICU4J: Treat whitespace after `.input` as optional
* ICU4J: Don't format unannotated number literals as numbers
* Both: Allow leading whitespace before complex messages (also a recent spec change)

Note that some manual changes to the spec tests are necessary; ICU4J currently returns best-effort output rather than throwing exceptions for resolution errors, so tests where a resolution error (e.g. `unknown-function` or `unsupported-statement`) is expected need to be annotated with an `ignoreJava` property. This can be removed once the error handling story has been resolved (see [MFWG issue 782](https://github.com/unicode-org/message-format-wg/issues/782)).

The more interesting changes needed to parse the test files according to the schema include:
* Updating error names according to schema
* Updating how test params are specified (as an array of objects with `name` and `value` properties, rather than as an object)
* Eliminating the `srcs` property to represent multi-line messages, and instead, allowing `src` to be either a single string or array of strings (consistent with [this schema change](https://github.com/unicode-org/conformance/pull/255))
* Handling default test properties

This also made it possible to move all the test filenames into one file in the ICU4J tests (`CoreTest.java`) and delete the others, since tests are handled uniformly now; analogously, removing extra methods in the ICU4C test reader.

Finally, the non-spec tests needed to be changed themselves; the main changes were to change the format of the `params` property (mentioned above) and changing `errors` to `expErrors`. In addition, one file (`icu-test-selectors.json`) needed to be "expanded out" because the schema doesn't support the `shared` and `variations` properties.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22834


catamorphism:
I can't request reviews, but: @markusicu @srl295 @echeran - and @mihnita might want to take a look at the Java changes.

catamorphism:
Also, I'll be on vacation from now until September 9, so I won't see any review comments until I return.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -414,25 +416,185 @@ static FormattedPlaceholder notANumber(const FormattedPlaceholder& input) {
...
+    }
+    case EIGHT: {
+        return 8;
+    }
+    case NINE: {
+        return 9;
+    }
+    default: {
+        // Should be unreachable
+        return 0;
```
srl295, line 461:
```suggestion
    default: {
        // Should be unreachable
        U_ASSERT(isDigit(c));
        return 0;
```

should this handle other numbering systems though?

srl295: and not to bikeshed on this part but there is `ufmt_digitvalue()` and you could ignore values <0 or ≥10 (it parses hex)

icu4c/source/i18n/messageformat2_macros.h
```diff
@@ -46,6 +46,17 @@ using namespace pluralimpl;
 #define LOWERCASE_E ((UChar32)0x0065)
 #define UPPERCASE_E ((UChar32)0x0045)
 
+#define DIGIT_ZERO ((UChar32)0x0030)
+#define DIGIT_ONE ((UChar32)0x0031)
+#define DIGIT_TWO ((UChar32)0x0032)
```
srl295, line 51: why do these have 'DIGIT'?

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -1718,15 +1673,18 @@ void Parser::parseUnsupportedStatement(UErrorCode& status) {
     dataModel.addUnsupportedStatement(builder.build(status), status);
 }
 
-// Terrible hack to get around the ambiguity between `matcher` and `reserved-statement`
-bool Parser::nextIsMatch() const {
+// Terrible hack to get around the ambiguity between unsupported keywords
+// and supported keywords
+// (It's convenient that all supported keywords have the same length)
+bool Parser::nextIs(const UChar32 (&keyword)[7]) const {
     for(int32_t i = 0; i < 6; i++) {
```
srl295, line 1680: move 6 / 7 into a named constant

srl295 (approved):

srl295:
as a followon i'd move the number parsing into ufmt_cmn.cpp

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: > should this handle other numbering systems though?

No, because it's operating on a Number Operand as defined in [the function registry spec](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#number-operands); it's not a localized number.

> and not to bikeshed on this part but there is ufmt_digitvalue() and you could ignore values <0 or ≥10 (it parses hex)

I tried just now to use that, but I realized that `ufmt_digitvalue()` is part of the `io` library and `io` depends on `i18n`, so that would be a circular dependency.  I guess it could be moved into `i18n`, but I'm hesitant to do a refactor like that just to avoid duplicating a relatively small function. What do you think?

icu4c/source/i18n/messageformat2_macros.h
catamorphism: The `DIGIT_` was to work around an error where an `enum` with `ZERO`, `ONE`, and `TWO` as variants is defined in `standardplural.h`, which was included indirectly in `messageformat2_function_registry.cpp` through `number_types.h`, and the messageformat2 macros were getting applied to the enum definition and confusing the compiler. However, it turns out the include of `number_types.h` wasn't needed anymore, so I'm removing it.

catamorphism: Never mind -- it's still necessary because `itformat.cpp` in the test suite indirectly includes both `standardplural.h` and `messageformat2_macros.h`.

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Done in [3e5b079](https://github.com/unicode-org/icu/pull/3063/commits/3e5b0799dd8ae1ca44b7d5f53e2cdba11238853a)

catamorphism:
> as a followon i'd move the number parsing into ufmt_cmn.cpp

See above comment about introducing a circular dependency between `io` and `i18n`.

icu4c/source/i18n/messageformat2_macros.h
markusicu: Why do we need macros like this? Why not use `u'0'`, `u'1'`, ...?

markusicu: More generally, preprocessor macros with generic names in a public header file are a bad idea. We should use regular C++ character and string literals, or C++ constants in an appropriate namespace, or, if we really need macros, use a UXYZ_ prefix, maybe UPRV_ for internal stuff.

And anything that's in a public header file and not in a private class section needs an API doc comment, maybe `/** @internal */`.

srl295: that's a better solution. But perhaps they aren't needed at all.

icu4c/source/i18n/messageformat2_function_registry.cpp
srl295: I didn't realize it was in io, good call. fine as is then.

icu4c/source/i18n/messageformat2_macros.h
```diff
@@ -46,6 +46,9 @@ using namespace pluralimpl;
 #define LOWERCASE_E ((UChar32)0x0065)
 #define UPPERCASE_E ((UChar32)0x0045)
 
+// The first three macro names are prefixed with DIGIT_
+// in order to avoid substituting the macros into the
+// definition of StandardPlural::Form (see standardplural.h)
```
srl295, line 51: if these are needed they should be UMSG_ or similar prefix, or else constants in a namespace.

srl295 (commented):
I think we should be more cautious about the defines in the public header, otherwise looking good.

icu4c/source/i18n/messageformat2_macros.h
```diff
@@ -46,6 +46,9 @@ using namespace pluralimpl;
...
 #define DIGIT_ZERO ((UChar32)0x0030)
 #define DIGIT_ONE ((UChar32)0x0031)
 #define DIGIT_TWO ((UChar32)0x0032)
```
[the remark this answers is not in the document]
markusicu: ... except: what's the point of even creating a named constant for `u'2'` or `U'2'` at all?

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -20,6 +20,9 @@ using namespace pluralimpl;
 
 using namespace data_model;
 
+// All supported keywords happen to have the same length. See Parser::nextIs()
+#define KEYWORD_LENGTH 7
```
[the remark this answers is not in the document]
markusicu: C++ constants are better than CPP macros.
Do we need length constants like this? Note that several often-used UnicodeString functions (incl. constructor, append(), and operator==) now work efficiently with UTF-16 string literals (avoiding u_strlen() calls).
In other cases, we may benefit from std::u16string_view literals.

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -46,22 +46,38 @@ public static MFDataModel.Message parse(String input) throws MFParseException {
     // Parser proper
     private MFDataModel.Message parseImpl() throws MFParseException {
         MFDataModel.Message result;
+        // Determine if message is simple or complex; this requires
+        // looking through whitespace.
+        int savedPosition = input.getPosition();
+        while (true) {
```
mihnita, line 52: Can we use skipOptionalWhitespaces() for this loop?

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -46,22 +46,38 @@ public static MFDataModel.Message parse(String input) throws MFParseException {
...
         if (cp == '.') { // declarations or .match
+            // No need to restore whitespace
             result = getComplexMessage();
         } else if (cp == '{') { // `{` or `{{`
             cp = input.readCodePoint();
             cp = input.peekChar();
             if (cp == '{') { // `{{`, complex body without declarations
                 input.backup(1); // let complexBody deal with the wrapping {{ and }}
                 MFDataModel.Pattern pattern = getQuotedPattern();
+                skipOptionalWhitespaces();
```
mihnita, line 70: I think it would really help to have the ebnf in a comment above this.
To show if there should be some optional whitespaces there or not.

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -589,6 +605,7 @@ private MFDataModel.Message getComplexMessage() throws MFParseException {
             // complex-message   = *(declaration [s]) complex-body
             checkCondition(cp != EOF, "Expected a quoted pattern or .match; got end-of-input");
             MFDataModel.Pattern pattern = getQuotedPattern();
+            skipOptionalWhitespaces(); // Trailing whitespace is allowed
```
mihnita, line 608: Same, ebnf in a comment.
---
In general what I did with the ebnf in comments was to grep them, and compare the result with the official ebnf, to see what needs updating.
And to "prove" that I am following the spec to the letter (as much as I could :-)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/Sources.java
```diff
@@ -0,0 +1,28 @@
...
+    final List<String> sources;
+
+    Sources(List<String> sources) {
+        this.sources = sources;
+    }
+
+    @Override
+    public String toString() {
+        StringJoiner result = new StringJoiner(", ", "[", "]");
+        result.add(sources.toString());
```
mihnita, line 25: This is basically "{" + sources + "}", there is no need for a StringJoiner.

The sources.toString() is just one string.
So the ", " separator in StringJoiner does not matter.
It would, if the code would be something like 
`for (String s : sources) { result.add(s); }`

mihnita:
There is an email thread on how to organize the test data.
Where the idea is to have some kind of shared folder in the root of the repo, with separate sub-folders for testdata ICU only and testdata from cldr.

Because some of the json files here are from cldr "as is" (or at least they should be).
And some are ICU only, shared between C++/Java, but not from cldr.

The cldr test files are copied to icu by an ant script.

---

If you don't know what thread I'm talking about (with vacation and all it is understandable if it was lost), tell me and I'll ping you by email with a link.

mihnita (commented):
Thank you.

Am I wrong, of there was another PR that was kind of similar, but not really?
That I reviewed.
I can't find it now.

icu4c/source/i18n/messageformat2_macros.h
catamorphism: Eliminated them in [1bdaa31](https://github.com/unicode-org/icu/pull/3063/commits/1bdaa3191bbc894d9ed3799c7f944c70216fb53f).

icu4c/source/i18n/messageformat2_macros.h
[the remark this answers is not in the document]
catamorphism: Eliminated these in [1bdaa31](https://github.com/unicode-org/icu/pull/3063/commits/1bdaa3191bbc894d9ed3799c7f944c70216fb53f).

icu4c/source/i18n/messageformat2_parser.cpp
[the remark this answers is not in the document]
catamorphism: Eliminated in [1bdaa31](https://github.com/unicode-org/icu/pull/3063/commits/1bdaa3191bbc894d9ed3799c7f944c70216fb53f).

catamorphism:
@mihnita: 
> Am I wrong, of there was another PR that was kind of similar, but not really?

Yes, there's a draft PR that I intend as a follow-up to this one: https://github.com/unicode-org/icu/pull/3092

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
catamorphism: Not straightforwardly, because if the message is simple, the whitespace can't be skipped (has to be kept as part of the message text). To me it seemed simplest to write code that just looks ahead without consuming any input.

catamorphism: Done in [9bd9556](https://github.com/unicode-org/icu/pull/3063/commits/9bd9556025ad236bb0c94b7954b0efe857f4b73a)

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
catamorphism: Done in [c14417b](https://github.com/unicode-org/icu/pull/3063/commits/c14417b9c098b383f1a5134a99456863cfdb3f91)

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
catamorphism: Done

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/Sources.java
catamorphism: Done in [3c0aa5f](https://github.com/unicode-org/icu/pull/3063/commits/3c0aa5ffa53b92e45f3845c7c69951c318f7ba54)

icu4c/source/i18n/messageformat2_macros.h
catamorphism: Eliminated these in [1bdaa31](https://github.com/unicode-org/icu/pull/3063/commits/1bdaa3191bbc894d9ed3799c7f944c70216fb53f)

catamorphism:
Rebasing now, since I think all the line comments have been addressed.

srl295 (dismissed):
Update LGTM

catamorphism:
@echeran It looks like the "enforce-all-checks" check failed, but that was the only check that failed. I think this is the same issue described in the icu-team email thread from yesterday?

mihnita (approved):
Thank you.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -421,25 +423,185 @@ static FormattedPlaceholder notANumber(const FormattedPlaceholder& input) {
...
+    while (isDigit(s[i])) {
+        i++;
+    }
+    int32_t numDigits = i - firstDigit;
+    int32_t placeValue = 1;
+    int32_t result = 0;
+    for (int32_t j = firstDigit + numDigits - 1; j >= firstDigit; j--) {
+        result += placeValue * parseDigit(s[j]);
+        placeValue *= 10;
+    }
```
echeran, line 485: simpler way to parse a stringified decimal number from front to back rather than back to front
```suggestion
    int32_t result = 0;
    while (isDigit(s[i])) {
        result *= 10;
        result += parseDigit(s[i]);
        i++;
    }
```

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -421,25 +423,185 @@ static FormattedPlaceholder notANumber(const FormattedPlaceholder& input) {
...
+    while (isDigit(s[i])) {
+        i++;
+    }
+    int32_t numDigits = i - firstDigit;
+    double placeValue = 10;
+    double result = 0;
+    for (int32_t j = firstDigit; j < firstDigit + numDigits; j++) {
+        result += ((double) parseDigit(s[j])) / placeValue;
+        placeValue *= 10;
+    }
```
echeran, line 501: 1. can we reuse `parseDigits`, or is there a reason against?
2. can we reduce the number of times we do division to reduce floating point precision loss?

```suggestion
    int32_t firstDigit = i;
    double result = (double) parseDigits(s, i);
    int32_t denominator = 1;
    for (int32_t j = i; j > firstDigit; j--) {
        denominator *= 10;
    }
    result /= denominator;
```

echeran (commented):
Looks good except I have a couple of questions / suggestions. Otherwise, I'm happy to rubber stamp since @mihnita is on board.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -421,25 +423,174 @@ static FormattedPlaceholder notANumber(const FormattedPlaceholder& input) {
...
+    double result = (double) parseDigits(s, i);
+    int32_t denominator = 1;
+    for (int32_t j = i; j > firstDigit; j--) {
+        denominator *= 10;
+    }
+    result /= denominator;
+    return result;
+}
+
+static double parseNumberLiteral(const FormattedPlaceholder& input, UErrorCode& errorCode) {
```
echeran, line 494: Suggestions from @sffc 

1. Use function from `double-conversion.h`
2. Use function from `number_decimalquantity.h`
3. Use function from `util.h` called `parseNumber` to replace the helper methods that parse the integer and fractional portions.

Either option 1 or 2 should be sufficient to do what we need here. Option 1 gives a `double`, option 2 gives a `DecimalQuantity`. At least option 1 should be able to handle sign, decimal point, and exponent, and probably option 2. Option 3 is helpful but doesn't replace as much code as the other 2 options.

catamorphism: I tried both (1) and (2), and the problem is that there are some strings that both functions accept, which are not valid number literals according to the MF2 grammar: for example, `01` (leading zero with no decimal point), `+1` (leading '+'), `.1` (decimal point with no leading 0), and `1.` (trailing decimal point).

Since these cases are relatively simple to check for, I just added a check before calling into `StringToDouble` (option 1). 

I didn't like (3) since calling `parseNumber` to parse the decimal part still means having to do math to add the two parts together, so it doesn't seem to simplify the code that much.

Let me know what you think.

echeran: Looks great! Option 1 seemed the most likely & directly applicable.

echeran (approved):
LGTM!

After you squash, I'm happy to reapprove, and also to click the merge button on your behalf if need be.

catamorphism:
@echeran Squashed -- I think you don't need to reapprove since nothing changed, but if you could click the merge button, that would be great!

mihnita:
Thank you!
M.

PR #3067, opened 2024-07-25
ICU-22838 Add WebAssembly/WASI cross-compilation support

kateinoigakukun:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22838


markusicu:
@srl295 @sffc could you help review these changes?

markusicu:
Please rebase on current main and resolve conflicts. Note that the autoconf scripts have been updated in another PR.
I will try to be better about nudging pull requests along...

kateinoigakukun:
@markusicu Thanks for updating autoconf stuff. I rebased the branch on the latest main.

icu4c/source/common/umutex.cpp
```diff
@@ -55,9 +56,11 @@ std::once_flag initFlag;
...
 
 }  // Anonymous namespace
+#endif
 
 U_CDECL_BEGIN
 static UBool U_CALLCONV umtx_cleanup() {
+#if U_HAVE_ATOMICS
     initMutex->~mutex();
     initCondition->~condition_variable();
     UMutex::cleanup();
```
markusicu, line 66: UMutex::cleanup() is not empty. Should it not still be called? And if so, then umtx_init() should still do `ucln_common_registerCleanup(UCLN_COMMON_MUTEX, umtx_cleanup)`.

icu4c/source/common/umutex.h
```diff
@@ -61,6 +64,7 @@ template struct std::atomic<std::mutex *>;
 #endif
 #endif
 
+#endif
```
markusicu, line 67:
```suggestion
#endif  // U_HAVE_ATOMICS
```

icu4c/source/common/umutex.h
```diff
@@ -88,6 +94,29 @@ inline int32_t umtx_atomic_dec(u_atomic_int32_t *var) {
...
+
+inline int32_t umtx_atomic_inc(u_atomic_int32_t *var) {
+    return ++(*var);
+}
+
+inline int32_t umtx_atomic_dec(u_atomic_int32_t *var) {
+    return --(*var);
+}
+
+#endif
```
markusicu, line 119:
```suggestion
#endif  // U_HAVE_ATOMICS
```

icu4c/source/common/unifiedcache.cpp
```diff
@@ -13,15 +13,42 @@
...
 #include "uassert.h"
 #include "uhash.h"
 #include "ucln_cmn.h"
 
 static icu::UnifiedCache *gCache = nullptr;
+#if U_HAVE_ATOMICS
 static std::mutex *gCacheMutex = nullptr;
 static std::condition_variable *gInProgressValueAddedCond;
+
+static void unifiedcache_init_mutex() {
```
markusicu, line 29: - static functions don't need a filename prefix
- please use camelCase function names

icu4c/source/common/umutex.cpp
kateinoigakukun: On second thought, I realized we don't need `umtx_cleanup` and `umtx_init` at all because:
1. `gListHead` is not used, it's always nullptr (thus `UMutex::cleanup` is always no-op)
2. We don't even use actual std::mutex, so we don't need global shared mutex (`initMutex`, `initCondition`, and `globalMutex`)

icu4c/source/i18n/decimfmt.cpp
```diff
@@ -1637,8 +1642,13 @@ void DecimalFormat::touch(UErrorCode& status) {
     setupFastFormat();
 
     // Delete the parsers if they were made previously
+#if U_HAVE_ATOMICS
     delete fields->atomicParser.exchange(nullptr);
     delete fields->atomicCurrencyParser.exchange(nullptr);
+#else
+    delete fields->atomicParser;
```
aheninger, line 1649: Probably need to assign these two parser fields to be nullptr. The owning object is not being deleted, leaving live invalid pointers.

icu4c/source/i18n/number_mapper.h
```diff
@@ -196,10 +198,18 @@ struct DecimalFormatFields : public UMemory {
     LocalizedNumberFormatter formatter;
 
     /** The lazy-computed parser for .parse() */
+#if U_HAVE_ATOMICS
     std::atomic<::icu::numparse::impl::NumberParserImpl*> atomicParser = {};
+#else
+    ::icu::numparse::impl::NumberParserImpl* atomicParser = nullptr;
```
aheninger, line 204: nullptr vs {} for initialization. The corresponding currency parser initialization, six lines down, is using {}. Should be consistent.

aheninger (dismissed):

markusicu:
@kateinoigakukun please don't squash any more until this PR is approved, to help GitHub track comments.

kateinoigakukun:
@aheninger Thank you for your feedback. They totally make sense to me. Addressed them in 051260b0e5c8355b9ab60946134b6abb1d8ce656

@markusicu Oops, sorry. I'll keep separate fixup commits for further changes.

sffc:
Are you aware of [ICU4X](https://github.com/unicode-org/icu4x), which we've built from the ground up to work well with WebAssembly? It is no_std so doesn't even need WASI in order to build.

icu4c/source/common/unicode/platform.h
```diff
@@ -313,6 +313,20 @@
...
 
+/**
+ * \def U_HAVE_ATOMICS
+ * Defines whether the platform supports atomic operations.
+ * @internal
+ */
+#ifdef U_HAVE_ATOMICS
+    /* Use the predefined value. */
+#elif defined(__wasi__) && !defined(_REENTRANT)
+    /* WASI does not support atomics when wasi-threads feature is not enabled */
```
sffc, line 324: Question: why not use wasi-threads, then? It seems error-prone to add a flag that skips mutexes.

icu4c/source/i18n/decimfmt.cpp
sffc: Please just implement the `exchange` function in your shim.

icu4c/source/i18n/decimfmt.cpp
```diff
@@ -1673,7 +1685,11 @@ const numparse::impl::NumberParserImpl* DecimalFormat::getParser(UErrorCode& sta
     }
 
     // First try to get the pre-computed parser
+#if U_HAVE_ATOMICS
     auto* ptr = fields->atomicParser.load();
+#else
+    auto* ptr = fields->atomicParser;
```
sffc, line 1691: Please just implement the load function in your shim.

sffc (changes requested):
I'm not a fan of U_HAVE_ATOMICS poking so many holes in the ICU4C code base. Synchronization code is among the hardest and most-error-prone code, and this just makes it more so.

It looks like wasi-threads is still stage 1; is that why you can't use it?

A better solution would be to compile only a subset of ICU4C that doesn't use atomics (I know of clients already doing this), and anything that needs atomics just doesn't get compiled.

Alternatively, see if ICU4X serves your needs.

aheninger:
sffc wrote:
> I'm not a fan of U_HAVE_ATOMICS poking so many holes in the ICU4C code base.

Me neither. I was assuming getting ICU running in this environment was considered worth the added complexity.

kateinoigakukun:
Thanks @sffc for your feedback

> Are you aware of [ICU4X](https://github.com/unicode-org/icu4x), which we've built from the ground up to work well with WebAssembly? It is no_std so doesn't even need WASI in order to build.

Yes, I'm aware of the ground-up work. For our needs, because the [Foundation project](https://github.com/swiftlang/swift-corelibs-foundation) heavily depends on ICU4C, the migration might not be trivial and take a long time to get all things stable (e.g. API migrations especially for APIs without no drop-in replacement like IDNA, build system integration with Rust toolchain, and so on). At least, it's hard to reason the migration decision just for WASI platform support.
But I personally agree it *might* be a direction we should take in the long term.


> It looks like wasi-threads is still stage 1; is that why you can't use it?

First, [wasi-threads](https://github.com/WebAssembly/wasi-threads) is a proposal to provide thread creation APIs and not to provide synchronization primitives. And also the proposal is considered legacy and to be replaced with another proposal (but no implementation for the future proposal yet)

The synchronization primitives are provided by a core spec proposal [WebAssembly/threads](https://github.com/WebAssembly/threads).

Under the circumstances, WASI-SDK currently supports two platform variations:
1. Only WASI Preview1
   - Pre-compiled libc and libcxx are compiled without `WebAssembly/threads` feature, no synchroniztion primitive APIs
2. WASI Preview1 + wasi-threads extension.
   - Pre-compiled libc and libcxx provide synchronization primive APIs

Given that wasi-threads is today consiered legacy, both of them don't satisfy the needs here.

So what we want here is synchrnoization primitives but not thread creation (WebAssembly/threads enabled but wasi-threads disabled). However, both WASI-SDK and even libcxx do not support such platform, "synchronization primitives available but no thread creation" for now.

I understand your concern about poking holes, and it's a typical pain point when porting existing software to WebAssembly. To fill the gap, there is an ongoing effort to provide stubs for synchronization primitive APIs without actual synchronization operations (https://github.com/WebAssembly/wasi-libc/pull/518) in wasi-libc and WASI-SDK side. However, it requires some more effort to bring libcxx's `std::mutex` and `std::atomic`, so we don't have a fixed timeline in this direction.


> A better solution would be to compile only a subset of ICU4C that doesn't use atomics (I know of clients already doing this), and anything that needs atomics just doesn't get compiled.

Yes, we considered that direction, however the Foundation projects depends on many ICU features including formatters, locales, and timezones and they all use `umutex.h`, so it's not a realistic approach for us unfortunately.

icu4c/source/i18n/decimfmt.cpp
kateinoigakukun: Do you mean we should add something like `icu::atomic<T>` to shim `std::atomic<T>`?

sffc: Style-wise, it seems cleaner if your new cfg would just polyfill `std::atomic` with something that works in WebAssembly. I don't think it's useful for code now and into the future to have to figure out how and where to put your atomic ifdefs.

sffc:
> Thanks @sffc for your feedback
> 
> > Are you aware of [ICU4X](https://github.com/unicode-org/icu4x), which we've built from the ground up to work well with WebAssembly? It is no_std so doesn't even need WASI in order to build.
> 
> Yes, I'm aware of the ground-up work. For our needs, because the [Foundation project](https://github.com/swiftlang/swift-corelibs-foundation) heavily depends on ICU4C, the migration might not be trivial and take a long time to get all things stable (e.g. API migrations especially for APIs without no drop-in replacement like IDNA, build system integration with Rust toolchain, and so on).

From browsing the project, it looks like most calls of ICU go through here:

https://github.com/swiftlang/swift-corelibs-foundation/blob/main/Sources/CoreFoundation/internalInclude/CFICULogging.h

The bulk of the functions in that list have ICU4X equivalents, including calendrical calculations, datetime formatting, custom datetime patterns and symbols, number formatting, relative datetime formatting (experimental in ICU4X), and list formatting. The only function in that list that sticks out where there's not yet an equivalent is number parsing. You mentioned IDNA; I don't see IDNA in that list, but ICU4X aims to export at least the primitives for IDNA (https://github.com/unicode-org/icu4x/issues/2614).

I can't comment on your concern about Rust toolchains except to observe that other projects with large amounts of legacy C/C++ code and complex build systems have adopted it in the last two years (e.g. [Chromium](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/rust.md)). Having worked on ICU4C for 5 years before working on ICU4X for 3 years, I feel much more confident in Rust's ability to deliver fast, reliable code in the i18n space.

> At least, it's hard to reason the migration decision just for WASI platform support. But I personally agree it _might_ be a direction we should take in the long term for various other advantages of ICU4X.

I don't think a migration would be too difficult, though it would be most successful with a Foundation contributor driving the process in collaboration with the ICU4X team who could fill in any potential gaps through the process.

My most recent talk on ICU4X, [at UTW 2023](https://www.youtube.com/watch?v=oKh2v2qx6Vc), goes over some of the metrics, showing how the team has delivered better code size, memory usage, and performance in most components, _especially_ formatters, which seems to be one of the main components you use in Foundation.

I hope to be giving off a healthy sense of enthusiasm in being a point of contact if you decide to go this route.

> I understand your concern about poking holes, and it's a typical pain point when porting existing software to WebAssembly. To fill the gap, there is an ongoing effort to provide stubs for synchronization primitive APIs without actual synchronization operations ([WebAssembly/wasi-libc#518](https://github.com/WebAssembly/wasi-libc/pull/518)) in wasi-libc and WASI-SDK side. However, it requires some more effort to bring libcxx's `std::mutex` and `std::atomic`, so we don't have a fixed timeline in this direction.

I see; so it sounds like there is likely room to continue making progress in this space in WASI itself. Do you think once WASI lands support for `std::mutex` and `std::atomic`, you might be able to revert this workaround in ICU4C?

Also, it seems like there should already exist polyfills for `std::mutex` and `std::atomic`. Can we land a copy of those in ICU4C (assuming compatible licenses) to avoid reinventing the wheel?

ArcaneNibble:
> However, it requires some more effort to bring libcxx's `std::mutex` and `std::atomic`, so we don't have a fixed timeline in this direction.

fwiw, if https://github.com/WebAssembly/wasi-libc/pull/518 manages to achieve consensus and get merged, the only thing that has to be done to enable `std::mutex` and `std::atomic` is changing a few lines of cmake to `ON` in WASI-SDK (achieving consensus on _that_ may or may not require yet another discussion, but it's not a difficult technical problem)

icu4c/source/i18n/decimfmt.cpp
srl295: I agree. I think you should just implement atomics in a shim instead of trying to disable it everywhere.

kateinoigakukun:
I'll back to this once a new wasi-sdk including the pthread stub +libcxx atomic support will be shipped.

kateinoigakukun:
Hi @sffc @srl295, I updated my patch based on recent WASI SDK improvements.

- https://github.com/unicode-org/icu/pull/3067/commits/06d4d9ed360b00c08931f6849fdedce8ddb436a9 Thanks to @ArcaneNibble and @whitequark's works, `std::mutex` and `std::atomic` are now available for `wasm32-unknown-wasip1` target, so the patch is now significantly simplified. (See https://github.com/WebAssembly/wasi-sdk/issues/546 for the background)
- https://github.com/unicode-org/icu/pull/3067/commits/a76738531dc0960f13f69db831e245dbb8a79a57 Added a new CI job to ensure a build with WASI-SDK is healthy

gb-jos:
@kateinoigakukun i tried the patches on the main, and it complains regarding `U_NL_LANGINFO_CODESET` in `putil.cpp`
Screenshot 2025-10-24 at 18 47 30

However adding `-DU_HAVE_NL_LANGINFO_CODESET=0` to `CXXFLAGS`, got rid of the usage and then everything builds fine. Thanks for the work on the PR.

kateinoigakukun:
@gb-jos Thank you for trying this :) I locally rebased on the top of the latest main with wasi sdk 27 but I couldn't reproduce it  🤔

kateinoigakukun:
@sffc @srl295 Friendly reminder in case this got buried, no rush, just wanted to keep it on your radar!

srl295:
> @sffc @srl295 Friendly reminder in case this got buried, no rush, just wanted to keep it on your radar!

Can you sign the cla ?

kateinoigakukun:
@srl295 Thanks! Signed CLA and squashed commits to pass single-commit check.

sffc:
This has been approved for several months and CI is green and the ticket is accepted so I'm going to push the merge button.

sffc (approved):

PR #3070, opened 2024-07-29
ICU-22696 Add implicit conversion from StringPiece to std::string_view

roubert:
This will allow ICU4C to seamlessly use `std::string_view` internally while continuing to use `StringPiece` in the public API.

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22696

### DIFF
--- a/icu4c/source/common/unicode/stringpiece.h
+++ b/icu4c/source/common/unicode/stringpiece.h
@@ -32,6 +32,7 @@
 #if U_SHOW_CPLUSPLUS_API
 
 #include <cstddef>
+#include <string_view>
 #include <type_traits>
 
 #include "unicode/uobject.h"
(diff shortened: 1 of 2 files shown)


markusicu (commented):
lgtm

Should this become public API (with a proposal)?
Should be harmless, and would be symmetric with the constructor-from-something-looking-like-string_view that you added in ICU 65.

icu4c/source/common/unicode/stringpiece.h
```diff
@@ -176,6 +177,16 @@ class U_COMMON_API StringPiece : public UMemory {
    */
   StringPiece(const StringPiece& x, int32_t pos, int32_t len);
 
+#ifndef U_HIDE_INTERNAL_API
+  /**
+   * Converts to an std::string_view().
```
markusicu, line 182: Nit: I prefer "a" not "an" -- assuming "std" is pronounced "standard" not "es tee dee"... and now I see that in PR #3071 you also wrote "a std::string_view" :-)  (uhash.h)

roubert: Done.

roubert:
> Should this become public API (with a proposal)?

That wouldn't really serve any purpose, code that calls ICU4C uses `icu::StringPiece` only to pass strings into ICU4C as part of ICU4C function signatures, there's no reason to ever use it in code outside of ICU4C itself (and therefore the only part of the public API that matters to users is the set of constructors).

markusicu (approved):

PR #3076, opened 2024-08-01
ICU-22843 UnicodeString <-> std::u16string_view / wstring_view

markusicu:
##### Checklist

### DIFF
--- a/icu4c/source/common/unicode/char16ptr.h
+++ b/icu4c/source/common/unicode/char16ptr.h
@@ -12,6 +12,7 @@
 #if U_SHOW_CPLUSPLUS_API
 
 #include <cstddef>
+#include <string_view>
 
 /**
  * \file
@@ -305,6 +306,36 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
 #endif
     return reinterpret_cast<OldUChar *>(p);
 }
+
+#ifndef U_FORCE_HIDE_INTERNAL_API
+/**
+ * Is T convertible to a std::u16string_view or to a 16-bit std::wstring_view?
+ * @internal
+ */
+template<typename T>
+constexpr bool ConvertibleToU16StringView =
+    std::is_convertible_v<T, std::u16string_view> ||
+    (U_SIZEOF_WCHAR_T==2 && std::is_convertible_v<T, std::wstring_view>);
+
+namespace internal {
+/**
+ * Pass-through overload.
+ * @internal
+ */
+inline std::u16string_view toU16StringView(std::u16string_view sv) { return sv; }
+
+#if U_SIZEOF_WCHAR_T==2
+/**
+ * Basically undefined behavior but sometimes necessary conversion
+ * from std::wstring_view to std::u16string_view.
+ * @internal
+ */
+inline std::u16string_view toU16StringView(std::wstring_view sv) {
+    return { ConstChar16Ptr(sv.data()), sv.length() };
+}
+#endif
+}  // internal
+#endif  // U_FORCE_HIDE_INTERNAL_API
 
 U_NAMESPACE_END
 
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -33,6 +33,7 @@
 #if U_SHOW_CPLUSPLUS_API
 
 #include <cstddef>
+#include <string_view>
 #include "unicode/char16ptr.h"
 #include "unicode/rep.h"
 #include "unicode/std_string.h"
@@ -97,16 +98,21 @@ class UnicodeStringAppendable;  // unicode/appendable.h
 #define US_INV icu::UnicodeString::kInvariant
 
 /**
- * Unicode String literals in C++.
+ * \def UNICODE_STRING
+ * Obsolete macro approximating UnicodeString literals.
  *
- * Note: these macros are not recommended for new code.
- * Prior to the availability of C++11 and u"unicode string literals",
- * these macros were provided for portability and efficiency when
+ * Prior to the availability of C++11 and u"UTF-16 string literals",
+ * this macro was provided for portability and efficiency when
  * initializing UnicodeStrings from literals.
  *
- * They work only for strings that contain "invariant characters", i.e.,
- * only latin letters, digits, and some punctuation.
- * See utypes.h for details.
+ * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+ * length determination:
+ * \code
+ * #include &lt;string_view&gt;
+ * using namespace std::string_view_literals;
+ * UnicodeString str(u"literal"sv);
+ * if (str == u"other literal"sv) { ... }
+ * \endcode
  *
  * The string parameter must be a C string literal.
  * The length of the string, not including the terminating
@@ -121,16 +127,12 @@ class UnicodeStringAppendable;  // unicode/appendable.h
 
 /**
  * Unicode String literals in C++.
- * Dependent on the platform properties, different UnicodeString
- * constructors should be used to create a UnicodeString object from
- * a string literal.
- * The macros are defined for improved performance.
- * They work only for strings that contain "invariant characters", i.e.,
- * only latin letters, digits, and some punctuation.
- * See utypes.h for details.
+ * Obsolete macro approximating UnicodeString literals.
+ * See UNICODE_STRING.
  *
  * The string parameter must be a C string literal.
  * @stable ICU 2.0
+ * @see UNICODE_STRING
  */
 #define UNICODE_STRING_SIMPLE(cs) UNICODE_STRING(cs, -1)
 
@@ -326,6 +328,32 @@ public:
    * @stable ICU 2.0
    */
   inline bool operator== (const UnicodeString& text) const;
+
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Equality operator. Performs only bitwise comparison with `text`
+   * which is, or which is implicitly convertible to,
+   * a std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view.
+   *
+   * For performance, you can use std::u16string_view literals with compile-time
+   * length determination:
+   * \code
+   * #include &lt;string_view&gt;
+   * using namespace std::string_view_literals;
+   * UnicodeString str = ...;
+   * if (str == u"literal"sv) { ... }
+   * \endcode
+   * @param text The string view to compare to this string.
+   * @return true if `text` contains the same characters as this one, false otherwise.
+   * @draft ICU 76
+   */
+  template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
+  inline bool operator==(const S &text) const {
+    std::u16string_view sv(internal::toU16StringView(text));
+    uint32_t len;  // unsigned to avoid a compiler warning
+    return !isBogus() && (len = length()) == sv.length() && doEquals(sv.data(), len);
+  }
+#endif  // U_HIDE_DRAFT_API
 
   /**
    * Inequality operator. Performs only bitwise comparison.
@@ -1897,6 +1925,24 @@ public:
    */
   UnicodeString &fastCopyFrom(const UnicodeString &src);
 
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Assignment operator. Replaces the characters in this UnicodeString
+   * with a copy of the characters from the `src`
+   * which is, or which is implicitly convertible to,
+   * a std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view.
+   *
+   * @param src The string view containing the characters to copy.
+   * @return a reference to this
+   * @draft ICU 76
+   */
+  template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
+  inline UnicodeString &operator=(const S &src) {
+    unBogus();
+    return doReplace(0, length(), internal::toU16StringView(src));
+  }
+#endif  // U_HIDE_DRAFT_API
+
   /**
    * Move assignment operator; might leave src in bogus state.
    * This string will have the same contents and state that the source string had.
@@ -2146,6 +2192,23 @@ public:
    */
   inline UnicodeString& operator+= (const UnicodeString& srcText);
 
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Append operator. Appends the characters in `src`
+   * which is, or which is implicitly convertible to,
+   * a std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view,
+   * to the UnicodeString object.
+   *
+   * @param src the source for the new characters
+   * @return a reference to this
+   * @draft ICU 76
+   */
+  template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
+  inline UnicodeString& operator+=(const S &src) {
+    return doAppend(internal::toU16StringView(src));
+  }
+#endif  // U_HIDE_DRAFT_API
+
   /**
    * Append the characters
    * in `srcText` in the range
@@ -2191,8 +2254,8 @@ public:
             int32_t srcLength);
 
   /**
-   * Append the characters in `srcChars` to the UnicodeString object
-   * at offset `start`. `srcChars` is not modified.
+   * Append the characters in `srcChars` to the UnicodeString object.
+   * `srcChars` is not modified.
    * @param srcChars the source for the new characters
    * @param srcLength the number of Unicode characters in `srcChars`;
    *                  can be -1 if `srcChars` is NUL-terminated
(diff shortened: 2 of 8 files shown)


icu4c/source/common/unicode/unistr.h
```diff
@@ -327,6 +328,53 @@ class U_COMMON_API UnicodeString : public Replaceable
    */
   inline bool operator== (const UnicodeString& text) const;
 
+#ifndef U_HIDE_DRAFT_API
+  /**
+   * Equality operator. Performs only bitwise comparison with a
+   * std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view.
+   * TODO: Should we document this as what it is, "convertible to one of these string views",
+   * rather than claiming that it needs to *be* one?
```
markusicu, line 336: WDYT?

icu4c/source/common/unicode/unistr.h
```diff
@@ -327,6 +328,53 @@ class U_COMMON_API UnicodeString : public Replaceable
...
+   *
+   * @param text The string view to compare to this string.
+   * @return true if `text` contains the same characters as this one, false otherwise.
+   * @draft ICU 76
+   */
+  template<typename StringView, typename = std::enable_if_t<ConvertibleToU16StringView<StringView>>>
+  // TODO: I am tempted to stick this template line into a #define macro like
+  // U_TEMPLATE_StringView_IF_CONVERTIBLE_TO_U16STRING_VIEW
```
markusicu, line 344: WDYT?

icu4c/source/common/unicode/unistr.h
```diff
@@ -3485,7 +3658,17 @@ class U_COMMON_API UnicodeString : public Replaceable
...
+    return doEquals(text.getArrayStart(), len);
+  }
+  UBool doEquals(const char16_t *text, int32_t len) const;
+#ifdef MAYBE_EQUALS_PTR
+  // TODO: This could be a useful optimization, by avoiding u_strlen(text).
+  // However, it is possible that constructing a std::u16string_view is even better
+  // when the text is a string literal, since the compiler should be able to
+  // determine the length at compile time.
+  UBool doEquals(const char16_t *text) const;
+#endif  // MAYBE_EQUALS_PTR
```
markusicu, line 3671: I had originally added this overload for disambiguation.

Then I realized that constructing a UnicodeString or a string_view from a NUL-terminated string calls strlen() which is wasteful when comparing for equality. So I optimized it.

However, I believe that when a string_view is constructed from a string literal, the length is determined at compile time, so creating a string_view and letting the other new overload take it would be fastest in a common use case.

So I plan to remove this again because (a) we no longer need it for the stated purpose of this PR and (b) as an optimization it may backfire.

WDYT?

icu4c/source/common/unicode/unistr.h
```diff
@@ -327,6 +328,53 @@ class U_COMMON_API UnicodeString : public Replaceable
...
+   * Equality operator. Performs only bitwise comparison with a
+   * std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view.
+   * TODO: Should we document this as what it is, "convertible to one of these string views",
+   * rather than claiming that it needs to *be* one?
+   *
+   * @param text The string view to compare to this string.
+   * @return true if `text` contains the same characters as this one, false otherwise.
+   * @draft ICU 76
+   */
+  template<typename StringView, typename = std::enable_if_t<ConvertibleToU16StringView<StringView>>>
```
markusicu, line 342: Also, since this does not have to _be_ a string view but only has to be convertertible to one, I am tempted to change the typename StringView to simply S. As in
```
template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
```

WDYT?

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -22,6 +22,42 @@
 #include "cmemory.h"
 #include "charstr.h"
 
+
+extern void tempCompileTest() {
```
markusicu, line 26: FYI just to make sure that the sv16 & str16 don't work before adding the overloads, and then make sure that they at least compile. This will eventually turn into real test code that actually looks at the objects.

icu4c/source/i18n/number_decimalquantity.cpp
```diff
@@ -1082,7 +1082,7 @@ UnicodeString DecimalQuantity::toScientificString() const {
     result.append(u'E');
     int32_t _scale = upperPos + scale + exponent;
     if (_scale == INT32_MIN) {
-        result.append({u"-2147483648", -1});
+        result.append(UnicodeString(true, u"-2147483648", -1));
```
markusicu, line 1085: FYI
Adding overloads can still break certain call sites, but I think this is rare.
In this case, the old code also copied the contents; now I am aliasing it.

icu4c/source/common/unicode/unistr.h
```diff
@@ -2925,6 +3022,34 @@ class U_COMMON_API UnicodeString : public Replaceable
...
+   * @return a string view of the contents of this string
+   * @draft ICU 76
+   */
+  inline operator std::u16string_view() const {
+    return { getBuffer(), (std::u16string_view::size_type)length() };
+  }
+
+#if U_SIZEOF_WCHAR_T==2 || defined(U_IN_DOXYGEN)
+  /**
+   * Converts to a std::wstring_view.
```
markusicu, line 3038: @eggrobin heads-up this is still here for Windows...

markusicu (commented):
Hi @richgillam I got help from @eggrobin who showed me (https://godbolt.org/z/5WMbWszoE) how to templatize the inputs so that in most cases we go from a group of four new overloads to just one templatized one. I like this much better! Do you agree?

See just the third commit for the changes compared with the proposal.

I added a few "WDYT?" comments for both of you.

icu4c/source/common/unicode/unistr.h
markusicu: @richgillam one of the input types that is convertible is a raw pointer. The input can also be a std::u16string, or, with this PR, a UnicodeString. That's why it seems a little misleading to just document the input as a string view (and to call the template type StringView).

icu4c/source/common/unicode/unistr.h
eggrobin: I agree StringView is not a great name here. No objections to S (for better or worse, something like `typename T` is _very_ idiomatic when one does not have any better ideas; at least S is slightly clearer than T here).

icu4c/source/common/unicode/unistr.h
eggrobin: I don’t think a macro is helpful. With a dedicated named trait this is very concise and idiomatic C++17, whereas a macro is always a puzzle.
(In C++20 it would get more compact, as discussed in another place.)

(Also I find a macro that declares a specific magic name to be very distasteful, though you could parameterize it on the type name. But my earlier point still stands.)

icu4c/source/common/unicode/unistr.h
eggrobin: > I believe that when a string_view is constructed from a string literal, the length is determined at compile time

I am not sure about that, but I would remove it anyway; if we have the templatized overload, we can be as smart as we want later on within that API surface.

(I think to determine the length at compile time you would need an overload templatized on the length that takes a reference to a char16_t array. That is fun, we can absolutely do it, but it arguably would not be an API change anymore, and it seems to be a job for another PR.)

icu4c/source/common/unicode/unistr.h
eggrobin: With my liaison officer’s hat on, this one makes me nervous, because it will be unsafe when we have a good solution on the C++ side.
With my Windows developer’s hat on, I need _something_.

I think it would be safer to have an actual member function here, that way once C++ figures things out, it can be made to return an RAII object that transmutes back after use.

But with my ICU-TC delegate’s hat on, I lean on keeping whatever we do on this particular conversion draft even as we undraft the rest, until we can see more clearly what will happen on the C++ side. Which means we might as well do this for now.

icu4c/source/common/unicode/char16ptr.h
```diff
@@ -306,6 +307,36 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
...
+constexpr bool ConvertibleToU16StringView =
+    std::is_convertible_v<T, std::u16string_view> ||
+    (U_SIZEOF_WCHAR_T==2 && std::is_convertible_v<T, std::wstring_view>);
+
+namespace internal {
+/**
+ * Pass-through overload.
+ * @internal
+ */
+inline std::u16string_view toU16StringView(std::u16string_view sv) { return sv; }
```
richgillam, line 325: What's this doing?  It looks like a no-op.

richgillam (dismissed):
Wow.  There's a lot of stuff here I don't understand (because I've never really worked in a modern dialect of C++), but I definitely like the conciseness.  You'll have to update the API proposal, of course, and maybe that's a good place to go into a little more detail about what you're actually doing.  I was having trouble following it, but maybe that's just me.  Regardless, I trust you guys...

icu4c/source/common/unicode/char16ptr.h
eggrobin: It is, it’s just there so we can call toU16StringView on something that may or may not be a u16string_view already in the templates.
(We could also do that with `if constexpr` but that seems overkill.)

icu4c/source/common/unicode/unistr.h
markusicu: Adding a note about keeping it draft.

icu4c/source/common/unicode/unistr.h
markusicu: rephrasing

icu4c/source/common/unicode/unistr.h
markusicu: I am not proposing (at this time) that we create our own overloads that take an array.
I thought that the compiler was doing this already, but I can't find a reference to that.
I removed this from-pointer overload.

Also, there is another way:
C++17 also added a string literal syntax (constexpr) for std::u16string_view, and that definitely determines the length at compile time. (Requires a "using" directive.)
https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv
I tried it out and added a little documentation.
We should change the ICU library code to use this style.

markusicu: Actually, considering the declaration
```
constexpr basic_string_view( const CharT* s );
```
How could this be a constexpr and not determine the string length at compile time?

eggrobin: Right, it can do the whole thing at compile time (but then the compiler is just doing a strlen).
Note that you can distinguish a thing that takes the length of the literal from one that does strlen, by passing `"something\0like this."`.

icu4c/source/common/unistr.cpp
```diff
@@ -107,12 +109,36 @@ UOBJECT_DEFINE_RTTI_IMPLEMENTATION(UnicodeString)
...
+  if (sumLengths != INT32_MAX) {
+    ++sumLengths;  // space for a terminating NUL if we need one
+  }
+  return UnicodeString(sumLengths, static_cast<UChar32>(0), 0).append(s1).append(s2);
 }
 
+U_COMMON_API UnicodeString U_EXPORT2
+unistr_internalConcat(const UnicodeString &s1, std::u16string_view s2) {
+  // TODO: version of uprv_add32_overflow() that takes a size_t for the second arg
+  // TODO: consider returning a std::optional<int32_t>
```
markusicu: I experimented with a slightly simpler version of that function for non-negative values where the second one is a size_t, but it didn't really make the total code simpler when it was effectively used in only one place.
```
namespace {

/** Version of uprv_add32_overflow() for non-negative numbers. */
std::optional<int32_t> addNonNegIntoInt32(int32_t a, size_t b) {
  U_ASSERT(a >= 0);
  if (b > INT32_MAX) {
    return {};
  }
  uint32_t sum = (uint32_t)a + (uint32_t)b;
  if (sum <= INT32_MAX) {
    return sum;
  } else {
    return {};
  }
}

}  // namespace
```

aheninger:
I like this  template-function approach. It's forcing me to learn something new (to me) with std::enable_if

richgillam (approved):
I got lost in some of the changes (I still haven't really gotten my head around `enable_if`), but it all sounds good and it looks like you've done appropriate testing...

markusicu:
I finally finished the unit test code. PTAL

markusicu:
I will look at the MSVC failures :-/

eggrobin:
> I will look at the MSVC failures :-/

Since I do my ICU4C work on MSVC, I am happy to look at them next week if you prefer.

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2400,3 +2411,173 @@ void UnicodeStringTest::TestLargeAppend() {
...
+    UnicodeString aliasFromPtr = UnicodeString::readOnlyAlias(p16);
+    assertTrue("aliasFromPtr pointer alias", aliasFromPtr.getBuffer() == p16);
+    assertEquals("aliasFromPtr length", 3, aliasFromPtr.length());
+
+    UnicodeString aliasFromSV = UnicodeString::readOnlyAlias(sv16);
+    assertTrue("aliasFromSV pointer alias", aliasFromSV.getBuffer() == sv16.data());
+    assertEquals("aliasFromSV length", (int32_t)sv16.length(), aliasFromSV.length());
+
+    UnicodeString aliasFromStr = UnicodeString::readOnlyAlias(str16);
+    assertTrue("aliasFromStr pointer alias", aliasFromStr.getBuffer() == str16.data());
```
eggrobin, line 2529: Here is your problem I think:
```suggestion
    assertTrue("aliasFromStr pointer alias", aliasFromStr.getBuffer() == reinterpret_cast<const char16_t*>(str16.data()));
```

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2400,3 +2411,173 @@ void UnicodeStringTest::TestLargeAppend() {
...
+    assertEquals("aliasFromStr length", (int32_t)str16.length(), aliasFromStr.length());
```
eggrobin, line 2530: I would note that we now consistently use static_cast for this since #3060, and that while it is verbose there is some value in not having to think about any of those C casts that might hide nasty stuff…

But really we should just not have to cast here, assertEquals should be templatized and should accept heterogeneous arguments, and then the call site can look unambiguously nicer.

markusicu: > I would note that we now consistently use static_cast for this since #3060, and that while it is verbose there is some value in not having to think about any of those C casts that might hide nasty stuff…

I think that static_cast for simple integer type casts adds a lot more clutter than value.

> But really we should just not have to cast here, assertEquals should be templatized and should accept heterogeneous arguments, and then the call site can look unambiguously nicer.

I don't know if templates would let us compare signed and unsigned types unhindered.

We have had a little discussion about switching ICU4C tests to using the Google unit test framework, but someone would have to work out how to do so. Probably only for new test code, which means adding a whole new, very different test suite, and one probably less portable than the primitive one we have.

markusicu:
It looks like I got MSVC to work where the wchar_t version of the test runs. Back to @richgillam & @eggrobin

icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2408,48 +2412,176 @@ void UnicodeStringTest::TestLargeAppend() {
...
-    UnicodeString fromPtr(p16);
-    UnicodeString fromSV(sv16);
-    UnicodeString fromSV2(u"sv16_2"sv);  // u16string_view literal
-    UnicodeString fromStr(str16);
+    // These copy the string contents.
+    UnicodeString fromPtr(p16);  // pointer is convertible to std::u16string_view
+    UnicodeString fromSV(sv16);  // std::u16string_view itself
+    UnicodeString fromSV2(u"sv16_2"sv);  // std::u16string_view literal
+    UnicodeString fromStr(str16);  // std::u16string is convertible to std::u16string_view
+    assertEquals("UnicodeString(const char16_t *pointer)", UnicodeString(u"p16", 3), fromPtr);
```
richgillam, line 2429: Would it make sense to do `UnicodeString(u"p16", -1)` instead of `UnicodeString(u"p16", 3)` here and in all the analogous places?  I'm a little worried somebody will change the string contents and forget they also have to change the length.

richgillam (approved):
Tests look good.  I had one nit, but I'm fine if you ignore it.

richgillam:
> Also, I am considering a follow-up proposal+PR for the same kind of template overload for operator!=. Seems silly not to have done it right away.

Ouch!  I'm bummed we didn't think of that.  I'm fine with that in a follow-up PR, although I kind of wish we'd thought about it in time to get it into this one.

icu4c/source/test/intltest/ustrtest.cpp
richgillam: > If you don't mind, I would like to squash and then merge as is.

That's fine with me.

PR #3077, opened 2024-08-01
ICU-22818 Fix memory leak during error in messageformat2_data_model.cpp

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22818

### DIFF
--- a/icu4c/source/i18n/messageformat2_data_model.cpp
+++ b/icu4c/source/i18n/messageformat2_data_model.cpp
@@ -1091,13 +1091,13 @@ MFDataModel::MFDataModel(const MFDataModel::Builder& builder, UErrorCode& errorC
         U_ASSERT(builder.selectors != nullptr);
         int32_t numVariants = builder.variants->size();
         int32_t numSelectors = builder.selectors->size();
-        Variant* variants = copyVectorToArray<Variant>(*builder.variants, errorCode);
-        Expression* selectors = copyVectorToArray<Expression>(*builder.selectors, errorCode);
+        LocalArray<Variant> variants(copyVectorToArray<Variant>(*builder.variants, errorCode), errorCode);
+        LocalArray<Expression> selectors(copyVectorToArray<Expression>(*builder.selectors, errorCode), errorCode);
         if (U_FAILURE(errorCode)) {
             bogus = true;
             return;
         }
-        body.emplace<Matcher>(Matcher(selectors, numSelectors, variants, numVariants));
+        body.emplace<Matcher>(Matcher(selectors.orphan(), numSelectors, variants.orphan(), numVariants));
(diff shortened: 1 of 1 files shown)


icu4c/source/i18n/messageformat2_data_model.cpp
```diff
@@ -1094,6 +1094,8 @@ MFDataModel::MFDataModel(const MFDataModel::Builder& builder, UErrorCode& errorC
         Variant* variants = copyVectorToArray<Variant>(*builder.variants, errorCode);
         Expression* selectors = copyVectorToArray<Expression>(*builder.selectors, errorCode);
         if (U_FAILURE(errorCode)) {
+            delete variants;
```
aheninger, line 1097: `copyVectorToArray` is doing an array new, so at a minimum the deletes here need to be array deletes.

Even better would be to use `LocalArray`

aheninger (dismissed):

icu4c/source/i18n/messageformat2_data_model.cpp
FrankYFTang: PTAL

icu4c/source/i18n/messageformat2_data_model.cpp
```diff
@@ -1091,13 +1091,13 @@ MFDataModel::MFDataModel(const MFDataModel::Builder& builder, UErrorCode& errorC
         U_ASSERT(builder.selectors != nullptr);
         int32_t numVariants = builder.variants->size();
         int32_t numSelectors = builder.selectors->size();
-        Variant* variants = copyVectorToArray<Variant>(*builder.variants, errorCode);
-        Expression* selectors = copyVectorToArray<Expression>(*builder.selectors, errorCode);
+        LocalArray<Variant> variants(copyVectorToArray<Variant>(*builder.variants, errorCode), errorCode);
+        LocalArray<Expression> selectors(copyVectorToArray<Expression>(*builder.selectors, errorCode), errorCode);
```
richgillam, line 1095: Do these arrays own their contents?  If so, you need to account for that.

FrankYFTang: I did.   I call orphan() to release the ownership

richgillam: orphan() releases ownership of the arrays themselves.  I was asking about the _elements_ of the arrays.  I don't know what `Variant` and `Expression` are.  But if I'm reading this right, they're inline structs, not pointers, and so there isn't an ownership issue here and I asked the question in error.  Sorry about that.

richgillam (approved):

PR #3079, opened 2024-08-01
ICU-22823 Fix memory leak while error in uniset.cpp

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22823

### DIFF
--- a/icu4c/source/common/uniset.cpp
+++ b/icu4c/source/common/uniset.cpp
@@ -981,6 +981,7 @@ void UnicodeSet::_add(const UnicodeString& s) {
     }
     strings->sortedInsert(t, compareUnicodeString, ec);
     if (U_FAILURE(ec)) {
+        delete t;
         setToBogus();
     }
 }


richgillam (approved):
LGTM

icu4c/source/common/uniset.cpp
```diff
@@ -981,6 +981,7 @@ void UnicodeSet::_add(const UnicodeString& s) {
     }
     strings->sortedInsert(t, compareUnicodeString, ec);
     if (U_FAILURE(ec)) {
+        delete t;
```
aheninger, line 984: I think this is incorrect, and that the original code is OK.

The `strings` UVector in UnicodeSet is set up with a deleter function. see uniset.cpp:1610.
Which means that `strings->sortedInsert(t, ...)` takes ownership the element being inserted, t. This is true always, failure or no. In the event of a failure, the `UVector` automatically deletes the item that it could not insert. See uvector.cpp:463

aheninger (changes requested):

icu4c/source/common/uniset.cpp
markusicu: Good catch! Thanks Andy!

FrankYFTang:
Invalid PR. Close it.

PR #3081, opened 2024-08-02
ICU-22825 Fix memLeak during error in tznames_impl.cpp

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22825

### DIFF
--- a/icu4c/source/i18n/tznames_impl.cpp
+++ b/icu4c/source/i18n/tznames_impl.cpp
@@ -235,24 +235,18 @@ TextTrieMap::put(const char16_t *key, void *value, UErrorCode &status) {
         LocalPointer<UVector> lpLazyContents(new UVector(status), status);
         fLazyContents = lpLazyContents.orphan();
     }
-    if (U_FAILURE(status)) {
-        if (fValueDeleter) {
-            fValueDeleter((void*) key);
+    if (U_SUCCESS(status)) {
+        U_ASSERT(fLazyContents != nullptr);
+        char16_t *s = const_cast<char16_t *>(key);
+        fLazyContents->addElement(s, status);
+        if (U_SUCCESS(status)) {
+            fLazyContents->addElement(value, status);
+            return;
         }
-        return;
     }
-    U_ASSERT(fLazyContents != nullptr);
-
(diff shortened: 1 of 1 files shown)


richgillam (approved):

icu4c/source/i18n/tznames_impl.cpp
```diff
@@ -731,6 +731,7 @@ class ZNames : public UMemory {
                 nameinfo->type = getTZNameType(static_cast<UTimeZoneNameTypeIndex>(i));
                 trie.put(name, nameinfo, status); // trie.put() takes ownership of the key
                 if (U_FAILURE(status)) {
+                    uprv_free(nameinfo);
```
markusicu, line 734: If trie.put() truly takes ownership, then it should do so even when something goes wrong. In that case, we would now get a double-delete problem.
Please check.
@yumaoka @aheninger FYI

markusicu (changes requested):

icu4c/source/i18n/tznames_impl.cpp
FrankYFTang: you are right. I think the real issue is the TextTrieMap::put() implement called the deleter on the wrong variable. it should call the deleter to delete the value, not the key

richgillam (approved):
I _think_ this is right.  I was concerned that you might need to take ownership of both the key and the value, but the documentation says this particular function is for cases where the key is in static memory owned by somebody else (the resource bundle), so I think you're doing the right thing.

FrankYFTang:
@sffc could you double check since you are the one who wrote the origional code.

sffc (approved):
The minimal fix would be to change `fValueDeleter((void*) key);` to `fValueDeleter(value);`

The rest of the refactor looks fine.

FrankYFTang:
> The minimal fix would be to change `fValueDeleter((void*) key);` to `fValueDeleter(value);`

no, that will not be enugh because for the last line of 
fLazyContents->addElement(value, status); in the origional line 255
we also need a "if fail if deleter delete block" in case the return status faile also. That is the leak the reporter detected.

PR #3082, opened 2024-08-02
ICU-22845 Better iterations for the ICU4J UnicodeSet

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22845

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
@@ -17,7 +17,12 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.SortedSet;
+import java.util.Spliterator;
 import java.util.TreeSet;
+import java.util.function.IntConsumer;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import com.ibm.icu.impl.BMPSet;
 import com.ibm.icu.impl.CharacterPropertiesImpl;
@@ -278,14 +283,22 @@ import com.ibm.icu.util.VersionInfo;
  *     </tr>
  *   </table>
  * </blockquote>
- * <p>To iterate over contents of UnicodeSet, the following are available:
- * <ul><li>{@link #ranges()} to iterate through the ranges</li>
- * <li>{@link #strings()} to iterate through the strings</li>
- * <li>{@link #iterator()} to iterate through the entire contents in a single loop.
- * That method is, however, not particularly efficient, since it "boxes" each code point into a String.
+ *
+ * <p>To iterate over contents of {@code UnicodeSet}, the following are available:
+ * <ul>
+ *   <li>to iterate over the ranges: {@link #ranges()}, {@link #rangeStream()}</li>
+ *   <li>to iterate over the strings: {@link #strings()}, {@link #stringStream()}</li>
+ *   <li>to iterate over the code points: {@link #codePoints()}, {@link #codePointStream()}</li>
+ *   <li>to iterate over the entire contents in a single loop: this class itself is {@link Iterable},
+ *       or use {@link #stream()}.<br>
+ *       All of these method are, however, not particularly efficient,
+ *       since they convert each individual code point to a {@code String}.
  * </ul>
- * All of the above can be used in <b>for</b> loops.
- * The {@link com.ibm.icu.text.UnicodeSetIterator UnicodeSetIterator} can also be used, but not in <b>for</b> loops.
+ *
+ * <p>The iterators and streams methods work as expected in idiomatic Java usage.<br>
+ * The {@link UnicodeSetIterator} cannot be used in <b>for</b> loops, and it is not very Java-idiomatic, because it is old.
+ * But it might be faster in certain use cases. We recommend that you measure in performance sensitive code.<br>
+ *
  * <p>To replace, count elements, or delete spans, see {@link com.ibm.icu.text.UnicodeSetSpanner UnicodeSetSpanner}.
  *
  * @author Alan Liu
@@ -5126,6 +5139,222 @@ public class UnicodeSet extends UnicodeFilter implements Iterable<String>, Compa
         // TODO: Check if the Unicode Tools or Unicode Utilities really need this.
         CharacterPropertiesImpl.clear();
         XSYMBOL_TABLE = xSymbolTable;
+    }
+
+    /**
+     * Returns a {@link Stream} of {@link EntryRange} values from this {@code UnicodeSet}.
+     *
+     * <p><b>Warnings:</b>
+     * <ul>
+     *   <li>The {@link EntryRange} instance is the same each time; the contents are just reset.
+     *   <li>To iterate over the full contents, you have to also iterate over the strings.
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return a {@link Stream} of {@link EntryRange}
+     *
+     * @draft ICU 76
+     */
+    public Stream<EntryRange> rangeStream() {
+        // Must use false to never make this parallel because the iterator always returns the same EntryRange object.
+        return StreamSupport.stream(ranges().spliterator(), false);
+    }
+
+    /**
+     * Returns a {@link Stream} of {@code String} values from this {@code UnicodeSet}.
+     *
+     * <p><b>Warnings:</b>
+     * <ul>
+     *   <li>To iterate over the full contents, you have to also iterate over the ranges or code points.
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return a {@link Stream} of {@code String}
+     *
+     * @draft ICU 76
+     */
+    public Stream<String> stringStream() {
+        return strings().stream();
+    }
+
+    /**
+     * Returns an {@link IntStream} of Unicode code point values from this {@code UnicodeSet}.
+     *
+     * <p><b>Warnings:</b>
+     * <ul>
+     *   <li>To iterate over the full contents, you have to also iterate over the strings.
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return an {@link IntStream} of Unicode code point values
+     *
+     * @draft ICU 76
+     */
+    public IntStream codePointStream() {
+        return StreamSupport.intStream(new CodePointSpliterator(this), false);
+    }
+
+    /**
+     * Returns a stream of {@code String} values from this {@code UnicodeSet}.
+     *
+     * <p><b>Warnings:</b>
+     * <ul>
+     *   <li>To iterate over the full contents, you have to also iterate over the strings.
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return a {@link Stream} of {@code String}
+     *
+     * @draft ICU 76
+     */
+    public Stream<String> stream() {
+        return StreamSupport.stream(spliterator(), false);
+    }
+
+    /**
+     * Returns an {@link Iterable} for iteration over all the code points in this set.
+     *
+     * <p><b>Warnings:</b>
+     * <ul>
+     *   <li>This is a convenience method, but comes with a performance penalty
+     *       because it boxes {@code int} into {@code Integer}.<br>
+     *       For an efficient but old alternative use {@link UnicodeSetIterator#next()}.
+     *   <li>To iterate over the full contents, you have to also iterate over the strings.
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return an {@link Iterable} over all the code points
+     *
+     * @draft ICU 76
+     */
+    public Iterable<Integer> codePoints() {
+        return new CodePointIterable(this);
+    }
+
+    private class CodePointIterable implements Iterable<Integer> {
+        private final UnicodeSet unicodeSet;
+
+        CodePointIterable(UnicodeSet unicodeSet) {
+            this.unicodeSet = unicodeSet;
+        }
+
+        @Override
+        public Iterator<Integer> iterator() {
+            return new CodePointIterator(unicodeSet);
+        }
+    }
+
+    private class CodePointIterator implements Iterator<Integer> {
+        private final CodePointIteratorInt cpi;
+
+        CodePointIterator(UnicodeSet unicodeSet) {
+            cpi = new CodePointIteratorInt(unicodeSet);
+        }
+
+        @Override
+        public boolean hasNext() {
+            return cpi.hasNext();
+        }
+
+        @Override
+        public Integer next() {
+            return cpi.next();
+        }
+
+        @Override
+        public void remove() {
+            throw new UnsupportedOperationException();
+        }
+    }
+
+    private static class CodePointSpliterator implements Spliterator.OfInt {
+        private final static int CHARACTERISTICS = Spliterator.SIZED | Spliterator.ORDERED | Spliterator.DISTINCT | Spliterator.NONNULL;
+
+        private final UnicodeSet unicodeSet;
+        private final CodePointIteratorInt cpi;
+
+        CodePointSpliterator(UnicodeSet unicodeSet) {
+            this.unicodeSet = unicodeSet;
+            cpi = new CodePointIteratorInt(unicodeSet);
+        }
+
+        @Override
+        public long estimateSize() {
+            return unicodeSet.size() - unicodeSet.strings.size();
+        }
+
+        @Override
+        public int characteristics() {
+            return unicodeSet.isFrozen() ? Spliterator.IMMUTABLE | CHARACTERISTICS : CHARACTERISTICS;
+        }
+
+        @Override
+        public Spliterator.OfInt trySplit() {
+            /* From the doc:
+             *   > This method may return null for any reason, including emptiness, inability to split after
+             *   > traversal has commenced, data structure constraints, and efficiency considerations.
+             */
+            return null;
+        }
+
+        @Override
+        public boolean tryAdvance(IntConsumer action) {
+            if (action == null) {
+                throw new NullPointerException();
+            }
+
+            if (cpi.hasNext()) {
+                action.accept(cpi.next());
+                return true;
+            }
+            return false;
+        }
+    }
+
+    /**
+     * This class is optimized to iterate on code points and will be used to implement both
+     * the Iterator<Integer> (Integer, boxed value) and the Spliterator.OfInt (int primitive).
+     * It looks exactly like an Iterator<Integer>, but works on the primitive int,
+     * so it can't implement Iterator.
+     */
+    static private class CodePointIteratorInt {
+        private final int[] list;
+        private final int lastRange; 
+        private int currentRange = 0;
+        private int rangeStart;
+        private int rangeLimit;
+
+        public CodePointIteratorInt(UnicodeSet unicodeSet) {
+            this.list = unicodeSet.list;
+            lastRange = unicodeSet.len - 1;
+            currentRange = 0;
+            rangeStart = list[currentRange++];
+            if (lastRange > 0) { // not an empty list
+                rangeLimit = list[currentRange++];
+            } else {
+                rangeLimit = rangeStart; // should be HIGH, the guard value
+            }
+        }
+
+        public boolean hasNext() {
+            return rangeStart < rangeLimit || currentRange < lastRange;
+        }
+
+        public int next() {
+            if (rangeStart >= rangeLimit) {
+                if (currentRange >= lastRange) {
+                  throw new NoSuchElementException();
+                }
+                rangeStart = list[currentRange++];
+                rangeLimit = list[currentRange++];
+            }
+            return rangeStart++;
+        }
     }
 }
 //eof
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
@@ -21,6 +21,7 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.StringJoiner;
@@ -60,6 +61,16 @@ import com.ibm.icu.util.OutputInt;
 public class UnicodeSetTest extends CoreTestFmwk {
 
     static final String NOT = "%%%%";
+
+    // Used to test iterators and streams
+    final static UnicodeSet[] UNICODE_SETS_TO_ITERATE = {
+            UnicodeSet.EMPTY,
+            new UnicodeSet(0, 0), // one code point, zero
+            new UnicodeSet(0x0000, 0x0010), // from zero
+            new UnicodeSet(0x10FFFF, 0x10FFFF), // one code point, max
+            new UnicodeSet(0x10FF00, 0x10FFFF), // end in max
+            populateUnicodeSet()
+    };
 
     private static final boolean isCccValue(int ccc) {
         switch (ccc) {
@@ -841,7 +852,7 @@ public class UnicodeSetTest extends CoreTestFmwk {
             }
         }
     }
-    
+
     @Test
     public void TestSetRelation() {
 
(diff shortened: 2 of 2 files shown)


mihnita:
For convenience the newly generated JavaDoc is here:
https://mihai-nita.net/shared/icu_javadoc/com/ibm/icu/text/UnicodeSet.html

---

Question: should I tag these new APIs as draft?
Or are they OK as is?

Thanks,
M.

markusicu:
> Question: should I tag these new APIs as draft? Or are they OK as is?

All public & protected APIs must have an API tag.
If they must be public or protected but are not intended for public use, then tag them as `@internal`.
Otherwise, all new APIs must be tagged with `@draft ICU 76` as usual.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -278,14 +283,25 @@
...
  * </blockquote>
- * <p>To iterate over contents of UnicodeSet, the following are available:
- * <ul><li>{@link #ranges()} to iterate through the ranges</li>
- * <li>{@link #strings()} to iterate through the strings</li>
- * <li>{@link #iterator()} to iterate through the entire contents in a single loop.
- * That method is, however, not particularly efficient, since it "boxes" each code point into a String.
+ *
+ * <p>To iterate over contents of {@code UnicodeSet}, the following are available:
+ * <ul>
+ *   <li>to iterate through the ranges: {@link #ranges()}, {@link UnicodeSetIterator#nextRange()},
```
markusicu, line 289: minor: I think it's "iterate over" not "iterate through"

markusicu, line 289: minor: I would move the UnicodeSetIterator versions to the ends of each list item

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -278,14 +283,25 @@
...
+ *       {@link #rangeStream()}</li>
+ *   <li>to iterate through the strings: {@link #strings()}, {@link UnicodeSetIterator#next()},
+ *       {@link #stringStream()}</li>
+ *   <li>to iterate through the code points: {@link UnicodeSetIterator#next()},
+ *       {@link #codePoints()}, {@link #codePointStream()}</li>
+ *   <li>to iterate through the entire contents in a single loop: {@link #iterator()},
```
markusicu, line 295: Instead of `{@link #iterator()}` please say something like "this class itself is Iterable".

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -278,14 +283,25 @@
...
+ *       {@link UnicodeSetIterator#next()}, {@link #stream()}.<br>
+ *       All of these method are, however, not particularly efficient,
+ *       since they convert each individual code point to a {@code String}.
  * </ul>
- * All of the above can be used in <b>for</b> loops.
- * The {@link com.ibm.icu.text.UnicodeSetIterator UnicodeSetIterator} can also be used, but not in <b>for</b> loops.
+ *
+ * <p>The iterators and streams methods work as expected in current idiomatic Java usage.<br>
+ * The {@link UnicodeSetIterator} cannot be in <b>for</b> loops, and it is not very Java-ideomatic, because it is old.
+ * But it might be faster in certain use cases. We recommend that you measure in performance sensitive code.<br>
```
markusicu, line 303:
```suggestion
 * <p>The iterators and streams methods work as expected in idiomatic Java usage.<br>
 * The {@link UnicodeSetIterator} cannot be used in <b>for</b> loops, and it is not very Java-idiomatic, because it is old.
 * But it might be faster in certain use cases. We recommend that you measure in performance sensitive code.<br>
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+     * @return a {@link Stream} of {@link EntryRange}
+     */
+    public Stream<EntryRange> rangeStream() {
+        return StreamSupport.stream(ranges().spliterator(), false);
+    }
+
+    /**
+     * Returns a {@link Stream} of {@code String} values from this {@code UnicodeSet}.
+     *
+     * <p><b>Warning:</b> To iterate over the full contents, you have to also iterate over the strings.
```
markusicu, line 5166:
```suggestion
     * <p><b>Warning:</b> To iterate over the full contents, you have to also iterate over the ranges or code points.
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+    public Stream<String> stream() {
+        return StreamSupport.stream(spliterator(), false);
+    }
+
+    /**
+     * Provide for iteration on code points.
+     *
+     * <p>Returns an {@link Iterable} over all the code points, "expanding" the ranges
+     * by iterating on all code points between the {@link EntryRange#codepoint} and
+     * {@link EntryRange#codepointEnd}.
```
markusicu, line 5210:
```suggestion
     * Returns an {@link Iterable} for iteration over all the code points in this set.
```
Talking about the ranges here seems unnecessary and distracting.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+     *
+     * <p><b>Warning:</b> This is a convenience method, but comes with a performance penalty
+     * because it boxes {@code int} into {@code Integer}.<br>
+     * For faster iteration use the {@link #codePointStream()} or {@link #charAt(int)}.
```
markusicu, line 5214: charAt(index) is not fast -- it linearly walks through each range until it finds the one with the index-th code point. Each time.
Better to point to UnicodeSetIterator.next() for an efficient but old alternative.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+
+    private class CodePointsIterator implements Iterator<Integer> {
+        private int pos = 0;
+        private int lastCp = -1; // don't call `charAt` twice, in `hasNext()` and `next()`
+
+        @Override
+        public boolean hasNext() {
+            lastCp = charAt(pos++);
+            return lastCp != -1;
+        }
```
markusicu, line 5242: 1. I wouldn't iterate in hasNext(). I would init the iterator with the number of code points or ranges, keep and index, and make hasNext() compare the index with the total number.
2. Don't call charAt(index). It's slow.
3. I suggest you init with getRangeCount(), keep a rangeIndex, and when you go to the next range then call getRangeStart(rangeIndex) and getRangeEnd(rangeIndex); store those, iterate over each code point of the current range.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+         *     > codepoints and strings."
+         *
+         * And from the Spliterator.SIZED doc:
+         *     > Characteristic value signifying that the value returned from
+         *     > {@code estimateSize()} prior to traversal or splitting represents a
+         *     > finite size that, in the absence of structural source modification,
+         *     > represents an exact count of the number of elements that would be
+         *     > encountered by a complete traversal.
+         * Note "exact count of the number of elements that would be encountered"
+         */
```
markusicu, line 5277: I don't understand. The exact number is known: `size - strings.size()`

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+        }
+
+        @Override
+        public long estimateSize() {
+            return uset.size();
+        }
+
+        @Override
+        public int characteristics() {
+            // TODO: consider calling `isFrozen()` and also return `Spliterator.IMMUTABLE` if true?
```
markusicu, line 5295: sure :-)

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5143,179 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+             *   > traversal has commenced, data structure constraints, and efficiency considerations.
+             */
+            return null;
+        }
+
+        @Override
+        public boolean tryAdvance(IntConsumer action) {
+            if (action == null)
+                throw new NullPointerException();
+            int cp = uset.charAt(index++);
```
markusicu, line 5312: not charAt()...

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -61,6 +61,13 @@ public class UnicodeSetTest extends CoreTestFmwk {
 
     static final String NOT = "%%%%";
 
+    // Used to test iterators and streams
+    final static UnicodeSet[] UNICODE_SETS_TO_ITERATE = {
+            UnicodeSet.EMPTY,
+            // UnicodeSet.ALL_CODE_POINTS, // this works, but the tests are visibly slow
```
markusicu, line 67: faster without charAt()?
if you don't intend to replace this with another set, then i suggest deleting this comment -- it's not unusual that doing something for all 1.1M Unicode code points would be slow

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        unicodeSet.applyPattern("\\p{Number}");
+        // Single code point
+        unicodeSet.add('X'); // adds a code point in the ASCII range
+        unicodeSet.add('Σ'); // adds a code point in the Greek block
+        unicodeSet.add(0x1F600); // adds a code point above BMP (😀)
+        // Code point ranges and from CharSequence
+        unicodeSet.add('A', 'F'); // adds a code point range in the ASCII range
+        unicodeSet.add('α', 'ζ'); // adds a code point range in the Greek block
+        unicodeSet.add(0x1F347, 0x1F353); // adds a code point range above BMP (🍇-🍓)
+        unicodeSet.addAll("hello"); // adds code points from CharSequence
```
markusicu, line 3223: this seems odd: coverage for addAll(CharSequence) but makes no difference for iteration?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+
+    @Test
+    public void testIterationMethodsCodepoints() {
+        for (UnicodeSet us : UNICODE_SETS_TO_ITERATE) {
+
+            // Build the reference test result
+            StringJoiner expected = newResult();
+            int i = 0;
+            while (true) {
+                int cp = us.charAt(i++);
```
markusicu, line 3238: ok for test code; although (optional) you could consider using UnicodeSetIterator as the reference implementation, rather than testing it again (since it has its own test code)

----
alternative, double-loop-iterate over the ranges and then their code points?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+    }
+
+    /*
+     * Helper methods for testing various iterations.
+     * The main goal is to make the results of any possible failures more readable
+     * by formatting code points to something like U+03A3(Σ) and wrapping strings in double quotes.
+     */
+
+    private static String codePointToString(int codepoint) {
+        String fromCodePoint = new String(Character.toChars(codepoint));
```
markusicu, line 3338: UTF16.valueOf(c)?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        assertEquals("code points + strings :: stream", expect.toString(), fromStream.toString());
+    }
+
+    /*
+     * Helper methods for testing various iterations.
+     * The main goal is to make the results of any possible failures more readable
+     * by formatting code points to something like U+03A3(Σ) and wrapping strings in double quotes.
+     */
+
+    private static String codePointToString(int codepoint) {
```
markusicu, line 3337: A "code point" variable should be called camelCase codePoint -- or simply cp or c.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        String fromCodePoint = new String(Character.toChars(codepoint));
+        return String.format("U+%04X(%s)", codepoint, fromCodePoint);
+    }
+
+    private static String rangeToString(EntryRange range) {
+        return rangeString(range.codepoint, range.codepointEnd);
+    }
+
+    private static String rangeString(int codepoint, int codepointEnd) {
```
markusicu, line 3346: totally optional: why not overload "rangeToString"?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+                if (cp == -1) {
+                    break;
+                }
+                expected.add(codePointToString(cp));
+            }
+
+            UnicodeSetIterator usi = new UnicodeSetIterator(us);
+            StringJoiner fromUnicodeSetIterator = newResult();
+            while (usi.next()) {
+                if (usi.codepoint != UnicodeSetIterator.IS_STRING) {
```
markusicu, line 3248: nit: break when you see the first string

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+            // Build the reference test result
+            StringJoiner expected = newResult();
+            int i = 0;
+            while (true) {
+                int cp = us.charAt(i++);
+                if (cp == -1) {
+                    break;
+                }
+                expected.add(codePointToString(cp));
+            }
```
markusicu, line 3243: How about renaming expected to joiner, and after the loop you do `String expected = joiner.toString();` only once?
(same for the other test functions)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+
+            StringJoiner fromStream = newResult();
+            us.codePointStream().forEach(cp -> fromStream.add(codePointToString(cp)));
+            assertEquals("code points :: codePointStream", expected.toString(), fromStream.toString());
+        }
+    }
+
+    @Test
+    public void testIterationMethodsRanges() {
+        final UnicodeSet us = populateUnicodeSet();
```
markusicu, line 3268: Why not `for (UnicodeSet us : UNICODE_SETS_TO_ITERATE)` ?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        assertEquals("ranges :: UnicodeSetIterator", expected.toString(), fromUnicodeSetIterator.toString());
+
+        StringJoiner fromStream = newResult();
+        us.rangeStream().forEach(r -> fromStream.add(rangeToString(r)));
+        assertEquals("ranges :: rangeStream", expected.toString(), fromStream.toString());
+    }
+
+    @Test
+    public void testIterationMethodsStrings() {
+        final UnicodeSet us = populateUnicodeSet();
```
markusicu, line 3292: Why not `for (UnicodeSet us : UNICODE_SETS_TO_ITERATE)` ?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+
+        StringJoiner fromStream = newResult();
+        us.stringStream().map(UnicodeSetTest::stringToString).forEach(fromStream::add);
+        assertEquals("strings :: stringStream", expected.toString(), fromStream.toString());
+    }
+
+    @Test
+    // Iterates on strings AND on code points (converted to strings)
+    public void testIterationMethodsCodepointsAndStrings() {
+        final UnicodeSet us = populateUnicodeSet();
```
markusicu, line 3317: Why not `for (UnicodeSet us : UNICODE_SETS_TO_ITERATE)` ?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        // Build the reference test result
+        StringJoiner expected = newResult();
+        for (EntryRange r : us.ranges()) {
+            expected.add(rangeToString(r));
+        }
+
+        UnicodeSetIterator usi = new UnicodeSetIterator(us);
+        StringJoiner fromUnicodeSetIterator = newResult();
+        while (usi.nextRange()) {
+            if (usi.codepoint != UnicodeSetIterator.IS_STRING) {
```
markusicu, line 3279: break at first string

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        // Build the reference test result
+        StringJoiner expected = newResult();
+        for (String str : us.strings()) {
+            expected.add(stringToString(str));
+        }
+
+        UnicodeSetIterator usi = new UnicodeSetIterator(us);
+        StringJoiner fromUnicodeSetIterator = newResult();
+        while (usi.nextRange()) {
+            if (usi.codepoint == UnicodeSetIterator.IS_STRING) {
```
markusicu, line 3303: - start with skipToStrings(), no need to test if it's a string
- nextRange() works, but next() seems more natural

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+    public void testIterationMethodsCodepointsAndStrings() {
+        final UnicodeSet us = populateUnicodeSet();
+
+        // Build the reference test result
+        StringJoiner expect = newResult();
+        for (String str : us) {
+            expect.add(stringToString(str));
+        }
+
+        StringJoiner fromStream = newResult();
```
markusicu, line 3325: curious omission of UnicodeSetIterator here...
(however, _testing_ that old class seems redundant here since it has old test code)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3205,178 @@ public void TestPatternCodePointComplement() {
...
+        System.out.println("codePointStream.normal   : " + timeNormal);
+
+        start = System.nanoTime();
+        int sumParallel = us.codePointStream().parallel().sum();
+        long timeParallel = System.nanoTime() - start;
+        System.out.println("codePointStream.parallel : " + timeParallel);
+
+        // On my machines this is about 1.2, so it is a bit faster.
+        // Unlikely to have any practical benefit, this is more to test that the
+        // parallel stream works the same as the normal one, by comparing the sums.
```
markusicu, line 3377: Why is there any difference if trySplit() always returns null? Doesn't that make it sequential?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
mihnita: Done

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
mihnita: After implementing an optimized version of codepoint iteration it is even faster (close to 1.5)

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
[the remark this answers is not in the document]
mihnita: I've implemented a suggestion from Markus.
See if it sounds better, if not I will drop it.

mihnita:
> > Question: should I tag these new APIs as draft? Or are they OK as is?
> 
> All public & protected APIs must have an API tag. If they must be public or protected but are not intended for public use, then tag them as `@internal`. Otherwise, all new APIs must be tagged with `@draft ICU 76` as usual.

Done

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
mihnita: Done

Refactored all the code to work directly on the list and not calling `charAt`

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
mihnita: Done
Refactored to work on the list, not calling `charAt`

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
mihnita: Done

It is faster without `charAt`, it is even a bit faster than the older `UnicodeSetIterator`

Interestingly enough, I don't see a significant speed difference even with `Iterator<Integer>` (boxed values instead of primitives).
I know that Java improved boxing performance bit time, but I did not expect it to be un-noticeable.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5140,237 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return a {@link Stream} of {@link EntryRange}
+     *
+     * @draft ICU 76
+     */
+    public Stream<EntryRange> rangeStream() {
+        return StreamSupport.stream(ranges().spliterator(), false);
```
markusicu, line 5160:
```suggestion
        // Must use false to never make this parallel because the iterator always returns the same EntryRange object.
        return StreamSupport.stream(ranges().spliterator(), false);
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5191,63 +5199,70 @@ public IntStream codePointStream() {
...
+        }
+
         @Override
         public Iterator<Integer> iterator() {
-            return new CodePointsIterator();
+            return new CodePointsIterator(unicodeSet);
         }
     }
 
     private class CodePointsIterator implements Iterator<Integer> {
```
markusicu, line 5251: nit: now that CodePointIterable is singular, this should also be singular CodePointIterator

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5277,23 +5292,22 @@ private static class CodePointsSpliterator implements Spliterator.OfInt {
...
 
-        private final UnicodeSet uset;
-        private int index;
+        private final UnicodeSet unicodeSet;
+        private final CodePointsIteratorInt cpi;
 
-        CodePointsSpliterator(UnicodeSet uset) {
-            this.uset = uset;
-            this.index = 0;
+        CodePointsSpliterator(UnicodeSet unicodeSet) {
```
markusicu, line 5298:
```suggestion
        CodePointSpliterator(UnicodeSet unicodeSet) {
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5309,13 +5323,54 @@ public Spliterator.OfInt trySplit() {
...
+
+            if (cpi.hasNext()) {
+                action.accept(cpi.next());
                 return true;
             }
             return false;
         }
     }
+
+    /* This class is optimized to iterate on code points and will be used to implement both
```
markusicu, line 5335:
```suggestion
    /**
     * This class is optimized to iterate on code points and will be used to implement both
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5140,237 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+     *   <li>For speed, {@code UnicodeSet} iteration does not check for concurrent modification.<br>
+     *       Do not alter the {@code UnicodeSet} while iterating.
+     * </ul>
+     *
+     * @return a {@link Stream} of {@code String}
+     *
+     * @draft ICU 76
+     */
+    public Stream<String> stringStream() {
+        return StreamSupport.stream(strings().spliterator(), false);
```
markusicu, line 5178: this should work, right?
```suggestion
        return strings().stream();
```

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5140,237 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+                // We can still iterate on code points in the current range
+                return true;
+            }
+            return currentRange < lastRange;
+        }
+
+        public int next() {
+            if (startRange >= endRange) {
+                startRange = unicodeSet.list[currentRange++];
+                endRange = unicodeSet.list[currentRange++];
```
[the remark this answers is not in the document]
mihnita: Done.

And added tests.
It didn't happen because it was protected by `hasNext()`
And since the correct use is `while (hasNext() ) { ...next() }` it was all good.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3204,149 @@ public void TestPatternCodePointComplement() {
...
+        unicodeSet.applyPattern("\\p{sc=Ethi}");
+        unicodeSet.applyPattern("\\p{Number}");
+        // Single code point
+        unicodeSet.add('X'); // adds a code point in the ASCII range
+        unicodeSet.add('Σ'); // adds a code point in the Greek block
+        unicodeSet.add(0x1F600); // adds a code point above BMP (😀)
+        // Code point ranges and from CharSequence
+        unicodeSet.add('A', 'F'); // adds a code point range in the ASCII range
+        unicodeSet.add('α', 'ζ'); // adds a code point range in the Greek block
+        unicodeSet.add(0x1F347, 0x1F353); // adds a code point range above BMP (🍇-🍓)
```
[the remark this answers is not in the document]
mihnita: It was 0x10FFFF
Fixed, added test (for both 0x0000 and 0x10FFFF)

mihnita (commented):
Implemented round 2 of feedback.
Thank you!
M

markusicu (commented):
almost there

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
mihnita: Done now.

Sorry, there was another class CodePointsFoo, and that was Done.
Off by one error.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/lang/UnicodeSetTest.java
```diff
@@ -3198,4 +3209,156 @@ public void TestPatternCodePointComplement() {
...
+        }
+    }
+
+    @Test(expected = NoSuchElementException.class)
+    public void testMisuseIterator() {
+        UnicodeSet us = new UnicodeSet(0x10FFFE, 0x10FFFF);
+        assertEquals("Even length, last range ends at HIGH", 2, us.size());
+        Iterator<Integer> cpIter = us.codePoints().iterator();
+        cpIter.next(); // 0x10FFFE
+        cpIter.next(); // 0x10FFFF
```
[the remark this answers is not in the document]
mihnita: Done

mihnita (commented):
Done.
Thanks!

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -5127,5 +5140,220 @@ public static void setDefaultXSymbolTable(XSymbolTable xSymbolTable) {
...
+             *   > This method may return null for any reason, including emptiness, inability to split after
+             *   > traversal has commenced, data structure constraints, and efficiency considerations.
+             */
+            return null;
+        }
+
+        @Override
+        public boolean tryAdvance(IntConsumer action) {
+            if (action == null)
+                throw new NullPointerException();
```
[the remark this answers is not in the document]
markusicu: The main part of my question was whether we need this check at all; seems redundant. See above.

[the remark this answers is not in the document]
mihnita: > The main part of my question was whether we need this check at all; seems redundant. See above.

I think it is needed. 

https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#tryAdvance-java.util.function.Consumer-

> Throws:
> NullPointerException - if the specified action is null

From this I guess that it's possible to be null. 

At the time when I wrote the code I checked some existing implementation in the jdk (but I don't remember on what class) 
And it was checking for null.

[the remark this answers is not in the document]
mihnita: The behavior would be different without this test.

Think `action == null` and `cpi.hasNext() == false`

With this test the method throws.
Without it the test would return `false`.

I think that input parameter validation should take priority.

---

When implementing it initially I checked what the JDK is doing.
And it does test for null and throws explicitly:
https://github.com/openjdk/jdk17/blob/master/src/java.base/share/classes/java/util/Spliterators.java#L2145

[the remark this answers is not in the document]
markusicu: Thanks for checking!

markusicu (commented):
lgtm pse squash

markusicu (approved):

mihnita:
Thank you!
M

PR #3089, opened 2024-08-07
ICU-22848 Show the problem of infinity loop

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22848


FrankYFTang:
@aheninger could you take a look at the test case I added. It is now hang. I think it is super slow

aheninger:
Looks like the state table builder is running in exponential time on the number of trailing dots in the test pattern, ".*X..................;"

This is deep in the guts of the dragon book state table construction algorithm. I doubt there is an easy fix - the algorithm is tricky and hard to fully understand.

I recommend just leaving this as a known issue. It's not a problem in actual real break rules.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -6133,6 +6134,13 @@ void RBBITest::TestBug22636() {
     assertEquals(WHERE, ec, U_ZERO_ERROR);
 }
 
+void RBBITest::TestBug22848() {
+    if (quick || logKnownIssue("ICU-22848", "Very slow case")) { return; }
```
aheninger, line 6138: I recommend
-  reduce the number of dots in the pattern to the minimum that causes the failure (copy from bug report comment). This will dramatically speed up the test.
-  add an AssertSuccess after creating the bi. So the test will actually fail, and not just hang.
-  re-title the PR to something like "Add test for rbbi rule builder failure."

The test should fail -- when the state table max size is exceeded the ec will return a failure.

aheninger (changes requested):

PR #3090, opened 2024-08-07
ICU-22849 Fix memLeak in RBBIRuleBuilder by using LocalMemory

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22849

### DIFF
--- a/icu4c/source/common/rbbirb.cpp
+++ b/icu4c/source/common/rbbirb.cpp
@@ -182,12 +182,12 @@ RBBIDataHeader *RBBIRuleBuilder::flattenData() {
     }
 #endif
 
-    RBBIDataHeader* data = static_cast<RBBIDataHeader*>(uprv_malloc(totalSize));
-    if (data == nullptr) {
+    LocalMemory<RBBIDataHeader> data(static_cast<RBBIDataHeader*>(uprv_malloc(totalSize)));
+    if (data.isNull()) {
         *fStatus = U_MEMORY_ALLOCATION_ERROR;
         return nullptr;
     }
-    uprv_memset(data, 0, totalSize);
+    uprv_memset(data.getAlias(), 0, totalSize);
 
 
     data->fMagic            = 0xb1a0;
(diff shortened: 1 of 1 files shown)


icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -155,6 +155,7 @@ void RBBITest::runIndexedTest( int32_t index, UBool exec, const char* &name, cha
     TESTCASE_AUTO(TestBug22585);
     TESTCASE_AUTO(TestBug22602);
     TESTCASE_AUTO(TestBug22636);
+    TESTCASE_AUTO(TestBug22849);
```
aheninger, line 158: This test case shouldn't be part of this PR - it's for a different issue.

aheninger (changes requested):

icu4c/source/test/intltest/rbbitst.cpp
FrankYFTang: it is the way I saw the memory leak. I do not know what better way to show that memroy leak.

richgillam (dismissed):
Looks okay to me, but this is Andy's code, so his comments should carry the day.

icu4c/source/test/intltest/rbbitst.cpp
roubert: This test case passes also without the bugfix, so it doesn't verify whether the bugfix works or not.

That said, I don't think this change needs 3 different reviewers, so I'll remove myself.

aheninger: Poking into this test a little bit, it turns out that not only does the run time go exponential with the number of '.'s in the rule pattern, but the size of the generated state table also goes exponential.

Maybe the fuzzer's docker environment has a lowish total heap size limit, which caused the allocation failure, showing the bug. But the test does not fail in a normal development environment, and I question whether it would fail in CI testing.

The size of the allocation triggered by this test is 526696 bytes, which is no big deal, but I don't know how much other stuff is left on the heap by the rule builder.

I still recommend removing this test. It's not going to produce consistent, reproducible results.

aheninger (changes requested):

icu4c/source/test/intltest/rbbitst.cpp
FrankYFTang: ok, I remove the tests

aheninger (approved):

PR #3092, opened 2024-08-08
ICU-22898 MF2: fix various parser bugs and add more tests

catamorphism:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22898

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
@@ -613,10 +613,8 @@ public class MFParser {
             }
             keys.add(key);
         }
-        // Only want to skip whitespace if we parsed at least one key
-        if (!keys.isEmpty()) {
-            skipOptionalWhitespaces();
-        }
+        // Trailing whitespace is allowed after the message
+        skipOptionalWhitespaces();
         if (input.atEnd()) {
             checkCondition(
                     keys.isEmpty(), "After selector keys it is mandatory to have a pattern.");
--- a/icu4c/source/i18n/messageformat2_parser.cpp
+++ b/icu4c/source/i18n/messageformat2_parser.cpp
@@ -24,21 +24,22 @@ using namespace data_model;
     The `ERROR()` macro sets a syntax error in the context
     and sets the offset in `parseError` to `index`. It does not alter control flow.
 */
-#define ERROR(parseError, errorCode, index)                                                             \
+#define ERROR(errorCode)                                                                                \
     if (!errors.hasSyntaxError()) {                                                                     \
         setParseError(parseError, index);                                                               \
         errors.addSyntaxError(errorCode);                                                               \
     }
 
-// Returns true iff `index` is a valid index for the string `source`
-static bool inBounds(const UnicodeString &source, uint32_t index) {
-    return static_cast<int32_t>(index) < source.length();
-}
+#define ERROR_AT(errorCode, i)                                                                          \
+    if (!errors.hasSyntaxError()) {                                                                     \
+        setParseError(parseError, i);                                                                   \
+        errors.addSyntaxError(errorCode);                                                               \
+    }
 
 // Increments the line number and updates the "characters seen before
-// current line" count in `parseError`, iff `source[index]` is a newline
+// current line" count in `parseError`, iff `peek()` is a newline
 void Parser::maybeAdvanceLine() {
-    if (source[index] == LF) {
+    if (peek() == LF) {
         parseError.line++;
         // add 1 to index to get the number of characters seen so far
         // (including the newline)
@@ -50,10 +51,15 @@ void Parser::maybeAdvanceLine() {
     Signals an error and returns either if `parseError` already denotes an
     error, or `index` is out of bounds for the string `source`
 */
-#define CHECK_BOUNDS(source, index, parseError, errorCode)                                              \
-    if (!inBounds(source, index)) {                                                                     \
-        ERROR(parseError, errorCode, index);                                                            \
-        return;                                                                                         \
+#define CHECK_BOUNDS(errorCode)                                                            \
+    if (!inBounds()) {                                                                     \
+        ERROR(errorCode);                                                                  \
+        return;                                                                            \
+    }
+#define CHECK_BOUNDS_1(errorCode)                                                          \
+    if (!inBounds(1)) {                                                                    \
+        ERROR_AT(errorCode, index + 1);                                                    \
+        return;                                                                            \
     }
 
 // -------------------------------------
@@ -203,6 +209,13 @@ static bool isQuotedChar(UChar32 c) {
         || c == RIGHT_CURLY_BRACE;
 }
 
+static bool isEscapableChar(UChar32 c) {
+    return c == PIPE
+        || c == BACKSLASH
+        || c == LEFT_CURLY_BRACE
+        || c == RIGHT_CURLY_BRACE;
+}
+
 // Returns true iff `c` can begin a `function` nonterminal
 static bool isFunctionStart(UChar32 c) {
     switch (c) {
@@ -247,15 +260,13 @@ static bool isKeyStart(UChar32 c) {
     return (c == ASTERISK || isLiteralStart(c));
 }
 
-inline bool isDeclarationStart(const UnicodeString& source, int32_t index) {
-    int32_t len = source.length();
-    int32_t next = index + 1;
-    return (source[index] == ID_LOCAL[0]
-            && next < len
-            && source[next] == ID_LOCAL[1])
-        || (source[index] == ID_INPUT[0]
-            && next < len
-            && source[next] == ID_INPUT[1]);
+bool Parser::isDeclarationStart() {
+    return (peek() == ID_LOCAL[0]
+            && inBounds(1)
+            && peek(1) == ID_LOCAL[1])
+        || (peek() == ID_INPUT[0]
+            && inBounds(1)
+            && peek(1) == ID_INPUT[1]);
 }
 
 // -------------------------------------
@@ -375,14 +386,14 @@ option, or the optional space before an attribute.
  Unless otherwise noted in a comment, all helper functions that take
     a `source` string, an `index` unsigned int, and an `errorCode` `UErrorCode`
     have the precondition:
-      `index` < `source.length()`
+      `index` < `len()`
     and the postcondition:
-      `U_FAILURE(errorCode)` || `index < `source.length()`
+      `U_FAILURE(errorCode)` || `index < `len()`
 */
 
 /*
   No pre, no post.
-  A message may end with whitespace, so `index` may equal `source.length()` on exit.
+  A message may end with whitespace, so `index` may equal `len()` on exit.
 */
 void Parser::parseWhitespaceMaybeRequired(bool required, UErrorCode& errorCode) {
     bool sawWhitespace = false;
@@ -391,7 +402,7 @@ void Parser::parseWhitespaceMaybeRequired(bool required, UErrorCode& errorCode) 
     // or when we see a non-whitespace character.
     while (true) {
         // Check if all input has been consumed
-        if (!inBounds(source, index)) {
+        if (!inBounds()) {
             // If whitespace isn't required -- or if we saw it already --
             // then the caller is responsible for checking this case and
             // setting an error if necessary.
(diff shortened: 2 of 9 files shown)


mihnita:
Sorry, the sharing of the test files PR started when I was in vacation and somehow didn't get on my radar when I got back.

But I just found that the change broke Eclipse, and that now the json test files are packaged in the release jar files of ICU.

---

I general having Maven consume files from outside it's own folder tree is hacky and error prone.
And when it finally works, we discover that it breaks the import to various IDEs.
I think that we would want to support at least Eclipse, IntelliJ, VS Code.

We currently have that problem with the LICENSE file (I can't import icu4j in Eclipse anymore)
I'm working on that...

---

The other problem with sharing is that it forces the c/c++ and Java implementations to be 100% in sync at all times.
Having them work the same is in general a good thing, of course.

But becomes a PITA when something changes and we need to update the code.
Because (often) the C++ and Java devs might not be the same.

This PR being an example.

---

TLDR: I am tempted to keep two copies of the test data.
There is an ant task in tools/cldr/ that (copy-cldr-testdata) that copies some test data from CLDR.

These json files are probably in the same bucket: they live in CLDR, but must be tested against in ICU.

But the idea is that the ant task can copy the files in two places, icu4c and icu4j.
If code breaks, but one has time / expertise for C/C++ only, they can open a ticket against icu4j, priority zero,
but revert the test files in icu4j until that issue is fixed.

I don't know if that is a good idea or not.
But this is what we already do for other tests.

Here is a fragment of the ant script:
```xml
        <fileset id="cldrTestData" dir="${cldrDir}/common/testData">
            <!-- Add directories here to control which test data is installed. -->
            <include name="localeIdentifiers/**"/> <!-- ... -->
            <include name="personNameTest/**"/> <!-- Used in ExhaustivePersonNameTest -->
            <include name="units/**"/> <!-- Used in UnitsTest tests -->
       </fileset>

        <copy todir="${testDataDir4C}">
            <fileset refid="cldrTestData"/>
        </copy>
        <copy todir="${testDataDir4J}">
            <fileset refid="cldrTestData"/>
        </copy>
```

We copy the same test files in `icu4c/source/test/testdata/cldr` and `icu4j/main/core/src/test/resources/com/ibm/icu/dev/data/cldr`

mihnita:
Try:
```
diff -r icu4c/source/test/testdata/cldr icu4j/main/core/src/test/resources/com/ibm/icu/dev/data/cldr
```

There are 3 sub-folders there: `localeIdentifiers`, `personNameTest`, `units`
I would propose we add another one (`messageformat2`) and do the same thing.

icu4j/main/core/src/main/java/com/ibm/icu/message2/InputSource.java
```diff
@@ -62,10 +62,22 @@ int readCodePoint() {
         return c;
     }
 
-    // Backup a number of characters.
+    // Backup a number of code points.
     void backup(int amount) {
```
mihnita, line 66: This method works on code units (Java `char`), not code points.

That's because `int readCodePoint()` also works on code units.
It returns a code point, but the current offset in the input (`cursor`) is expresses in code units.
So when it finds a code point above BMP it advances with 2

Reasons:
 - When one wants to parse something from an offset, it is faster to express the offset in the same kind of units we use for storage. We store using `char`, offset in code units. Otherwise we must iterate all the way to the offset in code points. Our input buffer uses `char`
 - Easier to deal with not-properly paired surrogates. The spec does not try to enforce the surrogate correctness, and I don't think it is the job of the MF2 to do that. If I pass `".....\uDC00...{$user}!` there is no reason to not format to `".....\uDC00...John!` 

Even if there is disagreement about the reasons, and we decide to express offsets in code points, this fix is not sufficient.

We would also have to change `getPosition()`, `skip(int amount)`, `gotoPosition(int position)`
Because we either work on code points, or on code units, we should not mix and match.

icu4j/main/core/src/main/java/com/ibm/icu/message2/InputSource.java
```diff
@@ -62,10 +62,22 @@ int readCodePoint() {
...
+        int amountBackedUp = 0;
         // TODO: validate
-        cursor -= amount;
+        while (amountBackedUp < amount) {
+            backupOneCodePoint();
+            amountBackedUp++;
+        }
+    }
+
+    void backupOneCodePoint() {
```
mihnita, line 75: What happens if it is an unpaired low surrogate?
There is nothing to prevent such a thing from showing up in a message.

True, it is incorrect.
But we are not trying to reject that, we accept it.
And this code would misbehave.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/FirstReleaseTests.java
mihnita, line 1: I would rather not delete these extra tests.
They cover functionality not covered by the tests in the spec.
And that can't be covered by the spec.

They test that the implementation works end-to-end.

Some might be moved to the standard (CLDR).

But some can't, because they test that custom functions work, or that `icu:skeletons` work (namespaced),
or that various types work (`Date`, `Calendar`, `java.time`, or even long with `:dateformat`.

And that the type "magically selects" the proper function (For example "...{$exp}..." formats as a date if the `exp` is a date-like type (`Date` / `Calendar` / `java.time`)

This comment also (or mostly?) apply to the other deleted files (`FunctionsTests`, `IcuFunctionsTest`)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/ParserSmokeTest.java
```diff
@@ -29,16 +29,5 @@ public void testNullInput() throws Exception {
         MFParser.parse(null);
     }
 
-    @Test
```
mihnita, line 32: There is not much / anything left if we remove this.

So the same comment that I had for deleted files applies: do we really want to delete extra tests?
I think that the more tests we have the better.
Even some don't come from the official CLDR spec.

We can separate these tests in a different folder for cleanliness.
But not delete them, unless they are redundant.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/SelectorsWithVariousArgumentsTest.java
mihnita, line 1: Same comment about deleting tests.

mihnita (commented):
First, thank you very much for taking a stab at updating the Java implementation to the latest spec.
I was planning to do that, but it looks like trying to share the test files precipitated things somewhat :-)

I added some comments.

But if there is a decisions to treat the mf2 test files the same way we treat other CLDR test files (units, people names, locale ids) and separate the C++ / Java tests, I can take over the Java part. And in a different PR.

Or you can continue it, as you already did a big chunk of it (all?)
Your call.

Thanks again,
M

catamorphism:
> Sorry, the sharing of the test files PR started when I was in vacation and somehow didn't get on my radar when I got back.
> 
> But I just found that the change broke Eclipse, and that now the json test files are packaged in the release jar files of ICU.
> 
> I general having Maven consume files from outside it's own folder tree is hacky and error prone. And when it finally works, we discover that it breaks the import to various IDEs. I think that we would want to support at least Eclipse, IntelliJ, VS Code.
> 

Of course; is there a way to add automated testing that changes don't break IDEs? 

> We currently have that problem with the LICENSE file (I can't import icu4j in Eclipse anymore) I'm working on that...
> 
> The other problem with sharing is that it forces the c/c++ and Java implementations to be 100% in sync at all times. Having them work the same is in general a good thing, of course.
> 
> But becomes a PITA when something changes and we need to update the code. Because (often) the C++ and Java devs might not be the same.
> 
> This PR being an example.
> 
> TLDR: I am tempted to keep two copies of the test data. There is an ant task in tools/cldr/ that (copy-cldr-testdata) that copies some test data from CLDR.

I don't want to unilaterally say yes or no to this; is it something to discuss in the TC meeting, maybe? (Note: I'll be on vacation from early next week until September 2, so I won't be at the next few meetings, but I can go back and read minutes later.)

icu4j/main/core/src/main/java/com/ibm/icu/message2/InputSource.java
catamorphism: Sounds like we need a test for this! (I'll add one.)

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -712,7 +721,7 @@ private MFDataModel.Declaration getDeclaration() throws MFParseException {
         MFDataModel.Expression expression;
         switch (declName) {
             case "input":
-                skipMandatoryWhitespaces();
+                skipOptionalWhitespaces();
```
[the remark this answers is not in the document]
catamorphism: I don't think so? If it was mandatory, it would say `input-declaration = input s variable-expression` (without the square brackets.)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/FirstReleaseTests.java
catamorphism: It might be less confusing to look at https://github.com/unicode-org/icu/pull/3063 first. This PR includes that PR, plus additional changes.

In the description for https://github.com/unicode-org/icu/pull/3063 I explained that I didn't delete any tests (unless they were duplicates), but rather, moved all the JSON filenames being read into a single file, because now the schema is the same for all the tests, so the same code can be re-used to read all of the files.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/ParserSmokeTest.java
catamorphism: See https://github.com/unicode-org/icu/pull/3092#discussion_r1712036379

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -589,6 +601,7 @@ private MFDataModel.Message getComplexMessage() throws MFParseException {
             // complex-message   = *(declaration [s]) complex-body
```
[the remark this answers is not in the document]
catamorphism: Fixed

icu4j/main/core/src/main/java/com/ibm/icu/message2/InputSource.java
catamorphism: Thanks for the explanation. The issue isn't really with `backup()`, but how it's used in conjunction with `readCodePoint()` (reading a code point that happens to be a wide char, and then calling `backup(1)` to "push it back", is a bug).

In [eee547c](https://github.com/unicode-org/icu/pull/3092/commits/eee547c25ba7275947dde0a22a4d34ab10a19974) I reverted the change to `backup()` and fixed the three places I could find where `readCodePoint()` and `backup()` were being used in that way.

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFDataModelFormatter.java
```diff
@@ -432,10 +432,8 @@ private static Object resolveLiteralOrVariable(
             Map<String, Object> arguments) {
         if (value instanceof Literal) {
             String val = ((Literal) value).value;
-            Number nr = OptUtils.asNumber(val);
-            if (nr != null) {
-                return nr;
-            }
+            // "The resolution of a text or literal MUST resolve to a string."
```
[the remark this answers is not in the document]
catamorphism: I don't think the plural matcher has to be changed? This change is to address situations where a number literal isn't annotated, like:

```
{
    "src": "Format {123456789.9876} number",
    "locale": "en-IN",
    "exp": "Format 123456789.9876 number",
    "comment": "Number literals are not formatted as numbers by default"
 }
```

This change doesn't change the behavior for `.match` on something with a `:number` annotation, or at least, there aren't any tests that show that.

catamorphism:
> First, thank you very much for taking a stab at updating the Java implementation to the latest spec. I was planning to do that, but it looks like trying to share the test files precipitated things somewhat :-)
> 

Right, first it was precipitated by sharing test files; then I wrote a random test generator and decided it was only fair to run ICU4J on some of the generated tests as well as ICU4C, so that's where some of the other changes came from.

> I added some comments.
> 
> But if there is a decisions to treat the mf2 test files the same way we treat other CLDR test files (units, people names, locale ids) and separate the C++ / Java tests, I can take over the Java part. And in a different PR.

Certainly (like I said in another comment, I think that has to be discussed with the TC).

catamorphism:
I've rebased this PR, so it's now ready for review.

Note that this includes several fixes related to unsupported expression/statement (reserved) parsing. Recently these features were removed from the spec. However, since the recent spec changes haven't been integrated yet, I'd prefer to land these changes. I'm open to suggestions, though.

mihnita (approved):
Thank you,
Mihai

echeran (approved):

PR #3096, opened 2024-08-12
ICU-23004 C++ Unicode string code point iterators

markusicu:
New C++ header-only APIs for iterating over the Unicode code points in a Unicode string, and more generally over the code units from a code unit iterator. These are modern C++ equivalents of some of the long-standing C macros for iterating over [UTF-8](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/utf8_8h.html) and [UTF-16](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/utf16_8h.html). This C++ API also supports UTF-32.

FYI: UTF-8 and UTF-16 encode code points with variable-length code unit sequences. A validating iterator needs to read and check all of the code units for one code point. When a code unit sequence is ill-formed, then the returned subsequence must be a prefix of a well-formed sequence. (Except we always return at least one code unit, so that we always progress.) (UTF-32 still has validation, but sequences always have length one.)

The API can read code units from a C++ [input_iterator or forward_iterator or bidirectional_iterator](https://en.cppreference.com/w/cpp/iterator/iterator_tags). The latter includes code unit pointers like `const char *` and `const char16_t *`. There is a convenience API for `std::string_view`s.

The main class is called `UTFIterator`. Its `operator*()` returns a value serving a variety of use cases: Class `CodeUnits` provides the code point, the start of its minimal subsequence, the number of code units, and whether they are well-formed. (All functions are declared inline. An optimizing compiler will usually omit fields that are not used, and the code for computing them.)

`UTFIterator` has the API of a C++ STL iterator. It has template parameters for the code unit iterator type, for the code point type, and for how to handle ill-formed subsequences. `std::make_reverse_iterator` works for making reverse-range iterators.

The convenience class `UTFStringCodePoints` turns a `std::string_view` (of variable code unit type) into a code point iteration “range” with begin()/end()/rbegin()/rend() functions.

There are convenience functions `utfIterator()` and `utfStringCodePoints()` to simplify call sites; they deduce the code unit and base iterator types.

For each of these classes and convenience functions, there is also an “unsafe” version, just like for the C macros. The normal versions validate the code unit sequences. The “unsafe” ones assume/require that the strings/sequences are well-formed. As a result, they yield much smaller and faster code.

##### Checklist

- [x] Rename utfiter.h to utfiterator.h as proposed, unless the TC decides on a different name.
- [x] Discuss API with TC, make changes as requested.
- [x] Finish API docs.
- [x] More comprehensive testing / API & implementation coverage.
- [ ] Squash...
- [x] Required: Issue filed: ICU-23004

ALLOW_MANY_COMMITS=true


markusicu:
Hi @eggrobin I think this is worth taking another look. I rebased on recent main, made changes from our discussions, and I think this looks roughly like a reasonable validating, forward-only (so far) Unicode 16-bit-string code point iterator. It no longer tries to be clever: It no longer reads & validates the code point while iterating, and no longer stores the result in the iterator.

Plenty of TODOs and questions left, but I would appreciate feedback on the shape of what I've got so far.

markusicu:
I experimented with godbolt, and found that the compiler does its best fusing operator*() and operator++() when they both call the same implementation function. This makes operator++() look horribly inefficient, but the machine code for a regular range-based for loop from the optimizing clang 19 looks very concise.

I then also made the iterator bidirectional and added a special version for efficient rbegin() & rend() using the same principles.

The bidirectional iterator also exposes explicit but non-colloquial functions.

eggrobin:
(As noted over email, I’ll take a look on Monday when I’m back from the holidays.)

markusicu:
FYI: What I have at the moment matches my API proposal, except for a couple of things after discussion with Robin which are highlighted in the proposal doc. One more exception is that I proposed the header filename as utfiterator.h while this PR still has it as utfiter.h at the moment. I am waiting for the proposal discussion to see if names shift.

richgillam (approved):
I'm going to have to take your word for it.  My eyes cross whenever I try hard to read this.  My overall impression is that this is a LOT of code to do something that seems like it should be pretty straightforward, but that's C++ for you.

markusicu:
> I'm going to have to take your word for it.

Let me know if you want a personal tour!

> My overall impression is that this is a LOT of code to do something that seems like it should be pretty straightforward, but that's C++ for you.

Yes. Well, the UTFImpl code is very similar to the C macros (so I claim that you can ignore that).

The UTFIterator code wraps sane iteration code (where reading contents goes together with moving the cursor) into the C++ an-iterator-looks-like-a-pointer model, where `*` and `++` / `--` are separate operations. It took me a while until I convinced the compiler to not write duplicate code. That required doing the sane thing and remembering it, so the iterators have state, and const functions working on mutable fields.

eggrobin:
>> I'm going to have to take your word for it.
>
>Let me know if you want a personal tour!

I’m happy to help with that too.

> this is a LOT of code to do something that seems like it should be pretty straightforward

Part of it is, as Markus noted, working around the fact that the natural operation here is *it++.

But a lot of it is also that this is really three different generic APIs wrapped into one—in some other languages these might be independent things, here they form the hierarchy of [iterator concepts](https://en.cppreference.com/w/cpp/iterator#Iterator_concepts_.28since_C.2B.2B20.29)—:
1. something that turns an input (single-pass) iterator over code units into an input iterator over code points, _e.g._, to read code points from standard input;
2. something that turns a forward (multi-pass) iterator over code units into a forward iterator over code points;
3. something that turns a bidirectional iterator over code units into a bidirectional iterator over code points.

And then for each of those there are both validating and non-validating versions (so now that’s six different APIs!).

And then bells and whistles so that if the iterator was into a string or string view to start with, you get to look at your code point as a string view (in C++20 we would do that based on whether the iterator is an iterator into contiguous memory).

richgillam:
>Let me know if you want a personal tour!

>I’m happy to help with that too.

If you guys think it'd be helpful to have another pair of eyes on this-- rather than just reviewing each other's work-- I'm happy to be the third pair of eyes, and I agree that the best way to proceed would probably be for you to walk me through it interactively.  I'd normally suggest right after the ICU TC meeting on Thursday, but I have a meeting then.  I'm relatively open tomorrow and Friday (except for from 11 to 12 on Friday).

icu4c/source/common/unicode/utfiter.h
```diff
@@ -0,0 +1,2045 @@
...
+        if (U8_IS_SINGLE(c)) {
+            return {c, 1, true, p0, p};
+        }
+        uint8_t length = 1;
+        uint8_t t = 0;
+        if (p != limit &&
+                // fetch/validate/assemble all but last trail byte
+                (c >= 0xe0 ?
+                    (c < 0xf0 ?  // U+0800..U+FFFF except surrogates
+                        U8_LEAD3_T1_BITS[c &= 0xf] & (1 << ((t = *p) >> 5)) &&
```
roubert, line 376: So this `c &= 0xf` here modifies the value of `c` halfway through the `if`-expression!? Just checking that this really is intentional …

icu4c/source/common/unicode/utfiter.h
```diff
@@ -0,0 +1,2045 @@
...
+        }
+        if (U8_IS_TRAIL(c) && p != start) {
+            UnitIter p1 = p;
+            uint8_t b1 = *--p1;
+            if (U8_IS_LEAD(b1)) {
+                if (b1 < 0xe0) {
+                    p = p1;
+                    c = ((b1 - 0xc0) << 6) | (c & 0x3f);
+                    return {c, 2, true, p, p0};
+                } else if (b1 < 0xf0 ?
```
roubert, line 447: As the previous block ends with `return`, it doesn't really make any sense to have an `else`-clause here, this could just be a plain separate `if`-statement.

eggrobin:
For example, see rangeLoop8 in https://godbolt.org/z/bc6Y9v7Yv and compare what it becomes if you add inline on operator* on line 532 and readAndInc on line 245.

icu4c/source/common/unicode/utfiter.h
```diff
@@ -0,0 +1,2041 @@
...
+
+/**
+ * \file
+ * \brief C++ header-only API: C++ iterators over Unicode strings (=UTF-8/16/32 if well-formed).
+ */
+
+#ifndef U_HIDE_DRAFT_API
+
+// Some defined behaviors for handling ill-formed Unicode strings.
+typedef enum UTFIllFormedBehavior {
```
roubert, line 96: Does this `typedef` really serve any purpose? Why not just a plain simple `enum` declaration:

```c++
enum UTFIllFormedBehavior {
    // …
};
```

Possibly with the underlying type specified, to eliminate any wasted storage:

```c++
enum UTFIllFormedBehavior : uint8_t {
    // …
};
```

eggrobin:
> That should not be possible, the removed inline specifiers should be perfectly redundant.

They are, from a standard standpoint: this does not change the legality nor the semantics of the program. But they affect inlining (try it in that godbolt example).

> If they really do have some effect, that needs to be clearly documented in the code, to prevent someone else to later get the same idea that I got now and delete them.

Good idea. Alternatively, I think we might want to make them force-inline so we don’t depend on the compiler taking the hint from the inline keyword (but then we need macros to do that, since MSVC and clang have different syntaxes).

markusicu:
> > That should not be possible, the removed inline specifiers should be perfectly redundant.
> 
> They are, from a standard standpoint: this does not change the legality nor the semantics of the program. But they affect inlining (try it in that godbolt example).

@roubert please revert your un-inlining commit.
And please don't just commit to this PR without discussion.

> > If they really do have some effect, that needs to be clearly documented in the code, to prevent someone else to later get the same idea that I got now and delete them.
> 
> Good idea. Alternatively, I think we might want to make them force-inline so we don’t depend on the compiler taking the hint from the inline keyword (but then we need macros to do that, since MSVC and clang have different syntaxes).

That's probably ok, but some of the validating UTF-8 code is pretty large.

icu4c/source/common/unicode/utfiter.h
markusicu: It is. As noted in comments, these implementations are nearly the same as the C macros, and I took a lot of time optimizing those some years ago.

Compare with https://github.com/unicode-org/icu/blob/release-77-1/icu4c/source/common/unicode/utf8.h#L389

icu4c/source/common/unicode/utfiter.h
markusicu: This is how we define enum types in ICU that work for both C and C++. Note also that it is outside any namespace.

roubert:
> @roubert please revert your un-inlining commit.

Done.

> And please don't just commit to this PR without discussion.

What would you have preferred instead? To me a simple commit, trivial to drop or revert, seems the by far easiest to deal with for everyone involved.

icu4c/source/common/unicode/utfiter.h
markusicu: Possible, but it doesn't really matter, and I think the explicit else helps understand the code.

This here is ported from https://github.com/unicode-org/icu/blob/release-77-1/icu4c/source/common/utf_impl.cpp#L243

icu4c/source/common/unicode/utfiter.h
eggrobin: Do we need this to work in C, considering that it is used only in some very ++ C++ ?

markusicu:
> > @roubert please revert your un-inlining commit.
> 
> Done.

Thanks!

> > And please don't just commit to this PR without discussion.
> 
> What would you have preferred instead? To me a simple commit, trivial to drop or revert, seems the by far easiest to deal with for everyone involved.

Please write a comment as usual. If you want to offer to do the work, that's great, but let's agree on it first.
In this case, we could have told you that we had added the inlines for a reason.

roubert:
> That should not be possible, the removed inline specifiers should be perfectly redundant.

This has been bugging me a lot. So I've looked closer at it and I'm not able to reproduce any difference in compiler output caused by the presence or absence of the `inline` keyword where it ought to be redundant here. I've put together a few test files and as far as I can see there really is no difference at all (which is exactly what I had originally expected):

[utfiter-inline-mystery.tar.gz](https://github.com/user-attachments/files/19393991/utfiter-inline-mystery.tar.gz)

Am I missing something here, or did maybe the idea that the `inline` keyword made some difference here actually happen to come from some other difference?

eggrobin:
@roubert see my earlier comment: 
> For example, see rangeLoop8 in https://godbolt.org/z/bc6Y9v7Yv and compare what it becomes if you add inline on operator* on line 532 and readAndInc on line 245.

roubert:
> Before: https://godbolt.org/z/bc6Y9v7Yv
> After: https://godbolt.org/z/64zo8f4PP

The problem is just that I can't reproduce that, when I do it the resulting assembly code is identical both with and without the redundant `inline` keywords, also in Godbolt:

https://godbolt.org/z/4azeYYMhb

markusicu:
> That said, `inline` is too weak a hint for me to be comfortable about the portability of any of this, so I would just want to [force inline](https://github.com/mockingbirdnest/Principia/blob/190d9f25c66e96528f0e560b57e8b6b5b8a16ae0/base/macros.hpp#L99-L110) the whole lot.

Why does ICC drop the other specifiers?
```
#elif PRINCIPIA_COMPILER_ICC
#  define FORCE_INLINE(specifiers) __attribute__((always_inline))
```

Why do some usages have specifiers=inline? Isn't that redundant?
```
template<FMAPolicy fma_policy>
FORCE_INLINE(inline)
Value SinImplementation(DoublePrecision<Argument> const θ_reduced) {
```

----
[U_FORCE_INLINE()](https://github.com/unicode-org/icu/pull/3096/commits/c869cbde95dd36d60ad07c90c205973b74e13c64) added

markusicu:
@eggrobin @richgillam I think I might be done. 🎉 

I have added a lot of test code and am out of ideas for what else we need.
Not much has changed in the implementation since our tour.

Both the new header file and the test code still contain the experimental/sample code. I will remove those after you give me a green light. (I will then of course need a re-rubber-stamp.)

**Please review again.**

You could go commit by commit since our tour... but there are a lot of commits, and some back & forth of adding code and then deduplicating/simplifying.
Probably best to look at the whole delta.
Let me know if you want another tour.

I also intend to make a copy of my branch before squashing. Try to remind me of that...

markusicu:
@roubert in case you have time: I added you back as a reviewer as well.
The implementation code has not changed a lot since you last looked.
Most of the recent changes are in the test code.
And you are more familiar with modern C++ than Rich.

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -0,0 +1,2502 @@
...
+    using difference_type = typename std::iterator_traits<UnitIter>::difference_type;
+    /** C++ iterator boilerplate @internal */
+    using iterator_category = std::conditional_t<
+        std::is_base_of_v<
+            std::bidirectional_iterator_tag,
+            typename std::iterator_traits<UnitIter>::iterator_category>,
+        std::bidirectional_iterator_tag,
+        std::forward_iterator_tag>;
+
+    // TODO: Should we enable this only for a bidirectional_iterator?
```
eggrobin, line 1039: The documentation seems to answer this TODO in the affirmative, and the implementation in the negative.

I think the documentation is right.

eggrobin: Actually the documentation talks about forward_iterator and the TODO about bidirectional_iterator.

Either way, as currently implemented, I think it does not actually require a forward_iterator (pretty sure you could construct it with an istreambuf_iterator), and that seems iffy.

markusicu: This TODO is only about the following three-parameter constructor. For a forward_iterator, providing a start sentinel is unnecessary but harmless.

I am not sure whether enable_if_t even works on a constructor. It resolves to a type (void or second tparam), and a constructor has no return type.

Note that when the UnitIter is only an input_iterator, then the following template specialization is used, and that one does not have this constructor signature.

Note also that I hide that specialization from doxygen, so the API docs above need to speak for both.

markusicu: I think we can resolve this TODO at a later time. I will make a note of it.

eggrobin: > Note that when the UnitIter is only an input_iterator, then the following template specialization is used, and that one does not have this constructor signature.

Ah yes, I forgot about that. Nevermind.

markusicu:
The ICU-TC today approved my API amendment.
- iterator default constructors
- switching each of the string-code-points factory functions from tparam StringView to 5 string_view overloads

I would really like to move forward here and get this work merged -- and try to use it for real.
If there is no severe feedback, then I want to remove the experimental & sample code, make a backup, and ask for approval, even if it's a rubber stamp.
I am willing to make changes later for good feedback.
I already have a short list of things to add or try.

richgillam (approved):
I'll be honest: My eyes started to cross midway through the unit tests.  But I think the implementation code looks good (as far as I could tell, anyway).

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -0,0 +1,1298 @@
...
+        }
+        }
+    }
+    longGood8.codePoints = longGood16.codePoints = longGood32.codePoints = longGood32.str;
+}
+
+template<TestMode mode, UTFIllFormedBehavior behavior, IterType type, typename Unit, typename Units>
+void UTFIteratorTest::checkUnits(
+        const Units &units, std::basic_string_view<Unit> part, UChar32 expectedCP) {
+    // printf("U+%04lx\n", (long)units.codePoint());
```
eggrobin, line 1224: Are the commented-out printfs meant to be here (and in the following function)?

markusicu: removing. cleanup commit to follow.

eggrobin (approved):

markusicu:
@eggrobin @richgillam I just committed my cleanup. I think it's ready to go, so please re-approve.

As for backup, Robin suggests that I squash-and-merge this PR, temporarily allowing that and being careful to prefix the commit message as required. That will avoid confusion with duplicate PRs while keeping the pre-squash-n-merge commit history in place.

PR #3098, opened 2024-08-13
ICU-22854 Implement subdivision validation

FrankYFTang:
Validate the value of rg and sd before considering the region part per UTS35.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22854


richgillam (commented):
I haven't looked at this in detail yet, but I want to bring up a couple of concerns based on pain I've suffered in the past.

1) We now have a new set of data tables that we're reading out of resource bundles, transforming into a C++-friendly format, and leaving sitting around in application memory until the process terminates.  We're already doing way too dang much of this in ICU4C, and my team gets yelled at for this kind of thing all the time.  When you have ICU included by practically every process on the system, and all of them have their own copies of the same 200K (or whatever) of cached static data sitting around in their heaps, you're doing something wrong.  We need to come up with a better way.

2) This has uncomfortable echos of the parent-locale data.  That data lived in a resource bundle, and yet I needed to have it live while I was loading other resource bundles, and you can't load a resource bundle while you're in the middle of loading another resource bundle.  I had to have a separate copy of the parent locale data in static space to avoid this problem.  I don't know, but I'm concerned that at some point rg-subtag handling is going to become part of resource bundle lookup and that we're going to have a similar problem here.

FrankYFTang:
Dear Rich:
Thanks for your early feedback. Very interesting points that I have not think about. 

> I haven't looked at this in detail yet, but I want to bring up a couple of concerns based on pain I've suffered in the past.
> 
> 1. We now have a new set of data tables that we're reading out of resource bundles, transforming into a C++-friendly format, and leaving sitting around in application memory until the process terminates.  We're already doing way too dang much of this in ICU4C, and my team gets yelled at for this kind of thing all the time.  When you have ICU included by practically every process on the system, and all of them have their own copies of the same 200K (or whatever) of cached static data sitting around in their heaps, you're doing something wrong.  We need to come up with a better way.

so... if reading inside heap is not a good idea, where will be the right place? Would precompiled them into memory as part of the static data inside the code make it better or worst?

> 2. This has uncomfortable echos of the parent-locale data.  That data lived in a resource bundle, and yet I needed to have it live while I was loading other resource bundles, and you can't load a resource bundle while you're in the middle of loading another resource bundle.  I had to have a separate copy of the parent locale data in static space to avoid this problem.  I don't know, but I'm concerned that at some point rg-subtag handling is going to become part of resource bundle lookup and that we're going to have a similar problem here.

That is a good point. so.... should this information compiled into binary and memory in instead?

richgillam:
> so... if reading inside heap is not a good idea, where will be the right place? Would precompiled them into memory as part of the static data inside the code make it better or worst?

There may be better approaches, but the two I happen to be able to think of are 1) to compile the subdivision-validity data into C++ source code (probably in an include file) that defines some number of static variables that can live in the code segment and 2) to have the data in `icuXXl.dat`or some other file that's memory-mapped, and where it's structured in such a way that you don't have to copy it out of the `.dat` file and into some other in-memory data structure.

The problem obviously goes way beyond the new stuff you're adding-- we've been doing things this way forever and not really paying attention to the runtime impact.  It begs for a global solution.  I'm not asking you to come up with that global solution as part of this PR; since we're feeling the pain the most, I'm thinking maybe the Apple team can tackle it over the next year or two.  (Although if you really _want_ to do this, I wouldn't stand in your way.  😄 ) But if there's a reasonably cheap way to avoid making the problem worse, I'd like to see that.

An alternative approach would be what I suggested in comments on the ticket itself: do we really _need_ to validate the subdivision code here, or is it enough to ignore it and just look at the region code?  Are we actually _using_ the subdivision code?  That sidesteps the whole memory-usage issue completely.

I used option 1 above for the parent-locale data, so that might be the approach you want to look at for the subdivision data, if we really need it during resource-loading time.  I can dredge up the PR if you need me to.

I should probably also mention that we used to do a lot more with generating source files in the build process and have tried to get away from that, so maybe my parent-locale thing _isn't_ something we want to be doing more of.  But maybe it'll get us through until we can take a decent whack at the bigger problem.

macchiati:
> An alternative approach would be what I suggested in comments on the
ticket itself: do we really *need* to validate the subdivision code here,
or is it enough to ignore it and just look at the region code? Are we
actually *using* the subdivision code? That sidesteps the whole
memory-usage issue completely.

I had a suggestion along these lines in the ticket.

FrankYFTang:
@richgillam @markusicu @macchiati Please take another look. My newest implementation has no dependency on loading resource for validation but depends on a 22 32 int bitmap and a small internal utility class. I also add unit test , which will load the resource during the test, to ensure this bitmap sync with the resource and generate updated values if later CLDR change the resource and add additional code. The same is change to the C++ and Java. I origionally want to use pre-existing code to do this, but found the current pre-existing code is buggy and may need more complicate solution. Also that approach will still hit the same issue Rich was concerning about - loading resource during loading resource. But my current approach won't

richgillam (approved):
I like this, but I do have a few thoughts.

1. It took me a while to grok what the `RegionValidMap` is doing.  The thing that's interesting is that you defined it as a mutable object.  It seems like at runtime, all you need is the static global table and the `isSet()` and `value()` functions.  It's mutable so that you can use the unit tests to figure out what the value _should_ be if we're missing something we want.  That's kind of cool, but I'd rather have that logic in the unit tests.  I'm worried about the cost of the `memcpy()` and the copy of the table if somebody's coming through here a lot.
2. The table is nice and small and probably very quick too, but it's opaque.  You might want to add some documentation to help the reader understand what you're doing.  Also, how much do you lose if you just make this an array of strings (or one big string containing all the codes) and linear- (or binary-) search it?  That'd be simpler and clearer and the list isn't very big.  (On the other hand, this _is_ pretty nice, so maybe I'm wrong...)
3. You pointed out some bugs and other issues in the `Region` object.  Can this code be brought to bear to improve the `Region` object at all?
4. Now that I think about it, I guess I'm a little unclear what the process is for updating the static table as we add new valid region codes.  Can you talk about that a little?  I was thinking maybe it happened automagically as you ran the unit tests, but that doesn't seem to be what the unit test does, so now I'm a little confused.

FrankYFTang:
@richgillam 
Please take another look. I split the implementation and make the code inside the library only for checking and make the object immutable. And move the part needed for the test only into the test code. 

 bugs and other issues in the Region object is more complicated. The involved issue could be more complex than dealing with subdivision so I intend to address them separately in https://github.com/unicode-org/icu/pull/3113 after I discuss with you and Mark about what that class should behave. While I solve that issue, I may consider reuse some implementation here w/ or w/o modification. But I think it is too early to talk about that. That API is more public and may have more backward compatability issue we need to consider than this one and I don't want to tight them together to make progress. 


> how much do you lose if you just make this an array of strings (or one big string containing all the codes) and linear- (or binary-) search it? That'd be simpler and clearer and the list isn't very big

Well, you stated on the top of this review:
>When you have ICU included by practically every process on the system, and all of them have their own copies of the same 200K (or whatever) of cached static data sitting around in their heaps, you're doing something wrong. We need to come up with a better way.

so here I present you an implementation with several simple function plus 88 bytes of data total. I assume with the additional code size, it probably increase run time by 120 -150 bytes? (assuming 32-62 bytes of compiled code) not sure that is a fair estimate.

richgillam (approved):
I think this looks great.  No notes.

richgillam:
> Please take another look. I split the implementation and make the code inside the library only for checking and make the object immutable. And move the part needed for the test only into the test code.

Looks great!  Thanks for the changes.

> bugs and other issues in the Region object is more complicated. The involved issue could be more complex than dealing with subdivision so I intend to address them separately in #3113 after I discuss with you and Mark about what that class should behave. While I solve that issue, I may consider reuse some implementation here w/ or w/o modification. But I think it is too early to talk about that. That API is more public and may have more backward compatability issue we need to consider than this one and I don't want to tight them together to make progress.

Yeah, that's perfectly reasonable.

> > how much do you lose if you just make this an array of strings (or one big string containing all the codes) and linear- (or binary-) search it? That'd be simpler and clearer and the list isn't very big
> 
> Well, you stated on the top of this review:
> 
> > When you have ICU included by practically every process on the system, and all of them have their own copies of the same 200K (or whatever) of cached static data sitting around in their heaps, you're doing something wrong. We need to come up with a better way.
> 
> so here I present you an implementation with several simple function plus 88 bytes of data total. I assume with the additional code size, it probably increase run time by 120 -150 bytes? (assuming 32-62 bytes of compiled code) not sure that is a fair estimate.

This looks great, and you certainly can't argue with the size.  Forget I said anything.  :-)

FrankYFTang:
This PR implement what I *think* Mark @macchiati  is going to put into UTS35 based my understanding from his comments in the bug. But after he land that UTS35 spec change, I may need to take a look at what actually wrote and ensure that this implementation agree w/ that.

PR #3100, opened 2024-08-13
ICU-22696 Avoid unnecessary copies of already NUL terminated strings

roubert:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22696

### DIFF
--- a/icu4c/source/common/ulocale.cpp
+++ b/icu4c/source/common/ulocale.cpp
@@ -21,7 +21,10 @@ U_NAMESPACE_USE
 ULocale*
 ulocale_openForLocaleID(const char* localeID, int32_t length, UErrorCode* err) {
     if (U_FAILURE(*err)) { return nullptr; }
-    CharString str(length < 0 ? StringPiece(localeID) : StringPiece(localeID, length), *err);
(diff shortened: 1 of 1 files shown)


icu4c/source/common/ulocale.cpp
```diff
@@ -21,7 +21,10 @@ U_NAMESPACE_USE
 ULocale*
 ulocale_openForLocaleID(const char* localeID, int32_t length, UErrorCode* err) {
     if (U_FAILURE(*err)) { return nullptr; }
-    CharString str(length < 0 ? StringPiece(localeID) : StringPiece(localeID, length), *err);
+    if (length < 0 || localeID[length] == '\0') {
```
markusicu, line 24: When we are given a non-negative length, we should not read beyond it, not even to check whether there is a NUL there.

roubert: Hm, there's already existing code that does that elsewhere in this code base, but sure, it might be a good idea to not add more of that. Done.

markusicu (approved):

PR #3101, opened 2024-08-13
ICU-22723 cldr 46 alpha0 to main

DraganBesevic:
##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22723

Alpha 0 integration for CLDR 46. 

Several unit test for currency display names started failing during this integration. 

One problem was related to inheritance, and it was exposed by unit test for Slovenian. 
The issue is that getCurrencyLongNameData is iterating over the CurrencyUnitPatterns is a locale to determine valid plural cases, 
but with inheritance there are actually no entries for this in most locales. The fix was to populate missing entries with the plural rules from locale.

The other problem still exists and it's log as a known issue (https://unicode-org.atlassian.net/browse/ICU-22856). 
Currency display name for locales which has no plural information and no fallback returns null now. It is not clear what intend is here.

Changes in unit tests

1. Updated the date formatting unit tests for `ak` that doesn't fallback to root any longer  
2. Updated the hour cycle unit tests for en_GB and en_AU to account for adding a comma to GyMMMEEEEd formatting
3. Known issue: `ak` use chars not in exemplar set for month names ()
4. Updated the default calendar unit tests for SA that was changed from islamic-umalqura to gregorian
5. Remove entries for deprecated time zoned from unit test for aliases EST, MST, HST (https://unicode-org.atlassian.net/browse/CLDR-17111)
6. Update unit test for short time formats for fi, using numeric months now

ALLOW_MANY_COMMITS=true


icu4c/source/i18n/number_longnames.cpp
```diff
@@ -519,10 +522,35 @@ void getCurrencyLongNameData(const Locale &locale, const CurrencyUnit &currency,
...
     if (U_FAILURE(status)) { return; }
     ures_getAllChildrenWithFallback(unitsBundle.getAlias(), "CurrencyUnitPatterns", sink, status);
     if (U_FAILURE(status)) { return; }
+    // Here the outArray[] entries are filled in with any CurrencyUnitPatterns data for locale,
+    // or if there is no CurrencyUnitPatterns data for locale since the patterns all inherited
+    // from the "other" pattern in root (which is true for many locales in CLDR 46), then only
+    // the "other" entry has a currency pattern. So now what we do is: For all valid plural keywords
+    // for the locale, if the corresponding outArray[] entry is bogus, fill it in from the "other"
+    // entry. In the longer run, clients of this should instead consider using CurrencyPluralInfo
+    // (see i18n/unicode/currpinf.h).
```
richgillam, line 536: I'm sorry; I didn't understand this at all.

icu4c/source/test/cintltst/creststn.c
```diff
@@ -2185,7 +2185,7 @@ static void TestFallback(void)
         UResourceBundle* myResB = ures_open(NULL,"no_NO_NY",&err);
         UResourceBundle* resLocID = ures_getByKey(myResB, "Version", NULL, &err);
         const UChar* version = NULL;
-        static const UChar versionStr[] = u"45"; // 45 in nn_NO or in a parent bundle/root
+        static const UChar versionStr[] = u"46"; // 45 in nn_NO or in a parent bundle/root
```
richgillam, line 2188: You probably want to update the comment here too.  Or if not, can you explain what's going on here?

icu4c/source/test/intltest/numfmtst.cpp
```diff
@@ -2108,7 +2108,7 @@ void NumberFormatTest::TestCurrencyNames() {
     assertTrue("ucurr_getName (en_US default)",
                     U_USING_DEFAULT_WARNING == ec || U_USING_FALLBACK_WARNING == ec, true);
 
-    ucurr_getName(CAD, "ti", UCURR_LONG_NAME, &isChoiceFormat,
+    ucurr_getName(CAD, "to", UCURR_LONG_NAME, &isChoiceFormat,
                             &len, &ec);
     assertTrue("ucurr_getName (ti default)",
```
richgillam, line 2113: Probably want to change the error message here.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/MeasureUnitTest.java
```diff
@@ -2872,6 +2872,205 @@ public void TestCompatible74() {
...
+                MeasureUnit.MILLILITER,
+                MeasureUnit.PINCH,
+                MeasureUnit.PINT,
+                MeasureUnit.PINT_METRIC,
+                MeasureUnit.QUART,
+                MeasureUnit.QUART_IMPERIAL,
+                MeasureUnit.TABLESPOON,
+                MeasureUnit.TEASPOON,
+        };
+        assertEquals("",  193, units.length);
```
richgillam, line 3071: What's this test doing?  At first glance it looks like it'd just always pass.

icu4j/main/core/src/main/java/com/ibm/icu/impl/number/LongNameHandler.java
```diff
@@ -440,6 +441,22 @@ private static void getCurrencyLongNameData(ULocale locale, Currency currency, S
...
+        PluralRules prules = PluralRules.forLocale(locale);
+        Set<String> keywords = prules.getKeywords();
+        String otherPattern = data.get("other");
+        if (keywords != null && otherPattern!= null) {
+            for (String keyword: keywords) {
+                if (!keyword.equals("other") && !data.containsKey(keyword)) {
+                    data.put(keyword, otherPattern);
+                }
+            }
+        }
```
richgillam, line 459: I'm still not sure what's going on here, but this code looks different from (and shorter than) the corresponding code in ICU4C above.  Are they really doing the same thing?

richgillam (approved):
I'm not completely following the currency-plural stuff.  Everything else looks good, I think, but I did have a few nits.

icu4j/main/core/src/main/java/com/ibm/icu/impl/number/LongNameHandler.java
DraganBesevic: It is really the same. The difference is, ICU4C is using an array, which has the bogus entries for the values that are not present. Those bogus entries are causing ICU4C to bail out. In ICU4J, a map is used instead of an array, so instead of updating an entry at an index, we just add a new entry with the key of "one" "two" "few" etc

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/MeasureUnitTest.java
DraganBesevic: Eh. That's just a compatibility test for units. You will notice there a lot of them before, in form of TestCompatibleZZ, where ZZ is a release number

icu4j/main/core/src/main/java/com/ibm/icu/impl/number/LongNameHandler.java
richgillam: Okay, I'll take your word for it.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/MeasureUnitTest.java
richgillam: Fair enough.

icu4c/source/test/cintltst/creststn.c
DraganBesevic: Duh! Will do!

icu4c/source/test/intltest/numfmtst.cpp
DraganBesevic: Yup. Done.

richgillam (approved):
Thanks for the fixes.  Still LGTM.

icu4c/source/i18n/number_longnames.cpp
pedberg-icu: > I'm sorry; I didn't understand this at all.

The code use to iterate through the currency unit patterns to populate the set of patterns needed for all of the currency plural cases in a locale. Those currency unit patterns mostly duplicated the root "other" case which they ultimately inherit from; so now CLDR does not bother providing them if they do. So we cannot iterate over them to get the set of currency patterns for the plural cases. Instead we use PluralRules to determine what the valid plural cases are for the locale, then flesh out the currency unit patterns by copying from the "other" entry any that are missing.

pedberg-icu (approved):

icu4c/source/i18n/number_longnames.cpp
richgillam: Thank you; that helps a lot.

PR #3104, opened 2024-08-15
ICU-22861 std::max deduced conflicting types in unifiedcache.cpp

EstherWx:
Declare the type 'int32_t' for std::max to fix error: no matching function for call to 'max(int, int32_t)'

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22861


eggrobin:
> no matching function for call to 'max(int, int32_t)'

So you are on some system where int and int32_t are different types?
Looking at https://en.cppreference.com/w/cpp/language/types#Data_models, that seems quite rare.
Are you targeting the Win16 API?

eggrobin (approved):

EstherWx:
> So you are on some system where int and int32_t are different types? Are you targeting the Win16 API?
I'm trying to use collation on ARM. 'int32_t' is alias of 'sigined long' in my machine and the compiler is arm-none-eabi-g++. std::max needs to be given two parameters of the exact same type. 'int' and 'long' are two different types.

PR #3106, opened 2024-08-15
ICU-22843 Enable constructing UnicodeString from literal in fixed time

roubert:
When passing a string literal to any of the legacy constructors that take just a plain pointer to a UTF-16 string it becomes necessary to iterate through the string to find its length, even though this length was known to the compiler (which just has no way of passing it on to the constructor).

But when calling the new templated string view constructor instead it becomes possible for the compiler to use the known length of a string literal to directly create a string view of the correct size and pass this on to the constructor.

By replacing the legacy constructors with the new constructor this is made the default behaviour.

Because non-templated functions take priority it's necessary to replace the legacy constructors or else the new constructor would never be called for any data type that can implicitly decay into a pointer (and this includes string literals). For the time being, this is gated on `U_HIDE_DRAFT_API` which will select either the old or the new constructors and even though the old constructors technically get deleted the new constructor is made possible to call in identical ways so that no changes to any call sites are necessary.

There is however a snag there and that's an old promise that `nullptr` is valid input and identical to the default constructor (ie. the empty string) but this is explicitly _not_ valid input for string views, so this requires a backward compatibility shim to catch any  `nullptr` before any attempt is made to use it to create a string view (which would crash most implementations of the standard library).

##### Checklist

### DIFF
--- a/icu4c/source/common/unicode/char16ptr.h
+++ b/icu4c/source/common/unicode/char16ptr.h
@@ -309,13 +309,19 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
 
 #ifndef U_FORCE_HIDE_INTERNAL_API
 /**
- * Is T convertible to a std::u16string_view or to a 16-bit std::wstring_view?
+ * Is T convertible to a std::u16string_view or some other 16-bit string view?
  * @internal
  */
 template<typename T>
 constexpr bool ConvertibleToU16StringView =
-    std::is_convertible_v<T, std::u16string_view> ||
-    (U_SIZEOF_WCHAR_T==2 && std::is_convertible_v<T, std::wstring_view>);
+    std::is_convertible_v<T, std::u16string_view>
+#if !U_CHAR16_IS_TYPEDEF
+    || std::is_convertible_v<T, std::basic_string_view<uint16_t>>
+#endif
+#if U_SIZEOF_WCHAR_T==2
+    || std::is_convertible_v<T, std::wstring_view>
+#endif
+    ;
 
 namespace internal {
 /**
@@ -323,6 +329,17 @@ namespace internal {
  * @internal
  */
 inline std::u16string_view toU16StringView(std::u16string_view sv) { return sv; }
+
+#if !U_CHAR16_IS_TYPEDEF
+/**
+ * Basically undefined behavior but sometimes necessary conversion
+ * from std::basic_string_view<uint16_t> to std::u16string_view.
+ * @internal
+ */
+inline std::u16string_view toU16StringView(std::basic_string_view<uint16_t> sv) {
+    return { ConstChar16Ptr(sv.data()), sv.length() };
+}
+#endif
 
 #if U_SIZEOF_WCHAR_T==2
 /**
@@ -334,6 +351,29 @@ inline std::u16string_view toU16StringView(std::wstring_view sv) {
     return { ConstChar16Ptr(sv.data()), sv.length() };
 }
 #endif
+
+/**
+ * Pass-through overload.
+ * @internal
+ */
+template <typename T,
+          typename = typename std::enable_if_t<!std::is_pointer_v<std::remove_reference_t<T>>>>
+inline std::u16string_view toU16StringViewNullable(const T& text) {
+    return toU16StringView(text);
+}
+
+/**
+ * In case of nullptr, return an empty view.
+ * @internal
+ */
+template <typename T,
+          typename = typename std::enable_if_t<std::is_pointer_v<std::remove_reference_t<T>>>,
+          typename = void>
+inline std::u16string_view toU16StringViewNullable(const T& text) {
+    if (text == nullptr) return {};  // For backward compatibility.
+    return toU16StringView(text);
+}
+
 }  // internal
 #endif  // U_FORCE_HIDE_INTERNAL_API
 
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -105,13 +105,11 @@ class UnicodeStringAppendable;  // unicode/appendable.h
  * this macro was provided for portability and efficiency when
  * initializing UnicodeStrings from literals.
  *
- * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+ * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
  * length determination:
  * \code
- * #include &lt;string_view&gt;
- * using namespace std::string_view_literals;
- * UnicodeString str(u"literal"sv);
- * if (str == u"other literal"sv) { ... }
+ * UnicodeString str(u"literal");
+ * if (str == u"other literal") { ... }
  * \endcode
  *
  * The string parameter must be a C string literal.
@@ -335,13 +333,11 @@ public:
    * which is, or which is implicitly convertible to,
    * a std::u16string_view or (if U_SIZEOF_WCHAR_T==2) std::wstring_view.
    *
-   * For performance, you can use std::u16string_view literals with compile-time
+   * For performance, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
    * UnicodeString str = ...;
-   * if (str == u"literal"sv) { ... }
+   * if (str == u"literal") { ... }
    * \endcode
    * @param text The string view to compare to this string.
    * @return true if `text` contains the same characters as this one, false otherwise.
@@ -3080,6 +3076,7 @@ public:
    */
   UNISTR_FROM_CHAR_EXPLICIT UnicodeString(UChar32 ch);
 
+#ifdef U_HIDE_DRAFT_API
   /**
    * char16_t* constructor.
    *
@@ -3088,20 +3085,19 @@ public:
    * on the compiler command line or similar.
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text The characters to place in the UnicodeString.  `text`
    * must be NUL (U+0000) terminated.
    * @stable ICU 2.0
    */
-  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const char16_t *text);
+  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const char16_t *text) :
+      UnicodeString(text, -1) {}
 
 #if !U_CHAR16_IS_TYPEDEF
   /**
@@ -3113,20 +3109,18 @@ public:
    * on the compiler command line or similar.
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text NUL-terminated UTF-16 string
    * @stable ICU 59
    */
   UNISTR_FROM_STRING_EXPLICIT UnicodeString(const uint16_t *text) :
-      UnicodeString(ConstChar16Ptr(text)) {}
+      UnicodeString(ConstChar16Ptr(text), -1) {}
 #endif
 
 #if U_SIZEOF_WCHAR_T==2 || defined(U_IN_DOXYGEN)
@@ -3140,21 +3134,20 @@ public:
    * on the compiler command line or similar.
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text NUL-terminated UTF-16 string
    * @stable ICU 59
    */
   UNISTR_FROM_STRING_EXPLICIT UnicodeString(const wchar_t *text) :
-      UnicodeString(ConstChar16Ptr(text)) {}
+      UnicodeString(ConstChar16Ptr(text), -1) {}
 #endif
+#endif  // U_HIDE_DRAFT_API
 
   /**
    * nullptr_t constructor.
@@ -3172,13 +3165,11 @@ public:
    * char16_t* constructor.
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text The characters to place in the UnicodeString.
@@ -3195,13 +3186,11 @@ public:
    * Delegates to UnicodeString(const char16_t *, int32_t).
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text UTF-16 string
@@ -3219,13 +3208,11 @@ public:
    * Delegates to UnicodeString(const char16_t *, int32_t).
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param text UTF-16 string
@@ -3259,9 +3246,9 @@ public:
    * @draft ICU 76
    */
   template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
-  explicit inline UnicodeString(const S &text) {
+  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const S &text) {
     fUnion.fFields.fLengthAndFlags = kShortString;
-    doAppend(internal::toU16StringView(text));
+    doAppend(internal::toU16StringViewNullable(text));
   }
 #endif  // U_HIDE_DRAFT_API
 
@@ -3280,13 +3267,11 @@ public:
    * so that both strings then alias the same readonly-text.
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString alias = UnicodeString::readOnlyAlias(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString alias = UnicodeString::readOnlyAlias(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param isTerminated specifies if `text` is `NUL`-terminated.
@@ -3369,13 +3354,11 @@ public:
    * the constructor that takes a US_INV (for its enum EInvariant).
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * It is recommended to mark this constructor "explicit" by
@@ -3485,13 +3468,11 @@ public:
    * \endcode
    *
    * Note, for string literals:
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
    * \endcode
    *
    * @param src String using only invariant characters.
--- a/icu4c/source/common/unistr.cpp
+++ b/icu4c/source/common/unistr.cpp
@@ -230,11 +230,6 @@ UnicodeString::UnicodeString(UChar32 ch) {
   }
 }
 
-UnicodeString::UnicodeString(const char16_t *text) {
-  fUnion.fFields.fLengthAndFlags = kShortString;
-  doAppend(text, 0, -1);
-}
-
 UnicodeString::UnicodeString(const char16_t *text,
                              int32_t textLength) {
   fUnion.fFields.fLengthAndFlags = kShortString;


roubert:
Please take a look at this.

I have intentionally not done anything about the documentation yet, thinking that it'll be a better use of time to first make sure that we all agree that this change should be made at all, and if so, that the implementation is sound.

As can be seen directly in this PR no existing call sites should need to be changed, the updated new templated constructor will accept all existing callers with legacy behaviour kept intact, so I think that the change could be deemed acceptable for the ICU API stability promises. But it must be mentioned that the old constructors really do get removed so in case anyone anywhere has code that does something like taking the address of one of the old constructors then that code will cease to work after this change. As the change is gated on `U_HIDE_DRAFT_API` there should be ample time to find and address any such unusual usage, if any actually exists at all.

icu4c/source/common/unicode/unistr.h
```diff
@@ -3101,7 +3104,8 @@ class U_COMMON_API UnicodeString : public Replaceable
    * must be NUL (U+0000) terminated.
    * @stable ICU 2.0
    */
-  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const char16_t *text);
+  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const char16_t *text) :
```
markusicu, line 3107: The old from-pointer constructors are not "explicit" by default, and many users probably use the implicit conversion. (Inside ICU we make it explicit.)

The new from-string_view constructors are explicit, so this change would break many users.

Should we use UNISTR_FROM_STRING_EXPLICIT on the new constructor, now that it becomes the successor for the old ones? That means that the default usage would hide the string contents copy when constructing a string_view from something other than a raw pointer. Win some, lose some?

markusicu (commented):
In principle, I like this a lot. While it adds yet more template magic, it also magically makes existing call sites that use literals faster.

One question/concern below.

Once we have an answer for the explicit-constructor question, this wants to be proposed as an API change via email to icu-design.
(I am not too worried about people getting the address of one of the old constructors.)

@eggrobin what do you think?
@richgillam heads-up...

icu4c/source/common/unicode/unistr.h
roubert: The constructor proposed in this PR works by not being marked `explicit`, that's what makes it possible for it to be correctly invoked on existing code like `UnicodeSet(u"[∞]", status)` which wouldn't be possible otherwise.

markusicu: I saw that you were modifying the new templatized constructor but missed that you were removing the "explicit inline" part. I suggest that we put UNISTR_FROM_STRING_EXPLICIT on the replacement then, so that people who did turn that on don't backslide.

markusicu (commented):
Are you planning to do the same treatment for the other functions that have overloads for both raw-pointer and template-S?

roubert:
> Are you planning to do the same treatment for the other functions that have overloads for both raw-pointer and template-S?

I haven't made any such plans for I haven't really looked at those functions at all (so I don't know if there's anything that would need to be done there).

icu4c/source/common/unicode/unistr.h
roubert: Done.

markusicu (commented):
Hi @roubert, the ICU 76 feature freeze is in less than 2 weeks. It would be great if you could finish your work here. Please take a look at the other new functions that take some ConvertibleToU16StringView, and have raw-pointer overloads, to see if they should get the same treatment.

When you have a complete set of changes, remember to write an API change proposal; you could reply to my proposal with an addendum, like I just did.

markusicu:
PS: Please don't squash until after approval, except when you need to rebase.

roubert:
@eggrobin, @richgillam as you were the reviewers of PR #3076, I'd like to know your opinion on these proposed changes here.

icu4c/source/common/unicode/unistr.h
```diff
@@ -3258,10 +3263,13 @@ class U_COMMON_API UnicodeString : public Replaceable
    * @param text UTF-16 string
    * @draft ICU 76
    */
-  template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
-  explicit inline UnicodeString(const S &text) {
+  template <typename S,
+            typename = std::enable_if_t<
+                ConvertibleToU16StringView<S> &&
+                !std::is_same_v<std::remove_cv_t<std::remove_reference_t<S>>, UnicodeString>>>
+  UNISTR_FROM_STRING_EXPLICIT UnicodeString(S&& text) {
```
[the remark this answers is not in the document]
roubert: Declaring the parameter as `S&&` goes together with then passing it on using `std::forward<S>()` and this is what's called _perfect forwarding_ and this is what delays the possible construction of any new string view object until after it has been verified that the parameter isn't a `nullptr`.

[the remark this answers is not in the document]
eggrobin: > this is what delays the possible construction of any new string view object until after it has been verified that the parameter isn't a nullptr.

What? If S is not a string view already, but is convertible to a string view, then const S& is a const reference to an S, not a string view; the construction only happens when you pass it to a function that actually takes a string_view (so now, it happens when you call toU16StringView in toU16StringViewNullable).

Perfect forwarding is about preserving the value category, for instance, to allow a move constructor to be called through another function (as in the `emplace` member function on containers). I don’t think it helps here.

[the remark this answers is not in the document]
roubert: > […]; the construction only happens when you pass it to a function that actually takes a string_view (so now, it happens when you call toU16StringView in toU16StringViewNullable).

Exactly, that was what I was trying to express.

icu4c/source/common/unicode/char16ptr.h
```diff
@@ -309,13 +310,19 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
...
+ * Is T convertible to a std::u16string_view or some other 16-bit string view?
  * @internal
  */
 template<typename T>
 constexpr bool ConvertibleToU16StringView =
-    std::is_convertible_v<T, std::u16string_view> ||
-    (U_SIZEOF_WCHAR_T==2 && std::is_convertible_v<T, std::wstring_view>);
+    std::is_convertible_v<T, std::u16string_view>
+#if !U_CHAR16_IS_TYPEDEF
+    || std::is_convertible_v<T, std::basic_string_view<uint16_t>>
```
eggrobin, line 320: I do not think that we should support `std::basic_string_view<uint16_t>`; uint16_t does not have a standard std::char_traits specification, so I am not sure how that will work (and indeed, how exactly it works could be changed by someone defining their own specialization), and it obviously is incompatible with string literals.

We support string views because we want to be compatible with idiomatic modern code. `std::basic_string_view<uint16_t>` may be modern, but idiomatic it is not.

Of course everything _exists_,  but compare the number of results for https://github.com/search?q=%22basic_string_view%3Cuint16_t%3E%22&type=code and https://github.com/search?q=%22std%3A%3Au16string_view%22&type=code or https://github.com/search?q=%22std%3A%3Awstring_view%22&type=code.

icu4c/source/common/unicode/unistr.h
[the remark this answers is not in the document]
eggrobin: But you don’t need perfect forwarding to achieve that, a const S& here and a const T& in toU16StringViewNullable  should also work.

eggrobin:
> I'd like to know your opinion on these proposed changes here.

I have opined. Apologies for the delayed review, I was on vacation last week.
Other than the points discussed above, this seems like an improvement (and reduces the number of constructors, except for the draft transition period).

markusicu:
Reminder...

> > Are you planning to do the same treatment for the other functions that have overloads for both raw-pointer and template-S?
> 
> I haven't made any such plans for I haven't really looked at those functions at all (so I don't know if there's anything that would need to be done there).

icu4c/source/common/unicode/char16ptr.h
markusicu: FWIW I didn't think that supporting uint16_t, given ICU's and its users' history, was a bad idea. (We know that some people still use UChar=uint16_t, although UChar is nearly gone from ICU C++).

But if uint16_t as a string unit type is not fully supported by the STL, then that could be more frustrating than useful.

icu4c/source/common/unicode/unistr.h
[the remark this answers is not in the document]
markusicu: I see that you also backed out the condition `!std::is_same_v<std::remove_cv_t<std::remove_reference_t<S>>, UnicodeString>`. Did you originally add that to yield to the copy constructor? Did you now conclude that you don't need it because that is a more direct signature match anyway?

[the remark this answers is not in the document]
roubert: That condition was to not match the user-defined conversion function from `UnicodeString` to `std::u16string_view` which otherwise would have matched `ConvertibleToU16StringView`.

icu4c/source/common/unicode/char16ptr.h
roubert: What exactly does "not fully supported" mean here? Is it something that could actually effect the operations that are done here, which are nothing more than calling the `data()` and `size()` members?

roubert:
> Reminder...

I just think that it's a good idea to 1) make sure that we all agree that this should be done at all and 2) agree on how exacrly to do it, before spending any time on doing any further changes.

icu4c/source/common/unicode/char16ptr.h
eggrobin: According to my reading https://eel.is/c++draft/string.view.template#general-1, `std::basic_string_view<uint16_t>` should be ill-formed unless `char_traits<uint16_t>` contains what is needed, which the standard does not mandate; so depending on the whims of the various bits of the standard library implementation, there is no guarantee that this class compiles.

icu4c/source/common/unicode/char16ptr.h
```diff
@@ -334,6 +352,21 @@ inline std::u16string_view toU16StringView(std::wstring_view sv) {
...
+
+template <typename T,
+          typename = typename std::enable_if_t<!std::is_pointer_v<std::remove_reference_t<T>>>>
+inline std::u16string_view toU16StringViewNullable(const T& text) {
+    return toU16StringView(text);
+}
+
+template <typename T,
+          typename = typename std::enable_if_t<std::is_pointer_v<std::remove_reference_t<T>>>,
+          typename = void>
```
markusicu, line 364: just curious: what does this last `typename = void` do?

markusicu:
Suggestion:
- drop the additional handling of uint16_t in char16ptr.h
- keep the addition of toU16StringViewNullable
- I don't feel strongly about keeping or reverting the changes from checking U_SIZEOF_WCHAR_T==2 in the template logic vs. in the preprocessor
- keep the replacement of (fn(ptr) + fn(const S&) --> fn(const S &)+nullable

Note: I looked at the other overloads that I added, and AFAICT none of them has a sibling overload that just takes a pointer, so this might be it for now.

icu4c/source/common/unicode/char16ptr.h
eggrobin: Oh, this is a classic ugly trick stemming from the fact that enable_if is an ugly trick.
The previous one is a declaration of
`template<typename T, typename (unnamed)> std::u16string_view toU16StringViewNullable(const T&)`.
You cannot declare the same thing a second time with a different default template parameter (even though the intended usage here is that you never pass the unnamed template parameter, and that the defaults cause only one of the two to exist). So you declare something different,
`template<typename T, typename (unnamed), typename (also unnamed)> std::u16string_view toU16StringViewNullable(const T&)`.

This will all go away with C++20 and requires clauses, and then the requires clause will be part of what you declare and a small amount of sanity will be restored.

eggrobin: See https://godbolt.org/z/eGno9xrsP.

eggrobin: And for completeness, a glimpse into the bright future of the year 2020: https://godbolt.org/z/e4G48W36n.

icu4c/source/common/unicode/char16ptr.h
roubert: Hm, does that mean that even the two mentions of `std::basic_string_view<uint16_t>` as a possible type here would fail to compile, or just that any code that tried to actually use that could fail to compile with some standard library implementations?

Our CI tests this with the GNU, LLVM and MSVC standard library implementations and at least with those the mentions of it as a possible type here compiles just fine.

So to gauge whether this really is a useful thing to support, I now took a closer look at what Clang 18 actually does with the following piece of example code together with the implementation currently proposed in this PR:

```
int test() {
  static constexpr UChar TEXT[] = { 0x66, 0x6f, 0x6f, 0x00 };
  return do_something(icu::UnicodeString(TEXT));
}
```

Both with the default `UChar` = `char16_t` and also with `-DUCHAR_TYPE=uint16_t` this really did result in the compiler directly creating a string view object of known size, which is what I had assumed that it would do (before you pointed out that this data type wasn't standard), so as long as we keep having both `UChar` and `UCHAR_TYPE` in the public API, I think it's very nice that this just works (and that switching the data type doesn't suddenly and unexpectedly change the runtime cost from fixed to linear).

So unless there's any real reason to believe that this would make it impossible to compile ICU4C itself with some standard library implementation that it otherwise would be reasonable to expect to be used, I'd say that supporting this adds some real actual value.

roubert:
> @richgillam heads-up...

If there's any reason for why any of this wouldn't work for Apple, now would be a very good time to learn about that.

icu4c/source/common/unicode/char16ptr.h
markusicu: > So unless there's any real reason to believe that this would make it impossible to compile ICU4C itself with some standard library implementation that it otherwise would be reasonable to expect to be used, I'd say that supporting this adds some real actual value.

wfm, ok to keep the uint16_t support until we learn otherwise
and I think we can consider this part "draft" and change it if forced by the standard, or maybe with compiler ifdefs

roubert:
> Are you planning to do the same treatment for the other functions that have overloads for both raw-pointer and template-S?

I've now searched through the entire file without finding any such functions.

roubert:
OK, it seems that we've reached some kind of silent consensus here, so I've now updated the documentation comments and made this PR ready to merge (if the proposed API change gets approved by the TC).

icu4c/source/common/unicode/unistr.h
```diff
@@ -105,13 +105,11 @@ class UnicodeStringAppendable;  // unicode/appendable.h
...
- * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+ * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
  * length determination:
  * \code
- * #include &lt;string_view&gt;
- * using namespace std::string_view_literals;
- * UnicodeString str(u"literal"sv);
- * if (str == u"other literal"sv) { ... }
+ * UnicodeString str(u"literal");
+ * if (str == u"other literal") { ... }
```
markusicu, line 112: Please revert the operator== example. Users who make the constructor explicit don't get the vanilla literal auto-converted to a UnicodeString, and users who do get the auto conversion get a copy of the string literal contents. With your change here, they would avoid strlen() but not the contents copy.

Using the new operator==(const S &) would still be more efficient with an input u16string_view.

icu4c/source/common/unicode/unistr.h
```diff
@@ -335,13 +333,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
    *
-   * For performance, you can use std::u16string_view literals with compile-time
+   * For performance, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
    * UnicodeString str = ...;
-   * if (str == u"literal"sv) { ... }
+   * if (str == u"literal") { ... }
```
markusicu, line 340: ditto please revert this change

icu4c/source/common/unicode/unistr.h
```diff
@@ -3088,20 +3085,19 @@ class U_COMMON_API UnicodeString : public Replaceable
...
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
```
markusicu, line 3092: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3113,20 +3109,18 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3116: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3140,21 +3134,20 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3141: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3172,13 +3165,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3172: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3195,13 +3186,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3193: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3219,13 +3208,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3215: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3280,13 +3267,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString alias = UnicodeString::readOnlyAlias(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString alias = UnicodeString::readOnlyAlias(u"literal");
+   * if (str == u"other literal") { ... }
```
markusicu, line 3274: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3369,13 +3354,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
-   * Since C++17 and ICU 76, you can use std::u16string_view literals with compile-time
+   * Since C++17 and ICU 76, you can use UTF-16 string literals with compile-time
    * length determination:
    * \code
-   * #include &lt;string_view&gt;
-   * using namespace std::string_view_literals;
-   * UnicodeString str(u"literal"sv);
-   * if (str == u"other literal"sv) { ... }
+   * UnicodeString str(u"literal");
+   * if (str == u"other literal") { ... }
```
markusicu, line 3361: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
```diff
@@ -3485,13 +3468,11 @@ class U_COMMON_API UnicodeString : public Replaceable
...
```
markusicu, line 3475: ditto revert operator== sample code

icu4c/source/common/unicode/unistr.h
roubert: I guess that either I don't understand what you mean, or the compiler doesn't do what you think it does. To verify what the compiler really does, I wrote this piece of example code:

```c++
bool is_foo(const icu::UnicodeString& input) {
  return input == u"foo";
}
```

And this piece of example code, using an inlined `std::u16string_view` instead of the UTF-16 string literal:

```c++
bool is_foo(const icu::UnicodeString& input) {
  return input == u"foo"sv;
}
```

Testing with the Clang 18 compiler, these both result in the same, both get the known length hardcoded and this length together with a pointer to the string literal then gets passed directly to the underlying `doEquals()` function.

This is the behaviour that I had expected and the reason for why I proposed updating the documentation to recommend the simpler code that uses the string literal directly. Did you expect something else, or am I just not understanding why you prefer using `u""sv` instead?

roubert: Maybe this could become easier to understand by flipping the question around:

Why would `operator==(const UnicodeString&)` be invoked unless you already have a `UnicodeString` object?

The compiler knows how to directly make a `std::u16string_view` from a `u""` string literal.

markusicu: Ok. I understand that my assumption was wrong. A raw pointer or array is ConvertibleToU16StringView. I need to get my head around all of the effects from that.

markusicu: FYI The simplification of API docs here is timely after my recent PR that introduced those, but not related to the constructor changes in this PR. I think that's part of what tripped me up.

icu4c/source/common/unicode/unistr.h
```diff
@@ -3140,21 +3134,20 @@ class U_COMMON_API UnicodeString : public Replaceable
...
    * \endcode
    *
    * @param text NUL-terminated UTF-16 string
    * @stable ICU 59
    */
   UNISTR_FROM_STRING_EXPLICIT UnicodeString(const wchar_t *text) :
-      UnicodeString(ConstChar16Ptr(text)) {}
+      UnicodeString(ConstChar16Ptr(text), -1) {}
 #endif
+#endif  // U_HIDE_DRAFT_API
```
markusicu, line 3150: I just noticed that the from-nullptr constructor, which follows next, is not being removed. It's the only remaining version of the from-raw-pointer-only constructors.

Does the new toU16StringViewNullable not handle this case?

icu4c/source/common/unicode/unistr.h
roubert: > The simplification of API docs here is timely after my recent PR that introduced those, but not related to the constructor changes in this PR.

Oh, is it not? After I learned (PR #3102) that the constructor required passing it a string view in order to not recalculate the length, I just assumed that the reason for the documentation to recommend always passing a string view (even for functions where it wasn't actually needed) was so that people wouldn't have to keep in mind whether they were invoking the constructor or some other function, so after fixing the constructor it seemed obvious to me to also update this part of the documentation at the same time as I then no longer could see any advantage in continuing to recommend the use of string views instead of string literals.

markusicu: The simplification of the constructor call site is related to the constructor change in this PR.
The simplification of the operator== call site didn't require the constructor change, because it doesn't construct a UnicodeString.
Not having to think about when to use which is even better.

icu4c/source/common/unicode/unistr.h
roubert: I would be possible to use `toU16StringViewNullable()` to handle also `std::nullptr_t` but it's not necessary to do that so I thought it logical to not change more than necessary, it doesn't clash with anything and the specialized overload is possibly even ever so slightly faster than taking the detour through an empty string view.

markusicu (approved):
This API change was approved in the ICU-TC meeting yesterday.

icu4c/source/common/unicode/char16ptr.h
roubert: > Our CI tests this with the GNU, LLVM and MSVC standard library implementations and at least with those the mentions of it as a possible type here compiles just fine.

It turns out that our CI also tests with one more implementation, the CIFuzz tests, which are currently not run as presubmit, only as postsubmit (PR #3142), use some other standard library implementation than the others and that one doesn't define `std::char_traits<uint16_t>` and that makes the compilation fail spectacularly (no such thing as `std::is_convertible_v<>` failing silently when the type compared to isn't defined).

I failed to find what standard library implementation CIFuzz actually uses (it's a layer cake of Docker images, where I quickly lost track) but the error messages look exactly like error messages one gets from the LLVM release 19 libc++ so that seems like a really good guess.

I've now sent PR #3144 for review to simply disable the use of `std::basic_string_view<uint16_t>` when using a too new libc++. That should be easy to find and delete in the future, once the tested-for version is too old to be relevant anymore.

richgillam: (Late to the party-- just back from vacation)  I feel obligated to point out that Apple _does_ still define `UChar` as `uint16_t`, as we have too many internal clients that are assuming that and I'm not sure I can get them all to change.  But unless Robin's issue is causing real problems in practice, I'm fine with what Fredrik is proposing here.

richgillam:
> > @richgillam heads-up...
> 
> If there's any reason for why any of this wouldn't work for Apple, now would be a very good time to learn about that.

See my comment above.  Apple _does_ still define `UChar` as `uint16_t`.  I tried taking that out a while ago and got all kinds of bug reports.  But it sounds like maybe Robin's objections are theoretical, at least in Apple's case, so maybe we leave what Fredrik is doing the way it is?

richgillam:
> See my comment above. Apple _does_ still define `UChar` as `uint16_t`. I tried taking that out a while ago and got all kinds of bug reports. But it sounds like maybe Robin's objections are theoretical, at least in Apple's case, so maybe we leave what Fredrik is doing the way it is?

...And now I get to the bottom and am reminded that this was already merged.  I lost track-- did we decide to leave the `uint_16` stuff in there?

markusicu:
> I lost track-- did we decide to leave the `uint_16` stuff in there?

It's in, but https://github.com/unicode-org/icu/pull/3144 makes it conditional, removing it starting with libc++ 18. Making a string_view over uint16_t generates warnings with libc++ 18 and no longer works at all with libc++ 19.

It should continue to work if you use a standard library that still supports `std::basic_string_view<uint16_t>` -- but the standard does not condone it.

richgillam:
> > I lost track-- did we decide to leave the `uint_16` stuff in there?
> 
> It's in, but #3144 makes it conditional, removing it starting with libc++ 18. Making a string_view over uint16_t generates warnings with libc++ 18 and no longer works at all with libc++ 19.
> 
> It should continue to work if you use a standard library that still supports `std::basic_string_view<uint16_t>` -- but the standard does not condone it.

Okay.  Sounds like it's time for me to figure out how to get everybody to change to using `char16_t`...

icu4c/source/common/unicode/char16ptr.h
roubert: > I feel obligated to point out that Apple _does_ still define `UChar` as `uint16_t`,

The changes made should not break anything for Apple (or anyone else) and the macOS tests on our CI still pass as does testing locally on my own Mac. But it'd be most valuable if you could verify that everything still works as expected for all your internal use of ICU4C at Apple (whatever that might be) at commit 7cad791fb666e6759bd7aaea0461eebafbda9cfe or later.

richgillam: I'll be testing this pretty soon, although might not really know if we're in trouble until after ICU 76 goes out the door.  If there are problems, you'll definitely hear from me in the next couple months.

PR #3107, opened 2024-08-15
ICU-21205 Fix Eclipse failing to import the icu4j maven project

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21205

### DIFF
--- a/icu4j/main/core/pom.xml
+++ b/icu4j/main/core/pom.xml
@@ -50,6 +50,29 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>build-helper-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <!-- Test files shared between ICU4C and ICU4J -->
+            <id>add-test-resource</id>
+            <goals>
+              <goal>add-test-resource</goal>
+            </goals>
+            <configuration>
+              <resources>
+                <resource>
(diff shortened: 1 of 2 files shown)


srl295:
@catamorphism FYI

srl295 (dismissed):

mihnita:
I realized something after Steven's approval, and fixed it.
Sorry.

With my previous change ALL the shared test files would be copied in ALL the modules,
because the `build-helper-maven-plugin` :: `add-test-resource` was done in the `icu4j/pom.xml`.

So I changed it to copy only the `message2` files in `main/core`, where they are needed.
Following the same pattern we can later add other shared folders to `common_tests`, or `collate`.

---

Does it hurt to copy everything for every module?

Functionally not, these are test files in the `*-tests.jar` files.

But that is unclean.
And increases the number of files copied around when building, wasting time. 

Cons: instead of a lump copy configured in the root `pom.xml` we would have to add this plugin to each module, with slightly different settings. We have to be more deliberate and think what goes where.

srl295 (approved):

mihnita:
> or at least Eclipse

I also checked with IntelliJ and VS Code, looks good.
But I don't usually work with these IDEs much (I use VSCode for smaller stuff)
So it might still be possible that certain things don't behave.

But the import and basic build seems fine.

PR #3108, opened 2024-08-16
ICU-22669 Fix circular dependency in DateTimePatternGenerator::addICU…

richgillam:
…Patterns().

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22669


icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,62 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+    // just obviate the need for this whole function).  --rtg 2/1/24
+    CharString patternResourcePath;
+    patternResourcePath.append(DT_DateTimeCalendarTag, status)
+        .append('/', status)
+        .append(calendarTypeToUse, status)
+        .append('/', status)
+        .append(DT_DateTimePatternsTag, status);
+
+    LocalUResourceBundlePointer dateTimePatterns;
+    dateTimePatterns.adoptInstead(
```
roubert, line 838: Is there any reason to call `adoptInstead()` instead of just doing the initialization directly in the constructor?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,62 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+        ures_getByKeyWithFallback(rb.getAlias(), patternResourcePath.data(),
+                                  (UResourceBundle*)nullptr, &status));
```
roubert, line 840: Typecasting `nullptr` doesn't do anything useful, does it?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,62 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+    if (ures_getSize(dateTimePatterns.getAlias()) < 8 || ures_getType(dateTimePatterns.getAlias()) != URES_ARRAY) {
```
roubert, line 841: Shouldn't these two checks be reversed, first verifying that it is the expected type and then verifying the size of it? If it's an unknown type, it's unknown what the size means, isn't it?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,62 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+        status = U_INVALID_FORMAT_ERROR;
+        return;
+    }
+
+    for (int32_t i = 0; i < DateFormat::kDateTime; i++) {
+        LocalUResourceBundlePointer patternRes(ures_getByIndex(dateTimePatterns.getAlias(), i, nullptr, &status));
+        UnicodeString pattern;
+        if (ures_getType(patternRes.getAlias()) == URES_STRING) {
```
roubert, line 849: I suggest replacing these chained if-statements with a switch-statement, so that the call to `ures_getType()` won't need to be repeated.

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,62 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+    LocalUResourceBundlePointer dateTimePatterns;
+    dateTimePatterns.adoptInstead(
+        ures_getByKeyWithFallback(rb.getAlias(), patternResourcePath.data(),
+                                  (UResourceBundle*)nullptr, &status));
+    if (ures_getSize(dateTimePatterns.getAlias()) < 8 || ures_getType(dateTimePatterns.getAlias()) != URES_ARRAY) {
+        status = U_INVALID_FORMAT_ERROR;
+        return;
+    }
+
+    for (int32_t i = 0; i < DateFormat::kDateTime; i++) {
```
roubert, line 846: Wouldn't it be a good idea to also check `U_SUCCESS(status)` in the loop condition, so that the loop won't continue (doing nothing) after an error has occurred?

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -173,15 +167,28 @@ private void initData(ULocale uLocale, boolean skipStdPatterns) {
...
-            if (i == DateFormat.SHORT) {
-                consumeShortTimePattern(df.toPattern(), returnInfo);
+        ICUResourceBundle rb = (ICUResourceBundle) UResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME, uLocale);
+        String calendarTypeToUse = getCalendarTypeToUse(uLocale);
+        // TODO: this is a HACK to work around ICU-22668/CLDR-17892!  Remove once they're fixed!
+        if (calendarTypeToUse.equals("iso8601")) {
+            calendarTypeToUse = "gregorian";
+        }
+        ICUResourceBundle dateTimePatterns = rb.getWithFallback("calendar/" + calendarTypeToUse + "/DateTimePatterns");
+        if (dateTimePatterns.getSize() < 8 || dateTimePatterns.getType() != UResourceBundle.ARRAY) {
```
roubert, line 177: Shouldn't these two checks be reversed, first verifying that it is the expected type and then verifying the size of it? If it's an unknown type, it's unknown what the size means, isn't it?

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -173,15 +167,28 @@ private void initData(ULocale uLocale, boolean skipStdPatterns) {
...
+            throw new MissingResourceException("Resource in wrong format", "ICUResourceBundle", "calendar/" + calendarTypeToUse + "/DateTimePatterns");
+        }
+        for (int i = 0; i < 8; i++) { // no constants available for the resource indexes
+            String pattern;
+            UResourceBundle patternRes = dateTimePatterns.get(i);
+
+            if (patternRes.getType() == UResourceBundle.STRING) {
```
roubert, line 184: I suggest replacing these chained if-statements with a switch-statement, so that the call to `getType()` won't need to be repeated.

roubert (changes requested):

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -905,7 +929,9 @@ DateTimePatternGenerator::getCalendarTypeToUse(const Locale& locale, CharString&
             &localStatus);
         localeWithCalendarKey[ULOC_LOCALE_IDENTIFIER_CAPACITY-1] = 0; // ensure null termination
         // now get the calendar key value from that locale
-        destination = ulocimp_getKeywordValue(localeWithCalendarKey, "calendar", localStatus);
+        if (U_SUCCESS(localStatus)) {
```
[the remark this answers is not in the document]
richgillam: Good point.  It took me a while to figure this out too.  Added a comment.

icu4c/source/i18n/dtptngen.cpp
richgillam: I guess not.  Changed it.

icu4c/source/i18n/dtptngen.cpp
richgillam: I guess not...

icu4c/source/i18n/dtptngen.cpp
richgillam: Yeah, I guess you're right...

icu4c/source/i18n/dtptngen.cpp
richgillam: I don't really like the way that looks, but whatever...

icu4c/source/i18n/dtptngen.cpp
richgillam: I don't think it's going to make much of a difference in practice, but sure.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
richgillam: Done.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
richgillam: Done.

roubert (approved):
I know precious little about ICU resource loading logic, but I think the code looks good now.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -173,15 +167,31 @@ private void initData(ULocale uLocale, boolean skipStdPatterns) {
...
-            addPattern(df.toPattern(), false, returnInfo);
-
-            if (i == DateFormat.SHORT) {
-                consumeShortTimePattern(df.toPattern(), returnInfo);
+        ICUResourceBundle rb = (ICUResourceBundle) UResourceBundle.getBundleInstance(ICUData.ICU_BASE_NAME, uLocale);
+        String calendarTypeToUse = getCalendarTypeToUse(uLocale);
+        // TODO: this is a HACK to work around ICU-22668/CLDR-17892!  Remove once they're fixed!
+        if (calendarTypeToUse.equals("iso8601")) {
+            calendarTypeToUse = "gregorian";
+        }
```
markusicu, line 175: ICU-22668 applies to C/C++. Why is this comment and hack only in Java?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,63 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
 
-        df = DateFormat::createTimeInstance(style, locale);
-        if (df != nullptr && (sdf = dynamic_cast<SimpleDateFormat*>(df)) != nullptr) {
-            sdf->toPattern(dfPattern);
-            addPattern(dfPattern, false, conflictingString, status);
+    // HACK to get around the fact that the old SimpleDateFormat code (actually, Calendar::getCalendarTypeForLocale() )
+    // returns "gregorian" for ja_JP_TRADITIONAL instead of "japanese"
+    if (uprv_strcmp(locale.getBaseName(), "ja_JP_TRADITIONAL") == 0) {
+        calendarTypeToUse.clear().append("gregorian", status);
+    }
```
markusicu, line 818: Is this not a problem in Java?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,63 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
-            // C++ uses MEDIUM, but Java uses SHORT.
-            if ( i==DateFormat::kShort && !dfPattern.isEmpty() ) {
-                consumeShortTimePattern(dfPattern, status);
-            }
+    // NOTE: It probably makes more sense to also iterate over the DateTimeSkeletons resource and add the
+    // standard patterns with their corresponding skeletons, but this doesn't currently work because the
+    // CLDR data has too many examples of multiple (different) standard patterns with the same skeleton
+    // (not to mention standard patterns whose skeleton conflicts with an entry in availableFormats).
+    // So we probably can't do anything "smarter" without cleanup at the CLDR level (cleanup that might
+    // just obviate the need for this whole function).  --rtg 2/1/24
```
markusicu, line 829: I assume that this comment applies to both C++ and Java. If so, could you please either repeat it in Java, or at least add a brief comment there pointing here? Or maybe move this line of thought into a ticket and point there from both?

markusicu (changes requested):
The code changes look plausible, and it's definitely good not to call higher-level code from a lower level.
However, I am a little distressed by the differences in hacks and comments in what seems like should be parallel C++ and Java code.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
richgillam: I added that because I had a failing unit test on the Java side, but not on the C++ side.  The failing unit test is `DateTimeGeneratorTest.testIso8601()`, which was added by @pedberg-icu and purports to address ICU-21839, which sounds like the same thing I reported in ICU-22668/CLDR-17892.  Peter claims the problem only ever happened in Java, and that it went away by itself (he thinks because of my sideways-inheritance changes), and so he'd only added this unit test to make sure it didn't come back.

So what I'm getting from this is that there was something about calling back into `SimpleDateFormat` to get the "standard" patterns that was sidestepping the ISO 8601 problem, and that my getting the patterns straight out of the resource bundle brought it back.  Maybe there's a subtle difference in that logic between C++ and Java that meant we didn't ever see the symptom in C++.  I haven't tried it, but I'm going to guess my change causes the problem to manifest in C++ as well.

So...

I can port that unit test to C++ and (if necessary) add the "hack" there too.

Or...

I can remove the test from the Java side and assume this is just going to be broken until we address ICU-22668/CLDR-17892.

I prefer option 2.  I'm guessing you're going to want option 1.

icu4c/source/i18n/dtptngen.cpp
richgillam: Apparently not.  The failing test in ICU4C is `IntlTestDateTimePatternGeneratorAPI::testAPI()`, specifically the block of `ja_JP_TRADITIONAL` test cases at ll. 330-349.  I found that same block of test cases at ll. 580-599 of `DateTimeGeneratorTest.java`, with completely different expected results and a little note at the top that says "This is different in C++ and Java."  That bothers me-- I'd like to understand _why_ it's different and try to make it the same-- but that feels like a whole different problem than what I'm trying to solve here.

icu4c/source/i18n/dtptngen.cpp
richgillam: (sigh) Okay, I filed https://unicode-org.atlassian.net/browse/ICU-22867 and took the comment out of both source files.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
markusicu: Messy... I can live with either option.

icu4c/source/i18n/dtptngen.cpp
markusicu: ticket?

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -803,38 +803,57 @@ DateTimePatternGenerator::staticGetBaseSkeleton(
...
+    if (U_FAILURE(status)) {
+        return;
+    }
 
-            // TODO: C++ and Java are inconsistent (see #12568).
-            // C++ uses MEDIUM, but Java uses SHORT.
-            if ( i==DateFormat::kShort && !dfPattern.isEmpty() ) {
-                consumeShortTimePattern(dfPattern, status);
-            }
+    CharString patternResourcePath;
```
markusicu, line 824:
```suggestion
    // TODO: See ICU-22867
    CharString patternResourcePath;
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -173,15 +167,31 @@ private void initData(ULocale uLocale, boolean skipStdPatterns) {
...
+        ICUResourceBundle dateTimePatterns = rb.getWithFallback("calendar/" + calendarTypeToUse + "/DateTimePatterns");
```
markusicu, line 176:
```suggestion
        // TODO: See ICU-22867
        ICUResourceBundle dateTimePatterns = rb.getWithFallback("calendar/" + calendarTypeToUse + "/DateTimePatterns");
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
richgillam: Well, yuck.  I ported that test over to ICU4C, and it passes.  So we've got a subtle difference between C and J.  I added a comment on the J side explaining this, but am not filing any more tickets because I'm hoping the problem goes away once and for all when CLDR-17892 gets fixed.

icu4c/source/i18n/dtptngen.cpp
richgillam: Done: https://unicode-org.atlassian.net/browse/ICU-22870

markusicu (approved):
tnx pse squash

markusicu (approved):
looks like you rebased

richgillam:
I did rebase.  I guess I didn't strictly need to, but...

Thanks for the review!

markusicu:
> I did rebase. I guess I didn't strictly need to, but...

If you just squash and don't change any file contents, then the approval should carry over.

If you have a merge conflict, you need to rebase. If you don't, you need not.

PR #3118, opened 2024-08-26
ICU-22606 Make the ICU4J release easier & more predictable

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22606

---

This creates a GitHub workflow that builds the ICU4J artifacts using JDK 8 and publishes them to Maven Central.
It does everything that's needed: sources, javadoc, digitally signed.

The next step is to improve it and generate the javadoc ready to publish on the web, and to collect the artifacts, with md5sum for the GitHub Release.

That way the files we publish in GitHub Release / Maven Central + Javadoc site will be binary identical, and generated automatically by this GitHub action.

This will generate all ICU4J files in a GitHub Release.

### DIFF
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -1,0 +1,115 @@
+# This workflow will build a package using Maven and then publish it to GitHub packages when a release is created
+# For more information see: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#apache-maven-with-a-settings-path
+
+name: Publish to Nexus Sonatype OSS
+
+# For this to run you should define the follwing GitHub Secrets in the proper Environment.
+# In Settings -- Environments -- release_env -- Environment secrets:
+# * MAVEN_CENTRAL_USERNAME & MAVEN_CENTRAL_TOKEN:
+#   * Go to https://oss.sonatype.org and login with the `icu-robot` user
+#   * Top-right select "Profile"
+#   * Open the drop-down list showing "Summary" and select "User Token"
+#   * Click the "Access User Token" button
+#   * The `username` (first, shorter part of the token) goes in the `MAVEN_CENTRAL_USERNAME` secret
+#   * The `password` (second, longer part of the token) goes in the `MAVEN_CENTRAL_TOKEN` secret
+# * MAVEN_GPG_PRIVATE_KEY: an ASCII dump of the GPG private signing key:
+#     `gpg --output icu_release_signing.asc --armor --export-secret-key <key_id>`
+# * MAVEN_GPG_PASSPHRASE: the GPG password used to protect that key
+
+on:
+  # To trigger the Env Test workflow manually, follow the instructions in
+  # https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
+  workflow_dispatch:
+    inputs:
+      gitTag:
+        # TODO: make this mandatory and validate that it is in a release* branch and looks like
+        # 'release-\d+.\d+ or something like that.
+        # For now we don't do it so that we can test.
+        description: 'Git tag at which to sync for deploy and release'
+        type: string
+
+env:
+  SHARED_MVN_ARGS: '--no-transfer-progress --show-version --batch-mode'
+  RELEASE_FOLDER: 'target/release_files'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    environment: release-env
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+
+    - name: Checkout repo files
+      uses: actions/checkout@v4.1.7
+      with:
+        lfs: true
+
+    - name: Set up JDK
+      uses: actions/setup-java@v4.2.2
+      with:
+        java-version: '8' # The custom Taglets for javadoc (tools/build) are still Java 8. They need updating to use a different JDK version.
+        distribution: 'temurin'
+        server-id: icu4j-maven-repo # Value of the distributionManagement/repository/id field of the pom.xml
+        server-username: MAVEN_USERNAME # env variable for username in deploy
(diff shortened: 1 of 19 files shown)


mihnita:
I created small diff commits to make it clear what I did and why.

What is not visible in the PR is that I created an environment (called [`release_env`](https://github.com/unicode-org/icu/settings/environments/3801893756/edit)) with the secrets needed to sign and publish.

The signing gpg key & gpg password are new, something I created to test.
The idea is that Yoshito will update them to the official ones.

---

Also TODO is making the triggering stricter (for example to only be able to released from a release branch, and from a release tag)
But not yet, because if we do it now it will be so tight that we can't test anything.

srl295:
The PR title + description is long on promises but short on details :)

mihnita:
See this repo that I used to test that the action works properly:
https://github.com/mihnita/fake_icu/

The code is (right now) 100% in sync with this PR.

You can see the action run that matches the PR here:
https://github.com/mihnita/fake_icu/actions/runs/10602022101

* it is green
* we can check the logs
* we can download and check the `artifact.zip` (link at the bottom of the run) and compare  `icu4j-76.0.1-SNAPSHOT.jar` in the zip to what was deployed to Sonatype [`icu4j-76.0.1-20240828.183452-24.jar`](https://oss.sonatype.org/service/local/repo_groups/staging/content/com/ibm/icu/icu4j/76.0.1-SNAPSHOT/icu4j-76.0.1-20240828.183452-24.jar), they are identical. That is true for all the other files.

.github/workflows/maven-publish.yml
```diff
@@ -0,0 +1,116 @@
...
+        # TODO: make this mandatory and validate that it is in a release* branch and looks like
+        # 'release-\d+.\d+ or something like that.
+        # For now we don't do it so that we can test.
+        description: 'Git tag at which to sync for deploy and release'
+        type: string
+
+env:
+  SHARED_MVN_ARGS: '--no-transfer-progress --show-version --batch-mode'
+  # TODO: find a way to not have a version number (too many places to update)
+  ICU_VERSION: '76.0.1-SNAPSHOT'
```
srl295, line 34:
```shell
$ mvn --file=icu4j/pom.xml help:evaluate -Dexpression=project.version -q -DforceStdout
76.0.1-SNAPSHOT
```

should be able to echo this to GITHUB_OUTPUT and use it later.

srl295: oh yeah, you only need the version in one place.

.github/workflows/maven-publish.yml
```diff
@@ -0,0 +1,116 @@
...
+      env:
+        MAVEN_USERNAME: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
+        MAVEN_CENTRAL_TOKEN: ${{ secrets.MAVEN_CENTRAL_TOKEN }}
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
+
+    # These are files that will end in the "Release" section of the GitHub repo.
+    # So the jar files published in GitHub are binary identical to the ones in Maven Central.
+    - name: Prepare release folder
+      run: |
+        pushd icu4j
```
srl295, line 89:
```suggestion
        pushd icu4j
        ICU_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
```

srl295 (dismissed):
looks great to me, and it delivers on the promise of the title
(I was only kidding you)

make sure CLDR still builds…

mihnita:
Merged Steven's fix taking ICU_VERSION directly from Maven, thank you.

Tested in my "fake_icu" repo:
* Commit: https://github.com/mihnita/fake_icu/commit/ba0d53ab06ae478e0eeb2ca2e0c1014dbaa12a6c
* [New run](https://github.com/mihnita/fake_icu/actions/runs/10619788278), green, and I checked the results

All good.

---

> make sure CLDR still builds…

I tried it, it behaves like before.

The CLDR `pom.xml` lists https://maven.pkg.github.com/unicode-org/icu as Maven repository, and does not list the Sonatype Snapshots repository.
So whatever I deployed there in the last few days does not affect building CLDR.

We still have the action that publishes to the Github maven repo, working exactly like before.
So CLDR will take artifacts from there (if available), or from local if one did a `mvn install` in a local ICU4J.

It might be possible to drop the publishing to GitHub repo and use the Sonatype snapshots one. I don't know the pros and cons, I need to think a bit, and try things.

Top of my head:
* I think deploying from GitHub to GitHub is faster.
* But using Sonatype might mean that users wanting to use the snapshot don't need to configure the GitHub key only to download artifacts (https://cldr.unicode.org/development/maven). That would be nice!

Interesting to check.

But for now the two mechanisms are 100% independent.

mihnita:
Yes.

I cloned a 100% clean CLDR repo, I removed my `~/.m2/repository/`, moved `~/.m2/settings.xml`

Then I changed the repository declared in `<CLDR>/tools/pom.xml` from https://maven.pkg.github.com/unicode-org/icu to https://oss.sonatype.org/content/repositories/snapshots.
No other changes.

And it builds everything, no errors, with the icu4j 76.0.1-SNAPSHOT artifacts I published earlier in my tests with this PR.

I am not saying "let's do this" :-)
Still need to check that nothing else breaks...
But it is possible, and relatively easy.
And it means developers / CLDR / Unicode tools don't have to tinker with GitHub keys and `settings.xml`

mihnita:
@srl295 
Thanks for the review!
See comments above about building CLDR.

mihnita:
@srl295 

Sorry, squishing the branches resulted in a force push that canceled your approval.
Or maybe my mergins your suggestion + removing ICU_VERSION.

But the point is, I need your approval again.
Thank you!
M.

srl295 (approved):

mihnita:
Thank you!

PR #3120, opened 2024-08-28
ICU-22876 C++ UnicodeSet/USet easy item iteration

markusicu:
Modern C++ iteration over the code points, ranges, or strings, or all of the elements, in a C++ UnicodeSet or C USet.

API approved by ICU-TC on 2024sep05.
Minor change sep09 during implementation testing:
Changed CodePointRange::iterator::operator++() from void to returning iterator & (usual signature)

- UnicodeSet.ranges() returns a USetRanges which has begin() & end() which return USetRangeIterator which works like an idiomatic C++ iterator.
- Parallel for UnicodeSet.codePoints() & UnicodeSet.strings().
- For iterating over all elements (code points & strings) in C++, by popular demand, I added begin() & end() directly to class UnicodeSet, similar to Java UnicodeSet which *is an* Iterable. The header-only wrapper still has separate range and iterator classes.
- For strings, adding new C API to get strings without copying their contents.
- Also, UnicodeSet.strings() wants to be called that in C++ just like in Java, and just like codePoints() & ranges(). That then required me to rename the *field* strings to strings_.
- Operates on C API USet, so that C API users should be able to use it. C++ UnicodeSet.ranges() and it siblings are trivial convenience functions.
- As discussed some time ago, this is written as a header-only API. It is the first of its kind.
- As such, I am defining a U_HEADER_NESTED_NAMESPACE and a U_HEADER_ONLY_NAMESPACE which put the new API into either icu::header when compiling user code or icu::internal when compiling ICU library code, so that even if we instantiate the API inside ICU it won't be the same as what a caller sees.
  - Exception: This does not work on Windows where we explicitly export APIs, because the library does not export the member functions that return the icu::header types.
  - We also don't need this on Windows because we explicitly export APIs.
  - So on Windows the namespace is always icu::header, and the UnicodeSet member functions are exported, but the USet wrapper types are not.
- Naming: I am using USetXyz class names because these classes actually operate on a USet, the C wrapper around a C++ UnicodeSet, and they are defined in unicode/uset.h. This would also allow future additions of UnicodeSetXyz that don't go through the C API, if we ever wanted that.

Sample real, working call sites:
```
    UnicodeSet set(u"[abcçカ🚴]", errorCode);
    for (auto [start, end] : set.ranges()) {
        printf("set.range U+%04lx..U+%04lx\n", (long)start, (long)end);
    }

    UnicodeSet set(u"[abcçカ🚴{}{abc}{de}]", errorCode);
    for (auto el : set) {
        std::string u8;
        printf("set.string length %ld \"%s\"\n", (long)el.length(), el.toUTF8String(u8).c_str());
    }
```

Output:
```
   utility {
      UnicodeSetTest {
         TestUnicodeSetRanges {
set.range U+0061..U+0063
set.range U+00e7..U+00e7
set.range U+30ab..U+30ab
set.range U+1f6b4..U+1f6b4
         } OK:   TestUnicodeSetRanges 
      } OK:   UnicodeSetTest 
   } OK:   utility 

   utility {
      UnicodeSetTest {
         TestUnicodeSetElements {
set.string length 1 "a"
set.string length 1 "b"
set.string length 1 "c"
set.string length 1 "ç"
set.string length 1 "カ"
set.string length 2 "🚴"
set.string length 0 ""
set.string length 3 "abc"
set.string length 2 "de"
         } OK:   TestUnicodeSetElements 
      } OK:   UnicodeSetTest 
   } OK:   utility 
```

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22876

### DIFF
--- a/icu4c/source/common/unicode/uset.h
+++ b/icu4c/source/common/unicode/uset.h
@@ -33,7 +33,10 @@
 #include "unicode/uchar.h"
 
 #if U_SHOW_CPLUSPLUS_API
+#include <string_view>
+#include "unicode/char16ptr.h"
 #include "unicode/localpointer.h"
+#include "unicode/unistr.h"
 #endif   // U_SHOW_CPLUSPLUS_API
 
 #ifndef USET_DEFINED
@@ -955,7 +958,7 @@ uset_charAt(const USet* set, int32_t charIndex);
 
 /**
  * Returns the number of characters and strings contained in this set.
- * The last (uset_getItemCount() - uset_getRangeCount()) items are strings.
+ * The last uset_getStringCount() == (uset_getItemCount() - uset_getRangeCount()) items are strings.
  *
  * This is slower than uset_getRangeCount() and uset_getItemCount() because
  * it counts the code points of all ranges.
@@ -965,6 +968,8 @@ uset_charAt(const USet* set, int32_t charIndex);
  * contained in set
  * @stable ICU 2.4
  * @see uset_getRangeCount
+ * @see uset_getStringCount
+ * @see uset_getItemCount
  */
 U_CAPI int32_t U_EXPORT2
 uset_size(const USet* set);
@@ -975,10 +980,41 @@ uset_size(const USet* set);
  * @stable ICU 70
  * @see uset_getItemCount
  * @see uset_getItem
+ * @see uset_getStringCount
  * @see uset_size
  */
 U_CAPI int32_t U_EXPORT2
 uset_getRangeCount(const USet *set);
+
+#ifndef U_HIDE_DRAFT_API
+
+/**
+ * @param set the set
+ * @return the number of strings in this set.
+ * @draft ICU 76
+ * @see uset_getRangeCount
+ * @see uset_getItemCount
+ * @see uset_size
+ */
+U_CAPI int32_t U_EXPORT2
+uset_getStringCount(const USet *set);
+
+/**
+ * Returns the index-th string (empty or multi-character) in the set.
+ * The string may not be NUL-terminated.
+ * The output length must be used, and the caller must not read more than that many UChars.
+ *
+ * @param set the set
+ * @param index the string index, 0 .. uset_getStringCount() - 1
+ * @param pLength the output string length; must not be NULL
+ * @return the pointer to the string; NULL if the index is out of range or pLength is NULL
+ * @draft ICU 76
+ * @see uset_getStringCount
+ */
+U_CAPI const UChar* U_EXPORT2
+uset_getString(const USet *set, int32_t index, int32_t *pLength);
+
+#endif  // U_HIDE_DRAFT_API
 
 /**
  * Returns the number of items in this set.  An item is either a range
@@ -987,6 +1023,8 @@ uset_getRangeCount(const USet *set);
  * @return a non-negative integer counting the character ranges
  * and/or strings contained in set
  * @stable ICU 2.4
+ * @see uset_getRangeCount
+ * @see uset_getStringCount
  */
 U_CAPI int32_t U_EXPORT2
 uset_getItemCount(const USet* set);
@@ -1001,6 +1039,7 @@ uset_getItemCount(const USet* set);
  * If <code>itemIndex</code> is at least uset_getRangeCount() and less than uset_getItemCount(), then
  * this function copies the string into <code>str[strCapacity]</code> and
  * returns the length of the string (0 for the empty string).
+ * See uset_getString() for a function that does not copy the string contents.
  *
  * If <code>itemIndex</code> is out of range, then this function returns -1.
  *
@@ -1018,6 +1057,7 @@ uset_getItemCount(const USet* set);
  * @return the length of the string (0 or >= 2), or 0 if the item is a range,
  *         or -1 if the itemIndex is out of range
  * @stable ICU 2.4
+ * @see uset_getString
  */
 U_CAPI int32_t U_EXPORT2
 uset_getItem(const USet* set, int32_t itemIndex,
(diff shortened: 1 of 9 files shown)


icu4c/source/common/unicode/uset.h
```diff
@@ -1285,4 +1286,95 @@ U_CAPI UBool U_EXPORT2
...
+    bool operator==(const UnicodeSetRangeIterator &other) const {
+        // same set, and both or neither are done, and same index if not done
+        return uset == other.uset &&
+            (rangeIndex >= rangeCount) == (other.rangeIndex >= other.rangeCount) &&
+            ((rangeIndex >= rangeCount) || rangeIndex == other.rangeIndex);
+    }
+
+    bool operator!=(const UnicodeSetRangeIterator &other) const { return !operator==(other); }
+
+    std::pair<UChar32, UChar32> operator*() const {
```
eggrobin, line 1317: My immediate thought is that `std::pair<UChar32, UChar32>` is too nondescript a type for a range of code points.
I would expect it to be, well, a [range](https://en.cppreference.com/w/cpp/ranges/range).
In particular, I would expect to be able to iterate over it:
```C++
for (auto const range: set.ranges()) {
  if (MayContainCats(range)) {
    for (UChar32 const cp : range) {
      if (u_iscat(cp)) {
        u_pet(cp);
      }
    }
  }
}
```

There is a bit of boilerplate needed to get that, but I just remembered that conveniently some egg wrote something like it back in October: https://github.com/unicode-org/unicodetools/blob/ce2ff440fca747e32e028594d8484db6e1d33afb/c/uca/sifter/ucd.hpp#L29-L91.

markusicu: > My immediate thought is that `std::pair<UChar32, UChar32>` is too nondescript a type for a range of code points. I would expect it to be, well, a [range](https://en.cppreference.com/w/cpp/ranges/range).

Makes sense, and it's tempting, but then either I can't use the nifty `for (auto [start, end] : set.ranges())` loop, or else I have to make funny range field names public. (See below.)

I also don't see why someone would want to do the double loop -- the next thing I am going to add is a code point iterator. Just like in Java, we give easy iteration over either code points or ranges, and callers should choose what's most useful to them. They should not need a double loop.

```
class CodePointRange {
public:
    CodePointRange(UChar32 start, UChar32 end) : start_(start), end_(end) {}
    CodePointRange(const CodePointRange &other) = default;
    inline operator std::pair<UChar32, UChar32>() const { return std::make_pair(start_, end_); }
    UChar32 begin() const { return start_; }
    UChar32 end() const { return end_ + 1; }

private:
    UChar32 start_, end_;
};

usettest.cpp:4274:16: error: cannot decompose private member 'start_' of 'icu_76::header::CodePointRange'
    for (auto [start, end] : set.ranges()) {
```

eggrobin: > but then either I can't use the nifty for (auto [start, end] : set.ranges()) loop, or else I have to make funny range field names public. (See below.)

I think _especially_ If you want to allow this nifty syntax, you should use a proper type, and then those member variables can have names and comments to clarify their semantics (iirc ICU and standard C++ have different opinions on what end means?). I see no problem with making it a struct with two public member variables while also giving it the structure of a range.

As for the use case, that double loop was a toy example, but being able to treat it as a range, and to feed it to various things that operate on ranges, seems useful. (certainly having a size() seems convenient).

eggrobin: > iirc ICU and standard C++ have different opinions on what end means

Indeed looking at your example above
```C++
    for (auto [start, end] : set.ranges()) {
        printf("set.range U+%04lx..U+%04lx\n", (long)start, (long)end);
    }
```
it looks like those are inclusive ranges, which are not the norm in C++. This is fine, since those _are_ the norm in Unicode, but then they really want to be their own type, not a nondescript pair.
The member variables can probably be called first and last or somesuch to avoid the word `end` which would be interpreted as something else here.

markusicu: Ok, done. Added CodePointRange and got it to work.
Note that in ICU, especially in UnicodeSet/USet, ranges have _always_ been inclusive start..end.

markusicu:
Hi everyone:
- I think I have a complete set of new APIs now for this purpose, including the popular "iterate over all elements" thing.
- I have updated the PR description, please read.
- As noted there, I am considering changing the prefixes of the new classes from UnicodeSetXyz to USetXyz, since they are really operating on a C USet. Ok? (One could counter-argue that in C++ we use UnicodeSet. Or that few people will actually type these class names.)
- Please take a look at the sample code in usettest.cpp. (No real unit test code yet.) Looks nice and colloquial?
- Please take a look at the API *signatures*.
- Please take a look at the NOTEs and TODOs.
- @eggrobin I would like to ask you in particular to look at the implementation code as well.

markusicu:
@richgillam do you care whether the prefix of the new classes, which operate on USet but are in C++ and also created by UnicodeSet, is USetXyz or UnicodeSetXyz? For example,
- USetCodePointIterator
- UnicodeSetCodePointIterator

Right now, I am leaning towards the shorter version which seems slightly more accurate.
I should really send the approval email today...

@aheninger opinion?

markusicu:
I used Craig as a sounding board for the class prefix naming question. What tipped him over the top is that the new types are in uset.h, so he would expect them to start with "USet". I like that. Will rename.

markusicu:
Renamed. This would also allow future additions of UnicodeSetXyz that don't go through the C API, if we ever wanted that.

richgillam:
> @richgillam do you care whether the prefix of the new classes, which operate on USet but are in C++ and also created by UnicodeSet, is USetXyz or UnicodeSetXyz? For example,
> 
> * USetCodePointIterator
> * UnicodeSetCodePointIterator

Yes, I like USet here, for the reasons you outline.

icu4c/source/common/unicode/uset.h
```diff
@@ -1285,4 +1325,414 @@ U_CAPI UBool U_EXPORT2
...
+#if U_SHOW_CPLUSPLUS_API
+#ifndef U_HIDE_DRAFT_API
+
+namespace U_HEADER_ONLY_NAMESPACE {
+
+// TODO: Naming: For now, I am working with UnicodeSetXyz class prefixes.
+// This might be a little confusing, since these classes actually operate on a USet,
+// the C wrapper around a C++ UnicodeSet.
+// I am thinking of renaming the type prefixes to just USetXyz.
+// Opinions?
```
richgillam, line 1337: I think you can take this comment out now.

markusicu: whoops! thanks -- done

richgillam (approved):
Wow, that's a lot of code to do something that seems like it should be fairly simple!

I'm not sure I completely followed everything, but it all seems to make sense.  But I'll defer to the other reviewers, all of whom understand what's going on here better than I do.

markusicu:
> Wow, that's a lot of code to do something that seems like it should be fairly simple!

I agree!
It's all because C++ didn't create a real iterator pattern and instead stuck with implementing iterators that look and quack like pointers.

icu4c/source/common/unicode/uset.h
```diff
@@ -1285,4 +1325,408 @@ U_CAPI UBool U_EXPORT2
...
+        // NOTE: I think we don't need to compare rangeCount & end given private constructor
+        // and assuming we don't compare iterators across the set being modified.
+        // We might even skip comparing uset.
+        // Unless we want operator==() to be "correct" for more than iteration.
+        return uset == other.uset && c == other.c && index == other.index;
+    }
+
+    bool operator!=(const USetElementIterator &other) const { return !operator==(other); }
+
+    UnicodeString operator*() const {
```
eggrobin, line 1636: Hm, when we discussed this one in ICU-TC, I was thinking of std::whateverstring, but of course we probably want UnicodeString here.
Does UnicodeString have the small-string optimization though?

icu4c/source/common/unicode/uset.h
```diff
@@ -1285,4 +1325,408 @@ U_CAPI UBool U_EXPORT2
...
+ */
+class USetElementIterator {
+public:
+    USetElementIterator(const USetElementIterator &other) = default;
+
+    bool operator==(const USetElementIterator &other) const {
+        // NOTE: I think we don't need to compare rangeCount & end given private constructor
+        // and assuming we don't compare iterators across the set being modified.
+        // We might even skip comparing uset.
+        // Unless we want operator==() to be "correct" for more than iteration.
```
eggrobin, line 1630: It is common for container modification to invalidate iterators, so I would not worry about this.

icu4c/source/common/unicode/uset.h
markusicu: > Hm, when we discussed this one in ICU-TC, I was thinking of std::whateverstring, but of course we probably want UnicodeString here.

I chose to return a UnicodeString because it's easy to make one from a code point, and fits well with other ICU APIs.

> Does UnicodeString have the small-string optimization though?

yes

markusicu:
@eggrobin FYI commit ed5e8e30513b8d7d9071047e28681af8abb0ba57 addresses the ICU-TC API proposal review

markusicu:
I think I am all done.
It seems to all work, the API reflects what has been approved, with one minor exception noted in the PR description.
API docs all there, although minimal for boilerplate C++ idioms.
Tested via range-based for loops as well as explicit calls of each new function.

It might still make sense to review this one commit at a time.
Once approved I will squash, then merge.

markusicu:
Dear reviewers, it would be lovely to get this reviewed, feedback addressed, approved, and merged... fairly quickly, to unblock @roubert's follow-up work with the header-only collator predicates. I would be happy to jump into a video meeting to talk things through.

richgillam (approved):
Looks like all that's changed since the last time I reviewed this is documentation and tests.  The new documentation and tests look good.

icu4c/source/common/unicode/uset.h
```diff
@@ -1618,6 +1622,115 @@ class UnicodeSetStrings {
...
+        if (index < rangeCount) {
+            // Fetch the first range.
+            operator++();
+        }
+        // Otherwise don't move beyond the (index - rangeCount)-th string.
+    }
+
+    const USet *uset;
+    int32_t index;
+    int32_t rangeCount, totalCount;
```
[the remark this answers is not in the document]
markusicu: done, PTAL

echeran (approved):
LGTM!

richgillam (approved):
New documentation also looks good.

PR #3129, opened 2024-09-05
ICU-22716 Fix GCC warning and turn warning to errors

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22716


roubert:
The gcc-10-stdlib17 tests still fail with `-Werror=return-local-addr`.

icu4c/source/common/ucnvmbcs.cpp
```diff
@@ -3144,13 +3144,14 @@ ucnv_MBCSGetNextUChar(UConverterToUnicodeArgs *pArgs,
     }
 
     if(c<0) {
-        if(U_SUCCESS(*pErrorCode) && source==sourceLimit && lastSource<source) {
+        if(U_SUCCESS(*pErrorCode) && source==sourceLimit && lastSource<source &&
+           size_t(source-lastSource) <= sizeof(cnv->toUBytes)) {
```
markusicu, line 3148:
```suggestion
        if(U_SUCCESS(*pErrorCode) && source==sourceLimit && lastSource<source &&
                (size_t)(source-lastSource) <= sizeof(cnv->toUBytes)) {
```

markusicu: suggesting style fixes -- but i am fairly sure that this condition is supposed to always be true; don't have time to dig into details, so ok with adding it

icu4c/source/common/ucnvmbcs.cpp
```diff
@@ -3144,13 +3144,14 @@ ucnv_MBCSGetNextUChar(UConverterToUnicodeArgs *pArgs,
...
             /* incomplete character byte sequence */
             uint8_t *bytes=cnv->toUBytes;
             cnv->toULength = static_cast<int8_t>(source - lastSource);
             do {
                 *bytes++=*lastSource++;
-            } while(lastSource<source);
+            } while(lastSource<source && bytes <= &(cnv->toUBytes[sizeof(cnv->toUBytes)-1]));
```
markusicu, line 3154: FYI only -- pretty sure that this additional condition is guaranteed true

icu4c/source/common/ushape.cpp
```diff
@@ -747,6 +748,10 @@ handleGeneratedSpaces(char16_t *dest, int32_t sourceLength,
         }
     }
 
+    if (static_cast<size_t>((sourceLength + 1)) > (std::numeric_limits<size_t>::max() / U_SIZEOF_UCHAR)) {
+        *pErrorCode = U_MEMORY_ALLOCATION_ERROR;
```
markusicu, line 752: The error code does not seem right. Set index-out-of-bounds-error instead?

icu4c/source/test/cintltst/custrtrn.c
```diff
@@ -1554,7 +1554,7 @@ static void Test_strToJavaModifiedUTF8(void) {
     p=u_strToJavaModifiedUTF8(dest, (int32_t)sizeof(dest), &length,
                               src, UPRV_LENGTHOF(src), &errorCode);
     if( U_FAILURE(errorCode) || p!=dest ||
-        length!=UPRV_LENGTHOF(expected) || 0!=memcmp(dest, expected, length) ||
+        length!=(UPRV_LENGTHOF(expected)-1) || 0!=memcmp(dest, expected, length) ||
```
markusicu, line 1557: Why do we need to NUL-terminate expected when we never look at the NUL?

FrankYFTang: strstr expect both argument to be null terminiated, if there are no "\xc0\x80" inside expected, strstr may look at memory pass the end of the memory allocated for expected.

icu4c/source/common/ucnvmbcs.cpp
FrankYFTang: removed

FrankYFTang:
@markusicu 
I have trouble to reproduce the following in my local machine and understand these messages. Do you have any clue how to fix it?

```
In file included from ../../../icu4c/source/tools/toolutil/ppucd.cpp:21:
In member function ‘icu_76::UniProps& icu_76::UniProps::operator=(const icu_76::UniProps&)’,
    inlined from ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’ at ../../../icu4c/source/tools/toolutil/ppucd.cpp:239:25:
../../../icu4c/source/tools/toolutil/ppucd.h:43:23: error: writing 16 bytes into a region of size 0 [-Werror=stringop-overflow=]
   43 | struct U_TOOLUTIL_API UniProps {
      |                       ^~~~~~~~
../../../icu4c/source/tools/toolutil/ppucd.h: In member function ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’:
../../../icu4c/source/tools/toolutil/ppucd.h:49:13: note: at offset 0 to object ‘icu_76::UniProps::start’ with size 4 declared here
   49 |     UChar32 start, end;
      |             ^~~~~
In member function ‘icu_76::UniProps& icu_76::UniProps::operator=(const icu_76::UniProps&)’,
    inlined from ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’ at ../../../icu4c/source/tools/toolutil/ppucd.cpp:235:25:
../../../icu4c/source/tools/toolutil/ppucd.h:43:23: error: writing 16 bytes into a region of size 0 [-Werror=stringop-overflow=]
   43 | struct U_TOOLUTIL_API UniProps {
      |                       ^~~~~~~~
../../../icu4c/source/tools/toolutil/ppucd.h: In member function ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’:
../../../icu4c/source/tools/toolutil/ppucd.h:49:13: note: at offset 0 to object ‘icu_76::UniProps::start’ with size 4 declared here
   49 |     UChar32 start, end;
      |             ^~~~~
In member function ‘icu_76::UniProps& icu_76::UniProps::operator=(const icu_76::UniProps&)’,
    inlined from ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’ at ../../../icu4c/source/tools/toolutil/ppucd.cpp:248:21:
../../../icu4c/source/tools/toolutil/ppucd.h:43:23: error: writing 16 bytes into a region of size 0 [-Werror=stringop-overflow=]
   43 | struct U_TOOLUTIL_API UniProps {
      |                       ^~~~~~~~
../../../icu4c/source/tools/toolutil/ppucd.h: In member function ‘const icu_76::UniProps* icu_76::PreparsedUCD::getProps(icu_76::UnicodeSet&, UErrorCode&)’:
../../../icu4c/source/tools/toolutil/ppucd.h:49:13: note: at offset 0 to object ‘icu_76::UniProps::start’ with size 4 declared here
   49 |     UChar32 start, end;
      |             ^~~~~
cc1plus: all warnings being treated as errors
*** Failed compilation command follows: ----------------------------------------------------------
```

icu4c/source/common/ucnvmbcs.cpp
```diff
@@ -3144,7 +3144,8 @@ ucnv_MBCSGetNextUChar(UConverterToUnicodeArgs *pArgs,
     }
 
     if(c<0) {
-        if(U_SUCCESS(*pErrorCode) && source==sourceLimit && lastSource<source) {
+        if(U_SUCCESS(*pErrorCode) && source==sourceLimit && lastSource<source &&
+           (size_t)(source-lastSource) <= sizeof(cnv->toUBytes)) {
```
markusicu, line 3148:
```suggestion
                (size_t)(source-lastSource) <= sizeof(cnv->toUBytes)) {
```
(indentation)

icu4c/source/test/cintltst/custrtrn.c
markusicu: It will never actually run past the end because there *is* C0 80 in that array... but I understand. tnx

icu4c/source/common/ucnvmbcs.cpp
FrankYFTang: Actually, I have to add them back to make my build work. Here are the errors I got if I remove them in my local build

```make[1]: Entering directory '/usr/local/google/home/ftang/icu/icu/icu4c/source/common'
   g++	 ...  ucnvmbcs.cpp
ucnvmbcs.cpp: In function ‘UChar32 ucnv_MBCSGetNextUChar(UConverterToUnicodeArgs*, UErrorCode*)’:
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
In file included from ucnvmbcs.cpp:55:
ucnv_bld.h:190:13: note: at offset 16 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 16 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 17 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [1, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 17 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 18 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [2, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 18 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 19 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [3, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 19 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 20 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [4, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 20 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 21 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [5, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 21 into destination object ‘UConverter::toUBytes’ of size 7
ucnvmbcs.cpp:3153:25: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
 3153 |                 *bytes++=*lastSource++;
      |                 ~~~~~~~~^~~~~~~~~~~~~~
ucnv_bld.h:190:13: note: at offset 22 into destination object ‘UConverter::toUBytes’ of size 7
  190 |     uint8_t toUBytes[UCNV_MAX_CHAR_LEN-1];/* more "toU status"; keeps the bytes of the current character */
      |             ^~~~~~~~
ucnv_bld.h:190:13: note: at offset [6, 7] into destination object ‘UConverter::toUBytes’ of size 7
ucnv_bld.h:190:13: note: at offset 22 into destination object ‘UConverter::toUBytes’ of size 7
cc1plus: all warnings being treated as errors
*** Failed compilation command follows: ----------------------------------------------------------
g++ -D_REENTRANT -DU_HAVE_ELF_H=1 -DU_HAVE_TZNAME=0 -DU_HAVE_STRTOD_L=1 -DU_HAVE_XLOCALE_H=0 -I. -DDEFAULT_ICU_PLUGINS="/usr/local/lib/icu"  -DU_ALL_IMPLEMENTATION -DU_ATTRIBUTE_DEPRECATED= -DU_COMMON_IMPLEMENTATION -O3 -Werror -W -Wall -pedantic -Wpointer-arith -Wwrite-strings -Wno-long-long -std=c++17 -c -DPIC -fPIC -o ucnvmbcs.o ucnvmbcs.cpp
--- ( rebuild with "make VERBOSE=1 all" to show all param
```

markusicu: toUBytes[] collects bytes of an incomplete byte sequence for encoding one character. With size 7 for some incomplete byte sequence, the next longer complete byte sequence would be 8 bytes long. I don't think we support any conversion where a single-character byte sequence is longer than 8 bytes.

Is it possible to figure out which converter it is using, and what the input byte sequence is?

If we have to add these checks now, then please create a follow-up ticket to investigate, and add TODO comments next to the new checks.

FrankYFTang:
PTAL

.github/workflows/icu4c.yml
```diff
@@ -185,6 +185,25 @@ jobs:
           cd $PREFIX/bin;
           LD_LIBRARY_PATH=../lib ./icuinfo
 
+  gcc-warnings-as-errors:
```
roubert, line 188: Is there any advantage in adding yet one more CI job, instead of simply adding the additonal error checking to the existing `gcc-debug-build-and-test` job?

.github/workflows/icu4c.yml
```diff
@@ -185,6 +185,25 @@ jobs:
...
+          CXXFLAGS: -Wextra -Werror -Wno-return-local-addr
+          CFLAGS: -Werror
+        run: |
+          mkdir build;
+          cd build;
+          ../icu4c/source/runConfigureICU Linux/gcc --disable-layout --disable-layoutex --prefix=$PREFIX;
+          make -j -l4.5 check;
+          make -j -l4.5 install;
+          cd $PREFIX/bin;
+          LD_LIBRARY_PATH=../lib ./icuinfo
```
roubert, line 205: What's the purpose of doing `cd` and setting `LD_LIBRARY_PATH` and then not doing anything else after that?

icu4c/source/test/cintltst/uformattedvaluetst.c
```diff
@@ -143,7 +143,7 @@ static void AssertAllPartsEqual(
 
     UErrorCode status = U_ZERO_ERROR;
 
-    char message[256];
+    char message[257];
```
roubert, line 146: Instead of changing the size of the buffer and thereby pushing the underlying problem ahead until someone feels the need to change the size of it again in the future, I propose that you use `UPRV_LENGTHOF()` to replace the hardcoded buffer size whenever the buffer is being used, here and elsewhere in this PR.

roubert (commented):
With the warnings added to the `gcc-debug-build-and-test` job, you got yet one more _maybe-uninitialized_ that'll need to be addressed.

icu4c/source/i18n/decNumber.cpp
```diff
@@ -267,8 +267,10 @@ static decNumber * decExpOp(decNumber *, const decNumber *,
 static void        decFinalize(decNumber *, decContext *, Int *, uInt *);
 static Int         decGetDigits(Unit *, Int);
 static Int         decGetInt(const decNumber *);
+#ifdef UNUSED_AND_WARNING_FUNCTIONS
```
roubert, line 270: What is this `UNUSED_AND_WARNING_FUNCTIONS` preprocessor macro? I can't find it mentioned anywhere but in this PR, it's not set anywhere, not documented anywhere.

If it's something you created just now, it needs some kind of documentation to explain its purpose, who's going to set it and when, is it a temporary thing or something that ICU4C is expected to keep using for the foreseeable future, etc. And it needs to follow ICU4C naming conventions, which is to begin with `U_`.

So depending on what the purpose here really is, it should probably be named something like `U_TO_BE_DELETED_ICU_22767` or `U_INTERNAL_DEBUG` or something else along those lines that follows the naming conventions and makes the purpose easy to guess from the name.

icu4c/source/i18n/formattedvalue.cpp
```diff
@@ -193,6 +193,12 @@ ucfpos_close(UConstrainedFieldPosition* ptr) {
 }
 
 
+// -Wreturn-local-addr first found in https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/Warning-Options.html#Warning-Options
+#if U_GCC_MAJOR_MINOR >= 409
+#pragma GCC diagnostic push
+#pragma GCC diagnostic warning "-Wreturn-local-addr"
```
roubert, line 199: This line doesn't serve any purpose, does it? Then it should be deleted.

icu4c/source/i18n/formattedvalue.cpp
```diff
@@ -193,6 +193,12 @@ ucfpos_close(UConstrainedFieldPosition* ptr) {
 }
 
 
+// -Wreturn-local-addr first found in https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/Warning-Options.html#Warning-Options
```
roubert, line 196: A much more useful comment would be to explain _why_ it's safe to return this pointer and in what way the compiler is mistaken when it warns about that being done.

icu4c/source/i18n/number_rounding.cpp
```diff
@@ -283,6 +283,10 @@ FractionPrecision Precision::constructFraction(int32_t minFrac, int32_t maxFrac)
     settings.fMaxFrac = static_cast<digits_t>(maxFrac);
     settings.fMinSig = -1;
     settings.fMaxSig = -1;
+    // The following are not used under RND_FRACTION but initialize to avoid
+    // warning.
+    settings.fPriority = UNUM_ROUNDING_PRIORITY_RELAXED;
+    settings.fRetain = false;
```
roubert, line 289: Instead of doing this, another alternative which requires less code and will continue to work even if new fields are added in the future would be to just default-initialize `settings` before setting only those values that need to be set to something other than the default:

```
    FractionSignificantSettings settings{};
```

icu4c/source/test/cintltst/utf16tst.c
```diff
@@ -397,6 +397,10 @@ static void TestNextPrevChar(void){
 
 }
 
+#if U_GCC_MAJOR_MINOR >= 406
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
```
roubert, line 402: Please add a comment to explain why the code is OK and the compiler is mistaken in warning about it.

icu4c/source/test/cintltst/utf8tst.c
```diff
@@ -449,6 +449,10 @@ static void TestNextPrevChar(void) {
     }
 }
 
+#if U_GCC_MAJOR_MINOR >= 406
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Warray-bounds"
```
roubert, line 454: This one also needs a comment.

icu4c/source/i18n/formattedvalue.cpp
FrankYFTang: The explaination about that is already stated inside the function previusly as 
```
// Note: this line triggers -Wreturn-local-addr, but it is safe because toTempString is
// defined to return memory owned by the ufmtval argument.
```
I do not think I need to move it here.

icu4c/source/i18n/formattedvalue.cpp
FrankYFTang: I can remove it. I think I hit an issue that some version of the compiler won't support the ignore and I have to add this line. But I will try to remove it to see will it still cause problem.

icu4c/source/i18n/decNumber.cpp
```diff
@@ -176,6 +176,13 @@
 /*      **  -- raise to the power                                     */
 /* ------------------------------------------------------------------ */
 
+// Notes:
+// Some of the functions are not used inside ICU and cause warnning. We
+// mark them with a macro
+// #ifdef U_NOT_USED_INSIDE_ICU_AND_CAUSE_WARNING
+// so it will not be compiled. 
+// Project which want to use them could define U_NOT_USED_INSIDE_ICU_AND_CAUSE_WARNING
+
```
roubert, line 185: I don't understand this, all of this is internal to ICU, it couldn't possibly be used from anywhere else.

If this is something that ICU can be expected to start using sometime in the future, I'd say that it'd be much better to fix it now, instead of hiding it behind `#ifdef`s that never are defined which will increase the likelihood of bitrot.

But if this isn't anything that ICU is expected to ever use, why not simply delete it instead?

FrankYFTang: @sffc explain this to me and I have hard time to explain it myself.

sffc: ICU is the source of truth for the decNumber library, a popular library for decimals in C++.

Citation: https://speleotrove.com/decimal/decnumber.html

(also see Google Chat)

FrankYFTang: ok

icu4c/source/test/cintltst/utf16tst.c
FrankYFTang: @markusicu Could you look at this test? This test is written and submitted by you in https://github.com/unicode-org/icu/commit/edac6e7206ee832a05d54b62da2a5e09ead3f907
From what I can tell, the legnth cannot be -1 for these macro, based on the macro description. 
```
 * @param i string offset, must be i<length
 * @param length string length
```

but you pass in -1 for length when calling   U16_NEXT(), U16_NEXT_OR_FFFD(),  U16_FWD_1(),  U16_FWD_N(), U16_GET(),  U16_SET_CP_LIMIT() 

I think that is one of the reason casuing the warning.

FrankYFTang:
PTAL  I removed unncessary changes

.github/workflows/icu4c.yml
```diff
@@ -67,6 +67,8 @@ jobs:
 
       - name: ICU4C with gcc
         env:
+          CXXFLAGS: -Wextra -Werror -Wno-return-local-addr
```
roubert, line 70:
```suggestion
          CXXFLAGS: -Wextra -Werror -Wno-error=return-local-addr
```

I recommend not hiding this warning entirely, but instead just not treat it as an error.

roubert (changes requested):

FrankYFTang:
During the ICUTC this morning. I agree to Markus we should hold this till 77.1

FrankYFTang:
PTAL

FrankYFTang:
@roubert PTAL

icu4c/source/i18n/number_rounding.cpp
```diff
@@ -278,23 +278,23 @@ Precision IncrementPrecision::withMinFraction(int32_t minFrac) const {
 }
 
 FractionPrecision Precision::constructFraction(int32_t minFrac, int32_t maxFrac) {
-    FractionSignificantSettings settings;
+    FractionSignificantSettings settings{};
```
markusicu, line 281: This seems fine, but why isn't it the same as the old `FractionSignificantSettings settings;`?
Either one should invoke the default constructor, right?

markusicu (approved):

icu4c/source/i18n/number_rounding.cpp
markusicu: > FractionSignificantSettings is a struct, not a class

... but these are the same thing, except that struct members default to public rather than private.

> https://en.cppreference.com/w/c/language/struct_initialization

That's for C, not for C++.

For C++ I found https://en.cppreference.com/w/cpp/language/aggregate_initialization
In particular the section “Implicitly initialized elements” there.

> FractionSignificantSettings settings{};
> 
> initializing the object of struct , which call the default constructor.

I was confused because there is no default constructor, and there are no [default member initializers](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization).
The aggregate_initialization page says “Otherwise, if the element is not a reference, the element is copy-initialized from an empty initializer list.”

So these fields get whatever null values their types have.

I find this difference between using and not using `{}` confusing -- but it's not the first thing I find confusing about modern C++.

PR #3130, opened 2024-09-05
ICU-22707 adjust UTS46 for Unicode 16

markusicu:
- I compared the ICU implementation with the UTS46 changes in Unicode [15.1](https://www.unicode.org/reports/tr46/tr46-31.html#Modifications) & [16.0](https://www.unicode.org/reports/tr46/tr46-33.html#Modifications).
  - Code & spec are not totally parallel because (a) ICU makes some optimizations and (b) where the spec has options for performing certain checks ICU always performs those but records errors in an output where callers can ignore them.
  - I checked some cases via temporary test code to make sure that the outcome was as desired.
- Added a specific check for "starts with xn--" to the code path after Punycode-decoding a label, so that we can set an invalid-ACE-label error rather than a vanilla has-hyphens-at-offsets-3-and-4 error.
  - In IDNA2008, this is checked via round-trip conversion back to Punycode. So it makes sense to me to treat this as an ACE label error.
- Added support for the new test file syntax for explicitly empty strings. (Blank fields mean "same as another field".)
- Worked around test cases where the only failure is from an empty root label. ICU always performs label length checks, but deliberately does not report the empty root label as failing that check. I filed ICU-22882 for adding a separate error flag for that.

##### Checklist

### DIFF
--- a/icu4c/source/common/uts46.cpp
+++ b/icu4c/source/common/uts46.cpp
@@ -756,7 +756,12 @@ UTS46::processLabel(UnicodeString &dest,
         if(U_FAILURE(errorCode)) {
             return labelLength;
         }
-        if(!isValid) {
+        // Unicode 15.1 UTS #46:
+        // Added an additional condition in 4.1 Validity Criteria to
+        // disallow labels such as xn--xn---epa., which do not round-trip.
+        // --> Validity Criteria new criterion 4:
+        // If not CheckHyphens, the label must not begin with “xn--”.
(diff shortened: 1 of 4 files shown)


macchiati (dismissed):

markusicu:
Hi @macchiati , thanks for the early review. I have now ported the changes to C++ (very parallel), and filled out the PR description. PTAL

markusicu:
PS: The best part is that, with the code changes and the test workaround, ICU is back to passing the UTS46 conformance test cases.

eggrobin (approved):

PR #3144, opened 2024-09-09
ICU-22843 Check libc++ version for std::basic_string_view<uint16_t>

roubert:
As was pointed out in PR #3106, the standard doesn't mandate the existence of `std::char_traits<uint16_t>` and libc++ did actually deprecate it in release 18 and delete it in release 19.

As ICU4C uses this for backward compatibility for old code that still depends on setting `-DUCHAR_TYPE=uint16_t` it'd be nice to keep using it as long as possible but once it's gone from the standard library there isn't really much useful that ICU4C could do about that so conditionally compiling the code when the standard library version is old enough should suffice for keeping backward compatibility for yet some time.

https://github.com/llvm/llvm-project/commit/cce062d

##### Checklist

### DIFF
--- a/icu4c/source/common/unicode/char16ptr.h
+++ b/icu4c/source/common/unicode/char16ptr.h
@@ -315,7 +315,7 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
 template<typename T>
 constexpr bool ConvertibleToU16StringView =
     std::is_convertible_v<T, std::u16string_view>
-#if !U_CHAR16_IS_TYPEDEF
+#if !U_CHAR16_IS_TYPEDEF && (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION < 180000)
     || std::is_convertible_v<T, std::basic_string_view<uint16_t>>
 #endif
 #if U_SIZEOF_WCHAR_T==2
@@ -330,7 +330,7 @@ namespace internal {
  */
 inline std::u16string_view toU16StringView(std::u16string_view sv) { return sv; }
 
-#if !U_CHAR16_IS_TYPEDEF
+#if !U_CHAR16_IS_TYPEDEF && (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION < 180000)
 /**
  * Basically undefined behavior but sometimes necessary conversion
  * from std::basic_string_view<uint16_t> to std::u16string_view.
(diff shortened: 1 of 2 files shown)


icu4c/source/common/unicode/char16ptr.h
```diff
@@ -315,7 +315,7 @@ inline OldUChar *toOldUCharPtr(char16_t *p) {
 template<typename T>
 constexpr bool ConvertibleToU16StringView =
     std::is_convertible_v<T, std::u16string_view>
-#if !U_CHAR16_IS_TYPEDEF
+#if !U_CHAR16_IS_TYPEDEF && (!defined(_LIBCPP_VERSION) || _LIBCPP_VERSION < 180000)
```
markusicu, line 318: Don't you want to use 190000 as the threshold?
From your explanation, libc++ 18 merely deprecated using uint16_t, and then 19 removed it.

roubert:
> […], and you do not have, e.g., string literals for uint16_t strings,

I don't have any actual data on this, but I believe there's code out there that really has `UChar[]` string constants and when those are used to initialize `UnicodeString` objects I find it helpful for our users if this behaves the same regardless of the `UCHAR_TYPE` setting, whenever possible.

> I would say just remove it;

That is, of course, also a possibility. Eventually we'll want to delete the `UCHAR_TYPE` feature altogether (the sooner, the better, if you ask me), but as long as it still exists it just seems nice to me to make the use of it as seemless as possible for as long as possible.

eggrobin:
> but as long as it still exists it just seems nice to me to make the use of it as seemless as possible for as long as possible

Fair enough; but since we are on the thin ice of nonstandard features, and we know that this is an area where implementations will actually exercise their right to break us, the tradeoff does not seem so good.

> Eventually we'll want to delete the `UCHAR_TYPE` feature altogether (the sooner, the better, if you ask me)

Relatedly and perhaps achievable sooner, `U_CHAR16_IS_TYPEDEF` should be false in any conformant compiler now…

roubert:
> […], the tradeoff does not seem so good.

Well, I don't have any strong feelings about this, but considering how small a part these few extra `#ifdef`s are of the entire backward compatibility quagmire we have here and that they'll naturally get deleted once the `UCHAR_TYPE` feature goes the way of the dodo or the standard library implementations still supporting this become too old to matter (whichever comes first), it feels to me as if the scale tips towards support.

markusicu (approved):
I am fine either way, but we should quickly move on one way or another.

eggrobin (approved):

roubert:
> I am fine either way, but we should quickly move on one way or another.

Then I'll merge this now, that requires no more waiting and it'll also tell us whether it really fixes the CIFuzz tests. Further changes can always be done later, when we know more.

roubert:
> […], so I haven't been able to verify that it really will fix the CI,

But it turns out I did a good guess there, this PR really did fix the CIFuzz tests.

PR #3150, opened 2024-09-11
ICU-22843 Disambiguate UnicodeString::readOnlyAlias() for MSVC

roubert:
Both `std::u16string_view` and `std::wstring_view` are possible matches for `UnicodeString` as a template parameter, but adding an explicit overload avoids both having to make that choice and taking the detour through creating any string view at all.

##### Checklist

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -3600,6 +3600,29 @@ public:
   static inline UnicodeString readOnlyAlias(const S &text) {
     return readOnlyAliasFromU16StringView(internal::toU16StringView(text));
   }
+
+  /**
+   * Readonly-aliasing factory method.
+   * Aliases the same buffer as the input `text`.
+   *
+   * The text will be used for the UnicodeString object, but
+   * it will not be released when the UnicodeString is destroyed.
+   * This has copy-on-write semantics:
+   * When the string is modified, then the buffer is first copied into
+   * newly allocated memory.
+   * The aliased buffer is never modified.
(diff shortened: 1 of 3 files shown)


roubert:
The test case in the first commit fails like this with MSVC:

https://github.com/unicode-org/icu/actions/runs/10816568529/job/30008039132?pr=3150

The fix in the second commit makes the test case work also with MSVC.

icu4c/source/common/unicode/unistr.h
```diff
@@ -3600,6 +3600,10 @@ class U_COMMON_API UnicodeString : public Replaceable
   static inline UnicodeString readOnlyAlias(const S &text) {
     return readOnlyAliasFromU16StringView(internal::toU16StringView(text));
   }
+  /** @internal */
```
markusicu, line 3603: It doesn't seem right to declare this internal, when we expect this version to be called for UnicodeString inputs.
Please copy the API docs from the previous new function (with adjustments) and send an addendum email reply to the icu-design thread.

markusicu:
@richgillam FYI

markusicu:
> The test case in the first commit fails like this with MSVC:

good catch!

icu4c/source/common/unicode/unistr.h
roubert: I thought of disambiguating overloads as an implementation detail, but I've changed the docstring now.

markusicu (approved):

PR #3055, opened 2024-06-29
ICU-22721 Add workflow that enforces required checks

echeran:
This PR gives instructions and a workflow so that ICU can reestablish all of the CI checks as being required while still retaining the new workflow structure that optimizes time by conditionally executing workflows based on affected file paths.

Examples: I tested this in my fork and saw the new GHA workflow work as advertised. Workflow [instance 1 log](https://github.com/echeran/icu/actions/runs/9720753145/job/26832555994?pr=66), [instance 2 log](https://github.com/echeran/icu/actions/runs/9720943698/job/26833014842?pr=66).

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22721
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [X] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [X] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [X] API docs and/or User Guide docs changed or added, if applicable

PR #3160, opened 2024-09-12
ICU-22721 Set delay for CI enforce checks job to allow other jobs to init

echeran:
CIFuzz jobs take a while to start. Until they do, within a workflow run, their neutral status and warning about "can't find configs" was causing the enforce-all-checks jobs to falsely fail.

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22721

### DIFF
--- a/.github/workflows/wait-for-checks.yml
+++ b/.github/workflows/wait-for-checks.yml
@@ -29,5 +29,9 @@ jobs:
         # It won't pass until all of the pipeline's checks pass.
         # For the ICU4J and ICU4C pipelines, this is called `CI-ICU4J` and `CI-ICU4C`.
         # These checks will return a "neutral" status, which this GH Action interprets as a failure.
+        # ClusterFuzzLite/CIFuzz appears to be spawned after a CIFuzz job finishes, but its status of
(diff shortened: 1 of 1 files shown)


markusicu:
The change looks fine, but the "docs minor fixes" ticket does not seem appropriate. Which ticket did you use for adding this check?

srl295 (approved):

echeran:
> The change looks fine, but the "docs minor fixes" ticket does not seem appropriate. Which ticket did you use for adding this check?

PR #3055 used the other release ticket, so I'm reusing that one.

One further inspection with @FrankYFTang, I think the cause is different than what I identified. It looks like the CIFuzz Github Action is spawning another job (ex: a cleanup job? a reporting job?) every time a CIFuzz job finishes. That's what's interfering with our enforce-all-checks job. We can add it to the ignore list in the enforce-all-checks config. There's nothing wrong with setting up a delay of 2 minutes, but adding `ClusterFuzzLite/CIFuzz` to the ignorelist might solve the real problem.

markusicu (approved):

PR #3161, opened 2024-09-12
ICU-22876 add U_SHOW_CPLUSPLUS_HEADER_API for C++ header-only APIs

markusicu:
This lets users who define U_SHOW_CPLUSPLUS_API=0 to not rely on unstable C++ DLL exports use C++ header-only APIs on top of stable C exports.

As discussed in the 2024-sep-12 ICU-TC meeting.

(FYI All of these visibility controls are marked `@internal`.)

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22876

### DIFF
--- a/icu4c/source/common/unicode/uset.h
+++ b/icu4c/source/common/unicode/uset.h
@@ -1325,7 +1325,7 @@ U_CAPI UBool U_EXPORT2
 uset_getSerializedRange(const USerializedSet* set, int32_t rangeIndex,
                         UChar32* pStart, UChar32* pEnd);
 
-#if U_SHOW_CPLUSPLUS_API
+#if U_SHOW_CPLUSPLUS_API || U_SHOW_CPLUSPLUS_HEADER_API
 #ifndef U_HIDE_DRAFT_API
 
 namespace U_HEADER_ONLY_NAMESPACE {
@@ -1893,6 +1893,6 @@ private:
 }  // namespace U_HEADER_ONLY_NAMESPACE
 
 #endif  // U_HIDE_DRAFT_API
-#endif  // U_SHOW_CPLUSPLUS_API
+#endif  // U_SHOW_CPLUSPLUS_API || U_SHOW_CPLUSPLUS_HEADER_API
 
 #endif  // __USET_H__
--- a/icu4c/source/common/unicode/utypes.h
+++ b/icu4c/source/common/unicode/utypes.h
@@ -54,21 +54,36 @@
  * integer and other types.
  */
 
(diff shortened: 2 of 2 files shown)


richgillam (approved):
Looks great.

By the way, I checked, and even though we prohibit use of C++ APIs, we (Apple) still (for some reason) explicitly define `U_SHOW_CPLUSPLUS_API` to be `1` in our own makefiles.  I'm going to guess it's so that we could still run the C++ unit tests, since the C unit tests don't replicate all the checks that are in the C++ ones, but that's just a guess.  There might be other reasons, and there might be things we could do to allow testing through the C++ interfaces without exposing them to the world (or maybe I'm wrong and this works fine as it is).  @pedberg-icu might be able to fill in some background here.

pedberg-icu:
> By the way, I checked, and even though we prohibit use of C++ APIs, we (Apple) still (for some reason) explicitly define `U_SHOW_CPLUSPLUS_API` to be `1` in our own makefiles. I'm going to guess it's so that we could still run the C++ unit tests, since the C unit tests don't replicate all the checks that are in the C++ ones, but that's just a guess. There might be other reasons, and there might be things we could do to allow testing through the C++ interfaces without exposing them to the world (or maybe I'm wrong and this works fine as it is). @pedberg-icu might be able to fill in some background here.

@richgillam As I recall, in the Apple makefiles, U_SHOW_CPLUSPLUS_API is set to 1 for when Apple  builds ICU and runs unit tests. But when Apple clients are building against ICU headers, they are not getting the makefile override, and they normally get the default setting of U_SHOW_CPLUSPLUS_API in utypes.h, which Apple changes to be 0. And furthermore, if a client externally overrides U_SHOW_CPLUSPLUS_API to be 1, they will get all sorts of compile warnings about that resulting in code that is not binary-compatible across updates and cannot be used in production.

richgillam:
> @richgillam As I recall, in the Apple makefiles, U_SHOW_CPLUSPLUS_API is set to 1 for when Apple builds ICU and runs unit tests. But when Apple clients are building against ICU headers, they are not getting the makefile override, and they normally get the default setting of U_SHOW_CPLUSPLUS_API in utypes.h, which Apple changes to be 0. And furthermore, if a client externally overrides U_SHOW_CPLUSPLUS_API to be 1, they will get all sorts of compile warnings about that resulting in code that is not binary-compatible across updates and cannot be used in production.

@pedberg-icu   Ah!  That totally makes sense.  I do see our local change to do that now.  Thanks for the explanation!

PR #3162, opened 2024-09-12
ICU-22722 toTitle does not use UTR 21

markusicu:
Fix API docs that claim that titlecasing functions use Unicode Technical Report #&zwnj;21 for the titlecase break iterator.
That was the case only for Unicode 3.2.
Use the wording from the Java API docs instead, which generically refers to the Unicode case mapping algorithms so that we need not adjust the docs every time this changes.

(Titlecasing has been using the word break iterator since Unicode 4.)

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22722


richgillam (approved):
Looks fine to me.

PR #3165, opened 2024-09-13
ICU-22856 Fix ICUCurrencyDisplayInfoProvider.fetchUnitPatterns() so t…

richgillam:
…hat it doesn't skip inheriting patterns

from root.

##### Checklist

- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22856

### DIFF
--- a/icu4j/main/currdata/src/main/java/com/ibm/icu/impl/ICUCurrencyDisplayInfoProvider.java
+++ b/icu4j/main/currdata/src/main/java/com/ibm/icu/impl/ICUCurrencyDisplayInfoProvider.java
@@ -325,7 +325,7 @@ public class ICUCurrencyDisplayInfoProvider implements CurrencyDisplayInfoProvid
             Map<String, String> result = unitPatternsCache;
             if (result == null) {
                 result = new HashMap<>();
-                CurrencySink sink = new CurrencySink(!fallback, CurrencySink.EntrypointTable.CURRENCY_UNIT_PATTERNS);
+                CurrencySink sink = new CurrencySink(false, CurrencySink.EntrypointTable.CURRENCY_UNIT_PATTERNS);
                 sink.unitPatterns = result;
                 rb.getAllItemsWithFallback("CurrencyUnitPatterns", sink);
                 unitPatternsCache = result;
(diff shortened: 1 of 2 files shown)


sffc:
Seems plausible, but I don't _think_ this is code I have touched before?

richgillam:
> Seems plausible, but I don't _think_ this is code I have touched before?

The last person to touch it was you, but that was seven years ago, according to `git blame`.

icu4j/main/currdata/src/main/java/com/ibm/icu/impl/ICUCurrencyDisplayInfoProvider.java
```diff
@@ -325,7 +325,7 @@ Map<String, String> fetchUnitPatterns() {
             Map<String, String> result = unitPatternsCache;
             if (result == null) {
                 result = new HashMap<>();
-                CurrencySink sink = new CurrencySink(!fallback, CurrencySink.EntrypointTable.CURRENCY_UNIT_PATTERNS);
+                CurrencySink sink = new CurrencySink(false, CurrencySink.EntrypointTable.CURRENCY_UNIT_PATTERNS);
```
sffc, line 328: Aha, I guess I did touch this code...

The `fallback` flag is supposed to prevent falling back to the root bundle. As far as I can tell, that is all it is doing here. I don't see how changing this variable would cause it to find `US-Dollar` when it didn't find it before.

richgillam: `de` doesn't define have that resource at all, and that parameter is keeping us from inheriting the default value from `root`.

I think I ended up getting rid of that parameter altogether on the C side because of problems like this, but didn't want to do anything that drastic this late in the cycle.

sffc: Which resource? `US-Dollar` definitely doesn't look like something that would be in root. Is it an alias or something not being found?

richgillam: `CurrencyUnitPatterns`.

pedberg-icu (approved):
I am reasonably familiar with this code, having looked through it (mostly on the C side) during one of the recent CLDR-ICU integrations, and also familiar with why the CurrencyUnitPatters are now inherited from root. The changes here look fine to me.

DraganBesevic (approved):

PR #3166, opened 2024-09-14
ICU-22890 Add test to show lone surrogate cause infinity loop

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22890

### DIFF
--- a/icu4c/source/i18n/messageformat2_parser.cpp
+++ b/icu4c/source/i18n/messageformat2_parser.cpp
@@ -1785,7 +1785,7 @@ UnicodeString Parser::parseText(UErrorCode& status) {
         return str;
     }
 
-    if (!(isTextChar(source[index] || source[index] == BACKSLASH))) {
+    if (!(isTextChar(source[index]) || source[index] == BACKSLASH)) {
         // Error -- text is expected here
         ERROR(parseError, status, index);
         return str;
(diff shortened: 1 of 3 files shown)


FrankYFTang:
@catamorphism and @echeran  This PR show the the problem

catamorphism:
Here's the fix:

```
$ git diff
diff --git a/icu4c/source/i18n/messageformat2_parser.cpp b/icu4c/source/i18n/messageformat2_parser.cpp
index fe5bfb48480..3d5ea6e560d 100644
--- a/icu4c/source/i18n/messageformat2_parser.cpp
+++ b/icu4c/source/i18n/messageformat2_parser.cpp
@@ -1785,7 +1785,7 @@ UnicodeString parser::parseText(UErrorCode& status) {
         return str;
     }
 
-    if (!(isTextChar(source[index] || source[index] == BACKSLASH))) {
+    if (!(isTextChar(source[index]) || source[index] == BACKSLASH)) {
         // Error -- text is expected here
         ERROR(parseError, status, index);
         return str;
```

The parentheses in the error check in `parseText()` were misplaced and the error condition was always false.

Note that #3092 fixes wide character parsing in general, but the above change fixes the infinite loop.

catamorphism:
Unfortunately, applying the fix exposes other parser bugs, which I think are fixed by #3063, which hasn't landed yet. I'll see if I can create a more minimal change set that will fix the tests that fail here.

catamorphism:
OK, see https://github.com/unicode-org/icu/pull/3167 for a complete fix. Feel free to cherry-pick the extra patch out of there and add it to this PR, or close this PR in favor of that one, whichever you prefer.

FrankYFTang:
close this PR in favor of [that one, whichever you prefer](https://github.com/unicode-org/icu/pull/3167).

PR #3167, opened 2024-09-14
ICU-22890 MF2: Add lone surrogate test to parser

catamorphism:
Also see #3166

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22890

### DIFF
--- a/icu4c/source/test/intltest/messageformat2test.cpp
+++ b/icu4c/source/test/intltest/messageformat2test.cpp
@@ -31,6 +31,8 @@ TestMessageFormat2::runIndexedTest(int32_t index, UBool exec,
     TESTCASE_AUTO(testAPISimple);
     TESTCASE_AUTO(testDataModelAPI);
     TESTCASE_AUTO(testFormatterAPI);
+    TESTCASE_AUTO(testHighLoneSurrogate);
+    TESTCASE_AUTO(testLowLoneSurrogate);
     TESTCASE_AUTO(dataDrivenTests);
     TESTCASE_AUTO_END;
 }
@@ -335,6 +337,34 @@ void TestMessageFormat2::testAPICustomFunctions() {
     // "informal" is the default formality and "length" is the default length
     assertEquals("testAPICustomFunctions", "Hello John", result);
     delete person;
+}
+
+// ICU-22890 lone surrogate cause infinity loop
+void TestMessageFormat2::testHighLoneSurrogate() {
+    IcuTestErrorCode errorCode(*this, "testHighLoneSurrogate");
+    UParseError pe = { 0, 0, {0}, {0} };
+    // Lone surrogate with only high surrogate
+    UnicodeString loneSurrogate({0xda02, 0});
+    icu::message2::MessageFormatter msgfmt1 =
+      icu::message2::MessageFormatter::Builder(errorCode)
+      .setPattern(loneSurrogate, pe, errorCode)
+      .build(errorCode);
+    UnicodeString result = msgfmt1.formatToString({}, errorCode);
+    errorCode.expectErrorAndReset(U_MF_SYNTAX_ERROR, "testHighLoneSurrogate");
+}
+
+// ICU-22890 lone surrogate cause infinity loop
+void TestMessageFormat2::testLowLoneSurrogate() {
+    IcuTestErrorCode errorCode(*this, "testLowLoneSurrogate");
+    UParseError pe = { 0, 0, {0}, {0} };
+    // Lone surrogate with only low surrogate
+    UnicodeString loneSurrogate({0xdc02, 0});
+    icu::message2::MessageFormatter msgfmt2 =
+      icu::message2::MessageFormatter::Builder(errorCode)
+      .setPattern(loneSurrogate, pe, errorCode)
+      .build(errorCode);
+    UnicodeString result = msgfmt2.formatToString({}, errorCode);
+    errorCode.expectErrorAndReset(U_MF_SYNTAX_ERROR, "testLowLoneSurrogate");
 }
 
 void TestMessageFormat2::dataDrivenTests() {
--- a/icu4c/source/test/intltest/messageformat2test.h
+++ b/icu4c/source/test/intltest/messageformat2test.h
@@ -89,6 +89,8 @@ private:
     void testMessageFormatDateTimeSkeleton(message2::TestCase::Builder&, IcuTestErrorCode&);
     void testMf1Behavior(message2::TestCase::Builder&, IcuTestErrorCode&);
 
+    void testHighLoneSurrogate(void);
+    void testLowLoneSurrogate(void);
 }; // class TestMessageFormat2
 
 U_NAMESPACE_BEGIN


catamorphism:
Note that #3063 (not landed yet) and #3092 (draft) have more general fixes for keyword lookahead and wide character parsing, but I wanted to include a minimal solution here to address the infinite loop bug.

FrankYFTang:
@echeran you review other MessageFormatter changes before. Could you also review this PR? Thanks

FrankYFTang:
Could you change the title of this PR from "Fix infinity loop in parser" to "Add lone surrogate test to MessageFormatter2 Parser" Thanks

FrankYFTang:
Could you also add similar test to Java ?

catamorphism:
@FrankYFTang First, I added checks to your ICU4C test so as to require a syntax error.

I also added a test for Java. I was hoping to add this to the shared data-driven tests, but I couldn't figure out how to escape the unpaired surrogate strings for JSON. So, the tests are separate for now.

ICU4J wasn't erroring out on this case (though there was no infinite loop either), so I fixed that -- cc @mihnita. But maybe that should be a separate PR, since the ICU4J bug isn't critical. What do you think?

icu4c/source/test/intltest/messageformat2test.cpp
```diff
@@ -263,6 +265,36 @@ void TestMessageFormat2::testAPICustomFunctions() {
...
+    UParseError pe = { 0, 0, {0}, {0} };
+    // Lone surrogate with only high surrogate
+    UnicodeString loneSurrogate({0xda02, 0});
+    icu::message2::MessageFormatter msgfmt1 =
+      icu::message2::MessageFormatter::Builder(errorCode)
+      .setPattern(loneSurrogate, pe, errorCode)
+      .build(errorCode);
+    UnicodeString result = msgfmt1.formatToString({}, errorCode);
+    assertEquals("testHighLoneSurrogate", U_MF_SYNTAX_ERROR, errorCode);
+    errorCode.reset();
```
FrankYFTang, line 280: why do we reset the errorCode here?

icu4c/source/test/intltest/messageformat2test.cpp
```diff
@@ -263,6 +265,36 @@ void TestMessageFormat2::testAPICustomFunctions() {
...
+    UParseError pe = { 0, 0, {0}, {0} };
+    // Lone surrogate with only low surrogate
+    UnicodeString loneSurrogate({0xdc02, 0});
+    icu::message2::MessageFormatter msgfmt2 =
+      icu::message2::MessageFormatter::Builder(errorCode)
+      .setPattern(loneSurrogate, pe, errorCode)
+      .build(errorCode);
+    UnicodeString result = msgfmt2.formatToString({}, errorCode);
+    assertEquals("testLowLoneSurrogate", U_MF_SYNTAX_ERROR, errorCode);
+    errorCode.reset();
```
FrankYFTang, line 295: why do we reset the errorCode here?

icu4c/source/test/intltest/messageformat2test.cpp
catamorphism: I thought that in intltest, each test has to reset the error code so it doesn't carry over to the next test -- is that not the case?

FrankYFTang (approved):
LGTM

icu4c/source/test/intltest/messageformat2test.cpp
FrankYFTang: no, errorCode is a local variable in this test.

FrankYFTang: please remove the reset

catamorphism: If I remove the `errorCode.reset()` call, then the test fails. I assume this has to do with the reference to `this`:

```
    IcuTestErrorCode errorCode(*this, "testHighLoneSurrogate");
```

and that there's some shared state that results in an error if the error code is non-success at the end of a test.

A lot of other tests in intltest have this pattern (`errorCode` declared as a local `IcuTestErrorCode` variable, and then `errorCode.reset()` at the end of the method.)

FrankYFTang: so sorry, I got confused, it should be 
```
errorCode.expectErrorAndReset(U_MF_SYNTAX_ERROR, "testHighLoneSurrogate");
```
instead of
```
assertEquals("testHighLoneSurrogate", U_MF_SYNTAX_ERROR, errorCode);
errorCode.reset();
```
same below

catamorphism:
I haven't been able to reproduce the fuzzer failure in the latest CI run yet, but I'll keep trying.

mihnita:
Unpaired surrogates are not an error, according to the spec.
We had a long argument about that a while ago, I can try to dig it out.

catamorphism:
[135a911](https://github.com/unicode-org/icu/pull/3167/commits/135a9111f2f0cc3abbd85a2d7a8d8f294603ba9f) should fix the fuzzer error.

catamorphism:
@mihnita: 
> Unpaired surrogates are not an error, according to the spec. We had a long argument about that a while ago, I can try to dig it out.

I think they're a syntax error?

```
simple-message    -> o [simple-start pattern]
simple-start -> simple-start-char / escaped-char / placeholder
simple-start-char -> content-char / "@" / "|"
```

And the definition of `content-char` excludes surrogates. Am I missing something?

icu4c/source/test/intltest/messageformat2test.cpp
catamorphism: OK, fixed in [a9a6de2](https://github.com/unicode-org/icu/pull/3167/commits/a9a6de272b531fd63818f2334d4f58f244589103)

FrankYFTang (approved):

catamorphism:
I won't merge this right away so that @mihnita has a chance to reply further re: whether unpaired surrogates are a syntax error.

mihnita:
https://github.com/unicode-org/message-format-wg/blob/main/meetings/2022/notes-2022-06-13.md

> RGN: Surrogate code points. Those are code points reserved for representing code points in UTF-16 that are beyond the first plane (BMP) of 2^16 code points.
> 
> MIH: I understand what you mean. But we also implement this is C and Java, and so on. So what should we do if we receive a message with invalid UTF-8 code points. Do we expect to replace them with the replacement character, or do we just pass them through?
> 
> RGN: I think what you're asking about, using JavaScript as a concrete example, is that a JS string is allowed to have unpaired surrogates. So the question is a question for the JS adapter / implementation, but that's not a question for the standard itself.
> 
> MIH: So we leave it to the implementation?
> 
> RGN: Yes.
> 
> MIH: Okay, that is fine with me.

If the spec ended up saying something else, I am really not happy about it.
I will try to track it down when this got in and how.

I don't want to reopen that discussion.
But it is not the job of the MF2 to validate correct / incorrect UTF sequences.
Not in the message proper.
I can live with it in the code part. But once we are in the pattern, we should not.

And I would not make such a change so very late.

One might make an argument about the C++ implementation.

But Java is UTF-16 everywhere, and it does not "explode" on improperly paired surrogates.
And was explicitly mentioned and accepted in the quoted thread.

(OK, deep in the belly of `String` there is an optimization making some strings Latin1, but that is not visible in the public APIs, it is only an implementation detail, for size)

catamorphism:
@mihnita Looks like https://github.com/unicode-org/message-format-wg/pull/290 is the PR that introduced this (from Aug. 2022).

mihnita:
I will open an issue with MessageFormat WG.

This is really unfriendly for Java, JavaScript, Windows C "wide APIs" (that take a wchar_t, which is 16 bit).

It is not the job of a formatting function to validate and reject UTF corectness, at least in the above mentioned environments.

For example in Java we have `String.format`, `java.text.MessageFormat`, `com.ibm.icu.text.MessageFormat`, all of them work fine with unpaired surrogates.

I will make my case in the WG issue.

But I am against making this change in Java with this PR, sorry.

mihnita (changes requested):
Please remove the Java changes from the PR.

markusicu:
We normally treat unpaired surrogates like unassigned code points.

catamorphism:
OK -- I've removed the Java test and changes. Needs a re-approval.

mihnita (approved):
Thank you very much!
Mihai

mihnita:
I've created https://github.com/unicode-org/message-format-wg/issues/895

catamorphism:
Sorry @mihnita -- one more approval? I had to rebase because I landed #3083 first.

mihnita (approved):

PR #3169, opened 2024-09-16
ICU-22874 Update ICU4C to use a current version of the autotools

roubert:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22874

ALLOW_MANY_COMMITS=true


icu4c/source/configure.ac
```diff
@@ -6,15 +6,16 @@ AC_COPYRIGHT([ Copyright (C) 2016 and later: Unicode, Inc. and others. License &
...
+AC_PREREQ([2.72])
 
 # Process this file with autoreconf to produce a configure script
 AC_INIT([ICU],
-        m4_esyscmd_s([sed -n 's/^[ 	]*#[ 	]*define[ 	]*U_ICU_VERSION[ 	]*"\([^"]*\)".*/\1/p' "./common/unicode/uvernum.h]"),
-        [http://icu-project.org/bugs],
-        [International Components for Unicode],
-        [http://icu-project.org])
+        m4_esyscmd_s([sed -n 's/^[ 	]*#[ 	]*define[ 	]*U_ICU_VERSION[ 	]*"\([^"]*\)".*/\1/p' common/unicode/uvernum.h]),
+        [https://icu.unicode.org/bugs],
```
srl295, line 16: it must be 2024!

srl295 (approved):

icu4c/source/configure.ac
roubert: 🤣 Updating the URLs wasn't strictly necessary to make autoconf 2.72 accept this without warnings, but I just couldn't resist sneaking that change in …

markusicu:
Thank you!
@roubert or @srl295 do we need to update the relevant LICENSE file sections for this new version?
https://github.com/unicode-org/icu/blob/main/LICENSE#L464-L525

@yumaoka FYI

roubert:
> @roubert or @srl295 do we need to update the relevant LICENSE file sections for this new version?

I can't imagine why that would be necessary (but on the other hand, I'm generally not very good at imagining such stuff).

markusicu:
> > @roubert or @srl295 do we need to update the relevant LICENSE file sections for this new version?
> 
> I can't imagine why that would be necessary (but on the other hand, I'm generally not very good at imagining such stuff).

Well, the license texts should be in sync between files that have special license and our collective LICENSE file.

I just checked, and only one character changed in the license text in one of these files
- https://github.com/unicode-org/icu/pull/3182

roubert:
> I just checked, and only one character changed in the license text in one of these files
> 
> * [ICU-22874 sync config.guess license #3182](https://github.com/unicode-org/icu/pull/3182)

OK, changing that semicolon to a comma is definitely something that I could never have imagined that it would matter,

markusicu:
It probably doesn't matter, but it's good to be in sync.

PR #3181, opened 2024-09-18
ICU-22897 Fix memory leak and int overflow

FrankYFTang:
1. Rewrite to use LocalPointer to prevent memory leak
2. Rewrite the if/else to switch to make the logic clear
3. Delete the rule if not remember inside the rule set to fix memory leak.
4. Check base value calculation to avoid int64_t overflow.
5. Add memory leak test

##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22897

### DIFF
--- a/icu4c/source/i18n/nfrule.cpp
+++ b/icu4c/source/i18n/nfrule.cpp
@@ -19,6 +19,7 @@
 
 #if U_HAVE_RBNF
 
+#include <limits>
 #include "unicode/localpointer.h"
 #include "unicode/rbnf.h"
 #include "unicode/tblcoll.h"
@@ -116,9 +117,9 @@ NFRule::makeRules(UnicodeString& description,
     // new it up and initialize its basevalue and divisor
     // (this also strips the rule descriptor, if any, off the
     // description string)
-    NFRule* rule1 = new NFRule(rbnf, description, status);
+    LocalPointer<NFRule> rule1(new NFRule(rbnf, description, status));
     /* test for nullptr */
-    if (rule1 == nullptr) {
+    if (rule1.isNull()) {
         status = U_MEMORY_ALLOCATION_ERROR;
         return;
     }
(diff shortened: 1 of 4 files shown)


richgillam (approved):
This all looks reasonable...

icu4c/source/i18n/nfrule.cpp
```diff
@@ -289,7 +290,14 @@ NFRule::parseRuleDescriptor(UnicodeString& description, UErrorCode& status)
...
                 if (c >= gZero && c <= gNine) {
-                    val = val * ll_10 + static_cast<int32_t>(c - gZero);
+                    int32_t single_digit = static_cast<int32_t>(c - gZero);
+                    if ((val > 0 && val > (std::numeric_limits<int64_t>::max() - single_digit) / 10) ||
+                        (val < 0 && val < (std::numeric_limits<int64_t>::min() - single_digit) / 10)) {
+                        // out of int64_t range
+                        status = U_PARSE_ERROR;
+                        return;
+                    }
+                    val = val * ll_10 + single_digit;
```
grhoten, line 300: Why is the math being done inconsistently here.  Why are we using 10 for part of the math and ll_10 for another part of the math?

Where is the test for this if statement? I only see a test for the memory leak.

Wouldn't it be just as easy to check if the sign flipped instead of redundantly performing this math?

FrankYFTang: The memory leak test will also cause the usan report int overflow
The sign flipped  will happen after int overflow and will cause usan report the error

grhoten: Isn't this issue applicable to Java too? This check should be kept in sync with ICU4J.

Can you please use and/or rename ll_10 too? I'm not a fan of magic numbers.  The rules are base 10, and it would be helpful to the next maintainer to be more descriptive on why 10 is used.

PR #3192, opened 2024-09-19
ICU-22900 Remove deprecated SimpleNumber truncateStart

sffc:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22900

### DIFF
--- a/icu4c/source/i18n/unicode/simplenumberformatter.h
+++ b/icu4c/source/i18n/unicode/simplenumberformatter.h
@@ -87,16 +87,6 @@ class U_I18N_API SimpleNumber : public UMemory {
     void setMaximumIntegerDigits(uint32_t maximumIntegerDigits, UErrorCode& status);
 #endif // U_HIDE_DRAFT_API
 
-#ifndef U_HIDE_DEPRECATED_API
-    /**
-     * Alias for setMaximumIntegerDigits.
-     * Will be removed after ICU 75.
-     *
-     * @deprecated ICU 75
-     */
-    void truncateStart(uint32_t maximumIntegerDigits, UErrorCode& status);
-#endif // U_HIDE_DEPRECATED_API
-
     /**
      * Pads the beginning of the number with zeros up to the given minimum number of integer digits.
      *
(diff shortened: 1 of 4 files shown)


markusicu:
> This should ship in ICU 77

I don't see why this shouldn't go into 76. We are not removing anything that has ever been stable. The two functions that are to be removed went from draft ICU 73 to deprecated ICU 75, so anyone limiting themselves to stable API has never used them.

icu4c/source/i18n/unicode/simplenumberformatter.h
```diff
@@ -78,24 +78,12 @@ class U_I18N_API SimpleNumber : public UMemory {
      */
     void roundTo(int32_t power, UNumberFormatRoundingMode roundingMode, UErrorCode& status);
 
-#ifndef U_HIDE_DRAFT_API
     /**
      * Sets the number of integer digits to the given amount, truncating if necessary.
      *
-     * @draft ICU 75
+     * @stable ICU 75
```
markusicu, line 84: We agreed to remove truncateStart() in ICU 76. I think we should in fact do so.
Of course, we should not promote this replacement to stable in 76.
Even if we pushed removing truncateStart() to 77, we should leave the promotion to the usual BRS task.

So please don't do the promotion in this PR.

icu4c/source/i18n/unicode/usimplenumberformatter.h
```diff
@@ -156,27 +156,13 @@ U_CAPI void U_EXPORT2
 usnum_setMinimumFractionDigits(USimpleNumber* unumber, int32_t minimumFractionDigits, UErrorCode* ec);
 
 
-#ifndef U_HIDE_DRAFT_API
 /**
  * Sets the number of integer digits to the given amount, truncating if necessary.
  *
- * @draft ICU 75
+ * @stable ICU 75
```
markusicu, line 162: ditto

richgillam (commented):
I think Markus is right that we shouldn't jump the gun on promoting anything or do it in a special PR.  I can take care of that the way I always do.

I'm agnostic on whether we should delete these APIs now or in 77, but I'm leaning toward Markus's position.  The longer they're in there, the longer we have for people to start using them, making it worse when we get rid of them.  And they're already marked as deprecated.  I'm not sure I see the value in keeping them around.

icu4c/source/i18n/unicode/simplenumberformatter.h
sffc: > Even if we pushed removing truncateStart() to 77, we should leave the promotion to the usual BRS task.

Ok that's reasonable

markusicu (approved):
Thanks -- please merge this today.

PR #3195, opened 2024-09-20
ICU-22903 Svace: Check values for NULL before access

ligurio:
##### Checklist

- [x] Required: Issue filed:  https://unicode-org.atlassian.net/browse/ICU-22903

### DIFF
--- a/icu4c/source/common/utrie2.cpp
+++ b/icu4c/source/common/utrie2.cpp
@@ -574,7 +574,15 @@ enumEitherTrie(const UTrie2 *trie,
                     c+=UTRIE2_DATA_BLOCK_LENGTH;
                 } else {
                     for(j=0; j<UTRIE2_DATA_BLOCK_LENGTH; ++j) {
-                        value=enumValue(context, data32!=nullptr ? data32[block+j] : idx[block+j]);
+                        if (data32!=nullptr) {
+                            value=enumValue(context, data32[block+j]);
+                        } else if (idx!=nullptr) {
+                            value=enumValue(context, idx[block+j]);
+                        } else {
+                            /* data32 and idx are not supposed to be NULL at the same time */
+                            U_ASSERT(false);
+                            return;
+                        }
                         if(value!=prevValue) {
                             if(prev<c && !enumRange(context, prev, c-1, prevValue)) {
                                 return;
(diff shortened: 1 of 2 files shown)


icu4c/source/common/utrie2.cpp
```diff
@@ -574,7 +574,15 @@ enumEitherTrie(const UTrie2 *trie,
...
-                        value=enumValue(context, data32!=nullptr ? data32[block+j] : idx[block+j]);
+                        if (data32!=nullptr) {
+                            value=enumValue(context, data32[block+j]);
+                        } else if (idx!=nullptr) {
+                            value=enumValue(context, idx[block+j]);
+                        } else {
+                            /* data32 and idx are not supposed to be NULL at the same time */
+                            U_ASSERT(false);
+                            return;
+                        }
```
markusicu, line 585: No.
- A UTrie2 either stores 32-bit values (via data32), or it stores 16-bit values (via idx). One or the other is not nullptr.
- If this was ever not the case, then we would get a crash -- and we don't.
- If we did want to make this super-duper-safe, but less readable, then we would need to add real error handling here. I don't see it warranted given the invariants.

icu4c/source/i18n/unum.cpp
```diff
@@ -177,7 +177,9 @@ unum_clone(const UNumberFormat *fmt,
     } else {
         const RuleBasedNumberFormat* rbnf = dynamic_cast<const RuleBasedNumberFormat*>(nf);
         U_ASSERT(rbnf != nullptr);
-        res = rbnf->clone();
+        if (rbnf != nullptr) {
+            res = rbnf->clone();
+        }
```
markusicu, line 182: This does not look right. If the input was a nullptr, then the following code would misleadingly report a memory-allocation-error.

I am fine with guarding against fmt==nullptr although it's a little weird to call a cloning function with a nullptr.
We would need to look at other C API uprefix_clone() functions for consistency:
We could detect fmt==nullptr early, right after the U_FAILURE check, and either just return nullptr, or maybe set a U_ILLEGAL_ARGUMENT_ERROR.

markusicu (changes requested):

icu4c/source/i18n/unum.cpp
ligurio: > We would need to look at other C API uprefix_clone() functions for consistency:
> We could detect fmt==nullptr early, right after the U_FAILURE check, and either just return nullptr, or maybe set a U_ILLEGAL_ARGUMENT_ERROR.

I don't have a spare time to rework a patch this way, sorry.

PR #3203, opened 2024-09-22
ICU-22912 fix coverity warning in gencnval.c

mike-fabian:
See: https://unicode-org.atlassian.net/browse/ICU-22912

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22912

### DIFF
--- a/icu4c/source/tools/gencnval/gencnval.c
+++ b/icu4c/source/tools/gencnval/gencnval.c
@@ -273,6 +273,12 @@ main(int argc, char* argv[]) {
 
     const char* sourcedir = options[SOURCEDIR].value;
     if (sourcedir != NULL && *sourcedir != 0) {
+        if (strlen(sourcedir) + strlen(path) + 2 > sizeof(pathBuf)) {
+            fprintf(stderr,
+                    "The source file name is too long, it must be less than %d in bytes.\n",
+                    (int) sizeof(pathBuf) - 1);
+            exit(U_ILLEGAL_ARGUMENT_ERROR);
+        }
         char *end;
         uprv_strcpy(pathBuf, sourcedir);
         end = uprv_strchr(pathBuf, 0);


markusicu (approved):
Thanks for the straightforward fix in old-fashioned code. We should really switch these tools to using std::string -- or our own, internal icu::CharString which has a nice function for appending a path part.

icu4c/source/tools/gencnval/gencnval.c
```diff
@@ -273,6 +273,11 @@ main(int argc, char* argv[]) {
 
     const char* sourcedir = options[SOURCEDIR].value;
     if (sourcedir != NULL && *sourcedir != 0) {
+        if (strlen(sourcedir) + strlen(path) + 1 >= 512) {
+            fprintf(stderr,
+                    "Length of sourcedir + path is too long, must be <= 510.\n");
```
grhoten, line 278: I don't like magic numbers because updates can get these values out of sync. Please switch `>= 512` to `> sizeof(pathBuf)` so that the relationship is clear. You also already have the +1 for the null.

Please also provide a more informative error message that a person unfamiliar with the tool can understand. For example, this might be more informative:

```
 fprintf(stderr, "The source filename is too long. It must be less than %i in bytes.\n", sizeof(pathBuf) - 1);
```

Optionally, you can consider switching the size of pathBuf to MAX_PATH/PATH_MAX from limits.h, but that may take time to get right from all of the platforms. A length smaller than necessary is fine, since most of the ICU tools should not be used normally.

grhoten (changes requested):

icu4c/source/tools/gencnval/gencnval.c
grhoten: Markus's suggestion is also an acceptable path forward.

mike-fabian: > I don't like magic numbers because updates can get these values out of sync. Please switch `>= 512` to `> sizeof(pathBuf)` so that the relationship is clear. You also already have the +1 for the null.

On top of the null byte at the end, a byte for the file separation character might be needed:

```
        if(*(end-1)!=U_FILE_SEP_CHAR) {
            *(end++)=U_FILE_SEP_CHAR;
        }
```

right?

mike-fabian: > > I don't like magic numbers because updates can get these values out of sync. Please switch `>= 512` to `> sizeof(pathBuf)` so that the relationship is clear. You also already have the +1 for the null.
> 
> On top of the null byte at the end, a byte for the file separation character might be needed:
> 
> ```
>         if(*(end-1)!=U_FILE_SEP_CHAR) {
>             *(end++)=U_FILE_SEP_CHAR;
>         }
> ```
> 
> right?

So this should work, I think:

```c
    if (sourcedir != NULL && *sourcedir != 0) {
        if (strlen(sourcedir) + strlen(path) + 2 > sizeof(pathbuf)) {
            fprintf(stderr,
                    "The source file name is too long, it must be less than %d in bytes.\n",
                    sizeof(pathbuf) - 1);
            exit(U_ILLEGAL_ARGUMENT_ERROR);
        }
        char *end;
        uprv_strcpy(pathBuf, sourcedir);
        end = uprv_strchr(pathBuf, 0);
        if(*(end-1)!=U_FILE_SEP_CHAR) {
            *(end++)=U_FILE_SEP_CHAR;
        }
        uprv_strcpy(end, path);
        path = pathBuf;
    }
```

markusicu: Yes. Using sizeof addresses George's well-founded dislike of magic numbers.
Doesn't matter whether you test `+ 1 >= sizeof` or `+ 2 > sizeof`.

In the error message, you will need to cast the value to (int) so that it fits the %d.

We can switch this code to icu::CharString in a follow-up PR.

grhoten: Yes the proposed fix in the comment works for me too.

mike-fabian: > Yes the proposed fix in the comment works for me too.

I pushed again with the fix as in the comment (adding the cast to int and fixing the typos, pathbuf -> pathBuf).

grhoten (approved):

markusicu (approved):

PR #3198, opened 2024-09-21
ICU-22908 MF2: Update spec tests and update implementation for recent spec changes

catamorphism:
Updating the spec tests requires two implementation changes:
* Match on variables rather than expressions -- see https://github.com/unicode-org/message-format-wg/pull/877
* Require attribute values to be literals (if present) -- see https://github.com/unicode-org/message-format-wg/pull/894

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22908
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [x] API docs and/or User Guide docs changed or added, if applicable

PR #3206, opened 2024-09-22
ICU-22908 DRAFT!!! MF2 ICU4J: Update spec tests and update implementation for recent spec changes

mihnita:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22908


mihnita:
Updated the Java code to pass the tests equivalent to https://github.com/unicode-org/icu/pull/3198

The .json files are not identical, the ones here "bring back" the `javaIgnore`.
Many of them (close to 60) are about issue 782, which is about error handling (fallback vs strict).

That is implemented, but it is an either / or, with a flag.
And I didn't update the test framework to do that.

TLDR: if we merge the 2 PRs into one, then it should contain these updated .json files, not only the Java code changes.

---

I still think it is a bad idea to try and push for this soo late in the game.

icu4j/main/core/src/main/java/com/ibm/icu/message2/MFParser.java
```diff
@@ -571,17 +571,18 @@ private MFDataModel.SelectMessage getMatch(List<MFDataModel.Declaration> declara
...
                 break;
             }
-            checkCondition(
-                    !(expression instanceof MFDataModel.Markup), "Cannot do selection on markup");
+            MFDataModel.VariableRef variableRef = getVariableRef();
+            if (variableRef == null) {
+                break;
+            }
+            MFDataModel.Expression expression =
+                    new MFDataModel.VariableExpression(variableRef, null, new ArrayList<>());
```
catamorphism, line 585: Is it possible to change the data model so that selectors are constrained to be `Variable`s rather than `Expression`s? (If you're trying to avoid a public API change, I understand... although that means it's possible to construct a data model that doesn't correspond to any syntactically correct message.)

testdata/message2/spec/syntax-errors.json
```diff
@@ -122,6 +122,10 @@
     {
       "src": "bad {:placeholder @attribute=@foo}"
     },
+    {
+      "src": "bad {:placeholder @attribute=$foo}",
+      "ignoreJava": "Unclear why would this be an error?"
```
catamorphism, line 127: See https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf#L38

macchiati (dismissed):

PR #3214, opened 2024-09-25
ICU-22919 Enable CI with JDK 21 and fix the json parsing

mihnita:
At the code of the problem is a change in how Java parses strings to `Double`:

```
System.out.println(Double.valueOf("1234567890123456789"));
```

The results are:
jdk 17 : `1.23456789012345677E18`
jdk 21 : `1.2345678901234568E18`

The json parser was seeing a number and parsed it to a `Double`.
With the change the json parser sees a string and passes that on to our test suite, and the string is parsed / interpreted by our code.

So we just took control from the json parser.

Java supports a `long value = 1234567890123456789L`, there is no rounding, and the result from `MessageFormatter` is `"1,234,567,890,123,456,789"`, as one would expect.

So the current fix does not need to change any code in `MessageFormatter`.

It only bypasses the json parsing of numbers and informs the test code on how to parse the string value.

Before (json parser in control):
  `value: 1234567890123456789`
After (our test suite in control):
  `value: {"decimal": "1234567890123456789"}`

##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22919

### DIFF
--- a/.github/workflows/icu4j.yml
+++ b/.github/workflows/icu4j.yml
@@ -67,7 +67,7 @@ jobs:
         run: |
           cd icu4j;
           mvn ${SHARED_MVN_ARGS} dependency:go-offline -P '!old_jdk_taglet'
-          
+
   # ICU4J build and unit test using Maven
   # https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven
   icu4j-mvn-build-and-test:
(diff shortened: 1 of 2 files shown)


cjchapman (approved):
I confirmed that all tests pass and I get `BUILD SUCCESS` when I run `mvn verify` locally with these changes. Thanks!

mihnita:
> I confirmed that all tests pass and I get BUILD SUCCESS when I run mvn verify locally with these changes. Thanks!

Thank you!
And thanks for "finding" it :-)

markusicu (approved):

PR #3222, opened 2024-09-30
ICU-22606 Create an (almost) full release from CI

mihnita:
Automates most of the tasks required for release.

There is no need to build on Windows / Ubuntu / Fedora, there (almost) no need to upload files by hand to the GitHub Release, there is no need for PGP signing, keys, etc.

Not final, close enough.

#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22606

ALLOW_MANY_COMMITS=true


markusicu (approved):
rslgtm -- I might not be the best person to review this.
@echeran @rp9-next @richgillam @srl295 @yumaoka FYI

.github/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Run
+```
+docker login ghcr.io
+```
+
+When prompted use these:
+
+* **User:** the github user
+* **Password:** the github key
```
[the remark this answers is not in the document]
echeran: Agree that the instructions are missing words that disambiguate what is currently written. I also see `secrets.DOCKER_CONTAINER_USER_NAME` and `secrets.DOCKER_CONTAINER_REGISTRY_TOKEN`, but it's not clear how to create `DOCKER_CONTAINER_REGISTRY_TOKEN`, and whether that's the key in question or not.

.github/workflows/icu4c.yml
```diff
@@ -910,9 +924,15 @@ jobs:
       - name: Zip
         run: |
           cd icu4c/source/icuexportdata
-          zip -r ../../../icuexportdata_tag-goes-here.zip .
+          zip -r ../../../icuexportdata_${{ inputs.gitReleaseTag }}.zip .
```
echeran, line 927: The suffix in the filename after the `_` underscore needs to replace all `/` slash characters with a `-` dash. The release tag names for ICU4X specific preview-releases  for the purposes of icuexportdata will have slashes in the name.

docs/processes/release/tasks/publish/release.md
```diff
@@ -0,0 +1,82 @@
+---
+layout: default
+title: Publish - Version 76.1 - WIP
```
echeran, line 3: I find the filename and the Markdown page title confusing, as well as the parent directory being called "publish". Is there a more specific name to help distinguish what this page represents, specifically?

.github/README.md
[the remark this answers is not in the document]
mihnita: I agree that it might not be obvious.
But can't really go and document github "fundamentals"

Anything "secrets.FOO" is to be found in the GitHub secrets area.
In our case https://github.com/unicode-org/icu/settings/secrets/actions

> What is "the github key"?

Sorry, will update.
It is a github token (not key) as documented here: https://cldr.unicode.org/development/maven

It can be typed (with ambition, because it is about 40 char long :-), but can definitely be copy-pasted.

.github/workflows/icu4c.yml
mihnita: I tried to keep this to a minimum, to get the exact same results as before.

The trouble is that everything else uses the ICU version (with 75.1 for Java, 75_1 for C).

But in fact this does not use the version, it uses the GitHub tag.
Which includes the `release-` part.
See the url here: https://github.com/unicode-org/icu/releases/tag/release-74-2 for example.

To really unify this I would need to also change the Makefile for icu4c (OK, the stuff used to generate the Makefile :-)

TLDR: I didn't want to shake that boat now, so close to RC.
I'm fine to do it between PR and release (if acceptable), or for the next version.

docs/processes/release/tasks/publish/release.md
mihnita: This is in the same folder and uses the same header as the `index.md` file.
And that is because this file replaces big chunks of the `index.md` file.

mihnita:
I've improved the doc a bit. I will squash at the end if you want.

.github/README.md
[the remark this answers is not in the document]
mihnita: I've updated the README.md

docs/processes/release/tasks/publish/release.md
echeran: Okay, this is a WIP anyways. Maybe after this gets merged, we can consider renaming the title (in the Markdown "frontmatter" section) as a sufficient resolution. Let's skip this for now.

echeran (approved):
LGTM

mihnita:
Thank you all!

I am happy to improve the doc / anything else.
But at least we now have in place everthing we need for the release.

mohd-akram:
Putting a README file in the `.github` directory seems to cause it to be displayed on the [project page](https://github.com/unicode-org/icu#readme) (instead of the README file in the root directory).

PR #3223, opened 2024-09-30
ICU-22933 Update PGP key information for release signing

mihnita:
#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22933


rp9-next (commented):
Judging by the size of the PGP Key block, 
I think this is an unsigned one which might cause issues if someone tries to verify the binaries signed by this key for the bot. ([Signing](https://www.phildev.net/pgp/gpgsigning.html) involves hashing the new key with a key that is already known to the world)
@roubert, was my saviour last time when I needed to get them signed.
Although I think anybody listed on the KEYS page could do it. Whether a bot should have a signed key is another question, that I don't have the answer to.

roubert:
> Judging by the size of the PGP Key block, I think this is an unsigned one […]

I've just verified that this assumption is correct, the key currently being added by this PR is only self-signed.

> Whether a bot should have a signed key is another question,

That question could be asked this way: What value does it add to the users of ICU that things are signed using a only self-signed key? I can't immediately imagine any (maybe there is some, but it's at least not obvious to me), in which case it'd be simpler for everyone involved to just not sign at all.

mihnita:
>  it'd be simpler for everyone involved to just not sign at all.

In general agree.

BUT (there is always a but :-)
Maven Central requires all artifacts to be PGP signed.

So even if we don't sign the files we post in our GitHub release, we must sign at least that part.
It's good if we do that with a signed key :-)

roubert:
> It's good if we do that with a signed key :-)

If you can find the person who knows the passphrase to the secret key then just have them contact me and I'll sign the public key.

mihnita:
> If you can find the person who knows the passphrase to the secret key then just have them contact me and I'll sign the public key.

I know it :-)
Can we consider this to be me contacting you?

Thanks,
M.

mihnita:
You can also send an email to icu to `icu-robot@unicode.org` and I will answer it, if that helps.

roubert:
> I know it :-)

OK, after our little signing session I've now updated the key in this PR with my signature.

mihnita:
@markusicu:
Are you OK if I accept the bug?
Or you want to do it?
Or leave it for the TC meeting?

Thank you,
M

mihnita:
Thank you Fredrik for signing the key!
Helpful as always.

mihnita:
And thank you Rahul for noticing and bring it up!
M

markusicu:
> @markusicu: Are you OK if I accept the bug? Or you want to do it? Or leave it for the TC meeting?

I am ok, but please send an email to the TC list, ask if anyone objects, and if not, then accept it.

----
Should this PR be redirected to the maint/maint-76 branch? Or do the workflows need it in main?

mihnita:
Both would be good.
Needed it in maint for release.
Can get it into main at any point, no rush.

I'm redirecting it to maint

markusicu (approved):

PR #3232, opened 2024-10-03
ICU-22936 Replace all ICU4C code that uses UBool as an integer

roubert:
#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22936

### DIFF
--- a/icu4c/source/i18n/scriptset.cpp
+++ b/icu4c/source/i18n/scriptset.cpp
@@ -285,19 +285,19 @@ uhash_equalsScriptSet(const UElement key1, const UElement key2) {
     return (*s1 == *s2);
 }
 
-U_CAPI int8_t U_EXPORT2
+U_CAPI int32_t U_EXPORT2
 uhash_compareScriptSet(UElement key0, UElement key1) {
     icu::ScriptSet *s0 = static_cast<icu::ScriptSet *>(key0.pointer);
     icu::ScriptSet *s1 = static_cast<icu::ScriptSet *>(key1.pointer);
     int32_t diff = s0->countMembers() - s1->countMembers();
-    if (diff != 0) return static_cast<UBool>(diff);
+    if (diff != 0) return diff;
     int32_t i0 = s0->nextSetBit(0);
     int32_t i1 = s1->nextSetBit(0);
     while ((diff = i0-i1) == 0 && i0 > 0) {
         i0 = s0->nextSetBit(i0+1);
         i1 = s1->nextSetBit(i1+1);
     }
-    return (int8_t)diff;
+    return diff;
 }
 
 U_CAPI int32_t U_EXPORT2
--- a/icu4c/source/test/cintltst/cloctst.c
+++ b/icu4c/source/test/cintltst/cloctst.c
@@ -2879,7 +2879,7 @@ static void TestDisplayNameWarning(void) {
  * starts with `prefix' plus an additional element, that is, string ==
  * prefix + '_' + x, then return 1.  Otherwise return a value < 0.
  */
-static UBool _loccmp(const char* string, const char* prefix) {
+static int32_t _loccmp(const char* string, const char* prefix) {
     int32_t slen = (int32_t)uprv_strlen(string),
             plen = (int32_t)uprv_strlen(prefix);
     int32_t c = uprv_strncmp(string, prefix, plen);
--- a/icu4c/source/test/cintltst/creststn.c
+++ b/icu4c/source/test/cintltst/creststn.c
@@ -3035,7 +3035,6 @@ tres_getString(const UResourceBundle *resB,
     const char *s8;
     UChar32 c16, c8;
     int32_t length16, length8, i16, i8;
-    UBool forceCopy;
 
     if(length == NULL) {
         length = &length16;
@@ -3053,7 +3052,7 @@ tres_getString(const UResourceBundle *resB,
     length16 = *length;
 
     /* try the UTF-8 variant of ures_getStringXYZ() */
-    for(forceCopy = false; forceCopy <= true; ++forceCopy) {
+    for (int8_t forceCopy = 0; forceCopy <= 1; ++forceCopy) {
         p8 = buffer8;
         length8 = (int32_t)sizeof(buffer8);
         if(idx >= 0) {
--- a/icu4c/source/test/cintltst/cucdtst.c
+++ b/icu4c/source/test/cintltst/cucdtst.c
@@ -431,12 +431,12 @@ compareUSets(const USet *a, const USet *b,
              const char *a_name, const char *b_name,
              UBool diffIsError) {
     /*
-     * Use an arithmetic & not a logical && so that both branches
+     * Use temporary variables so that both branches
      * are always taken and all differences are shown.
      */
-    return
-        showAMinusB(a, b, a_name, b_name, diffIsError) &
-        showAMinusB(b, a, b_name, a_name, diffIsError);
+    UBool ab = showAMinusB(a, b, a_name, b_name, diffIsError);
+    UBool ba = showAMinusB(b, a, b_name, a_name, diffIsError);
+    return ab && ba;
 }
 
 /* test isLetter(u_isapha()) and isDigit(u_isdigit()) */
--- a/icu4c/source/test/cintltst/udatatst.c
+++ b/icu4c/source/test/cintltst/udatatst.c
@@ -1139,18 +1139,7 @@ static void TestICUDataName(void)
     switch(U_CHARSET_FAMILY)
     {
     case U_ASCII_FAMILY:
-          switch((int)U_IS_BIG_ENDIAN)
-          {
-          case 1:
-                typeChar = 'b';
-                break;
-          case 0:
-                typeChar = 'l';
-                break;
-          default:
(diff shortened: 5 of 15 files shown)


icu4c/source/tools/toolutil/ucm.cpp
```diff
@@ -354,7 +354,7 @@ checkBaseExtUnicode(UCMStates *baseStates, UCMTable *base, UCMTable *ext,
         /* compare the base and extension mappings */
         cmp=compareUnicode(base, mb, ext, me);
         if(cmp<0) {
-            if(intersectBase && (intersectBase!=2 || mb->bLen>1)) {
```
eggrobin, line 357: What was going on here? Did someone expect an UBool to be 2?

eggrobin (approved):

icu4c/source/test/intltest/collationtest.cpp
```diff
@@ -1754,10 +1754,10 @@ UBool CollationTest::checkCompareTwo(const char *norm, const UnicodeString &prev
...
     // only that those two methods yield the same order.
     //
-    // Use bit-wise OR so that getMergedCollationKey() is always called for both strings.
-    if((getMergedCollationKey(prevString.getBuffer(), prevString.length(), prevKey, errorCode) |
-                getMergedCollationKey(s.getBuffer(), s.length(), key, errorCode)) ||
-            errorCode.isFailure()) {
+    // Use two variables so that getMergedCollationKey() is always called for both strings.
+    if (UBool prev = getMergedCollationKey(prevString.getBuffer(), prevString.length(), prevKey, errorCode),
+              curr = getMergedCollationKey(s.getBuffer(), s.length(), key, errorCode);
+        prev || curr || errorCode.isFailure()) {
```
richgillam, line 1760: This seems like it'd read better if we just moved the definitions of `prev` and `curr ` out of the condition of the `if` statement.

icu4c/source/test/cintltst/creststn.c
```diff
@@ -3053,7 +3053,8 @@ tres_getString(const UResourceBundle *resB,
     length16 = *length;
 
     /* try the UTF-8 variant of ures_getStringXYZ() */
-    for(forceCopy = false; forceCopy <= true; ++forceCopy) {
+    for (int iteration = 0; iteration < 2; ++iteration) {
+        forceCopy = iteration;  // Repeat loop for false & true.
```
richgillam, line 3057: Maybe `forceCopy = iteration == 1`?

richgillam (dismissed):
LGTM.  A couple nits you can ignore.

icu4c/source/test/intltest/caltest.cpp
```diff
@@ -3918,7 +3918,7 @@ void CalendarTest::TestClearMonth() {
     if (failure(status, "Calendar::get(UCAL_MONTH)")) return;
     cal->clear(UCAL_MONTH);
     assertEquals("Calendar::isSet(UCAL_MONTH) after clear(UCAL_MONTH)", false, !!cal->isSet(UCAL_MONTH));
-    assertEquals("Calendar::get(UCAL_MONTH after clear(UCAL_MONTH))", UCAL_JANUARY, !!cal->get(UCAL_MONTH, status));
+    assertEquals("Calendar::get(UCAL_MONTH after clear(UCAL_MONTH))", UCAL_JANUARY, cal->get(UCAL_MONTH, status));
```
markusicu, line 3921: FYI: looks like a copy-paste bug from the previous line which normalizes what should be a boolean

icu4c/source/test/intltest/collationtest.cpp
markusicu, line 1760: I don't find this more readable than the original.
What's wrong with a bit-wise = arithmetic OR for boolean values to force evaluation of both sides?

icu4c/source/tools/toolutil/ucm.cpp
markusicu: I don't remember the details of this code from somewhere around 2002... but I would be very surprised that I wrote this when intersectBase can never be 2.
It would be safer to change the type to int8_t.

icu4c/source/tools/toolutil/ucm.cpp
```diff
@@ -442,14 +442,6 @@ checkBaseExtBytes(UCMStates *baseStates, UCMTable *base, UCMTable *ext,
             }
             mb=base->mappings+baseMap[b];
 
-            if(intersectBase==2 && mb->bLen==1) {
```
markusicu, line 445: ditto

markusicu (changes requested):

icu4c/source/i18n/scriptset.cpp
```diff
@@ -290,14 +290,14 @@ uhash_compareScriptSet(UElement key0, UElement key1) {
     icu::ScriptSet *s0 = static_cast<icu::ScriptSet *>(key0.pointer);
     icu::ScriptSet *s1 = static_cast<icu::ScriptSet *>(key1.pointer);
     int32_t diff = s0->countMembers() - s1->countMembers();
-    if (diff != 0) return static_cast<UBool>(diff);
+    if (diff != 0) return diff;
```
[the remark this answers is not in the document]
roubert: > Or change the return type to int32_t, as you do with _loccmp() in another file.

Done.

icu4c/source/test/cintltst/cloctst.c
```diff
@@ -2879,7 +2879,7 @@ static void TestDisplayNameWarning(void) {
  * starts with `prefix' plus an additional element, that is, string ==
  * prefix + '_' + x, then return 1.  Otherwise return a value < 0.
  */
-static UBool _loccmp(const char* string, const char* prefix) {
+static int32_t _loccmp(const char* string, const char* prefix) {
```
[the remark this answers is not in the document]
roubert: I chose `int32_t` because that's the normal integer type in ICU4C, an attempt to signal that this is "just an integer, nothing special".

icu4c/source/test/cintltst/creststn.c
roubert: > `for(int8_t forceCopy = 0; forceCopy <= 1; ++forceCopy) {`

Done.

icu4c/source/test/cintltst/cucdtst.c
```diff
@@ -435,7 +435,7 @@ compareUSets(const USet *a, const USet *b,
      * are always taken and all differences are shown.
      */
     return
-        showAMinusB(a, b, a_name, b_name, diffIsError) &
+        showAMinusB(a, b, a_name, b_name, diffIsError) &&
```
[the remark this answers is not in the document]
roubert: The problem with the existing code is that it won't compile if `UBool = bool`, so I've now instead changed it to use temporary variables in order to ensure that both branches are always taken.

icu4c/source/test/cintltst/udatatst.c
```diff
@@ -1139,18 +1139,7 @@ static void TestICUDataName(void)
...
-                typeChar = 'b';
-                break;
-          case 0:
-                typeChar = 'l';
-                break;
-          default:
-                log_err("Expected 1 or 0 for U_IS_BIG_ENDIAN, got %d!\n", U_IS_BIG_ENDIAN);
-                /* return; */
-          }
+          typeChar = U_IS_BIG_ENDIAN ? 'b' : 'l';
```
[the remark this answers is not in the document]
roubert: If we really want to support several kinds of endianness, the variable should be made an enum, not a boolean.

icu4c/source/test/intltest/collationtest.cpp
roubert: The problem with the existing code is that it won't compile if `UBool = bool`.

icu4c/source/tools/toolutil/ucm.cpp
roubert: > [...] but I would be very surprised that I wrote this when intersectBase can never be 2.

It totally could be, if someone were to write any code that called it in that way, it just never is.

> It would be safer to change the type to int8_t.

Done.

icu4c/source/test/cintltst/cucdtst.c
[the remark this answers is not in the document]
markusicu: (C++ rant)
> The problem with the existing code is that it won't compile if `UBool = bool`

Adding the bool type but then limiting the use of bit-wise and/or to more-than-one-bit integers is just dumb :-(

> so I've now instead changed it to use temporary variables in order to ensure that both branches are always taken.

ok

icu4c/source/test/intltest/collationtest.cpp
markusicu: optional: I agree that if we can't use the bit-wise OR any more then pulling out the two variables, like in the other file, is more readable. Mostly because of the fairly long function calls.

markusicu (approved):
thanks!

icu4c/source/test/intltest/collationtest.cpp
roubert: Maybe, but I don't think that's something that'd be worth spending another round of review/approval on.

PR #3234, opened 2024-10-04
ICU-22935 Add ZWG to static currency lists

DraganBesevic:
#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22935

Adding ZWG to few static lists. 

Followed instructions in https://unicode-org.github.io/icu/processes/cldr-icu.html#5-generate-cldr-production-data-and-convert-for-icu (step 5f).

Also, based on discussion in https://github.com/unicode-org/icu/pull/2961, there is no lists to update in Java, so only done in C++.

### DIFF
--- a/icu4c/source/common/ucurr.cpp
+++ b/icu4c/source/common/ucurr.cpp
@@ -2024,6 +2024,7 @@ static const struct CurrencyList {
     {"ZRN", UCURR_COMMON|UCURR_DEPRECATED},
     {"ZRZ", UCURR_COMMON|UCURR_DEPRECATED},
     {"ZWD", UCURR_COMMON|UCURR_DEPRECATED},
+    {"ZWG", UCURR_COMMON|UCURR_NON_DEPRECATED},
     {"ZWL", UCURR_COMMON|UCURR_DEPRECATED},
     {"ZWR", UCURR_COMMON|UCURR_DEPRECATED},
     { nullptr, 0 } // Leave here to denote the end of the list.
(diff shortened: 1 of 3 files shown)


pedberg-icu (approved):

markusicu (changes requested):
> Also, based on discussion in #2961, there is no lists to update in Java

Actually, based on that discussion, the table in NumericCodeData.java should be updated with the mapping between the letter code and the numeric code, if there is one.

For XCG at the time, there was no known numeric code.

For ZWG, https://en.wikipedia.org/wiki/Zimbabwean_ZiG says that the numeric code is 924.

Please add that.

markusicu:
... and please update https://unicode-org.github.io/icu/processes/cldr-icu.html#5-generate-cldr-production-data-and-convert-for-icu as well.

DraganBesevic:
> > Also, based on discussion in #2961, there is no lists to update in Java
> 
> Actually, based on that discussion, the table in NumericCodeData.java should be updated with the mapping between the letter code and the numeric code, if there is one.
> 
> For XCG at the time, there was no known numeric code.
> 
> For ZWG, https://en.wikipedia.org/wiki/Zimbabwean_ZiG says that the numeric code is 924.
> 
> Please add that.

Ooops! Should read it more carefully!

I will update it in the morning

DraganBesevic:
> NumericCodeData.java

Actually, Yoshito already updated that

https://github.com/unicode-org/icu/commit/74d5e6d2580fd0f888d2533033229490492d77fb

markusicu (approved):
thanks!

PR #3236, opened 2024-10-08
ICU-22940 DRAFT: MF2 ICU4C: Update for bidi support

catamorphism:
DRAFT

#### Checklist
- [ ] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22940
  - For minor changes you can use one of the following per-release tickets:
  - ICU 77 code warnings/version updates: ICU-22920 — Fix compiler warnings. Update versions of code-related dependencies (e.g., dependabot).
  - ICU 77 docs minor fixes: ICU-22921 — User Guide & API docs typos etc., and version updates (e.g., dependabot for User Guide)


icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -153,8 +159,8 @@ static bool isDigit(UChar32 c) { return inRange(c, 0x0030, 0x0039); }
 
 static bool isNameStart(UChar32 c) {
     return isAlpha(c) || c == UNDERSCORE || inRange(c, 0x00C0, 0x00D6) || inRange(c, 0x00D8, 0x00F6) ||
-           inRange(c, 0x00F8, 0x02FF) || inRange(c, 0x0370, 0x037D) || inRange(c, 0x037F, 0x1FFF) ||
-           inRange(c, 0x200C, 0x200D) || inRange(c, 0x2070, 0x218F) || inRange(c, 0x2C00, 0x2FEF) ||
+           inRange(c, 0x00F8, 0x02FF) || inRange(c, 0x0370, 0x037D) || inRange(c, 0x037F, 0x061B) ||
+           inRange(c, 0x061D, 0x200D) || inRange(c, 0x2070, 0x218F) || inRange(c, 0x2C00, 0x2FEF) ||
```
srl295, line 163: @catamorphism  i see ALM is not namestart, but this change makes U+2000…U+200B isNameStart true.  they are dashes and spaces, and not `ID_Start`

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -125,7 +125,13 @@ static bool isContentChar(UChar32 c) {
            || inRange(c, 0xE000, 0x10FFFF);
 }
 
-// See `s` in the MessageFormat 2 grammar
+// See `bidi` in the MF2 grammar
+static bool isBidi(UChar32 c) {
```
srl295, line 129: maybe isBidiControl might be better?

icu4c/source/i18n/messageformat2_parser.cpp
macchiati: BTW, I think this would be far more reliable in using a UnicodeSet. That can be created as the C++ equivalent of a static final immutable object.

catamorphism: @srl295 Fixed

@macchiati Done in [780a947](https://github.com/unicode-org/icu/pull/3236/commits/780a947d1a44d6705a2948ed0ff77504af17dc40)

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Done in [780a947](https://github.com/unicode-org/icu/pull/3236/commits/780a947d1a44d6705a2948ed0ff77504af17dc40)

srl295 (dismissed):

srl295:
needs squash but LGTM, seems like all issues addressed

catamorphism:
I added [5eb2b7d](https://github.com/unicode-org/icu/pull/3236/commits/5eb2b7d588963f45e1dcffbd2b1c7ae27623bfe6) to fix a bug that actually was in a previous PR, https://github.com/unicode-org/icu/pull/3239 , but wasn't caught when merging that PR because of a bug in the test runner.

This PR incidentally fixes that bug, so the test failure showed up here when I rebased this PR against main, and I'm fixing it here. (Will squash after @srl295 gets a chance to look at it, I left it unsquashed so far just so it's clear what I changed after he reviewed the PR.)

srl295 (approved):

catamorphism:
The fuzzer is reporting a timeout bug with this test data: https://github.com/unicode-org/icu/actions/runs/12241776312/artifacts/2295676981

I downloaded the artifact and tried running it. The test string is `"\u007b\u0000\u002f\u0000\u0067\u0020\u007d\u0000\u0000\u0000\u0000\u000c"`. I can't reproduce the timeout locally.

catamorphism:
I think I've found the cause of the timeout bug. [41994fa](https://github.com/unicode-org/icu/pull/3236/commits/41994fa1ea8a6194f26017f0e197fe8a8be62945) fixes it, if so; there's a loop in the parser that's checking for the presence of a syntax error and exiting if there already is one so that it doesn't loop infinitely, but if the `UErrorCode` is set, the operation that adds a syntax error will fail and `hasSyntaxError()` will return true. 

What I don't understand is why the `UErrorCode` is being set when running this test. I still can't reproduce the failure outside of the fuzzer, and I only have limited ability to run gdb on the fuzzer since it's chroot'ed.

srl295 (approved):
fixes LGTM

icu4c/source/i18n/messageformat2_parser.h
```diff
@@ -84,15 +105,55 @@ namespace message2 {
...
+            source(input), index(0), errors(e), normalizedInput(normalizedInputRef), dataModel(dataModelBuilder) {
+            (void) status;
 	  parseError.line = 0;
 	  parseError.offset = 0;
 	  parseError.lengthBeforeCurrentLine = 0;
 	  parseError.preContext[0] = '\0';
 	  parseError.postContext[0] = '\0';
 	}
 
+        UnicodeSet initContentChars(UErrorCode& status);
```
FrankYFTang, line 132: are the return type of these initX method "UnicodeSet" ? or it should be "UnicodeSet* "?

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+inline UnicodeSet* getImpl(Key key) {
+    return gUnicodeSets[key];
+}
+
+icu::UInitOnce gMF2ParseUniSetsInitOnce {};
+}
+
+UnicodeSet* initContentChars(UErrorCode& status) {
+    if (U_FAILURE(status)) {
+        return {};
```
FrankYFTang, line 114: why you return {} here but return nullptr later? What is the differences?

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+    }
+
+    UnicodeSet* result = new UnicodeSet();
+    if (result == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return nullptr;
+    };
+    result->addAll(*unisets::getImpl(unisets::NAME_START));
```
FrankYFTang, line 220: same as above

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+    if (U_FAILURE(status)) {
+        return {};
+    }
+
+    UnicodeSet* result = new UnicodeSet();
+    if (result == nullptr) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return nullptr;
+    };
+    result->addAll(*unisets::getImpl(unisets::CONTENT));
```
FrankYFTang, line 241: Also here

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
```
FrankYFTang, line 260: and here

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+    gUnicodeSets[unisets::NAME_CHAR] = initNameChars(status);
+    gUnicodeSets[unisets::TEXT] = initTextChars(status);
+    gUnicodeSets[unisets::QUOTED] = initQuotedChars(status);
+    gUnicodeSets[unisets::ESCAPABLE] = initEscapableChars(status);
+}
+
+const UnicodeSet* get(Key key) {
+    UErrorCode localStatus = U_ZERO_ERROR;
+    umtx_initOnce(gMF2ParseUniSetsInitOnce, &initMF2ParseUniSets, localStatus);
+    if (U_FAILURE(localStatus)) {
```
FrankYFTang, line 317: If the first time the code call initMF2ParseUniSets the error happen, then localStatus will be error and return nullptr, but the gUnicodeSets will still be partially initialize. For example, gUnicodeSets[unisets::TEXT]  may be nullptr if initNameChars failed. then later on you will pass a nullptr into contentChars and the later code contentChars->contains() will deref a nullptr

FrankYFTang:
I found the error handling code in this PR very weak. If any of the operation of new operation during initMF2ParseUniSets return nullptr due to out of memory, the code will deref nullptr later on.

icu4c/source/i18n/messageformat2_parser.h
```diff
@@ -84,15 +105,55 @@ namespace message2 {
...
+               MFDataModel::Builder& dataModelBuilder,
+               StaticErrors& e,
+               UnicodeString& normalizedInputRef,
+               UErrorCode& status)
+	  : contentChars(unisets::get(unisets::CONTENT)),
+            whitespaceChars(unisets::get(unisets::WHITESPACE)),
+            bidiControlChars(unisets::get(unisets::BIDI)),
+            alphaChars(unisets::get(unisets::ALPHA)),
+            digitChars(unisets::get(unisets::DIGIT)),
+            nameStartChars(unisets::get(unisets::NAME_START)),
```
FrankYFTang, line 118: notice, with the current code, some of these unisets::get() may return nullptr while other not and cause nullptr deref later

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+    UnicodeSet* result = new UnicodeSet(UnicodeString("[:number:]"), status);
+    if (U_FAILURE(status)) {
+        return {};
+    }
+    result->freeze();
+    return result;
+}
+
+UnicodeSet* initNameStartChars(UErrorCode& status) {
+    if (U_FAILURE(status)) {
```
FrankYFTang, line 183: you have an implicit requirement that initNameStartChars must be called after initAlpha or it may crash here. I think this is dangerous. Could we enforce that explicitly, for example, by calling initAlpha in the beginning of  initNameStartChars() ? Also, there is an implicit assumption that the status contains the status passed to initAlpha so if initAlpha fail the status here will contain that failure.

icu4c/source/i18n/messageformat2_parser.h
catamorphism: Should be `UnicodeSet*`, but these declarations aren't actually necessary as the functions are defined before they're used.

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: There is no difference, but I'll change it to `nullptr` for consistency.

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+        return {};
+    }
+
+    UnicodeSet* result = new UnicodeSet(*unisets::getImpl(unisets::ALPHA));
```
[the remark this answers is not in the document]
catamorphism: Will do this in a future PR.

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Will fix in a future PR

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Will fix in a future PR

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Will fix in a future PR

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Will fix in a future PR (by making `get()` take a `UErrorCode`).

icu4c/source/i18n/messageformat2_parser.cpp
catamorphism: Will fix in a future PR

icu4c/source/i18n/messageformat2_parser.cpp
```diff
@@ -93,14 +96,235 @@ static void copyContext(const UChar in[U_PARSE_CONTEXT_LEN], UChar out[U_PARSE_C
...
+
+void U_CALLCONV initMF2ParseUniSets(UErrorCode& status) {
+    ucln_i18n_registerCleanup(UCLN_I18N_MF2_UNISETS, cleanupMF2ParseUniSets);
+
+    gUnicodeSets[unisets::CONTENT] = initContentChars(status);
+    gUnicodeSets[unisets::WHITESPACE] = initWhitespace(status);
+    gUnicodeSets[unisets::BIDI] = initBidiControls(status);
+    gUnicodeSets[unisets::ALPHA] = initAlpha(status);
+    gUnicodeSets[unisets::DIGIT] = initDigits(status);
+    gUnicodeSets[unisets::NAME_START] = initNameStartChars(status);
```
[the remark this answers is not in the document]
catamorphism: Will fix in a future PR

catamorphism:
@FrankYFTang I submitted https://github.com/unicode-org/icu/pull/3306 to apply the fixes you suggested. Thanks!

PR #3237, opened 2024-10-08
ICU-22789 Add Segmenter API to conveniently wrap BreakIterator in ICU4J

echeran:
In order to "modernize" the `BreakIterator` API, this PR introduces a new  wrapper using a more convenient, modern API design around a `Segmenter` interface.

A few of the goals that motivate the new `Segmenter` API:

* Use newer Java features from Java 8 that support the `Stream` API which underlies a functional programming style
* Create instances that are immutable (reduces complexity borne of statefulness; allows user code to be more referentially transparent)
* Create a wrapper class around the iteration. This allows the decoupling of the iteration of a source string from the construction of the BreakIterator such that we can perform iteration over one string in isolation from other strings
* Use interfaces to properly decouple and abstract. APIs built on top of interfaces can allow user-created implementations to participate in such higher level APIs.

More details in the [design doc](https://docs.google.com/document/d/17C02RNdwD41e-sKTHcwyM8uPJhYNvo0oNI7LBs8un58/edit?pli=1).

This PR will focus on the ICU4J side of the work.

#### Checklist
- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22789

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
@@ -1,0 +1,85 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.segmenter.Segments.IterationDirection;
+
+/**
+ * An iterator of segmentation boundaries that can operate in either the forwards or reverse
+ * direction.
+ *
+ * <p>When constructed to operate in the forwards direction, the iterator will return all boundaries
+ * that are strictly after the input index value provided to the constructor. However, when
+ * constructed to operate in the backwards direction, if the input index is already a segmentation
+ * boundary, then it will be included as the first value that the iterator returns as it iterates
+ * backwards.
+ */
+class BoundaryIteratorOfInts {
+  private BreakIterator breakIter;
+  private IterationDirection direction;
+  private int currIdx;
+
+  BoundaryIteratorOfInts(BreakIterator breakIter, CharSequence sourceSequence, IterationDirection direction, int startIdx) {
+    this.breakIter = breakIter;
+    this.direction = direction;
+
+    if (direction == IterationDirection.FORWARDS) {
+      currIdx = breakIter.following(startIdx);
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+
+      // When iterating backwards over boundaries, adjust the initial index to be the boundary
+      // that is either startIdx or else the one right before startIdx.
+      //
+      // Note: we have to set the initial index indirectly because there is no way to statelessly
+      // query whether an index is on a boundary. Instead, BreakIterator.isBoundary() will mutate
+      // state when the input is not on a boundary, before it returns the value indicating a
+      // boundary.
+      int sourceLength = sourceSequence.length();
+      if (startIdx == 0) {
+        currIdx = breakIter.first();
+      } else if (startIdx == sourceLength) {
+        currIdx = breakIter.last();
+      } else {
+        boolean isOnBoundary =
+            0 <= startIdx
+                && startIdx <= sourceLength
+                && breakIter.isBoundary(startIdx);
+
+        // The previous call to BreakIterator.isBoundary(startIdx) will have advanced breakIter's
+        // current position forwards to the next boundary if the argument, startIdx, is not a
+        // boundary. Therefore, in that case, we have to move back to the previous boundary.
+        //
+        // BreakIterator.isBoundary(startIdx) should have cached the surrounding 2 boundaries in the
+        // BreakIterator, which means that BreakIterator.preceding(startIdx) shouldn't cost
+        // significant extra time.
+        //
+        // BreakIterator.preceding(startIdx) is used in initialization instead of a simple call to
+        // BreakIterator.previous() since BreakIterator.preceding() can accept arguments larger than
+        // the last boundary and return the last boundary, whereas .previous() would return DONE.
+        // Thus, .preceding() provides symmetrical behavior to .following(), which we use in the
+        // forwards direction.
+        currIdx = isOnBoundary ? startIdx : breakIter.preceding(startIdx);
+      }
+    }
+  }
+
+  public boolean hasNext() {
+    return currIdx != BreakIterator.DONE;
+  }
+
+  public Integer next() {
+    int result = currIdx;
+
+    if (direction == IterationDirection.FORWARDS) {
+      currIdx = breakIter.next();
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+      currIdx = breakIter.previous();
+    }
+
+    return result;
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundarySpliterator.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundarySpliterator.java
@@ -1,0 +1,64 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.segmenter.Segments.IterationDirection;
+import java.util.Spliterator;
+import java.util.function.IntConsumer;
+
+class BoundarySpliterator implements Spliterator.OfInt {
+
+  private final BoundaryIteratorOfInts iter;
+
+  BoundarySpliterator(BreakIterator breakIter, CharSequence sourceSequence, IterationDirection direction, int startIdx) {
+    iter = new BoundaryIteratorOfInts(breakIter, sourceSequence, direction, startIdx);
+  }
+
+  @Override
+  public OfInt trySplit() {
+    // The elements of the Stream represent an iteration through a string, and is thus inherently
+    // stateful. Therefore, splitting this Stream does not make sense. Ex: splitting the Stream
+    // is tantamount to discarding the segment subtended by the end value (index into the input
+    // string) of one substream and the beginning value of the next substream.
+    return null;
+  }
+
+  @Override
+  public long estimateSize() {
+    // The number of segments per input size depends on language, script, and
+    // the content of the input string, and thus is hard to estimate without
+    // sacrificing performance. Thus, returning `Long.MAX_VALUE`, according
+    // to the API, to mean "unknown, or too expensive to compute".
+    return Long.MAX_VALUE;
+  }
+
+  @Override
+  public int characteristics() {
+    return
+        // BreakIterator always advances
+        Spliterator.DISTINCT
+        // The design of the Segmenter API is to provide an immutable view of
+        // segmentation by preventing the input string from mutating
+        // in the underlying BreakIterator.
+        | Spliterator.IMMUTABLE
+        // primitive int is non-null
+        | Spliterator.NONNULL
+        // BreakIterator always advances, and in a single direction.
+        | Spliterator.ORDERED;
+  }
+
+  @Override
+  public boolean tryAdvance(IntConsumer action) {
+    if (action == null) {
+      throw new NullPointerException();
+    }
+    if (iter.hasNext()) {
+      action.accept(iter.next());
+      return true;
+    } else {
+      return false;
+    }
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
@@ -1,0 +1,137 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.util.ULocale;
+import java.util.Locale;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/**
+ * Performs segmentation according to the rules defined for the locale.
+ */
+public class LocalizedSegmenter implements Segmenter {
+
+  private BreakIterator breakIterPrototype;
+
+  /**
+   * Returns a {@link Segments} object that encapsulates the segmentation of the input
+   * {@code CharSequence}. The {@code Segments} object, in turn, provides the main APIs to support
+   * traversal over the resulting segments and boundaries via the Java {@code Stream} abstraction.
+   * @param s input {@code CharSequence} on which segmentation is performed. The input must not be
+   *     modified while using the resulting {@code Segments} object.
+   * @return A {@code Segments} object with APIs to access the results of segmentation, including
+   *     APIs that return {@code Stream}s of the segments and boundaries.
+   * @draft ICU 78
+   */
+  @Override
+  public Segments segment(CharSequence s) {
+    return new SegmentsImpl(breakIterPrototype, s);
+  }
+
+  /**
+   * @return a builder for constructing {@code LocalizedSegmenter}
+   * @draft ICU 78
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  private LocalizedSegmenter(ULocale locale, SegmentationType segmentationType) {
+    switch (segmentationType) {
+      case LINE:
+        breakIterPrototype = BreakIterator.getLineInstance(locale);
+        break;
+      case SENTENCE:
+        breakIterPrototype = BreakIterator.getSentenceInstance(locale);
+        break;
+      case WORD:
+        breakIterPrototype = BreakIterator.getWordInstance(locale);
+        break;
+      case GRAPHEME_CLUSTER:
+        breakIterPrototype = BreakIterator.getCharacterInstance(locale);
+        break;
+    }
+  }
+
+  /**
+   * The type of segmentation to be performed. See the ICU User Guide page
+   * <a
+   * href="https://unicode-org.github.io/icu/userguide/boundaryanalysis/#four-types-of-breakiterator">Boundary Analysis</a>
+   * for further details.
+   * @draft ICU 78
+   */
+  public enum SegmentationType {
+    GRAPHEME_CLUSTER,
+    WORD,
+    LINE,
+    SENTENCE,
+  }
+
+  /**
+   * Builder for {@link LocalizedSegmenter}
+   * @draft ICU 78
+   */
+  public static class Builder {
+
+    private ULocale locale = ULocale.ROOT;
+
+    private SegmentationType segmentationType = null;
+
+    private Builder() { }
+
+    /**
+     * Set the locale for which segmentation rules will be loaded
+     * @param locale an ICU locale object
+     * @draft ICU 78
+     */
+    public Builder setLocale(ULocale locale) {
+      if (locale == null) {
+        throw new IllegalArgumentException("locale cannot be set to null.");
+      }
+      this.locale = locale;
+      return this;
+    }
+
+    /**
+     * Set the locale for which segmentation rules will be loaded
+     * @param locale a Java locale object
+     * @draft ICU 78
+     */
+    public Builder setLocale(Locale locale) {
+      if (locale == null) {
+        throw new IllegalArgumentException("locale cannot be set to null.");
+      }
+      this.locale = ULocale.forLocale(locale);
+      return this;
+    }
+
+    /**
+     * Set the segmentation type to be performed.
+     * @param segmentationType
+     * @draft ICU 78
+     */
+    public Builder setSegmentationType(SegmentationType segmentationType) {
+      if (segmentationType == null) {
+        throw new IllegalArgumentException("segmentationType cannot be set to null.");
+      }
+      this.segmentationType = segmentationType;
+      return this;
+    }
+
+    /**
+     * Builds the {@code Segmenter}
+     * @return the constructed {@code Segmenter} instance
+     * @draft ICU 78
+     */
+    public Segmenter build() {
+      if (segmentationType == null) {
+        throw new IllegalArgumentException("segmentationType is null and must be set to a specific value.");
+      }
+      return new LocalizedSegmenter(locale, segmentationType);
+    }
+
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
@@ -1,0 +1,90 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.text.RuleBasedBreakIterator;
+import java.io.InputStream;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/**
+ * Performs segmentation according to the provided rule string. The rule string must follow the
+ * same guidelines as for {@link RuleBasedBreakIterator#RuleBasedBreakIterator(String)}.
+ * @draft ICU 78
+ */
+public class RuleBasedSegmenter implements Segmenter {
+
+  private final BreakIterator breakIterPrototype;
+
+  /**
+   * Returns a {@link Segments} object that encapsulates the segmentation of the input
+   * {@code CharSequence}. The {@code Segments} object, in turn, provides the main APIs to support
+   * traversal over the resulting segments and boundaries via the Java {@code Stream} abstraction.
+   * @param s input {@code CharSequence} on which segmentation is performed. The input must not be
+   *     modified while using the resulting {@code Segments} object.
+   * @return A {@code Segments} object with APIs to access the results of segmentation, including
+   *     APIs that return {@code Stream}s of the segments and boundaries.
+   * @draft ICU 78
+   */
+  @Override
+  public Segments segment(CharSequence s) {
+    return new SegmentsImpl(breakIterPrototype, s);
+  }
+
+  /**
+   * @return a builder for constructing {@code RuleBasedSegmenter}
+   * @draft ICU 78
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  private RuleBasedSegmenter(BreakIterator breakIter) {
+    breakIterPrototype = breakIter;
+  }
+
+  /**
+   * Builder for {@link RuleBasedSegmenter}
+   * @draft ICU 78
+   */
+  public static class Builder {
+
+    private BreakIterator breakIter = null;
+
+    private Builder() { }
+
+    /**
+     * Sets the rule string for segmentation.
+     * @param rules rule string.  The rule string must follow the same guidelines as for
+     *     {@link RuleBasedBreakIterator#getInstanceFromCompiledRules(InputStream)}.
+     * @draft ICU 78
+     */
+    public Builder setRules(String rules) {
+      if (rules == null) {
+        throw new IllegalArgumentException("rules cannot be set to null.");
+      }
+      try {
+        breakIter = new RuleBasedBreakIterator(rules);
+        return this;
+      } catch (RuntimeException rte) {
+        throw new IllegalArgumentException("The provided rule string is invalid"
+            + " or there was an error in creating the RuleBasedSegmenter.", rte);
+      }
+    }
+
+    /**
+     * Builds the {@code Segmenter}
+     * @return the constructed {@code Segmenter} instance
+     * @draft ICU 78
+     */
+    public Segmenter build() {
+      if (breakIter == null) {
+        throw new IllegalArgumentException("A rule string must be set.");
+      } else {
+        return new RuleBasedSegmenter(breakIter);
+      }
+    }
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segment.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segment.java
@@ -1,0 +1,34 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+/**
+ * A simple struct to represent an element of the segmentation result. The {@code start} and
+ * {@code limit} indices correspond to {@code source}, the input {@code CharSequence} that was
+ * originally passed to the {@code Segmenter}. {@code start} and {@code limit} are inclusive and
+ * exclusive boundaries, respectively.
+ * @draft ICU 78
+ */
+public class Segment {
+  public final int start;
+  public final int limit;
+  public final int ruleStatus = 0;
+  private final CharSequence source;
+
+  Segment(int start, int limit, CharSequence source) {
+    this.start = start;
+    this.limit = limit;
+    this.source = source;
+  }
+
+  /**
+   * Returns the subsequence represented by this {@code Segment}
+   * @return a new {@code CharSequence} object that is the subsequence represented by this
+   * {@code Segment}.
+   * @draft ICU 78
+   */
+  public CharSequence getSubSequence() {
+    return source.subSequence(start, limit);
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
@@ -1,0 +1,77 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.segmenter.Segments.IterationDirection;
+import java.util.Iterator;
+
+class SegmentIterator implements Iterator<Segment> {
+  private BreakIterator breakIter;
+  private final IterationDirection direction;
+  private int start;
+  private int limit;
+  private final CharSequence source;
+
+  SegmentIterator(BreakIterator breakIter, IterationDirection direction, int startIdx, CharSequence source) {
+    this.breakIter = breakIter;
+    this.direction = direction;
+    this.source = source;
+
+    // Note: BreakIterator.isBoundary() is a stateful operation. It resets the position in the
+    // BreakIterator, and thus doesn't just return whether the input is on a boundary.
+    boolean startIdxIsBoundary = breakIter.isBoundary(startIdx);
+
+    if (direction == IterationDirection.FORWARDS) {
+      if (startIdxIsBoundary) {
+        start = startIdx;
+        limit = breakIter.next();
+      } else {
+        // if startIdx wasn't on a boundary, then the call to isBoundary will have advanced it to
+        // the next boundary, which is the limit of the segment
+        limit = breakIter.current();
+        // go back to get the start of the segment
+        start = breakIter.previous();
+        // reset current position of BreakIterator to be limit of segment
+        breakIter.isBoundary(limit);
+      }
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+      if (startIdxIsBoundary) {
+        limit = breakIter.current();
+      } else {
+        // if startIdx was not on boundary, then the breakIter state moved forward past startIdx
+        // after the call to BreakIterator.current(), so we need to move to the previous boundary
+        // before startIdx to start the iteration
+        limit = breakIter.previous();
+      }
+      start = breakIter.previous();
+    }
+  }
+
+  @Override
+  public boolean hasNext() {
+    if (direction == IterationDirection.FORWARDS) {
+      return limit != BreakIterator.DONE;
+    } else {
+      return start != BreakIterator.DONE;
+    }
+  }
+
+  @Override
+  public Segment next() {
+    Segment result = new Segment(start, limit, source);
+
+    if (direction == IterationDirection.FORWARDS) {
+      start = limit;
+      limit = breakIter.next();
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+      limit = start;
+      start = breakIter.previous();
+    }
+
+    return result;
+  }
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segmenter.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segmenter.java
@@ -1,0 +1,43 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+
+/**
+ * An interface that defines APIs for segmentation in terms of segments and boundaries, and
+ * enforces immutable stateless iteration over the segmentation result yielded from an input
+ * {@code CharSequence}.
+ *
+ * <p>{@code Segmenter} is designed to be a followup to the {@code BreakIterator} in providing
+ * segmentation functionality. {@code Segmenter} provides immutable iteration, higher level
+ * constructs like {@code Segment}s and {@code CharSequence}s as return types, and Java programmer
+ * conveniences like {@code Stream}s in its APIs.
+ *
+ * <p>Iteration over the input sequences is made immutable by separating the design into two parts,
+ * each represented by an interface. The {@code Segmenter} interface represents the construction of
+ * the object that encapsulates the segmentation logic. The {@link Segments} interface represents
+ * the result of segmentation being performed for a specific given input {@code CharSequence}.
+ * {@code Segments} APIs also provide {@code Stream}s to support iteration over the segmentation
+ * results in a stateless manner.
+ *
+ * @see Segments
+ * @see BreakIterator
+ * @draft ICU 78
+ */
+public interface Segmenter {
+
+  /**
+   * Returns a {@link Segments} object that encapsulates the segmentation of the input
+   * {@code CharSequence}. The {@code Segments} object, in turn, provides the main APIs to support
+   * traversal over the resulting segments and boundaries via the Java {@code Stream} abstraction.
+   * @param s input {@code CharSequence} on which segmentation is performed. The input must not be
+   *     modified while using the resulting {@code Segments} object.
+   * @return A {@code Segments} object with APIs to access the results of segmentation, including
+   *     APIs that return {@code Stream}s of the segments and boundaries.
+   * @draft ICU 78
+   */
+  Segments segment(CharSequence s);
+
+}
--- a/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
@@ -1,0 +1,106 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+class SegmentsImpl implements Segments {
+
+  private CharSequence source;
+
+  private BreakIterator breakIterPrototype;
+
+  SegmentsImpl(BreakIterator breakIter, CharSequence source) {
+    this.source = source;
+
+    // We are creating a clone of the Segmenter's prototype BreakIterator field so that this
+    // concrete Segments object can avoid sharing state with the other Segments object instances
+    // that get spawned from the Segmenter. This allows difference source CharSequences to be used
+    // in each Segments object.
+    //
+    // In turn, the cloned BreakIterator becomes a prototype to be stored in the Segments object,
+    // which then gets cloned and used in each of the Segments APIs' implementations. The second
+    // level of cloning that happens when the Segments object's local BreakIterator prototype
+    // gets cloned allows the iteration state to be separate whenever an Segments API is called.
+    // Otherwise, there is a chance that multiple API calls on the same Segments object might
+    // mutate the same position/index, if done concurrently.
+    breakIterPrototype = (BreakIterator) breakIter.clone();
+    // It's okay to perform .setText on the object that we want to clone later because we should
+    // then not have to call .setText on the clones.
+    breakIterPrototype.setText(source);
+  }
+
+  @Override
+  public Segment segmentAt(int i) {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+    int start;
+    int limit;
+
+    if (i < 0 || i >= source.length()) {
+      throw new IndexOutOfBoundsException(i);
+    }
+
+    boolean isBoundary = breakIter.isBoundary(i);
+
+    if (isBoundary) {
+      start = i;
+      limit = breakIter.next();
+    } else {
+      // BreakIterator.isBoundary(i) will advance forwards to the next boundary if the argument
+      // is not a boundary.
+      limit = breakIter.current();
+      start = breakIter.previous();
+    }
+
+    assert start != BreakIterator.DONE && limit != BreakIterator.DONE;
+
+    return new Segment(start, limit, source);
+  }
+
+  @Override
+  public boolean isBoundary(int i) {
+    return ((BreakIterator) breakIterPrototype.clone()).isBoundary(i);
+  }
+
+  @Override
+  public Stream<Segment> segmentsFrom(int i) {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    SegmentIterable iterable = new SegmentIterable(breakIter, IterationDirection.FORWARDS, i,
+        source);
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+
+  @Override
+  public Stream<Segment> segmentsBefore(int i) {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    SegmentIterable iterable = new SegmentIterable(breakIter, IterationDirection.BACKWARDS, i,
+        source);
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+
+  @Override
+  public IntStream boundariesAfter(int i) {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    return StreamSupport.intStream(new BoundarySpliterator(breakIter, source, IterationDirection.FORWARDS,
+        i), false);
+  }
+
+  @Override
+  public IntStream boundariesBackFrom(int i) {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    return StreamSupport.intStream(new BoundarySpliterator(breakIter, source, IterationDirection.BACKWARDS,
+        i), false);
+  }
+
+}
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
@@ -1,0 +1,64 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.dev.test.segmenter;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import com.ibm.icu.dev.test.CoreTestFmwk;
+import com.ibm.icu.segmenter.LocalizedSegmenter;
+import com.ibm.icu.segmenter.LocalizedSegmenter.SegmentationType;
+import com.ibm.icu.segmenter.Segmenter;
+import com.ibm.icu.segmenter.Segments;
+import com.ibm.icu.util.ULocale;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Locale;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class LocalizedSegmenterTest extends CoreTestFmwk {
+
+  @Test
+  public void testLocaleInLocalizedSegmenter() {
+
+    Object[][] casesData = {
+        {"de",
+            "Das 21ste Jahrh. ist das beste.",
+            Arrays.asList("Das", " ", "21ste", " ", "Jahrh", ".", " ", "ist", " ", "das", " ", "beste", ".")},
+    };
+
+    for (Object[] caseDatum : casesData) {
+      String localeTag = (String) caseDatum[0];
+      String source = (String) caseDatum[1];
+      List<CharSequence> expWords = (List<CharSequence>) caseDatum[2];
+
+      // create functions that abstract over the different types of locales so that, later on,
+      // we can test using both in the same way
+      Function<LocalizedSegmenter.Builder, LocalizedSegmenter.Builder> setJLocaleFn =
+          builder -> builder.setLocale(Locale.forLanguageTag(localeTag));
+      Function<LocalizedSegmenter.Builder, LocalizedSegmenter.Builder> setULocaleFn =
+          builder -> builder.setLocale(ULocale.forLanguageTag(localeTag));
+
+      // Loop/iterate over both Java Locale and ICU ULocale such that we always test with both
+      // types of locale objects.
+      for (Function<LocalizedSegmenter.Builder, LocalizedSegmenter.Builder> setLocIntoBuilder : Arrays.asList(setJLocaleFn, setULocaleFn)) {
+
+        LocalizedSegmenter.Builder builder = LocalizedSegmenter.builder()
+            .setSegmentationType(SegmentationType.WORD);
+        LocalizedSegmenter.Builder builderWithLocale = setLocIntoBuilder.apply(builder);
+        Segmenter wordSeg = builderWithLocale.build();
+        Segments segments = wordSeg.segment(source);
+
+        List<CharSequence> actWords = segments.subSequences().collect(Collectors.toList());
+
+        assertThat(actWords, is(expWords));
+      }
+    }
+  }
+}
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
@@ -1,0 +1,58 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.dev.test.segmenter;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import com.ibm.icu.dev.test.CoreTestFmwk;
+import com.ibm.icu.segmenter.RuleBasedSegmenter;
+import com.ibm.icu.segmenter.Segmenter;
+import com.ibm.icu.segmenter.Segments;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class RuleBasedSegmenterTest extends CoreTestFmwk {
+
+  @Test
+  public void testRules() {
+    Object[][] casesData = {
+        {"ASCII lowercase a-z only",
+            "Kühlschränke kühlen Getränke",
+            "[a-z]+;",
+            Arrays.asList("K", "ü", "hlschr", "ä", "nke", " ", "k", "ü", "hlen", " ", "G", "etr", "ä", "nke")},
+        {"ASCII upper- and lowercase a-z",
+            "Kühlschränke kühlen Getränke",
+            "[A-Za-z]+;",
+            Arrays.asList("K", "ü", "hlschr", "ä", "nke", " ", "k", "ü", "hlen", " ", "Getr", "ä", "nke")},
+        {"ASCII upper- and lowercase a-z and some letters with diaeresis",
+            "Kühlschränke kühlen Getränke",
+            "[A-Za-züä]+;",
+            Arrays.asList("Kühlschränke", " ", "kühlen", " ", "Getränke")},
+    };
+
+    for (Object[] caseDatum : casesData) {
+      String desc = (String) caseDatum[0];
+      String source = (String) caseDatum[1];
+      String rule = (String) caseDatum[2];
+      List<CharSequence> expWords = (List<CharSequence>) caseDatum[3];
+
+      Segmenter seg = RuleBasedSegmenter.builder()
+          .setRules(rule)
+          .build();
+      Segments segments = seg.segment(source);
+
+      List<CharSequence> actWords = segments.subSequences().collect(Collectors.toList());
+
+      assertThat(desc, actWords, is(expWords));
+    }
+
+  }
+
+}
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
@@ -1,0 +1,371 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.dev.test.segmenter;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import com.ibm.icu.dev.test.CoreTestFmwk;
+import com.ibm.icu.segmenter.LocalizedSegmenter;
+import com.ibm.icu.segmenter.LocalizedSegmenter.SegmentationType;
+import com.ibm.icu.segmenter.Segment;
+import com.ibm.icu.segmenter.Segmenter;
+import com.ibm.icu.segmenter.Segments;
+import com.ibm.icu.util.ULocale;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class SegmentsTest extends CoreTestFmwk {
+
+  @Test
+  public void testSegments() {
+    Segmenter enWordSegmenter =
+        LocalizedSegmenter.builder()
+            .setLocale(ULocale.ENGLISH)
+            .setSegmentationType(SegmentationType.WORD)
+            .build();
+
+    String source1 = "The quick brown fox jumped over the lazy dog.";
+
+    // Create new Segments for source1
+    Segments segments1 = enWordSegmenter.segment(source1);
+
+    List<Segment> segments = segments1.segments().collect(Collectors.toList());
+
+    assertEquals("first segment start", 0, segments.get(0).start);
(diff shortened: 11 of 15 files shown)


macchiati (approved):

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,48 @@
...
+      this.currIdx = BreakIterator.DONE;
+      return;
+    }
+
+    if (direction == IterationDirection.FORWARDS) {
+      this.currIdx = breakIter.following(startIdx);
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+      this.currIdx = breakIter.preceding(startIdx);
+    }
```
richgillam, line 29: If `startIdx` is a boundary, there's no way to tell from this API.  Is that intentional?  Is that handled somewhere else?  Do we always assume `startIdx` is a boundary?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+    this.segmentationType = segmentationType;
+  }
+
+  /**
+   * @internal
+   * @deprecated This API is ICU internal only.
+   */
+  @Override
+  @Deprecated
+  public BreakIterator getNewBreakIterator() {
```
richgillam, line 45: Why does this function need to be declared `public`?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+    private CharSequence source;
+
+    private LocalizedSegmenter segmenter;
+
+    private BreakIterator breakIter;
+
+    private LocalizedSegments(CharSequence source, LocalizedSegmenter segmenter) {
+      this.source = source;
+      this.segmenter = segmenter;
+      this.breakIter = this.segmenter.getNewBreakIterator();
```
richgillam, line 142: I'm guessing `LocalizedSegmented.getNewBreakIterator()` is public because you're calling it here.  Would it make more sense to either have it be a method on `LocalizedSegments`, or for the segmenter to call that (now-private) function and pass the result to this constructor?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,137 @@
...
+    this.rules = rules;
+  }
+
+  /**
+   * @internal
+   * @deprecated This API is ICU internal only.
+   */
+  @Override
+  @Deprecated
+  public BreakIterator getNewBreakIterator() {
```
richgillam, line 47: Same thoughts as the version of this method on `LocalizedSegmenter`.  I think you could avoid having this be public by either having the segmenter pass the break iterator to the `Segments` object on construction or by having this actually be a method on the `Segments` object.

richgillam (approved):
Looks great!  One quibble.

richgillam:
> Looks great! One quibble.

Actually, one other observation.  As things stand, `Segmenter` and its subclasses don't do much-- they just create `Segments` objects, which wrap `BreakIterator` objects that do all the work.  The API is a lot cleaner and clearer, but the implementation isn't.  I assume the plan in the future at some point is to move to an implementation where the `Segmenter` actually owns the state and category tables and the `Segments` object just handles iteration over a particular string?  (I'm not saying you need to do this now; just clarifying that that's in the plan.)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: Good observation. Yes, there is a difference between the semantics of the `Segments.backFrom(...)` public API that includes a boundary and the behavior of `BreakIterator.preceding()` that always moved backwards at least 1.

I had the logic for adjusting `startIdx` in the backwards direction condition to account for that difference is in the helper utility fn `SegmentsImplUtils.boundariesBackFrom()`.

This comment reminds me that it would make sense to have this logic closer to where it is actually acted upon, which is here in `BoundaryIteratorOfInts`. So I just now went ahead and refactored that logic out from the helper util class and into the constructor of `BoundaryIteratorOfInts`. Even though `BoundaryIteratorOfInts` is an internal helper class, the refactoring makes it behave more like the public APIs we're supporting without exposing BreakIterator impl details to private internal callers, which is cleaner, even if transparent to users.

I didn't do so already because I would need to additionally pass a reference to the input `CharSequence` to the constructor of `BoundaryIteratorOfInts`, which is technically redundant because it already needs a `BreakIterator` as an argument that theoretically contains that `CharSequence`. However, I can't get a `CharSequence` back from a `BreakIterator`; I only get back a `CharacterIterator`. Even though it's redundant, I figure the cost of adding a reference to `CharSequence` is minimal and acceptable.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: This question is a question of why the interface `Segmenter` needs to include a method `getNewBreakIterator()`, which then implies that the method is public.

My reason for including this is based on the need for isolation (separation of state) across all concrete instances of the `Segments` interface that are spawned each time that a concrete `Segmenter` object is fed a new input `CharSequence`. In order to ensure isolation in the `Segments` implementations, they need to create a new `BreakIterator` instance.

Given the way things are implemented, you could argue that the interface method `getNewBreakIterator()` could be removed without much hassle. I wanted to include it so that the interface communicates the common denominator requirements to fulfill the contract.

I wished that there could be a way to say "make this private (non-user facing) but still make it a part of the contract that has to be implemented by concrete implementing classes". The Java 9 update to interfaces to allow private methods is not fully what I want -- it only allows those private methods to be used as helpers when implementing default implementation methods, but it does not get included in the contract for implementing classes.

One the one hand, I like the principle of the current design in defining a contract for code that implements `Segmenter` & `Segments`, but on the other hand, I don't like that we have to mention `BreakIterator` again, even if it's in a deprecate public method.

I could be convinced to remove the (public) `getBreakIterator` interface method. One possible argument is that Markus mentioned the possibility of changing the internals of the concrete classes to be more efficient by rewriting them to not depend on & wrap `BreakIterator`.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: See conversation in [thread above](https://github.com/unicode-org/icu/pull/3237/files#r2049615459).

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: See conversation in [thread above](https://github.com/unicode-org/icu/pull/3237/files#r2049615459).

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
richgillam: Thank you for the explanation.  That all makes sense and sounds good.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
richgillam: It's been a fairly long time since I last through about this stuff, so I'm not sure I'm completely up on what we were talking about here, and I apologize for that.  I think I had two thoughts when I wrote that comment (although who knows now?): One was that this is an implementation detail and it's unfortunate to have it in the public API, even declared `@Deprecated`.  But, of course, you can't always avoid that in Java, and I think you've made a good case for it.  The other was that it kind of assumes that all concrete subclasses of `LocalizedSegmenter` will use a `BreakIterator` for their implementation.  If we want to move away from having `BreakIterator` inside these things at some point, or if we want clients to be able to create subclasses that do their own thing, this reference stands in the way.  I'm guessing in cases like that, you could just have it return null, and that'd be fine, but it'd still be kind of a wart.  On the other hand, if we don't think there's anything to be gained from moving away from `BreakIterator`, then it's fine the way it is and you can ignore me.  :-)

richgillam: Rereading my comments below, it looks like I was suggesting that you move the `getNewBreakIterator()` method into the classes where it's getting used and where it could be made private so that you don't have to expose it to the public API.  And I think I'm reading your response to say that you can't do that-- the information necessary to create a new break iterator lives in `Segmenter` and there isn't a good way to pass it all along.  I'll buy that-- this might just be a case where we have to live with the deprecated public method.

echeran: Actually, I think the question you raised about the possibility of future work that would reimplement the internals to bypass BreakIterator is a good question. It makes me think that your original point of excluding the public deprecated (internal) method from the interface is a good one if there is even just a slight chance of such future work. So I'm leaning in favor of removing it, and saying that doing so is the proper decision.

Perhaps the counterargument is, you could say, that it's easier to have a deprecated interface method that you can ignore in the future (and have future implementations just implement it by returning null, like you said), rather than adding a method to an interface later on. You can't really add a method to an interface or abstract class without breaking the contract for existing implementations, as [this comment in `BreakIterator.preceding()`](https://github.com/unicode-org/icu/blob/b30c63d1b930610850489a67433b9c3ba55d6f43/icu4j/main/core/src/main/java/com/ibm/icu/text/BreakIterator.java#L347-L349) points out, ironically enough.

So I think the question is back to what you said, is can Segmenter stand on its own independently, or is it defined intrinsically as a wrapper of BreakIterator? I think the answer is the former, which is why I'm leaning in favor of removing it, like you said.

Maybe @markusicu could weigh in?

markusicu: I would like the old BreakIterator and the new Segmenter to eventually _share implementation code_ without the Segmenter having to _have a_ BreakIterator. So if we can avoid mentioning BreakIterator on public Segmenter API then that would be good.

(sorry for getting here so late)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segment.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  /**
+   * Returns the subsequence represented by this {@code Segment}
+   * @return a new {@code CharSequence} object that is the subsequence represented by this
+   * {@code Segment}.
+   * @draft ICU 78
+   */
+  public CharSequence getSubSequence() {
+    return source.subSequence(this.start, this.limit);
```
markusicu: why `this.` here?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * @return a builder for constructing {@code LocalizedSegmenter}
+   * @draft ICU 78
+   */
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  LocalizedSegmenter(ULocale locale, SegmentationType segmentationType) {
```
markusicu: please try to make as many "package" methods & types private as possible -- this one for example should work

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+   * Builder for {@link LocalizedSegmenter}
+   * @draft ICU 78
+   */
+  public static class Builder {
+
+    private ULocale locale = ULocale.ROOT;
+
+    private SegmentationType segmentationType = SegmentationType.GRAPHEME_CLUSTER;
+
+    Builder() { }
```
markusicu: private?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Builder for {@link LocalizedSegmenter}
+   * @draft ICU 78
+   */
+  public static class Builder {
+
+    private ULocale locale = ULocale.ROOT;
+
+    private SegmentationType segmentationType = SegmentationType.GRAPHEME_CLUSTER;
```
markusicu: Did we discuss this default? I don't remember.
While the locale default is reasonably obvious, a default segmentation type does not seem obvious to me.
I think it would be better to init this to null and have the build() method throw an exception if it's not set.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+
+    Builder() { }
+
+    /**
+     * Set the locale for which segmentation rules will be loaded
+     * @param locale an ICU locale object
+     * @draft ICU 78
+     */
+    public Builder setLocale(ULocale locale) {
+      this.locale = locale;
```
markusicu, line 94: each setter should guard against null values; NPE or illegal argument?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,188 @@
...
+
+  class LocalizedSegments implements Segments {
+
+    private CharSequence source;
+
+    private LocalizedSegmenter segmenter;
+
+    private BreakIterator breakIter;
+
+    private LocalizedSegments(CharSequence source, LocalizedSegmenter segmenter) {
```
markusicu: nit/optional: i would swap the parameters

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  static class RuleBasedSegments implements Segments {
+    private CharSequence source;
+
+    private RuleBasedSegmenter segmenter;
+
+    private BreakIterator breakIter;
+
+    RuleBasedSegments(CharSequence source, RuleBasedSegmenter segmenter) {
```
markusicu: nit/optional: i would swap the parameters -- same as in LocalizedSegments

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,137 @@
...
+      return this;
+    }
+
+    /**
+     * Builds the {@code Segmenter}
+     * @return the constructed {@code Segmenter} instance
+     * @draft ICU 78
+     */
+    public Segmenter build() {
+      return new RuleBasedSegmenter(this.rules);
```
markusicu: This defers parsing/building the rules too much. If the rules have problems, we should throw an exception right here (or in the RuleBasedSegmenter constructor).

The RuleBasedSegmenter should have a field for the RBBI, not for the rule string. getNewBreakIterator() should clone the prototype.

markusicu (commented):
sorry for the late review!
so far i have mostly checked the API here against the proposal -- and found that the proposal omitted one method, see the design doc.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done. We didn't discuss the default. The rationale for a null for segmentation type, and how that differs from locale, makes sense. I added the validation & exception when the default null isn't overwritten.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done. I followed [this advice](https://stackoverflow.com/a/47710/2077918) that strongly advocated for illegal argument over NPE, giving multiple good reasons.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done. `getNewBreakIterator()` is no longer in the `Segmenter` interface, so the concrete method is now private.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done. Ditto.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segment.java
echeran: Done. I removed `this.` in every place where it was superfluous.

echeran (commented):
Addressed new feedback from @markusicu. PTAL.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
markusicu: tnx

nit/optional: the exception message should not need to say "In LocalizedSegmenter" because that's available via the exceptions stack trace

... and similar for other such throw statements

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -31,7 +31,7 @@ public static Builder builder() {
     return new Builder();
   }
 
-  private LocalizedSegmenter(ULocale locale, SegmentationType segmentationType) {
+  private LocalizedSegmenter(SegmentationType segmentationType, ULocale locale) {
```
markusicu: for these parameters I prefer locale then type like you had it...

my comment was about the LocalizedSegments constructor -- which you swapped in your last commit today -- not this constructor here

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -35,11 +35,17 @@ public static Builder builder() {
   }
 
   private RuleBasedSegmenter(String rules) {
-    this.rules = rules;
+    try {
+      this.prototypeRbbi = new RuleBasedBreakIterator(rules);
+    }
+    catch (Exception e) {
+      throw new IllegalArgumentException("In RuleBasedSegmenter, the provided rule string is"
+          + " invalid or there was an error in creating the RuleBasedSegmenter.");
```
markusicu: please add `e` as the chained exception

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,65 @@
...
+ * direction.
+ *
+ * <p>When constructed to operate in the forwards direction, the iterator will return all boundaries
+ * that are strictly after the input index value provided to the constructor. However, when
+ * constructed to operate in the backwards direction, if the input index is already a segmentation
+ * boundary, then it will be included as the first value that the iterator returns as it iterates
+ * backwards.
+ */
+class BoundaryIteratorOfInts {
+  private BreakIterator breakIter;
```
markusicu, line 20: please consistently use an indent of 4

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,65 @@
...
+    if (direction == IterationDirection.FORWARDS) {
+      currIdx = breakIter.following(startIdx);
+    } else {
+      assert direction == IterationDirection.BACKWARDS;
+
+      // When iterating backwards over boundaries, adjust the start index forwards by 1 to
+      // counteract the behavior from BreakIterator.preceding(), which we use to initialize the
+      // BreakIterator state, that always moves backwards by at least 1. We want to support an
+      // API that includes the input index when it is itself a boundary, unlike the behavior of
+      // BreakIterator.preceding().
```
markusicu: - first thought: I don't understand. You are already checking isBoundary(startIdx). If it is, we should just stay there and not call preceding() -- rather than adding 1 and calling preceding() to rediscover this same index. Or am I missing something?
- second thought: I see that BreakIterator has no setIndex(), so this is a workaround; seems dumb but that's the old API; please add this to your comment
- third: is there an internal/hidden BreakIterator or RBBI setIndex()?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,65 @@
...
+      int sourceLength = sourceSequence.length();
+      boolean isOnBoundary =
+          0 <= startIdx
+          && startIdx <= sourceLength
```
markusicu: should the caller (the Segments impl) check for a valid index and otherwise throw an exception?

(you are anyway not validating the index for forwards)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,65 @@
...
+          && breakIter.isBoundary(startIdx);
+      int backFromIdx = isOnBoundary ? startIdx + 1 : startIdx;
```
markusicu: if startIdx == 0 then call breakIter.first(); if startIdx == sourceLength then call breakIter.last()

we only need this dance when we are in the middle

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,194 @@
...
+ * Performs segmentation according to the rules defined for the locale.
+ */
+public class LocalizedSegmenter implements Segmenter {
+
+  private ULocale locale;
+
+  private SegmentationType segmentationType;
+
+  @Override
+  public Segments segment(CharSequence s) {
```
markusicu, line 30: all public methods need api javadoc

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,194 @@
...
+    this.locale = locale;
+    this.segmentationType = segmentationType;
+  }
+
+  private BreakIterator getNewBreakIterator() {
+    BreakIterator breakIter;
+    switch (segmentationType) {
+      case LINE:
+        breakIter = BreakIterator.getLineInstance(locale);
+        break;
```
markusicu: nit/minor: why not just directly return BreakIterator.getXyzInstance(...)?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
```diff
@@ -0,0 +1,194 @@
...
+      return this;
+    }
+
+    /**
+     * Builds the {@code Segmenter}
+     * @return the constructed {@code Segmenter} instance
+     * @draft ICU 78
+     */
+    public Segmenter build() {
+      if (this.segmentationType == null) {
```
markusicu:
```suggestion
      if (segmentationType == null) {
```

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,143 @@
...
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.text.RuleBasedBreakIterator;
+import java.io.InputStream;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+/**
+ * Performs segmentation according to the provided rule string. The rule string must follow the
+ * same guidelines as for {@link RuleBasedBreakIterator#getInstanceFromCompiledRules(InputStream)}.
```
markusicu: I think that function takes a binary of, well, _compiled_ rules.

Please refer to the [RBBI constructor](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/RuleBasedBreakIterator.html#RuleBasedBreakIterator-java.lang.String-).

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+
+class SegmentsImplUtils {
+
+  static boolean isBoundary(BreakIterator breakIter, CharSequence source, int i) {
+    return breakIter.isBoundary(i);
+  }
```
markusicu: Why does this take a source param?
Why not inline this function?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
```diff
@@ -0,0 +1,91 @@
...
+      } else {
+        return null;
+      }
+    } catch (IllegalArgumentException iae) {
+      // Catch the error that is thrown by the implementation helper method inside BreakIterator
+      // (called checkOffset) whenever the index passed to `.isBoundary(int)` is out of bounds
+      // in the original string. Since IndexOutOfBoundsException is more appropriate, throw that
+      // instead.
+      throw new IndexOutOfBoundsException(i);
+    }
```
markusicu: you should be able to remove this since you merged https://github.com/unicode-org/icu/pull/3508

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentSpliterator.java
```diff
@@ -0,0 +1,60 @@
...
+// License & terms of use: https://www.unicode.org/copyright.html
+
+package com.ibm.icu.segmenter;
+
+import com.ibm.icu.text.BreakIterator;
+import com.ibm.icu.segmenter.Segments.IterationDirection;
+import java.util.Spliterator;
+import java.util.function.IntConsumer;
+
+class SegmentSpliterator implements Spliterator.OfInt {
```
markusicu: better BoundarySpliterator

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
```diff
@@ -0,0 +1,74 @@
...
+        // position to increment forwards when the return value is false, we should call
+        // BreakIterator.previous() to update the iterator position while getting the start value
+        // of the segment at startIdx
+        this.start = breakIter.previous();
+      }
+      this.limit = getDirectionBasedNextIdx();
+    }
+  }
+
+  private int getDirectionBasedNextIdx() {
```
markusicu: totally optional: i would call it simply nextIndex(); direction-aware just like next()

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
```diff
@@ -0,0 +1,74 @@
...
+
+  @Override
+  public boolean hasNext() {
+    return this.limit != BreakIterator.DONE;
+  }
+
+  @Override
+  public Segment next() {
+    Segment result;
+    if (this.limit < this.start) {
```
markusicu: I find this very confusing: With fields called start & limit, I would expect that we always have start < limit until one or both are DONE. Which means that for backwards we would advance with (limit = start, start = nextIndex()).

As it is, I got very confused in the constructor about which field gets which value.

I don't know what field names to use if you want to keep the logic as is.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
```diff
@@ -0,0 +1,74 @@
...
+
+  SegmentIterator(BreakIterator breakIter, IterationDirection direction, int startIdx, CharSequence source) {
+    this.breakIter = breakIter;
+    this.direction = direction;
+    this.source = source;
+
+    Segment segmentAtIdx = SegmentsImplUtils.segmentAt(breakIter, source, startIdx);
+
+    if (segmentAtIdx == null) {
+      this.start = BreakIterator.DONE;
```
markusicu: don't need `this.` for start & limit...

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
```diff
@@ -0,0 +1,74 @@
...
+    } else if (direction == IterationDirection.FORWARDS) {
+      this.start = segmentAtIdx.start;
+      this.limit = breakIter.following(this.start);
```
markusicu: I was going to ask “why not limit = segmentAtIdx.limit?” but then figured out that we need not only the indexes, but we also need to set the breakIter to the forward/backward boundary so that next() will do the right thing.

I am bothered by the potential for duplicate work (and complexity and confusion). For example, when we go forward from 0, segmentAt() will have found that 0 is a boundary (which it always is) and called next(). Then we call following(start) here to find the same boundary as the segment limit.

One option: Get more information out of segmentAt().

Instead, I think it would be better to tell segmentAt() which direction we want to take; either as a parameter, but probably cleaner as two separate functions.
- segmentAtForward() would ensure that the breakIter is on the segment limit before returning
- segmentAtBackward() would ensure that the breakIter is on the segment start before returning

Then this code here should be able to pick up the segment boundaries without having to fiddle any further with the breakIter.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentSpliterator.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public int characteristics() {
+    return Spliterator.DISTINCT  // BreakIterator always advances
+        | Spliterator.IMMUTABLE // design of Segmenter API is to provide an immutable view of
+        // segmentation by preventing the input string from mutating
+        // in the underlying BreakIterator
+        | Spliterator.NONNULL   // primtive int is non-null
+        | Spliterator.ORDERED   // BreakIterator always advances, and in a single direction
+        ;
```
markusicu:
```suggestion
    return
        // BreakIterator always advances
        Spliterator.DISTINCT
        // The design of Segmenter the API is to provide an immutable view of
        // segmentation by preventing the input string from mutating
        // in the underlying BreakIterator.
        | Spliterator.IMMUTABLE
        // primitive int is non-null
        | Spliterator.NONNULL
        // BreakIterator always advances, and in a single direction.
        | Spliterator.ORDERED;
```

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class LocalizedSegmenterTest extends CoreTestFmwk {
+
+  @Test
+  public void testLocaleInLocalizedSegmenter() {
+    String source = "Das 21ste Jahrh. ist das beste.";
```
markusicu: should be an entry in the casesData

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
```diff
@@ -0,0 +1,70 @@
...
+      Segments segments = wordSeg.segment(source);
+
+      List<CharSequence> actWords = segments.subSequences().collect(Collectors.toList());
+
+      assertThat(actWords, is(expWords));
+    }
+  }
+
+  @Test
+  public void testJavaLocaleInLocalizedSegmenter() {
```
markusicu: I would not write a whole separate test function for ULocale vs. Locale, complete with duplicate test data etc.

Instead I would duplicate some of the test body, maybe a mini loop over the two variants which build wordSeg with ULocale vs. Locale and then come out to using the same test code with segments/assertThat.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class RuleBasedSegmenterTest extends CoreTestFmwk {
+
+  @Test
+  public void testRules() {
+    String source = "hejsan k:a tack";
```
markusicu: move into casesData

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    Object[][] casesData = {
+        {"default",   ".*;",       Arrays.asList("hejsan k:a tack")},
+        // TODO: add more cases once RBBI rule syntax is understood
```
markusicu: time to add test cases :-)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
```diff
@@ -0,0 +1,52 @@
...
+    };
+
+    for (Object[] caseDatum : casesData) {
+      String desc = (String) caseDatum[0];
+      String subrule = (String) caseDatum[1];
+      List<CharSequence> expWords = (List<CharSequence>) caseDatum[2];
+
+      // the following rule substring was taken as a subset from BreakIteratorRules_en_US_TEST.java:
+      String rules = subrule;
```
markusicu: you are just renaming/aliasing the variable??

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: This should be handled automatically by the source formatter once we are ready to pull the trigger and merge https://github.com/unicode-org/icu/pull/3516. I'd prefer to wait until then. My IDE currently likes to give me a mix of 2 and 4 space indents, apparently. So it makes more sense for one automated tool to override another tool's automatic default behavior than spending human time fighting that.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: done: updated comments about `isBoundary` behavior

discussed offline: `isBoundary(startIdx)` is primarily setting the current index position to a boundary if the index isn't already a boundary, and then it returns whether the input was a boundary.  It's a very misleading name a type signature. `BreakIterator` has no other way to re-/set the current index position held in internal state.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: It's a little more of a functional style that I get from Clojure in which nil punning and/or empty sequences makes more sense than throwing errors. In the case of sequences (Java `Stream`s), returning an empty sequence makes just as much sense. So here, if I want to get all the `Segmenter` sequences back from index -2, then I'll get an empty `Stream`. Which makse sense: "there are no segments back from -2". Any caller that consumes it (ex: a stateful "foreach" `for` loop or the `.forEach` on a Stream) which just do nothing because they traverse an empty Stream. Knowing that your program blow up might be a benefit.

You could return a null, or throw an exception, but then you put extra burden on the caller's code - maybe extra verbosity or complexity. If we return an empty Stream, then they can figure out what to do and/or debug their own situation if it's GIGO.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done. I shortened the open-ended description when copying from the interface to the concrete classes, but kept the text in all of the other fields (`@param`, `@return`) the same. The interface on `Segments` has extra details about invariants because that's where I put the text after refactoring the design doc. I didn't think all of those extra details in the description needed to be copied beyond  the high level description.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: It's more of a functional style. If the switch were an expression (had a return value) rather than just statements, you could get rid of the `break`s and just put `return` in front of `switch`. [Switch expressions](https://docs.oracle.com/en/java/javase/17/language/switch-expressions-and-statements.html) exist in Java 17, fwiw.

If it's more efficient to get rid of the temporary variable, then I'm happy to remove it.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/LocalizedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segment.java
echeran: I left a lot of instances of `this`. in the previous push. I hope I got the rest of them.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
echeran: I addressed this by adjusting both the initialization (constructor) and the iterative step (`hasNext`) in each of the 2x2=4 scenarios: direction={forwards, backwards} x startIdx={boundary, not boundary}.

The tests still pass. In the process, also I removed a helper method.

PTAL to see if this addresses the duplicate work, and if so, whether the code is less confusing.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
echeran: Removed this method while simplifying the initialization code in the constructor.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentIterator.java
echeran: After changing the initialization & iteration code, start < limit is now an upheld invariant throughout.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentSpliterator.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
echeran: Done. Removed. :-|

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundarySpliterator.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public int characteristics() {
+    return
+        // BreakIterator always advances
+        Spliterator.DISTINCT
+        // The design of Segmenter the API is to provide an immutable view of
+        // segmentation by preventing the input string from mutating
+        // in the underlying BreakIterator.
+        | Spliterator.IMMUTABLE
```
echeran, line 45: @markusicu On second thought, I don't think that the Spliterator is immutable, only because we support `isBoundary`, and we delegate that to an impl of that fn that mutates its own state as a side effect.

For the Spliterator API's docs for [IMMUTABLE](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#IMMUTABLE), if you indicate neither IMMUTABLE nor CONCURRENT, then you have to implement something to indicate ConcurrentModificationException.

But an alternative to indicating ConcurrentModificationException is for our `Segments::isBoundary` to first stash the local BreakIterator's current index, then call/delegate to `BreakIterator::isBoundary`, and then reset the local BreakIterator's position. IIRC, any API that takes an index will always call isBoundary internally, so the BreakIterator internal state's position always ends on a boundary by the time we would see it, so I think this alternative should work.

WDYT?

echeran: Discussed offline: one of the biggest goals in this API is to achieve an immutable and thread-safe view over the segmentation results. With all of the APIs in `Segments` that returns `Stream`s, as this PR stands currently, they all keep iteration state by delegating to the one `BreakIterator` that is instantiated in the concrete `Segments` object. They all depend on that `BreakIterator` instance maintain state, but that means that if they're run concurrently, they will mutate each other's state. In this way, `isBoundary` also cannot be run at the same time as these other APIs, with the current PR's code impl design. Instead, we should push down the instantiations (clonings) of `BreakIterator`s into each of the API calls—one instance per call—to ensure true thread-safe immutable semantics for the user. This is desirable, even if it might cost more overhead than optimal. Once we achieve the behavior that we want, we can come back and change & optimize the implementation later on.

echeran: Done after more discussion offline. Offline discussion:

* in the concrete impl class of `Segmenter`, create a prototype `BreakIterator` instance. The data loading part of constructing that object can be expensive, so doing that once and cloning is cheaper than doing it repeatedly
* in the concrete impl class of `Segments`, clone the `BreakIterator` prototype and store that locally. This allows each instance of the concrete `Segments` class to avoid state (the same source string / `CharSeqence`) with each other
* in turn, the local field in `Segments` that holds a `BreakIterator` will be a prototype that gets cloned every time an API on `Segments` gets called. This is because the `BreakIterator` instance is also holding iteration state (the current index/position), so we need to have a separate instance per API so that we can guarantee that concurrent calls to these functions acting on the same `Segments` instance won't cause an inconsistent state.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segmenter.java
```diff
@@ -0,0 +1,43 @@
...
+ * results in a stateless manner.
+ *
+ * @see Segments
+ * @see BreakIterator
+ * @draft ICU 78
+ */
+public interface Segmenter {
+
+  /**
+   * Return a {@link Segments} object that encapsulates the segmentation of the input
```
markusicu: nit/optional: API docs read more nicely in third-person-descriptive rather than imperative: "Returns a Segments object..."

(I think Google style recommends this)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,212 @@
...
+    public Segmenter build() {
+      if (breakIter == null) {
+        throw new IllegalArgumentException("A rule string must be set.");
+      } else {
+        return new RuleBasedSegmenter(breakIter);
+      }
+    }
+  }
+
+  private class RuleBasedSegments implements Segments {
```
markusicu: It looks like this is now almost the exact same code as LocalizedSegments.
If so, please pull them both out into a package-private class, something like SegmentsImpl.

It seems like the only code change would be for the constructor to take a BreakIterator instead of an XyzSegmenter. You could just clone that in the constructor rather than calling segmenter.getNewBreakIterator() which might then no longer be necessary.

Come to think of it, you could then move the SegmentsImplUtils methods into SegmentsImpl, make them non-static, and stop delegating from the XyzSegments to the Utils.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
```diff
@@ -0,0 +1,212 @@
...
+      breakIterPrototype.setText(source);
+    }
+
+    /**
+     * Returns the segment that contains index {@code i}. Containment is inclusive of the start index
+     * and exclusive of the limit index.
+     * @param i index in the input {@code CharSequence} to the {@code Segmenter}
+     * @return A segment that either starts at or contains index {@code i}
+     * @draft ICU 78
+     */
```
markusicu: You don't need elaborate API docs on methods of a private class :-)

At a minimum, please remove the API status tags (`@draft`).

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,76 @@
...
+      // API that includes the input index when it is itself a boundary, unlike the behavior of
+      // BreakIterator.preceding().
+      //
+      // Note: we have to set the initial index indirectly because there is no way to statelessly
+      // query whether an index is on a boundary. Instead, BreakIterator.isBoundary() will mutate
+      // state when the input is not on a boundary, before it returns the value indicating a
+      // boundary.
+      int sourceLength = sourceSequence.length();
+      if (startIdx == 0) {
+        this.currIdx = breakIter.first();
```
markusicu: Please remove `this.` from currIdx :-)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -0,0 +1,76 @@
...
+      } else if (startIdx == sourceLength) {
+        this.currIdx = breakIter.last();
+      } else {
+        boolean isOnBoundary =
+            0 <= startIdx
+                && startIdx <= sourceLength
+                && breakIter.isBoundary(startIdx);
+        int backFromIdx = isOnBoundary ? startIdx + 1 : startIdx;
+
+        this.currIdx = breakIter.preceding(backFromIdx);
```
markusicu: Can we not simplify?

```suggestion
        currIdx = isOnBoundary ? startIdx : breakIter.previous();
```

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
```diff
@@ -0,0 +1,79 @@
...
+        // BreakIterator::isBoundary(i) will advance forwards to the next boundary if the argument
+        // is not a boundary.
+        limit = breakIter.current();
+        start = breakIter.previous();
+      }
+
+      if (start != BreakIterator.DONE && limit != BreakIterator.DONE) {
+        return new Segment(start, limit, sourceSequence);
+      } else {
+        return null;
```
markusicu: This is not documented. Assuming that BI.isBoundary(i) throws an exception for i not in 0..sourceLength, it seems like this can only happen for i==sourceLength, where limit == DONE. Should we return an empty Segment in that case? Or throw an exception? -- I am thinking return empty. WDYT?

I think we can and should assert start != DONE.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
```diff
@@ -0,0 +1,79 @@
...
+
+  static Stream<Segment> segmentsBefore(BreakIterator breakIter, CharSequence sourceSequence, int i) {
+     breakIter.setText(sourceSequence);
+
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    SegmentIterable iterable = new SegmentIterable(breakIter, IterationDirection.BACKWARDS, i, sourceSequence);
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+
+  static Function<Segment, CharSequence> segmentToSequenceFn(CharSequence sourceSequence) {
```
markusicu: is this called from somewhere? Segment now has a source.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    Object[][] casesData = {
+        {"de",
+            "Das 21ste Jahrh. ist das beste.",
+            Arrays.asList("Das", " ", "21ste", " ", "Jahrh", ".", " ", "ist", " ", "das", " ", "beste", ".")},
+    };
+
+    for (Object[] caseDatum : casesData) {
+      String localeTag = (String) caseDatum[0];
+      // ULocale locale = ULocale.forLanguageTag(localeTag);
```
markusicu:
```suggestion
```

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
```diff
@@ -0,0 +1,358 @@
...
+    // Create new Segments for source1
+    Segments segments1 = enWordSegmenter.segment(source1);
+
+    List<Segment> segments = segments1.segments().collect(Collectors.toList());
+
+    assertEquals("first range start", 0, segments.get(0).start);
+    assertEquals("first range limit", 3, segments.get(0).limit);
+
+    assertEquals("second range start", 3, segments.get(1).start);
+    assertEquals("second range limit", 4, segments.get(1).limit);
```
markusicu: s/range/segment/

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
```diff
@@ -0,0 +1,358 @@
...
+            .setLocale(ULocale.ENGLISH)
+            .setSegmentationType(LocalizedSegmenter.SegmentationType.WORD)
+            .build();
+
+    String source1 = "The quick brown fox jumped over the lazy dog.";
+
+    // Create new Segments for source1
+    Segments segments1 = enWordSegmenter.segment(source1);
+
+    Object[][] casesData = {
```
markusicu: optional: You could test this by calling segments(), collect all of the limits (and 0), and looping over the whole string and checking that isBoundary() is true exactly for the collected segment boundaries. Less fragile/onerous than explicit boundary data, and you test segments() elsewhere.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
```diff
@@ -0,0 +1,358 @@
...
+    // Create new Segments for source
+    Segments segments1 = enWordSegmenter.segment(source);
+
+    Object[][] casesData = {
+        {"index before beginning",                       -2,                 null,              null},
+        {"index at beginning",                           0,                  0,                 3},
+        {"index in the middle of the first segment",     2,                  0,                 3},
+        {"index in the middle of the third segment",     5,                  4,                 9},
+        {"index at the end",                             source.length()-1,  44,                45},
+        {"index after the end",                          source.length()+1,  null,              null},
```
markusicu: Please also test with source.length() itself.

markusicu: should this one not throw an exception?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/Segmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/RuleBasedSegmenter.java
echeran: Done. Problem removed when the inner classes implementing `Segments` were deduplicated into the helper `SegmentsImpl` class.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: Done.

echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
echeran: Done. Offline discussion: throwing an exception is better than returning a `Segment` with empty length because the existence of a `Segment` object (as opposed to null) implies that it is a real segment resulting from segmentation. A `Segment` with empty length means that its fields `start` and `limit` would be artificial yet specific, and set to the same value. It would be confusing.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImplUtils.java
echeran: Done. Removed. This is a vestige of old code that got obviated, as you said, after the `Segment` class gained a field to point to the source `CharSequence`.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/LocalizedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
echeran: Done.

echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
```diff
@@ -0,0 +1,128 @@
...
+      return new Segment(start, limit, source);
+    } else {
+      return null;
+    }
+  }
+
+  @Override
+  public Stream<Segment> segments() {
+    BreakIterator breakIter = (BreakIterator) breakIterPrototype.clone();
+    breakIter.setText(source);
```
markusicu: needed? the constructor did this on the prototype

several more in this file

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+    if (start != BreakIterator.DONE && limit != BreakIterator.DONE) {
+      return new Segment(start, limit, source);
+    } else {
+      return null;
+    }
+  }
+
+  @Override
+  public Stream<Segment> segments() {
```
markusicu: call segmentsFrom(0) to reduce code?

maybe even do so in the interface definition?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    breakIter.setText(source);
+
+    // create a Stream from a Spliterator of an Iterable so that the Stream can be lazy, not eager
+    SegmentIterable iterable = new SegmentIterable(breakIter, IterationDirection.BACKWARDS, i,
+        source);
+    return StreamSupport.stream(iterable.spliterator(), false);
+  }
+
+  @Override
+  public IntStream boundaries() {
```
markusicu: call boundariesAfter(-1) to reduce code?

maybe even do so in the interface definition?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
```diff
@@ -50,9 +50,8 @@ class BoundaryIteratorOfInts {
             0 <= startIdx
                 && startIdx <= sourceLength
                 && breakIter.isBoundary(startIdx);
-        int backFromIdx = isOnBoundary ? startIdx + 1 : startIdx;
 
-        currIdx = breakIter.preceding(backFromIdx);
+        currIdx = isOnBoundary ? startIdx : breakIter.preceding(startIdx);
```
markusicu, line 64: why not
```
currIdx = isOnBoundary ? startIdx : breakIter.previous();
```

?

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
```diff
@@ -40,6 +40,10 @@ public Segment segmentAt(int i) {
     int start;
     int limit;
 
+    if (i < 0 || i >= breakIterPrototype.last()) {
```
markusicu: Calling breakIterPrototype.last() is not thread-safe.
Please just use source.length() here.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
```diff
@@ -104,22 +106,23 @@ public void testIsBoundary() {
...
-        {"end of string",                        source1.length(),      true},
-    };
-
-    for (Object[] caseDatum : casesData) {
-      String desc = (String) caseDatum[0];
-      int idx = (int) caseDatum[1];
-      boolean exp = (boolean) caseDatum[2];
-
-      assertThat(desc, segments1.isBoundary(idx) == exp);
+    BitSet actBoundaries = new BitSet(source1.length());
```
markusicu: `new BitSet(source1.length() + 1);`

+1 so that the length itself fits

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/BitSet.html#%3Cinit%3E(int)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
```diff
@@ -104,22 +106,23 @@ public void testIsBoundary() {
...
+    for (Segment seg : segments) {
+      actBoundaries.set(seg.start);
+      actBoundaries.set(seg.limit);
     }
+
+    IntStream.rangeClosed(0, source1.length() - 1).forEach(
```
markusicu: Please remove the `- 1` so that we test the boundary at length as well.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
```diff
@@ -27,6 +27,14 @@ public void testRules() {
             "Kühlschränke kühlen Getränke",
             "[a-z]+;",
             Arrays.asList("K", "ü", "hlschr", "ä", "nke", " ", "k", "ü", "hlen", " ", "G", "etr", "ä", "nke")},
+        {"ASCII upper- and lowercase a-z",
+            "Kühlschränke kühlen Getränke",
+            "[A-Za-z]+;",
+            Arrays.asList("K", "ü", "hlschr", "ä", "nke", " ", "k", "ü", "hlen", " ", "Getr", "ä", "nke")},
+        {"ASCII upper- and lowercase a-z and some letters with diaeresis",
+            "Kühlschränke kühlen Getränke",
+            "[A-Za-z{ü}{ä}]+;",
```
markusicu: nit: unless the accented characters are encoded in NFD, they don't need the curly braces

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
echeran: Done. Removed all of the `.setText()` calls on the cloned BreakIterator

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: I probably did that just to make it more readable. Your formulation is correct, and saves a few instructions and a few characters, but it requires knowledge of the statefulness of BreakIterator (which is unwieldy and sometimes surprising-in-a-bad-way). The cost of the current formulation is minimal since it's in a constructor, so not in a loop, and BreakIterator avois getting re-constructed often because we clone it instead.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/SegmentsTest.java
echeran: Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/segmenter/RuleBasedSegmenterTest.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/SegmentsImpl.java
echeran: Done.

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
markusicu: I don't think it's more readable, and everywhere in this implementation we need to know intimately how the BreakIterator works. I prefer making this as succinct as in segmentAt(i) -- we know that isBoundary() returned false and moved to the following boundary, and previous() will get us back from there to where we want to be. But I don't know if it actually saves anything. Either way we hopefully have at least the two surrounding boundaries cached. Ok either way; maybe just a comment that we *could* call previous().

markusicu (approved):
lean and mean :-)

icu4j/main/core/src/main/java/com/ibm/icu/segmenter/BoundaryIteratorOfInts.java
echeran: Fair points. I went ahead and tried to implement your suggestion, but I couldn't because a test broke. The breaking test is `SegmentsTest.testBoundariesBackFrom` with the  test case of the starting index being after the last boundary. `BreakIterator.previous()` returned `DONE` in that test case (leading ultimately to an empty Stream), whereas `BreakIterator.preceding(i)` returns `BreakIterator.last()` when `i` is after the last boundary. So I only updated comments on the existing code, which includes these details.

markusicu: Thanks for giving it a good try!

PR #3242, opened 2024-10-16
ICU-22946 Fix github breakage on pip3 install hjson

FrankYFTang:
Fixing icu4c-without-collation-rule-strings
Ref: https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3

#### Checklist
- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22946

### DIFF
--- a/.github/workflows/icu4c.yml
+++ b/.github/workflows/icu4c.yml
@@ -807,7 +807,8 @@ jobs:
       - name: Install hjson dependency
         run: |
           sudo apt-get install python3-pip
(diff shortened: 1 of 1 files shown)


FrankYFTang:
The fix is wrong and the problem go away. Probably because someone in github fix it.

FrankYFTang:
the problem is fixed. this is the wrong approach

PR #3247, opened 2024-10-18
ICU-22946 Pin all -latest runners per ICU-TC 2024-10-17 🗡️

eggrobin:
Pin ubuntu-latest, windows-latest, and macos-latest to their current values, with a note that those are updated in the BRS (I have added a post-release task to the BRS for that).

It does not seem trivial to have that configuration in a single place ([configuration variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-configuration-variables-for-multiple-workflows) would work, but those are not part of the repository itself, so they are not source-controlled and they are only accessible to repository owners; I am also not sure how they would behave with PRs from forks). It might be possible to reduce this to one occurrence per file, but 22 files is enough that I would want to three global substitutes anyway, at which point it does not make a difference that there are 83 of them.

#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22946

### DIFF
--- a/.github/workflows/cache_retain.yml
+++ b/.github/workflows/cache_retain.yml
@@ -35,7 +35,7 @@ permissions:
 jobs:
   retain-maven-cache:
     name: Run all tests with Maven
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04  # Updated in BRS
     steps:
       - name: Checkout and setup
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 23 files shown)


markusicu:
I reopened @FrankYFTang 's ticket in order to unblock this PR.

markusicu:
I am happy to rubber-stamp these changes, but we should also have documentation for the new BRS task.

I suggest adding a section here:
- https://unicode-org.github.io/icu/processes/release/tasks/build.html
- https://github.com/unicode-org/icu/blob/main/docs/processes/release/tasks/build.md

Maybe just before the obsolete "Update the pool bundles".

markusicu (approved):
tnx!

mihnita (approved):
Wow, that's a lot of files!

We should probably look into some cleanup...
(not now, of course)

Thank you for this one,
Mihai

PR #3250, opened 2024-10-21
ICU-22950 Do not initialize currency for NumberPropertyMapper when useCurrency is false

nightelf3:
resolveCurrency() slows down simple number formatting requests like:
```
for (int i = 0; i < 9999; i++)
  u_snprintf_u(buff, 100, u"%d", i);
```
Since icu4c doesn't use currency there, we can skip its initialization.

#### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22950

### DIFF
--- a/icu4c/source/i18n/number_mapper.cpp
+++ b/icu4c/source/i18n/number_mapper.cpp
@@ -74,9 +74,11 @@ MacroProps NumberPropertyMapper::oldToNew(const DecimalFormatProperties& propert
             !properties.currencyPluralInfo.fPtr.isNull() ||
             !properties.currencyUsage.isNull() ||
             warehouse.affixProvider.get().hasCurrencySign());
-    CurrencyUnit currency = resolveCurrency(properties, locale, status);
-    UCurrencyUsage currencyUsage = properties.currencyUsage.getOrDefault(UCURR_USAGE_STANDARD);
+    CurrencyUnit currency;
+    UCurrencyUsage currencyUsage;
     if (useCurrency) {
+        currency = resolveCurrency(properties, locale, status);
+        currencyUsage = properties.currencyUsage.getOrDefault(UCURR_USAGE_STANDARD);
         // NOTE: Slicing is OK.
         macros.unit = currency; // NOLINT
     }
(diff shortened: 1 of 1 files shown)


eggrobin (dismissed):

icu4c/source/i18n/number_mapper.cpp
```diff
@@ -74,9 +74,11 @@ MacroProps NumberPropertyMapper::oldToNew(const DecimalFormatProperties& propert
             !properties.currencyPluralInfo.fPtr.isNull() ||
             !properties.currencyUsage.isNull() ||
             warehouse.affixProvider.get().hasCurrencySign());
-    CurrencyUnit currency = resolveCurrency(properties, locale, status);
-    UCurrencyUsage currencyUsage = properties.currencyUsage.getOrDefault(UCURR_USAGE_STANDARD);
+    CurrencyUnit currency;
+    UCurrencyUsage currencyUsage;
```
markusicu, line 78: I expanded the rest of this (long) function, searched for "currency", and found some code that uses these variables but is gated by conditions other than `useCurrency`.

- `if (!properties.currencyUsage.isNull())`
- `if (exportedProperties != nullptr)`

If we think we can guarantee that useCurrency is true there, then we should add `U_ASSERT(useCurrency)` there.
For the second one, maybe a default-constructed `currency` is fine if not `useCurrency`?
Otherwise, we may need to add a runtime check for useCurrency.

@sffc @mihnita @richgillam @eggrobin

richgillam: Markus is right; this is a concern.

For `if (!properties.currencyUsage.isNull())`, I wonder if this could just be changed to `useCurrency`?

For `exportedProperties->currency`, I agree that a default-constructed `currency` ought to be fine.

For `if (precision.fType == Precision::PrecisionType::RND_CURRENCY)`, I wonder if we could change this to `if (useCurrency && precision.fType == Precision::PrecisionType::RND_CURRENCY)`?

nightelf3: Added an assertion to `if (!properties.currencyUsage.isNull())`, since useCurrency is true when `!properties.currencyUsage.isNull()`
Used `if (useCurrency && precision.fType == Precision::PrecisionType::RND_CURRENCY)`, thank you!

richgillam (approved):
Looks good.  Thanks for the changes!

markusicu:
The tests hadn't run again. I rebased, squashed, fixed a couple of typos in the commit message, force-pushed, and approved test runs.

PR #3268, opened 2024-11-13
ICU-22927 Duplicate (back) the MF2 test data between icu4c and icu4j

mihnita:
The C++ and Java implementations are done by two different people different companies. With different time constraints and availability.
The spec is still not final (although it is close), and the implementations are still quite a bit behind.
Sharing these test files slows down development, by forcing any C++ and Java changes to happen in the same time.

There are other components that share test files that are not shared yet, even if they are more stable. So I don't know why we would force this on MF2 only.

This is temporary, and the data files will be de-duplicated again at a later time, when the two implementations are more stable.
That de-duplication will include other shared files, and in a slightly different structure (we have a doc and a discution on this topic).

#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22927
  - For minor changes you can use one of the following per-release tickets:
  - ICU 77 code warnings/version updates: ICU-22920 — Fix compiler warnings. Update versions of code-related dependencies (e.g., dependabot).
  - ICU 77 docs minor fixes: ICU-22921 — User Guide & API docs typos etc., and version updates (e.g., dependabot for User Guide)


mihnita:
> Is there a JIRA issue to deduplicate the tests again in the future after the spec settles down?

The same as before: https://unicode-org.atlassian.net/browse/ICU-22927

Note that the issue is not about MF2. And only explicitly lists
`localeMatcherTest.txt`, `break_rules/`, `BidiCharacterTest.txt`, `BidiTest.txt`, `IdnaTestV2.txt`, `numberformattestspecification.txt`, `numberpermutationtest.txt` and `collate/`

It is a clone of an older issue, opened in July 31, 2020
https://unicode-org.atlassian.net/browse/ICU-21205?focusedCommentId=157500

So I don't know why it all became about MF2.

---

And not only that it slows MF2 dev down, it is not in "the maven spirit", (even if the problems with Eclipse are solved, it was done in a hacky way).

Finally, there is a automated step that takes CLDR test files and copies them in the icu structure.
That was not updated.

TLDR: this is not nice and clean anyway.

markusicu (approved):
rslgtm

PR #3270, opened 2024-11-16
ICU-22922 Merge maint/maint-76 to main

sffc:
https://unicode-org.atlassian.net/browse/ICU-22922

Follow-up to #3258 and #3260


markusicu (approved):
thank you!

markusicu:
So... this was supposed to be a routine merge of the straggler maint/maint-76 CLDR-to-ICU integration commit from @DraganBesevic 's https://github.com/unicode-org/icu/pull/3260 to main.

However, we have two types of check failures here:

1. There is a [C++ exhaustive intltest failure from utility/LocaleTest/TestDataDrivenLikelySubtags](https://github.com/unicode-org/icu/actions/runs/11826108766/job/32951358852?pr=3270). This is the [same test failure that we have been seeing on main already](https://github.com/unicode-org/icu/actions/runs/11867025958), so it's not a regression from this commit. I was hoping that this failure would get fixed with integrating the late CLDR 46 likely subtags change, but no :-/

*Since this is not a regression from this PR, I propose that we ignore it for this PR.*
We do need to debug & fix it on main, but not in a merge-from-maint PR.

2. Three of the nine MSVC checks fail with this:
```
"D:\a\icu\icu\icu4c\source\allinone\allinone.sln" (default target) (1) ->
"D:\a\icu\icu\icu4c\source\data\makedata_uwp.vcxproj" (default target) (56) ->
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VisualStudio\v17.0\AppxPackage\Microsoft.AppXPackage.Targets(225,25): error MSB4184: The expression "[MSBuild]::VersionGreaterThanOrEquals('', 9.0)" cannot be evaluated. Version string was not in a correct format. [D:\a\icu\icu\icu4c\source\data\makedata_uwp.vcxproj]
```

This looks like an MSVC config bug, unrelated to this PR. *I propose that we ignore this, too, for now.*
I will restart these checks to see if they were flakes.

@rp9-next @richgillam ok to ignore these and let Shane merge?

richgillam:
Markus—

> @rp9-next <https://github.com/rp9-next> @richgillam <https://github.com/richgillam> ok to ignore these and let Shane merge?
> 

Fine with me.

—Rich

markusicu:
Thanks, Rick.
I reran some of the MSVC checks once or twice, and they keep failing. But I don't see how these failures could be related to these changes.
We need to put a lid on ICU 76 and focus on fixing problems on main.
@sffc please merge.

PR #3278, opened 2024-11-21
ICU-22977 Fix MSVC builds - Forcing target version

mihnita:
Defined `TargetFrameworkVersion`, following this doc:
https://learn.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=msvc-170#target-framework-ccli-project-only

I am not sure if it is the right fix. It worked just fine for so many years.

Will see what the GitHub team says.

#### Checklist
- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22977

### DIFF
--- a/icu4c/source/common/common_uwp.vcxproj
+++ b/icu4c/source/common/common_uwp.vcxproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <ItemGroup Label="ProjectConfigurations">
     <ProjectConfiguration Include="Debug|Win32">
@@ -38,6 +38,7 @@
     <ProjectGuid>{C10CF34B-3F79-430E-AD38-5A32DC0589C2}</ProjectGuid>
     <Keyword>DynamicLibrary</Keyword>
     <DefaultLanguage>en-US</DefaultLanguage>
+    <TargetFrameworkVersion>10.0</TargetFrameworkVersion>
   </PropertyGroup>
   <PropertyGroup Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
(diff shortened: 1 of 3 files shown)


rp9-next:
Thanks for working on this! I too was looking into some solutions such as retargeting platform version (#3277), but couldn't get it to work without reverting to windows-2019 which made me believe this might be another vm runner update of MSVC issue. 

May I ask, how did you get to `<TargetFrameworkVersion>10.0</TargetFrameworkVersion>` ?

Looking at [docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-target-framework-and-target-platform?view=vs-2022), it seems to be the .NET Framework version the latest of which is .NET 4.8 or (.NET 8 for .NET core). Was this set through visual studio IDE ?

icu4c/source/common/common_uwp.vcxproj
```diff
@@ -106,7 +107,7 @@
       <TurnOffAssemblyGeneration>true</TurnOffAssemblyGeneration>
       <IgnoreSpecificDefaultLibraries>vccorlib.lib;msvcrt.lib</IgnoreSpecificDefaultLibraries>
       <!-- The icudt.lib is for U_ICUDATA_ENTRY_POINT -->
-      <AdditionalDependencies>icudt.lib;onecore.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
rp9-next, line 109: Perhaps the [arm build](https://github.com/unicode-org/icu/actions/runs/11951328578/job/33314736962#step:5:4548) is failing due to this change ?

rp9-next:
> .NET 8 for .NET core

My bad, .NET 9 was released a few days ago that gets included in Visual Studio 17.12. Perhaps that's the source of the build failures.
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

mihnita:
@rp9-next
Do you think I should make it 9.0 instead of 10.0?

mihnita:
> Perhaps the [arm build](https://github.com/unicode-org/icu/actions/runs/11951328578/job/33314736962#step:5:4548) is failing due to this change ?

Yes, already fixed (6 hours ago)

mihnita:
> May I ask, how did you get to <TargetFrameworkVersion>10.0</TargetFrameworkVersion> ?

The error pointed me to the right file and line.
I diff-ed the file pre / post update (from a Windows snapshot).

I removed the condition on version and the failure was fixed.
So then I tried to make it pass the test.

I have doubts about 10.0
Maybe should be 9.0. Or maybe v4.8

>  Was this set through visual studio IDE ?

No, it was "by hand" following this doc:
https://learn.microsoft.com/en-us/cpp/build/how-to-modify-the-target-framework-and-platform-toolset?view=msvc-170#target-framework-ccli-project-only

rp9-next:
> I have doubts about 10.0
> Maybe should be 9.0. Or maybe v4.8

According to https://learn.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions, 9.0 seems like a valid target version. Perhaps 9.0 (or 8.0 if for some reason it doesn't work).
I think this is better fix (vs #3277) if CI gets green.

PR #3287, opened 2024-11-29
ICU-22984 Generate old monkeys

eggrobin:
C++ only for now, I will do the [old monkeys from Java](https://en.wikipedia.org/wiki/Javan_surili) separately (that will be more of the same but with subtle differences, as we will need to expand the UnicodeSets ourselves before feeding them to Pattern). This uses ICU regexes to match the context before and after, or to match the left-hand side of a remap rule, as in UAX14 and UAX29. The regexes are the ones used to generate the conformance tests in the Unicode tools; they now closely match the ones in the UAX.
Also line breaking only; the other ones will be very similar (and simpler).

The generated partition and rules are from https://github.com/unicode-org/unicodetools/pull/979. Note that the generation involves transforming `&` and `-` (UnicodeSet syntax) to `&&` and `--` (ICU regex character class syntax).

#### Checklist
- [x] Required: Issue filed: ICU-22984

ALLOW_MANY_COMMITS=true


macchiati (approved):
Looks like a good optimization (and cleanup)

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -2789,11 +2797,16 @@ RBBILineMonkey::RBBILineMonkey() :
 
     std::list<std::pair<std::string, UnicodeSet>> partition;
 
-    // TODO(egg): The following two are a workaround for what seems to be an ICU bug.
+    // TODO(egg): The following two workarounds for what seems to be ICU bugs;
+    // with UREGEX_DOTALL (but not UREGEX_MULTILINE):
+    // 1. /.*\u000A/ does not match CR LF;
```
markusicu, line 2858: I think that's because . matches CR LF, so it won't match just half of that.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -2963,14 +2976,14 @@ RBBILineMonkey::RBBILineMonkey() :
...
 
-    /* TODO Something about SA, but not this.
-    for (int lb = 0; lb < U_LB_COUNT; ++lb) {
-        const std::string lbValueShortName = u_getPropertyValueName(UCHAR_LINE_BREAK, lb, U_SHORT_PROPERTY_NAME);
-        if (lbValueShortName == "SA") {
-          continue;
+    const UnicodeSet lbSA(uR"(\p{lb=SA})", status);
+    for (auto it = partition.begin(); it != partition.end();) {
+        if (lbSA.containsAll(it->second)) {
+          it = partition.erase(it);
```
markusicu, line 3038: pse check & fix indentations...

markusicu (commented):
very nice replacement of hardcoded rules with generated ones!

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -2606,6 +2607,143 @@ RBBISentMonkey::~RBBISentMonkey() {
...
 
 
+class SegmentationRule {
+  public:
+    enum Resolution {
+        BREAK = u'÷',
+        NO_BREAK = u'×',
+    };
+    struct BreakContext {
+        BreakContext(std::size_t index) : indexInRemapped(index) {}
```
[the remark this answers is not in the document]
eggrobin: I will do another pass to try to make more things signed. As you can see from the commit history, I made some of them unsigned to deal with (perhaps overzealous) signed-unsigned comparison warnings, and then realized that I had written code that was warning-free but also UB, so went back to int64_t…

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -2606,6 +2607,143 @@ RBBISentMonkey::~RBBISentMonkey() {
...
+
+class RemapRule : public SegmentationRule {
+  public:
+    RemapRule(const std::u16string_view name, const std::u16string_view pattern,
+              const std::u16string_view replacement)
+        : SegmentationRule(name), replacement_(replacement) {
+        UParseError parseError;
+        UErrorCode status;
+        pattern_.reset(RegexPattern::compile(pattern, parseError, status));
+        U_ASSERT(U_SUCCESS(status));
```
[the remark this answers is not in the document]
eggrobin: Ah, I did not know about IcuTestErrorCode. I will look into that.

icu4c/source/test/intltest/rbbitst.cpp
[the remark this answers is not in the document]
markusicu: Why not int32_t? None of the test strings will be >1B characters.

icu4c/source/test/intltest/rbbitst.cpp
eggrobin: That may well be the explanation, but even with regex greed, that still feels like a bug: `/(rn|[a-z])*n/` still matches `rn`.

icu4c/source/test/intltest/rbbitst.cpp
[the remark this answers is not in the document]
eggrobin: I ended up just `std::terminate`ing with a message as I have done elsewhere, since bubbling up the error all the way to the test would make debugging tricky.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -2606,6 +2607,143 @@ RBBISentMonkey::~RBBISentMonkey() {
...
+                if (*resolved[i].indexInRemapped > matcher->start(status)) {
+                    break;
+                }
+                *resolved[i].indexInRemapped += offset;
+            }
+            for (;; ++i) {
+                if (!resolved[i].indexInRemapped.has_value()) {
+                    continue;
+                }
+                if (*resolved[i].indexInRemapped == matcher->end(status)) {
```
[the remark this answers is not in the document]
eggrobin: Done. `>` should really never happen unless you throw some extremely evil rules at this though. Added a long comment to that effect.

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -1641,6 +1641,25 @@ class RemapRule : public SegmentationRule {
...
+        // units, and apply the following two rules:
+        //   1. (?<X>\P{lb=SP}) \p{lb=CM}* → ${X}
+        //   2. \p{lb=CM} → A
+        // The string remapped and the indexInRemapped values change as follows:
+        // indexInRemapped                   remapped string  rule        final
+        // (aligned on the initial string)                    applied     offset
+        //    𒀀     ◌́     ␠     ◌𝅲
+        // 0  1  2     3     4  5  6         ⟨ 𒀀, ◌́, ␠, ◌𝅲 ⟩   (none)
+        // 0  ␀  ␀     2     3  4  5         ⟨ 𒀀,    ␠, ◌𝅲 ⟩   1                -1
+        // 0  ␀  ␀     2     3  ␀  4         ⟨ 𒀀,    ␠, A ⟩   2                -1
```
markusicu: should the final offset be -2?

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -1691,7 +1710,10 @@ class RemapRule : public SegmentationRule {
             std::terminate();
         }
         remapped = result;
-        U_ASSERT(U_SUCCESS(status));
+        if (!U_SUCCESS(status)) {
```
markusicu, line 1731: Please change  if (!U_SUCCESS(status))  to if (U_FAILURE(status))  -- at least two of these

icu4c/source/test/intltest/rbbitst.cpp
```diff
@@ -1602,8 +1602,8 @@ class SegmentationRule {
         NO_BREAK = u'×',
     };
     struct BreakContext {
-        BreakContext(std::int64_t index) : indexInRemapped(index) {}
-        std::optional<std::int64_t> indexInRemapped;
+        BreakContext(int index) : indexInRemapped(index) {}
```
markusicu, line 1605: In ICU, we normally use int32_t not int. In the old days, an int could be just 16 bits wide. I suppose it could become 64 bits on some machine. Anyway, the functions involved explicitly return int32_t, so that would guarantee a type match.

markusicu (commented):
great comments!

icu4c/source/test/intltest/rbbitst.cpp
eggrobin: No, the `offset` variable is with respect to the previous value of indexInRemapped, so 6 + -1 + -1 = 4.

icu4c/source/test/intltest/rbbitst.cpp
eggrobin: https://www.youtube.com/watch?v=vC7C4BEvhRs

Done.

icu4c/source/test/intltest/rbbitst.cpp
eggrobin: This file is full of `int`s, but done.

(But then if we are not going with the concision of `int`, why not `int64_t`? It’s not like there is a performance benefit these days, and we did [get bitten something being 32-bit here](https://github.com/unicode-org/icu/commit/cc64ec7c94a822ae97dea5cd5a4fa972b2e465ef)…)

markusicu: Where 32 bits are plenty, I still prefer to use them. It's also consistent with long-standing ICU practice, and consistent with Java whose "int" is precisely int32_t, and is used for string indexes etc., so can help a little with porting between the languages.

markusicu: ... and we universally use `int32_t` not `std::int32_t` but I won't block the PR on that now...

markusicu (approved):

icu4c/source/test/intltest/rbbitst.cpp
eggrobin: > ... and we universally use int32_t not std::int32_t but I won't block the PR on that now...

But I did it anyway, so now you get to restamp the rubber :-p

markusicu (approved):

markusicu:
Remember to squash the commits. It's ok if you want to merge multiple commits if they make sense as separate ones.
Each one's message needs to start with `ICU-22984 `
Squashing without contents change will not require re-approval.

eggrobin:
> It's ok if you want to merge multiple commits if they make sense as separate ones.

I tried with four steps (unicodetools-like implementation, optimization, code motion which wants to be on its own because it is undiffable, post-motion cleanup). Let me know if you like that, otherwise I can squish it all into one commit.

markusicu:
> > It's ok if you want to merge multiple commits if they make sense as separate ones.
> 
> I tried with four steps (unicodetools-like implementation, optimization, code motion which wants to be on its own because it is undiffable, post-motion cleanup). Let me know if you like that, otherwise I can squish it all into one commit.

wfm tnx
I added the allow-many-commits magic into the description.

FrankYFTang:
The Exhaustive Tests for ICU #22  is broken between e025466 and  2e57f07 in TestMonkey

https://github.com/unicode-org/icu/actions/runs/12209763924 

Likely caused by this PR

eggrobin:
Yes, fixing in #3296.

PR #3290, opened 2024-12-05
ICU-22773 Migrate the CLDR conversion tool to Maven

mihnita:
This also touches / fixes (at least partially?) other issues:
* [ICU-20998](https://unicode-org.atlassian.net/browse/ICU-20998) Fix conflicting uses of CLDR_DIR in cldr-to-icu process/tools
* [ICU-21426](https://unicode-org.atlassian.net/browse/ICU-21426) Updates/changes to CLDR-to-ICU build tools to handle CLDR Maven migration
* [ICU-21748](https://unicode-org.atlassian.net/browse/ICU-21748) Change ICU CLDR tools to use CLDR artifact directly
* Fixed the unit tests. They didn't work before.
* more

I will search and close / comment on these bugs and others once this lands.

**Summary**

* Java code changes (the tool proper)
  * moved the XML configuration part in a separate file, out of the ant build xml
  * converted the 3 Ant `Task` classes to regular classes, and invoked them from a Main class.
  * the 3 files load the config from the separate xml file
  * created a "Main class", with a command line parser that recognizes all the options passed before from the ant build xml file
* The CLDR Library (`cldr-code`) is now built and installed in the local Maven repo, not in a folder, and the extra `lib` artifact is gone (ICU-21748)
* Updated all the documentation I was able to find

**ToDo / wish list**
* renames / reorg? (see below)
* I have a change that makes the generation 4-5 times faster
* put together a small script (Python?) that invokes the 3-4 commands that now are in the `README.md` and need to be executed by hand, with copy/paste
* There is still one ant build file one level up that imports test files from CLDR to ICU
But it not `tools/cldr/cldr-to-icu`. Maybe move to Python?
* But the last 2 points should maybe be part of a bigger task (automating things with python and moving away from copy-paste from .md files)

---

I tried to keep the changes at minimum.

So I preserved the old files with a "fake" `Task` class, with `init()` and `execute()` methods, as if it is an Ant Task.
Also, kept the files in the `.../ant` package.

This should (hopefully) help with the review.

I can do a cleaning after this with cleanup, refactoring (renaming and so on).

I did quite a bit of testing, but mostly manual, running the conversion tool on Linux / MacOS / Windows (which didn't work before, but works now), with all kind of options.

#### Checklist
- [x] Required: Issue filed: ICU-22773

[ICU-20998]: https://unicode-org.atlassian.net/browse/ICU-20998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-21426]: https://unicode-org.atlassian.net/browse/ICU-21426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-21748]: https://unicode-org.atlassian.net/browse/ICU-21748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/cldr/cldr-to-icu/README.md
```diff
@@ -40,65 +64,132 @@ relies on a pre-processing step, and the CLDR data must come from the separate
 "staging" repository (i.e. https://github.com/unicode-org/cldr-staging) or be
 pre-processed locally into a different directory.
 
+:point_right: **Note**: the 3 folders can also be overridden:
```
markusicu, line 67: optional: do we need three ways to do this?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -86,7 +90,6 @@ public static final class Retain extends Task {
 
         // Don't use "Path" for the argument type because that always makes an absolute path (e.g.
         // relative to the working directory for the Ant task). We want relative paths.
```
markusicu, line 92: optional: remove references to and/Ant in comments?

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -340,4 +338,77 @@ private static ImmutableList<String> readLinesFromResource(String name) {
...
+            Document doc = builder.parse(new File(fileName));
+            Element root = doc.getDocumentElement();
+            if (!"config".equals(root.getTagName())) {
+                System.err.println("The root of the config file should be <config>");
+                return null;
+            }
+
+            NodeList outputDirectories = root.getElementsByTagName("outputDirectories");
+            if (outputDirectories.getLength() != 1) {
+                System.err.println("Only one <outputDirectories> element allowed and required");
```
markusicu, line 388:
```suggestion
                System.err.println("Exactly one <outputDirectories> element allowed and required");
```

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/ConvertIcuDataTask.java
```diff
@@ -408,4 +387,128 @@ private static <T extends Enum<T>> T resolve(Class<T> enumClass, String name) {
...
+            Document doc = builder.parse(new File(fileName));
+            Element root = doc.getDocumentElement();
+            if (!"config".equals(root.getTagName())) {
+                System.err.println("The root of the config file should be <config>");
+                return null;
+            }
+
+            NodeList convertNodes = root.getElementsByTagName("convert");
+            if (convertNodes.getLength() != 1) {
+                System.err.println("Only one <convert> element allowed and required");
```
markusicu, line 476:
```suggestion
                System.err.println("Exactly one <convert> element allowed and required");
```

tools/cldr/cldr-to-icu/README.md
mihnita: Unfortunately some is legacy.
We have a lot of documentation using 3 folders.
Pointing to the 3 paths where the 3 repos are cloned (icu, cldr, and cldr-staging).

There is also code relaying on that.
The cldr-code expects CLDR_DIR to be set. And not set in environment, but as Java property.
The ant did not allow one to set them from command line, but from the ant script itself, and (in some cases) from environment.
And then it used that set value(s) to set both environment AND the Java property.
Somewhere in the code you will see that I explicityl do a System.setProperty to keep thinks the way ant set them up.

So unfortunately it was tangled pretty badly.
I only kept things as they were.

There is quite a bit of refactoring that I would like to do.
And this can be one of them.
But has to also be coordinated with cldr-code, and a lot of doc.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
mihnita: Done

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
mihnita: Accepted. Done.

mihnita:
> Thanks! This looks very good and plausible, but I didn't try to understand all of the Task replacements and configs. Some minor feedback, nothing intelligent.

Thank you.

---

Yes, it also took me quite some time to figure it out, and understand what Ant was doing.
So I am basically doing a minimal Ant replacement :-)

On the way I tried to replace the config with json.
Then I tried to use xml and parse it with Jackson, but Jackson didn't call the setters that Ant was calling, and could not do xml attributes.
To make it work would have required a lot more refactoring.

What I ended up doing (short version)
I create a main class (Cldr2Icu) that simulates (to a degree) the work that Ant was doing.

Basically load the "task classes", load the xml fragment corresponding to each task, call `init()` to do extra initialization and validation, then call `execute()` to run the task proper. Which is kind of what Ant does with each task.

The config file is 100% sub-tree(s) that existed in the Ant build file.
The `project/target[@name="convert-impl"]/convert` in the old xml is now `config/convert`
And all the attributes on that old xml node become CLI options.
(while diffing them now I found some extra comments that I removed, so now they are really the same)

tools/cldr/cldr-to-icu/README.md
markusicu: My question is not about the three folders, but about the three ways of defining them:
- as a shell environment variable
- as a JVM property with -Dname=value
- as a command-line option

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
markusicu: this one is not done

mihnita: Weird. I see it locally, and here:
https://github.com/unicode-org/icu/pull/3290/commits/67705cc76fbeb939ceebfbc831a2bfa7dd06dbca

docs/processes/release/tasks/versions.md
```diff
@@ -53,7 +53,7 @@ need to be correspondingly updated. See below for more files to be updated and s
 [icu4c/source/data/misc/icuver.txt](https://github.com/unicode-org/icu/blob/main/icu4c/source/data/misc/icuver.txt)
 needs to be updated with the correct version number for ICU and its data.
 
-#### Since ICU 68
+#### Since ICU 68, but before ICU 77
```
echeran, line 56: Should there be a "Since ICU 77" version added?  Even if it says to look somewhere else?

The README in the codebase in the `cldr-to-icu` folder points to the `cldr-icu.md` page in the ICU User Guide...

tools/cldr/cldr-to-icu/.classpath
echeran, line 1: Can we get rid of these Eclipse IDE config files, rather than modifying the existing files and adding new ones? Especially since this change moves from Ant to Maven, so Eclipse can't possibly be the standard way to build / test things anymore using Ant.

tools/cldr/cldr-to-icu/README.md
markusicu: Not a blocker, but maybe worth thinking about?

markusicu (dismissed):

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/Cldr2Icu.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private void run(String[] args) {
+        options.processArgs(args);
+        clean();
+        generate();
+    }
+
+    public static void main(String[] args) {
+        new Cldr2Icu().run(args);
```
echeran, line 72: nit: This distinction between `main` and `run` seems artificial. Instead of making `Cldr2Icu` a class that you instantiate, can we not just have every method here be `static`? And then pass options as an argument to the helper methods?

There's nothing inherently stateful, nor are we creating multiple instances of `Cldr2Icu`, so it doesn't "feel" right -- it doesn't seem appropriate for what an instance is (regardless of whether that is semi-idiomatic in Java).

tools/cldr/cldr-to-icu/README.md
mihnita: The existing code and doc assumed both shell environment variable and JVM property

The docs say "set this environment variable" everywhere.
And the defined values are used in a lot of copy / paste commands.

Then the code (in several places, including cldr-code, which is not even icu) assumed that some are set as Java properties.
And failed if not.
Ant was "magically" taking some of the environment variables and making them Java properties

For example:
```xml
<condition property="cldrDataDir" value="${env.CLDR_DATA_DIR}">
    <isset property="env.CLDR_DATA_DIR"/>
</condition>
```

And some were "fixed" properties in the ant script:
```xml
<property name="outDir" value="${basedir}/../../../icu4c/source/data/"/>
```

Then the Ant `convert` invoked Ant again as "convert-impl" by defining CLDR_DIR in environment, and passing other things as Java properties :-(
```xml
<exec executable="ant" searchpath="true" failonerror="true">
    <env key="CLDR_DIR" value="${cldrDataDir}" />
    ...
    <arg value="-DoutDir=${outDir}"/>
    ...
```

And finally the real "convert" task (in the "convert-impl" target) took some of them and invoked setters in the associated Java class (extending Task):
```xml
<convert cldrDir="${cldrDir}" outputDir="${outDir}"  ... />
```

But the point is: some of these were both Java properties and environment variables at the same time.
I only preserved the existing behavior.
AND it is only for the folders.

So the Java code never looked at environment vars, only Ant did.
Java looked at (a few) properties + setters.

---

I had a chat with Elango (and a whiteboard) on what to do about it.

The main question is really between JVM properties vs command line options.
I was split between them, each have pros and cons.

In the end CLI won. The rest needs to be unified.
But not a hight priority, and not in this CL.
It will need to (among others) change the cldr-code library in the cldr repo.

Note that only the 3 folders have this "override behavior", not the rest.
Because they had it already.
Grep the old code (before my CL):
```sh
$ grep -r   System.getProperty src/

src//test/java/org/unicode/icu/tool/cldrtoicu/SupplementalDataTest.java:        Path cldrRoot = Paths.get(System.getProperty("CLDR_DIR"));
src//main/java/org/unicode/icu/tool/cldrtoicu/IcuConverterConfig.java:        Optional.ofNullable(System.getProperty("ICU_DIR", null))
src//main/java/org/unicode/icu/tool/cldrtoicu/DebugWriter.java:        String cldrPath = System.getProperty("CLDR_DIR", System.getenv("CLDR_DIR"));
```

And note this in `DebugWriter.java`:
```java
String cldrPath = System.getProperty("CLDR_DIR", System.getenv("CLDR_DIR"));
```

This assumption was there.
And it was too much surgery to fix in one step.
That's why I was forced to do a `System.setProperty("CLDR_DIR", cldrDir);` (in `Cldr2IcuCliOptions.java`)

---

TLDR: I am not arguing that it is good, or that I like it.
What I am trying to say is that was unclean before, it is the same now.

Deserves a cleanup, but it is not mavenization, and is not as simple as removing this override that I implemented here.

For example can't remove it from cldr-code, because CLDR itself also depends on it.
So the processes / scripts / docs in CLDR would need to be updated BEFORE removing it from this code.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
```diff
@@ -340,4 +338,77 @@ private static ImmutableList<String> readLinesFromResource(String name) {
...
+
+    private static Dir getDirectory(Element element) {
+        if (!"dir".equals(element.getTagName())) {
+            return null;
+        }
+        String name = element.getAttribute("name");
+        Dir dir = new Dir();
+        dir.setName(name);
+        Node node = element.getFirstChild();
+        while (node != null) {
```
echeran, line 360: Any reason why you aren't iterating over `element.getChildNodes()`, and instead manually getting the first and then calling iteratively `getNextSibling()`? Happens elsewhere in the PR.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
mihnita: Ah, same fix in two files.
My bad.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/Task.java
```diff
@@ -0,0 +1,25 @@
+// © 2019 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+package org.unicode.icu.tool.cldrtoicu.ant;
+
+public class Task {
```
echeran, line 5: It looks like `Task` should be an interface with default methods. I don't think that stops you from having inner classes inside it, so such a change would be minimal.

tools/cldr/cldr-to-icu/.classpath
mihnita: +1

But to do that I need to bring back the .gitignore in `tools/cldr`, which I don't quite like...

I think it would be good to do a .gitignore cleanup.
We have 9 of them.

mihnita: Done

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
mihnita: `element.getChildNodes()` returns a `NodeList`

And the only way to iterate a `NodeList` is something like this:
```
NodeList tmp = outputDirectories.item(0).getChildNodes();
for (int i = 0; i < tmp.getLength(); i++) {
    Node node = tmp.item(i);
    ...
}
```
Which means it loads all the nodes in memory.
The getFirst / getNext loads nodes as needed.

I don't think that memory would be a problem for the kind of files here.
It just felt a bit cleaner. No a strong preference.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/Task.java
mihnita: That is what I initially did.

But there are are many classes extending Task, not just the top 3: \
`LocaleIds`, `Directory`, `ForcedAlias`, `AltPath` in `ConvertIcuDataTask` \
`Retain`, `Dir` in `CleanOutputDirectoryTask`

Some implement init, some implement exec, some implement both.

So to make the Task an interface I would need to either add some empty implementations in each of these classes, or add default implementations in the interface.

First option added too much noise (my goal was to keep things as close to the original).

Second option means that instead of:
```java
class Task {
   log() { print }
   init() {}
   exec() {}
}
```
I would have:
```java
interface Task {
   default log() { print }
   default init() {}
   default exec() {}
}
```

It does not look like an improvement.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/Cldr2Icu.java
mihnita: It just felt bad to have a class with 8 static methods.

Options (what do you find cleaner):

1. Leave as is

2. Do all the work in main, without a `run()`
```java
public static void main(String[] args) {
    Cldr2Icu self = new Cldr2Icu();
    self.options.processArgs(args);
    self.clean();
    self.generate();
}
```

3. Close to 2, but parsing arguments in a constructor (feels like an initialization)
```java

public Cldr2Icu(String[] args) {
    options.processArgs(args);
    // or even something like
    // options = Cldr2IcuCliOptions.processArgs(args);
}

public static void main(String[] args) {
    Cldr2Icu self = new Cldr2Icu(args);
    self.clean();
    self.generate();
}
```

I can also make `options` a non-member:
```java
public static void main(String[] args) {
    Cldr2IcuCliOptions options = Cldr2IcuCliOptions.processArgs(args);
    Cldr2Icu self = new Cldr2Icu();
    self.clean(options);
    self.generate(options);
}
```

docs/processes/release/tasks/versions.md
mihnita: "Since ICU 77" would be a NOP :-)

echeran: I think it's worth adding, even if were a no-op. But the fact that the configuration file is no longer in `build-icu-data.xml` but instead in some new thing called `config.xml` is worth pointing out. And you can also point to the README which has usage tips and examples. It's simple & not much, because it doesn't need to be much, and it's more significant than nothing.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/Task.java
echeran: You're right, not worth doing.  An interface's behaviors should be implemented by all participants. If that's not the case, then maybe it's a sign of 2 different behaviors, each with their own interface. This Task design is inherited from the Ant-based way of doing things, and since it works, it's better to leave as is.

markusicu:
sgtm -- unless Elango objects, I can offer a rubber stamp...

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
echeran: Ah, not worth changing. Not more readable, and a minor perf penalty, too.  Guava has `Streams.stream(Iterator)`, which we can't use either because there's no Guava and no Iterator.

The reason I thought to ask is because the code you wrote would be smaller and more readable & maintainable with an FP style via `Stream`.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/Cldr2Icu.java
echeran: My preference is option 2. I think nothing of 8 static methods because it's a hint of statelessness. No stateful work being done , so no chance for mutation of state. I feel bad to see 8 member methods for a stateful class when no state is truly needed. It would also entail making `options` a non-member and passing it throughout, as you suggested in the last example. Overall, that's the "cleanest".

I'm okay with option 1 if you don't think it's worth the time.

My reaction is based on the feeling that it looks very OOP Java idiomatic in a way that feels unnecessarily complicated, which is how a lot of OOP code tends to look.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/ant/CleanOutputDirectoryTask.java
markusicu: FYI: The CLDR-to-ICU converter depends on CLDR which I believe uses Guava, so Guava should be fine here. However, it sounds like this code is good as is anyway.

tools/cldr/cldr-to-icu/src/main/java/org/unicode/icu/tool/cldrtoicu/Cldr2Icu.java
mihnita: Thank you. Went with option 2.

DONE.

docs/processes/release/tasks/versions.md
mihnita: DONE

mihnita:
I also managed to fix the tests.

They were cause by real changes in the data (for example the likely subtags refactoring).
But these didn't fail, because they are not under `icu4j/main`, so nobody runs them.

echeran (dismissed):
LGTM

markusicu (dismissed):

docs/processes/cldr-icu.md
```diff
@@ -415,62 +424,62 @@ appropriate, and repeating from step 4 or 8 as appropriate.
...
 mvn install -pl main/common_tests -Dtest=MeasureUnitTest#TestGreek
 ```
 
-13b. Optionally run the tests in exhautive mode
+13b. Optionally run the tests in exhaustive mode
 
 Optionally run before committing changes, or run to diagnose failures from
-running exhastive CI tests in the PR using `/azp run CI-Exhaustive`: 
-```
+running exhaustive CI tests in the PR using `/azp run CI-Exhaustive`:
```
pedberg-icu, line 475: `/azp run CI-Exhaustive` is no longer valid, right?

markusicu: good catch... --> https://unicode-org.github.io/icu/userguide/dev/ci.html#exhaustive-tests

could be done in a follow-up PR

mihnita: Oups... sorry. Already did it in this one.

I think it is a left-over from Azure.
The command below the description is correct.

I updated this to what I think was the intention.

docs/processes/cldr-icu.md
```diff
@@ -472,12 +472,15 @@ mvn install -pl main/common_tests -Dtest=MeasureUnitTest#TestGreek
 13b. Optionally run the tests in exhaustive mode
 
 Optionally run before committing changes, or run to diagnose failures from
-running exhaustive CI tests in the PR using `/azp run CI-Exhaustive`:
+running exhaustive CI tests in the PR using exhaustive localy:
```
markusicu: running exhaustive CI tests locally:

docs/processes/cldr-icu.md
```diff
@@ -472,12 +472,15 @@ mvn install -pl main/common_tests -Dtest=MeasureUnitTest#TestGreek
...
 ```sh
 cd $ICU4J_ROOT
 mvn install -DICU.exhaustive=10 2>&1 | tee $NOTES/icu4j-newData-mvnCheckEx.txt
 ```
 
+Exhaustive tests in CL can ve triggered by running the "Exhaustive Tests for ICU"
+action from the GitHub web UI.
```
markusicu: typo fix & adding link:
```
Exhaustive tests in CL can be triggered by running the "Exhaustive Tests for ICU"
action from the GitHub web UI:
https://unicode-org.github.io/icu/userguide/dev/ci.html#exhaustive-tests
```
except nicer if we can link to the .md page with a relative link, and GitHub Pages turns that into the link to the .html page

docs/processes/cldr-icu.md
```diff
@@ -471,14 +471,13 @@ mvn install -pl main/common_tests -Dtest=MeasureUnitTest#TestGreek
 
 13b. Optionally run the tests in exhaustive mode
 
-Optionally run before committing changes, or run to diagnose failures from
-running exhaustive CI tests in the PR using exhaustive localy:
+Optionally run exhaustive tests localy before committing changes:
```
markusicu, line 474: typo: localy --> locally

markusicu: reminder

mihnita: DONE

docs/processes/cldr-icu.md
mihnita: Added a link, hope it works.
I used CI (not CL) intentionally. Because that is a CI test.

markusicu: I know, I had missed the "CL" typo

markusicu (approved):

mihnita:
Thank you all!

PR #3294, opened 2024-12-09
ICU-22773 Faster generation (4x) with multithreading

mihnita:
About 4 times faster. Tested on Windows, MacOS, and Linux.

Right now it is opt-in.
Do you think it should be opt-out?

---

All on reasonably modern machines (Win & Linux: Intel CPU, 16 cores (8 Performance + 8 Economy), MacOS on a MacBook M3 Pro, 12 cores).

Why not 10 times faster (or more)? I guess the bottleneck moves somewhere else (disk?)
Good enough for now.

#### Checklist
- [x] Required: Issue filed: ICU-22773


echeran (approved):
LGTM

macchiati:
Did you verify that the same results obtain?

(My only worry is that some of the code called by the tool might not be fully thread-safe.)

mihnita:
> Did you verify that the same results obtain?
> (My only worry is that some of the code called by the tool might not be fully thread-safe.)

Yes, I did, on all 3 platforms.

And that is also why I activate it from command line, and opt-in.
So if something looks fishy we can always do things the slow-but-reliable way.

markusicu (approved):

mihnita:
Thank you!

pedberg-icu (approved):

PR #3305, opened 2024-12-13
ICU-22920 Fix raw type warnings in icu4j core tests

mihnita:
To make the review easier this only fixes one kind of warning:
`XYZ is a raw type. References to generic type XYZ should be parameterized.`

This PR removes about 440 warnings.

Only fixed the warnings in tests, leaving the main code untouched.

The non-test code might needs a bit more care.
Some might be implementation details, private methods / members, easy to change.
But some might affect public APIs

---

There is also a leftover class in tests that need fixing.

`LocaleAliasTest` fails if I fix it, because it is already broken, but no tests run.

There is a generic (typeless) `Map`, initialized with some `String` (key) =>  `Strings` (value).

Later on there are `get(ULocale)` calls, which return `null`, and that test case if skipped.
So no test is executed, because in fact all test cases are skipped.

If I fix the type of the map to `<ULocale, ULocale>` (as it should be), then the real tests are executed and some are failing.

#### Checklist
- [x] Required: Issue filed: ICU-22920

### DIFF
--- a/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
+++ b/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
@@ -129,12 +129,12 @@ public abstract class TestBoilerplate<T> extends TestFmwk {
      * @return clone
      */
     protected T _clone(T a) throws Exception {
-        Class aClass = a.getClass();
+        Class<?> aClass = a.getClass();
         try {
             Method cloner = aClass.getMethod("clone", (Class[])null);
             return (T) cloner.invoke(a,(Object[])null);
         } catch (NoSuchMethodException e) {
-            Constructor constructor = aClass.getConstructor(new Class[] {aClass});
+            Constructor<?> constructor = aClass.getConstructor(new Class[] {aClass});
             return (T) constructor.newInstance(new Object[]{a});
         }
     }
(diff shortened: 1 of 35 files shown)


icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
```diff
@@ -15,6 +15,7 @@
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
```
markusicu, line 18: Would you mind using Map.Entry in the code, rather than importing it?
I find that these nested-class imports generally lead to less readable call sites.
"What in the world is this Entry?"...

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
```diff
@@ -150,26 +151,25 @@ public static boolean verifySetsIdentical(AbstractTestLog here, UnicodeSet set1,
...
+        Set<T> temp;
         TestFmwk.errln("Values differ:");
         TestFmwk.errln("UnicodeMap - HashMap");
-        temp = new TreeSet(values1);
+        temp = new TreeSet<>(values1);
         temp.removeAll(values2);
         TestFmwk.errln(show(temp));
         TestFmwk.errln("HashMap - UnicodeMap");
-        temp = new TreeSet(values2);
+        temp = new TreeSet<T>(values2);
```
markusicu, line 163: Can't you remove the T from the <>?

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/UnicodeMapTest.java
```diff
@@ -27,6 +27,7 @@
 import org.junit.runners.JUnit4;
 
 import com.ibm.icu.impl.UnicodeMap;
+import com.ibm.icu.impl.UnicodeMap.EntryRange;
```
markusicu, line 30: optional: slight preference for not importing this

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/util/ICUServiceTestSample.java
```diff
@@ -60,13 +59,11 @@ public void helloServiceChanged() {
...
         private void display() {
-            Map names = HelloService.getDisplayNames(ULocale.US);
+            Map<String, String> names = HelloService.getDisplayNames(ULocale.US);
             System.out.println("displaying " + names.size() + " names.");
-            Iterator iter = names.entrySet().iterator();
-            while (iter.hasNext()) {
-                Entry entry = (Entry)iter.next();
-                String displayName = (String)entry.getKey();
-                HelloService service = HelloService.get((String)entry.getValue());
+            for (Entry<String, String> entry : names.entrySet()) {
```
markusicu, line 64: ditto optional: not import Map.Entry (I know, pre-existing)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
mihnita: I don't mind. That is also my preference.
I guess Eclipse added it and didn't notice it.

Done.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/TestBoilerplate.java
mihnita: Done

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/UnicodeMapTest.java
mihnita: Done

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/util/ICUServiceTestSample.java
mihnita: Done

markusicu (approved):

PR #3311, opened 2024-12-18
Fix circular include on MS Visual Studio

ignazp:
Including ucal.h from a C file on MS VisualStudio triggers a circular include, which triggers a build fail.

The circualr include can be resumed as:
utypes.h -> umachine.h -> ptypes.h -> uchar.h -> utypes.h

The first reported error is
```
...\x64\Release\include\unicode\ucpmap.h(12,1): error C2054: expected '(' to follow 'U_CDECL_BEGIN'
70>(compiling source file 'test.c')
```

### DIFF
--- a/icu4c/source/common/unicode/platform.h
+++ b/icu4c/source/common/unicode/platform.h
@@ -728,12 +728,12 @@
     /*
      * Notes:
      * C++11 and C11 require support for UTF-16 literals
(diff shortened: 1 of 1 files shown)


sffc (changes requested):
This seems overly broad? Most MSVC should have `char16_t`. Is this intended to specifically target C but not C++?

PR #3321, opened 2024-12-24
ICU-23000 Replace CharString for LocaleBased

FrankYFTang:
These changes improve the overall peak memory usage and slightly improve
performance
```
	main avg	main stdev	branch avg	branch stdev	diff avg	diff avg %
User time (seconds):	81.19	0.37	81.36	0.09	0.17	0.21%
System time (seconds):	5.03	0.08	4.86	0.05	-0.17	-3.38%
Percent of CPU this job got:	237.33%	0.01	235.33%	0.01	-2.00%	-0.84%
Maximum resident set size (kbytes):	227,141	99.14	226,667	366.48	-475	-0.21%
Minor (reclaiming a frame) page faults:	46,009	510.78	43,775	294.47	-2,234	-4.86%
Voluntary context switches:	333,442	4513.76	316,506	8850.48	-16,936	-5.08%
Involuntary context switches:	467	30.60	515	22.65	48	10.36%
```
Raw data in
https://docs.google.com/spreadsheets/d/10LQ3kO83EwvsFfoiySTzsTbIzfUV00Te4jQnX0b_FM4/edit?usp=sharing

#### Checklist
- [X] Required: Issue filed: ICU-23000

### DIFF
--- a/icu4c/source/common/brkiter.cpp
+++ b/icu4c/source/common/brkiter.cpp
@@ -59,7 +59,7 @@ BreakIterator::buildInstance(const Locale& loc, const char *type, UErrorCode &st
 {
     char fnbuff[256];
     char ext[4]={'\0'};
-    CharString actualLocale;
+    CharString actual;
     int32_t size;
     const char16_t* brkfname = nullptr;
     UResourceBundle brkRulesStack;
@@ -94,7 +94,7 @@ BreakIterator::buildInstance(const Locale& loc, const char *type, UErrorCode &st
 
         // Use the string if we found it
         if (U_SUCCESS(status) && brkfname) {
-            actualLocale.append(ures_getLocaleInternal(brkName, &status), -1, status);
+            actual.append(ures_getLocaleInternal(brkName, &status), -1, status);
 
             char16_t* extStart=u_strchr(brkfname, 0x002e);
             int len = 0;
@@ -123,10 +123,9 @@ BreakIterator::buildInstance(const Locale& loc, const char *type, UErrorCode &st
     if (U_SUCCESS(status) && result != nullptr) {
         U_LOCALE_BASED(locBased, *(BreakIterator*)result);
 
-        locBased.setLocaleIDs(ures_getLocaleByType(b, ULOC_VALID_LOCALE, &status), 
-                              actualLocale.data());
-        uprv_strncpy(result->requestLocale, loc.getName(), ULOC_FULLNAME_CAPACITY);
-        result->requestLocale[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
+        locBased.setLocaleIDs(ures_getLocaleByType(b, ULOC_VALID_LOCALE, &status),
+                              actual.data(), status);
+        LocaleBased::setLocaleID(loc.getName(), result->requestLocale, status);
     }
 
     ures_close(b);
@@ -206,26 +205,32 @@ BreakIterator::getAvailableLocales(int32_t& count)
 
 BreakIterator::BreakIterator()
 {
-    *validLocale = *actualLocale = *requestLocale = 0;
 }
 
 BreakIterator::BreakIterator(const BreakIterator &other) : UObject(other) {
-    uprv_strncpy(actualLocale, other.actualLocale, sizeof(actualLocale));
-    uprv_strncpy(validLocale, other.validLocale, sizeof(validLocale));
-    uprv_strncpy(requestLocale, other.requestLocale, sizeof(requestLocale));
+    UErrorCode status = U_ZERO_ERROR;
+    U_LOCALE_BASED(locBased, *this);
+    locBased.setLocaleIDs(other.validLocale, other.actualLocale, status);
+    LocaleBased::setLocaleID(other.requestLocale, requestLocale, status);
+    U_ASSERT(U_SUCCESS(status));
 }
 
 BreakIterator &BreakIterator::operator =(const BreakIterator &other) {
     if (this != &other) {
-        uprv_strncpy(actualLocale, other.actualLocale, sizeof(actualLocale));
-        uprv_strncpy(validLocale, other.validLocale, sizeof(validLocale));
-        uprv_strncpy(requestLocale, other.requestLocale, sizeof(requestLocale));
+        UErrorCode status = U_ZERO_ERROR;
+        U_LOCALE_BASED(locBased, *this);
+        locBased.setLocaleIDs(other.validLocale, other.actualLocale, status);
+        LocaleBased::setLocaleID(other.requestLocale, requestLocale, status);
+        U_ASSERT(U_SUCCESS(status));
     }
     return *this;
 }
 
 BreakIterator::~BreakIterator()
 {
+    delete validLocale;
+    delete actualLocale;
+    delete requestLocale;
 }
 
 // ------------------------------------------
@@ -394,7 +399,7 @@ BreakIterator::createInstance(const Locale& loc, int32_t kind, UErrorCode& statu
         // revisit this in ICU 3.0 and clean it up/fix it/remove it.
         if (U_SUCCESS(status) && (result != nullptr) && *actualLoc.getName() != 0) {
             U_LOCALE_BASED(locBased, *result);
-            locBased.setLocaleIDs(actualLoc.getName(), actualLoc.getName());
+            locBased.setLocaleIDs(actualLoc.getName(), actualLoc.getName(), status);
         }
         return result;
     }
@@ -488,6 +493,7 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
     }
 
     if (U_FAILURE(status)) {
+        delete result;
         return nullptr;
     }
 
@@ -496,20 +502,25 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
 
 Locale
 BreakIterator::getLocale(ULocDataLocaleType type, UErrorCode& status) const {
-    if (type == ULOC_REQUESTED_LOCALE) {
-        return {requestLocale};
+    if (U_FAILURE(status)) {
+        return Locale::getRoot();
     }
-    U_LOCALE_BASED(locBased, *this);
-    return locBased.getLocale(type, status);
+    if (type == ULOC_REQUESTED_LOCALE) {
+        return requestLocale == nullptr ?
+            Locale::getRoot() : Locale(requestLocale->data());
+    }
+    return LocaleBased::getLocale(validLocale, actualLocale, type, status);
 }
 
 const char *
 BreakIterator::getLocaleID(ULocDataLocaleType type, UErrorCode& status) const {
-    if (type == ULOC_REQUESTED_LOCALE) {
-        return requestLocale;
+    if (U_FAILURE(status)) {
+        return nullptr;
     }
-    U_LOCALE_BASED(locBased, *this);
-    return locBased.getLocaleID(type, status);
+    if (type == ULOC_REQUESTED_LOCALE) {
+        return requestLocale == nullptr ?  "" : requestLocale->data();
+    }
+    return LocaleBased::getLocaleID(validLocale, actualLocale, type, status);
 }
 
 
@@ -536,8 +547,10 @@ int32_t BreakIterator::getRuleStatusVec(int32_t *fillInVec, int32_t capacity, UE
 }
 
 BreakIterator::BreakIterator (const Locale& valid, const Locale& actual) {
+  UErrorCode status = U_ZERO_ERROR;
   U_LOCALE_BASED(locBased, (*this));
-  locBased.setLocaleIDs(valid, actual);
+  locBased.setLocaleIDs(valid.getName(), actual.getName(), status);
+  U_ASSERT(U_SUCCESS(status));
 }
 
 U_NAMESPACE_END
--- a/icu4c/source/common/locbased.cpp
+++ b/icu4c/source/common/locbased.cpp
@@ -12,44 +12,84 @@
 */
 #include "locbased.h"
 #include "cstring.h"
+#include "charstr.h"
 
 U_NAMESPACE_BEGIN
 
-Locale LocaleBased::getLocale(ULocDataLocaleType type, UErrorCode& status) const {
-    const char* id = getLocaleID(type, status);
+Locale LocaleBased::getLocale(const CharString* valid, const CharString* actual,
+                              ULocDataLocaleType type, UErrorCode& status) {
+    const char* id = getLocaleID(valid, actual, type, status);
     return Locale(id != nullptr ? id : "");
 }
 
-const char* LocaleBased::getLocaleID(ULocDataLocaleType type, UErrorCode& status) const {
+const char* LocaleBased::getLocaleID(const CharString* valid, const CharString* actual,
+                                     ULocDataLocaleType type, UErrorCode& status) {
     if (U_FAILURE(status)) {
         return nullptr;
     }
 
     switch(type) {
     case ULOC_VALID_LOCALE:
-        return valid;
+        return valid == nullptr ? "" : valid->data();
     case ULOC_ACTUAL_LOCALE:
-        return actual;
+        return actual == nullptr ? "" : actual->data();
     default:
         status = U_ILLEGAL_ARGUMENT_ERROR;
         return nullptr;
     }
 }
 
-void LocaleBased::setLocaleIDs(const char* validID, const char* actualID) {
-    if (validID != nullptr) {
-      uprv_strncpy(valid, validID, ULOC_FULLNAME_CAPACITY);
-      valid[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
-    }
-    if (actualID != nullptr) {
-      uprv_strncpy(actual, actualID, ULOC_FULLNAME_CAPACITY);
-      actual[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
+void LocaleBased::setLocaleIDs(const CharString* validID, const CharString* actualID, UErrorCode& status) {
+    setValidLocaleID(validID, status);
+    setActualLocaleID(actualID,status);
+}
+void LocaleBased::setLocaleIDs(const char* validID, const char* actualID, UErrorCode& status) {
+    setValidLocaleID(validID, status);
+    setActualLocaleID(actualID,status);
+}
+
+void LocaleBased::setLocaleID(const char* id, CharString*& dest, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    if (id == nullptr || *id == 0) {
+        delete dest;
+        dest = nullptr;
+    } else {
+        if (dest == nullptr) {
+            dest = new CharString(id, status);
+            if (dest == nullptr) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+                return;
+            }
+        } else {
+            dest->copyFrom(id, status);
+        }
     }
 }
 
-void LocaleBased::setLocaleIDs(const Locale& validID, const Locale& actualID) {
-  uprv_strcpy(valid, validID.getName());
-  uprv_strcpy(actual, actualID.getName());
+void LocaleBased::setLocaleID(const CharString* id, CharString*& dest, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    if (id == nullptr || id->isEmpty()) {
+        delete dest;
+        dest = nullptr;
+    } else {
+        if (dest == nullptr) {
+            dest = new CharString(*id, status);
+            if (dest == nullptr) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+                return;
+            }
+        } else {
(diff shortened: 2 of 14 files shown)


icu4c/source/common/brkiter.cpp
```diff
@@ -123,10 +123,16 @@ BreakIterator::buildInstance(const Locale& loc, const char *type, UErrorCode &st
...
         U_LOCALE_BASED(locBased, *(BreakIterator*)result);
 
-        locBased.setLocaleIDs(ures_getLocaleByType(b, ULOC_VALID_LOCALE, &status), 
-                              actualLocale.data());
-        uprv_strncpy(result->requestLocale, loc.getName(), ULOC_FULLNAME_CAPACITY);
-        result->requestLocale[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
+        locBased.setLocaleIDs(ures_getLocaleByType(b, ULOC_VALID_LOCALE, &status),
+                              actualLocale.data(), status);
+        if (U_FAILURE(status)) {
+            return nullptr;
```
markusicu, line 129: These returns skip the `ures_close(b)` and `delete result` below. If you stop checking for failures and returning early, you avoid this problem.

icu4c/source/common/locbased.h
```diff
@@ -19,13 +19,14 @@
 /**
  * Macro to declare a locale LocaleBased wrapper object for the given
  * object, which must have two members named `validLocale' and
- * `actualLocale' of size ULOC_FULLNAME_CAPACITY
+ * `actualLocale' of CharString.
```
markusicu, line 22:
```suggestion
 * `actualLocale' which are pointers to the internal icu::CharString.
```

icu4c/source/common/locbased.h
```diff
@@ -41,13 +42,13 @@ class U_COMMON_API LocaleBased : public UMemory {
...
      */
-    inline LocaleBased(char* validAlias, char* actualAlias);
+    inline LocaleBased(CharString* validAlias, CharString* actualAlias);
 
     /**
      * Construct a LocaleBased wrapper around the two const pointers.
      * These will be aliased for the lifetime of this object.
      */
-    inline LocaleBased(const char* validAlias, const char* actualAlias);
+    inline LocaleBased(const CharString* validAlias, const CharString* actualAlias);
```
markusicu, line 51: This, and the implementation which casts away const, looks wrong.
None of the classes that use LocaleBased should have `const CharString`.

Please try to remove this constructor completely.

icu4c/source/common/locbased.cpp
```diff
@@ -27,29 +28,38 @@ const char* LocaleBased::getLocaleID(ULocDataLocaleType type, UErrorCode& status
...
-void LocaleBased::setLocaleIDs(const char* validID, const char* actualID) {
+void LocaleBased::setLocaleIDs(const char* validID, const char* actualID, UErrorCode& status) {
+    if (U_FAILURE(status)) {
+        return;
+    }
     if (validID != nullptr) {
-      uprv_strncpy(valid, validID, ULOC_FULLNAME_CAPACITY);
-      valid[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
+      valid->clear();
+      valid->append(validID, -1, status);
```
markusicu, line 46: Hm. We should really have a CharString function overload
```
    CharString &copyFrom(StringPiece s, UErrorCode &errorCode);
```
to simplify such call sites. In particular, that would eliminate error checking because it wouldn't clear & replace when there is already a failure.

I can work on that as a follow-up if you like.

icu4c/source/common/brkiter.cpp
```diff
@@ -204,28 +210,34 @@ BreakIterator::getAvailableLocales(int32_t& count)
...
-    *validLocale = *actualLocale = *requestLocale = 0;
 }
 
-BreakIterator::BreakIterator(const BreakIterator &other) : UObject(other) {
-    uprv_strncpy(actualLocale, other.actualLocale, sizeof(actualLocale));
-    uprv_strncpy(validLocale, other.validLocale, sizeof(validLocale));
-    uprv_strncpy(requestLocale, other.requestLocale, sizeof(requestLocale));
+BreakIterator::BreakIterator(const BreakIterator &other) : UObject(other), actualLocale(new CharString()), validLocale(new CharString), requestLocale(new CharString()) {
+    UErrorCode status = U_ZERO_ERROR;
+    actualLocale->copyFrom(*(other.actualLocale), status);
```
markusicu, line 219: Before this line, please check if any of the three pointers is a nullptr. If so, then set the memory-allocation-error. Otherwise, copy.

icu4c/source/common/brkiter.cpp
```diff
@@ -535,9 +547,11 @@ int32_t BreakIterator::getRuleStatusVec(int32_t *fillInVec, int32_t capacity, UE
     return 1;
 }
 
-BreakIterator::BreakIterator (const Locale& valid, const Locale& actual) {
+BreakIterator::BreakIterator (const Locale& valid, const Locale& actual) : actualLocale(new CharString()), validLocale(new CharString), requestLocale(new CharString()) {
   U_LOCALE_BASED(locBased, (*this));
-  locBased.setLocaleIDs(valid, actual);
+  UErrorCode status = U_ZERO_ERROR;
+  locBased.setLocaleIDs(valid, actual, status);
```
markusicu, line 553: Before this line, please check if any of the three pointers is a nullptr. If so, then set the memory-allocation-error.

icu4c/source/i18n/calendar.cpp
```diff
@@ -705,10 +705,10 @@ fTime(0),
...
-fSkippedWallTime(UCAL_WALLTIME_LAST)
+fSkippedWallTime(UCAL_WALLTIME_LAST),
+validLocale(new CharString()),
+actualLocale(new CharString())
 {
-    validLocale[0] = 0;
-    actualLocale[0] = 0;
     clear();
     if (U_FAILURE(success)) {
         return;
```
markusicu, line 714: 3 lines below: also check if the two locale pointers are nullptr

icu4c/source/i18n/calendar.cpp
```diff
@@ -733,10 +733,10 @@ fTime(0),
...
-fSkippedWallTime(UCAL_WALLTIME_LAST)
+fSkippedWallTime(UCAL_WALLTIME_LAST),
+validLocale(new CharString()),
+actualLocale(new CharString())
 {
-    validLocale[0] = 0;
-    actualLocale[0] = 0;
     if (U_FAILURE(success)) {
         delete zone;
         return;
```
markusicu, line 742: 2 lines below: check if the two locale pointers are nullptr; if so: mem-alloc-error, and delete the zone if you didn't use a LocalPointer for it

icu4c/source/i18n/calendar.cpp
```diff
@@ -733,10 +733,10 @@ fTime(0),
...
 fRepeatedWallTime(UCAL_WALLTIME_LAST),
-fSkippedWallTime(UCAL_WALLTIME_LAST)
+fSkippedWallTime(UCAL_WALLTIME_LAST),
+validLocale(new CharString()),
+actualLocale(new CharString())
 {
-    validLocale[0] = 0;
-    actualLocale[0] = 0;
     if (U_FAILURE(success)) {
         delete zone;
```
markusicu, line 741: optional: in order to track the adopted zone, we could stick it into a LocalPointer at the start of the constructor body, and orphan() it when we copy the pointer into fZone

icu4c/source/i18n/calendar.cpp
```diff
@@ -768,10 +768,10 @@ fTime(0),
...
 fRepeatedWallTime(UCAL_WALLTIME_LAST),
-fSkippedWallTime(UCAL_WALLTIME_LAST)
+fSkippedWallTime(UCAL_WALLTIME_LAST),
+validLocale(new CharString()),
+actualLocale(new CharString())
 {
-    validLocale[0] = 0;
-    actualLocale[0] = 0;
     if (U_FAILURE(success)) {
         return;
```
markusicu, line 776: 2 lines below: check if the two locale pointers are nullptr

icu4c/source/i18n/calendar.cpp
```diff
@@ -828,10 +830,10 @@ Calendar::operator=(const Calendar &right)
         fWeekendCease            = right.fWeekendCease;
         fWeekendCeaseMillis      = right.fWeekendCeaseMillis;
         fNextStamp               = right.fNextStamp;
-        uprv_strncpy(validLocale, right.validLocale, sizeof(validLocale)-1);
-        validLocale[sizeof(validLocale)-1] = 0;
-        uprv_strncpy(actualLocale, right.actualLocale, sizeof(actualLocale)-1);
-        actualLocale[sizeof(validLocale)-1] = 0;
+        UErrorCode status = U_ZERO_ERROR;
+        validLocale->copyFrom(*right.validLocale, status);
```
markusicu, line 834: first check both for nullptr

icu4c/source/i18n/dcfmtsym.cpp
```diff
@@ -99,26 +99,25 @@ static const char *gNumberElementKeys[DecimalFormatSymbols::kFormatSymbolCount]
 // Initializes this with the decimal format symbols in the default locale.
 
 DecimalFormatSymbols::DecimalFormatSymbols(UErrorCode& status)
-        : UObject(), locale() {
+        : UObject(), locale(), actualLocale(new CharString()), validLocale(new CharString()) {
     initialize(locale, status, true);
```
markusicu, line 103: make sure that initialize() checks the locale pointers for nullptr

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -291,13 +291,13 @@ DateFormatSymbols::createForLocale(
 
 DateFormatSymbols::DateFormatSymbols(const Locale& locale,
                                      UErrorCode& status)
-    : UObject()
+    : UObject(), validLocale(new CharString()), actualLocale(new CharString())
 {
   initializeData(locale, nullptr,  status);
```
markusicu, line 296: make sure that initializeData() checks the locale pointers for nullptr

icu4c/source/i18n/format.cpp
```diff
@@ -70,22 +71,23 @@ FieldPosition::clone() const {
 // default constructor
 
 Format::Format()
-    : UObject()
+    : UObject(), actualLocale(new CharString()), validLocale(new CharString())
 {
```
markusicu, line 75: I am troubled by constructors here and elsewhere that don't take a UErrorCode and therefore cannot report failures when the allocations failed (out of memory).

Would it be feasible to track down subclass constructors that invoke the respective super class constructors?

---
Other idea:
- We could try to default-construct these locale string pointers with nullptr
- construct LocaleBased with *references* to the two pointers, so that it can allocate CharString's if needed
- handle nullptr in the getters: return ""
- handle nullptr in the setters: if nullptr then try to allocate

This would also save a little memory when these are not set, or set to the empty string.

icu4c/source/common/brkiter.cpp
FrankYFTang: according https://en.cppreference.com/w/cpp/memory/new/operator_new if the new CharString() fail, it will not return nullptr but throw exception. So none of these three pointers could be a nullptr right?

roubert: As far as I know, it is possible to build ICU4C in environments where operator `new` can return `nullptr` upon failure.

FrankYFTang:
PTAL

icu4c/source/common/brkiter.cpp
markusicu: - ICU does not work with exceptions
- ICU allows build-time customization of uprv_malloc() & uprv_free()
- ICU new & delete are overridden in base class UMemory to call uprv_malloc() & uprv_free()
- thus, nullptr is what we get when out of memory

icu4c/source/i18n/dcfmtsym.cpp
```diff
@@ -201,10 +206,37 @@ DecimalFormatSymbols::operator==(const DecimalFormatSymbols& that) const
             return false;
         }
     }
+    if (locale != that.locale) return false;
+
+    if (validLocale == nullptr) {
```
markusicu, line 211: This is begging for a new
```
static boolean LocaleBased::equalIDs(const CharString* left, const CharString* right);
```

markusicu (changes requested):

FrankYFTang:
PTAL

icu4c/source/common/brkiter.cpp
```diff
@@ -488,6 +493,9 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
     }
 
     if (U_FAILURE(status)) {
+        if (result != nullptr) {
+            delete result;
+        }
```
markusicu, line 498: no need to check
```suggestion
        delete result;
```

icu4c/source/common/brkiter.cpp
```diff
@@ -496,20 +504,25 @@ BreakIterator::makeInstance(const Locale& loc, int32_t kind, UErrorCode& status)
...
 }
 
 const char *
 BreakIterator::getLocaleID(ULocDataLocaleType type, UErrorCode& status) const {
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
     if (type == ULOC_REQUESTED_LOCALE) {
-        return requestLocale;
+        return requestLocale == nullptr ?  nullptr : requestLocale->data();
```
markusicu, line 523: avoid behavior change; non-error return used to never be nullptr
```suggestion
        return requestLocale == nullptr ?  "" : requestLocale->data();
```

icu4c/source/common/locbased.cpp
```diff
@@ -12,44 +12,83 @@
...
-const char* LocaleBased::getLocaleID(ULocDataLocaleType type, UErrorCode& status) const {
+const char* LocaleBased::getLocaleID(const CharString* valid, const CharString* actual, ULocDataLocaleType type, UErrorCode& status) {
     if (U_FAILURE(status)) {
         return nullptr;
     }
 
     switch(type) {
     case ULOC_VALID_LOCALE:
-        return valid;
+        return valid == nullptr ? nullptr : valid->data();
```
markusicu, line 32: maintain behavior
```suggestion
        return valid == nullptr ? "" : valid->data();
```

icu4c/source/common/locbased.cpp
```diff
@@ -12,44 +12,83 @@
...
     case ULOC_ACTUAL_LOCALE:
-        return actual;
+        return actual == nullptr ? nullptr : actual->data();
```
markusicu, line 34:
```suggestion
        return actual == nullptr ? "" : actual->data();
```

icu4c/source/common/locbased.cpp
```diff
@@ -12,44 +12,83 @@
...
+    setActualLocaleID(actualID,status);
+}
+void LocaleBased::setLocaleIDs(const char* validID, const char* actualID, UErrorCode& status) {
+    setValidLocaleID(validID, status);
+    setActualLocaleID(actualID,status);
+}
+
+void LocaleBased::setLocaleID(const char* id, CharString*& dest, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    if (id == nullptr) {
```
markusicu, line 52: normalize on input
```suggestion
    if (id == nullptr || *id == 0) {
```

icu4c/source/common/locbased.cpp
```diff
@@ -12,44 +12,83 @@
...
+        }
     }
-    if (actualID != nullptr) {
-      uprv_strncpy(actual, actualID, ULOC_FULLNAME_CAPACITY);
-      actual[ULOC_FULLNAME_CAPACITY-1] = 0; // always terminate
+}
+
+void LocaleBased::setLocaleID(const CharString* id, CharString*& dest, UErrorCode& status) {
+    if (U_FAILURE(status)) { return; }
+    if (id == nullptr) {
```
markusicu, line 70:
```suggestion
    if (id == nullptr || id->isEmpty()) {
```

markusicu (commented):
- I like it now :-)
- I realized that the pending changes return nullptr in non-error cases; we should go back to returning an empty string as before -- hopefully I caught the relevant places in the code
- a few stylistic changes

richgillam:
I'm sorry I'm getting to this late.  Markus has already worked this PR over pretty thoroughly, so I may not be bringing anything useful to the party (or at least anything useful enough to justify reopening the PR or doing a follow-on PR).  But I have a couple thoughts:

1. There's a spot in the documentation for `LocaleBased` that says it's likely to be removed in ICU 3.0.  That obviously didn't happen, so you might want to update that comment.

2. Even with all the code in this PR that uses it, I couldn't get my mind around `LocaleBased` or its intended usage model.  I kinda get it, and I think it's probably fine (and it's clearly been there forever), but it feels weird.  Some of that might be the name, which seems strange and non-reflective of a utility class that lives for a short period of time on the stack.

3. I'm also uncomfortable with this usage of `CharString`.  As I understand it, `CharString` is generally intended to be used on the stack or as an inline member of a class; having a bunch objects that contain _pointers_ to `CharString` feels uncomfortable and perhaps likely to lead to errors.  Furthermore, the whole idea of `CharString` is to allow the user to use an inline `char` buffer for small strings and seamless move to a heap-allocated buffer for larger strings.  If everything is a pointer to `CharString`, you're _always_ using a heap-allocated buffer, and when the string is long, you're using _two_ heap-allocated buffers.  But it does give you the advantage of the empty string being just a null pointer.  (Unless, of course, all these pointers are pointers to inline or stack-allocated `CharString`s that are owned by something else, but that isn't what it looked like.)

I can't deny the memory-usage advantages you're getting here-- this is clearly a big win over what we had before-- but I can't help wondering if `CharString`, at least as it's currently implemented, is really the best tool for the job, or whether you want something new that's more tailored to the job you're actually doing here.

At the very least, it might be nice to encapsulate the pointer-ness in a new `CharStringOrNull` class that does the empty-string conversion automatically and could be swapped out later for something less heap-dependent.

But I'm fine if you don't do any of this, and I'm _definitely_ fine if you just file a ticket to do it later.  What you have is a clear improvement over what was there before.

markusicu:
> 1. There's a spot in the documentation for `LocaleBased` that says it's likely to be removed in ICU 3.0.  That obviously didn't happen, so you might want to update that comment.

good point

> 2. Even with all the code in this PR that uses it, I couldn't get my mind around `LocaleBased` or its intended usage model.  I kinda get it, and I think it's probably fine (and it's clearly been there forever), but it feels weird.  Some of that might be the name, which seems strange and non-reflective of a utility class that lives for a short period of time on the stack.

Yes, it's very weird. It's really just a temporary helper for setting and fetching two common locale IDs.

It would be more natural to have an additional base class for Format and BreakIterator etc., but I suspect that this hack was introduced because additional base classes are messy in C++ and add a vtable pointer.

> 3. I'm also uncomfortable with this usage of `CharString`.  As I understand it, `CharString` is generally intended to be used on the stack or as an inline member of a class; having a bunch objects that contain _pointers_ to `CharString` feels uncomfortable and perhaps likely to lead to errors.  Furthermore, ...

The reason for using CharString here is mainly to stop having a fixed-size buffer, and to avoid adding manual, error-prone buffer-and-pointer-and-length management.

It would be much better to have a string-by-value field, but CharString is internal, and used in public APIs, so the best we can do is forward-declare it in the public headers and have pointer fields.

> I can't deny the memory-usage advantages you're getting here

The nullptr-for-empty-ID part is not really for saving memory but for keeping constructors manageable.

> At the very least, it might be nice to encapsulate the pointer-ness in a new `CharStringOrNull` class that does the empty-string conversion automatically and could be swapped out later for something less heap-dependent.

I think that that would suffer from the same internal vs. public API problem.

It seems like a cleaner way around that would be to create a public-but-`@internal` clone of CharString, maybe simplified a bit. But that would be more work. Or make CharString itself public-but-`@internal`?? Or public???

richgillam:
> It would be more natural to have an additional base class for Format and BreakIterator etc., but I suspect that this hack was introduced because additional base classes are messy in C++ and add a vtable pointer.

Would it need to be a base class, or could it just be a separate class that the client classes include as an inline member?  Or are their uses too different for that?

> It would be much better to have a string-by-value field, but CharString is internal, and used in public APIs, so the best we can do is forward-declare it in the public headers and have pointer fields.

Ick.  I forgot about that.  Okay, you've convinced me.  It's probably not worth messing with further, at least right now.  :-)

markusicu:
> > It would be more natural to have an additional base class for Format and BreakIterator etc., but I suspect that this hack was introduced because additional base classes are messy in C++ and add a vtable pointer.
> 
> Would it need to be a base class, or could it just be a separate class that the client classes include as an inline member? Or are their uses too different for that?

A class-as-value-field should work. The class would have to be public-but-`@internal`.

FrankYFTang:
> A class-as-value-field should work. The class would have to be public-but-`@internal`.

ok, I file a new bug to track that in https://unicode-org.atlassian.net/browse/ICU-23005 I will work on that.

roubert:
> I'm also uncomfortable with this usage of `CharString`.

That has now been resolved by PR #3596.

PR #3326, opened 2025-01-06
ICU-22979 Support inverse rule for [] span in RBNF

grhoten:
This feature enhancement only affects the documentation and RBNF syntax. Some changes were made to keep both the ICU4J and ICU4C implementations in sync.  Some compiler warnings were also fixed.

I’d like to extend the RBNF syntax to support more complex grammar.  I’d like to change the omission rule with square brackets.  By default, everything between the square brackets are omitted when the remainder is 0.  My proposal will not change this behavior by default, unless a “|” (pipe symbol) is present between the square brackets.  You can think of it performing like an else statement.  Everything between the beginning square bracket and the pipe acts as it currently does.  Everything between the pipe symbol and the end square bracket will be used instead of omitting the text.

This behavior is important for supporting large ordinals in slavic languages.  It’s convenient for other languages, like English.

The test case in the prototype and the ticket provides more examples of the change.  Below is a simplified example of the new syntax.  Right now, we have the following ordinals in English.
```
%%tieth:
0: tieth;
1: ty-=%spellout-ordinal=;
%spellout-ordinal:
...
20: twen>%%tieth>;
30: thir>%%tieth>;
40: for>%%tieth>;
50: fif>%%tieth>;
```
That could be simplified to the following rules instead.
```
%spellout-ordinal:
...
20: twent[y->>|ieth];
30: thirt[y->>|ieth];
40: fort[y->>|ieth];
50: fift[y->>|ieth];
```
The cardinal and ordinal rules will work on either side of the pipe symbol.

#### Checklist
- [x] Required: Issue filed: ICU-22979

### DIFF
--- a/icu4c/source/i18n/nfrule.cpp
+++ b/icu4c/source/i18n/nfrule.cpp
@@ -64,6 +64,7 @@ NFRule::~NFRule()
 
 static const char16_t gLeftBracket = 0x005b;
 static const char16_t gRightBracket = 0x005d;
+static const char16_t gVerticalLine = 0x007C;
 static const char16_t gColon = 0x003a;
 static const char16_t gZero = 0x0030;
 static const char16_t gNine = 0x0039;
@@ -146,6 +147,7 @@ NFRule::makeRules(UnicodeString& description,
         // then it's really shorthand for two rules (with one exception)
         LocalPointer<NFRule> rule2;
         UnicodeString sbuf;
+        int32_t orElseOp = description.indexOf(gVerticalLine);
 
         // we'll actually only split the rule into two rules if its
         // base value is an even multiple of its divisor (or it's one
@@ -193,9 +195,13 @@ NFRule::makeRules(UnicodeString& description,
             rule2->radix = rule1->radix;
             rule2->exponent = rule1->exponent;
 
-            // rule2's rule text omits the stuff in brackets: initialize
-            // its rule text and substitutions accordingly
+            // By default, rule2's rule text omits the stuff in brackets,
+            // unless it contains a | between the brackets.
+            // Initialize its rule text and substitutions accordingly.
             sbuf.append(description, 0, brack1);
+            if (orElseOp >= 0) {
+                sbuf.append(description, orElseOp + 1, brack2 - orElseOp - 1);
+            }
             if (brack2 + 1 < description.length()) {
                 sbuf.append(description, brack2 + 1, description.length() - brack2 - 1);
             }
@@ -206,7 +212,12 @@ NFRule::makeRules(UnicodeString& description,
         // the brackets themselves: initialize _its_ rule text and
         // substitutions accordingly
         sbuf.setTo(description, 0, brack1);
-        sbuf.append(description, brack1 + 1, brack2 - brack1 - 1);
+        if (orElseOp >= 0) {
+            sbuf.append(description, brack1 + 1, orElseOp - brack1 - 1);
+        }
+        else {
+            sbuf.append(description, brack1 + 1, brack2 - brack1 - 1);
+        }
         if (brack2 + 1 < description.length()) {
             sbuf.append(description, brack2 + 1, description.length() - brack2 - 1);
         }
@@ -404,7 +415,7 @@ NFRule::parseRuleDescriptor(UnicodeString& description, UErrorCode& status)
     // finally, if the rule body begins with an apostrophe, strip it off
     // (this is generally used to put whitespace at the beginning of
     // a rule's rule text)
-    if (description.length() > 0 && description.charAt(0) == gTick) {
+    if (!description.isEmpty() && description.charAt(0) == gTick) {
         description.removeBetween(0, 1);
     }
 
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
@@ -1951,4 +1951,92 @@ public class RbnfTest extends CoreTestFmwk {
             }
         }
     }
+
+    /**
+     * This test is a little contrived for English, but the grammar is relevant for several languages, including:
+     * Latin, Germanic, Slavic and Indic.
+     * It's pretty common, especially for ordinals, to use different words as a magnitude unit and when it's the final word.
+     * Several languages need grammatical agreement between the final and non-final magnitude unit
+     * with the numerical quantity before the unit. This test is the equivalent seen in other languages.
+     */
+    @Test
+    public void testOmissionReplacementWithPluralRules() {
+        final String rules = "%cardinal:\n" +
+                "-x: minus >>;\n" +
+                "x.x: << point >>;\n" +
+                "Inf: infinite;\n" +
+                "NaN: not a number;\n" +
+                "zero; one; two; three; four; five; six; seven; eight; nine;\n" +
+                "ten; eleven; twelve; thirteen; fourteen; fifteen; sixteen; seventeen; eighteen; nineteen;\n" +
+                "20: twenty[->>];\n" +
+                "30: thirty[->>];\n" +
+                "40: forty[->>];\n" +
+                "50: fifty[->>];\n" +
+                "60: sixty[->>];\n" +
+                "70: seventy[->>];\n" +
+                "80: eighty[->>];\n" +
+                "90: ninety[->>];\n" +
+                "100: << hundred[ >>];\n" +
+                "1000: << thousand[ >>];\n" +
+                "1000000: << million[ >>];\n" +
+                "1000000000: << billion[ >>];\n" +
+                "1000000000000: << trillion[ >>];\n" +
+                "1000000000000000: =#,##0=;\n" +
+                "%ordinal:\n" +
+                "-x: minus >>;\n" +
+                "x.x: =#,##0.#=;\n" +
+                "Inf: infinitieth;\n" +
+                "zeroth; first; second; third; fourth; fifth; sixth; seventh; eighth; ninth;\n" +
+                "tenth; eleventh; twelfth;\n" +
+                "13: =%cardinal=th;\n" +
+                "20: twent[y->>|ieth];\n" +
+                "30: thirt[y->>|ieth];\n" +
+                "40: fort[y->>|ieth];\n" +
+                "50: fift[y->>|ieth];\n" +
+                "60: sixt[y->>|ieth];\n" +
+                "70: sevent[y->>|ieth];\n" +
+                "80: eight[y->>|ieth];\n" +
+                "90: ninet[y->>|ieth];\n" +
+                "100: <%cardinal< [$(cardinal,one{hundred}other{hundreds})$ >>|$(cardinal,one{hundredth}other{hundredths})$];\n" +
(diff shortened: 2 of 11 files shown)


icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1951,4 +1951,89 @@ public void TestInfiniteRecursion() {
...
+                "50: fift[y->>|ieth];\n" +
+                "60: sixt[y->>|ieth];\n" +
+                "70: sevent[y->>|ieth];\n" +
+                "80: eight[y->>|ieth];\n" +
+                "90: ninet[y->>|ieth];\n" +
+                "100: <%cardinal< [$(cardinal,one{hundred}other{hundreds})$ >>|$(cardinal,one{hundredth}other{hundredths})$];\n" +
+                "1000: <%cardinal< [$(cardinal,one{thousand}other{thousands})$ >>|$(cardinal,one{thousandth}other{thousandths})$];\n" +
+                "1000000: <%cardinal< [$(cardinal,one{million}other{millions})$ >>|$(cardinal,one{millionth}other{millionths})$];\n" +
+                "1000000000: <%cardinal< [$(cardinal,one{billion}other{billions})$ >>|$(cardinal,one{billionth}other{billionths})$];\n" +
+                "1000000000000: <%cardinal< [$(cardinal,one{trillion}other{trillions})$ >>|$(cardinal,one{trillionth}other{trillionths})$];\n" +
```
grhoten, line 2001: This style of large ordinals is currently impossible to support when you have more than 2 cardinal states to consider.  It's excessively tedious to repeatedly split and copy the rules.  It's worse for languages that are highly inflectional with many grammatical cases.

Languages that use private rules like English's `%%tieth` and `%%th` ordinals rules are candidates to use this syntax. Those private rules only define 0 and 1.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
```diff
@@ -1902,40 +1870,34 @@ private StringBuilder stripWhitespace(String description) {
...
             {
                 ++start;
             }
 
-            //if the first non-whitespace character is semicolon, skip it and continue
-            if (start < descriptionLength && description.charAt(start) == ';') {
-                start += 1;
-                continue;
-            }
-
```
grhoten, line 1918: Some of this logic in stripWhitespace was out of sync between C++ and Java. Both sides have been brought in sync in this pull request.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1951,4 +1951,89 @@ public void TestInfiniteRecursion() {
...
+                "1000000000000000: =#,##0=$(ordinal,one{st}two{nd}few{rd}other{th})$;";
```
richgillam, line 2002: I'm a little unclear on the `$` syntax.  It looks like you're just using normal plural rules.  Do the English plural rules define `few` as just 3, or you doing that somewhere else in here just for the purposes of the test?

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
```diff
@@ -1951,4 +1951,89 @@ public void TestInfiniteRecursion() {
...
+                {"29", "twenty-ninth"},
+                {"30", "thirtieth"},
+                {"31", "thirty-first"},
+                {"39", "thirty-ninth"},
+                {"100", "one hundredth"},
+                {"101", "one hundred first"},
+                {"200", "two hundredths"},
+                {"201", "two hundreds first"},
+                {"300", "three hundredths"},
+                {"301", "three hundreds first"},
```
richgillam, line 2017: Obviously, this looks weird in English, but I presume you've got exactly this pattern in other languages?

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
```diff
@@ -416,106 +401,91 @@
...
  *   <tr>
- *     <td style="width: 37;"></td>
- *     <td style="width: 23;"></td>
- *     <td style="width: 165; vertical-align: top;">in improper-fraction rule</td>
- *     <td>Omit the optional text if the number is between 0 and 1 (same as specifying both an
- *     x.x rule and a 0.x rule)</td>
+ *     <td style="white-space: nowrap; vertical-align: top; padding-left: 1em; padding-right: 1em;">in improper-fraction rule</td>
+ *     <td>Omit the optional text and rules if the number is between 0 and 1 (same as specifying both an
+ *     x.x rule and a 0.x rule).
+ *     If the | symbol is present, the text and rules after the symbol and before the right bracket will be used instead of omission.</td>
```
richgillam, line 434: This sentence (in all the places it appears) sounds a little clunky to me.  Maybe something like "Omit the optional text and rules if (blah blah blah).  If the | symbol is present, use the text before the | symbol if (blah blah blah) and use the text after the | symbol if (NOT blah blah blah)".  It's longer and more work (you can't just copy-paste), but it'd be clearer and read a little better.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
```diff
@@ -1350,13 +1319,12 @@ public Number parse(String text, ParsePosition parsePosition) {
             workingPos.setIndex(0);
         }
 
-        // add the high water mark to our original parse position and
+        // add the high watermark to our original parse position and
```
richgillam, line 1322: I think this should actually be "high-water mark"...

icu4c/source/i18n/nfrule.cpp
```diff
@@ -196,6 +198,9 @@ NFRule::makeRules(UnicodeString& description,
             // rule2's rule text omits the stuff in brackets: initialize
             // its rule text and substitutions accordingly
```
richgillam, line 199: Maybe update the comment?

richgillam (dismissed):
I think this looks great.  A couple things you included as comments on this PR probably ought to be comments in the actual tests or code, and I had a couple small suggestions on your documentation, but the code looks great.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
grhoten: The code editor that I was using does spelling and grammar checking. It highlighted this along with the other oddities, like inro, and ned.

It seems like a valid English word to me. https://en.wiktionary.org/wiki/watermark

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
grhoten: Sure I’ll contemplate better wording.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
grhoten: Yes several languages have this grammar style for large ordinals, and they’re currently unable to support them without huge amounts of copied and tweaked rules. Yes, I agree that this is contrived for English, but it’s readable. It also doesn’t require extensive research on the correct spelling of ordinals in an affected language.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
grhoten: The other languages need to use the cardinal form for large ordinals. English doesn’t define few. I just wanted to verify that the syntax is usable, and not add additional plural rule tests. I think that these new tests demonstrate that.

I could try to pick a difficult language to test, like Lithuanian. I didn’t have patience to write such rules, and I wanted something readable, reviewable, and quick to write.

Do you really want the masculine singular nominative Lithuanian ordinals written in a test? I was hoping to defer such work and use the Number Format Tester instead of hard coded in tests. As an alternative, I can give you just the thousands line for documentation purposes. That’s simpler.

grhoten: I think a small sample of Lithianian thousands of the positive degree nominative masculine non-pronomial ordinal would be the following:

```
1000: [tūkstantis >>|tūkstantas];
2000: <%spellout-cardinal-masculine< [$(cardinal,one{tūkstantis}few{tūkstančiai}other{tūkstančių})$ >>|$(cardinal,one{tūkstantas}few{tūkstanti}other{tūkstantų})$];
```

Of course, this sample is unvetted, but the structure is what is needed. I got the spellings from Wiktionary. See [tūkstantis](https://en.wiktionary.org/wiki/t%C5%ABkstantis), [tūkstantas](https://en.wiktionary.org/wiki/t%C5%ABkstantas), and [pirmas](https://en.wiktionary.org/wiki/pirmas#Lithuanian) for the inflection tables. This pull request is needed to support ordinals larger than 9,999 in Lithuanian. To fully support Lithuanian ordinals, I'd likely need to copy the structure of these 2 rules less than 162 times for only the ordinals. The current Lithianian ordinal rules are a little clunky.

The structure is pretty close to the English example in the test.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
grhoten: Actually the original was right. I'll revert this change.

https://en.wikipedia.org/wiki/High_water_mark

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
richgillam: What you had is okay with me; I was just trying to understand it.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
richgillam: Fine with me; just trying to understand what you're doing.

icu4c/source/i18n/nfrule.cpp
grhoten: This has been updated.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedNumberFormat.java
grhoten: Please check the new wording.  I had to shuffle the order around and use bullet points so that it's easier to read.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/RbnfTest.java
grhoten: I have added a couple of sentences before these 2 tests to clarify to describe what's going on. Please check it out.

richgillam (approved):
I took another look.  I like your updates to the documentation.  No notes.

PR #3327, opened 2025-01-07
ICU-22991 Reduce Calendar object size

FrankYFTang:
Gregorian monthOfYear, dayOfMonth only need one byte dayOfYear only need two bytes to store.
These fields are used to store calculated value of day of year, day of month and month of year according to Gregorian calendar, and therefore will max to 365, 31 and 12 so it is a waste to store them with 4 bytes. We only need 9 bits, 5 bits, 4 bits for them. We can use bit fields to store but still need 18 bits which is more than 2 bytes. In this PR, I change to use 2 + 1 + 1 = 4 bytes to store them. Save 8 bytes per Calendar object  

#### Checklist
- [X] Required: Issue filed: ICU-22991

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -1524,14 +1524,13 @@ void Calendar::computeGregorianFields(int32_t julianDay, UErrorCode& ec) {
     if (U_FAILURE(ec)) {
         return;
     }
-    int32_t gregorianDayOfWeekUnused;
     if (uprv_add32_overflow(
             julianDay, -kEpochStartAsJulianDay, &julianDay)) {
         ec = U_ILLEGAL_ARGUMENT_ERROR;
         return;
     }
     Grego::dayToFields(julianDay, fGregorianYear, fGregorianMonth,
-                       fGregorianDayOfMonth, gregorianDayOfWeekUnused,
+                       fGregorianDayOfMonth,
                        fGregorianDayOfYear, ec);
 }
 
--- a/icu4c/source/i18n/basictz.cpp
+++ b/icu4c/source/i18n/basictz.cpp
@@ -160,12 +160,13 @@ BasicTimeZone::getSimpleRulesNear(UDate date, InitialTimeZoneRule*& initial,
               || (tr.getFrom()->getDSTSavings() != 0 && tr.getTo()->getDSTSavings() == 0))
             && (date + MILLIS_PER_YEAR > nextTransitionTime)) {
  
-            int32_t year, month, dom, dow, doy, mid;
+            int32_t year, mid;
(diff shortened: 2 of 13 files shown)


icu4c/source/i18n/calendar.cpp
```diff
@@ -1530,9 +1530,16 @@ void Calendar::computeGregorianFields(int32_t julianDay, UErrorCode& ec) {
...
-    Grego::dayToFields(julianDay, fGregorianYear, fGregorianMonth,
-                       fGregorianDayOfMonth, gregorianDayOfWeekUnused,
-                       fGregorianDayOfYear, ec);
+    int32_t gregorianMonth, gregorianDayOfMonth, gregorianDayOfYear;
+    Grego::dayToFields(julianDay, fGregorianYear, gregorianMonth,
+                       gregorianDayOfMonth, gregorianDayOfWeekUnused,
+                       gregorianDayOfYear, ec);
+    U_ASSERT(gregorianMonth <= 12);
+    U_ASSERT(gregorianDayOfMonth <= 31);
+    U_ASSERT(gregorianDayOfYear <= 366);
```
markusicu, line 1539: optional: Does Grego::dayToFields() not guarantee these? If the computations there don't necessarily guarantee these, then consider moving these asserts there.

icu4c/source/i18n/calendar.cpp
```diff
@@ -1530,9 +1530,16 @@ void Calendar::computeGregorianFields(int32_t julianDay, UErrorCode& ec) {
...
+    fGregorianMonth = gregorianMonth;
+    fGregorianDayOfMonth = gregorianDayOfMonth;
+    fGregorianDayOfYear = gregorianDayOfYear;
```
markusicu, line 1542: optional: Would it make sense to change Grego::dayToFields() to write to narrower values, so that you need not change the call site?

markusicu (commented):
optional: Java?

richgillam (approved):
Looks fine to me.  I second both of Markus's comments, but have none of my own to add.

FrankYFTang:
PTAL

markusicu (approved):
Whoa! I guess I need to be careful what I ask for... I glanced at many of the changes, and they look plausible, and it looks like you are reducing the amount of work and reducing unused variables. I am stamping.

As for Java, I suspect that it might matter less there whether we store byte/short vs. int, but it might be nice to port the other changes there if they make sense. (Don't know what the Grego functions look like there since Java does not have output reference types.)

@richgillam ok?

richgillam (approved):
Looked at it again.  LGTM.

FrankYFTang:
> optional: Java?

I will deal with that in a separate PR later.

PR #3337, opened 2025-01-20
ICU-22781 Adding support for constant denominators (C++)

younies:
# Adding support for constant denominators

- Introduced a new field for constant denominators in MeasureUnitImpl.
- Added methods to set and retrieve constant denominators in MeasureUnit.
- Updated parsing logic to handle unit constants correctly.
- Enhanced unit tests to cover new functionality, including constant denominator scenarios.
- Improved code readability and consistency across MeasureUnit and SingleUnitImpl classes.

# TODO in the following PRs
1. Add support to unit constant during units conversion.
2. Add support to unit constant during units formatting.

#### Checklist

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -15,6 +15,7 @@
 #include "charstr.h"
 #include "cmemory.h"
 #include "cstring.h"
+#include "double-conversion-string-to-double.h"
 #include "measunit_impl.h"
 #include "resource.h"
 #include "uarrsort.h"
@@ -30,12 +31,14 @@
 #include "unicode/ustringtrie.h"
 #include "uresimp.h"
 #include "util.h"
+#include <climits>
 #include <cstdlib>
-
 U_NAMESPACE_BEGIN
 
 
 namespace {
+
+using icu::double_conversion::StringToDoubleConverter;
 
 // TODO: Propose a new error code for this?
 constexpr UErrorCode kUnitIdentifierSyntaxError = U_ILLEGAL_ARGUMENT_ERROR;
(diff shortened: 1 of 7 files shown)


icu4c/source/i18n/number_fluent.cpp
```diff
@@ -654,9 +654,6 @@ void LocalizedNumberFormatter::formatImpl(impl::UFormattedNumberData* results, U
     } else {
         NumberFormatterImpl::formatStatic(fMacros, results, status);
     }
-    if (U_FAILURE(status)) {
-        return;
-    }
     results->getStringRef().writeTerminator(status);
```
younies, line 657: revert

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -530,8 +548,116 @@ class Token {
         return fMatch - kSimpleUnitOffset;
     }
 
+    // TODO: Consider split function as a utility function.
+    // Parse the given string to a unsigned long value.
+    // If the value is not positive integer, it will return `kUnitIdentifierSyntaxError`.
+    uint64_t parseStrigToLong(const StringPiece str, UErrorCode &status) {
```
richgillam, line 554: Should this be `parseStringToLong()`?  (Missing `n`)

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -530,8 +548,116 @@ class Token {
...
+        }
+
+        // Check if the result is within the valid range (1 to max_int64)
+        if ( result < 1 || result > max_int64) {
+            status = kUnitIdentifierSyntaxError;
+            return result; // add it for code consistency.
+        }
+
+        return result;
+    }
```
richgillam, line 655: Two comments about this whole function:

1) Since this is guaranteed to be a decimal number using ASCII characters in a `char` string, couldn't you just use `atoi()` instead of writing the whole thing yourself?  Afterward, you could just check the result from `atoi()` to decide whether it's valid (e.g., disallowing negative numbers).

2) The logic to parse the mantissa and the logic to parse the exponent are basically the same (except maybe for which values are valid).  It seems like you could use the same code to handle each piece.  In fact, if `atoi()` doesn't support scientific notation, you could at least use it to separately parse the mantissa and exponent and then just have your own logic for handling the `e` and putting the pieces together (or use `atof()`, although that'll give you back a `float`, which probably isn't helpful).

3) Come to think of it, I don't think the logic above works right with EBCDIC, although I don't know if we still care about that...

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -530,8 +548,116 @@ class Token {
...
+            status = kUnitIdentifierSyntaxError;
+            return result;
+        }
+
+        // Handle exponent
+        for (int i = exponentIndex; i < str.length(); ++i) {
+            char c = str.data()[i];
+
+            // handle sign
+            if (i == exponentIndex && c == '+') {
```
richgillam, line 609: Is `+` actually a legal character in a unit identifier?

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -658,12 +849,33 @@ class Parser {
...
             fIndex = previ;
+            return {match};
         }
-        return {match};
+
+        // the index of the character after the last character of the constant denominator.
+        int32_t endOfConstantIndex = -1;
+        // If no match was found, we check if the token is a constant denominator.
+        // 1. find the first `-` from the `currentFIndex` to the end.
+        for (int32_t i = currentFIndex; i < fSource.length(); ++i) {
```
richgillam, line 861: Can't you use `sdrchr()` to do this?

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -695,27 +907,30 @@ class Parser {
         bool atStart = fIndex == 0;
         Token token = nextToken(status);
         if (U_FAILURE(status)) {
-            return result;
+            return SingleUnitOrConstant::singleUnitValue(singleUnitResult);
```
richgillam, line 910: There are a lot of repetitions of this line in this function, and it's not immediately obvious that what you're doing here is just returning a default-constructed `SingleUnitImpl` when you get a parse error.  It's both really verbose and not clear.

Actually, as I look at this more, it's _not_ always a default-constructed `SingleUnitImpl`-- it's the actual result you're constructing, as far as you got before getting to the error.  Is that really what you want to be doing?

I can't help thinking you might be better off creating a default-constructed `SingleUnitOrConstant` (or a `SingleUnitOrConstant` containing a default-constructed `SingleUnitImpl`) up front, calling it `errorResult`, and just returning `errorResult` everywhere you're currently returning `SingleUnitOrConstant(singleUnitResult)` to indicate a parse error.

Or maybe I'm just horribly misreading this function...

richgillam (commented):
I think this generally looks okay, but I have a few questions.

This also doesn't actually _do_ anything, but I see that you're planning to do that part in separate PRs, so that's okay for now.

icu4c/source/i18n/measunit_extra.cpp
younies: I have used StringToDoubleConverter for now to mimic the same behavior in Java. However, we need to implement our own logic in a separate class to return an error when there is a limit exceed instead of returning any random number.

I am trying to implement the long converter here: https://github.com/unicode-org/icu/pull/3339

icu4c/source/i18n/measunit_extra.cpp
younies: Only if it is part of the unit constant:

For example, "meter-per-+10".

However, it looks a bit weird, and when the user returns the identifier, it will be removed.

This is now handled by the `StringToDoubleConverter`.

icu4c/source/i18n/measunit_extra.cpp
younies: sure, I found that we can use `StringPiece::find` instead.

icu4c/source/i18n/measunit_extra.cpp
younies: The idea is that I have changed the return type from `SingleUnitImpl` to a new type called `SingleUnitOrConstant`.

Therefore, if we need to return a result, it needs to be packed in a `SingleUnitOrConstant` object. After each check for the `status`, if there is a failure, I need to encapsulate the result (which is `SingleUnitImpl`) in a `SingleUnitOrConstant` object.

younies: Okay, you are right, I have found a way to get rid of them by returning `{}`

richgillam: I'm not questioning what you were actually doing; I was just saying it wound up being verbose and hard to read and was suggesting you try to find a more concise way of expressing it.  I haven't looked at the changes yet, but on the face of it, `{}` sounds good to me.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -483,7 +486,7 @@ class Token {
 
   static Token constantToken(StringPiece str, UErrorCode &status) {
       Token result;
-      auto value = result.parseStrigToLong(str, status);
+      auto value = Token::parseStrigToLong(str, status);
```
richgillam, line 489: You still have `parseStrigToLong()` instead of `parseStriNgToLong()`.

richgillam (dismissed):
This looks a lot better!  Thank you for the changes.

icu4c/source/i18n/measunit_extra.cpp
younies: done, thanks for the note :)

richgillam (approved):
Thank you!

PR #3348, opened 2025-01-25
ICU-23006 Fix Chinese Calendar getActualMaximize

FrankYFTang:
Read the IS_LEAP_MONTH field correctly while calling getActualMaximize.

Chinese Calendar need to read additional IS_LEAP_MONTH field while calculate the start and length of month, make the value read from the out most calls and passed into function parameter. 

#### Checklist
- [X] Required: Issue filed: ICU-23006

### DIFF
--- a/icu4c/source/i18n/chnsecal.cpp
+++ b/icu4c/source/i18n/chnsecal.cpp
@@ -253,11 +253,16 @@ int32_t ChineseCalendar::handleGetExtendedYear(UErrorCode& status) {
  * @stable ICU 2.8
  */
 int32_t ChineseCalendar::handleGetMonthLength(int32_t extendedYear, int32_t month, UErrorCode& status) const {
+    bool isLeapMonth = internalGet(UCAL_IS_LEAP_MONTH) == 1;
+    return handleGetMonthLengthWithLeap(extendedYear, month, isLeapMonth, status);
+}
+
+int32_t ChineseCalendar::handleGetMonthLengthWithLeap(int32_t extendedYear, int32_t month, bool leap, UErrorCode& status) const {
     const Setting setting = getSetting(status);
     if (U_FAILURE(status)) {
         return 0;
     }
-    int32_t thisStart = handleComputeMonthStart(extendedYear, month, true, status);
+    int32_t thisStart = handleComputeMonthStartWithLeap(extendedYear, month, leap, status);
     if (U_FAILURE(status)) {
         return 0;
     }
@@ -333,6 +338,14 @@ struct MonthInfo computeMonthInfo(
  * @stable ICU 2.8
  */
 int64_t ChineseCalendar::handleComputeMonthStart(int32_t eyear, int32_t month, UBool useMonth, UErrorCode& status) const {
+    bool isLeapMonth = false;
+    if (useMonth) {
+        isLeapMonth = internalGet(UCAL_IS_LEAP_MONTH) != 0;
+    }
+    return handleComputeMonthStartWithLeap(eyear, month, isLeapMonth, status);
+}
+
+int64_t ChineseCalendar::handleComputeMonthStartWithLeap(int32_t eyear, int32_t month, bool isLeapMonth, UErrorCode& status) const {
     if (U_FAILURE(status)) {
        return 0;
     }
@@ -361,12 +374,6 @@ int64_t ChineseCalendar::handleComputeMonthStart(int32_t eyear, int32_t month, U
     int32_t newMoon = newMoonNear(setting.zoneAstroCalc, theNewYear + month * 29, true, status);
     if (U_FAILURE(status)) {
        return 0;
-    }
-
-    // Ignore IS_LEAP_MONTH field if useMonth is false
-    bool isLeapMonth = false;
-    if (useMonth) {
-        isLeapMonth = internalGet(UCAL_IS_LEAP_MONTH) != 0;
     }
 
     int32_t newMonthYear = Grego::dayToYear(newMoon, status);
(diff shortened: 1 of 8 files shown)


icu4c/source/i18n/chnsecal.cpp
```diff
@@ -253,11 +253,16 @@ int32_t ChineseCalendar::handleGetExtendedYear(UErrorCode& status) {
  * @stable ICU 2.8
  */
 int32_t ChineseCalendar::handleGetMonthLength(int32_t extendedYear, int32_t month, UErrorCode& status) const {
+    bool isLeapMonth = internalGet(UCAL_IS_LEAP_MONTH) == 1;
+    return handleGetMonthLengthWithLeap(extendedYear, month, isLeapMonth, status);
+}
+
+int32_t ChineseCalendar::handleGetMonthLengthWithLeap(int32_t extendedYear, int32_t month, bool leap, UErrorCode& status) const {
```
richgillam, line 260: Is it really worth it to split `handleGetMonthLength()` into two functions like this?  Wouldn't it be easier to just make the change inside `handleGetMonthLength()`?

icu4c/source/i18n/chnsecal.cpp
```diff
@@ -333,6 +338,14 @@ struct MonthInfo computeMonthInfo(
  * @stable ICU 2.8
  */
 int64_t ChineseCalendar::handleComputeMonthStart(int32_t eyear, int32_t month, UBool useMonth, UErrorCode& status) const {
```
richgillam, line 340: Same question here: Is it really worth it to split `handleComputeMonthStart()` into two functions like this?

icu4c/source/i18n/chnsecal.cpp
```diff
@@ -1184,6 +1191,27 @@ ChineseCalendar::Setting ChineseCalendar::getSetting(UErrorCode&) const {
...
+        LocalPointer<ChineseCalendar> cal(clone(), status);
+        if(U_FAILURE(status)) {
+            return 0;
+        }
+        cal->setLenient(true);
+        cal->prepareGetActual(field,false,status);
+        int32_t year = cal->get(UCAL_EXTENDED_YEAR, status);
+        int32_t month = cal->get(UCAL_MONTH, status);
+        bool leap = cal->get(UCAL_IS_LEAP_MONTH, status) != 0;
+        return handleGetMonthLengthWithLeap(year, month, leap, status);
```
richgillam, line 1210: I'm guessing this is a partial answer to my question is here, but if you just left `handleGetMonthLength()` as one function and updated it to handle the leap year, couldn't you just call it here (it'd then be getting the `UCAL_IS_LEAP_MONTH` field instead of getting it here).

icu4c/source/test/intltest/caltest.cpp
```diff
@@ -5354,6 +5355,18 @@ void CalendarTest::TestLimitsOrdinalMonth() {
...
+    GregorianCalendar gc(status);
+    gc.set(2025, UCAL_AUGUST, 8);
+    LocalPointer<Calendar> cal(Calendar::createInstance(Locale("en@calendar=chinese"), status), status);
+    cal->setTime(gc.getTime(status), status);
+    int32_t beforeCallingGet = cal->getActualMaximum(UCAL_DAY_OF_MONTH, status);
+    cal->get(UCAL_DAY_OF_MONTH, status);
+    int32_t afterCallingGet = cal->getActualMaximum(UCAL_DAY_OF_MONTH, status);
+    assertEquals("getActualMaximum() should return same value before/after calling get()",
+                 beforeCallingGet, afterCallingGet);
+}
```
richgillam, line 5369: Is `getActualMaximum()` the only public function on this calendar that's changed its behavior?  This test seem to test the leap-month case, as well as checking that you get the right thing whether or not you've done a `get()` on the `UCAL_DAY_OF_MONTH` field, but do you have existing test to verify that this function still works correctly for non-leap months?

I'm guessing there's another `getMaximum()` method that returns the absolute maximum for the dat-of-month field, that it was already returning the right thing, and that there are already tests for it?

icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
```diff
@@ -450,13 +450,14 @@ protected int handleGetExtendedYear() {
     /**
      * Override Calendar method to return the number of days in the given
      * extended year and month.
-     *
-     * <p>Note: This method also reads the IS_LEAP_MONTH field to determine
-     * whether or not the given month is a leap month.
-     * @stable ICU 2.8
```
richgillam, line 456: Isn't this comment still true even with your change?

richgillam (approved):
Looks fine, I think, and I think I'm okay with you checking in as-is, so I went ahead and approved, but I do have some questions you might want to consider.  Most of those questions apply to both the C++ and Java versions of your change, but I didn't repeat them on the Java code.

icu4c/source/i18n/chnsecal.cpp
FrankYFTang: The problem is one usage need to use get (to force the calculation of all fields) and the other need to use internalGet (to avoid infinity recursion) and I cannot change the function signature (because it is defined in the base class as a protected method for all subclass of Calendar). The tricky part of this fix is it is very easy to get into infinity recursion

richgillam: Okay.  Thanks for the explanation.

markusicu:
rslgtm, leaving it to you two -- tnx!

icu4c/source/test/intltest/caltest.cpp
FrankYFTang: I added more test cases

icu4c/source/i18n/chnsecal.cpp
FrankYFTang: It is necessary because we cannot change the signature of handleComputeMonthStart since it is defined by the superclass but we need to pass in the leap information separately to avoid infinity recusion

FrankYFTang:
PTAL

richgillam (approved):
Thanks for your answers to my questions.  I'm convinced.

markusicu:
tests pass -- ready to merge?

PR #3374, opened 2025-02-06
ICU-22954 Make UCollator predicates usable without U_SHOW_CPLUSPLUS_API

roubert:
Until now, the implementation of the `UCollator` predicates has been using `UnicodeString` and `StringPiece` as convenient wrappers for converting from standard C++ data types to ICU4C data types.

But as that doesn't work when the client uses ICU4C built without `U_SHOW_CPLUSPLUS_API` this is now changed to instead perform these conversions directly.

(It's a bit more code, but does just the same thing in the end.)

Then add a build test for all header files that use `U_SHOW_CPLUSPLUS_HEADER_API` to verify that they now all can be built without `U_SHOW_CPLUSPLUS_API`.

With this all done, ICU-22954 should be resolved.

#### Checklist
- [x] Required: Issue filed: ICU-22954

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/test/intltest/cplusplus_header_api_build_test.cpp
+++ b/icu4c/source/test/intltest/cplusplus_header_api_build_test.cpp
@@ -1,0 +1,11 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html#License
+
+// ICU-22954 Test that client code can be built with U_SHOW_CPLUSPLUS_API=0.
+#undef U_ALL_IMPLEMENTATION
+#undef U_SHOW_CPLUSPLUS_API
+#define U_SHOW_CPLUSPLUS_API 0
+#include "unicode/char16ptr.h"
+#include "unicode/ucol.h"
+#include "unicode/uset.h"
+#include "unicode/utypes.h"
--- a/icu4c/source/test/intltest/ucolheaderonlytest.cpp
+++ b/icu4c/source/test/intltest/ucolheaderonlytest.cpp
@@ -1,0 +1,96 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
+// Test header-only ICU C++ APIs. Do not use other ICU C++ APIs.
+// Non-default configuration:
+#define U_SHOW_CPLUSPLUS_API 0
+
+#include "unicode/utypes.h"
+
+#if !UCONFIG_NO_COLLATION
+
+#include <string_view>
+
+#include "unicode/ucol.h"
+
+#include "intltest.h"
+
+namespace {
+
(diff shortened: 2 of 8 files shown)


icu4c/source/test/intltest/cplusplus_header_api_build_test.cpp
```diff
@@ -0,0 +1,11 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html#License
+
+// ICU-22954 Test that client code can be built with U_SHOW_CPLUSPLUS_API=0.
+#undef U_ALL_IMPLEMENTATION
+#undef U_SHOW_CPLUSPLUS_API
+#define U_SHOW_CPLUSPLUS_API 0
+#include "unicode/char16ptr.h"
+#include "unicode/ucol.h"
```
markusicu, line 9: This proves that ucol.h does not cause compiler errors when included in this configuration, but it doesn't prove that these predicates work like this. This file here would be just as happy if ucol.h looked empty...

Why not move the relevant unit tests into a file like https://github.com/unicode-org/icu/blob/main/icu4c/source/test/intltest/usetheaderonlytest.cpp ?

----
otherwise lgtm tnx!

roubert: Done.

icu4c/source/test/intltest/ucolheaderonlytest.cpp
```diff
@@ -0,0 +1,96 @@
...
+void UColHeaderOnlyTest::runIndexedTest(int32_t index, UBool exec, const char*& name, char* /*par*/) {
+    if (exec) {
+        logln("TestSuite UColHeaderOnlyTest: ");
+    }
+    TESTCASE_AUTO_BEGIN;
+    TESTCASE_AUTO(TestPredicateTypes);
+    TESTCASE_AUTO_END;
+}
+
+class UCollatorWrapper {
```
markusicu, line 36: FYI: We can use std::unique_ptr here.

markusicu (approved):
lgtm tnx!

icu4c/source/test/intltest/ucolheaderonlytest.cpp
roubert: > FYI: We can use std::unique_ptr here.

I can't see how, but please send a follow-up PR if you have some clever idea.

markusicu: You would provide ucol_close as the deleter: https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr
Not important to change.

roubert: I still can't imagine how that would lead to simpler code, but please send a follow-up PR if you do.

PR #3378, opened 2025-02-07
ICU-22967 Remove explicit manipultaion of old-xmls.tar.bz2 in workflows

mihnita:
The workflow was trying to repair old-xmls.tar.bz2

Which was broken (the first commit in my pull request https://github.com/unicode-org/icu/pull/3375)
Although the right fix would have been to restore the file for good.
But this "fix" does not work now, on a removed file :-) (the second commit in the mentioned PR)
Anyway... the whole folder 'tools/multi' is not needed, so it should be good now.

#### Checklist
- [x] Required: Issue filed: ICU-22967

### DIFF
--- a/.github/workflows/icu_merge_ci.yml
+++ b/.github/workflows/icu_merge_ci.yml
@@ -349,11 +349,6 @@ jobs:
 
       - name: Build and run unicodesetperf test
         run: |
-          # This file needs to be restored  otherwise GHA cannot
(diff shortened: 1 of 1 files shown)


roubert (approved):
These changes look perfectly reasonable to me and I trust that you've verified that they work.

markusicu (approved):

mihnita:
Thank you!

PR #3383, opened 2025-02-10
ICU-23032 Optimize complexity caching of MeasureUnit (Java)

younies:
# Description: 
Optimize MeasureUnit complexity retrieval by:
- Adding complexity and isComplexitySet fields
- Caching complexity value after first retrieval
- Reducing redundant complexity calculations

#### Checklist
- [x] Required: Issue filed: ICU-23032


younies:
The problem, that this approach will break the immutability !

icu4j/main/core/src/main/java/com/ibm/icu/util/MeasureUnit.java
```diff
@@ -496,8 +502,14 @@ public String getIdentifier() {
      * @stable ICU 68
      */
     public Complexity getComplexity() {
+        if (isComplexitySet) {
+            return complexity;
+        }
```
macchiati, line 507: This is not thread-safe. Please read https://en.wikipedia.org/wiki/Double-checked_locking

macchiati (changes requested):

PR #3385, opened 2025-02-10
ICU-22781 Uncomment and enable constant denominator tests

younies:
# Description: 

Re-enable previously commented out tests for constant denominators in units tests for both ICU4C and ICU4J. This includes tests for:
- Large power of 10 denominators
- Scientific notation denominators
- Specific unit denominators

Added notes about potential CLDR-related edge cases for certain unit representations.

#### Checklist

### DIFF
--- a/icu4c/source/test/intltest/units_test.cpp
+++ b/icu4c/source/test/intltest/units_test.cpp
@@ -1192,6 +1192,7 @@ void UnitsTest::testUnitsConstantsDenomenator() {
     } testCases[]{
         {"meter-per-1000", 1000},
         {"liter-per-1000-kiloliter", 1000},
+        {"meter-per-100-kilometer", 100}, // Failing: ICU-23045
         {"liter-per-kilometer", 0},
         {"second-per-1000-minute", 1000},
         {"gram-per-1000-kilogram", 1000},
@@ -1205,6 +1206,7 @@ void UnitsTest::testUnitsConstantsDenomenator() {
         {"portion-per-7", 7},
         {"portion-per-8", 8},
         {"portion-per-9", 9},
+
         // Test for constant denominators that are powers of 10
         {"portion-per-10", 10},
         {"portion-per-100", 100},
@@ -1214,16 +1216,18 @@ void UnitsTest::testUnitsConstantsDenomenator() {
         {"portion-per-1000000", 1000000},
         {"portion-per-10000000", 10000000},
         {"portion-per-100000000", 100000000},
-        // ICU-22781: {"portion-per-1000000000", 1000000000},
-        // ICU-22781: {"portion-per-10000000000", 10000000000},
-        // ICU-22781: {"portion-per-100000000000", 100000000000},
-        // ICU-22781: {"portion-per-1000000000000", 1000000000000},
-        // ICU-22781: {"portion-per-10000000000000", 10000000000000},
-        // ICU-22781: {"portion-per-100000000000000", 100000000000000},
-        // ICU-22781: {"portion-per-1000000000000000", 1000000000000000},
-        // ICU-22781: {"portion-per-10000000000000000", 10000000000000000},
-        // ICU-22781: {"portion-per-100000000000000000", 100000000000000000},
-        // ICU-22781: {"portion-per-1000000000000000000", 1000000000000000000},
+        {"portion-per-1000000000", 1000000000}, // Failing: ICU-23045
+        {"portion-per-10000000000", 10000000000},
+        {"portion-per-100000000000", 100000000000},
+        {"portion-per-1000000000000", 1000000000000},
+        {"portion-per-10000000000000", 10000000000000},
+        {"portion-per-100000000000000", 100000000000000},
+        {"portion-per-1000000000000000", 1000000000000000},
+        {"portion-per-10000000000000000", 10000000000000000},
+        {"portion-per-100000000000000000", 100000000000000000},
+        {"portion-per-1000000000000000000", 1000000000000000000},
+        {"portion-per-1e3-kilometer", 1000},
+
         // Test for constant denominators that are represented as scientific notation
         // numbers.
         {"portion-per-1e1", 10},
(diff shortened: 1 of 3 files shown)


icu4c/source/test/intltest/units_test.cpp
```diff
@@ -1238,23 +1240,35 @@ void UnitsTest::testUnitsConstantsDenomenator() {
...
         {"meter", 0},
         {"meter-per-second", 0},
         {"meter-per-square-second", 0},
-        // NOTE: The following constant denominator should be 0. However, since
+
+        // NOTE: The following constant denominator should be 1e9. However, since
+        // `portion-per-1e9` is treated as a unit in CLDR,
+        // the unit does not have a constant denominator.
+        // This issue should be addressed in CLDR & ICU.
+        {"portion-per-1000000000", 0},
```
richgillam, line 1262: Is there an outstanding ticket for this?  If so, you should probably include the ticket number (or a link) here.

richgillam (commented):
Where's the part where you actually test that formatting does he right thing with this unit?

icu4c/source/test/intltest/units_test.cpp
macchiati: `portion-per-1e9` is treated as a _precomputed form_ for translation, not a single unit.

[Some units already have 'precomputed' forms, such as **kilometer-per-hour**; where such units exist, they should be used in preference.](https://unicode.org/reports/tr35/tr35-general.html#Unit_Identifier_Uniqueness:~:text=Some%20units%20already%20have%20%27precomputed%27%20forms%2C%20such%20as%20kilometer%2Dper%2Dhour%3B%20where%20such%20units%20exist%2C%20they%20should%20be%20used%20in%20preference.)

We shouldn't test for the wrong behavior. 
That being said, given the schedule, and the fact that we are going fix the problem fully in 48, it's ok for `{"portion-per-1000000000", 1000000000},` be present but commented out.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/MeasureUnitTest.java
```diff
@@ -1344,38 +1345,59 @@ class ConstantDenominatorTestCase {
...
                 new ConstantDenominatorTestCase("meter-per-square-second", 0),
-                // NOTE: The following constant denominator should be 0. However, since
+
+                // NOTE: The following constant denominator should be 1e9. However, since
+                // `portion-per-1e9` is treated as a unit in CLDR,
+                // the unit does not have a constant denominator.
+                // This issue should be addressed in CLDR & ICU.
+                new ConstantDenominatorTestCase("portion-per-1000000000", 0),
+                new ConstantDenominatorTestCase("portion-per-1e9", 0),
+                new ConstantDenominatorTestCase("portion-per-1E9", 0),
```
macchiati, line 1393: Same here.

macchiati (changes requested):

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -1205,6 +1205,7 @@ void UnitsTest::testUnitsConstantsDenomenator() {
         {"portion-per-7", 7},
         {"portion-per-8", 8},
         {"portion-per-9", 9},
```
mihnita, line 1207: These cases don't exist in the Java tests.
Is there a good reason to have them?
Are they testing something interesting, that is not well covered by other cases?
They they should also be in Java.

If they don't actually exercise anything interesting, we should not have them in C/C++ either.

In general the tests in C++ and Java should be the same, unless there are very good reasons (for example Java testing for all kind of types that don't exist in C++)

icu4c/source/test/intltest/units_test.cpp
younies: Sure, I have added it back with `logKnownIssue` so it can be addressed in the next release.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/MeasureUnitTest.java
younies: done, I added also a `logKnownIssue`

icu4c/source/test/intltest/units_test.cpp
younies: now , c++ and java are identical, sorry for this small confusion.

younies:
> Where's the part where you actually test that formatting does he right thing with this unit?

@richgillam : for testing the formatting:
https://github.com/unicode-org/icu/pull/3389
https://github.com/unicode-org/icu/pull/3390

icu4c/source/test/intltest/units_test.cpp
```diff
@@ -1214,16 +1216,18 @@ void UnitsTest::testUnitsConstantsDenomenator() {
...
+        {"portion-per-1000000000", 1000000000},
+        {"portion-per-10000000000", 10000000000},
+        {"portion-per-100000000000", 100000000000},
+        {"portion-per-1000000000000", 1000000000000},
+        {"portion-per-10000000000000", 10000000000000},
+        {"portion-per-100000000000000", 100000000000000},
+        {"portion-per-1000000000000000", 1000000000000000},
+        {"portion-per-10000000000000000", 10000000000000000},
+        {"portion-per-100000000000000000", 100000000000000000},
+        {"portion-per-1000000000000000000", 1000000000000000000},
```
richgillam, line 1228: Same issue as in the other PRs: If these are all expected to fail and are being covered by `logKnownIssue()` in the code, put some sort of comment next to the actual test case to indicate that (the ticket number you had in the earlier version of these changes works, for example).  Same thing further down in here.

An alternative approach would be to add a `knownIssueTicketNumber` property to your `TestCase` object-- the test code would look for that and, if it's populated, skip the test and pass it to `logKnownIssue`.

richgillam (commented):
Looks okay to me, but I think you should get approval from somebody who raised objections the last time around.

icu4c/source/test/intltest/units_test.cpp
richgillam: Is line 1219 the only one that's failing?  If so, we're good here.

younies:
@macchiati , can you take a look ?

macchiati (approved):

PR #3386, opened 2025-02-10
ICU-22781 Add tests for portion unit formatting (Java)

younies:
# Description:

1. Add test cases for formatting portion units, including tests for "portion-per-1e9" and other arbitrary portion units. 
2. The tests verify the formatting of portions with different scales and highlight a known CLDR data issue (CLDR-18274).

#### Checklist


icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatTest.java
```diff
@@ -7120,4 +7120,63 @@ public TestData(String unitIdentifier, Integer inputValue, UnitWidth width, ULoc
...
+        assertEquals("1 portion per 1e9", "1 part per billion", formatted1);
+
+        String formatted2 = formatter.format(2).toString();
+        assertEquals("1000 portion per 1e9", "2 parts per billion", formatted2);
+
+        String formatted3 = formatter.format(1000000).toString();
+        assertEquals("1000000 portion per 1e9", "1,000,000 parts per billion", formatted3);
+
+        String formatted4 = formatter.format(1000000000).toString();
+        assertEquals("1000000000 portion per 1e9", "1,000,000,000 parts per billion", formatted4);
```
richgillam, line 7139: These four test cases all seem to be doing the same thing.  In all four of them, the "unit" part of the string is exactly the same.  Why do you need all four, and is there _other_ testing you should be doing where the "unit" part of the string _does_ change?  (Okay, I'm wrong: _one_ of these _is_ different and the other _three_ are all the same...)

And if all you're doing here is making sure that "portion-per-1e9" comes out properly localized (in English) as "parts per billion", isn't there a more general test you could just add this too?

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatTest.java
```diff
@@ -7120,4 +7120,63 @@ public TestData(String unitIdentifier, Integer inputValue, UnitWidth width, ULoc
...
+        for (TestData testCase : testData) {
+            try {
+                MeasureUnit unit = MeasureUnit.forIdentifier(testCase.unitIdentifier);
+                LocalizedNumberFormatter formatter = NumberFormatter.withLocale(ULocale.ENGLISH).unit(unit)
+                        .unitWidth(UnitWidth.FULL_NAME);
+                formatter.format(testCase.inputValue);
+                // NOTE: All these test cases should pass, but because CLDR data is missing for
+                // `portion`, an error is thrown.
+                // See: CLDR-18274
+                fail("Expected IllegalArgumentException for unit: " + testCase.unitIdentifier
```
richgillam, line 7174: This is the thing you're waiting on Mihai for, right?  Shouldn't you be using `logKnownIssue()` for this?  And don't you already have what you need from Mihai?

Also, I assume when the localized names for "portion" are there, you'll update this to check the actual returned value, right?

richgillam (approved):
This generally looks okay, I think, but I had a couple of concerns.

Also, what's the story on the C++ side?  It seems like it'd be a lot better to have the C++ and Java stuff in the same PRs instead of in different PRs the way you've been doing it.  Easier to make sure you're doing the same thing in both places.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatTest.java
younies: From Mihai, I have only `portion-per-1e9` data, not `portion` , therefore, I can format `portion-per-1e9` but not `portion-per-XYZ` , see CLDR-18274.

Therefore, I am adding these test cases that expected a failure.

by the way, what is `logKnownIssue` ?

richgillam: >From Mihai, I have only portion-per-1e9 data, not portion , therefore, I can format portion-per-1e9 but not portion-per-XYZ , see [CLDR-18274](https://unicode-org.atlassian.net/browse/CLDR-18274).

Oh, I see.  Okay, then.

>by the way, what is logKnownIssue ?

There's a function called `logKnownIssue()` that you can call to indicate that a particular case is known to fail and that we have a ticket outstanding for fixing it.  I think that's what you want to be using here.  I don't remember the details off the top of my head, but if you search the code base for it, you'll find it being used a lot.

younies: okay, I will merge this PR and have a following PR with `logKnownIssue` function

PR #3391, opened 2025-02-11
ICU-22901 Update ulocimp_canonicalize() &co. to use std::string_view

roubert:
#### Checklist
- [x] Required: Issue filed: ICU-22901

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/uloc.cpp
+++ b/icu4c/source/common/uloc.cpp
@@ -589,7 +589,7 @@ compareKeywordStructs(const void * /*context*/, const void *left, const void *ri
 }  // namespace
 
 U_EXPORT CharString
-ulocimp_getKeywords(const char* localeID,
+ulocimp_getKeywords(std::string_view localeID,
                     char prev,
                     bool valuesToo,
                     UErrorCode& status)
@@ -606,7 +606,7 @@ ulocimp_getKeywords(const char* localeID,
 }
 
 U_EXPORT void
-ulocimp_getKeywords(const char* localeID,
+ulocimp_getKeywords(std::string_view localeID,
                     char prev,
                     ByteSink& sink,
                     bool valuesToo,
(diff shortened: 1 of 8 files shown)


icu4c/source/common/uloc.cpp
```diff
@@ -618,84 +618,81 @@ ulocimp_getKeywords(const char* localeID,
...
+            equalSign = localeID.find('=');
+            semicolon = localeID.find(';');
             /* lack of '=' [foo@currency] is illegal */
             /* ';' before '=' [foo@currency;collation=pinyin] is illegal */
-            if(!equalSign || (semicolon && semicolon<equalSign)) {
+            if (equalSign == std::string_view::npos ||
+                (semicolon != std::string_view::npos && semicolon < equalSign)) {
+                status = U_INVALID_FORMAT_ERROR;
+                return;
+            }
```
markusicu, line 648: nit: as in google code, the condition continuation should be indented more (+4 spaces on the second highlighted line here) than the if-body

icu4c/source/common/uloc.cpp
```diff
@@ -1748,29 +1748,29 @@ _canonicalize(const char* localeID,
...
+        if (localeID.size() >= 2 && localeID.find('_') != std::string_view::npos && localeID[1] != '-' && localeID[1] != '_') {
+            localeIDWithHyphens.append(localeID, err);
             if (U_SUCCESS(err)) {
                 for (char* p = localeIDWithHyphens.data(); *p != '\0'; ++p) {
                     if (*p == '_') {
                         *p = '-';
                     }
                 }
-                localeIDPtr = localeIDWithHyphens.data();
+                localeIDPtr = localeIDWithHyphens.toStringPiece();
```
markusicu, line 1768: nit: consider renaming localeIDPtr now that it's not a pointer

markusicu (approved):
lgtm except for two minor things

icu4c/source/common/uloc.cpp
roubert: One day I'll update `.clang-format` to match ICU style (in as much as such a thing actually exists), but until then the existing formatting is already inconsistent enough for this to not matter.

icu4c/source/common/uloc.cpp
roubert: Considered, but getting a smaller (and thus easier to read and verify) diff found more valuable, it's still being used like some kind of pointer (and this code is already full of poorly chosen and inconsistent variable names).

PR #2923, opened 2024-03-25
ICU-22534 BRS75 J API Signature file and API change report

yumaoka:
##### Checklist

- [x] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-22534
- [x] Required: The PR title must be prefixed with a JIRA Issue number. 
- [x] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [x] Issue accepted (done by Technical Committee after discussion)
- [ ] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

PR #3409, opened 2025-02-21
ICU-22922 BRS77 Add ICU4J API signature file

mihnita:
BRS 77 sheet, line 64, "Add ICU4J API signature file"

#### Checklist
- [x] Required: Issue filed: ICU-22922


markusicu:
Should we delete icu4j67.api3.gz to keep the number of these files the same, similar to @yumaoka 's https://github.com/unicode-org/icu/pull/3221 and https://github.com/unicode-org/icu/pull/2923 ?

It also looks like Yoshito used to clean up javadoc warnings in these PRs. Are there any?

markusicu (approved):

mihnita:
> Should we delete icu4j67.api3.gz to keep the number of these files the same, similar to @yumaoka 's #3221 and #2923 ?

I was thinking to do that in the release BRS.
But sure, no problem.

---

> It also looks like Yoshito used to clean up javadoc warnings in these PRs. Are there any?

I cleaned some javadoc warnings a couple of days ago.
https://github.com/unicode-org/icu/pull/3397

The doc for this step says nothing about javadoc.
Would be a bit weird to have that here

I was talking with Elango about this a couple of days ago though.
The javadoc got stricter in time.

There is a comment in the `icu4j/pom.xml`:

```xml
<!--
  The ant project specified <doclint>reference,html</doclint>.
  But that fails in main\core\src\main\java because maven enforces dependencies more strictly:
    [ERROR] com\ibm\icu\text\UnicodeFilter.java:15: error: reference not found
    [ERROR]  * {@link com.ibm.icu.text.Transliterator}
    [ERROR] com\ibm\icu\util\ULocale.java:100: error: reference not found
    [ERROR]  * {@link com.ibm.icu.text.Collator}
-->
<doclint>html</doclint>
```

I wanted to actually propose that:
- we make the `<doclint>` stricter, and fix ALL complaints
- we keep it strict, and enable it in CI, so the problems are detected asap

References are problematic though, when the project is modular like this.

Because you can't have core depend on charset and charset depend on core (circular dependency).
So you can't have javadoc in core refer to classes in charset.

Anyway...
Right now we don't have any javadoc errors

PR #3411, opened 2025-02-21
ICU-23053 In the ICU4X mode of the collation builder, mark middle starters in contractions

hsivonen:
This change uses a contraction ce32 for starters that occur in the middle of contractions but do not themselves start contractions. In practice, this affects two characters in the root (U+0CC2 and U+0DCF) and none in the CLDR tailorings.

This enables finding out whether a pair of starters might participate in a contraction without having to look further back, which should enable the identical prefix optimization directly from ce32s.

#### Checklist
- [x] Required: Issue filed: ICU-23053

I have tested this using the ICU4X tree (not landed yet).


sffc:
Hi, it seems that this blocks ICU4X 2.0 since it breaks data. Can the review be prioritized so that we can tag an icuexportdata release?

hsivonen:
Related: https://github.com/unicode-org/unicodetools/issues/1101

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -1391,7 +1419,71 @@ CollationDataBuilder::buildMappings(CollationData &data, UErrorCode &errorCode)
...
+        // that has the `CONTRACT_HAS_STARTER` flag set.
+
+        UCharsTrieBuilder contractionBuilder(errorCode);
+        // Intentionally unpaired low surrogate to make it never
+        // match well-formed UTF-16 which ICU4X feeds to the
+        // matcher.
+        UnicodeString placeholder(0xDC00);
+
+        UnicodeSetIterator iter(contractionMiddleStarter);
+        while(iter.next() && !iter.isString()) {
```
markusicu, line 1437: optional but nice to have:
something like this should work since ICU 76 --
```
for (UChar32 c : contractionMiddleStarter.codePoints()) {
```

see https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeSet.html#a537a13a1b3adcaebc140f715876aca63

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -1391,7 +1419,71 @@ CollationDataBuilder::buildMappings(CollationData &data, UErrorCode &errorCode)
...
+                if (fromBase) {
+                    // This case does not actually happen as of February 2025.
+                    ce32 = copyFromBaseCE32(c, ce32, true, errorCode);
+                }
+                if (Collation::hasCE32Tag(ce32, Collation::CONTRACTION_TAG)) {
+                    // This middle starter is also the first character of another
+                    // contraction, but that contraction does not have the
+                    // CONTRACT_HAS_STARTER flag. Let's add the flag to make
+                    // mark this at the expense of pessimizing the matching
+                    // of this contraction.
```
markusicu, line 1455:
```suggestion
                    // CONTRACT_HAS_STARTER flag. Let's add the flag to
                    // mark this at the expense of pessimizing the matching
                    // of this contraction.
```

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -1391,7 +1419,71 @@ CollationDataBuilder::buildMappings(CollationData &data, UErrorCode &errorCode)
...
+                    // As of February 2025, this case does not actually occur
+                    // in CLDR.
+                    ce32 |= Collation::CONTRACT_HAS_STARTER;
+                } else {
+                    // This middle starter is not also the first character
+                    // in another contraction.
+
+                    // The UCharsTrie needs to contain some placeholder. Trying to
+                    // build an empty UCharsTrie causes an error. We build a trie
```
markusicu, line 1464:
```suggestion
                    // The UCharsTrie needs to contain some placeholder
                    // because it cannot be empty. We build a trie
```

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -1391,7 +1419,71 @@ CollationDataBuilder::buildMappings(CollationData &data, UErrorCode &errorCode)
...
+                    // that never actually matches anything that ICU4X can try to
+                    // match, since ICU4X always passes well-formed UTF-16 to the
+                    // matcher and we put an unpaired low surrogate into the trie.
+                    // This pessimizes the character to CE mapping of the `c`,
+                    // since useless trie matching will be attempted but as of
+                    // February 2025, only two characters are affected.
```
markusicu, line 1470:
```suggestion
                    // February 2025, only two relatively rare characters are affected.
```

icu4c/source/i18n/collationdatabuilder.h
```diff
@@ -254,6 +254,12 @@ class U_I18N_API CollationDataBuilder : public UObject {
     int32_t contextsEra = 0;
 protected:
     UnicodeSet unsafeBackwardSet;
+    /**
+     * For ICU4X only: The starters that occur in some contraction
+     * in a position that is neither the first nor the last scalar
```
markusicu, line 259:
```suggestion
     * in a position that is neither the first nor the last code point
```

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -655,6 +655,20 @@ CollationDataBuilder::addCE32(const UnicodeString &prefix, const UnicodeString &
...
             }
+            int32_t sCount = s.countChar32();
+            UChar32 sUtf32[32];
+            int32_t sLen = s.toUTF32(sUtf32, 32, errorCode);
+            if (sLen != sCount) {
+                // If this error is ever reached, just increase the buffer
+                // size above.
+                errorCode = U_UNSUPPORTED_ERROR;
+                return;
+            }
```
markusicu, line 666: FYI: no real need to convert to a UTF-32 string; you could consider some alternatives, or keep this as is

- UnicodeString lets you iterate over its code points using char32At() and moveIndex32(index, delta=1)
- ICU 78 adds a real C++ string code point iterator ([unicode/utfiterator.h](https://github.com/unicode-org/icu/blob/main/icu4c/source/common/unicode/utfiterator.h) function `utfIterator<UChar32, UTF_BEHAVIOR_SURROGATE>(s)` makes a “range”)
  - You could use `auto limit = range.end();` then `--limit;` to skip the last code point, skip `auto start = range.begin();` via `++start` if it's not already at the limit, and then explicitly loop `for (; start != limit; ++start)` and use `start->codePoint()`

FYI 2
I also feel like the builder probably doesn't allow contractions longer than 31 UChars but can't find a constant for that so maybe that's not currently a safe/enforced assumption

markusicu:
Clever; I should try to do something like this in ICU4C/J if I ever find time to work on that again...

What do you do in identical-prefix matching with ccc!=0 characters? Additional lookup in normalization data?

icu4c/source/i18n/collationdatabuilder.cpp
```diff
@@ -1391,7 +1419,71 @@ CollationDataBuilder::buildMappings(CollationData &data, UErrorCode &errorCode)
...
+                    // matcher and we put an unpaired low surrogate into the trie.
+                    // This pessimizes the character to CE mapping of the `c`,
+                    // since useless trie matching will be attempted but as of
+                    // February 2025, only two characters are affected.
+                    contractionBuilder.clear();
+                    contractionBuilder.add(placeholder, static_cast<int32_t>(ce32), errorCode);
+
+                    int32_t index = addContextTrie(ce32, contractionBuilder, errorCode);
+                    if(U_FAILURE(errorCode)) { return; }
+                    if(index > Collation::MAX_INDEX) {
```
markusicu, line 1476: Does addContextTrie() not already check this?

markusicu (commented):
Mostly lgtm. Thanks for the comprehensive comments.
And sorry for the long delay.

icu4c/source/i18n/collationdatabuilder.cpp
hsivonen: If it is checked, it's not obvious. Pre-existing calls to the method in this file had this check right after, which is why I copied the check here.

icu4c/source/i18n/collationdatabuilder.cpp
hsivonen: I think I'll leave this as-is for now in case a need to backport to ICU4C earlier than 78 shows up.

icu4c/source/i18n/collationdatabuilder.cpp
hsivonen: Changed.

hsivonen:
I believe I have now addressed the review comments.

markusicu:
> I believe I have now addressed the review comments.

Thanks, will look. In the future, please just add commits when responding to feedback, and squash only at the very end after approval.

markusicu (approved):

markusicu:
@hsivonen are you ready for me to merge? Can I also close the ticket, or is there more to be done?

hsivonen:
> > I believe I have now addressed the review comments.
> 
> Thanks, will look. In the future, please just add commits when responding to feedback, and squash only at the very end after approval.

OK.

> @hsivonen are you ready for me to merge? Can I also close the ticket, or is there more to be done?

Yes, please merge. It's OK to close the ticket on the ICU4C side.

Thanks!

PR #3413, opened 2025-02-22
ICU-23056 Add workflow to generate commit checker report

echeran:
Adds a CI workflow that is manually invoked that automates the instructions for the BRS task to generate the commit checker report.

Details:
* Output: After the workflow completes, the summary page for that workflow instance will render the markdown of the report at the bottom, like so:
https://github.com/echeran/icu/actions/runs/13466516414
* Authentication: When the commit range includes a sensitive ticket, the tool will fail unless the user provides the optional  needed for this purpose. Otherwise, the credentials are optional.
* Privacy: The workflow avoids needing to print the user email & API token. In addition, the workflow will mask the user email & API token in the future hypothetical scenario that they are needed in some way that gets captured by the log. (This is a feature after invoking the `add-mask` directive in Github Actions). 

Note: The instructions for running the commit checker tool locally are in the BRS tasks section of the User Guide here: https://unicode-org.github.io/icu/processes/release/tasks/miscellaneous.html#check-integrity-of-jira-issues-in-commit-messages

#### Checklist
- [X] Required: Issue filed: ICU-23056

### DIFF
--- a/.github/workflows/brs-commit-checker.yml
+++ b/.github/workflows/brs-commit-checker.yml
@@ -1,0 +1,93 @@
+name: BRS Commit Checker Report
+on:
+      workflow_dispatch:
+        inputs:
+            fix_version:
+                type: string
+                required: true
+                description: The ICU Jira "Fix Version" semver
+            from_git_ref:
+                type: string
+                required: true
+                description: The git ref start of comparison range. Prefix branches with `origin/`.
+            end_git_ref:
+                type: string
+                required: true
+                description: The git ref end of comparison range. Must be descendant of `from_git_ref`. Prefix branches with `origin/`.
+            # Jira user name & API token is used for processing sensitive tickets
+            jira_user_email:
+                type: string
+                required: false
+                description: (optional to include sensitive tickets) Email address of Jira user
+            # Jira user name & API token is used for processing sensitive tickets
+            # We protect the API token as a secret by preventing its value from pasted at the command line
+            # by using add-mask to obfuscate secret values in the jobs
+            # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions
+            jira_user_api_token:
+                type: string
+                required: false
+                description: (optional to include sensitive tickets) API token of Jira user
+
+jobs:
+    commit-report:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                fetch-tags: true
+                fetch-depth: 0
+            # workaround for bug in checkout action. this step should be redundant.
+            # https://github.com/actions/checkout/issues/1471
+            # https://github.com/actions/checkout/issues/1781
+            # https://github.com/actions/checkout/issues/701#issuecomment-1133937950
+            - name: Fetch all tags
+              run: |
+                git fetch --tags origin
+            - name: Fetch all branches
+              run: |
+                git fetch origin
+            - name: Setup Python
+              uses: actions/setup-python@v5
+              with:
+                python-version: '3.12.8'
+                cache: 'pipenv'
+                cache-dependency-path: |
+                    tools/commit-checker/Pipfile
+                    tools/commit-checker/Pipfile.lock
+            - name: Install pipenv
+              run: |
+                sudo pip3 install pipenv
+              # Use add-mask to obfuscate secret values in the jobs
+              # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions
+              # https://stackoverflow.com/a/71698374/2077918
+            - name: Store Jira credentials if provided
+              run: |
+                JIRA_USERNAME=$(cat $GITHUB_EVENT_PATH | jq -r '.inputs.jira_user_email' )
+                echo "::add-mask::${JIRA_USERNAME}"
+                JIRA_PASSWORD=$(cat $GITHUB_EVENT_PATH | jq -r '.inputs.jira_user_api_token' )
+                echo "::add-mask::${JIRA_PASSWORD}"
+                if [ ! -z "${JIRA_USERNAME}" ] && [ ! -z "${JIRA_PASSWORD}" ]
+                then
+                    echo "::add-mask::${JIRA_USERNAME}"
+                    echo "::add-mask::${JIRA_PASSWORD}"
+                    pushd ./tools/commit-checker
+                    echo "appending username and password to .env file"
+                    echo "JIRA_USERNAME=${JIRA_USERNAME}" >> ./.env
+                    echo "JIRA_PASSWORD=${JIRA_PASSWORD}" >> ./.env
+                    popd
+                else
+                    echo "Either jira_user_email or jira_user_api_token is empty. Not storing credentials file to enable access to sensitive tickets."
+                fi
+            - name: Generate report
+              run: |
+                pushd ./tools/commit-checker
+                pipenv install
+                pipenv run python3 check.py \
+                    --jira-query "project=ICU AND fixVersion=${{ inputs.fix_version }}" \
+                    --rev-range "${{ inputs.from_git_ref }}..${{ inputs.end_git_ref }}" > REPORT.md
+                popd
+            # https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/
+            - name: Reproduce report as workflow job summary
+              run: |
+                cat ./tools/commit-checker/REPORT.md >> $GITHUB_STEP_SUMMARY
+                echo "View the Summary page of this GHA Workflow instance to view the rendered Markdown of this report."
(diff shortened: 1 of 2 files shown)


.github/workflows/brs-commit-checker.yml
```diff
@@ -0,0 +1,93 @@
...
+            # Jira user name & API token is used for processing sensitive tickets
+            jira_user_email:
+                type: string
+                required: false
+                description: (optional to include sensitive tickets) Email address of Jira user
+            # Jira user name & API token is used for processing sensitive tickets
+            # We protect the API token as a secret by preventing its value from pasted at the command line
+            # by using add-mask to obfuscate secret values in the jobs
+            # https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions
+            jira_user_api_token:
```
mihnita, line 26: Did you consider making this a GitHub secret?

.github/workflows/brs-commit-checker.yml
```diff
@@ -0,0 +1,93 @@
+name: BRS Commit Checker Report
+on:
+      workflow_dispatch:
+        inputs:
+            fix_version:
+                type: string
+                required: true
+                description: The ICU Jira "Fix Version" semver
+            from_git_ref:
```
mihnita, line 9: Do we need this to be so flexible, from git-ref to git-ref?
Working with tags would be a lot easier and more readable
(think from `release-76-1` to `release-77-rc`)

.github/workflows/brs-commit-checker.yml
```diff
@@ -0,0 +1,93 @@
...
+              # https://stackoverflow.com/a/71698374/2077918
+            - name: Store Jira credentials if provided
+              run: |
+                JIRA_USERNAME=$(cat $GITHUB_EVENT_PATH | jq -r '.inputs.jira_user_email' )
+                echo "::add-mask::${JIRA_USERNAME}"
+                JIRA_PASSWORD=$(cat $GITHUB_EVENT_PATH | jq -r '.inputs.jira_user_api_token' )
+                echo "::add-mask::${JIRA_PASSWORD}"
+                if [ ! -z "${JIRA_USERNAME}" ] && [ ! -z "${JIRA_PASSWORD}" ]
+                then
+                    echo "::add-mask::${JIRA_USERNAME}"
```
mihnita, line 71: Sorry, not an expert, but this has a bit of "code smell"
When working with GitHub secrets there is a lot of masking done automatically.

I would have to dig a bit for best practices.

.github/workflows/brs-commit-checker.yml
echeran: This derives from how git works. When you issue a command (ex: `checkout`, `reset`, etc.), what you provide as an argument is called a git ref. If "ref" is an interface/supertype, then concrete types of a ref include tags, commit hashes, and branch names.

As you know, with git, the implementation is the specification. So I presume the flexibility comes from git.

I also personally don't think the the flexibility is a problem in the first place.

.github/workflows/brs-commit-checker.yml
```diff
@@ -0,0 +1,93 @@
...
+              uses: actions/setup-python@v5
+              with:
+                python-version: '3.12.8'
+                cache: 'pipenv'
+                cache-dependency-path: |
+                    tools/commit-checker/Pipfile
+                    tools/commit-checker/Pipfile.lock
+            - name: Install pipenv
+              run: |
+                sudo pip3 install pipenv
```
mihnita, line 59: Can't `venv` do the same kind of things (out of the box, no need to install anything)

I did a very quick scan of several articles and there seems to be agreement that `pipenv` is more powerful.

But I've also seen several recommendations to start with `venv`, get familiar with virtual environments, and when/if you need more power switch to `pipenv`.

More power often means added complexity :-)

.github/workflows/brs-commit-checker.yml
echeran: See the links that I pasted in the comments above this step. Within the first link, see [this section about `add-mask`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#masking-a-value-in-a-log)

.github/workflows/brs-commit-checker.yml
echeran: I considered that out of scope. This is just automating the existing instructions. 

I also don't know enough about either to comment, but I would be interested to learn more, assuming that this topic is orthogonal to the objective of the PR.

.github/workflows/brs-commit-checker.yml
mihnita: I didn't suggest making it an input. 
Make it a secret, always. 
The same way we have the git token to manipulate the release. 
Or the gpg password and gpg keys used to sign the release files.

.github/workflows/brs-commit-checker.yml
mihnita: Does it mean that I can use `release-76-rc`?
And that's a git-ref in GitHub lingo?

In fact in git a tag is basically an alias to a ref, or a pointer to a ref (however you want to look at it)

So maybe it's not really a GitHub lingo, but a git one :-)

.github/workflows/brs-commit-checker.yml
sffc: I wrote this script and chose pipenv due to its simplicity and reproducibility. It behaves more like a modern package manager. I've tried many ways to install Python dependencies and this is by far my favorite.

.github/workflows/brs-commit-checker.yml
sffc: I think I wrote the script such that environment variables for the secrets would work. I use a dotenv file locally but you don't need to in CI if there is a better option

.github/workflows/brs-commit-checker.yml
echeran: Yes, `release-76-rc` is a git tag, which is concrete type of a git ref, so it will work. And yes, that's an underlying git thing, not a Github-specific thing.

.github/workflows/brs-commit-checker.yml
echeran: It's a good question. The docs from Jira said that an API token cannot last more than 1 year, which seemed short to me. At most, 2 cycles. Creating a user token is fairly easy to do. Also, the Jira API token is tied to a specific user. And a quick search indicates that there's no possibility of a [user-neutral service account](https://community.atlassian.com/t5/Jira-questions/is-there-a-way-to-generate-API-for-a-service-account-or-team/qaq-p/2302499).

So for all those reasons, I felt like it's not much extra work for people to make their own token when they need it to the run the report.

But do you think it's still a net benefit to make the token into a GH secret, and just have a BRS task only to change out the API token every 6 months? Even though it's tied to a specific person? I could buy that. Just wondering what you think.

.github/workflows/brs-commit-checker.yml
echeran: I didn't see instructions about environment variables, But if that's possible, that would make using the Jira API tokens in the GHA workflows even safer via GH tokens.

echeran:
After updating the script to use Github secrets stored in the repository, the workflow still works. 

Demo/example:
- [workflow run](https://github.com/echeran/icu/actions/runs/13534183328/job/)
- [workflow run summary page - contains formatted report](https://github.com/echeran/icu/actions/runs/13534183328)

.github/workflows/brs-commit-checker.yml
```diff
@@ -0,0 +1,64 @@
...
+
+jobs:
+    commit-report:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+              with:
+                fetch-tags: true
+                fetch-depth: 0
+            # workaround for bug in checkout action. this step should be redundant.
```
mihnita, line 28: If it's a problem in the checkout action does it mean we have to do it in all of our actions?

.github/workflows/brs-commit-checker.yml
mihnita: > I don't like venv because you make a profile global to the system user

I am not sure I understand this.

With venv I create a virtual environment in whatever folder I want.
I can even create it under my file tree covered by a git repo.
There is nothing global, or system user.

> behaves more like a modern package manager

Not necessarily a plus in my book :-)

---

Not enough to block the pr on this.
It's in a cloud machine, whatever.

mihnita (approved):

.github/workflows/brs-commit-checker.yml
echeran: No, because most actions only care about the latest version of code that was created as a result of the PR's latest commit. The default behavior of `actions/checkout` is to give you that. 

In contrast, this workflow actually cares about being able to see all branches and tags of the repository, which is why we have to deal with this workaround here (and not yet anywhere else).

sffc (approved):

PR #3416, opened 2025-02-25
ICU-23060 Fix heap-buffer-overflow when UnicodeString size too large

FrankYFTang:
#### Checklist
- [X] Required: Issue filed: ICU-23060

### DIFF
--- a/icu4c/source/common/unistr.cpp
+++ b/icu4c/source/common/unistr.cpp
@@ -1945,6 +1945,11 @@ UnicodeString::cloneArrayIfNeeded(int32_t newCapacity,
       growCapacity = newCapacity;
     } else if(newCapacity <= US_STACKBUF_SIZE && growCapacity > US_STACKBUF_SIZE) {
       growCapacity = US_STACKBUF_SIZE;
+    } else if(newCapacity > growCapacity) {
+      return false;  // bad inputs
+    }
+    if(growCapacity > kMaxCapacity) {
+      return false;
     }
 
     // save old values
--- a/icu4c/source/test/intltest/ustrtest.cpp
+++ b/icu4c/source/test/intltest/ustrtest.cpp
@@ -77,6 +77,7 @@ void UnicodeStringTest::runIndexedTest( int32_t index, UBool exec, const char* &
     TESTCASE_AUTO(TestNullPointers);
     TESTCASE_AUTO(TestUnicodeStringInsertAppendToSelf);
     TESTCASE_AUTO(TestLargeAppend);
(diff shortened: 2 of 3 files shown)


icu4c/source/test/intltest/ustrtest.cpp
```diff
@@ -2351,6 +2352,14 @@ void UnicodeStringTest::TestUnicodeStringInsertAppendToSelf() {
     assertEquals("", u"abbcdcde", str);
 }
 
+void UnicodeStringTest::TestLargeMemory() {
+    IcuTestErrorCode status(*this, "TestLargeMemory");
+    constexpr uint32_t len = 2147483643;
+    char16_t *buf = new char16_t[len];
```
roubert, line 2358: This causes a MSVC compile error:

```
ustrtest.cpp(2358,25): error C2148: total size of array must not exceed 0x7fffffff bytes
```

roubert (changes requested):

icu4c/source/test/intltest/ustrtest.cpp
markusicu: Right, and it won't work on a 32-bit platform either.
I guess we need some conditional compilation, either collecting platforms where this does not work, or collecting some where it does.

How about
```
#if U_PLATFORM_IS_LINUX_BASED || U_PLATFORM_IS_DARWIN_BASED
...
#endif
```

... and also:
```
    if (buf == nullptr) { return; }
```

icu4c/source/common/unistr.cpp
```diff
@@ -1947,6 +1947,11 @@ UnicodeString::cloneArrayIfNeeded(int32_t newCapacity,
       growCapacity = US_STACKBUF_SIZE;
     }
 
+    // The newCapacity is too large.
+    if (newCapacity > growCapacity) {
+      return false;
+    }
```
markusicu, line 1953: let's first check for valid arguments,
then check the allocation size for the named max constant
```suggestion
    } else if(newCapacity > growCapacity) {
      return false;  // bad inputs
    }

    if(growCapacity > kMaxCapacity) {
      return false;
    }
```

markusicu (changes requested):
please re-target this PR to the maint/maint-77 branch

markusicu (approved):

richgillam (approved):
LGTM

roubert:
This seems to have broken the _Exhaustive Tests for ICU_:

https://github.com/unicode-org/icu/actions/runs/13533127126/job/37819569299

PR #3417, opened 2025-02-25
ICU-23061 LowercaseTransliterator has poor scalability due to synchronized state

stevenschlansker:
LowercaseTransliterator tries to reuse buffers to avoid allocation. Historically, this may have been an important optimization.
However on modern machines you are much more likely to use multiple CPUs, and modern JVMs are very efficient at GC, especially for local temporary data. Contending on a monitor is a big deal nowadays, especially for something so low level and hidden.

I've prepared a benchmark using [JMH](https://github.com/openjdk/jmh) to show the change in scalability:

```
Linux 6.12.13-200.fc41.x86_64 AMD Ryzen 7 3700X 8-Core Processor
JMH version: 1.37
JDK 23.0.2, OpenJDK 64-Bit Server VM, 23.0.2+7

Original `synchronized`:

1 Thread Benchmark                         Mode  Cnt     Score    Error   Units
LowercaseTransliteratorPerf.testSentence  thrpt    5   789.169 ±  2.779  ops/ms
LowercaseTransliteratorPerf.testShort     thrpt    5  5217.250 ± 35.501  ops/ms

8 Thread Benchmark                         Mode  Cnt     Score    Error   Units
LowercaseTransliteratorPerf.testSentence  thrpt    5   759.934 ±  4.880  ops/ms
LowercaseTransliteratorPerf.testShort     thrpt    5  3916.430 ± 70.545  ops/ms

New `stateless`:

1 Thread Benchmark                         Mode  Cnt     Score    Error   Units
LowercaseTransliteratorPerf.testSentence  thrpt    5   779.504 ±  6.507  ops/ms
LowercaseTransliteratorPerf.testShort     thrpt    5  5641.104 ± 29.747  ops/ms

8 Thread Benchmark                         Mode  Cnt      Score      Error   Units
LowercaseTransliteratorPerf.testSentence  thrpt    5   6340.495 ±  240.436  ops/ms
LowercaseTransliteratorPerf.testShort     thrpt    5  39132.015 ± 7102.311  ops/ms

```

The current implementation exhibits negative scalability: adding more threads hurts throughput (~5% loss going from 1T to 8T)
By removing the `synchronized` state, the implementation now scales perfectly (~8x increase going from 1T to 8T for the longer test case)
There is a very minor (~1%) penalty for pure single-threaded throughput since we no longer reuse state.
I believe this is acceptable in the context of allowing multiple threads to work without coordination.

#### Checklist
- [x] Required: Issue filed: ICU-23061

### DIFF
--- a/icu4j/main/translit/src/main/java/com/ibm/icu/text/BreakTransliterator.java
+++ b/icu4j/main/translit/src/main/java/com/ibm/icu/text/BreakTransliterator.java
@@ -22,8 +22,6 @@ import com.ibm.icu.util.ULocale;
 final class BreakTransliterator extends Transliterator {
     private BreakIterator bi;
     private String insertion;
-    private int[] boundaries = new int[50];
-    private int boundaryCount = 0;
 
     public BreakTransliterator(String ID, UnicodeFilter filter, BreakIterator bi, String insertion) {
         super(ID, filter);
@@ -52,8 +50,9 @@ final class BreakTransliterator extends Transliterator {
     public BreakIterator getBreakIterator() {
         // Defer initialization of BreakIterator because it is slow,
         // typically over 2000 ms.
-        if (bi == null) bi = BreakIterator.getWordInstance(new ULocale("th_TH"));
-        return bi;
+        // Using a holder class for safe init without a volatile-read.
+        if (bi == null) bi = WordBreakIteratorHolder.BI;
+        return (BreakIterator) bi.clone();
     }
 
     ///CLOVER:OFF
@@ -74,10 +73,11 @@ final class BreakTransliterator extends Transliterator {
         | (1<<Character.ENCLOSING_MARK)
         ;
     @Override
-    protected synchronized void handleTransliterate(Replaceable text, Position pos, boolean incremental) {
-        boundaryCount = 0;
+    protected void handleTransliterate(Replaceable text, Position pos, boolean incremental) {
+        int[] boundaries = new int[50];
+        int boundaryCount = 0;
         int boundary = 0;
-        getBreakIterator(); // Lazy-create it if necessary
+        BreakIterator bi = getBreakIterator(); // Lazy-create it if necessary
         bi.setText(new ReplaceableCharacterIterator(text, pos.start, pos.limit, pos.start));
         // TODO: fix clumsy workaround used below.
         /*
@@ -416,4 +416,7 @@ final class BreakTransliterator extends Transliterator {
         }
     }
 
+    private static class WordBreakIteratorHolder {
+        static final BreakIterator BI = BreakIterator.getWordInstance(new ULocale("th_TH"));
+    }
 }
--- a/icu4j/main/translit/src/main/java/com/ibm/icu/text/LowercaseTransliterator.java
+++ b/icu4j/main/translit/src/main/java/com/ibm/icu/text/LowercaseTransliterator.java
@@ -42,8 +42,6 @@ class LowercaseTransliterator extends Transliterator{
     private final ULocale locale;
 
     private final UCaseProps csp;
-    private ReplaceableContextIterator iter;
-    private StringBuilder result;
     private int caseLocale;
 
     /**
@@ -54,8 +52,6 @@ class LowercaseTransliterator extends Transliterator{
         super(_ID, null);
         locale = loc;
         csp=UCaseProps.INSTANCE;
-        iter=new ReplaceableContextIterator();
-        result = new StringBuilder();
         caseLocale = UCaseProps.getCaseLocale(locale);
     }
 
@@ -63,7 +59,7 @@ class LowercaseTransliterator extends Transliterator{
      * Implements {@link Transliterator#handleTransliterate}.
      */
     @Override
-    protected synchronized void handleTransliterate(Replaceable text,
+    protected void handleTransliterate(Replaceable text,
                                        Position offsets, boolean isIncremental) {
         if(csp==null) {
             return;
@@ -73,8 +69,10 @@ class LowercaseTransliterator extends Transliterator{
             return;
         }
 
+        ReplaceableContextIterator iter = new ReplaceableContextIterator();
+        StringBuilder result = new StringBuilder();
+
         iter.setText(text);
-        result.setLength(0);
         int c, delta;
 
         // Walk through original string
--- a/icu4j/perf-tests/pom.xml
+++ b/icu4j/perf-tests/pom.xml
@@ -43,6 +43,73 @@
       <artifactId>commons-cli</artifactId>
       <version>${commons-cli.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-core</artifactId>
+      <version>${jmh.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <version>${jmh.version}</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
+
+  <profiles>
+    <profile>
+      <id>jmh_benchmark</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <annotationProcessors>
+                <annotationProcessor>org.openjdk.jmh.generators.BenchmarkProcessor</annotationProcessor>
+              </annotationProcessors>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>build-jmh-classpath</id>
+                <goals>
+                  <goal>build-classpath</goal>
+                </goals>
+                <configuration>
+                  <includeScope>runtime</includeScope>
+                  <outputProperty>jmhClasspath</outputProperty>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <configuration>
+              <mainClass>org.openjdk.jmh.Main</mainClass>
+              <arguments>
+                <argument>-f</argument>
+                <argument>1</argument>
+                <argument>-wi</argument>
+                <argument>5</argument>
+                <argument>-i</argument>
+                <argument>10</argument>
+              </arguments>
+              <systemProperties>
+                <property>
+                  <key>java.class.path</key>
+                  <value>${project.build.outputDirectory}${path.separator}${jmhClasspath}</value>
+                </property>
+              </systemProperties>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
 
 </project>
--- a/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
+++ b/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
@@ -1,0 +1,10 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+package com.ibm.icu.text;
+
+public class BreakTransliteratorAccess {
+    // Non-public access
+    public static Transliterator newInstance() {
+        return new BreakTransliterator("Any-Break", UnicodeSet.ALL_CODE_POINTS);
+    }
+}
--- a/icu4j/main/translit/src/main/java/com/ibm/icu/text/CaseFoldTransliterator.java
+++ b/icu4j/main/translit/src/main/java/com/ibm/icu/text/CaseFoldTransliterator.java
@@ -39,8 +39,6 @@ class CaseFoldTransliterator extends Transliterator{
     }
 
     private final UCaseProps csp;
-    private ReplaceableContextIterator iter;
-    private StringBuilder result;
 
     /**
      * Constructs a transliterator.
@@ -49,15 +47,13 @@ class CaseFoldTransliterator extends Transliterator{
     public CaseFoldTransliterator() {
         super(_ID, null);
         csp=UCaseProps.INSTANCE;
-        iter=new ReplaceableContextIterator();
-        result = new StringBuilder();
     }
 
     /**
      * Implements {@link Transliterator#handleTransliterate}.
      */
     @Override
-    protected synchronized void handleTransliterate(Replaceable text,
+    protected void handleTransliterate(Replaceable text,
                                        Position offsets, boolean isIncremental) {
         if(csp==null) {
             return;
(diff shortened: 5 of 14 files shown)


macchiati (approved):
It all looks good to me. 

NOTE: I can't assess the additional dependencies; someone else should check that they look secure.

stevenschlansker:
I needed to add the license header to the performance test.

stevenschlansker:
I removed the `:` from my commit messages, I think it was tripping up the checker. Also rebased on top of `main`.

richgillam:
The change itself looks reasonable, assuming the performance advantages are as billed.  I can't speak to the POM changes or the performance test, as that's not stuff I'm really familiar with.

Usually, bug fixes in ICU need to be done in both C++ and Java when appropriate.  Do we have a similar problem on the C++ side?  If so, I'd like to see a fix for that here too.

grhoten:
These changes remind me of something. I think this Java lowercase transliterator still uses the locale.  The ICU4C implementation [removed that long ago](https://github.com/unicode-org/icu/commit/f6860c5a759a13482aa15700e082b7e647c698d8) because of how the registry works with sharing the singleton objects. Though the implementation in Java hard codes it to be US English casing only. So it's not much different.

I thought the transliterator casing functionality in ICU4C was doing something funny with thread safety long ago, but I can't seem to find the concerning code anymore. Perhaps it was removed, or I missed it.

stevenschlansker:
> I can't speak to the POM changes or the performance test, as that's not stuff I'm really familiar with.

The introduction of JMH is optional, and if the project maintainers don't want to introduce it, I can remove these changes.
I see that ICU already has its own benchmark harness. JMH is nice because it handles a lot of the boilerplate for you - help messages, warming up the JVM, timing runs - in a way that the "JVM experts" have thought long and hard about. The benchmark could be converted to the ICU runner, or not committed, if that's preferred. But I think you'll find that more and more outside contributors would already come in with JMH familiarity.

richgillam:
We really need one or more people from the ICU4J side of things to comment here...

stevenschlansker:
I rebased and resolved conflicts.

stevenschlansker:
Anything I should do to move this PR forward? If the JMH benchmark is acceptable, I can squash the commits together. I didn't want to do that until we confirmed it would be accepted.

icu4j/perf-tests/pom.xml
```diff
@@ -43,6 +44,64 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.openjdk.jmh</groupId>
+      <artifactId>jmh-generator-annprocess</artifactId>
+      <version>${jmh.version}</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
+  <build>
```
mihnita, line 60: Can you please make this a profile?
So that this happens only when we want it, not all the time?

Also, please use [java-exec](exec:java) instead of shading + running java "by hand"

So we run the tests with `mvn exec:java ...`

Combined with a profile the end result would look something like this:
```
mvn exec:java -f tools/build/ -P jmh_benchmarks
```

See `icu4j/tools/build/pom.xml` with the `errorprone_report` profile for an example.

icu4j/main/translit/src/main/java/com/ibm/icu/text/LowercaseTransliterator.java
```diff
@@ -54,16 +52,14 @@ public LowercaseTransliterator(ULocale loc) {
...
-        result = new StringBuilder();
         caseLocale = UCaseProps.getCaseLocale(locale);
     }
 
     /**
      * Implements {@link Transliterator#handleTransliterate}.
      */
     @Override
-    protected synchronized void handleTransliterate(Replaceable text,
+    protected void handleTransliterate(Replaceable text,
```
mihnita, line 62: Please do the same thing for UppercaseTransliterator, TitlecaseTransliterator, CaseFoldTransliterator, BreakTransliterator, all in the same folder.

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/LowercaseTransliteratorPerf.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.ibm.icu.text.Transliterator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class LowercaseTransliteratorPerf {
```
mihnita, line 21: I didn't try it, but this looks way too small to show any improvement.
What are the results of this test with / without the fix?

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/LowercaseTransliteratorPerf.java
```diff
@@ -0,0 +1,35 @@
+// © 2016 and later: Unicode, Inc. and others.
```
mihnita, line 1: This is a new file, so please update the year to 2025, and remove the `2002-2008, International Business Machines ...` block, which is legacy.

mihnita (commented):
Added some comments.
Thank you very much!
Mihai

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/LowercaseTransliteratorPerf.java
mihnita: Sorry, I noticed the test results in the initial comment.

icu4j/perf-tests/pom.xml
```diff
@@ -15,6 +15,7 @@
 
   <properties>
     <module-name>perf_tests</module-name>
+    <jmh.version>1.37</jmh.version>
```
mihnita, line 18: Please declare this in the `icu4j/pom.xml`, together with the versions for all other dependencies.

icu4j/main/translit/src/main/java/com/ibm/icu/text/BreakTransliterator.java
```diff
@@ -52,8 +50,9 @@ public void setInsertion(String insertion) {
     public BreakIterator getBreakIterator() {
         // Defer initialization of BreakIterator because it is slow,
         // typically over 2000 ms.
-        if (bi == null) bi = BreakIterator.getWordInstance(new ULocale("th_TH"));
-        return bi;
+        // Using a holder class for safe init without a volatile-read.
+        if (bi == null) bi = WordBreakIteratorHolder.BI;
+        return (BreakIterator) bi.clone();
```
stevenschlansker, line 55: The Break transliterator is a bit different - since the `bi` is shared, we now have to clone it for each operation to get rid of `synchronized`. This assumes that `clone()` is reasonably efficient.

macchiati: The high bit is that it should be reasonably efficient.

I poked around, and It does have some state. But I think the only things that require much processing are some int[] buffers, either directly or via fBreakCache and fDictionaryCache.

The biggest chunk of data is in fData. But that is immutable, and so is just copied; there is no need for a deep clone.

macchiati (approved):
LGTM

icu4j/main/translit/src/main/java/com/ibm/icu/text/BreakTransliterator.java
stevenschlansker: Thanks for confirming. The benchmark also indicates that it's faster to clone a new break iterator, than it is to synchronize on a shared one, with multiple cores involved.

icu4j/main/translit/src/main/java/com/ibm/icu/text/BreakTransliterator.java
```diff
@@ -74,10 +73,11 @@ public void setBreakIterator(BreakIterator bi) {
         | (1<<Character.ENCLOSING_MARK)
         ;
     @Override
-    protected synchronized void handleTransliterate(Replaceable text, Position pos, boolean incremental) {
-        boundaryCount = 0;
+    protected void handleTransliterate(Replaceable text, Position pos, boolean incremental) {
+        int[] boundaries = new int[50];
+        int boundaryCount = 0;
```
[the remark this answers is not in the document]
stevenschlansker: One concern is that `ArrayList<Integer>` would also need to box the primitive values into wrapper objects, with an additional performance penalty. The existing primitive `int[]` does not have this problem

macchiati (dismissed):

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/BreakTransliteratorPerf.java
```diff
@@ -0,0 +1,30 @@
...
+import com.ibm.icu.text.BreakTransliteratorAccess;
+import com.ibm.icu.text.Transliterator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class BreakTransliteratorPerf {
```
mihnita:
```suggestion
public class BreakTransliteratorPerfTest {
```

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/LowercaseTransliteratorPerf.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.ibm.icu.text.Transliterator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class LowercaseTransliteratorPerf {
```
mihnita:
```suggestion
public class LowercaseTransliteratorPerfTest {
```

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/TitlecaseTransliteratorPerf.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.ibm.icu.text.Transliterator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class TitlecaseTransliteratorPerf {
```
mihnita:
```suggestion
public class TitlecaseTransliteratorPerfTest {
```

icu4j/perf-tests/src/main/java/com/ibm/icu/dev/test/perf/UppercaseTransliteratorPerf.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.ibm.icu.text.Transliterator;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+public class UppercaseTransliteratorPerf {
```
mihnita:
```suggestion
public class UppercaseTransliteratorPerfTest {
```

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
```diff
@@ -0,0 +1,8 @@
+package com.ibm.icu.text;
+
+public class BreakTransliteratorAccess {
```
mihnita, line 5: Why add this public class?
I can only see it used once, in BreakTransliteratorPerf.java
```
static final Transliterator TITLE = BreakTransliteratorAccess.newInstance();
```

Can we just do this?
```
static final Transliterator TITLE =
    new BreakTransliterator("Any-Break", UnicodeSet.ALL_CODE_POINTS);
```

mihnita (commented):
Thank you very much for the updates.

I only have a few comments, hopefully non-controversial.

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
stevenschlansker: The `BreakTransliterator` class is not public, so attempting to access it from the package the performance tests is in fails. The `Access` class is in the `.icu.text` package to access the non-public constructor.

I can remove this test, or move the performance test to `.icu.text` instead if preferred.

stevenschlansker:
I made the suggested renamings. Please let me know what you would like to do with the BreakTransliterator non-public access case.

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
mihnita: Thank you, I understand now.
I think the current solution is reasonable.

I know that the established pattern is to put the `Foo` and `FooTest` in the same namespace, but ICU is pretty old, with (in some cases) outdated practices.
We might tackle this at some point and move the tests.
But not now.

LGTM, and thank you for the patience.

echeran:
It looks like all of the issues have been addressed on this PR and we have been ready to merge, we just overlooked it. I did some minor bookkeeping details to push it over the finish line and finally merge it.

FYI on bookkeeping details needed before the merge: CI needed to be manually enabled again (not sure why, once has been good enough for new contributors). The CI copyright check job failed because we're adding a new file in this PR, and it didn't have that verbiage, so I amended the PR with the diff below to add it, and then did a force push back to the PR branch (and then reapproved and merged). 

```diff
diff --git a/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java b/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
index 5afbe1e92bd..a80b5da2fa7 100644
--- a/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
+++ b/icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
@@ -1,3 +1,5 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
 package com.ibm.icu.text;
 
 public class BreakTransliteratorAccess {
```

PR #3418, opened 2025-02-25
ICU-23073 Use string_view constants for option names and formatter names

catamorphism:
This PR implements most of the suggested changes from https://github.com/unicode-org/icu/pull/3216 that weren't handled before merging.

#### Checklist
- [x] Required: Issue filed: ICU-23073

[ICU-22907]: https://unicode-org.atlassian.net/browse/ICU-22907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/icu4c/source/i18n/messageformat2_formatter.cpp
+++ b/icu4c/source/i18n/messageformat2_formatter.cpp
@@ -24,6 +24,18 @@ U_NAMESPACE_BEGIN
 
 namespace message2 {
 
+    namespace functions {
+    static constexpr std::u16string_view DATETIME = u"datetime";
+    static constexpr std::u16string_view DATE = u"date";
+    static constexpr std::u16string_view TIME = u"time";
+    static constexpr std::u16string_view NUMBER = u"number";
+    static constexpr std::u16string_view INTEGER = u"integer";
+    static constexpr std::u16string_view TEST_FUNCTION = u"test:function";
+    static constexpr std::u16string_view TEST_FORMAT = u"test:format";
+    static constexpr std::u16string_view TEST_SELECT = u"test:select";
+    static constexpr std::u16string_view STRING = u"string";
+    }
+
     // MessageFormatter::Builder
 
     // -------------------------------------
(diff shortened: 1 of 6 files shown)


catamorphism:
> @catamorphism [ICU-22097](https://unicode-org.atlassian.net/browse/ICU-22097) redirects to a CLDR ticket. Please check the ticket number and fix PR title & description, and the commit messages.

@markusicu The ticket number was correct -- see https://unicode-org.atlassian.net/browse/ICU-22907 -- but for some reason, the link as pasted in the PR description was going to the CLDR ticket. I'm not sure why (it appeared to be the same), but I pasted the full link in instead of just leaving it was "ICU-22097" and it works now.

markusicu:
> > @catamorphism [ICU-22097](https://unicode-org.atlassian.net/browse/ICU-22097) redirects to a CLDR ticket. Please check the ticket number and fix PR title & description, and the commit messages.
> 
> @markusicu The ticket number was correct -- see https://unicode-org.atlassian.net/browse/ICU-22907 -- but for some reason, the link as pasted in the PR description was going to the CLDR ticket. I'm not sure why (it appeared to be the same), but I pasted the full link in instead of just leaving it was "[ICU-22097](https://unicode-org.atlassian.net/browse/ICU-22097)" and it works now.

So apparently the correct ticket is ICU-22907 but the commit message and the PR title still say ICU-22097 (note: swapping the 9 vs. 0).

We are about to release ICU 77. I think this PR is too late.

@catamorphism @richgillam

icu4c/source/i18n/messageformat2_function_registry_internal.h
```diff
@@ -39,7 +39,7 @@ namespace message2 {
 
         private:
         static UnicodeString getStringOption(const FunctionOptions& opts,
-                                             const UnicodeString& optionName,
+                                             const std::u16string_view optionName,
```
markusicu, line 42: Please remove the `const` from the string_view-by-value parameter, here and in the other functions.

icu4c/source/i18n/messageformat2_formatter.cpp
```diff
@@ -130,18 +142,18 @@ namespace message2 {
...
-            .adoptFormatter(FunctionName(UnicodeString("number")), number, success)
-            .adoptFormatter(FunctionName(UnicodeString("integer")), integer, success)
-            .adoptFormatter(FunctionName(UnicodeString("test:function")), new StandardFunctions::TestFormatFactory(), success)
-            .adoptFormatter(FunctionName(UnicodeString("test:format")), new StandardFunctions::TestFormatFactory(), success)
-            .adoptSelector(FunctionName(UnicodeString("number")), new StandardFunctions::PluralFactory(UPLURAL_TYPE_CARDINAL), success)
-            .adoptSelector(FunctionName(UnicodeString("integer")), new StandardFunctions::PluralFactory(StandardFunctions::PluralFactory::integer()), success)
-            .adoptSelector(FunctionName(UnicodeString("string")), new StandardFunctions::TextFactory(), success)
-            .adoptSelector(FunctionName(UnicodeString("test:function")), new StandardFunctions::TestSelectFactory(), success)
-            .adoptSelector(FunctionName(UnicodeString("test:select")), new StandardFunctions::TestSelectFactory(), success);
+        standardFunctionsBuilder.adoptFormatter(FunctionName(functions::DATETIME), dateTime, success)
```
markusicu, line 145: Why don't you just use the string literals directly? The compiler figures out the string_view length, so there is no runtime cost for strlen() or similar.
```suggestion
        standardFunctionsBuilder.adoptFormatter(FunctionName(u"datetime"), dateTime, success)
```

markusicu (changes requested):
I took a look now. This needs more work than just fixing the ticket numbers.

icu4c/source/i18n/messageformat2_formatter.cpp
catamorphism: The constants are only used once in this PR, but a future change adds a special case for the `:number` function - see https://github.com/unicode-org/icu/pull/3423 and specifically line 329 of messageformat2_evaluation.cpp in that PR. So I'd rather define them in one place and use the constants, to prevent typos.

icu4c/source/i18n/messageformat2_function_registry_internal.h
catamorphism: Fixed in [8500b13](https://github.com/unicode-org/icu/pull/3418/commits/8500b1344d108b4d471d973448b06146fc741acb)

icu4c/source/i18n/messageformat2_formatter.cpp
markusicu: I think the benefit of such constants (whose names are the same as their contents) is very low, but... ok.

icu4c/source/i18n/messageformat2_formatter.cpp
```diff
@@ -24,6 +24,18 @@ U_NAMESPACE_BEGIN
 
 namespace message2 {
 
+    namespace functions {
```
markusicu, line 27: Please do not indent the contents of a namespace -- here and elsewhere.

markusicu:
> > We are about to release ICU 77. I think this PR is too late.
> 
> Understood.

Thank you. Please create a new ticket for this work for ICU 78 (maybe a larger topic of which this is a part?) and redirect this PR to that.

icu4c/source/i18n/messageformat2_formatter.cpp
catamorphism: Fixed in [38b9065](https://github.com/unicode-org/icu/pull/3418/commits/38b9065748d0f7092f571fc23e855c277a350b2f) -- I didn't un-indent the rest of this file (the contents of `namespace message2` outside of `namespace functions`) to avoid making the diff too big, but I can do that if you prefer.

markusicu: Reformatting is better in a separate PR.

FYI: On the way from "tech preview" to "draft" for 78, we all need to take another look at the API vs. spec and vs. C++ coding standards, and there may be more cleanup changes from that.

markusicu (approved):
lgtm now tnx!
@richgillam ok?

srl295 (approved):
LGTM3

markusicu:
ticket accepted

PR #3423, opened 2025-03-01
ICU-23059 ICU4C MF2: Spec test updates

catamorphism:
Update spec tests to current version from message-format-wg

  - Update parser for changed name-start grammar rule
  - Validate number literals in :number implementation (since parser no longer does this)
  - Disallow `:number`/`:integer` select option set from variable

    See https://github.com/unicode-org/message-format-wg/pull/1016

    As part of this, un-skip tests where the `bad-option` error is expected, and implement validating digit size options (pending PR https://github.com/unicode-org/icu/pull/2973 is intended to do this more fully)

#### Checklist
- [x] Required: Issue filed: ICU-23059


icu4c/source/i18n/messageformat2_errors.h
```diff
@@ -66,6 +66,15 @@ namespace message2 {
     enum DynamicErrorType {
         UnresolvedVariable,
         FormattingError,
+        BadOptionError,
```
srl295, line 69: If these are meant as documentation, it would be better to put them where Doxygen can find them, not just when scrolling through.

```suggestion
        BadOptionError, //< Used to signal…
```

or

```suggestion
        /** Used to signal … */
        BadOptionError,
```

icu4c/source/i18n/messageformat2_errors.h
```diff
@@ -66,6 +66,15 @@ namespace message2 {
...
+        // This is used to signal errors from :number and :integer when a
+        // bad `select` option is passed. In this case, fallback output
+        // is not used, so it must be distinguished from a regular bad
+        // option error (but it maps to a bad option error in the final
+        // error code).
+        // See https://github.com/unicode-org/message-format-wg/blob/main/spec/functions/number.md#number-selection
+        // ("The formatting of the _resolved value_ is not affected by the `select` option.")
+        SoftBadOptionError,
```
srl295, line 77: docs on SoftBadOption ?

icu4c/source/i18n/messageformat2_evaluation.cpp
```diff
@@ -319,15 +348,50 @@ PrioritizedVariant::~PrioritizedVariant() {}
         }
         FormattedPlaceholder arg = std::move(*std::get_if<FormattedPlaceholder>(&p->argument));
 
+        // This condition can't be checked in the selector.
+        // Effectively, there are two different kinds of "bad option" errors:
+        // one that can be recovered from (used for select=$var) and one that
+        // can't (used for bad digit size options and other cases).
+        // The checking of the "soft" error has to be done here; otherwise,
```
srl295, line 355: maybe 'recoverable' is better than 'soft'?

icu4c/source/tools/toolutil/json-json.hpp
```diff
@@ -7757,22 +7757,22 @@ class lexer : public lexer_base<BasicJsonType>
                                     }
                                     else
                                     {
-                                        error_message = "invalid string: surrogate U+D800..U+DBFF must be followed by U+DC00..U+DFFF";
-                                        return token_type::parse_error;
+                                        // ICU PATCH
+                                        codepoint = codepoint1;
```
srl295, line 7761: This should be upstreamed as an option.
Also, ICU PATCH here should probably link to a future ticket noting the patch is here, `// TODO ICU-nnnn Patched because …`

srl295 (approved):
looks good, comments about:

- upstreaming patch (check with ICU-TC)
- improving documentation on enum and possibly name of 'soft' error

srl295 (commented):
Actually, what is the impact of the JSON changes on other clients of the JSON library if any? 

Does the MF library supporting malformed UTF-8 mean that the JSON parser must also? is there another solution for test files, such as using numeric escapes?

catamorphism:
> Actually, what is the impact of the JSON changes on other clients of the JSON library if any?

Well, there aren't any other clients within ICU that use the JSON library, as far as I know. If you mean "what would the impact be if the changes were upstreamed?", I'm not sure.

> Does the MF library supporting malformed UTF-8 mean that the JSON parser must also? is there another solution for test files, such as using numeric escapes?

The test file already does use a numeric escape: https://github.com/unicode-org/message-format-wg/blob/main/test/tests/syntax-errors.json#L195

but the (unpatched) JSON parser rejects it even so. Note that this particular test case tests for a syntax error, but the message string still has to be representable within the test file, so this particular problem is not caused by the MF library supporting malformed UTF-8 (although it does), but rather, the storage format for test cases needing to represent malformed UTF-8.

Hope that makes sense, if not, let me know.

icu4c/source/i18n/messageformat2_errors.h
catamorphism: Fixed in [e525cac](https://github.com/unicode-org/icu/pull/3423/commits/e525cacb99e2185ec4e6a10965015c84004f80d8)

icu4c/source/i18n/messageformat2_errors.h
catamorphism: Fixed in [e525cac](https://github.com/unicode-org/icu/pull/3423/commits/e525cacb99e2185ec4e6a10965015c84004f80d8)

icu4c/source/tools/toolutil/json-json.hpp
catamorphism: I'm not sure if it would be accepted upstream; see https://json.nlohmann.me/home/faq/#parse-errors-reading-non-ascii-characters --

"Only UTF-8 encoded input is supported which is the default encoding for JSON according to [RFC 8259](https://tools.ietf.org/html/rfc8259.html#section-8.1)."

Since the codepoint in the test .json file is escaped, I'm not sure why it's considered to be invalid UTF-8, but maybe I'm not understanding something.

srl295:
> > Actually, what is the impact of the JSON changes on other clients of the JSON library if any?
> 
> Well, there aren't any other clients within ICU that use the JSON library, as far as I know. If you mean "what would the impact be if the changes were upstreamed?", I'm not sure.

Nlohmann wrote somewhere that rejecting unpaired surrogates was required by the JSON standard, but I don't find support for that in either json.org or ECMA-404. 

> > Does the MF library supporting malformed UTF-8 mean that the JSON parser must also? is there another solution for test files, such as using numeric escapes?
> 
> The test file already does use a numeric escape: https://github.com/unicode-org/message-format-wg/blob/main/test/tests/syntax-errors.json#L195
> 
> but the (unpatched) JSON parser rejects it even so. Note that this particular test case tests for a syntax error, but the message string still has to be representable within the test file, so this particular problem is not caused by the MF library supporting malformed UTF-8 (although it does), but rather, the storage format for test cases needing to represent malformed UTF-8.

It may be better to have another representation in the test files, perhaps supporting BASE64 or something.  Will other implementations also run into trouble using these same test files on other platforms/environments?

something parallel to bidiIsolation

```json
    {
       "description": "various characters as unquoted-literal",
       "src": "e8Kh2J3hmoHigIvigJDigLDigaDigarjgIHugIDvt7B9",
       "exp": "wqHYneGageKAi+KAkOKAsOKBoOKBquOAge6AgO+3sA==",
       "encoding": "base64"
     }
```

> Hope that makes sense, if not, let me know.

ICU-TC should be aware of the vendoring/patch, and i think it would be a good idea to open a ticket upstream at least. In fact, it would make a pretty simple #ifdef switch upstream.

srl295 (approved):
approving - i think other issues are good to discuss but don't need to block

catamorphism:
I created https://github.com/unicode-org/message-format-wg/issues/1062 to air your suggestion about using base64 in the test.

catamorphism:
I also created https://unicode-org.atlassian.net/browse/ICU-23090 to track the ICU patch to the JSON parser, and will add that as a comment in the file.

catamorphism:
@srl295 I squashed and added a reference to [ICU-23090](https://unicode-org.atlassian.net/browse/ICU-23090) as a comment in json-json.hpp. That's the only change in the force-push, if you don't mind re-approving one more time.

srl295 (approved):

PR #3426, opened 2025-03-05
ICU-23062 Added null check in one spot in genrb where we know we've s…

richgillam:
…een crashes.

The ticket describes a null-pointer dereference on line 300 in `parse.cpp` (in the `genrb` tool).  Turns out we already had a fix for that in the Apple branch of ICU, but it never got applied back to open-source ICU.  This PR does that.  (No new unit tests are added, because this is an internal tool that runs all the time as part of the build process, and because the problem is intermittent.)

#### Checklist
- [x] Required: Issue filed: ICU-23062

### DIFF
--- a/icu4c/source/tools/genrb/parse.cpp
+++ b/icu4c/source/tools/genrb/parse.cpp
@@ -297,7 +297,7 @@ static char *getInvariantString(ParseState* state, uint32_t *line, struct UStrin
 
     if(!uprv_isInvariantUString(tokenValue->fChars, tokenValue->fLength)) {
         *status = U_INVALID_FORMAT_ERROR;
-        error(*line, "invariant characters required for table keys, binary data, etc.");
+        error((line == nullptr) ? 0 : *line, "invariant characters required for table keys, binary data, etc.");
         return nullptr;
     }
 


pedberg-icu (approved):
Seems reasonable, but it might be good to know why line would be null

richgillam:
> Seems reasonable, but it might be good to know why line would be null

`parseInteger()` and `parseIntVector()` are explicitly passing null for `line`.  I don't know if that's correct.  I think `line` is an output parameter, and we've had a convention in other places that if you don't care about a particular result, you can pass null to ignore it, so I'm guessing those functions have decided they don't care about `line` and changed it without checking whether that's okay?  I could change this or add internal documentation, but I wonder if that's worth it.

@pedberg-icu I'm guessing you made the change on the Apple side that I'm porting over now, so I was kind of assuming you knew more about what was going on here than I did...

pedberg-icu:
> @pedberg-icu I'm guessing you made the change on the Apple side that I'm porting over now, so I was kind of assuming you knew more about what was going on here than I did...

Well, it does not look familiar to me...but that is not saying much

PR #3433, opened 2025-03-13
ICU-23052 Fix the addLikelySubtags code

FrankYFTang:
ICU-23052 Fix the addLikelySubtags code

1. Remove the hacks from the cldr to icu converter code
2. Fix the addLikelySubtags code to handle the case where the data does not have und_Latn => mapping
3. Handle the fallback matching $lang_$region and $lang after matching $lang_$script_$region and $lang_$script

#### Checklist
- [X] Required: Issue filed: ICU-23052


FrankYFTang:
I am coding this on the maint77 branch only to make sure the data generation is correct. I can move the patch to the main branch if we do not want to land into 77

FrankYFTang:
move to https://github.com/unicode-org/icu/pull/3445

PR #3439, opened 2025-03-18
ICU-23077 Correct the internalGet call inside Calendar

FrankYFTang:
#### Checklist
- [X] Required: Issue filed: ICU-23077

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -4251,17 +4251,20 @@ int32_t Calendar::internalGetMonth(UErrorCode& status) const {
     if (U_FAILURE(status)) {
         return 0;
     }
-    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, status);
+    if (resolveFields(kMonthPrecedence) == UCAL_ORDINAL_MONTH) {
+        return internalGet(UCAL_ORDINAL_MONTH);
     }
-    return internalGet(UCAL_ORDINAL_MONTH, status);
+    return internalGet(UCAL_MONTH);
 }
 
-int32_t Calendar::internalGetMonth(int32_t defaultValue, UErrorCode& /* status */) const {
-    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, defaultValue);
+int32_t Calendar::internalGetMonth(int32_t defaultValue, UErrorCode& status) const {
+    if (U_FAILURE(status)) {
+        return 0;
     }
-    return internalGet(UCAL_ORDINAL_MONTH);
+    if (resolveFields(kMonthPrecedence) == UCAL_ORDINAL_MONTH) {
+        return internalGet(UCAL_ORDINAL_MONTH);
+    }
+    return internalGet(UCAL_MONTH, defaultValue);
 }
 
 BasicTimeZone*
--- a/icu4c/source/i18n/chnsecal.cpp
+++ b/icu4c/source/i18n/chnsecal.cpp
@@ -1177,10 +1177,14 @@ int32_t ChineseCalendar::internalGetMonth(int32_t defaultValue, UErrorCode& stat
     if (U_FAILURE(status)) {
         return 0;
     }
-    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, defaultValue);
+    switch (resolveFields(kMonthPrecedence)) {
+        case UCAL_MONTH:
+            return internalGet(UCAL_MONTH);
+        case UCAL_ORDINAL_MONTH:
+            return internalGetMonth(status);
+        default:
+            return defaultValue;
     }
-    return internalGetMonth(status);
 }
 
 ChineseCalendar::Setting ChineseCalendar::getSetting(UErrorCode&) const {


FrankYFTang:
@anba

icu4c/source/i18n/calendar.cpp
```diff
@@ -4247,14 +4247,11 @@ Calendar::internalSet(EDateFields field, int32_t value)
...
 }
 
-int32_t Calendar::internalGetMonth(UErrorCode& status) const {
-    if (U_FAILURE(status)) {
-        return 0;
-    }
+int32_t Calendar::internalGetMonth(UErrorCode& /* status */ ) const {
     if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, status);
+        return internalGet(UCAL_MONTH);
```
richgillam, line 4252: Are you sure this is what you want to do?  I thought I remembered there was another overload of `internalGet` that took _three_ parameters, the third of which was `status`.  Would it make more sense to use that one, with a dummy value for the second parameter, or does that not make sense (or waste time getting the same result?

icu4c/source/i18n/calendar.cpp
```diff
@@ -4247,14 +4247,11 @@ Calendar::internalSet(EDateFields field, int32_t value)
...
     }
-    return internalGet(UCAL_ORDINAL_MONTH, status);
```
richgillam, line 4257: Same comment as above.

richgillam (commented):
Are you _sure_ we don't care about the status anymore?

icu4c/source/i18n/calendar.cpp
```diff
@@ -4247,14 +4247,11 @@ Calendar::internalSet(EDateFields field, int32_t value)
     internalSet(static_cast<UCalendarDateFields>(field), value);
 }
 
-int32_t Calendar::internalGetMonth(UErrorCode& status) const {
-    if (U_FAILURE(status)) {
-        return 0;
-    }
```
markusicu, line 4253: Please keep this, and please add it into the following overload.

icu4c/source/i18n/calendar.cpp
```diff
@@ -4247,14 +4247,11 @@ Calendar::internalSet(EDateFields field, int32_t value)
...
+int32_t Calendar::internalGetMonth(UErrorCode& /* status */ ) const {
     if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, status);
+        return internalGet(UCAL_MONTH);
     }
-    return internalGet(UCAL_ORDINAL_MONTH, status);
+    return internalGet(UCAL_ORDINAL_MONTH);
```
markusicu, line 4254: Why does the corresponding line in the following overload not pass in the defaultValue?

icu4c/source/i18n/calendar.cpp
richgillam: I think I was thinking of this:

```
int32_t Calendar::internalGetMonth(int32_t defaultValue, UErrorCode& /* status */) const {
    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
        return internalGet(UCAL_MONTH, defaultValue);
    }
    return internalGet(UCAL_ORDINAL_MONTH);
}
```

So does it make sense for the version of `internalGetMonth()` that Frank is changing to just call the other version?

Then again, since that other version isn't doing anything with the error code either, maybe it doesn't really make a difference...

icu4c/source/i18n/calendar.cpp
```diff
@@ -4251,17 +4251,20 @@ int32_t Calendar::internalGetMonth(UErrorCode& status) const {
...
 
-int32_t Calendar::internalGetMonth(int32_t defaultValue, UErrorCode& /* status */) const {
-    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, defaultValue);
+int32_t Calendar::internalGetMonth(int32_t defaultValue, UErrorCode& status) const {
+    if (U_FAILURE(status)) {
+        return 0;
+    }
+    if (resolveFields(kMonthPrecedence) == UCAL_ORDINAL_MONTH) {
+        return internalGet(UCAL_ORDINAL_MONTH);
```
markusicu, line 4265: why not
```suggestion
        return internalGet(UCAL_ORDINAL_MONTH, defaultValue);
```
?

icu4c/source/i18n/chnsecal.cpp
```diff
@@ -1177,10 +1177,14 @@ int32_t ChineseCalendar::internalGetMonth(int32_t defaultValue, UErrorCode& stat
     if (U_FAILURE(status)) {
         return 0;
     }
-    if (resolveFields(kMonthPrecedence) == UCAL_MONTH) {
-        return internalGet(UCAL_MONTH, defaultValue);
+    switch (resolveFields(kMonthPrecedence)) {
+        case UCAL_MONTH:
+            return internalGet(UCAL_MONTH);
+        case UCAL_ORDINAL_MONTH:
+            return internalGetMonth(status);
```
markusicu, line 1184: why are you no longer passing the default values into these functions?
circular calls?
is that a problem for both of these?
should they both call internalGet() instead of internalGetMonth()?

icu4c/source/i18n/calendar.cpp
FrankYFTang:
```
    inline int32_t internalGet(UCalendarDateFields field, int32_t defaultValue) const {return fStamp[field]>kUnset ? fFields[field] : defaultValue;}
```
The default value is used when the value of that field is unset  the field is UCAL_ORDINAL_MONTH is set here, so there are no point to pass with it.

icu4c/source/i18n/chnsecal.cpp
FrankYFTang: if we already know the UCAL_MONTH is reolved to, then it must not be unset. In that case, there are no need to pass in the defaultValue since we already know we won't use the default. 
For the case of UCAL_ORDINAL_MONTH, we know there are value stored in the UCAL_ORDINAL_MONTH field, so we will use that value to calculate the month, therefore no need to pass in the default value. 
The default value is used in the condition that neither the field UCAL_MONTH is set nor UCAL_ORDINAL_MONTH is set.

FrankYFTang (commented):
1.

richgillam (approved):
Okay, I'm sold...

icu4c/source/i18n/chnsecal.cpp
markusicu: Ok, thanks. These things might benefit from comments, but maybe they are obvious to someone who knows this code.

markusicu (approved):

PR #3445, opened 2025-03-20
ICU-23052 Fix addLikelySubtags

FrankYFTang:
1. Remove hacks inside CLDR to ICU conversion tool
2. Relax the assertion check inside the CLDR to ICU conversion tool and allow no mapping for the case of und_Latn without mapping
3. Fix the maximization code to fallback to und_$region  => and und => . 
4. Regenarate langInfo.txt and langInfo.res

#### Checklist
- [X] Required: Issue filed: ICU-23052


FrankYFTang:
ping

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -715,13 +715,28 @@ LSR LikelySubtags::maximize(StringPiece language, StringPiece script, StringPiec
             } else {
                 iter.resetToState64(state);
                 value = trieNext(iter, "", 0);
-                U_ASSERT(value > 0);
+                // For the case of und_Latn
+                if (value < 0) {
```
richgillam, line 719: Looks like we went from `U_ASSERT(value > 0)` to `if (value < 0)` here.  What if `value == 0`?  Does that case never happen?

icu4j/main/core/src/main/java/com/ibm/icu/impl/locale/LikelySubtags.java
```diff
@@ -355,11 +355,26 @@ private LSR maximize(String language, String script, String region, boolean retu
                 } else {
                     iter.resetToState64(state);
                     value = trieNext(iter, "", 0);
-                    assert value > 0;
+                    if (value < 0) {
```
richgillam, line 358: Same question here: What if `value` is 0?

icu4c/source/common/loclikelysubtags.cpp
```diff
@@ -715,13 +715,28 @@ LSR LikelySubtags::maximize(StringPiece language, StringPiece script, StringPiec
...
+                    U_ASSERT(value == 0);
+                    int64_t trieUndEmptyState = iter.getState64();
+                    value = trieNext(iter, region, 0);
+                    // Fallback to und =>
+                    if (value < 0) {
+                        iter.resetToState64(trieUndEmptyState);
+                        value = trieNext(iter, "", 0);
+                        U_ASSERT(value > 0);
+                    }
+                }
```
richgillam, line 735: I've read this change several times, and I have absolutely no idea what it's doing.  Maybe I'm not the best person to review this.  I don't know that you need to add documentation, but since I remember `loclikelysubtags.cpp` being wildly complicated and arcane and I haven't yet had to spend the time to understand it, perhaps you could walk me through what's going on here?

richgillam (commented):
I can rubber-stamp this if you're in a hurry, but I'm kind of lost as to what's going on in this change and I'm wondering if I'm the best person to have review it...

richgillam (approved):
Thanks for the assertion change.  I'm still not following the main logic of this fix, but I trust you.

PR #3454, opened 2025-03-26
ICU-22885 Add parsing of approximately sign

sffc:
This adds support for parsing the approximately sign and fixes the bug observed in ICU-22885.

#### Checklist
- [x] Required: Issue filed: ICU-22885

### DIFF
--- a/icu4c/source/common/static_unicode_sets.cpp
+++ b/icu4c/source/common/static_unicode_sets.cpp
@@ -187,7 +187,13 @@ void U_CALLCONV initNumberParseUniSets(UErrorCode& status) {
     U_ASSERT(gUnicodeSets[PERCENT_SIGN] != nullptr);
     U_ASSERT(gUnicodeSets[PERMILLE_SIGN] != nullptr);
 
+    // The following don't currently have parseLenients in data.
+    U_ASSERT(gUnicodeSets[INFINITY_SIGN] == nullptr);
     gUnicodeSets[INFINITY_SIGN] = new UnicodeSet(u"[∞]", status);
+    U_ASSERT(gUnicodeSets[APPROXIMATELY_SIGN] == nullptr);
+    // This set of characters was manually curated from the
+    // values of the approximatelySign element of CLDR common/main/*.xml files.
+    gUnicodeSets[APPROXIMATELY_SIGN] = new UnicodeSet(u"[∼~≈≃約]", status);
     if (U_FAILURE(status)) { return; }
 
     U_ASSERT(gUnicodeSets[DOLLAR_SIGN] != nullptr);
--- a/icu4c/source/i18n/numparse_symbols.cpp
+++ b/icu4c/source/i18n/numparse_symbols.cpp
@@ -195,4 +195,21 @@ void PlusSignMatcher::accept(StringSegment& segment, ParsedNumber& result) const
 }
 
 
+ApproximatelySignMatcher::ApproximatelySignMatcher(
+        const DecimalFormatSymbols& dfs, bool allowTrailing) :
+            SymbolMatcher(
+                dfs.getConstSymbol(DecimalFormatSymbols::kApproximatelySignSymbol),
+                unisets::APPROXIMATELY_SIGN),
+            fAllowTrailing(allowTrailing) {
+}
+
+bool ApproximatelySignMatcher::isDisabled(const ParsedNumber& result) const {
+    return !fAllowTrailing && result.seenNumber();
+}
+
+void ApproximatelySignMatcher::accept(StringSegment& segment, ParsedNumber& result) const {
+    result.setCharsConsumed(segment);
+}
+
+
 #endif /* #if !UCONFIG_NO_FORMATTING */
--- a/icu4c/source/i18n/numparse_symbols.h
+++ b/icu4c/source/i18n/numparse_symbols.h
@@ -163,6 +163,23 @@ class U_I18N_API PlusSignMatcher : public SymbolMatcher {
     bool fAllowTrailing;
 };
 
+
+// Exported as U_I18N_API for tests
+class U_I18N_API ApproximatelySignMatcher : public SymbolMatcher {
+  public:
+    ApproximatelySignMatcher() = default;  // WARNING: Leaves the object in an unusable state
+
+    ApproximatelySignMatcher(const DecimalFormatSymbols& dfs, bool allowTrailing);
+
+  protected:
+    bool isDisabled(const ParsedNumber& result) const override;
+
+    void accept(StringSegment& segment, ParsedNumber& result) const override;
+
+  private:
+    bool fAllowTrailing;
+};
+
 } // namespace numparse::impl
 U_NAMESPACE_END
 
--- a/icu4c/source/test/intltest/numfmtst.cpp
+++ b/icu4c/source/test/intltest/numfmtst.cpp
@@ -156,6 +156,7 @@ void NumberFormatTest::runIndexedTest( int32_t index, UBool exec, const char* &n
   TESTCASE_AUTO(TestFormatAttributes);
   TESTCASE_AUTO(TestFieldPositionIterator);
   TESTCASE_AUTO(TestDecimal);
+  TESTCASE_AUTO(TestDecimalFormatParse7E);
   TESTCASE_AUTO(TestCurrencyFractionDigits);
   TESTCASE_AUTO(TestExponentParse);
   TESTCASE_AUTO(TestExplicitParents);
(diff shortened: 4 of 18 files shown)


FrankYFTang:
Do we also need a Java fix for this?

icu4c/source/common/static_unicode_sets.cpp
```diff
@@ -187,7 +187,11 @@ void U_CALLCONV initNumberParseUniSets(UErrorCode& status) {
     U_ASSERT(gUnicodeSets[PERCENT_SIGN] != nullptr);
     U_ASSERT(gUnicodeSets[PERMILLE_SIGN] != nullptr);
 
+    // The following don't currently have parseLenients in data.
+    U_ASSERT(gUnicodeSets[INFINITY_SIGN] == nullptr);
     gUnicodeSets[INFINITY_SIGN] = new UnicodeSet(u"[∞]", status);
+    U_ASSERT(gUnicodeSets[APPROXIMATELY_SIGN] == nullptr);
+    gUnicodeSets[APPROXIMATELY_SIGN] = new UnicodeSet(u"[∼~≈≃約]", status); // this set was manually curated
```
FrankYFTang, line 194: how is this set determeind? What does it based on ? having "約" in this set is strange? Could we have a comments about this?

sffc: I claim no moral authority for how this set was formed beyond "this set was manually curated".

What I did was open the xml files and look for characters used in the approximately pattern in various locales.

FrankYFTang: OK, here is what I tried
```
find common/main/* |xargs egrep approximatelySign |egrep -v "↑↑↑|unconfirmed"|cut -d '>' -f 2|cut -d '<' -f 1|sort -u
-
~
∼
≃
≈
ca.
dáàshì
dáàṣì
約
```

so... maybe adding comment as "This set of characters is gathered from the values of approximatelySign element of CLDR  common/main/*.xml files." /

sffc:
Added Java and fixed the comment.

richgillam:
I'm coming to the party late, so I apologize if these are dumb questions, but what are you actually doing here, and why is that the appropriate response to the original issue?  If I'm reading this correctly, this makes the number parser explicitly aware of the approximately sign (in all the various locales), but just basically ignores it in parsing.  Is that right, and is that what we want to do?
1. What does this symbol mean in practice?  Why would somebody be using it in text that we parse?
2. I get why calling `abort()` is a bad idea, but why wouldn't it be better to just signal a parse error?
3. If ignoring the character is the right thing to do, why do we need code to explicitly identify it as the approximately sign?  Couldn't you just have a generic list of characters that should be ignored in parsing?

sffc:
The bug uncovered that we didn't handle the approximately sign in parsing, even though it is supported in patters and in formatting, which I added relatively recently (a few years ago).

Treating the approximately sign the same way as the plus sign makes sense to me. With the plus sign, we accept it and it doesn't impact the resulting parsed value. I mostly copied the plus sign code to make the approximately sign code.

richgillam:
>The bug uncovered that we didn't handle the approximately sign in parsing, even though it is supported in patters and in formatting, which I added relatively recently (a few years ago).
>
>Treating the approximately sign the same way as the plus sign makes sense to me. With the plus sign, we accept it and it doesn't impact the resulting parsed value. I mostly copied the plus sign code to make the approximately sign code.

Okay, I'll accept that.  Thanks for the explanation.  Given that, the code looks okay to me.

richgillam (approved):
LOKTM

icu4c/source/common/static_unicode_sets.cpp
```diff
@@ -187,7 +187,12 @@ void U_CALLCONV initNumberParseUniSets(UErrorCode& status) {
     U_ASSERT(gUnicodeSets[PERCENT_SIGN] != nullptr);
     U_ASSERT(gUnicodeSets[PERMILLE_SIGN] != nullptr);
 
+    // The following don't currently have parseLenients in data.
+    U_ASSERT(gUnicodeSets[INFINITY_SIGN] == nullptr);
     gUnicodeSets[INFINITY_SIGN] = new UnicodeSet(u"[∞]", status);
+    U_ASSERT(gUnicodeSets[APPROXIMATELY_SIGN] == nullptr);
+    // This set of characters was manually curated from the values of the approximatelySign element of CLDR common/main/*.xml files.
```
FrankYFTang, line 194: please wrap the line in the comment. This is way too long I think.

icu4c/source/i18n/numparse_symbols.cpp
```diff
@@ -195,4 +195,18 @@ void PlusSignMatcher::accept(StringSegment& segment, ParsedNumber& result) const
 }
 
 
+ApproximatelySignMatcher::ApproximatelySignMatcher(const DecimalFormatSymbols& dfs, bool allowTrailing)
+        : SymbolMatcher(dfs.getConstSymbol(DecimalFormatSymbols::kApproximatelySignSymbol), unisets::APPROXIMATELY_SIGN),
```
FrankYFTang, line 199: line wrap for these two lines please

icu4c/source/test/intltest/numfmtst.cpp
```diff
@@ -6981,6 +6982,29 @@ void NumberFormatTest::TestDecimal() {
...
+    status = U_ZERO_ERROR;
+    dfmt = icu::DecimalFormat(u"~0", status);
+    ASSERT_SUCCESS(status);
+    dfmt.parse(u"200", result, status);
+    ASSERT_EQUALS(status, U_INVALID_FORMAT_ERROR);
+    status = U_ZERO_ERROR;
+    dfmt.parse(u"≈200", result, status);
+    ASSERT_SUCCESS(status);
+    if (result.getInt64() != 200) {
+        errln(UnicodeString(u"Got unexpected parse result: ") + DoubleToUnicodeString(result.getInt64()));
```
FrankYFTang, line 7004: line wrap please

icu4j/main/core/src/main/java/com/ibm/icu/impl/StaticUnicodeSets.java
```diff
@@ -263,7 +264,10 @@ public void put(com.ibm.icu.impl.UResource.Key key, Value value, boolean noFallb
         assert unicodeSets.containsKey(Key.PERCENT_SIGN);
         assert unicodeSets.containsKey(Key.PERMILLE_SIGN);
 
+        // The following don't currently have parseLenients in data.
         unicodeSets.put(Key.INFINITY_SIGN, new UnicodeSet("[∞]").freeze());
+        // This set of characters was manually curated from the values of the approximatelySign element of CLDR common/main/*.xml files.
```
FrankYFTang, line 269: line wrap te comment plese.

markusicu:
@sffc please take a look at the feedback from @FrankYFTang

icu4c/source/i18n/numparse_symbols.h
```diff
@@ -163,6 +163,23 @@ class U_I18N_API PlusSignMatcher : public SymbolMatcher {
     bool fAllowTrailing;
 };
 
+
+// Exported as U_I18N_API for tests
+class U_I18N_API ApproximatelySignMatcher : public SymbolMatcher {
+  public:
+    ApproximatelySignMatcher() = default;  // WARNING: Leaves the object in an unusable state
```
FrankYFTang, line 170: what is the "WARNING:" comment for?

sffc: It's copied from above. The default constructor needs to exist because otherwise anything that contains a field of this type isn't constructable. You can verify that the only way that the code references one of these is after calling the main constructor.

sffc:
I used the tool to squash the branch because @FrankYFTang's comments are stylistic only. It would take me more time to find the branch and fix the style. I might try to get to it this week but I'm currently overwhelmed with urgent items and I want to make this mergeable for ICU 78 RC.

icu4c/source/common/static_unicode_sets.cpp
markusicu: done

icu4c/source/i18n/numparse_symbols.cpp
markusicu: done

icu4c/source/test/intltest/numfmtst.cpp
markusicu: done

icu4j/main/core/src/main/java/com/ibm/icu/impl/StaticUnicodeSets.java
markusicu: done

markusicu:
Let's not have this be stuck for a second release just on line wrapping... I have amended the commit to address @FrankYFTang 's feedback. @richgillam had approved this already in March, minus the new line wrapping. @sffc ok with my changes? I will also self-approve, and try to remember to merge by EOD if I don't hear otherwise.

markusicu (approved):

richgillam (approved):
Still looks good to me.

sffc:
Sure

PR #3456, opened 2025-03-26
ICU-23087 Pass null to applyPropertyAlias for a unary property query

eggrobin:
This only affects users of the internal perpetually-`@draft` XSymbolTable (in particular, unicode-org/unicodetools). It makes it possible to distinguish \p{X} (the characters that have binary property X, or that have gc X or script X) from \p{X=} (the characters for which property X is the empty string).

Note that ICU allows \p{Whitespace=} for \p{Whitespace}, or \p{L=} for \p{L}. This remains the case with this change, so the public API is unaffected.

#### Checklist
- [x] Required: Issue filed: ICU-23087

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
@@ -3601,7 +3601,7 @@ public class UnicodeSet extends UnicodeFilter implements Iterable<String>, Compa
             }
         }
 
-        if (valueAlias.length() > 0) {
+        if (valueAlias != null && !valueAlias.isEmpty()) {
             p = UCharacter.getPropertyEnum(propertyAlias);
 
             // Treat gc as gcm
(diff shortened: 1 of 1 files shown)


macchiati (dismissed):
Great idea!

markusicu:
> Note that ICU allows \p{Whitespace=} for \p{Whitespace}, or \p{L=} for \p{L}. This remains the case with this change, so the public API is unaffected.

I think we can & should propose to change those to errors.

icu4j/main/core/src/main/java/com/ibm/icu/text/UnicodeSet.java
```diff
@@ -3601,7 +3601,7 @@ public UnicodeSet applyPropertyAlias(String propertyAlias,
             }
         }
 
-        if (valueAlias.length() > 0) {
+        if (valueAlias != null && valueAlias.length() > 0) {
```
markusicu, line 3604: optional:
```suggestion
        if (valueAlias != null && !valueAlias.isEmpty()) {
```

eggrobin:
> I think we can & should propose to change those to errors.

I agree. But there will be more changes to UnicodeSet behaviour coming as part of the standardization, so I would like to do that when we have a draft standard to point to.

markusicu (dismissed):

markusicu (approved):

PR #3461, opened 2025-04-03
ICU-23095 Add fuzzer for DecimalFormatSymbols

FrankYFTang:
Add a fuzzer for DecimalFormatSymbols to explore possible cause of https://g-issues.chromium.org/issues/407149905

#### Checklist
- [X] Required: Issue filed: ICU-23095

### DIFF
--- a/icu4c/source/test/fuzzer/Makefile.in
+++ b/icu4c/source/test/fuzzer/Makefile.in
@@ -40,6 +40,7 @@ FUZZER_TARGETS = \
 		 collator_rulebased_fuzzer \
 		 converter_fuzzer date_format_fuzzer \
 		 date_time_pattern_generator_fuzzer \
+		 decimal_format_symbols_fuzzer \
 		 dtfmtsym_fuzzer \
 		 list_format_fuzzer locale_fuzzer \
 		 locale_morph_fuzzer \
--- a/icu4c/source/test/fuzzer/decimal_format_symbols_fuzzer.cpp
+++ b/icu4c/source/test/fuzzer/decimal_format_symbols_fuzzer.cpp
@@ -1,0 +1,95 @@
+// © 2025 and later: Unicode, Inc. and others.
+// License & terms of use: http://www.unicode.org/copyright.html
+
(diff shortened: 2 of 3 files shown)


richgillam (approved):
Seems reasonable to me.  You might also want @cjchapman to take a look.

cjchapman:
@FrankYFTang With the intermittent leaks we've seen of `DecimalFormatSymbols`, the allocation call stack always involves `DecimalFormat::touch()`.

If I understand this fuzzing test code, correctly, it is exercising `DecimalFormatSymbols` directly and not exercising the `DecimalFormat` code.

If that's the case, could you please extend this test to exercise `DecimalFormat` as well? At least enough to ensure that `DecimalFormat::touch()` gets called?

Thanks!

FrankYFTang:
The crash we observed in
i18n/numfmt.cpp on the line
```
       symbolsToAdopt.adoptInsteadAndCheckErrorCode(new DecimalFormatSymbols(desiredLocale, status), status);
```

it crash inside a uprv_free called from MaybeStackArray<char,40>::releaseArray I assume it is a double free or somehow the pointer is corrupted and pointing to somewere else

DecimalFormat is already test under 

https://github.com/unicode-org/icu/blob/e296ac93b57b006e5d715269cd3831d8abc2a3c3/icu4c/source/test/fuzzer/number_format_fuzzer.cpp#L20


I could change number_format_fuzzer.cpp to cause it to touch DecimalFormat::touch() but that should not be part of this fuzzer I think. I will open another PR for that.

cjchapman:
> I could change number_format_fuzzer.cpp to cause it to touch DecimalFormat::touch() but that should not be part of this fuzzer I think. I will open another PR for that.

Thanks!

FrankYFTang:
I try number_format_fuzzer, it already trigger DecimalFormat::touch()

cjchapman:
> I try number_format_fuzzer, it already trigger DecimalFormat::touch()

Ah, cool. I'm glad you're testing it. I take it you haven't seen any leaks with it yet?

PR #3480, opened 2025-04-22
ICU-23102 Simplify Calendar code

FrankYFTang:
Avoid calling unnecessary computeGregorianFields() Refactor and reduced handleComputeFields in CECalendar

#### Checklist
- [X] Required: Issue filed: ICU-23102

### DIFF
--- a/icu4c/source/i18n/ethpccal.h
+++ b/icu4c/source/i18n/ethpccal.h
@@ -162,19 +162,25 @@ protected:
      */
     virtual int32_t handleGetExtendedYear(UErrorCode& status) override;
 
-    /**
-     * Compute fields from the JD
-     * @internal
-     */
-    virtual void handleComputeFields(int32_t julianDay, UErrorCode &status) override;
-
     DECLARE_OVERRIDE_SYSTEM_DEFAULT_CENTURY
 
     /**
      * Return the date offset from Julian
      * @internal
      */
-    virtual int32_t getJDEpochOffset() const override;
+    int32_t getJDEpochOffset() const override;
+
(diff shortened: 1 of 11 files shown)


icu4c/source/i18n/ethpccal.h
```diff
@@ -330,6 +330,18 @@ class EthiopicAmeteAlemCalendar : public EthiopicCalendar {
...
+     * Compute the era from extended year.
+     * @internal
+     */
+    int32_t extendedYearToEra(int32_t extendedYear) const override;
+
+    /**
+     * Compute the year from extended year.
+     * @internal
+     */
+    int32_t extendedYearToYear(int32_t extendedYear) const override;
```
richgillam, line 344: It looks like you have the same thing here twice.  Is that intentional?

richgillam (commented):
Again, I'm not 100% sure, but I _think_ I see what you're doing here, and think it makes sense...

icu4c/source/i18n/ethpccal.h
FrankYFTang: There are two calnedar class here EthiopicCalendar and EthiopicAmeteAlemCalendar
each one of them have two functions extendedYearToEra and extendedYearToYear so there are total 4 added into this header

richgillam: Oh, wait.  It's two different function names.  Sorry, I missed that when I read this the first time.

richgillam (approved):
Sorry, I thought I approved this the last time I looked at it...

PR #3481, opened 2025-04-23
ICU-23097 Fix memory leak while fail

FrankYFTang:
#### Checklist
- [X] Required: Issue filed: ICU-23097

### DIFF
--- a/icu4c/source/common/lstmbe.cpp
+++ b/icu4c/source/common/lstmbe.cpp
@@ -809,7 +809,15 @@ U_CAPI const LSTMData* U_EXPORT2 CreateLSTMDataForScript(UScriptCode script, UEr
 
 U_CAPI const LSTMData* U_EXPORT2 CreateLSTMData(UResourceBundle* rb, UErrorCode& status)
 {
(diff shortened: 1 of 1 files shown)


richgillam (approved):
LGTM

icu4c/source/common/lstmbe.cpp
```diff
@@ -809,7 +809,15 @@ U_CAPI const LSTMData* U_EXPORT2 CreateLSTMDataForScript(UScriptCode script, UEr
 
 U_CAPI const LSTMData* U_EXPORT2 CreateLSTMData(UResourceBundle* rb, UErrorCode& status)
 {
-    return new LSTMData(rb, status);
+    if (U_FAILURE(status)) {
+        return nullptr;
+    }
+    const LSTMData* result = new LSTMData(rb, status);
```
aheninger, line 815: Need to check for a null pointer return from the new, and set status U_MEMORY_ALLOCATION_ERROR if it happens. The LocalPointer constructor with an error code is convenient for this.

PR #630, opened 2019-04-22
ICU-11533 Myanmar calendar support - suggestion to add [WIP]

mapmeld:
This is still a work in progress, but as an ICU and C++ newbie, I am reaching out for advice. Any help on setup, please contact me here or @mapmeld on Twitter.

Update: earlier I was overwriting BuddhCal.cpp with MyanCal.cpp to get it to run; for the PR to be more legible I have changed them back to separate files

##### Checklist

- [x] Issue filed: https://unicode-org.atlassian.net/browse/ICU-11533
- [x] Updated PR title and link in previous line to include Issue number
- [x] Issue accepted
- [x] Tests included
- [ ] Documentation is changed or added

CLDR ticket: https://unicode-org.atlassian.net/browse/CLDR-8081
Unicode / ICU ticket: https://unicode-org.atlassian.net/browse/ICU-11533

The code implements the Burmese/Myanmar lunisolar calendar for festival / holiday dates.  Some people in Myanmar use this calendar on official documents and forms, e.g. their birthdate, and it is not simple to convert between them.

### Comparing to a working reference

I am using the open source C++ and JS implementations from https://github.com/yan9a/mcal and making changes:
- renaming variables so their meaning is easier to understand
- changing variable type from long to int where a long is not needed (e.g. day / month / year )
- numbering months in chronological order
- using ClockMath.floorDivide instead of importing math.h ... but is this necessary? Following other calendar implementations
Currently the tests appear to have off-by-one errors (1 month and 1 day off) in most cases.

### Help with development

~~As mentioned earlier, development would be better if I can create calendar=myanmar separate from calendar=buddhist. Also~~ I currently run *all* icu4c tests with ```make check``` instead of the ones which I am targeting - Any help on setup, please contact me here or @mapmeld on Twitter.

### CLA

I have signed the CLA. Because the C++ code is based on Yan Naing Aye's repo, I've filed an Issue asking if he would sign the CLA as well.

PR #3482, opened 2025-04-23
ICU-11533 limited Myanmar calendar support

mapmeld:
**Background**
A while ago I proposed PR #630 to add the Myanmar lunar-based calendar (aka [Burmese calendar](https://en.wikipedia.org/wiki/Burmese_calendar)) to ICU. This calendar is frequently used for birthdays and on official documents.
After difficulties with supporting and testing the complete calendar and older dates, I closed the PR last year.

**Goals**
This new PR adds a limited Myanmar calendar that changes only the year; for example today (April 23, 2025 AD) would be represented as April 23, 1387 ME. This is similar to the modern Thai calendar (Buddhist / buddhcal in ICU).
A place in ICU and CLDR sets a foundation to support the months and days of the civil lunar cycle either in future updates or a region-specific API.

**Representing Myanmar new year**
The Myanmar new year ([Thingyan](https://en.wikipedia.org/wiki/Thingyan)) consistently falls in mid-April. This is possible by occurring at different dates or even different months on the complete Myanmar calendar; for example the current Myanmar year started on the 20th day of Tagu, and the next will start on the 2nd day of Kason.

This PR would increment years on April 17th:
- The year returned by this code is correct between April 17, 2000 AD and April 16, 2035 AD inclusive
- Other new year dates become more frequent over time, unless a secondary leap year (big watat) is declared 
- By nature, a discrepancy caused by a different new year lasts for only one day
- Corrections for individual years are not feasible, for example April 16, 1362 ME could be reversed to either 2000 or 2001 AD.

### DIFF
--- a/icu4c/source/i18n/calendar.cpp
+++ b/icu4c/source/i18n/calendar.cpp
@@ -50,6 +50,7 @@
 #include "coptccal.h"
 #include "dangical.h"
 #include "ethpccal.h"
+#include "myancal.h"
 #include "unicode/calendar.h"
 #include "cpputils.h"
 #include "servloc.h"
@@ -162,6 +163,7 @@ static const char * const gCalTypes[] = {
     "gregorian",
     "japanese",
     "buddhist",
+    "myanmar",
     "roc",
     "persian",
     "islamic-civil",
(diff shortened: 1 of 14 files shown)


FrankYFTang:
I think it is a very bad idea to support an incomplete (and therefore nobody use) calendar into ICU. See https://www.frontiermyanmar.net/en/the-myanmar-calendar/ the Myanmar calendar is not Gregorian based.

PR #3486, opened 2025-04-24
ICU-23113 Add nullptr handle for  in CollationDataBuilder::addCE32

wooffie:
#### Checklist
- [x] Required: Issue filed: ICU-23113


FrankYFTang:
can you add a unit test to trigger the bug if not fix?

markusicu:
It is an invariant that a FALLBACK_CE32 only occurs in a delta (tailoring) table, which only functions with a non-nullptr base pointer to the root table. A FALLBACK_CE32 means that the delta table has no data for a character, and without the base it could not process that character.

From icu4c/source/i18n/collation.h:
```
    /**
     * A CE32 is special if its low byte is this or greater.
     * Impossible case bits 11 mark special CE32s.
     * This value itself is used to indicate a fallback to the base collator.
     */
    static const uint8_t SPECIAL_CE32_LOW_BYTE = 0xc0;
    static const uint32_t FALLBACK_CE32 = SPECIAL_CE32_LOW_BYTE;
```

PR #3489, opened 2025-04-30
ICU-23107 Remove unnecessary virtual specifiers

DKLoehr:
Clang has recently added a warning for code with the `virtual` keyword on methods of `final` classes that don't override anything. Such classes can never be subclassed, so the keyword is meaningless. This fixes instances of that warning in ICU.

#### Checklist
- [X] Required: Issue filed: ICU-23107
- [ ] Tests included, if applicable (N/A)
- [ ] API docs and/or User Guide docs changed or added, if applicable (N/A)

### DIFF
--- a/icu4c/source/common/unicode/uniset.h
+++ b/icu4c/source/common/unicode/uniset.h
@@ -495,7 +495,7 @@ public:
      * @return <tt>true</tt> if the specified set is equal to this set.
      * @stable ICU 2.0
      */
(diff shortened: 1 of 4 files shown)


markusicu:
Thanks, I will take a look. I suspect that some of these don't want to just lose `virtual` but should actually have `override`.

roubert (approved):
OK, I've now verified that these deletions are all correct, those of the methods that need to have the `override` keyword already have that.

PR #3499, opened 2025-05-20
ICU-23004 Throw some C++20 at utfiterator.h

eggrobin:
Surprisingly, it looks like we got the iterators right.

We will need to tweak things a bit in order to properly support ranges though, see the TODOs.

#### Checklist
- [ ] Required: Issue filed: ICU-23004


eggrobin:
(Trying to fix the API bug in the return type of codePoints brought up ICU-22995 again, hence the flurry of intltest commits.)

eggrobin:
gcc11-cpp20 is failing because it does not have P2210R2; I need to tweak the feature test #if.

markusicu (commented):
very nifty!

it sounds like what's missing is
- the input specialization of UTFIterator needs a constructor where the limit unit iterator can be a different type (sentinel) than the start unit iterator
- the (unsafe)utfStringCodePoints want to be reformulated to work with ... at least a [std::range](https://en.cppreference.com/w/cpp/ranges.html)?
- the (unsafe)UTFStringCodePoints want to work with a range for code units, and *be* a range (or a view?) of code points, possibly a different type of code point range depending on the code unit range

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -488,7 +488,17 @@ class UTFIteratorTest : public IntlTest {
...
-        // UTFIterator<char32_t, UTF_BEHAVIOR_FFFD, std::ranges::iterator_t<CodeUnitRange>> it;
+        // Markus: I got the iterator as such to compile, but the constructor
+        // sees a sentinel type for the limit (as expected),
+        // so this still does not compile.
+        // TODO: Do we have to change all of the iterator and range templates to
+        // take two iterator template parameters (UnitIter, LimitIter)
+        // ... and change all iterator call sites?
+        // I can't find a way to deduce the sentinel type from the iterator type.
+        // TODO: If we have to change the template API, can we limit the damage to
+        // the input iterator specialization?
```
markusicu: @eggrobin I think I made progress, but I don't know how to deal with the sentinel types from here except to break all of the template invocations.

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -487,18 +487,8 @@ class UTFIteratorTest : public IntlTest {
...
-        // TODO: Do we have to change all of the iterator and range templates to
-        // take two iterator template parameters (UnitIter, LimitIter)
-        // ... and change all iterator call sites?
-        // I can't find a way to deduce the sentinel type from the iterator type.
-        // TODO: If we have to change the template API, can we limit the damage to
-        // the input iterator specialization?
-        // UTFIterator<char32_t, UTF_BEHAVIOR_FFFD, std::ranges::iterator_t<CodeUnitRange>> it(
-        //     codeUnits.begin(), codeUnits.end());
+        auto it = utfIterator<char32_t, UTF_BEHAVIOR_FFFD>(codeUnits.begin(), codeUnits.end());
+        // TODO: Use `it`.
```
markusicu: I left this for you :-)

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -487,8 +487,23 @@ class UTFIteratorTest : public IntlTest {
...
+        // TODO(egg): UTFStringCodePoints could support this.
+        struct CodePoints : std::ranges::view_interface<CodePoints> {
+            using iterator =
+                UTFIterator<char32_t, UTF_BEHAVIOR_FFFD, std::ranges::iterator_t<CodeUnitRange>,
+                            std::ranges::sentinel_t<CodeUnitRange>>;
+            CodePoints(CodeUnitRange codeUnits) : codeUnits(codeUnits) {}
+            iterator begin() { return iterator(codeUnits.begin(), codeUnits.end()); }
+            std::default_sentinel_t end() {
+                return std::default_sentinel;
+            }
```
markusicu: Why not something like
```
auto end() { return codeUnits.end(); }
```
?

eggrobin: Because I wrote `bool operator==(default_sentinel_t)`, instead of `template<sentinel_for<UnitIter> Sentinel> bool operator==(Sentinel s)`.

I think since it’s an iterator that knows its end, it makes sense to make the default sentinel a sentinel for it, see https://eel.is/c++draft/default.sentinel.

For the one that does not know its end, we should have `template<sentinel_for<UnitIter> Sentinel> bool operator==(Sentinel s)`, and there the code point range should return the sentinel of the underlying code unit range.

markusicu: I am trying to make it work with C++17 as well, see my attempt in the generic (non-input-iterator) UTFIterator.

I assume that if someone uses any kind of sentinel type, then I can compare the current UnitIter with the sentinel object. This should work with both validating and unsafe code point iterators which would be very nice.

I also found bugs with using the right kind of UTFImpl, and your operator==(sentinel) is missing a condition which should make it not work correctly.

eggrobin: > I assume that if someone uses any kind of sentinel type, then I can compare the current UnitIter with the sentinel object.

Yes.

But in addition to that, our validating iterators (and not the unsafe ones) are iterators that know their bound, so it seems nice to make the default sentinel a sentinel for them, regardless of whether the underlying code unit iterator has a separate sentinel.

Effectively, this is giving them a .hasNext(), except that it is spelled != std::default_sentinel.

eggrobin: > your operator==(sentinel) is missing a condition which should make it not work correctly.

Ah yes, it should be incorrect if it is used after *. Coming up with a test that actually exercises that seems tricky though; normally you do the == after the ++ (to see that you can *), _then_ you *.

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -625,6 +598,25 @@ class UTFIteratorTest : public IntlTest {
         // Bidirectional but not contiguous (there are holes where the bytes are FF).
         static_assert(std::ranges::bidirectional_range<CodeUnitRange>);
         static_assert(!std::ranges::contiguous_range<CodeUnitRange>);
+#if 0
+        using CodePoints = decltype(utfStringCodePoints<char32_t, UTF_BEHAVIOR_FFFD>(codeUnits));
+        using CodeUnits = decltype(*std::declval<CodePoints>().begin());
+        // TODO: fails static_assert(std::ranges::common_range<CodePoints>) below
+        // due to the bidi_iter branch of UTFStringCodePoints::end()
+        // failing with this root error:
```
markusicu: @eggrobin this one doesn't work yet

eggrobin: Interestingly, this is also the one where I have a test that fails at runtime in MSVC (see email).

I’ll take a look.

icu4c/source/test/intltest/utfiteratortest.cpp
```diff
@@ -680,6 +648,8 @@ class UTFIteratorTest : public IntlTest {
         const std::u16string codeUnits = u"𒀭𒊺𒉀​𒍠𒊩";
         static_assert(std::ranges::contiguous_range<decltype(codeUnits)>);
         auto codePoints = utfStringCodePoints<char32_t, UTF_BEHAVIOR_FFFD>(codeUnits);
+        auto end = codePoints.end();
+        static_assert(std::sentinel_for < decltype(end), decltype(codePoints.begin())>);
```
markusicu: @eggrobin Is this the debug assertion that you want to remove?
Is it useful to keep? (Except remove the spaces around `<` ?)

eggrobin: Oh right, I forgot to push that commit…

eggrobin:
Note: I will take a detour through ICU-23132, because it is seriously harming my productivity on this code.

icu4c/source/test/intltest/utfiteratortest.cpp
eggrobin: And it looks like it fails with 0f63cd5, too (so that commit successfully fixes the build configuration to run with stdcpplatest as intended).

markusicu: Testing at the "bleeding edge" (stdcpplatest) seems scary, since that might not be very stable. Just like early C++20 support wasn't stable, with some things missing and other things needing fixes that became unnecessary with later C++20 revisions and implementations.

I know that the workflow already tried to do stdcpplatest. But maybe we should fix that at, say, C++23 or even C++20, and review it after a release?

eggrobin: It would be nice to know that it works for people who are using the latest greatest compiler, though perhaps that should be a postsubmit best-effort check.

eggrobin:
It would probably be a good idea to make utfStringCodePoints a range adaptor object rather than a function…

eggrobin:
```
    // TODO: If I just take a const Range &unitRange and keep the reference in this object,
    // then some tests fail & others crash.
    // Should we take the reference and change the test code?
```
No.

The way this happens with, e.g., applying filter_view to a std::string is that we get a ref_view of the std::string; we would want the same thing to happen here, but we still want it to be safe to pass a temporary string.

eggrobin:
> It would probably be a good idea to make utfStringCodePoints a range adaptor object rather than a function…

Ah, we can only do that in C++23.

markusicu:
> > It would probably be a good idea to make utfStringCodePoints a range adaptor object rather than a function…
> 
> Ah, we can only do that in C++23.

We can't make a thing in C++17 that happens to satisfy the requirements for a range adaptor?

markusicu:
CI fails even with C++20...

eggrobin:
> We can't make a thing in C++17 that happens to satisfy the requirements for a range adaptor?

I think we need to inherit from `std::ranges::range_adaptor_closure`. But we could have a function object that only inherits from that in C++23, and then you can use the pipe syntax in C++23, and in C++20 it behaves like the current function.

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -1743,7 +1743,7 @@ class UTFStringCodePoints {
  */
 template<typename CP32, UTFIllFormedBehavior behavior>
 struct UTFStringCodePointsAdaptor
-#if defined(__cpp_lib_ranges) && __cpp_lib_ranges >= 2022'02
+#if U_CPLUSPLUS_VERSION >= 23 && defined(__cpp_lib_ranges) && __cpp_lib_ranges >= 2022'02
```
markusicu: Can we test U_CPLUSPLUS_VERSION alone? Do we need to also test __cpp_lib_ranges?

eggrobin: I would prefer to also test the feature flags, because support for a new version of C++ tends to come piecemeal. See the comments on the other feature tests where the GCC we use doesn’t have everything yet.

eggrobin: I can add a comment with a link to the corresponding P-document though.

markusicu: > I can add a comment with a link to the corresponding P-document though.

nice to have, and might be useful since we are (unusually for ICU) near the bleeding edge

eggrobin: Also, we don’t need to test for definedness, since that follows from C++23.

eggrobin: Done.

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -2454,9 +2457,10 @@ struct UnsafeUTFStringCodePointsAdaptor
 
 
 /**
- * UnsafeUTFStringCodePoints factory function for a "range" of code points in a code unit range.
- * The string must be well-formed.
- * Deduces the Range template parameter from the input.
+ * Range adaptor function object returning a UnsafeUTFStringCodePoints object that represents a
```
markusicu:
```suggestion
 * Range adaptor function object returning an UnsafeUTFStringCodePoints object that represents a
```

markusicu:
@eggrobin might we be done here? I am running out of reasonable TODOs.

markusicu:
By the way, I think we should squash the commits here between reviewer approval and merging. The many little bug fix commits don't seem worth preserving.

eggrobin:
Yes, we definitely don’t want to rebase this mess as-is.

0f63cd5f9d81e21e92a06b3b8f91a4d320d330a7 probably wants to be its own PR, too (by the way, it turns out stdcpplatest is how we get C++23, so we really want to keep that one in pre-merge CI; we could add an MSVC C++20 if you want).

The commit history reflects some amount of interesting back-and-forth, for instance, those commits where you described your compiler errors in an `#if 0`; I wonder if we should do the same as #3096 and exceptionally squash-merge it.

markusicu:
@richgillam or @roubert could either of you give at least a cursory look at this PR?

This is to make the new code point iterators work with C++20-style code unit range iterators, and make the code point "range" objects work with more code unit "ranges" and C++20/23 algorithms.

I have sent an amendment email about the API changes to the icu-design list. As you can see from the test code, they are backward compatible: The existing test code works with these changes, but of course we added a bunch of new test code.

markusicu:
> [0f63cd5](https://github.com/unicode-org/icu/commit/0f63cd5f9d81e21e92a06b3b8f91a4d320d330a7) probably wants to be its own PR, too

merged:
- https://github.com/unicode-org/icu/pull/3525

richgillam (approved):
I'm going to go ahead and approve this, although I'll be honest: even though I read through all the code changes, it was hard to understand what was going on-- there's too much exotic new-C++ stuff going on in here that I'm not familiar with and don't have time to really learn right now.  I know Markus also asked Fredrik to review this, and I hope he does and will defer to his judgment, as his C++-fu is (at this point) much stronger than mine.

The explanation of what you're doing any why in the API proposal all made sense to me and sounded like a win, and it looks like you managed to simplify the code quite a bit by factoring stuff out into separate classes and declarations, and that all looks really good to me.

The one thing that looks ugly and that I don't really understand is the stuff to deal with having limit sentinels that are a different type from the iterator type itself.  This is forcing you to jump through a bunch of ugly hoops to get the `==` and `!=` operators to work right.  The explanation of this most recent round of changes takes that for granted, so I guess the different types have been "a thing" in here for a while, but why?  What's that difference in types getting you?

eggrobin:
I can’t sleep because of jetlag, and breakfast isn’t open yet, so this sounds like a great time to teach some modern C++. Hopefully this explanation is vaguely coherent…

> The one thing that looks ugly and that I don't really understand is the stuff to deal with having limit sentinels that are a different type from the iterator type itself. This is forcing you to jump through a bunch of ugly hoops to get the == and != operators to work right

Part of the ugliness here is that we are doing this in C++17 as much as possible (even though the main benefit of separate sentinel types is to C++20).

> The explanation of this most recent round of changes takes that for granted, so I guess the different types have been "a thing" in here for a while, but why? What's that difference in types getting you?

There are a bunch of standard library C++20 ranges, especially ones that come from algorithms, that have these, so that is the motivation.
There are some examples in [utfiteratortest.cpp](https://github.com/unicode-org/icu/pull/3499/files#diff-af84540a5d80d23a4c359902f7452e467623ad2950224c3b68226949db78aa01) here.
A range that has a sentinel type which is the same as the iterator is called a [common_range](https://en.cppreference.com/w/cpp/ranges/common_range.html), so the relevant examples where these types are different have _uncommon_ in their name.

The standard library does that, for instance, when it has an algorithm that intrinsically knows when it is done, and doesn’t have a preordained end position, such as « iterate over the subsequence that starts at the beginning and ends at the first `#` » in `testUncommonBidirectionalRange`. There isn’t a meaningful `end` value, because you don’t know where the `#` is until you get there; but while iterating you can tell that you are done, it is when you see a `#`. So the sentinel is some dummy type (typically [default_sentinel_t](https://en.cppreference.com/w/cpp/iterator/default_sentinel.html)) and the == operator asks « are we done yet ? ».

richgillam:
> I can’t sleep because of jetlag, and breakfast isn’t open yet, so this sounds like a great time to teach some modern C++. Hopefully this explanation is vaguely coherent…

I think it is, and your explanation makes sense.  Thanks for that!

> > The one thing that looks ugly and that I don't really understand is the stuff to deal with having limit sentinels that are a different type from the iterator type itself. This is forcing you to jump through a bunch of ugly hoops to get the == and != operators to work right
> 
> Part of the ugliness here is that we are doing this in C++17 as much as possible (even though the main benefit of separate sentinel types is to C++20).

Does that mean that this ugliness will go away when we move to C++20 (whenever that might happen)?

eggrobin:
> Does that mean that this ugliness will go away when we move to C++20 (whenever that might happen)?

Yes, _a lot_ of this will become vastly more readable in C++20. In particular, all of the `enable_if`s will disappear and be replaced by concepts (sometimes used in requires clauses). You can see a preview of that glorious future here:
https://github.com/unicode-org/icu/blob/7f4fa13af2ddc51e749ec4a73a7569b65a0b3a89/icu4c/source/common/unicode/utfiterator.h#L292-L299

That is equivalent to
```C++
     template<typename Iter = UnitIter> 
     std::enable_if_t<std::contiguous_iterator<Iter>, std::basic_string_view<Unit>> stringView() const { 
```
but it uses a concept (`contiguous_iterator`) as a constraint instead of enable_if.

markusicu:
> > [0f63cd5](https://github.com/unicode-org/icu/commit/0f63cd5f9d81e21e92a06b3b8f91a4d320d330a7) probably wants to be its own PR, too
> 
> merged:
> 
> * [ICU-23054 Make the MSVC stdcpplatest build work as intended #3525](https://github.com/unicode-org/icu/pull/3525)

@eggrobin can you please rebase on main so that that PR replaces your earlier commit?

I can then try to give the whole thing another quick look at rubber-stamp. Then we can try the squash-n-merge dance.

eggrobin:
Rebasing a long history like that is tricky (I am getting conflicts that have been resolved by subsequent commits).

But merging main back into it removed the .props/.vcxproj from the diff in https://github.com/unicode-org/icu/pull/3499/files, which I think should make it reviewable again.

(And of course the squish-and-merge means we won’t get those merge commits in the actual history in main.)

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -166,6 +166,71 @@ typedef enum UTFIllFormedBehavior {
 
 namespace U_HEADER_ONLY_NAMESPACE {
 
+namespace prv {
```
roubert, line 169: In the rest of ICU4C, the names `internal` or `impl` are generally used for such namespaces.

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -1010,19 +1063,73 @@ class UTFIterator {
...
      * @param other Another iterator
-     * @return true if this operator is not at the same position as the other one
+     * @return true if this iterator is not at the same position as the other one
      * @draft ICU 78
      */
     U_FORCE_INLINE bool operator!=(const UTFIterator &other) const { return !operator==(other); }
 
+    // Asymmetric equality & nonequality with a sentinel type.
+    // C++17: Need to define all four combinations of == / != vs. parameter order.
+    // Once we require C++20, we could remove all but the first == because
```
roubert: You might want to `#if` this to make it abundantly clear what it is that can be removed in the future and already now verify that this works.

icu4c/source/common/unicode/utfiterator.h
markusicu: I believe that we use `impl` only in Java.
I chose not to use `internal` here because when we compile ICU itself, the namespace is already `icu::internal`, and nesting another `internal` would be confusing and might be error-prone.

icu4c/source/common/unicode/utfiterator.h
eggrobin: Good idea. Done.

markusicu: You made only one of four sets of these C++17-only. I just did the same for the rest.

shyeyian:
compilation failed on g++:
```sh
g++ -std=c++26 -Wall -fdiagnostics-color=always -fmodules -fmodule-mapper=./bin/debug/module/mapper.txt -g -O0 -DDEBUG -I ./bin/debug/third_party/install/include -c ./module/icu.cpp -o ./bin/debug/module/icu.o
```

```text
In file included from ./module/icu.cpp:100:
 1750 |     : std::ranges::range_adaptor_closure<UTFStringCodePointsAdaptor<CP32, behavior>>
./bin/debug/third_party/install/include/unicode/utfiterator.h:1750:41: error: expected ‘{’ before ‘<’ token
 2455 |     : std::ranges::range_adaptor_closure<UnsafeUTFStringCodePointsAdaptor<CP32>>
./bin/debug/third_party/install/include/unicode/utfiterator.h:2455:41: error: expected ‘{’ before ‘<’ token
In module imported from main:
In module imported from icu:
./bin/debug/third_party/install/include/unicode/utfiterator.h:1750:41: error: expected template-name before ‘<’ token [-Wtemplate-body]
      |                                         ^
./bin/debug/third_party/install/include/unicode/utfiterator.h:2455:41: error: expected template-name before ‘<’ token [-Wtemplate-body]
```

eggrobin:
@AnonymousPC It sounds like you have a combination of compiler and standard library that claims to be for C++26 and in particular that has `__cpp_lib_ranges >= 202202L`, but that does not actually provide `std::ranges::range_adaptor_closure`. Which versions of GCC and libstdc++ are you using?

shyeyian:
> @AnonymousPC It sounds like you have a combination of compiler and standard library that claims to be for C++26 and in particular that has `__cpp_lib_ranges >= 202202L`, but that does not actually provide `std::ranges::range_adaptor_closure`. Which versions of GCC and libstdc++ are you using?

Linux, g++ version=15.1, libstdc++ default with g++.
I finally found that `utfiterator.h` forgots to `#include <ranges>` at the file beginning :)

icu4c/source/common/unicode/utfiterator.h
shyeyian, line 1:
```cpp
#if U_CPLUSPLUS_VERSION >= 20 && __cpp_lib_ranges
    #include <ranges>
#endif
    
```

markusicu: Proposed fix in https://github.com/unicode-org/icu/pull/3543

markusicu: merged

PR #3506, opened 2025-05-22
ICU-22995 templatize assert[Not]Equals of integral and enum types

eggrobin:
#### Checklist
- [x] Required: Issue filed: ICU-22995

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/test/intltest/intltest.cpp
+++ b/icu4c/source/test/intltest/intltest.cpp
@@ -18,6 +18,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <cinttypes>
 #include <cmath>
 #include <math.h>
 #include <string_view>
@@ -73,12 +74,12 @@ UCharToUnicodeString(char16_t c) { return {c}; }
 // [rtg] Just to get things working
 UnicodeString
 operator+(const UnicodeString& left,
-      long num)
+      int64_t num)
 {
     char buffer[64];    // nos changed from 10 to 64
     char danger = 'p';  // guard against overrunning the buffer (rtg)
 
-    snprintf(buffer, sizeof(buffer), "%ld", num);
+    snprintf(buffer, sizeof(buffer), "%" PRId64, num);
     assert(danger == 'p');
 
     return left + buffer;
@@ -86,12 +87,12 @@ operator+(const UnicodeString& left,
 
 UnicodeString
 operator+(const UnicodeString& left,
-      unsigned long num)
+      uint64_t num)
 {
     char buffer[64];    // nos changed from 10 to 64
     char danger = 'p';  // guard against overrunning the buffer (rtg)
 
-    snprintf(buffer, sizeof(buffer), "%lu", num);
+    snprintf(buffer, sizeof(buffer), "%" PRIu64, num);
     assert(danger == 'p');
 
     return left + buffer;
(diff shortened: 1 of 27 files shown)


icu4c/source/test/intltest/intltest.cpp
```diff
@@ -2091,12 +2084,22 @@ UBool IntlTest::assertEquals(const char* message, std::u16string_view expected,
...
 }
 
-UBool IntlTest::assertEquals(const char* message,
-                             int32_t expected,
-                             int32_t actual) {
+bool IntlTest::assertSigned64Equals(const char *message, int64_t expected, int64_t actual) {
     if (expected != actual) {
-        errln(UnicodeString("FAIL: ") + message + "; got " +
-              actual + "=0x" + toHex(actual) +
+        errln(UnicodeString("FAIL: ") + message + "; got " + actual + "; expected " + expected);
```
markusicu, line 2089: no hex?

icu4c/source/test/intltest/intltest.cpp
```diff
@@ -241,6 +230,10 @@ UnicodeString toString(UBool b) {
   return b ? UnicodeString("true"):UnicodeString("false");
```
markusicu, line 230: would it make sense to print the actual value if it's not 0 or 1, in case someone does assertEquals with tiny ints that are not booleans?

icu4c/source/test/intltest/intltest.cpp
```diff
@@ -2143,18 +2146,29 @@ UBool IntlTest::assertEquals(const char* message,
...
 #ifdef VERBOSE_ASSERTIONS
     else {
-      logln(UnicodeString("Ok: ") + message + "; got " + toString(actual));
+        logln(UnicodeString("Ok: ") + message + "; got " + toString(actual));
+    }
+#endif
+    return true;
+}
+
+bool IntlTest::assertBooleanNotEquals(const char *message, int8_t expected, int8_t actual) {
```
markusicu, line 2163: for ...NotEquals(), should we share implementations with the positive versions and parameterize?

markusicu (commented):
very nice, thank you!

icu4c/source/test/intltest/intltest.cpp
eggrobin: See also the comments in the .h.

markusicu: I did know where we were ;-)
I am just wondering why we show hex for int32_t but only decimal for int64_t.
Sometimes large values would also be more readable in hex, e.g., for collation elements.

Not required here, but maybe consider adding hex output here?

icu4c/source/test/intltest/intltest.cpp
eggrobin: Doesn’t seem to be worth it for now. If we want to factorize more, I would probably go instead in the direction of a single assertEquals and a single assertNotEquals implementation, with a debugString function to show the values.

icu4c/source/test/intltest/intltest.cpp
eggrobin: The idea was that showing hex gets a bit spammy in general, e.g., when comparing sizes.

But if we think we want hex always, we can simplify the code drastically, by dropping the int32_t overload entirely.

eggrobin: Hmm, wait, no, it gets messy.  This 64-bit overload gets used for all numeric types.
If we have negative numbers, we don’t want to log the hex of the unsigned representation of their widening.
(That is, if we have an int8_t that is -1, we want to show FF, not FFFFFFFF; the latter would be massively confusing if you are debugging some byte-twiddling code.)

So this would require lifting some kind of debugString out of the assert functions as I wrote in the other comment, which is fine, but then it would require templatizing it so it doesn’t widen. That gets messy because we cannot use UnicodeString in the header.

Let’s not go there.

eggrobin: > That gets messy because we cannot use UnicodeString in the header.

OK, admittedly we don’t _need_ to use UnicodeString here, std::string would do the job just fine for printing out integers.

It would be a nontrivial rewrite though, and we have stuff on top of this. How about doing it in a future PR on the same ticket?

markusicu: > It would be a nontrivial rewrite though, and we have stuff on top of this. How about doing it in a future PR on the same ticket?

ok

PR #3507, opened 2025-05-23
ICU-22987 Return DONE for BreakIterator.preceding on negative index

echeran:
Make `BreakIterator.preceding()` return `DONE` when given a negative index for the input string. A negative index is already out of bounds and comes before the start of the string. The current behavior is to return 0, which doesn't make sense and is misleading. By contrast, the current behavior also returns 0 for an index of 1, which does make sense.

#### Checklist
- [X] Required: Issue filed: ICU-22987


markusicu:
Please also test C++ -- and fix if needed.

markusicu (approved):

PR #3518, opened 2025-06-06
ICU-20392 Split the Locale payload into nested and heap allocated

roubert:
All the most commonly used `Locale` objects have very little payload, most of them don't use any extensions, don't use a language tag longer than 3 characters and don't use more than a single variant.

There's room for all that data in a simple 32 byte large payload object, which can be nested directly in the `Locale` object.

Any payload larger than that can instead be heap allocated as needed, in order to save storage for the most commonly used objects while retaining the ability to create arbitrarily large and complex `Locale` objects.

This reduces the storage requirements for all `Locale` objects.

For nested payloads, this reduction is from 224 bytes to 48 bytes.

For payloads that need to be heap allocated, the reduction depends on several factors, but for most cases there's some reduction. There are also cases where this refactoring actually increases the storage used, because `CharString` allocates more storage than necessary. There are a number of ways in which this could be improved upon, such as optimizing `CharString` to not allocate more than necessary when copying a string of known length, not allocating any empty `CharString` objects or possibly replacing `CharString` with a new class for fixed length strings.

The public API remains unchanged but the operations which can lead to `U_MEMORY_ALLOCATION_ERROR` change.

#### Checklist
- [X] Required: Issue filed: ICU-20392

ALLOW_MANY_COMMITS=true


icu4c/source/common/locid.cpp
```diff
@@ -232,9 +234,88 @@ locale_get_default()
     return Locale::getDefault().getName();
 }
 
+namespace {
+
+template <auto FIELD, typename T>
+void copy_to_array(T* that, std::string_view sv) {
```
markusicu: ICU wants to use lowerCamelCase functions --> copyToArray

markusicu: In C++ we _mostly_ use (source, destination) param order. Parameter names that suggest which is input/source vs. output/target/destination also help.

icu4c/source/common/locid.cpp
```diff
@@ -408,64 +471,27 @@ Locale& Locale::operator=(const Locale& other) {
...
+        setToBogus();
+    } else if (const Nest* nest = std::get_if<Nest>(&other.payload)) {
+        payload.emplace<Nest>(*nest);
+    } else if (const std::unique_ptr<Heap>* heap = std::get_if<std::unique_ptr<Heap>>(&other.payload)) {
+        U_ASSERT(*heap);
+        if (UErrorCode status = U_ZERO_ERROR;
+            !payload.emplace<std::unique_ptr<Heap>>(std::make_unique<Heap>(**heap, status)) ||
+            U_FAILURE(status)) {
+            setToBogus();
+        }
```
markusicu: Please indent the 2nd & 3rd "if" lines so that the condition is not at the same level as the if body.

icu4c/source/common/locid.cpp
roubert: The initialization code that uses this helper function calls either `uprv_memcpy()` or `copy_to_array()`, depending on what's being copied to where, so I though it'd increase readability if they both looked similar, in naming as well as in parameter ordering:

```c++
uprv_memcpy(this, &other, sizeof *this);
copy_to_array<&Nest::language>(this, language);
```

icu4c/source/common/locid.cpp

roubert:
For the first step here, I'd very much like to use the well-tested and type-safe standard library `std::variant<>`, as that means less new code that needs to be written, adding no risk of introducing any additional bugs.

The next most important optimization after that would then be to eliminate the storage currently wasted by `CharString` allocating more than needed for strings of known size.

With that done, these changes will have resulted in significant size reductions for all kinds of `Locale` objects, finally making them small enough to use everywhere and making it possible to then start removing workarounds (such as ICU-23005) that currently use different kinds of strings instead of using `Locale` objects directly (saving some storage at the cost of having to repeatedly re-parse these strings).

At that point the primary problem will have been fully solved and then we can start looking into further improvements. I myself would then like to start out by taking a critical look at `Locale::initBaseName()`, which seems overly convoluted and maybe could be eliminated altogether by some thoughtful refactoring. And then, sure, replacing `std::variant<>` with something specialized could both save a few more bytes and be quite interesting to implement.

icu4c/source/common/locid.cpp
markusicu: I don't think this helps. Let's go with ICU conventions.

markusicu:
> For the first step here, I'd very much like to use the well-tested and type-safe standard library `std::variant<>`, as that means less new code that needs to be written, adding no risk of introducing any additional bugs.

Ok for a first step / first PR.

> The next most important optimization after that would then be to eliminate the storage currently wasted by `CharString` allocating more than needed for strings of known size.

I partially disagree. I think that making the Locale object smaller for commonly used locale IDs is most important.
I am ok if that's in a second PR.

CharString is "only" used for less-common locale IDs. sizeof(Heap) should be reduced after sizeof(Locale).

I agree that then there are other opportunities, such as making the parser/initBaseName() less convoluted, and even storing keywords in an at least slightly structured way, so that we need not traipse through the string all the time.

icu4c/source/common/locid.cpp

PR #3521, opened 2025-06-10
ICU-20392 DEBUG Windows DLL [DO NOT MERGE]

roubert:
#### Checklist
- [X] Required: Issue filed: ICU-20392

ALLOW_MANY_COMMITS=true


icu4c/source/common/servloc.h
```diff
@@ -422,7 +422,11 @@ class U_COMMON_API ICUResourceBundleFactory : public LocaleKeyFactory
  ******************************************************************
  */
 
-class U_COMMON_API ICULocaleService : public ICUService 
+class
+#if !(U_PF_WINDOWS <= U_PLATFORM && U_PLATFORM <= U_PF_CYGWIN)
```
markusicu: same as
```
#if !U_PLATFORM_HAS_WIN32_API
```

see https://github.com/unicode-org/icu/blob/0b253b86acecb862d6be75d98e50d0a5282794e9/icu4c/source/common/unicode/platform.h#L264-L265

markusicu: Feedback from the TC meeting: Change this to appropriately define something like U_COMMON_API_CLASS -- same as U_COMMON_API except empty on Windows.

roubert: Done.

roubert:
There's a `testtagsguards.sh` failure in clang-release-build-and-test, what's the correct solution for that?

icu4c/source/common/unicode/locid.h
```diff
@@ -1165,6 +1177,11 @@ class U_COMMON_API Locale : public UObject {
     char fullNameBuffer[ULOC_FULLNAME_CAPACITY];
     // name without keywords
     char* baseName;
+
+    struct Foo;
+    std::unique_ptr<Foo> foo;
+    std::variant<std::monostate, int, std::unique_ptr<Foo>> bar;
```
markusicu: This looks like a leftover from experimenting. Please remove.

icu4c/source/common/unicode/locid.h
```diff
@@ -369,7 +372,7 @@ class U_COMMON_API Locale : public UObject {
      * @system
      * @stable ICU 2.0
      */
-    static const Locale& U_EXPORT2 getDefault();
+    U_EXPORT static const Locale& getDefault();
```
markusicu: why are some methods U_EXPORT instead of U_COMMON_API?

icu4c/source/common/unicode/utypes.h
```diff
@@ -384,6 +384,81 @@ typedef double UDate;
 #define U_TOOLUTIL_API U_IMPORT
 #endif
 
+/**
+ * \def U_DATA_API_CLASS
+ * Set to export library symbols from inside the stubdata library,
+ * and to import them from outside, to be used on a class.
+ * @draft ICU 78
```
markusicu, line 393: > There's a testtagsguards.sh failure in clang-release-build-and-test, what's the correct solution for that?

You need to surround the new @draft definitions with
```
#ifndef U_HIDE_DRAFT_API
#endif  // U_HIDE_DRAFT_API
```

Or, in this case, since we can't compile without these, “use the force” -- U_FORCE_HIDE_DRAFT_API

markusicu, line 393: Please also send a [proposal email](https://icu.unicode.org/processes/proposal-template) for the new definitions.

icu4c/source/common/unicode/utypes.h
```diff
@@ -384,6 +384,81 @@ typedef double UDate;
...
+ */
+
+// When used on Windows, the U_..._API macros expand to __declspec(dllexport)
+// and __declspec(dllimport), which when used on a class results in all members
+// of the class being exported, including private members, which is problematic
+// for classes that have private members that can't be exported (such as
+// templates from the standard library):
+//
+// https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4251
+//
```
markusicu, line 445: mention ICU-23136 ?

icu4c/source/common/unicode/locid.h
roubert: I don't know, but it looks to me as if the pattern was that static members were U_EXPORT2 (while non-static members just inherited U_…_API from the class). Should it not be like that?

icu4c/source/common/unicode/locid.h
roubert: That's not a leftover, that's very intentionally there to verify that this all really works, just as the commit description says: "TEST: std::unique_ptr & std::variant members. [DO NOT MERGE]".

icu4c/source/common/unicode/utypes.h
roubert: Done.

icu4c/source/common/unicode/utypes.h
roubert: The text in the ticket and the text in the comment is the same, both copy-pasted from the mail I sent to icu-team@ to ask about this in the first place, so I don't think mentioning the ticket also in the comment adds any value (and the ticket is already mentioned in the revision log).

icu4c/source/common/unicode/utypes.h
roubert: Done: https://groups.google.com/a/unicode.org/g/icu-design/c/-Q82BiedfTQ

icu4c/source/common/unicode/locid.h
markusicu: I wasn't looking one commit at a time. As long as you remember to remove this commit before merging it's ok.

icu4c/source/common/unicode/locid.h
markusicu: The whole song and dance with these macros has grown over a long period of time trying to make various platforms and compilers happy.

I assume that U_COMMON_API on a whole class also applies to static members, that's why I am surprised you are not adding it to them as well. I think we needed the additional U_EXPORT2 for some consistency in calling conventions.

I also don't know what effect it has that you are replacing U_EXPORT2 after the return type with U_EXPORT before the return type. These kinds of things are largely compiler-specific. I seem to remember that the mainframe compilers were especially picky, but we don't have them in CI. I guess we will see if anyone complains after ICU 78 that their platform is now broken, and then we ask them to come up with a pull request that fixes theirs and doesn't break others.

Why are you replacing U_EXPORT2 after the return type with U_EXPORT before the return type?

icu4c/source/common/unicode/utypes.h
markusicu: Thanks. For reviewing, it would be nice if you didn't squash until the end. GitHub is bad with tracking comments, and worse across squashing/rebasing.

icu4c/source/common/unicode/locid.h
roubert: > Why are you replacing U_EXPORT2 after the return type with U_EXPORT before the return type?

That I do know! That's necessary to do, for otherwise MSVC will fail with a compilation error.

markusicu: > > Why are you replacing U_EXPORT2 after the return type with U_EXPORT before the return type?
> 
> That I do know! That's necessary to do, for otherwise MSVC will fail with a compilation error.

... but it has worked so far. Would it work with the U_EXPORT2 in its original place if the public static members also had U_COMMON_API like they (I assume) used to inherit from the enclosing class?

roubert: > Would it work with the U_EXPORT2 in its original place if the public static members also had U_COMMON_API like they (I assume) used to inherit from the enclosing class?

I don't know.

markusicu: > > Would it work with the U_EXPORT2 in its original place if the public static members also had U_COMMON_API like they (I assume) used to inherit from the enclosing class?
> 
> I don't know.

Please try this with one of the static functions and see if it passes CI.
I would be more comfortable if this was a more consistent move-U_COMMON_API-onto-every-public-member change.

roubert: > Please try this with one of the static functions and see if it passes CI.

I don't have any reasonably easy way to try that. These changes are already done through trial-and-error until I found something that happened to work. Trying out more things just to see if they might work too wouldn't be a meaningful use of time.

mihnita:
Digging in the export tables, and a bunch of entries seems to be dropped (not exported anymore).

I was unable to figure out a clear rule.
I've seen both static an non static methods disappearing, public / protected / private.

Most worrisome (I think) is the disappearance of some public entries, mostly constructors.
And not only the default ones, generated by the compiler, but constructors with parameters, that I can find declared in the header file.

The intriguing part is that not all constructors disappear, only some.
Again, I can't see the rule.

mihnita:
>
> Why are you replacing U_EXPORT2 after the return type with U_EXPORT before the return type?

> That I do know! That's necessary to do, for otherwise MSVC will fail with a compilation error.

I think that since `U_EXPORT` ends up being `__declspec(dllexport)`, this applies:

> To export functions, the __declspec(dllexport) keyword must appear to the left of the calling-convention keyword, if a keyword is specified. For example:
>
> ```
> __declspec(dllexport) void __cdecl Function1(void);
> ```
> To export all of the public data members and member functions in a class, the keyword must appear to the left of the class name as follows:
> ```
> class __declspec(dllexport) CExampleExport : public CObject
> { ... class definition ... };
> ```
https://learn.microsoft.com/en-us/cpp/build/exporting-from-a-dll-using-declspec-dllexport?view=msvc-170

So  `U_EXPORT` must be on the left of the return type.

---

`U_EXPORT2` maps to `__cdecl`

> Place the `__cdecl` modifier before a variable or a function name.

https://learn.microsoft.com/en-us/cpp/cpp/cdecl?view=msvc-170

I think that calling it `?_EXPORT?` is confusing, as these are not actually exported.
But hey, it is probably ancient history, not worth shaking the boat.

markusicu:
> Digging in the export tables, and a bunch of entries seems to be dropped (not exported anymore).
> [...]
> I've seen both static an non static methods disappearing, public / protected / private.
> Most worrisome (I think) is the disappearance of some public entries, mostly constructors.
> And not only the default ones, generated by the compiler, but constructors with parameters, that I can find declared in the header file.

C++ likes to auto-generate member functions. We have tried to define them explicitly, so that we can tag them with their status, but I think that's becoming less tenable with it generating more and more things like symmetric operator== and opterator!=.

If the auto-generated ones are no longer reliably exported, then that seems like a problem.

No longer reliably exporting explicitly declared methods is of course even worse :-(

> > Why are you replacing U_EXPORT2 after the return type with U_EXPORT before the return type?
> 
> > That I do know! That's necessary to do, for otherwise MSVC will fail with a compilation error.
> 
> I think that since `U_EXPORT` ends up being `__declspec(dllexport)`, this applies:

I reminded myself that [U_COMMON_API is the same as](https://github.com/unicode-org/icu/blob/09c6428b5fd88516704af53bb6a63a79b5c82e22/icu4c/source/common/unicode/utypes.h#L315) U_EXPORT or U_IMPORT or nothing depending on whether we compile ICU itself or user code.

So I think these static member functions should use U_COMMON_API instead of U_EXPORT.

And if it is harmless (= does not break CI) to also keep the U_EXPORT2 where it's always been, then so much the better.

IOW please try to treat the static member functions the same as the non-static ones.

@mihnita since you have a Windows machine, can you please try this?
For example:
```
U_COMMON_API static const Locale& U_EXPORT2 getDefault();
```

mihnita:
> @mihnita since you have a Windows machine, can you please try this?
> For example:
> U_COMMON_API static const Locale& U_EXPORT2 getDefault();

There is no difference in the export table between this and what Fredrik does.

mihnita:
> @mihnita since you have a Windows machine, can you please try this? For example:

I tried, there is no difference in the export table between this and the result with the current PR change.

And I am pretty sure I've commented this a few days ago, but somehow GitHub "forgot" it.
Or maybe my memory is playing tricks on me...

markusicu:
> > @mihnita since you have a Windows machine, can you please try this? For example:
> 
> I tried, there is no difference in the export table between this and the result with the current PR change.

Ok. So Windows/Mac/Linux don't care either way.

The current attribute combinations are a result of long-ago trial-and-error with a larger set of platforms. Is there any reason not to make the more conservative change? If not, then I am willing to add a commit to that effect.

roubert:
> Is there any reason not to make the more conservative change?

I would just like to avoid making any change at all before we find someone who can say what the preferred way of doing this really is. At the moment we have two potential options but we don't know whether any of them is the preferred way, there might be a third option that we're not yet aware of, so making any such change now would be wasted effort if it later needs to be changed again. That's all.

markusicu:
> > Is there any reason not to make the more conservative change?
> 
> I would just like to avoid making any change at all before we find someone who can say what the preferred way of doing this really is.

Well, we don't seem to have an actual expert.
You _are_ making changes, and it scares me a bit that they seem to do more than what we need.

Note also that the macros like U_COMMON_API resolve to either import or export depending on which code you are compiling. U_COMMON_API wants to _export_ from the common library, and _import_ into the i18n library, into ICU tools & tests, and into non-ICU code. Changing U_COMMON_API to just U_EXPORT seems misleading at best.

I will give it a try, see what CI says, and ask Mihai to try it on his Windows machine as well.

icu4c/source/i18n/collationtailoring.h
```diff
@@ -47,7 +47,7 @@ class UnicodeSet;
  *
  * It is shared, reference-counted, and auto-deleted; see SharedObject.
  */
-struct U_I18N_API CollationTailoring : public SharedObject {
+struct CollationTailoring : public SharedObject {
```
mihnita: Done we want to U_I18N_API_CLASS here?

icu4c/source/i18n/collationtailoring.h
```diff
@@ -94,7 +94,7 @@ struct U_I18N_API CollationTailoring : public SharedObject {
     CollationTailoring(const CollationTailoring &other) = delete;
 };
 
-struct U_I18N_API CollationCacheEntry : public SharedObject {
+struct CollationCacheEntry : public SharedObject {
```
mihnita: Done we want to U_I18N_API_CLASS here?

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -1519,7 +1519,7 @@ class U_I18N_API Padder : public UMemory {
 
 // Do not enclose entire MacroProps with #ifndef U_HIDE_INTERNAL_API, needed for a protected field
 /** @internal */
-struct U_I18N_API MacroProps : public UMemory {
+struct MacroProps : public UMemory {
```
mihnita: Done we want to U_I18N_API_CLASS here?

icu4c/source/i18n/unicode/numberrangeformatter.h
```diff
@@ -90,7 +90,7 @@ namespace impl {  // icu::number::impl
 
 // Do not enclose entire MacroProps with #ifndef U_HIDE_INTERNAL_API, needed for a protected field
 /** @internal */
-struct U_I18N_API RangeMacroProps : public UMemory {
+struct RangeMacroProps : public UMemory {
```
mihnita, line 93: Done we want to U_I18N_API_CLASS here?

Looks like you decided not to export `struct`s
Or is it because they are not public?

icu4c/source/i18n/collationtailoring.h
markusicu: I assume that with “Done we want...” you mean “Don't we want...”

icu4c/source/i18n/collationtailoring.h
markusicu: I haven't check, but this is not a public header, and if none of the tools & test code use the types here, then we need not export them.

icu4c/source/i18n/unicode/numberformatter.h
markusicu: not public API, so depends on whether tools+tests use it

mihnita:
There is a pretty long list of differences between what is exported "the old way" and "the new way"

There are differences even for members declared public, which I would expect to be preserved.
And there are still present a lot of members that are private or protected.
Which I would expect to stop being exported.

Let's say we don't worry too much about extra exports (although I kind of think was the point of this exercise?)

But missing public exports seems worrisome.

See here an archive with the old / new exports: https://mihai-nita.net/tmp/diff_exports.zip

---

There is no risk that I built at different commit points.
Because I went to head, built / dump / cleanup, then applied the PR as a patch, then built / dump.
So the only differences between the 2 dumps was the PR.

icu4c/source/i18n/collationtailoring.h
roubert: I thought that this had been tagged `U_I18N_API` by mistake, that it really shouldn't be exported as it isn't part of the public API. (Is that not so?)

icu4c/source/i18n/collationtailoring.h
roubert: I thought that this had been tagged `U_I18N_API` by mistake, that it really shouldn't be exported as it isn't part of the public API. (Is that not so?)

icu4c/source/i18n/unicode/numberformatter.h
roubert: I thought that this had been tagged `U_I18N_API` by mistake, that it really shouldn't be exported as it isn't part of the public API. (Is that not so?)

icu4c/source/i18n/unicode/numberrangeformatter.h
roubert: I thought that this had been tagged `U_I18N_API` by mistake, that it really shouldn't be exported as it isn't part of the public API. (Is that not so?)

roubert:
> There are differences even for members declared public, which I would expect to be preserved.

Apart from the ones that I believe to have been exported by mistake before, have you seen any other public declarations missing now? I can't find any such in your dump files (but I haven't checked every single entry that differs).

> And there are still present a lot of members that are private or protected.

Are any of those such members that the changes in this PR ought to remove? The changes in this PR only touch a tiny fraction of the code base, most private or protected members should be unaffected by these changes.

icu4c/source/i18n/collationtailoring.h
markusicu: See my previous comment. It's not in a public header file. We need to export non-public stuff if it's used in tools or in test code, but where that's not the case, we need not export. The U_I18N_API here was probably from copy-paste by the guy who rewrote the collation code.

icu4c/source/i18n/collationtailoring.h
markusicu: We checked. icu4c/source/tools/genrb/parse.cpp uses CollationTailoring, so we need to export it.

icu4c/source/i18n/collationtailoring.h
markusicu: used in .../tools/gencolusb/extract_unsafe_backwards.cpp so should export

icu4c/source/i18n/unicode/numberformatter.h
markusicu: used from intltest so should export

icu4c/source/i18n/unicode/numberrangeformatter.h
markusicu: not used from tools or tests, no need to export

markusicu:
> There is a pretty long list of differences between what is exported "the old way" and "the new way"

Mihai and I are looking at the diffs. Some of the symbols that are no longer supported are private functions, and as long as they are not called from inline code in the header file, that's fine.

I worry a bit about no longer exporting some of the vtables, for example
```
const icu_78::Locale::`vftable'
const icu_78::DateFormatSymbols::`vftable'
const icu_78::DateIntervalFormat::`vftable'
```
and a bunch of others.
Are they not needed?

On the other hand, if CI checks are happy... :shrug: ?

icu4c/source/i18n/collationtailoring.h
markusicu: I am adding a commit that restores this U_I18N_API.

icu4c/source/i18n/collationtailoring.h
markusicu: I am adding a commit that restores this U_I18N_API.

icu4c/source/i18n/unicode/numberformatter.h
markusicu: I am adding a commit that restores this U_I18N_API.

markusicu (commented):
Added another commit after Mihai and I reviewed the changes in symbol exports.

If this passes CI and no one has ideas for what else to test and what more to change, then I suggest we remove the first commit (the temporary test one), squash the rest, approve, and merge.

icu4c/source/i18n/unicode/numberformatter.h
roubert: > used from intltest so should export

If it was necessary to export this, how could then all the tests on the CI pass?

markusicu:
> Added another commit after Mihai and I reviewed the changes in symbol exports.
> 
> If this passes CI

I broke CI. I should have known better: The three internal classes that I exported again have Locale fields, which must be why Fredrik un-exported them. I changed these to U_I18N_API_CLASS class now.

icu4c/source/i18n/unicode/numberformatter.h
markusicu: We now do have U_I18N_API_CLASS here like Mihai asked.

> If it was necessary to export this, how could then all the tests on the CI pass?

I don't know. Some of this feels like black magic.

I would like stuff to make sense. If code outside the i18n library uses an i18n library class, then we should declare it with U_I18N_API[_CLASS]. I don't know why tools/tests worked successfully on Windows when we didn't export things that they use.

I also don't know why stuff seems to work on Windows when we are sloppy with switching "export" to "import". But we do have the mechanism for doing it in a way that looks proper, and we mostly use it, so I think we should try to be consistent.

The other problem is that we developed these things at a time when we had a greater variety of platforms, including OS/390, OS/400, Solaris, AIX, ... and test there frequently. Now we get infrequent complaints and contributions for some of these platforms.

I am trying to be conservative so that we don't break things that used to work, or else the complaints and fixes will be bigger the next time someone bothers to try a new version of ICU.

markusicu:
It looks like the MSVC builds, and thus CI checks overall, are happy now. So we are back to my suggestion:

> If this passes CI and no one has ideas for what else to test and what more to change, then I suggest we remove the first commit (the temporary test one), squash the rest, approve, and merge.

markusicu:
> If this passes CI and no one has ideas for what else to test and what more to change, then I suggest we remove the first commit (the temporary test one), squash the rest, approve, and merge.

Elsewhere, @mihnita gave me a "thumb up" for this. I have now removed the test commit 540a96af65a683793df3b667a3fe6e06055115c2 so this should be ready for review & approval. Once approved, let's squash, and then merge.

icu4c/source/common/unicode/locid.h
markusicu: I remove the test commit.

markusicu (approved):

mihnita (approved):
Thank you very much for the patience

icu4c/source/i18n/unicode/numberformatter.h
```diff
@@ -1517,9 +1517,10 @@ class U_I18N_API Padder : public UMemory {
     friend class impl::GeneratorHelpers;
 };
 
-// Do not enclose entire MacroProps with #ifndef U_HIDE_INTERNAL_API, needed for a protected field
+// Do not enclose entire MacroProps with #ifndef U_HIDE_INTERNAL_API, needed for a protected field.
+// U_I18N_API because intltest uses it.
 /** @internal */
-struct U_I18N_API MacroProps : public UMemory {
+struct U_I18N_API_CLASS MacroProps : public UMemory {
```
roubert, line 1523: I don't understand why this struct would ever need to be exported: It's fully defined in the header file, from which the compiler gets all of it for use in any compilation unit. What is there that could possibly be in the DLL that would need to be exported from there? My assumption was that it had simply been declared U_I18N_API because of overly eager copy-pasting, not because it had any reason to be. Am I missing something here?

markusicu: Compilers may or may not inline, and this is used from outside the i18n library.

PR #3529, opened 2025-06-20
ICU-23142 Fix tiny header guarder in `icu4c/source/i18n/regexcxt.h`

shyeyian:
Header `icu4c/source/i18n/regexcxt.h` is wrongly guarded by [`RBBIRPT_H`](https://github.com/unicode-org/icu/blob/main/icu4c/source/i18n/regexcst.h#L14). I made it a tiny change to `REGEXCST_H`. 

Recently I'm trying to build `icu` into a [`C++20 module`](https://cppreference.com/w/cpp/language/modules.html), and we need a correct guardian in order to build across multiple module partitions. **Thank you!**

*This is an amazing library an I will star it.*

### DIFF
--- a/icu4c/source/i18n/regexcst.h
+++ b/icu4c/source/i18n/regexcst.h
@@ -11,8 +11,8 @@
 //   and others. All rights reserved.  
 //
 //---------------------------------------------------------------------------------
-#ifndef RBBIRPT_H
-#define RBBIRPT_H
+#ifndef REGEXCST_H
+#define REGEXCST_H
 
 #include "unicode/utypes.h"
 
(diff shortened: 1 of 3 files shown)


icu4c/source/i18n/regexcst.h
```diff
@@ -11,8 +11,8 @@
 //   and others. All rights reserved.  
 //
 //---------------------------------------------------------------------------------
-#ifndef RBBIRPT_H
```
aheninger, line 14: This (regexcst.h) is a generated file. The fix needs to be made in the generation script, regexcst.pl, and then this file regenerated, to make sure nothing else has gone wrong over time.

aheninger (changes requested):
Good catches. Thanks.

icu4c/source/i18n/regexcst.h
shyeyian: I got it, thanks for clarifying.

icu4c/source/i18n/regexcst.pl
```diff
@@ -214,8 +214,8 @@
 print "//   and others. All rights reserved.  \n";
 print "//\n";
 print "//---------------------------------------------------------------------------------\n";
-print "#ifndef RBBIRPT_H\n";
```
shyeyian, line 217: Trace back to the root cause here :)

PR #3537, opened 2025-06-28
ICU-23151 Fix building on Windows using msbuild with clang-cl.

KaylinFinke:
When building the latest release 77.1 on Windows from the command line I encountered two issues when using Clang-CL as a compiler (cl compiled fine). I have described them in the ticket, but to reiterate the first is a failure to generate sbr files in debug builds and the second is nullptr passed to strcmp when building data in x64 builds. 

For the sbr files, I have removed browse information from the solution. If this is not an acceptable solution I am happy to revisit it, but my understanding is browse information isn't needed anymore per reading this documentation https://learn.microsoft.com/en-us/cpp/build/reference/creating-an-dot-sbr-file and the option could be fairly safely removed.

The second issue I fixed by piggy backing on the existing dll architecture option. This appears to currently specify an architecture for a building a data file instead of letting the system auto-detect it and seemed harmless to always perform. I Updated the code to store off this architecture, and pass it through to writeObjectCode allowing Clang-CL builds to generate the appropriate data file. I updated the data generation code to handle ARM in addition since that option was missing and couldn't be worse than the previous undefined behavior, and preserved support for lowercase options used by --cpu-arch with genccode. I did not add a new option as -c was already used, and having two different ways to specify the architecture seemed more bug prone, but I am open to suggestions.

I compiled the project using the 8 variants of: Win32/x64, Debug/Release, and MSVC/Clang-CL as well as ran the tests for all these configurations which passed. I also compiled all 8 of these variants to ensure Win32 Release Clang-CL as well as all 4 MSVC configurations passed without the change, while the other 3 Clang-CL variants which failed to build.

I was not able to test ARM / ARM64 though invocations of the tool did not change in the makedata.mak scripts for these platforms. My reason for adding the ARM variant is it can't be worse than the previous code here which if reached would have had the same issue with strcmp.

#### Checklist
- [x] Required: Issue filed: ICU-23151

### DIFF
--- a/icu4c/source/data/makedata.mak
+++ b/icu4c/source/data/makedata.mak
@@ -243,6 +243,9 @@ COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -u
 !IF "$(CFG)" == "x64\Release" || "$(CFG)" == "x64\Debug"
 COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -a X64
 !ENDIF
+!IF "$(CFG)" == "Release" || "$(CFG)" == "release" || "$(CFG)" == "Debug" || "$(CFG)" == "debug" || "$(CFG)" == "x86\Release" || "$(CFG)" == "x86\Debug"
+COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -a X86
+!ENDIF
 !IF "$(CFG)" == "ARM\Release" || "$(CFG)" == "ARM\Debug"
 COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -a ARM
 !ENDIF
--- a/icu4c/source/tools/toolutil/pkg_genc.cpp
+++ b/icu4c/source/tools/toolutil/pkg_genc.cpp
@@ -299,7 +299,8 @@ checkAssemblyHeaderName(const char* optAssembly) {
 
 U_CAPI UBool U_EXPORT2
 checkCpuArchitecture(const char* optCpuArch) {
-    return strcmp(optCpuArch, "x64") == 0 || strcmp(optCpuArch, "x86") == 0 || strcmp(optCpuArch, "arm64") == 0;
+    return strcmp(optCpuArch, "x64") == 0 || strcmp(optCpuArch, "x86") == 0 || strcmp(optCpuArch, "arm64") == 0 || strcmp(optCpuArch, "arm") == 0
+        || strcmp(optCpuArch, "X64") == 0 || strcmp(optCpuArch, "X86") == 0 || strcmp(optCpuArch, "ARM64") == 0 || strcmp(optCpuArch, "ARM") == 0;
 }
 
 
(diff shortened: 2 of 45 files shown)


richgillam (approved):
This all looks reasonable to me, but I haven't touched a Windows machine in 15 years, so I'm really not sure mine should be the final word on this subject...

icu4c/source/data/makedata.mak
```diff
@@ -243,6 +243,9 @@ COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -u
 !IF "$(CFG)" == "x64\Release" || "$(CFG)" == "x64\Debug"
 COMMON_ICUDATA_ARGUMENTS=$(COMMON_ICUDATA_ARGUMENTS) -a X64
 !ENDIF
+!IF "$(CFG)" == "Release" || "$(CFG)" == "release" || "$(CFG)" == "Debug" || "$(CFG)" == "debug" || "$(CFG)" == "x86\Release" || "$(CFG)" == "x86\Debug"
```
roubert, line 246: Why is this statement here testing `CFG` for both with and without titlecase, when the other such statements before and after don't do that?

icu4c/source/tools/toolutil/pkg_genc.cpp
```diff
@@ -299,7 +299,8 @@ checkAssemblyHeaderName(const char* optAssembly) {
 
 U_CAPI UBool U_EXPORT2
 checkCpuArchitecture(const char* optCpuArch) {
-    return strcmp(optCpuArch, "x64") == 0 || strcmp(optCpuArch, "x86") == 0 || strcmp(optCpuArch, "arm64") == 0;
+    return strcmp(optCpuArch, "x64") == 0 || strcmp(optCpuArch, "x86") == 0 || strcmp(optCpuArch, "arm64") == 0 || strcmp(optCpuArch, "arm") == 0
```
roubert, line 302: Where does this uppercase / lowercase difference come from?

roubert:
> This all looks reasonable to me, but I haven't touched a Windows machine in 15 years, so I'm really not sure mine should be the final word on this subject...

Pretty much the same for me there.

@StefanStojanovic & @rp9-next, you're both @microsoft and were the author and approver of PR #3023 which added the Windows ClangCL support in the first place, could you please be so kind and take a look at this?

icu4c/source/data/makedata.mak
KaylinFinke: Consistency with icu4c/source/extra/uconv/makedata.mak line 20. I am happy to remove them if desired.

icu4c/source/tools/toolutil/pkg_genc.cpp
KaylinFinke: the WIN_DLL_ARCH option to pkgdata is an existing option specified in uppercase used to specify the target platform to the linker [machine](https://learn.microsoft.com/en-us/cpp/build/reference/machine-specify-target-platform?view=msvc-170) The options added in the resolution of ICU-22787 added lowercase options for genccode and this optCpuArch option used with writeObjectCode. In the interest of not specifying the architecture twice for what is essentially the same purpose (target platform) once as uppercase and once as lowercase I am passing the value of WIN_DLL_ARCH to optCpuArch instead of making a new option. In the interest of not breaking external code relying on the lowercase parameters added in ICU-22787 I decided to support both the longstanding WIN_DLL_ARCH format and the new format added in ICU-22787.


I didn't think this solution was ideal -- but duplicating options isn't either, nor is breaking existing users. I'm open to other approaches if you would prefer a different one.

pandeyrah-msft:
tagging @mamtadhaka to take a look

KaylinFinke:
Please let me know if there's any way I can assist in moving this forward. I am happy to meet with people to discuss it if that's helpful.
-Kaylin

Mamtadhaka:
> tagging @Mamtadhaka to take a look

I do not have the permissions to approve the PR, 
I have gone through the changes, and looks good to me, @KaylinFinke can you please confirm if the changes are tested for windows.

KaylinFinke:
> > tagging @Mamtadhaka to take a look
> 
> I do not have the permissions to approve the PR, I have gone through the changes, and looks good to me, @KaylinFinke can you please confirm if the changes are tested for windows.

Yes. This is documented in the ticket including compilation steps to perform the tests I did. I compiled Win32 and x64 Debug and Release Clang-CL and CL using Windows 11 (26100) using Visual Studio 2022 17.14.7 Preview 1.0 and ran the associated tests as outlined in the documentation here https://unicode-org.github.io/icu/userguide/icu4c/build.html#running-the-tests-from-the-windows-command-line-cmd .

I did not test ARM nor ARM64. I do not have access to these platforms.

PR #3546, opened 2025-07-03
ICU-23155 Fix buffer overflow in SSearchTest::offsetTest()

gvictor:
https://unicode-org.atlassian.net/browse/ICU-23155

Use std::string instead.

Root cause analysis:
In printOffsets, while the pointer `s` is advanced, the remaining size `n` remains unchanged to be a constant 4096, and snprintf falsely thinks the buffer has 4096 bytes left.

#### Checklist
- [X] Required: Issue filed: ICU-23155

### DIFF
--- a/icu4c/source/test/intltest/ssearch.cpp
+++ b/icu4c/source/test/intltest/ssearch.cpp
@@ -28,6 +28,9 @@
 #include "ssearch.h"
 #include "xmlparser.h"
 
+#include <iomanip>
+#include <iostream>
+#include <sstream>
 #include <stdio.h>  // for snprintf
 
 char testId[100];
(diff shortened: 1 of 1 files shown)


gvictor:
@markusicu can you help review this test change?

icu4c/source/test/intltest/ssearch.cpp
```diff
@@ -524,40 +527,48 @@ UBool OrderList::matchesAt(int32_t offset, const OrderList &other) const
     return true;
 }
 
-static char *printOffsets(char *buffer, size_t n, OrderList &list)
+static char* printOffsets(std::string &buffer, OrderList &list)
```
roubert, line 530: This would all become considerably simpler if the caller wasn't required to pass in any buffer at all and the function simply returned an `std::string` object by value. That's what any normal C++ code would do nowadays.

icu4c/source/test/intltest/ssearch.cpp
```diff
@@ -524,40 +527,48 @@ UBool OrderList::matchesAt(int32_t offset, const OrderList &other) const
...
 {
     int32_t size = list.size();
-    char *s = buffer;
+    buffer.clear();
+    buffer.reserve(size * 10 + 2);
```
roubert, line 534: Is this some kind of performance optimization, to avoid having to dynamically resize the buffer? If so, I suggest removing it, this is test code, the possible performance gain is irrelevant compared to having a few extra lines of code to maintain.

roubert (changes requested):

markusicu:
@gvictor please take a look at the outstanding feedback

gvictor:
Looks like that the PR has bee submitted to fix this bug.

https://github.com/unicode-org/icu/pull/4153

PR #3550, opened 2025-07-08
ICU-23140 Use covariant return types in clone() in ICU4J

mihnita:
#### Checklist
- [x] Required: Issue filed: ICU-23140


mihnita:
I tried to split it in smaller commits.
I had to do it by area, with several classes at once, since certain things depended on each other.
With a fix at the end, because not everything was perfect from the first try :-)

icu4j/main/core/src/main/java/com/ibm/icu/text/TimeZoneFormat.java
```diff
@@ -3301,5 +3301,14 @@ public TimeZoneFormat cloneAsThawed() {
         copy._frozen = false;
         return copy;
     }
+
+    /**
+     * {@inheritDoc}
+     * @draft ICU 78
```
markusicu: I think we should mark this as `@stable ICU 49` since it has existed since then and is only really changing return type now, for which we don't change the other clone()s to draft either.

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -168,7 +168,7 @@ public Object clone()  {
         RuleBasedBreakIterator result;
         result = (RuleBasedBreakIterator)super.clone();
         if (fText != null) {
-            result.fText = (CharacterIterator)(fText.clone());
+            result.fText = (CharacterIterator)fText.clone();
```
markusicu, line 171: do we need this cast?

icu4j/main/core/src/main/java/com/ibm/icu/util/GregorianCalendar.java
```diff
@@ -924,4 +924,9 @@ public String factoryName() {
         return factory;
     }
     */
+
+    @Override
```
markusicu, line 932: please add javadoc, stable ICU 2.0

markusicu (commented):
nice!

icu4j/main/core/src/main/java/com/ibm/icu/util/GregorianCalendar.java
mihnita: Done

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
mihnita: Yes, that is `java.text.CharacterIterator`, and the clone() returns an `Object`.
And I can't change that :-)

icu4j/main/core/src/main/java/com/ibm/icu/text/TimeZoneFormat.java
mihnita: Done

markusicu (approved):

PR #3554, opened 2025-07-10
ICU-12717 Generate errorprone reports as markup and tab-separated-values

mihnita:
Generating tab-separate-values for easy import in a spreadsheet, and markup to use as reporting from GitHub actions.

#### Checklist
- [x] Required: Issue filed: ICU-12717

### DIFF
--- a/icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
+++ b/icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
@@ -11,8 +11,9 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -20,10 +21,13 @@ import java.util.Map;
 class ErrorProneReport {
     private static final String HTML_REPORT_FILE = "errorprone1.html";
     private static final String HTML_REPORT_FILE2 = "errorprone2.html";
+    private static final String TSV_REPORT_FILE = "errorprone.tsv";
+    private static final String MD_REPORT_FILE = "errorprone.md";
     private static final String SORTTABLE_JS_FILE = "sorttable.js";
     private static final String SORTTABLE_CSS_FILE = "errorprone.css";
     private static final String [] EMBEDDED_FILES =
         { SORTTABLE_JS_FILE, SORTTABLE_CSS_FILE };
+    private static final String MD_CHARS_TO_ESCAPE = "\\*_|#`[]{}()<>+-.!";
 
     public static void genReports(String icuDir, String mavenStdOut, String outDir, String baseUrl)
             throws IOException {
@@ -32,8 +36,10 @@ class ErrorProneReport {
                 ParseMavenOutForErrorProne.parse(icuDir, mavenStdOut);
 
         extractExtraFiles(outDir);
-        genReport1(errors, outDir, baseUrl);
-        genReport2(errors, outDir, baseUrl);
+        genReportHtml1(errors, outDir, baseUrl);
+        genReportHtml2(errors, outDir, baseUrl);
+        genReportTsv(errors, outDir);
+        genReportMd(errors, outDir);
     }
 
     // Extract additional files used by the reports (css, js, etc)
@@ -45,7 +51,7 @@ class ErrorProneReport {
         }
     }
 
-    private static void genReport1(Map<String, List<ErrorProneEntry>> errors,
+    private static void genReportHtml1(Map<String, List<ErrorProneEntry>> errors,
             String outDir, String baseUrl) throws IOException {
 
         Path outFileName = Paths.get(outDir, HTML_REPORT_FILE);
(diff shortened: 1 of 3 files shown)


icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
```diff
@@ -266,4 +268,107 @@ private static void outSummary(HtmlUtils hu, Map<String, List<ErrorProneEntry>>
...
+    private static void genReportTsv(Map<String, List<ErrorProneEntry>> errors,
+            String outDir) throws IOException {
+        Path outFileName = Paths.get(outDir, TSV_REPORT_FILE);
+        System.out.println("Report generated: " + outFileName);
+        try (PrintStream wrt = new PrintStream(outFileName.toString(), StandardCharsets.UTF_8)) {
+            wrt.println("Issue type"
+                    + "\tCount"
+                    + "\tSeverity"
+                    + "\tURL"
+            );
```
markusicu, line 281: optional: you could of course just write one string "Issue type\tCount\t..."

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
```diff
@@ -266,4 +268,107 @@ private static void outSummary(HtmlUtils hu, Map<String, List<ErrorProneEntry>>
...
+            for (Map.Entry<String, List<ErrorProneEntry>> e : errors.entrySet()) {
+                String errorType = e.getKey();
+                List<ErrorProneEntry> errorsOfType = e.getValue();
+                if (errorsOfType.isEmpty()) {
+                    continue;
+                }
```
markusicu, line 287: optional: each of the genReport... methods is doing some of the same stuff. could we simplify? for example, filter empty entries from the input so that the methods need not check for empty?

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
```diff
@@ -266,4 +268,107 @@ private static void outSummary(HtmlUtils hu, Map<String, List<ErrorProneEntry>>
...
+
+                for (ErrorProneEntry error : e.getValue()) {
+                    wrt.print("| ");
+                    wrt.print("[`" + errorType + "`](" + error.url + ")");
+                    String tags = ErrorProneUtils.getTags(errorType);
```
markusicu, line 323: optional: you can pull this out of the loop, together with computing either "" or `" " + escapeMd(tags)` --> just unconditionally append the resulting string here

markusicu (approved):

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
mihnita: The reason for that was to make it more readable and easy to keep in sync with the values, a bit below:

```java
wrt.println("Issue type"
        + "\tCount"
        + "\tSeverity"
        + "\tURL"
);
...
wrt.println(errorType
        + "\t" + errorsOfType.size()
        + "\t" + errorSeverity
        + "\t" + firstEntry.url
);
```

So if you add another column it is easier.

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
mihnita: Done

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
mihnita: The `isEmpty` part was just excess of care :-)
It should never happen.

I removed, since in the end this is not production code.
Even if it happens (how?) probably the fix should be somewhere else.
Let's fail, and fix it if it does.

---

In general the 4 reports are not that similar. There is not much to refactor.

Maybe the tab-separate-file is like the summary part of errorprone2.html ...
And the .md is similar to errorprone1.html

I actually considered using a templating system, like Freemarker (https://freemarker.apache.org/)
But that's another dependency, and one would need to learn that templating system in order to change anything.
So I decided against.

mihnita: BTW, if it helps here are the reports:
* https://mihai-nita.net/tmp/ep/errorprone1.html
* https://mihai-nita.net/tmp/ep/errorprone2.html
* https://mihai-nita.net/tmp/ep/errorprone.md
* https://mihai-nita.net/tmp/ep/errorprone.tsv

mihnita:
Thank you!
M

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
markusicu: > The `isEmpty` part was just excess of care :-) It should never happen.
> 
> I removed, since in the end this is not production code. Even if it happens (how?) probably the fix should be somewhere else. Let's fail, and fix it if it does.

If you did remove this check, then I guess you have a pending change on your local disk which didn't make it into the main merge.

> In general the 4 reports are not that similar. There is not much to refactor.

ok

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/errorprone/ErrorProneReport.java
markusicu: pending local change?

mihnita: Yes :-(
I merged before pushing the local. :-(

PR #3563, opened 2025-07-15
ICU-23054 Fix all 'ReferenceEquality' errorprone issues

mihnita:
#### Checklist

### DIFF
--- a/icu4j/main/collate/src/main/java/com/ibm/icu/text/RuleBasedCollator.java
+++ b/icu4j/main/collate/src/main/java/com/ibm/icu/text/RuleBasedCollator.java
@@ -461,6 +461,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #setUpperCaseFirst(boolean)
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public final void setCaseFirstDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -478,6 +479,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #isAlternateHandlingShifted()
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setAlternateHandlingDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -495,6 +497,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #isCaseLevel
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setCaseLevelDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -512,6 +515,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #setDecomposition(int)
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setDecompositionDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -529,6 +533,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #setFrenchCollation(boolean)
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setFrenchCollationDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -546,6 +551,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #getStrength
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setStrengthDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -562,6 +568,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @see #setNumericCollation
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
     public void setNumericCollationDefault() {
         checkNotFrozen();
         CollationSettings defaultSettings = getDefaultSettings();
@@ -737,6 +744,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @stable ICU 53
      */
     @Override
+    @SuppressWarnings("ReferenceEquality")
     public RuleBasedCollator setMaxVariable(int group) {
         // Convert the reorder code into a MaxVariable number, or UCOL_DEFAULT=-1.
         int value;
@@ -917,6 +925,7 @@ public final class RuleBasedCollator extends Collator implements Cloneable {
      * @stable ICU 4.8
      */
     @Override
+    @SuppressWarnings("ReferenceEquality")
     public void setReorderCodes(int... order) {
         checkNotFrozen();
         int length = (order != null) ? order.length : 0;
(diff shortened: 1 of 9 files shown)


mihnita:
I am not that happy with the tons of `@SuppressWarnings("ReferenceEquality")` that I had to use.

But it looks like in quite a few places we put something somewhere (setter, or store in a map) and then expect to get the exact same thing back.

Which might be intentional, or accidental, I don't know.

Works, for now. But prevents certain types of refactoring, for example doing a clone in the setter, or in the getter.

I would not expect the exact same value... but I did not want to change behavior.
So I tried to use some best judgement. And ended up suppressing the warning in many cases.
But at least we can enable checking for it in CI and we don't introduce more.
And if in the future we want to address (some of) this we can grep.

OR
I change most of these `==` into `.equals`, and we don't care that if you get what you set or not.

mihnita:
Please hold please.
I trying to reduce some of the @SuppressWarnings("ReferenceEquality") instances.

mihnita:
Just in case: this is unblocked for review.

I did everything I could, all remaining `@SuppressWarnings("ReferenceEquality")` instances are there to stay.

icu4j/main/collate/src/main/java/com/ibm/icu/text/RuleBasedCollator.java
```diff
@@ -461,6 +461,7 @@ public void setLowerCaseFirst(boolean lowerfirst) {
      * @see #setUpperCaseFirst(boolean)
      * @stable ICU 2.8
      */
+    @SuppressWarnings("ReferenceEquality")
```
markusicu, line 464: FYI
These `if(settings.readOnly() == defaultSettings)` are deliberate and necessary.

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/CollationAPITest.java
```diff
@@ -530,8 +530,7 @@ public void TestJunkCollator(){
 
         String colrules = ((RuleBasedCollator)col).getRules();
         String junkrules = ((RuleBasedCollator)junk).getRules();
-        doAssert(colrules == junkrules || colrules.equals(junkrules),
-                   "The default collation should be returned.");
+        doAssert(colrules.equals(junkrules), "The default collation should be returned.");
```
markusicu: FYI -- looks like that was an optimization which hopefully lives in the .equals() impl as well anyway

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/UCAConformanceTest.java
```diff
@@ -181,6 +181,7 @@ private static int normalizeResult(int result) {
         return result < 0 ? -1 : result == 0 ? 0 : 1;
     }
 
+    @SuppressWarnings("ReferenceEquality")
```
markusicu, line 184: still needed after changing to .equals()?

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/ime/indic/IndicInputMethodImpl.java
```diff
@@ -373,7 +373,7 @@ public int getRunStart() {
 
         public int getRunStart(AttributedCharacterIterator.Attribute attribute) {
             return (index >= committed &&
-                attribute == TextAttribute.INPUT_METHOD_UNDERLINE) ? committed : 0;
+                attribute.equals(TextAttribute.INPUT_METHOD_UNDERLINE)) ? committed : 0;
```
markusicu, line 376: - This is in .../tools/misc/.../dev/tool/ime/... so probably no one uses it.
- .equals() should be fine
- However, it is a little weird having to use .equals() for a `static final` constant -- an immutable singleton.
  https://docs.oracle.com/en/java/javase/11/docs/api/java.desktop/java/awt/font/TextAttribute.html#INPUT_METHOD_UNDERLINE

icu4j/main/collate/src/main/java/com/ibm/icu/text/RuleBasedCollator.java
mihnita: Yes, I figured, that's why I left them in place.

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/CollationAPITest.java
mihnita: From https://errorprone.info/bugpattern/ReferenceEquality:

* Most .equals (should) have a "fast path"
  ```java
  @Override
  public boolean equals(Object other) {
    if (this == other) {
      return true; // fast path, reference equality is allowed here
    }
    ...
  ```

Optional, I can replace with `assertSame(b, a)`
```java
assertSame(colrules, junkrules); // this uses ==
assertEquals(colrules, junkrules); // this uses .equals
```

Although in the original `if` it is an OR, and what I propose is an AND.

But basically having it an OR is too loose.
Even if `colrules != junkrules` the `colrules.equals(junkrules)` might succeed.
So my testing is not that good, I think.

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/UCAConformanceTest.java
mihnita: Yes, because in line 271 we have this:
```java
if (oldSk == sk1) {
    newSk = sk2;
} else {
    newSk = sk1;
}
```
Not a fun, but...

One of the things I don't like about `@SuppressWarnings` is that (in most cases) you can't just dump it in the middle of a block of code.
So you put it at the top of a method, to counter something that happens 100 lines down.

And replacing it with `.equals` fails.
I can debug it to understand why.
But it kind of shows why this kind of behavior is considered "FragileCode" (errorprone classification).

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/ime/indic/IndicInputMethodImpl.java
mihnita: I can "sprinkle" `@SuppressWarnings("ReferenceEquality")`

Even if it is safe to use `==`, that is not good enough for the linter.
It is not smart enough to "know" that.
All I can do is use `.equals`, or disable the warning.
You know how linters work :-)
If we want to enable this check in CI (and fail if new instances are introduced) then this must be "solved" (even if we know it is not a problem).

This being some (very much outdated) GUI tool (which most of the time waits for the user), I thought the performance is not a problem.

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/CollationAPITest.java
mihnita: I've replaced it with `assertSame(colrules, junkrules)` only.

There is no point to also test `assertEquals`, these are strings, and we know that `equals` is implemented correctly.
If `a == b` then `a.equals(b)` and `b.equals(a)`.
If that would fail for `String` it would be hilarious :-)

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/ime/indic/IndicInputMethodImpl.java
mihnita: TLDR: I don't feel strongly about `SuppressWarnings` or `.equals`. Coin toss.

mihnita:
Thank you for the review.

Please round two, but not outside business hours :-)

icu4j/main/collate/src/test/java/com/ibm/icu/dev/test/collator/CollationAPITest.java
markusicu: I actually thought that "equals" was fine here, just adding a comment that helped justify it. But if "same" works, that's ok too.

mihnita:
Thank you!
M

PR #3571, opened 2025-07-21
ICU-23076 Add gmtime null handler

wooffie:
#### Checklist
- [x] Required: Issue filed: ICU-23076
- [x] Required: The PR title must be prefixed with a JIRA Issue number. Example: "ICU-23076 Add gmtime null handler"
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. Example: "ICU-23076 Add gmtime null handler"


wooffie:
Use `exit` because some functions in this file can call it too. Maybe we can use "default" year, but better to say user about problem in his system

PR #184, opened 2018-10-01
ICU-20186 Adding test for semicolon spacing in DecimalFormat.

sffc:
##### Checklist

- [x] Issue filed at https://unicode-org.atlassian.net :  ICU-20186
- [x] Update PR title to include Issue number
- [ ] Issue accepted
- [x] Tests included
- [ ] Documentation is changed or added

PR #3572, opened 2025-07-22
ICU-23004 add ...CodePoints default constructors

markusicu:
add [Unsafe]UTFStringCodePoints default constructors

#### Checklist
- [x] Required: Issue filed: ICU-23004

### DIFF
--- a/icu4c/source/common/unicode/utfiterator.h
+++ b/icu4c/source/common/unicode/utfiterator.h
@@ -1753,6 +1753,12 @@ class UTFStringCodePoints {
     static_assert(sizeof(CP32) == 4, "CP32 must be a 32-bit type to hold a code point");
 public:
     /**
+     * Constructs an empty C++ "range" object.
+     * @draft ICU 78
+     */
+    UTFStringCodePoints() = default;
+
+    /**
      * Constructs a C++ "range" object over the code points in the string.
      * @param unitRange input range
      * @draft ICU 78
@@ -2463,6 +2469,12 @@ template<typename CP32, typename Range>
 class UnsafeUTFStringCodePoints {
     static_assert(sizeof(CP32) == 4, "CP32 must be a 32-bit type to hold a code point");
 public:
+    /**
+     * Constructs an empty C++ "range" object.
+     * @draft ICU 78
+     */
+    UnsafeUTFStringCodePoints() = default;
+
     /**
      * Constructs a C++ "range" object over the code points in the string.
      * @param unitRange input range
--- a/icu4c/source/test/intltest/utfiteratortest.cpp
+++ b/icu4c/source/test/intltest/utfiteratortest.cpp
@@ -349,6 +349,7 @@ public:
         TESTCASE_AUTO(testUnsafe32ZigzagReverse);
 
         TESTCASE_AUTO(testOwnership);
+        TESTCASE_AUTO(testCPDefaultConstructors);
 
         // C++20 ranges with all 2021 defect reports.  There is no separate
         // feature test macro value for https://wg21.link/P2210R2, but 2021'10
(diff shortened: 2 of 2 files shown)


icu4c/source/common/unicode/utfiterator.h
```diff
@@ -2463,6 +2472,15 @@ template<typename CP32, typename Range>
 class UnsafeUTFStringCodePoints {
     static_assert(sizeof(CP32) == 4, "CP32 must be a 32-bit type to hold a code point");
 public:
+    /** C++ iterator boilerplate @internal */
+    using iterator = UnsafeUTFIterator<CP32, decltype(std::declval<Range>().begin())>;
```
jmr: IMHO, clearer to replace `auto begin()`/`end()` with `iterator`.

markusicu:
I just added a second commit: I realized that for the validating UTFStringCodePoints I was defaulting UTFIterator's tparam LimitIter to its UnitIter. In order to make it work in general, we need to match LimitIter with what code unit range .end() returns.

icu4c/source/common/unicode/utfiterator.h
markusicu: I think we were intentionally vague so that the type could change, as long as calling code can be reasonably stable.

Also, end() may not return the same type as begin() -- C++20 merely requires that the iterator is _comparable_ to end() -- so at most begin() could possibly return "iterator". It all depends on the input code unit range.

icu4c/source/common/unicode/utfiterator.h
```diff
@@ -2463,6 +2475,15 @@ template<typename CP32, typename Range>
...
```
jmr: Can we also get `const_iterator`?  

Reverse iterators typedefs would be nice as well.

icu4c/source/common/unicode/utfiterator.h
jmr: > I think we were intentionally vague so that the type could change, as long as calling code can be reasonably stable.

You are now guaranteeing that `begin()` returns `iterator`, now, right?  Then I would make that clearer by declaring it with `iterator` and not `auto`.

I also took 'C++ "range"' to mean "something like a range", not "the range concept".  Is a more precise wording possible?

eggrobin: > You are now guaranteeing that `begin()` returns `iterator`, now, right?

I don’t think we are (I don’t think this is always the case depending on constness of `this`).
Note that `iterator` is not part of the stable API (it is marked `@internal`).
And it is certainly not the case that `end()` returns iterators, because of sentinels in C++20.

And yes, the intent is to implement the `range` concept, the comments should probably say so.

icu4c/source/common/unicode/utfiterator.h
eggrobin: This is designed to resemble a standard library iterator adaptor, not a standard library container, so we did not provide things that the iterator adaptors do not provide. Note that even _iterator_ is exposition-only in https://eel.is/c++draft/range.filter.

For reverse iterators in particular, `std::reverse_iterator` works everywhere (and it is what std::views::reverse uses).

The `iterator` alias makes some sense for users who are stuck in C++17, since they don’t get `std::ranges::iterator_t`.
I guess we could provide `const_iterator` similarly for users who don’t have the C++23 `std::ranges::const_iterator_t`.

icu4c/source/common/unicode/utfiterator.h
jmr: > > You are now guaranteeing that `begin()` returns `iterator`, now, right?
> 
> I don’t think we are (I don’t think this is always the case depending on constness of `this`). 

Then I don't see how `iterator` can be used if it's not guaranteed to be what `begin()` returns.

I was hoping to turn
```
using CodePoints = icu::header::UnsafeUTFStringCodePoints<char32_t, char>;
using CodePointsIterator = icu::header::UnsafeUTFIterator<char32_t, std::string_view::iterator>;
...
CodePointsIterator it = unsafeUTFStringCodePoints<char32_t>(...).begin();
```
into
```
using CodePoints = icu::header::UnsafeUTFStringCodePoints<char32_t, char>;
using CodePointsIterator = CodePoints::iterator;
...
```
(or eliminate `CodePointsIterator` completely).

So it seems my choices are leave it the old way or use `decltype()`.  Do you guarantee the old way keeps working?

eggrobin: > So it seems my choices are leave it the old way or use `decltype()`

(Or `auto it = `…, but I gather from discussions with Markus that you want to use that as a member of some struct?)

I think it makes sense to add type aliases `CodePoints::iterator` and `CodePoints::const_iterator` that are equivalent to `std::ranges::iterator_t<CodePoints>` and `std::ranges::const_iterator_t<CodePoints>` for those users who are stuck in C++17 and C++20 respectively and thus don’t have those two alias templates in their standard library. And then `begin` will return those by definition. (As noted in the other thread, for reverse_iterator, you can use std::reverse_iterator, so no need for a member alias here.)

I think those can be made part of the stable API (not `@internal`). But that needs to get reviewed in ICU-TC (on the 31st at the earliest).

Markus and I are busy with [UTC #184](https://www.unicode.org/timesens/logistics-utc184.html) this week, so we will probably get back to this next week.

markusicu: `using iterator = ...` is a lot more complicated to get right than I thought.
I think I should _remove_ these typedefs from this PR, and only add the default constructors now, which is what the PR title says after all. Then we can revisit "iterator" without time pressure.

eggrobin: SGTM. But at least now I have convinced myself that `using iterator = […]` (done right) is a good idea, so there is some progress here!

jmr: Removing `iterator` from this PR SGTM.

markusicu: removed `using iterator` -- now just adding default constructors

markusicu:
@richgillam you had previously approved this PR. I have since removed the `using iterator` stuff which needs more thought. Could you please re-approve?

Or maybe @eggrobin could stamp it?

richgillam (approved):
Iterator removal looks fine.  Already approved everything else.

richgillam:
> @richgillam you had previously approved this PR. I have since removed the `using iterator` stuff which needs more thought. Could you please re-approve?
> 
> Or maybe @eggrobin could stamp it?

Change looks fine.  Approved again.

PR #3575, opened 2025-07-23
ICU-23054 Fix all 'AlmostJavadoc' errorprone issues (for real this time)

roubert:
Follow-up to PR #3566 after PR #3573 made Error Prone actually find and report all warnings.

#### Checklist


mihnita (approved):
Thank you.

Although some of these seem quite silly, to the point where I am questioning the value of this check.
Too many false positives?
Can't I use block comments now?

I would probably argue that if a block comment does not contain any @tag, and it is not "attached" to a public API / member, then it is very likely NOT an "almost javadoc"

roubert:
> Too many false positives?

I think so, some of those really didn't look like Javadoc at all.

PR #3581, opened 2025-08-04
ICU-23142 In i18n/regexcst.h fix the header guard

aheninger:
Correct the bad header guard in the file regexcst.h, which contains the state table used for parsing regular expressions.

Because regexcst.h is a generated file, the fix needs to be made in the generation script, regexcst.pl, and the file regenerated.

Regeneration has, unfortunately, produced large diffs because much of the generated content depends on the order of enumeration of hashes in the script, which has always been undetermined, and with recent versions of Perl, is deliberately randomized.

In the end, this PR makes the following changes to the generation script:
- Produce the correct header guard.
- Enumerate hashes in sorted order, to avoid large diffs in the future.
- Use nullptr instead of 0 (a change previously made directly in the generated file)
- Update messages and names that incorrectly referred to RBBI rather than regex

#### Checklist
- [x] Required: Issue filed: ICU-23142


markusicu (approved):
thank you!

PR #3587, opened 2025-08-07
ICU-23114 Fix DateTimePatternGenerator numeric/alphabetic substitution

sffc:
This tweaks the DTPG behavior according to the ticket.

#### Checklist
- [x] Required: Issue filed: ICU-23114

### DIFF
--- a/icu4c/source/test/intltest/dtptngts.cpp
+++ b/icu4c/source/test/intltest/dtptngts.cpp
@@ -49,6 +49,7 @@ void IntlTestDateTimePatternGeneratorAPI::runIndexedTest( int32_t index, UBool e
         TESTCASE(13, testDateTimePatterns);
         TESTCASE(14, testISO8601);
         TESTCASE(15, testRegionOverride);
+        TESTCASE(16, testAlphabeticSubstitution23114);
         default: name = ""; break;
     }
 }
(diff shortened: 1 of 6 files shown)


markusicu (approved):

icu4c/source/test/intltest/dtptngts.cpp
```diff
@@ -1822,4 +1823,27 @@ void IntlTestDateTimePatternGeneratorAPI::doDTPatternTest(DateTimePatternGenerat
     }
 }
 
+void IntlTestDateTimePatternGeneratorAPI::testAlphabeticSubstitution23114() {
+    IcuTestErrorCode status(*this, "testAlphabeticSubstitution23114");
+
+    LocalPointer dtpg(DateTimePatternGenerator::createEmptyInstance(status), status);
```
markusicu:
```suggestion
    LocalPointer<DateTimePatternGenerator> dtpg(
        DateTimePatternGenerator::createEmptyInstance(status), status);
```
to fix
```
dtptngts.cpp:1829:5: error: 'LocalPointer' may not intend to support class template argument deduction [-Werror,-Wctad-maybe-unsupported]
 1829 |     LocalPointer dtpg(DateTimePatternGenerator::createEmptyInstance(status), status);
      |     ^
../../common/unicode/localpointer.h:191:7: note: add a deduction guide to suppress this warning
  191 | class LocalPointer : public LocalPointerBase<T> {
      |       ^
1 error generated.
```

@eggrobin @roubert it looks like LocalPointer would be a lot easier to use if we added a [User-defined deduction guide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction.html)? And that's supported in C++17?

Something like
```
template<class T>
LocalPointer(T *p, UErrorCode &errorCode) -> LocalPointer<remove_ref_something(*p)>;
```
?

Too bad that apparently the compiler can already do this implicitly, but with a warning, and we build with warnings-as-errors :-(
See the [LocalPointer constructor definition](https://github.com/unicode-org/icu/blob/0e20d64b5963f1e31b751d367869385e037be71f/icu4c/source/common/unicode/localpointer.h#L214).

roubert: > @eggrobin @roubert it looks like LocalPointer would be a lot easier to use if we added a [User-defined deduction guide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction.html)?

Well, that'd make `LocalPointer` instantiations a bit shorter, but it'd also hide from anyone reading the code what data type exactly is being stored in there. For a call like `DateTimePatternGenerator::createEmptyInstance()` here it's easy enough to guess with high accuracy that the data type will be `DateTimePatternGenerator`, but it's not always that easy to guess so the shorter declaration would come at some cost (which might not be worth it).

See also the recommendations here about pros and cons:

https://google.github.io/styleguide/cppguide.html#Type_deduction

sffc: It seems like we should allow unqualified `LocalPointer` in situations where we allow `auto`.

I don't intend to block this PR on resolving this question; it is more of a curiosity for the future.

PR #3588, opened 2025-08-08
ICU-23172 part 1: Convert era names from array to map, handle era codes with non-0 start, holes

pedberg-icu:
This is the first part of a multipart change to convert era data from an array to a map (only affects the name data, the era rules data was already a map), and then to allow era codes that do not start from 0, or may potentially have holes in the sequence (as long as the date ranges for the eras are contiguous). This is needed to unblock some changes to era data in CLDR, such as for coptic (eliminating the current first era) and japanese (eliminating eras before Meiji).
- This first part makes the CLDR-to-ICU changes to convert the era name data from array to map, and then implements the code changes to handle the new data in map form (will also support the current array form for compatibility with current data). It was tested by doing the CLDR-to-ICU conversion and generating the new map-format data (see example of the delta for root below), but that updated data is not part of this PR to avoid conflicts with current CLDR-to-ICU integration efforts; it will take effect for the next integration. Note that the public DateFormatSymbols interface for era names still returns a string array that runs from era 0 to the max era code for a calendar, so we fill out such an array from the resource data that may only have a partial set of era codes.
- The second part (in a future PR) will adjust the EraRules code to handle calendars whose era codes no longer start at 0 etc.

Here is the delta for `icu4c/source/data/locales/root.txt` (elided for Japanese eras) that results from the change in CLDR-to-ICU conversion:
```
@@ -1250,7 +1250,7 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "BE",
+                    0{"BE"}
                 }
                 narrow:alias{"/LOCALE/calendar/buddhist/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/buddhist/eras/abbreviated"}
@@ -1664,8 +1664,8 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "ERA0",
-                    "ERA1",
+                    0{"ERA0"}
+                    1{"ERA1"}
                 }
                 narrow:alias{"/LOCALE/calendar/coptic/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/coptic/eras/abbreviated"}
@@ -1740,8 +1740,8 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "ERA0",
-                    "ERA1",
+                    0{"ERA0"}
+                    1{"ERA1"}
                 }
                 narrow:alias{"/LOCALE/calendar/ethiopic/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/ethiopic/eras/abbreviated"}
@@ -1800,7 +1800,7 @@ root{
             dayNames:alias{"/LOCALE/calendar/ethiopic/dayNames"}
             eras{
                 abbreviated{
-                    "ERA0",
+                    0{"ERA0"}
                 }
                 narrow:alias{"/LOCALE/calendar/ethiopic-amete-alem/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/ethiopic-amete-alem/eras/abbreviated"}
@@ -1899,8 +1899,8 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "ERA0",
-                    "ERA1",
+                    0{"ERA0"}
+                    1{"ERA1"}
                 }
                 narrow:alias{"/LOCALE/calendar/generic/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/generic/eras/abbreviated"}
@@ -2217,8 +2217,8 @@ root{
             }
             eras{
                 abbreviated{
-                    "BCE",
-                    "CE",
+                    0{"BCE"}
+                    1{"CE"}
                 }
                 narrow:alias{"/LOCALE/calendar/gregorian/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/gregorian/eras/abbreviated"}
@@ -2437,7 +2437,7 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "AM",
+                    0{"AM"}
                 }
                 narrow:alias{"/LOCALE/calendar/hebrew/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/hebrew/eras/abbreviated"}
@@ -2498,7 +2498,7 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "Saka",
+                    0{"Saka"}
                 }
                 narrow:alias{"/LOCALE/calendar/indian/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/indian/eras/abbreviated"}
@@ -2555,8 +2555,8 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "AH",
-                    "BH",
+                    0{"AH"}
+                    1{"BH"}
                 }
                 narrow:alias{"/LOCALE/calendar/islamic/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/islamic/eras/abbreviated"}
@@ -2925,482 +2925,482 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "Taika (645–650)",
-                    "Hakuchi (650–671)",
-                    "Hakuhō (672–686)",
...
-                    "Keiō (1865–1868)",
-                    "Meiji",
-                    "Taishō",
-                    "Shōwa",
-                    "Heisei",
-                    "Reiwa",
+                    0{"Taika (645–650)"}
+                    1{"Hakuchi (650–671)"}
+                    10{"Tenpyō (729–749)"}
...
+                    2{"Hakuhō (672–686)"}
...
+                    231{"Keiō (1865–1868)"}
+                    232{"Meiji"}
+                    233{"Taishō"}
+                    234{"Shōwa"}
+                    235{"Heisei"}
+                    236{"Reiwa"}
...
+                    98{"Eiman (1165–1166)"}
+                    99{"Nin’an (1166–1169)"}
                 }
                 narrow{
-                    "Taika (645–650)",
-                    "Hakuchi (650–671)",
-                    "Hakuhō (672–686)",
...
-                    "Keiō (1865–1868)",
-                    "M",
-                    "T",
-                    "S",
-                    "H",
-                    "R",
+                    0{"Taika (645–650)"}
+                    1{"Hakuchi (650–671)"}
+                    10{"Tenpyō (729–749)"}
...
+                    2{"Hakuhō (672–686)"}
...
+                    231{"Keiō (1865–1868)"}
+                    232{"M"}
+                    233{"T"}
+                    234{"S"}
+                    235{"H"}
+                    236{"R"}
...
+                    98{"Eiman (1165–1166)"}
+                    99{"Nin’an (1166–1169)"}
                 }
                 wide:alias{"/LOCALE/calendar/japanese/eras/abbreviated"}
             }
@@ -3419,7 +3419,7 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "AP",
+                    0{"AP"}
                 }
                 narrow:alias{"/LOCALE/calendar/persian/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/persian/eras/abbreviated"}
@@ -3476,8 +3476,8 @@ root{
             dayNames:alias{"/LOCALE/calendar/gregorian/dayNames"}
             eras{
                 abbreviated{
-                    "Before R.O.C.",
-                    "R.O.C.",
+                    0{"Before R.O.C."}
+                    1{"R.O.C."}
                 }
                 narrow:alias{"/LOCALE/calendar/roc/eras/abbreviated"}
                 wide:alias{"/LOCALE/calendar/roc/eras/abbreviated"}
```

#### Checklist
- [x] Required: Issue filed: ICU-23172


icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1805,8 +1806,34 @@ protected void processResource(String path, UResource.Key key, UResource.Value v
...
+                    if (currentPath.startsWith("eras")) {
+                        // path is one of eras/wide, eras/abbreviated, eras/narrow
+                        UResource.Table rDataTable = value.getTable();
+                        int dataTableSize = rDataTable.getSize();
+                        TreeMap<Integer, String> eraCodeAndNames = new TreeMap<>();
+                        // Collect the era entries that we do have into map sorted by era code
+                        for (int dataTableIndex = 0; dataTableIndex < dataTableSize; dataTableIndex++) {
+                           rDataTable.getKeyAndValue(dataTableIndex, key, value);
+                           Integer eraCode = Integer.valueOf(key.toString());
+                           String eraName = (value.getType() == ICUResourceBundle.STRING)? value.getString(): "";
```
markusicu:
```suggestion
                           String eraName = (value.getType() == ICUResourceBundle.STRING) ? value.getString() : "";
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1805,8 +1806,34 @@ protected void processResource(String path, UResource.Key key, UResource.Value v
...
+                           eraCodeAndNames.put(eraCode, eraName);
+                        }
+                        // Now allocate an array running from era code 0 to the max era we have data for, then
+                        // fill in the entries we have data for with the real names and fill in empty string for
+                        // the others, and save the array.
+                        int maxEraCode = eraCodeAndNames.lastKey().intValue();
+                        int dataArraySize = maxEraCode + 1; // since it runs from 0 through maxEraCode
+                        String[] dataArray = new String[dataArraySize];
+                        for (int dataArrayIndex = 0; dataArrayIndex < dataArraySize; dataArrayIndex++) {
+                            String eraName = eraCodeAndNames.get(Integer.valueOf(dataArrayIndex));
```
markusicu: let's use auto-boxing
```suggestion
                            String eraName = eraCodeAndNames.get(dataArrayIndex);
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1805,8 +1806,34 @@ protected void processResource(String path, UResource.Key key, UResource.Value v
...
+                            dataArray[dataArrayIndex] = (eraName != null)? eraName: "";
```
markusicu:
```suggestion
                            dataArray[dataArrayIndex] = (eraName != null) ? eraName : "";
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1805,8 +1806,34 @@ protected void processResource(String path, UResource.Key key, UResource.Value v
...
                     arrays.put(currentPath, dataArray);
                 } else if (value.getType() == ICUResourceBundle.TABLE) {
-                    // We are not on a leaf, recursively process the subtable.
-                    processResource(currentPath, key, value);
+                    // We might have an eras table that is replacing an eras leaf array
+                    if (currentPath.startsWith("eras")) {
+                        // path is one of eras/wide, eras/abbreviated, eras/narrow
+                        UResource.Table rDataTable = value.getTable();
+                        int dataTableSize = rDataTable.getSize();
+                        TreeMap<Integer, String> eraCodeAndNames = new TreeMap<>();
```
markusicu: A TreeMap seems heavy for a possibly-sparse mapping from indexes, especially when we never need to lookup by the index key.

Suggestion 1: (I think I like suggestion 2 below best)
a)
```
class IndexedEra {
    int index;
    String name;
}
```

b) Make an `ArrayList<IndexedEra>`, initialized with the size of the table.

c) Read the bundle table into the list. Track the max index on the fly.

d) Allocate the final array, `Arrays.fill(array, "")`, iterate over the list, set `array[entry.index] = entry.name`

----
Suggestion 2:
More basic, probably ports more easily to C++:

```
var list = ArrayList<String>(table.size());
int maxIndex = 0;
for (table key, value) {  // pseudocode
    int index = parse key string;
    if (index > maxIndex) {
        list.ensureCapacity(index + 1);
        maxIndex = index;
    }
    list.set(index, eraName);
}
String[] array = list.toArray(new String[maxIndex + 1]);
for (int i = 0; i < array.length; ++i) {
    if (array[i] == null) {
        array[i] = "";
    }
}
```

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1805,8 +1806,34 @@ protected void processResource(String path, UResource.Key key, UResource.Value v
...
+                           String eraName = (value.getType() == ICUResourceBundle.STRING)? value.getString(): "";
+                           eraCodeAndNames.put(eraCode, eraName);
+                        }
+                        // Now allocate an array running from era code 0 to the max era we have data for, then
+                        // fill in the entries we have data for with the real names and fill in empty string for
+                        // the others, and save the array.
+                        int maxEraCode = eraCodeAndNames.lastKey().intValue();
+                        int dataArraySize = maxEraCode + 1; // since it runs from 0 through maxEraCode
+                        String[] dataArray = new String[dataArraySize];
+                        for (int dataArrayIndex = 0; dataArrayIndex < dataArraySize; dataArrayIndex++) {
```
markusicu: If you stick with the TreeMap, then please iterate over its `Map.Entry`s.

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1803,43 @@ struct CalendarDataSink : public ResourceSink {
                 arraySizes.puti(path, dataArraySize, errorCode);
                 if (U_FAILURE(errorCode)) { return; }
             } else if (value.getType() == URES_TABLE) {
-                // We are not on a leaf, recursively process the subtable.
-                processResource(path, key, value, errorCode);
+                // We might have an eras table that is replacing an eras leaf array
+                if (path.startsWith(kErasTagUChar, UPRV_LENGTHOF(kErasTagUChar))) {
```
markusicu: why not `u"eras"` ??

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1803,43 @@ struct CalendarDataSink : public ResourceSink {
...
+                    // path is one of eras/wide, eras/abbreviated, eras/narrow
+                    ResourceTable rDataTable = value.getTable(errorCode);
+                    int32_t dataTableSize = rDataTable.getSize();
+                    LocalArray<EraCodeAndName> eraCodeAndNames(new EraCodeAndName[dataTableSize], errorCode);
+                    // Collect the era entries that we do have, then sort by era code
```
markusicu: This seems roundabout... please take a look at my "suggestion 2" on the Java code. UVector should work in lieu of Java ArrayList.

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1490,6 +1490,27 @@ const char16_t kFormatTagUChar[] = {
...
     SOLIDUS, LOW_A, LOW_B, LOW_B, LOW_R, LOW_E, LOW_V, LOW_I, LOW_A, LOW_T, LOW_E, LOW_D
 };
+const char16_t kErasTagUChar[] = {
+    LOW_E, LOW_R, LOW_A, LOW_S
+};
+
+struct EraCodeAndName : public UObject {
+    int32_t       eraCode;
+    UnicodeString eraName;
+};
```
richgillam: I'm with Markus on this-- why not just use a string literal, now that that works?  Maybe it's even worth it to convert the other constants.

richgillam (commented):
I'm sorry; I kind of got lost in here, and I think it's because I don't have a good sense of the big picture.

Does ICU use the era numbers at all?  If not, couldn't you change the CLDR-to-ICU tool so that it just spits out what it did before?

I'm guessing we _do_ use the era numbers, that we don't want to have to renumber the eras when we delete some of them, and so you're converting to a table resource on the ICU side to enable this.  I wonder if it'd be simpler to keep it as an array on the ICU side and just leave holes for the deleted eras.  This seems ugly for the Japanese calendar, but not so bad for the others.  If we make it an actual table resource, I like Markus's idea of using an array as the intermediate form to facilitate sorting.

You also said something in here about preserving backward compatibility with the old resource file format.  Why do we need to do that?  It seems like that'd just make everything more complicated.

Sorry if these are stupid questions; I don't think all my brain cells are functioning today...

icu4c/source/i18n/dtfmtsym.cpp
richgillam: That wasn't the line I was trying to mark!  I meant this comment to apply to `kErasTagUChar` just above this.

pedberg-icu:
@richgillam 

> Does ICU use the era numbers at all? If not, couldn't you change the CLDR-to-ICU tool so that it just spits out what it did before?

ICU certainly does use era numbers and they are public, eras are one of the standard calendar fields you get when you convert a UDate (C) or milliseconds (J) to a date in a particular calendar. Certain values have corresponding defined constants, as for 0 (BCE) and 1 (CE) in the Gregorian calendar. 

In addition, DateFormatSymbols in both C and J can get a complete array of all era names of a particular type (wide, abbreviated, narrow) in a particular locale and calendar; this array is index by era code starting at 0 (with this PR, the array will contain empty strings for era codes that are no longer valid).

> I'm guessing we _do_ use the era numbers, that we don't want to have to renumber the eras when we delete some of them, and so you're converting to a table resource on the ICU side to enable this. I wonder if it'd be simpler to keep it as an array on the ICU side and just leave holes for the deleted eras. This seems ugly for the Japanese calendar, but not so bad for the others. If we make it an actual table resource, I like Markus's idea of using an array as the intermediate form to facilitate sorting.

The requirements from the CLDR-ICU design discussion were:
- current era codes continue to have their same meeting: if era 236 is the Japanese calendar is currently Reiwa, it should remain so regardless of whether we delete eras.
- the _valid_ era codes for a given calendar need to all increase monotonically with the time period for the era (that is, era code n+m needs to correspond to a time period greater than era code n), or need to all decrease monotonically; there cannot be chronological reverse in the sequence of era codes.
- the valid era codes need to cover a contiguous sequence of time periods (but the era codes themselves may not be contiguous); this allows us to, say, consolidate two eras into one.
- the _valid_ era codes need not start at 0.

For the public API we do need to return a 0-based array for the era names (with the index corresponding to the era code), but it may contain some empty strings for now-invalid eras. But there is not really a way to generate that in the CLDR-ICU converter from the sparse era data on the CLDR side; it is much easier to have the ICU data resource be sparse, but then generate the full array from that sparse resource data when a particular Calenda is instantiated. This also means a lot less wasted storage in ICU data, particularly when we start deleting the first 232 eras in the Japanese calendar  (we have about 70 instances of localized name sets for the japanese calendar in CLDR data)

> You also said something in here about preserving backward compatibility with the old resource file format. Why do we need to do that? It seems like that'd just make everything more complicated.

Actually that is already handled automatically for all date symbol array with one block of code, I did not need to add any code at all to do this. All I needed to do was add code to handle that case where we have a table instead of an array for the era names. Leaving in the capability to handle the as an array also makes the transition easier while separate CLDR-ICU integrations are going on (that is, my change does not need to be synchronized with any particular CLDR-ICU integration).

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Yes, will fix this

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Yes, will fix

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Yes, working on this now

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: > A TreeMap seems heavy for a possibly-sparse mapping from indexes, especially when we never need to lookup by the index key.

Well, that code _did_ lookup by the index key, see line 1829, which you commented on below. However, I think your suggestion #2 is a great improvement, I am not used to using the resizable aspect of ArrayList/UVector so that is a very nice approach. Working on it now (mostly in testing on the C side).

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done, thanks!

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: Revised generallyper suggestion (thanks!), and parallel with updated ICU4C.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: No longer using TreeMap, moot

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: No longer using this code

pedberg-icu:
@markusicu Revised, thanks, PTAL.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
pedberg-icu: Partly done

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1782,49 @@ struct CalendarDataSink : public ResourceSink {
...
                 if (U_FAILURE(errorCode)) { return; }
             } else if (value.getType() == URES_TABLE) {
-                // We are not on a leaf, recursively process the subtable.
-                processResource(path, key, value, errorCode);
+                // We might have an eras table that is replacing an eras leaf array
+                if (path.startsWith(u"eras", 4)) {
+                    // path is one of eras/wide, eras/abbreviated, eras/narrow
+                    ResourceTable rDataTable = value.getTable(errorCode);
+                    int32_t dataTableSize = rDataTable.getSize();
+                    LocalPointer<UVector> dataList(new UVector(uprv_deleteUObject, uhash_compareUnicodeString, dataTableSize, errorCode), errorCode);
```
markusicu: we should not need a LocalPointer for the local UVector
```suggestion
                    UVector dataList(uprv_deleteUObject, uhash_compareUnicodeString, dataTableSize, errorCode);
```

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1782,49 @@ struct CalendarDataSink : public ResourceSink {
...
+                    if (U_FAILURE(errorCode)) { return; }
+                    // Expand the UVector as necessary to have index from 0 up to the max
+                    // eraCode, and fill in the slots for the eras defined in the resource data
+                    // (filling in empty strings for other eras as we expand, since they would
+                    // otherwise not get set to anything in particular such as null).
+                    for (int32_t dataTableIndex = 0; dataTableIndex < dataTableSize; dataTableIndex++) {
+                        rDataTable.getKeyAndValue(dataTableIndex, key, value);
+                        int32_t listIndex = strtol(key, nullptr, 10);
```
markusicu: I don't think we use naked "strtol()" in ICU. I believe we either have a uprv_... version or some even more obscure Taligent-y T_ or class Utility function. Please check.

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1782,49 @@ struct CalendarDataSink : public ResourceSink {
...
+                            while (dataList->size() < listIndex + 1) {
+                                LocalPointer<UnicodeString> emptyString(new UnicodeString(), errorCode);
+                                dataList->adoptElement(emptyString.orphan(), errorCode);
+                            }
+                        }
+                        // Now set the eraName that we just read
+                        LocalPointer<UnicodeString> eraName((value.getType() == URES_STRING) ?
+                                new UnicodeString(value.getUnicodeString(errorCode)) : new UnicodeString(), errorCode);
+                        dataList->setElementAt(eraName.orphan(), listIndex);
+                    }
```
markusicu, line 1813: Please add another `if (U_FAILURE(errorCode)) { return; }` before this line.

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1782,8 +1782,49 @@ struct CalendarDataSink : public ResourceSink {
...
+                    // Now convert to array running from era code 0 to the max era we have data for, fill
+                    // in from the UVector
+                    int32_t dataArraySize = dataList->size();
+                    LocalArray<UnicodeString> dataArray(new UnicodeString[dataArraySize], errorCode);
+                    // Fill out dataArray from dataList (dataList->toArray did not seem to work as expected)
+                    for (int32_t daIdx = 0; daIdx < dataArraySize; daIdx++) {
+                        dataArray[daIdx].setTo(*((UnicodeString*)dataList->elementAt(daIdx)));
```
markusicu:
```suggestion
                        dataArray[daIdx] = std::move(*(reinterpret_cast<UnicodeString *>(dataList->elementAt(daIdx))));
```

richgillam (approved):
Looks good.  I agree with Markus's comments, but don't have any more of my own.

sffc (commented):
Shouldn't there be changes in the resource txt files since you are changing cldr-to-icu?

markusicu:
> Shouldn't there be changes in the resource txt files since you are changing cldr-to-icu?

Please read the PR description.

sffc:
Ok

> will also support the current array form for compatibility with current data

Am I correct that this is going above and beyond ICU's normal data stability policy? We don't guarantee that ICU 77 data works in ICU 78.

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: There is a uprv_strtol, used that. But I do see naked uprv_strtol used elsewhere in i18n (which is why I had used it).

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done

icu4c/source/i18n/dtfmtsym.cpp
pedberg-icu: Done

pedberg-icu:
> Ok
> 
> > will also support the current array form for compatibility with current data
> 
> Am I correct that this is going above and beyond ICU's normal data stability policy? We don't guarantee that ICU 77 data works in ICU 78.

@sffc I also responded to Rich's comment about this, see above. Basically: I did not need to add any code to retain compatibility with handling the current array-form data. There is a single block that handles array-form data for all of the calendar symbols. All I did was: If the data is not an array and is instead a table and is for eras, then handle the table. I did check this with data run through the converter to generate tables for the eras. But I am not including that data in this PR to avoid conflicts with the separate CLDR-ICU integration going on in parallel. The table-form data will be picked up for the next CLDR-ICU integration.

pedberg-icu:
@markusicu I updated per your second set of review comments, thanks.

markusicu (approved):
thanks!

PR #3591, opened 2025-08-11
ICU-23179 Test the error paths and the exact output of toPattern in UnicodeSet parsing

eggrobin:
Before this change, if I replace every `ec = U_MALFORMED_SET;` with `clear(); add(UnicodeString(u"something"));`, the tests pass. This seems like a bad starting point for a semantics-preserving restructuring of the parser (let alone adding more features after that).

Likewise, nothing tests what `toPattern()` actually looks like so long as it parses back to the right thing. But we probably do not want to turn `\p{Cuneiform}` into `[𒑰-𒑴 𒐕𒐞𒐬𒐴𒑏𒑘 𒐀𒐖𒐟𒐣𒐭𒐵𒑊𒑐𒑖𒑙 𒐁𒐈𒐗𒐠𒐤𒐥𒐮𒐯𒐶𒐷𒐺𒐻𒑋𒑑𒑗 𒐂𒐉𒐏𒐘𒐡𒐦𒐰𒐸𒐼-𒐿𒑌𒑒𒑓𒑩 𒐃𒐊𒐐𒐙𒐢𒐧𒐱𒐹𒑍𒑔𒑕𒑪 𒐄𒐋𒐑𒐚𒐨𒑀𒑎𒑫 𒐅𒐌𒐒𒐛𒐩𒑁-𒑃𒑬 𒐆𒐍𒐓𒐜𒐪𒑄𒑅𒑭 𒐇𒐎𒐔𒐝𒐫𒑆-𒑉𒑮 𒐲 𒐳 𒑚-𒑨 𒀀-𒉢 𒋔 𒋕 𒉣-𒋓 𒋖-𒎙 𒒀-𒕃]`; besides being obnoxious/unreadable, if someone is storing the result of `toPattern` and keeping it across Unicode versions, this would change the semantics.

#### Checklist
- [x] Required: Issue filed: ICU-23179

ALLOW_MANY_COMMITS=meow

### DIFF
--- a/icu4c/source/test/intltest/usettest.cpp
+++ b/icu4c/source/test/intltest/usettest.cpp
@@ -4353,6 +4353,12 @@ void UnicodeSetTest::TestToPatternOutput() {
             {u"[ - + - ]", uR"([+\-])"},
             {u"[ { Z e i c h e n k e t t e } Zeichenmenge ]", u"[Zceg-imn{Zeichenkette}]"},
             {uR"([ { \x5A e i c h e n k e t t e } \x5Aeichenmenge ])", u"[Zceg-imn{Zeichenkette}]"},
+            {u"[$d-za-c]", uR"([\$a-z])"},
+            {u"[a-c$d-z]", uR"([\$a-z])"},
+            {uR"([\uFFFFa-z])", uR"([a-z\uFFFF])"},
+            {u"[!-$z]", uR"([!-\$z])"},
+            {u"[-a-cd-z$-]", uR"([\$\-a-z])"},
+            {u"[-$-]", uR"([\$\-])"},
             // A property-query or named-element is kept as-is:
             {uR"(\p{ General_Category = Punctuation })", uR"(\p{ General_Category = Punctuation })"},
             {uR"(\p{P})", uR"(\p{P})"},
@@ -4370,6 +4376,7 @@ void UnicodeSetTest::TestToPatternOutput() {
             uR"([c-za-b\p{ General_Category = Punctuation }])"},
             {u"[^[c]]", uR"([^[c]])"},
             {uR"([ ^ [ \u0000-b d-\U0010FFFF ] ])", uR"([^[^c]])"},
+            {u"[$[]]", uR"([\$[]])"},
             // Spaces are eliminated within a string-literal even when the syntax is preserved.
             {u"[ {Z e i c h e n k e t t e } [] Zeichenmenge ]", u"[{Zeichenkette}[]Zeichenmenge]"},
             // Escapes are removed even when the syntax is preserved.
@@ -4378,6 +4385,10 @@ void UnicodeSetTest::TestToPatternOutput() {
             // A named-element is currently a nested set, so it is preserved and causes the syntax to be
             // preserved.
             {uR"([ \N{LATIN CAPITAL LETTER Z}eichenmenge ])", uR"([\N{LATIN CAPITAL LETTER Z}eichenmenge])"},
+            // An anchor also causes the syntax to be preserved.
+            {u"[ d-z a-c $ ]", u"[d-za-c$]"},
+            {u"[ - a-c d-z $ ]", uR"([\-a-cd-z$])"},
+            {u"[$$$]", uR"([\$\$$])"},
         }) {
         UErrorCode errorCode = U_ZERO_ERROR;
         const UnicodeSet set(expression, errorCode);
@@ -4416,6 +4427,7 @@ void UnicodeSetTest::TestParseErrors() {
             u"[{aa]",
             // "Unquoted '$'".
             u"[a-$]",
+            u"[!-$]",
             // "Invalid range".
             u"[a-a]",  // TODO(egg): Exclude in PDUTS61.
             u"[z-a]",


markusicu:
> we probably do not want to turn `\p{Cuneiform}` into `[𒑰-𒑴 𒐕𒐞...𒒀-𒕃]`; besides being obnoxious/unreadable, if someone is storing the result of `toPattern` and keeping it across Unicode versions, this would change the semantics.

Depends. Sometimes we want the original pattern back if the set has not been changed since parsing the pattern.

Sometimes we do want the properties to be resolved, for example for use in places that do not support properties or not the latest version of Unicode.

I thought that we do get the original pattern back unless we modify the set, and that places that want the properties resolved to something like complement twice or add & remove some code point.

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4334,3 +4335,63 @@ void UnicodeSetTest::TestElementIterator() {
...
+             u"[{aa}-{zz}]",
+             // "'&' not after set".
+             u"[a&z]",
+             u"[{aa}&{zz}]",
+             // "'^' not after '['"
+             u"[a^z]",  // TODO(egg): Exclude from literal-element in PDUTS61.
+             // "Missing operand after operator".
+             u"[a-{zz}]",
+             u"[[a]-{zz}]",
+             u"[[a]&{zz}]",
```
markusicu: FYI: It seems like this should have the same error message as `[a-[b]]` above...

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4334,3 +4335,63 @@ void UnicodeSetTest::TestElementIterator() {
     // begin() & end() return USetElementIterator for which explicit APIs are tested via USet
     // in a header-only unit test file.
 }
+
+void UnicodeSetTest::TestParseErrors() {
+    for (const auto expression : std::vector<std::u16string_view>{
+             // Java error message: "Char expected after operator".
```
markusicu: Please indent by 8 from the start of "for"

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4334,3 +4335,63 @@ void UnicodeSetTest::TestElementIterator() {
...
+             u"[a-$]",
+             // "Invalid range".
+             u"[a-a]",  // TODO(egg): Exclude in PDUTS61.
+             u"[z-a]",
+             // "Set expected after operator".
+             u"[[a]-z]",
+             u"[[a]&z]",
+             // "Missing ']'".
+             u"[a-z",
+         }) {
```
markusicu: please outdent 1 or merge with the previous line

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4334,3 +4335,63 @@ void UnicodeSetTest::TestElementIterator() {
...
+        UErrorCode errorCode = U_ZERO_ERROR;
+        const UnicodeSet set(expression, errorCode);
+        if (errorCode != U_MALFORMED_SET) {
+            UnicodeString s;
+            errln(expression + u": Expected U_MALFORMED_SET, got " + u_errorName(errorCode) +
+                  ", set is " + UnicodeSet(set).complement().complement().toPattern(s));
+        }
+    }
+    for (const auto expression : std::vector<std::u16string_view>{
+             // Java error message: "Invalid property pattern".
```
markusicu: indentation...

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4334,3 +4335,63 @@ void UnicodeSetTest::TestElementIterator() {
...
+             u"[:]",
+             uR"(\p)"
+             u"[:^]",
+             uR"(\P)",
+             uR"(\N)",
+         }) {
+        UErrorCode errorCode = U_ZERO_ERROR;
+        const UnicodeSet set(expression, errorCode);
+        if (errorCode != U_ILLEGAL_ARGUMENT_ERROR) {
```
markusicu, line 4458: FYI -- no idea why we use two different error codes...

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4336,6 +4337,60 @@ void UnicodeSetTest::TestElementIterator() {
...
+             {u"[c-za-b]", u"[a-z]"},
+             {u"[  c - z  a - b  ]", u"[a-z]"},
+             {uR"([ ^ \u0000-b d-\U0010FFFF ])", u"[c]"},
+             {uR"([ \u0000-b d-\U0010FFFF ])", u"[^c]"},
+             {u"[ - - ]", uR"([\-])"},
+             {u"[ - _ - ]", uR"([\-_])"},
+             {u"[ - + - ]", uR"([+\-])"},
+             {u"[ { Z e i c h e n k e t t e } Zeichenmenge ]", u"[Zceg-imn{Zeichenkette}]"},
+             {uR"([ { \x5A e i c h e n k e t t e } \x5Aeichenmenge ])", u"[Zceg-imn{Zeichenkette}]"},
+             // A property-query or named-element is kept as-is:
```
markusicu: FYI -- we _could_ remove ignored white space

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4336,6 +4337,60 @@ void UnicodeSetTest::TestElementIterator() {
...
+             {uR"(\p{P})", uR"(\p{P})"},
+             {uR"(\p{gc=P})", uR"(\p{gc=P})"},
+             {uR"([: general category = punctuation :])", uR"([: general category = punctuation :])"},
+             {uR"([: ^general category = punctuation :])", uR"([: ^general category = punctuation :])"},
+             {uR"(\P{ gc = punctuation })", uR"(\P{ gc = punctuation })"},
+             {uR"(\N{ latin small letter a })", uR"(\N{ latin small letter a })"},
+             // If there is any Restriction among the terms, its syntax is mostly as-is (spaces are
+             // still eliminated), with the exception that an initial UnescapedHyphenMinus gets escaped.
+             // This is applied recursively, so innermost ranges-only UnicodeSets get normalized.
+             {u"[ c-z a-b [c-f g-z] ]", u"[c-za-b[c-z]]"},
```
markusicu: FYI -- I am surprised that it doesn't just turn this into `[a-z]`

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4336,6 +4337,60 @@ void UnicodeSetTest::TestElementIterator() {
...
+             {u"[- + c-z a-b [c-f g-z] -]", uR"([\-+c-za-b[c-z]-])"},
+             {uR"([ c-z a-b \p{ General_Category = Punctuation } ])",
+              uR"([c-za-b\p{ General_Category = Punctuation }])"},
+             {u"[^[c]]", uR"([^[c]])"},
+             {uR"([ ^ [ \u0000-b d-\U0010FFFF ] ])", uR"([^[^c]])"},
+             // Spaces are eliminated within a string-literal even when the syntax is preserved.
```
markusicu: This feels like a bug

markusicu (commented):
what about Java? separate PR?

eggrobin:
> I thought that we do get the original pattern back unless we modify the set, and that places that want the properties resolved to something like complement twice or add & remove some code point.

That is pretty much the case, yes. And we do not want to change that and resolve everything all the time.

icu4c/source/test/intltest/usettest.cpp
eggrobin: We probably should at some point, yes.

icu4c/source/test/intltest/usettest.cpp
eggrobin: Same. That would probably be harmless. Ideally `[c-za-b[ab\p{Something}]]` could turn into `[a-z\p{Something}]` too.

icu4c/source/test/intltest/usettest.cpp
eggrobin: I think that is OK (spaces are ignored in a UnicodeSet string literal).

eggrobin:
> what about Java? separate PR?

Yes. I want to first rewrite the C++ parser (bug-compatibly) before I touch Java.

eggrobin:
(Obviously I will need to rewrite the Java parser in the same way before I can consider actually changing the API. But so long as I am trying not to do anything, one language at a time.)

icu4c/source/test/intltest/usettest.cpp
markusicu: Well, I am not sure if it's intentional. I would expect to be able to have a string that contains spaces.

eggrobin: I think I agree that it would be nicer if the string literals did not ignore spaces, but that is a pretty seriously breaking change (they are very consciously ignored in the UTS35 grammar and in both implementations and have been forever).

I would support a motion in ICU-TC to stop ignoring them, but I think that the use cases broken by changes to make the variable and `\N` mechanisms less mad are transparently wrong, changing the behaviour of spaces in string literals warrants big flashing signs to warn users.

eggrobin: Hmm, actually, UTS35 is unclear (as often). But anyway, I think people are more likely to be bitten by this than by defining `-` as a variable.

macchiati: Whoa, I never realized that it deleted spaces within a string literal; that is definitely unexpected:

`[a-b e-f { x y }] == [a b e f {xy}]`

You have to escape them if you want them. Ugg

`[a-b e-f {\ x\ y\ }]`

eggrobin: OK, if the designer of this did not expect it, I am definitely in favour of changing the behaviour! But still, we should have big flashing signs when we make this change. 🚨

macchiati:
Definitely 🚨🚨🚨

BTW, I didn't do the code for UnicodeSet; that was mostly Alan Liu with
some (important) refinements later by Markus; the only thing I remember
doing was the 'hooks' for extending properties.

icu4c/source/test/intltest/usettest.cpp
macchiati: I think the key goal behind retaining the original string (or something like it) was to retain all mentions of properties.

That is: if \p{something} occurs N times in the parsed string, then it occurs N times in the result.

As long as that is the case, I think we can do any natural simplification that is easy to do.

PR #3592, opened 2025-08-12
ICU-20392 Replace use of CharString with FixedString in Locale::Heap

roubert:
Add internal class `FixedString` to store fixed-length strings and use this to replace use of `CharString` in `Locale::Heap`. This avoids allocating any more storage than strictly needed, reducing the size of all `Locale` objects that need to be heap allocated.

The storage required for different kinds of `Locale` objects then looks like this, all of them now smaller than they were with the legacy implementation, sizes in bytes:

|| legacy | `CharString` | `FixedString` | _ratio_ |
| :--- | ---: | ---: | ---: | ---: |
| simple | 224 | 48 | 48 | 21% |
| complex | 227 | 208 | 117 | 52% |
| pathological | 245 | 368 | 237 | 97% |
| posix | 224 | 48 | 48 | 21% |
| fubar | 408 | 432 | 280 | 69% |

|||
| --- | --- |
| simple | `hi` |
| complex | `hi@t=en-h0-hybrid` |
| pathological | `sco__FONIPA_SCOTLAND@calendar=buddhist;collation=phonebook;t=tlh-fonxsamp-simple-h0-hybrid-m0-ungegn-1977;x=hello-world` |
| posix | `en_US_POSIX` |
| fubar | `vo__FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR` |

#### Checklist

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -50,6 +50,7 @@
 #include "charstrmap.h"
 #include "cmemory.h"
 #include "cstring.h"
+#include "fixedstring.h"
 #include "mutex.h"
 #include "putilimp.h"
 #include "uassert.h"
@@ -281,8 +282,8 @@ struct Locale::Heap : public UMemory {
     char script[ULOC_SCRIPT_CAPACITY];
     char region[ULOC_COUNTRY_CAPACITY];
     int32_t variantBegin;
-    CharString fullName;
-    CharString baseName;
+    FixedString fullName;
+    FixedString baseName;
 
     const char* getLanguage() const { return language; }
     const char* getScript() const { return script; }
@@ -303,11 +304,26 @@ struct Locale::Heap : public UMemory {
 
     Heap(const Heap& other, UErrorCode& status)
         : variantBegin(other.variantBegin),
-          fullName(other.fullName, status),
-          baseName(other.baseName, status) {
-        uprv_memcpy(language, other.language, sizeof language);
-        uprv_memcpy(script, other.script, sizeof script);
-        uprv_memcpy(region, other.region, sizeof region);
+          fullName(),
+          baseName() {
+        if (U_SUCCESS(status)) {
+            uprv_memcpy(language, other.language, sizeof language);
+            uprv_memcpy(script, other.script, sizeof script);
+            uprv_memcpy(region, other.region, sizeof region);
+            if (!other.fullName.isEmpty()) {
+                fullName = other.fullName;
+                if (fullName.isEmpty()) {
+                    status = U_MEMORY_ALLOCATION_ERROR;
+                } else {
+                    if (!other.baseName.isEmpty()) {
+                        baseName = other.baseName;
+                        if (baseName.isEmpty()) {
+                            status = U_MEMORY_ALLOCATION_ERROR;
+                        }
+                    }
+                }
+            }
+        }
     }
 
     // Move should be done on the std::unique_ptr object that owns this.
@@ -1898,19 +1914,14 @@ Locale& Locale::init(StringPiece localeID, UBool canonicalize)
         err = U_ZERO_ERROR;
         length = parse(localeID, fullName, sizeof Nest::baseName, err);
 
-        std::unique_ptr<CharString> fullNameBuffer;
+        FixedString fullNameBuffer;
         if (err == U_BUFFER_OVERFLOW_ERROR || length >= static_cast<int32_t>(sizeof Nest::baseName)) {
             /*Go to heap for the fullName if necessary*/
-            fullNameBuffer = std::make_unique<CharString>();
-            if (!fullNameBuffer) {
+            if (!fullNameBuffer.reserve(length + 1)) {
                 break; // error: out of memory
             }
+            fullName = fullNameBuffer.getAlias();
             err = U_ZERO_ERROR;
-            int32_t capacity;
-            fullName = fullNameBuffer->getAppendBuffer(length + 1, length + 1, capacity, err);
-            if (U_FAILURE(err)) {
-                break; // error: out of memory
-            }
             length = parse(localeID, fullName, length + 1, err);
         }
         if(U_FAILURE(err) || err == U_STRING_NOT_TERMINATED_WARNING) {
@@ -1980,7 +1991,7 @@ Locale& Locale::init(StringPiece localeID, UBool canonicalize)
 
         if (Nest::fits(length, language, script, region)) {
             U_ASSERT(fullName == nest.baseName);
-            U_ASSERT(!fullNameBuffer);
+            U_ASSERT(fullNameBuffer.isEmpty());
             nest.init(language, script, region, variantBegin);
         } else {
             std::unique_ptr<Heap>& heap = payload.emplace<std::unique_ptr<Heap>>(
@@ -1989,19 +2000,17 @@ Locale& Locale::init(StringPiece localeID, UBool canonicalize)
                 break; // error: out of memory
             }
             if (fullName == nest.baseName) {
-                U_ASSERT(!fullNameBuffer);
-                heap->fullName.copyFrom({fullName, length}, err);
+                U_ASSERT(fullNameBuffer.isEmpty());
+                heap->fullName = {fullName, static_cast<std::string_view::size_type>(length)};
+                if (heap->fullName.isEmpty()) {
+                    break; // error: out of memory
+                }
             } else {
-                U_ASSERT(fullNameBuffer);
-                fullNameBuffer->append(fullName, length, err);
-                heap->fullName = std::move(*fullNameBuffer);
-            }
-            if (U_FAILURE(err)) {
-                break;
+                U_ASSERT(!fullNameBuffer.isEmpty());
+                heap->fullName = std::move(fullNameBuffer);
             }
         }
 
-        err = U_ZERO_ERROR;
         initBaseName(err);
         if (U_FAILURE(err)) {
             break;
@@ -2071,8 +2080,11 @@ Locale::initBaseName(UErrorCode &status) {
                 status = U_MEMORY_ALLOCATION_ERROR;
                 return;
             }
-            copy->fullName.copyFrom(fullName, status);
-            if (U_FAILURE(status)) { return; }
+            copy->fullName = fullName;
+            if (copy->fullName.isEmpty()) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+                return;
+            }
             heap = &payload.emplace<std::unique_ptr<Heap>>(std::move(copy));
             if (!*heap) {
                 status = U_MEMORY_ALLOCATION_ERROR;
@@ -2088,7 +2100,12 @@ Locale::initBaseName(UErrorCode &status) {
             (*heap)->variantBegin = baseNameLength;
         }
 
-        (*heap)->baseName.copyFrom({fullName, baseNameLength}, status);
+        (*heap)->baseName = {fullName, static_cast<std::string_view::size_type>(baseNameLength)};
+        if ((*heap)->baseName.isEmpty()) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+            setToBogus();
+            return;
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


icu4c/source/common/fixedstring.h
```diff
@@ -0,0 +1,91 @@
...
+
+    char* getAlias() {
+        return ptr;
+    }
+
+    bool isEmpty() const {
+        return ptr == nullptr;
+    }
+
+    void reserve(size_t size) {
```
markusicu: This seems dangerous since it leaves the array uninitialized.
Can you make it private?
Otherwise please add API docs for this function.

icu4c/source/common/fixedstring.h
```diff
@@ -0,0 +1,91 @@
...
+ * A terminating NUL is always stored, but the length of the string isn't.
+ * An empty string is stored as nullptr, allocating no storage at all.
+ *
+ * This class wants to be convenient but is also deliberately minimalist.
+ * Please do not add methods if they only add minor convenience.
+ */
+class FixedString : public UMemory {
+public:
+    FixedString() = default;
+    ~FixedString() { operator delete[](ptr); }
```
markusicu, line 29: What is the advantage of calling `operator delete[](ptr)` instead of `uprv_free(ptr)` & same for `operator new[]` vs. `uprv_malloc()`? Seems ok but confusing; at first I thought you were calling global new & delete.

icu4c/source/common/locid.cpp
```diff
@@ -301,10 +302,10 @@ struct Locale::Heap : public UMemory {
         copyToArray<&Heap::region>(region, this);
     }
 
-    Heap(const Heap& other, UErrorCode& status)
+    Heap(const Heap& other)
         : variantBegin(other.variantBegin),
-          fullName(other.fullName, status),
-          baseName(other.baseName, status) {
+          fullName(other.fullName),
+          baseName(other.baseName) {
```
markusicu: This hides out-of-memory errors; in OOM one or both of these just look "empty". Instead, the Locale object should be .isBogus() when that happens.

icu4c/source/common/locid.cpp
```diff
@@ -477,9 +478,7 @@ Locale& Locale::operator=(const Locale& other) {
         payload.emplace<Nest>(*nest);
     } else if (const std::unique_ptr<Heap>* heap = std::get_if<std::unique_ptr<Heap>>(&other.payload)) {
         U_ASSERT(*heap);
-        if (UErrorCode status = U_ZERO_ERROR;
-                !payload.emplace<std::unique_ptr<Heap>>(std::make_unique<Heap>(**heap, status)) ||
-                U_FAILURE(status)) {
+        if (!payload.emplace<std::unique_ptr<Heap>>(std::make_unique<Heap>(**heap))) {
```
markusicu: This is a regression.

icu4c/source/common/locid.cpp
```diff
@@ -1898,19 +1897,15 @@ Locale& Locale::init(StringPiece localeID, UBool canonicalize)
...
         length = parse(localeID, fullName, sizeof Nest::baseName, err);
 
-        std::unique_ptr<CharString> fullNameBuffer;
+        FixedString fullNameBuffer;
         if (err == U_BUFFER_OVERFLOW_ERROR || length >= static_cast<int32_t>(sizeof Nest::baseName)) {
             /*Go to heap for the fullName if necessary*/
-            fullNameBuffer = std::make_unique<CharString>();
-            if (!fullNameBuffer) {
+            fullNameBuffer.reserve(length + 1);
+            if (fullNameBuffer.isEmpty()) {
```
markusicu: Checking isEmpty() as a proxy for "memory allocation failed" is very confusing.

Suggestion: Make reserve() return bool for success --> `if (!fullNameBuffer.reserve(...)) { OOM }`
Alternative, add a UErrorCode parameter to it.

icu4c/source/common/fixedstring.h
roubert: It just seemed natural to me to call `new[]` and `delete[]` from `UMemory` so that they'll resolve to either global `new[]` and `delete[]` or to `uprv_malloc()` and `uprv_free()` respectively, depending on `U_OVERRIDE_CXX_ALLOCATION`.

icu4c/source/common/fixedstring.h
roubert: Done.

icu4c/source/common/locid.cpp

icu4c/source/common/locid.cpp

icu4c/source/common/locid.cpp

icu4c/source/common/locid.cpp
```diff
@@ -302,10 +302,14 @@ struct Locale::Heap : public UMemory {
...
 
-    Heap(const Heap& other)
+    Heap(const Heap& other, UErrorCode& status)
         : variantBegin(other.variantBegin),
           fullName(other.fullName),
           baseName(other.baseName) {
+        if (fullName.isEmpty() != other.fullName.isEmpty() ||
+            baseName.isEmpty() != other.baseName.isEmpty()) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+        }
```
markusicu: Condition-continuation indentation should be larger than body indentation, so that we can tell one from the other.
```suggestion
        if (fullName.isEmpty() != other.fullName.isEmpty() ||
                baseName.isEmpty() != other.baseName.isEmpty()) {
            status = U_MEMORY_ALLOCATION_ERROR;
        }
```

icu4c/source/common/locid.cpp
```diff
@@ -302,10 +302,14 @@ struct Locale::Heap : public UMemory {
...
     }
 
-    Heap(const Heap& other)
+    Heap(const Heap& other, UErrorCode& status)
         : variantBegin(other.variantBegin),
           fullName(other.fullName),
           baseName(other.baseName) {
+        if (fullName.isEmpty() != other.fullName.isEmpty() ||
+            baseName.isEmpty() != other.baseName.isEmpty()) {
+            status = U_MEMORY_ALLOCATION_ERROR;
```
markusicu: Don't set a failure code if there is one already.

Suggestion:
- Don't copy-construct the names in the initializer
- First line here: check the status, do nothing if failure
- Then copy the names
- Then check if either copy failed and set the error

Alternative: Add a UErrorCode to the copy constructor, have that not copy when already failure, else set a failure if OOM. Then you can copy-construct in the initializers and need not check later that the copies worked. Simpler & safer.

icu4c/source/common/locid.cpp
```diff
@@ -2071,8 +2069,11 @@ Locale::initBaseName(UErrorCode &status) {
                 status = U_MEMORY_ALLOCATION_ERROR;
                 return;
             }
-            copy->fullName.copyFrom(fullName, status);
-            if (U_FAILURE(status)) { return; }
+            copy->fullName = fullName;
+            if (copy->fullName.isEmpty()) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+                return;
+            }
```
markusicu, line 2087: optional: It would be easier/safer if the FixedString-assign-from-string_view took a UErrorCode, like with CharString, but if you caught every place where it's called, this works.

icu4c/source/common/locid.cpp
```diff
@@ -2071,8 +2069,11 @@ Locale::initBaseName(UErrorCode &status) {
                 status = U_MEMORY_ALLOCATION_ERROR;
                 return;
             }
-            copy->fullName.copyFrom(fullName, status);
-            if (U_FAILURE(status)) { return; }
+            copy->fullName = fullName;
+            if (copy->fullName.isEmpty()) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+                return;
```
markusicu, line 2086: Some existing and new code calls `setToBogus()` when something goes wrong, other places don't. Please check the call sites (maybe they consistently setToBogus() after calling initBaseName()?) and make this function consistent.

icu4c/source/common/locid.cpp

icu4c/source/common/locid.cpp
roubert: This code was formatted by `clang-format` so it's correct according to ICU4C's `.clang-format` and such formatting can be found elsewhere in the code base.

icu4c/source/common/locid.cpp
roubert: I've checked and it's consistent as far as I can see: whether `setToBogus()` is called or not after an operation fails depends on whether the failure could have left the object in a bogus state. Like on this line, failure here would leave a fully working object, just with erroneously empty `fullName`, so not bogus.

icu4c/source/common/locid.cpp
roubert: That's just a question of where to put the complexity. All of this is internal so it'll be easy to refactor later, but for now it feels like a good idea to me to have as little complexity in `FixedString` as possible.

icu4c/source/common/locid.cpp
markusicu: Well, it is poorly readable. If you like to use clang-format then please try to configure it to produce readable output; or fix problems after running it.

icu4c/source/common/locid.cpp
markusicu: > failure here would leave a fully working object, just with erroneously empty `fullName`, so not bogus.

So this failure does not leave the object bogus, but the different getters return inconsistent values? That's bad.

icu4c/source/common/locid.cpp
roubert: That's purely a matter of taste. I personally agree with whoever wrote the current configuration and think it's beneficial for readability to have the statements lined up like they are now, but if you disagree with that you'll have to get the configuration changed, for trying to get everyone to manually edit the output of the formatter all the time won't be a productive path forward.

icu4c/source/common/locid.cpp
roubert: Well, that's the way it currently is. We could change it so that any failure while modifying the object always sets it to bogus, but that'd be out-of-scope for this PR.

markusicu: > Well, that's the way it currently is.

Fair enough --> ICU-23178

icu4c/source/common/locid.cpp
markusicu: I disagree very much. It is really confusing to have no separation between the if-condition and the if-body. In Google style, we always have a double indent for the continued condition. Google style uses an indent of 2, so in a code base where the indent is 4, doubling makes it 8.

We don't routinely use clang format in ICU, and it looks like we shouldn't apply it blindly. When it has a bug, we should fix its output.

roubert: > We don't routinely use clang format in ICU,

Some of us do, that's why ICU4C has had a custom `.clang-format` for many years now.

> When it has a bug, we should fix its output.

I fully agree with that, I just don't agree that there's something wrong with this formatting here, I quite like it.

markusicu: I don't accept
```
        if (fullName.isEmpty() != other.fullName.isEmpty() ||
            baseName.isEmpty() != other.baseName.isEmpty()) {
        }
```

I can accept
```
        if (fullName.isEmpty() != other.fullName.isEmpty() ||
                baseName.isEmpty() != other.baseName.isEmpty()) {
        }
```
(like in [this example](https://google.github.io/styleguide/cppguide.html#Boolean_Expressions))
or
```
        if (fullName.isEmpty() != other.fullName.isEmpty() ||
            baseName.isEmpty() != other.baseName.isEmpty()
        ) {
        }
```

roubert: Let's eliminate the entire formatting question by instead doing a microoptimization: `baseName` is always empty if `fullName` is empty and there's also no point in attempting to copy `baseName` if copying `fullName` failed.

markusicu (approved):

PR #3593, opened 2025-08-14
ICU-23172 part 2: Make EraRules work if era code sequence starts at >0 and/or has holes

pedberg-icu:
This is the second and final part of a multipart change to convert era data from an array to a map (only affects the name data, the era rules data was already a map), and then to allow era codes that do not start from 0, or may potentially have holes in the sequence (as long as the date ranges for the eras are contiguous). This is needed to unblock some changes to era data in CLDR, such as for coptic (eliminating the current first era) and japanese (eliminating eras before Meiji). The first part was done in PR #3588.

- This second part makes the internal EraRules class handle supplemental calendar eras data (which was already in map form) in which the era code sequence may not start at 0, or may have holes in it.
    - As part of this I added a method getMaxEraCode() which now may produce a different result than getNumberOfEras(); some clients want one behavior and some the other (at least in test code). I also renamed some methods to make it clear they were dealing with an era code, not an era index that might be assumed to be sequential from 0, and an index into something. It may also be useful to add getMinEraCode() in the future.
    - I tested this change partly by manually deleting eras 0-229 from the Japanese calendar data in C `data/misc/supplementalData.txt`, integrating that change over to ICU4J, and running the normal tests for both C and J with that change, and with some logging to verify proper construction of data in EraRules. That supplementalData.txt change is not in the PR however; it will happen in a near-future CLDR-ICU integration after the eras are deleted in CLDR.
    - There are currently some tests that try to calculate with era 0 in all calendars, which will need to be modified when we no longer have era 0 in some calendars. There are also tests that try to format dates in the Japanese calendar for Gregorian years like 1456 or 1776, these will produce a different formatted result if the corresponding Japanese-calendar eras are deleted. I added notes for these but thought it premature to change the tests themselves (until we actually remove the eras).

#### Checklist
- [x] Required: Issue filed: ICU-23172

### DIFF
--- a/icu4c/source/i18n/erarules.cpp
+++ b/icu4c/source/i18n/erarules.cpp
@@ -17,6 +17,7 @@
 #include "erarules.h"
 #include "gregoimp.h"
 #include "uassert.h"
+#include "uvectr32.h"
 
 U_NAMESPACE_BEGIN
 
@@ -102,9 +103,9 @@ static int32_t compareEncodedDateWithYMD(int encoded, int year, int month, int d
     }
 }
 
-EraRules::EraRules(LocalMemory<int32_t>& eraStartDates, int32_t numEras)
-    : numEras(numEras) {
-    startDates = std::move(eraStartDates);
+EraRules::EraRules(LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn, int32_t minEraIn, int32_t numErasIn)
+    : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn) {
+    startDates = std::move(startDatesIn);
     initCurrentEra();
 }
 
@@ -127,12 +128,10 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
     int32_t numEras = ures_getSize(rb.getAlias());
     int32_t firstTentativeIdx = MAX_INT32;
 
-    LocalMemory<int32_t> startDates(static_cast<int32_t *>(uprv_malloc(numEras * sizeof(int32_t))));
-    if (startDates.isNull()) {
-        status = U_MEMORY_ALLOCATION_ERROR;
+    UVector32 eraStartDates(numEras, status);
+    if (U_FAILURE(status)) {
         return nullptr;
     }
-    uprv_memset(startDates.getAlias(), 0 , numEras * sizeof(int32_t));
 
     while (ures_hasNext(rb.getAlias())) {
         LocalUResourceBundlePointer eraRuleRes(ures_getNextResource(rb.getAlias(), nullptr, &status));
@@ -141,16 +140,27 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
         }
         const char *eraIdxStr = ures_getKey(eraRuleRes.getAlias());
         char *endp;
-        int32_t eraIdx = static_cast<int32_t>(strtol(eraIdxStr, &endp, 10));
+        int32_t eraIdx = static_cast<int32_t>(uprv_strtol(eraIdxStr, &endp, 10));
         if (static_cast<size_t>(endp - eraIdxStr) != uprv_strlen(eraIdxStr)) {
             status = U_INVALID_FORMAT_ERROR;
             return nullptr;
         }
-        if (eraIdx < 0 || eraIdx >= numEras) {
+        if (eraIdx < 0) {
             status = U_INVALID_FORMAT_ERROR;
             return nullptr;
         }
-        if (isSet(startDates[eraIdx])) {
+        if (eraIdx + 1 > eraStartDates.size()) {
+            eraStartDates.ensureCapacity(eraIdx + 1, status); // needed only to minimize expansions
+            // Fill in 0 for all added slots (else they are undefined)
+            while (eraStartDates.size() < eraIdx + 1) {
+                eraStartDates.addElement(0, status);
+            }
+            if (U_FAILURE(status)) {
+                return nullptr;
+            }
+        }
+        // Now set the startDate that we just read
+        if (isSet(eraStartDates.elementAti(eraIdx))) {
             // start date of the index was already set
             status = U_INVALID_FORMAT_ERROR;
             return nullptr;
@@ -174,7 +184,7 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
                     status = U_INVALID_FORMAT_ERROR;
                     return nullptr;
                 }
-                startDates[eraIdx] = encodeDate(fields[0], fields[1], fields[2]);
+                eraStartDates.setElementAt(encodeDate(fields[0], fields[1], fields[2]), eraIdx);
             } else if (uprv_strcmp(key, "named") == 0) {
                 const char16_t *val = ures_getString(res.getAlias(), &len, &status);
                 if (u_strncmp(val, VAL_FALSE, VAL_FALSE_LEN) == 0) {
@@ -185,7 +195,7 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
             }
         }
 
-        if (isSet(startDates[eraIdx])) {
+        if (isSet(eraStartDates.elementAti(eraIdx))) {
             if (hasEnd) {
                 // This implementation assumes either start or end is available, not both.
                 // For now, just ignore the end rule.
@@ -194,7 +204,7 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
             if (hasEnd) {
                 // The islamic calendars now have an end-only rule for the
                 // second (and final) entry; basically they are in reverse order.
-                startDates[eraIdx] = MIN_ENCODED_START;
+                eraStartDates.setElementAt(MIN_ENCODED_START, eraIdx);
             } else {
                 status = U_INVALID_FORMAT_ERROR;
                 return nullptr;
@@ -213,47 +223,91 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
         }
     }
 
-    EraRules *result;
-    if (firstTentativeIdx < MAX_INT32 && !includeTentativeEra) {
-        result = new EraRules(startDates, firstTentativeIdx);
-    } else {
-        result = new EraRules(startDates, numEras);
+    // Remove from eraStartDates any tentative eras if they should not be included
+    // (these would be the last entries). Also reduce numEras appropriately.
+    if (!includeTentativeEra) {
+        while (firstTentativeIdx < eraStartDates.size()) {
+            int32_t lastEraIdx = eraStartDates.size() - 1;
+            if (isSet(eraStartDates.elementAti(lastEraIdx))) { // If there are multiple tentativeEras, some may be unset
+                numEras--;
+            }
+            eraStartDates.removeElementAt(lastEraIdx);
+        }
+        // Remove any remaining trailing unSet entries
+        // (can only have these if tentativeEras have been removed)
+        while (eraStartDates.size() > 0 && !isSet(eraStartDates.elementAti(eraStartDates.size() - 1))) {
+            eraStartDates.removeElementAt(eraStartDates.size() - 1);
+        }
     }
-
+    // Remove from eraStartDates any initial 0 entries, keeping the original index (eraCode)
+    // of the first non-zero entry as minEra; then we can add that back to the offset in the
+    // compressed array to get the correct eraCode.
+    int32_t minEra = 0;
+    while (eraStartDates.size() > 0 && !isSet(eraStartDates.elementAti(0))) {
+        eraStartDates.removeElementAt(0);
+        minEra++;
+    }
+    // Convert eraStartDates to int32_t array startDates and pass to EraRules constructor,
+    // along with startDatesLength, minEra and numEras (which may be different from startDatesLength)
+    LocalMemory<int32_t> startDates(static_cast<int32_t *>(uprv_malloc(eraStartDates.size() * sizeof(int32_t))));
+    if (startDates.isNull()) {
+        status = U_MEMORY_ALLOCATION_ERROR;
+        return nullptr;
+    }
+    for (int32_t eraIdx = 0; eraIdx < eraStartDates.size(); eraIdx++) {
+        startDates[eraIdx] = eraStartDates.elementAti(eraIdx);
+    }
+    EraRules *result = new EraRules(startDates, eraStartDates.size(), minEra, numEras);
     if (result == nullptr) {
         status = U_MEMORY_ALLOCATION_ERROR;
     }
     return result;
 }
 
-void EraRules::getStartDate(int32_t eraIdx, int32_t (&fields)[3], UErrorCode& status) const {
+void EraRules::getStartDate(int32_t eraCode, int32_t (&fields)[3], UErrorCode& status) const {
     if(U_FAILURE(status)) {
         return;
     }
-    if (eraIdx < 0 || eraIdx >= numEras) {
-        status = U_ILLEGAL_ARGUMENT_ERROR;
+    int32_t startDate = 0;
+    if (eraCode >= minEra) {
+        int32_t startIdx = eraCode - minEra;
+        if (startIdx < startDatesLength) {
+            startDate = startDates[startIdx];
+        }
+    }
+    if (isSet(startDate)) {
+        decodeDate(startDate, fields);
         return;
     }
-    decodeDate(startDates[eraIdx], fields);
+    // We did not find the requested eraCode in our data
+    status = U_ILLEGAL_ARGUMENT_ERROR;
+    return;
 }
 
-int32_t EraRules::getStartYear(int32_t eraIdx, UErrorCode& status) const {
+int32_t EraRules::getStartYear(int32_t eraCode, UErrorCode& status) const {
     int year = MAX_INT32;   // bogus value
     if(U_FAILURE(status)) {
         return year;
     }
-    if (eraIdx < 0 || eraIdx >= numEras) {
-        status = U_ILLEGAL_ARGUMENT_ERROR;
+    int32_t startDate = 0;
+    if (eraCode >= minEra) {
+        int32_t startIdx = eraCode - minEra;
+        if (startIdx < startDatesLength) {
+            startDate = startDates[startIdx];
+        }
+    }
+    if (isSet(startDate)) {
+        int fields[3];
+        decodeDate(startDate, fields);
+        year = fields[0];
         return year;
     }
-    int fields[3];
-    decodeDate(startDates[eraIdx], fields);
-    year = fields[0];
-
+    // We did not find the requested eraCode in our data
+    status = U_ILLEGAL_ARGUMENT_ERROR;
     return year;
 }
 
-int32_t EraRules::getEraIndex(int32_t year, int32_t month, int32_t day, UErrorCode& status) const {
+int32_t EraRules::getEraCode(int32_t year, int32_t month, int32_t day, UErrorCode& status) const {
     if(U_FAILURE(status)) {
         return -1;
     }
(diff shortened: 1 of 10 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -25,12 +27,12 @@ public class EraRules {
     private static final int MONTH_MASK = 0x0000FF00;
     private static final int DAY_MASK = 0x000000FF;
 
-    private int[] startDates;
+    private int[][] startDateAndEras; // startDate in [n][0], eraCode in [n][1]
```
pedberg-icu: This may not be the ideal representation for this data, but there are some limitations on what we can do since the EraRules getInstance method that uses startDateAndEras is declared static (so it cannot use a custom data class). I guess we could instead have two parallel int[] arrays, one for startDates and one for eraCodes, not sure that is better.

markusicu: I don't understand this comment. It is normal for getInstance() (a factory method) to be static, and I don't know what that has to do with the data structure.

You could define a class for the pair of values. It would make the code more readable, but it would cost a large number of Java objects.

Even `int[numEras][2]` allocates numEras+1 objects in Java -- the outer array, and numEras mini arrays. (Same number of objects as with a bespoke type.)

Please use two parallel arrays for startDates and eraCodes, and keep C++ reasonably parallel.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -25,12 +27,12 @@ public class EraRules {
...
     private int numEras;
-    private int currentEra;
+    private int currentEraStartIndex;
 
-    private EraRules(int[] startDates, int numEras) {
-        this.startDates = startDates;
+    private EraRules(int[][] startDateAndEras, int numEras) {
```
markusicu: if numEras == array.length then it is redundant and should be removed both as a parameter and as a field. (Java only)

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -56,11 +58,19 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
-            if (eraIdx < 0 || eraIdx >= numEras) {
+            if (eraIdx < 0) {
                 throw new ICUException("Era rule key:" + eraIdxStr + " in era rule data for " + calType.getId()
-                        + " must be in range [0, " + (numEras - 1) + "]");
+                        + " must be > 0");
             }
-            if (isSet(startDates[eraIdx])) {
+            if (eraIdx + 1 > eraStartDates.size()) {
+                eraStartDates.ensureCapacity(eraIdx + 1); // needed only to minimize expansions
+                // Fill in empty strings for all added slots
```
markusicu:
```suggestion
                // Fill in zeros for all added slots.
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -116,68 +126,103 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
             }
         }
 
+        // Now make array of just the eras we have rules for, with startDate and eraCode for each
+        int[][] startDateAndEras = new int[numEras][2];
+        int startDateIdx = 0;
+        for (int eraIdx = 0; eraIdx < eraStartDates.size(); eraIdx++) {
+            if (isSet(eraStartDates.get(eraIdx).intValue())) {
+                if (startDateIdx >= numEras) {
```
markusicu: This is not possible because of `numEras = erasRes.getSize();` above, right?

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -116,68 +126,103 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
+                    throw new ICUException(
+                            "With start date for era code " + eraIdx + ", found more start dates than era rule bundles (" + numEras + ")");
+                }
+                startDateAndEras[startDateIdx][0] = eraStartDates.get(eraIdx).intValue();
+                startDateAndEras[startDateIdx][1] = eraIdx;
+                startDateIdx++;
+            }
+        }
+        if (startDateIdx != numEras) {
```
markusicu: at most assert, I think

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: If I define a class for the pair of values, then assigning them requires reference to the "this" variable for that class, and that cannot be done from within a static method (according to there error message I was getting; presumably you cannot have instance variables in a class created inside a static method; not sure how it works for the JDK classes). But I can change to use parallel arrays.

markusicu: > If I define a class for the pair of values, then assigning them requires reference to the "this" variable for that class, and that cannot be done from within a static method

Aha!
If you define a class and don't declare the class `static` then yes it requires (and includes) an "outer this" pointer.
If you do define the inner class `static` then you can use it from static methods, and in fact if you don't need an "outer this" pointer, making the inner class `static` will save some space and performance.

> But I can change to use parallel arrays.

yes please

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: It is not necessarily the same; in some cases what is passed for numEras is `firstTentativeIdx` which is a smaller value; this is used in cases such as testing new eras for the Japanese calendar, when for some period of time the start date of the new era is known but its name may not yet be known. So we need to keep this param.

pedberg-icu: The value passed here for numEras is not necessarily the length of the array. In some cases (such as in testing the Japanese calendar when the start date of an era is known before the name is) the array(s) of era data may include "tentative" eras at the end, which do not yet have a name. There is an environment flag that can set whether to include these. If they are not included, the value passed for numEras only included the count of non-tentative eras. (And I realized the first version of this code was passing the wrong value: The era code of the first tentative era instead of the count of non-tentative eras, so that is fixed).

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Should not be, I guess I was overdoing the checks.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Probably another instance of overdoing the checks, I think this is actually not possible.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done, removed in C too

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Removed here and in C

pedberg-icu:
@markusicu Thanks for comments, updated.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -126,30 +128,27 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
             }
         }
-        if (startDateIdx != numEras) {
-            throw new ICUException(
-                     "number of start dates " + startDateIdx + "did not equal number of era rule bundles (" + numEras + ")");
-        }
 
         if (firstTentativeIdx < Integer.MAX_VALUE && !includeTentativeEra) {
-            return new EraRules(startDateAndEras, firstTentativeIdx);
+            return new EraRules(startDates, eraCodes, numErasNonTentative);
```
markusicu:
```suggestion
            // Trade-off:
            // This avoids shrinking/reallocating the arrays, but
            // wastes a little bit of space because the slots for the tentative eras are unused.
            // It also causes startDates.length != numEras, so EraRules need the numEras field.
            return new EraRules(startDates, eraCodes, numErasNonTentative);
```

icu4c/source/i18n/erarules.cpp
```diff
@@ -127,12 +129,10 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
     int32_t numEras = ures_getSize(rb.getAlias());
     int32_t firstTentativeIdx = MAX_INT32;
 
-    LocalMemory<int32_t> startDates(static_cast<int32_t *>(uprv_malloc(numEras * sizeof(int32_t))));
-    if (startDates.isNull()) {
-        status = U_MEMORY_ALLOCATION_ERROR;
+    UVector eraStartDates(numEras, status);
```
markusicu: did you mean UVector32?

icu4c/source/i18n/erarules.cpp
```diff
@@ -213,11 +224,31 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
         }
     }
 
+    // Now for just the eras we have rules for, make parallel arrays of startDates and eraCodes
+    LocalMemory<int32_t> startDates(static_cast<int32_t *>(uprv_malloc(numEras * sizeof(int32_t))));
+    LocalMemory<int32_t> eraCodes(static_cast<int32_t *>(uprv_malloc(numEras * sizeof(int32_t))));
```
markusicu: Idea:
If you computed firstTentativeIdx / numErasNonTentative on the fly while reading the data, then you could allocate exactly as much as you need, and not fill in the unused tentative slots, right?

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -116,68 +128,100 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
+        // Now for just the eras we have rules for, make parallel arrays of startDates and eraCodes
+        int[] startDates = new int[numEras];
+        int[] eraCodes = new int[numEras];
+        int startDateIdx = 0;
+        int numErasNonTentative = numEras;
+        for (int eraIdx = 0; eraIdx < eraStartDates.size(); eraIdx++) {
+            if (isSet(eraStartDates.get(eraIdx).intValue())) {
+                if (eraIdx >= firstTentativeIdx) {
+                    numErasNonTentative = startDateIdx;
+                }
```
markusicu: If there are multiple slots starting from firstTentativeIdx, aren't you setting numErasNonTentative to the last of them instead of the first of them?

pedberg-icu: Yes, thanks,  I did that in too much of a rush. I have already been rewriting to fix this and make other improvements, coming soon.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Yes, I went to a single trimmed array of startDates: It eliminates any tentativeEras if includeTentativeEra is false, and also deletes initial unset values, keeping a separate minEras value that gets added to the index to yield the era code. This simplified things a great deal.

icu4c/source/i18n/erarules.cpp
pedberg-icu: I cannot do that when filling the initial eraStartDates ArrayList/UVector because we may not know that an era is tentative until after we add it to the array (depending on the order in which we get the various subelements for the era bundle). But I do now trim the tentative eras from the ArrayList/UVector before converting it to the final int array if includeTentativeEras is false.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Fixed in the latest code.

icu4c/source/i18n/erarules.cpp
pedberg-icu: Indeed.

pedberg-icu:
@markusicu Rewrote again to use a single trimmed array of startDates (starts at first non-zero entry, trims any tentativeEras if includeTentativeEras is false, and keeps a minEra value so that minEra + index in startDates yields the eraCode). Much simpler this way. PTAL, thanks for all of the comments.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -56,11 +60,19 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
                 throw new ICUException("Era rule key:" + eraIdxStr + " in era rule data for " + calType.getId()
-                        + " must be in range [0, " + (numEras - 1) + "]");
+                        + " must be > 0");
             }
-            if (isSet(startDates[eraIdx])) {
+            if (eraIdx + 1 > eraStartDates.size()) {
+                eraStartDates.ensureCapacity(eraIdx + 1); // needed only to minimize expansions
+                // Fill in zeros for all added slots
+                while (eraStartDates.size() < eraIdx + 1) {
+                    eraStartDates.add(Integer.valueOf(0));
```
markusicu: let's use auto-boxing
```suggestion
                    eraStartDates.add(0);
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -78,7 +90,7 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
                                 "Invalid era rule date data:" + Arrays.toString(fields) + " in era rule data for "
                                 + calType.getId());
                     }
-                    startDates[eraIdx] = encodeDate(fields[0], fields[1], fields[2]);
+                    eraStartDates.set(eraIdx, Integer.valueOf(encodeDate(fields[0], fields[1], fields[2])));
```
markusicu:
```suggestion
                    eraStartDates.set(eraIdx, encodeDate(fields[0], fields[1], fields[2]));
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -56,11 +60,19 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
             } catch (NumberFormatException e) {
                 throw new ICUException("Invalid era rule key:" + eraIdxStr + " in era rule data for " + calType.getId());
             }
-            if (eraIdx < 0 || eraIdx >= numEras) {
+            if (eraIdx < 0) {
                 throw new ICUException("Era rule key:" + eraIdxStr + " in era rule data for " + calType.getId()
-                        + " must be in range [0, " + (numEras - 1) + "]");
+                        + " must be > 0");
```
markusicu: Judging from the code condition this should be
```suggestion
                        + " must be >= 0");
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -97,7 +109,7 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
                 if (hasEnd) {
                     // The islamic calendars now have an end-only rule for the
                     // second (and final) entry; basically they are in reverse order.
-                    startDates[eraIdx] = MIN_ENCODED_START;
+                    eraStartDates.set(eraIdx, Integer.valueOf(MIN_ENCODED_START));
```
markusicu:
```suggestion
                    eraStartDates.set(eraIdx, MIN_ENCODED_START);
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -116,102 +128,149 @@ public static EraRules getInstance(CalType calType, boolean includeTentativeEra)
...
+        int startDate = 0;
+        if (eraCode >= minEra) {
+            int startIdx = eraCode - minEra;
+            if (startIdx < startDates.length) {
+                startDate = startDates[startIdx];
+            }
         }
-        int[] fields = decodeDate(startDates[eraIdx], null);
-        return fields[0];
+        if (isSet(startDate)) {
```
markusicu, line 218: optional: share code with getStartDate()? move common code into a getStartDateImpl() which returns the raw startDate int? (same in C++)

icu4c/source/i18n/erarules.cpp
```diff
@@ -141,16 +140,27 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
...
-        if (isSet(startDates[eraIdx])) {
+        if (eraIdx + 1 > eraStartDates.size()) {
+            eraStartDates.ensureCapacity(eraIdx + 1, status); // needed only to minimize expansions
+            if (U_FAILURE(status)) {
+                return nullptr;
+            }
+            // Fill in 0 for all added slots (else they are undefined)
+            while (eraStartDates.size() < eraIdx + 1) {
+                eraStartDates.addElement(0, status);
+            }
```
markusicu:
```suggestion
            // Fill in 0 for all added slots (else they are undefined)
            while (eraStartDates.size() < eraIdx + 1) {
                eraStartDates.addElement(0, status);
            }
            if (U_FAILURE(status)) {
                return nullptr;
            }
```

icu4c/source/i18n/erarules.cpp
```diff
@@ -312,30 +363,12 @@ void EraRules::initCurrentEra() {
...
-            eraIdx--;
-        }
-        // Note: current era could be before the first era.
-        // In this case, this implementation returns the first era index (0).
-    }
-    currentEra = eraIdx;
+    // Now that getEraCode no longer depends on currentEra, we can just do this:
+    currentEra = getEraCode(year, month0 + 1 /* changes to 1-base */, dom, ec);
+    if (U_FAILURE(ec)) currentEra = minEra;
+    return;
```
markusicu:
```suggestion
```

icu4c/source/i18n/erarules.cpp
```diff
@@ -312,30 +363,12 @@ void EraRules::initCurrentEra() {
...
-            }
-            eraIdx--;
-        }
-        // Note: current era could be before the first era.
-        // In this case, this implementation returns the first era index (0).
-    }
-    currentEra = eraIdx;
+    // Now that getEraCode no longer depends on currentEra, we can just do this:
+    currentEra = getEraCode(year, month0 + 1 /* changes to 1-base */, dom, ec);
+    if (U_FAILURE(ec)) currentEra = minEra;
```
markusicu:
```suggestion
    if (U_FAILURE(ec)) {
        currentEra = minEra;
    }
```

icu4c/source/i18n/erarules.cpp
```diff
@@ -312,30 +363,12 @@ void EraRules::initCurrentEra() {
     int32_t year, mid;
     int8_t  month0, dom;
     Grego::timeToFields(localMillis, year, month0, dom, mid, ec);
+    currentEra = minEra;
     if (U_FAILURE(ec)) return;
```
markusicu:
```suggestion
    if (U_FAILURE(ec)) { return; }
```

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Decided to leave that as a future enhancement :-), but thanks. I hope to get this in soon to unblock Shane. Sorry for the muddled code earlier.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Done

icu4c/source/i18n/erarules.cpp
pedberg-icu: Done

icu4c/source/i18n/erarules.cpp
pedberg-icu: Done

icu4c/source/i18n/erarules.cpp
pedberg-icu: Done

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
pedberg-icu: Yes, of course, done

icu4c/source/i18n/erarules.cpp
pedberg-icu: Done

pedberg-icu:
@markusicu Fixed the last small stuff, except deferred extracting 5 lines of common code as getStartDateImpl() to a future update of this code.

markusicu:
Thanks, changes lgtm.

The Java 17 CI check failed:
```
DateFormatRoundTripTest.TestDateFormatRoundTrip:74->_test:141->_test:259->AbstractTestLog.errln:50 Pattern: M/d/yy, h:mm a failed to match; ms = -14541168992007
```

I don't know why Java 17 would fail when Java 11 & 21 are good.
Maybe just try to rerun?

sffc:
Thank you so much for your work here and on #3588. It seems like @markusicu and @richgillam are doing the code reviews. Do you need my attention, too?

pedberg-icu:
@markusicu The Java 17 CI test passed when I re-ran it. And then I squashed, and tests re-ran, and it passed again.

pedberg-icu:
> Thank you so much for your work here and on #3588. It seems like @markusicu and @richgillam are doing the code reviews. Do you need my attention, too?

@sffc Thanks, I think there has already been enough review of this.

pedberg-icu:
@markusicu or @richgillam, please re-stamp (Markus was happy with latest changes but then I needed to squash), then I can merge to unblock Shane's CLDR work, thanks!

markusicu (approved):

PR #3596, opened 2025-08-15
ICU-23005 Replace use of CharString with Locale to simplify the code

roubert:
This brings a lot of the code that uses `LocaleBased` back to how it was before the refactoring of ICU-23000, accessing the `validLocale` and `actualLocale` fields directly instead of through `LocaleBased` helper functions that now no longer are necessary.

This also makes it possible to remove the `U_LOCALE_BASED()` macro and `LocaleBased` then ends up becoming a trivial class that now contains nothing more than two static helper functions.

The purpose of this PR is to simplify the code, but it should also not have any negative performance impact so with the `/usr/bin/time` test from ICU-23000 averaged over 10 runs it looks like this, with a dramatic change in the user / system time division but no noticeable change in wall time and a tiny reduction in memory consumption:

|| before | after | _ratio_ |
| :--- | ---: | ---: | ---: |
| User time (seconds) | 69.22 | 60.04 | 86.73% |
| System time (seconds) | 1.98 | 9.96 | 504.00% |
| Percent of CPU this job got | 266.80% | 261.20% | 97.90% |
| Elapsed (wall clock) time (h:mm:ss or m:ss) | 00:26.63 | 00:26.75 | 100.43% |
| Average shared text size (kbytes) | 0 | 0 ||
| Average unshared data size (kbytes) | 0 | 0 ||
| Average stack size (kbytes) | 0 | 0 ||
| Average total size (kbytes) | 0 | 0 ||
| Maximum resident set size (kbytes) | 225148 | 224944 | 99.91% |
| Average resident set size (kbytes) | 0 | 0 ||
| Major (requiring I/O) page faults | 0 | 0 ||
| Minor (reclaiming a frame) page faults | 44183 | 44305 | 100.28% |
| Voluntary context switches | 461694 | 1436460 | 311.13% |
| Involuntary context switches | 3291 | 3898 | 118.44% |
| Swaps | 0 | 0 ||
| File system inputs | 0 | 0 ||
| File system outputs | 0 | 0 ||
| Socket messages sent | 0 | 0 ||
| Socket messages received | 0 | 0 ||
| Signals delivered | 0 | 0 ||
| Page size (bytes) | 4096 | 4096 | 100.00% |
| Exit status | 0 | 0 ||

#### Checklist
- [x] Required: Issue filed: ICU-23005

### DIFF
--- a/icu4c/source/common/locbased.cpp
+++ b/icu4c/source/common/locbased.cpp
@@ -11,85 +11,36 @@
 **********************************************************************
 */
 #include "locbased.h"
-#include "cstring.h"
-#include "charstr.h"
+#include "uresimp.h"
 
 U_NAMESPACE_BEGIN
 
-Locale LocaleBased::getLocale(const CharString* valid, const CharString* actual,
-                              ULocDataLocaleType type, UErrorCode& status) {
-    const char* id = getLocaleID(valid, actual, type, status);
-    return Locale(id != nullptr ? id : "");
-}
-
-const char* LocaleBased::getLocaleID(const CharString* valid, const CharString* actual,
+const Locale& LocaleBased::getLocale(const Locale& valid, const Locale& actual,
                                      ULocDataLocaleType type, UErrorCode& status) {
     if (U_FAILURE(status)) {
-        return nullptr;
+        return Locale::getRoot();
     }
 
     switch(type) {
     case ULOC_VALID_LOCALE:
-        return valid == nullptr ? "" : valid->data();
+        return valid;
     case ULOC_ACTUAL_LOCALE:
-        return actual == nullptr ? "" : actual->data();
+        return actual;
     default:
         status = U_ILLEGAL_ARGUMENT_ERROR;
+        return Locale::getRoot();
+    }
+}
+
+const char* LocaleBased::getLocaleID(const Locale& valid, const Locale& actual,
+                                     ULocDataLocaleType type, UErrorCode& status) {
+    const Locale& locale = getLocale(valid, actual, type, status);
+
+    if (U_FAILURE(status)) {
         return nullptr;
     }
-}
 
-void LocaleBased::setLocaleIDs(const CharString* validID, const CharString* actualID, UErrorCode& status) {
-    setValidLocaleID(validID, status);
-    setActualLocaleID(actualID,status);
-}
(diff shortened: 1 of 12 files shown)


icu4c/source/i18n/calendar.cpp
```diff
@@ -702,7 +702,9 @@ fAreAllFieldsSet(false),
 fAreFieldsVirtuallySet(false),
 fLenient(true),
 fRepeatedWallTime(UCAL_WALLTIME_LAST),
-fSkippedWallTime(UCAL_WALLTIME_LAST)
+fSkippedWallTime(UCAL_WALLTIME_LAST),
+validLocale(Locale::getRoot()),
```
markusicu, line 706: question: Could we move these into initializers in the class declarations?

icu4c/source/common/locbased.cpp
```diff
@@ -11,85 +11,36 @@
...
     }
-}
 
-bool LocaleBased::equalIDs(const CharString* left, const CharString* right) {
-    // true if both are nullptr
-    if (left == nullptr && right == nullptr) return true;
-    // false if only one is nullptr
-    if (left == nullptr || right == nullptr) return false;
-    return *left == *right;
+    return locale == Locale::getRoot() ? kRootLocaleName : locale.getName();
```
markusicu, line 43: kRootLocaleName is "root".
I believe that the current implementation just returns "", which I assume Locale::getRoot().getName() also returns, right?

markusicu (commented):
very nice!

icu4c/source/i18n/calendar.cpp
roubert: That should be possible (but would be out-of-scope for this PR).

icu4c/source/common/locbased.cpp
roubert: ICU4C uses two different strings to represent the root locale depending on context, which is why this expression here is needed to bridge between the two. The resource code uses the string "root" (`kRootLocaleName`) while the locale code uses the empty string and as `LocaleBased` belongs to the resource code this function currently returns the string "root" and needs to continue to do so for this PR to not change behaviour in any way.

FrankYFTang:
Thanks for working on this! Great Job!

PR #3597, opened 2025-08-18
ICU-23178 Always set to bogus if allocation fails when modifying Locale

roubert:
If any memory allocation error occurs while a `Locale` object is being modified this leaves the object in an unpredictable state (in contrast to other kinds of errors, like illegal argument errors, where the same input always leaves the object in the same state). When this happens, the safest path forward is therefore to set the object to bogus.

#### Checklist
- [x] Required: Issue filed: ICU-23178


roubert:
> Please add unit tests.

Does ICU4C have some mechanism for simulating memory allocation errors in unit tests?

PR #3598, opened 2025-08-19
ICU-20392 Simplify code by removing Locale::initBaseName()

roubert:
Whatever the purpose of this helper function originally was, subsequent refactorings have now made it obsolete.

It requires both less code and less processing to assign `baseName` together with the other data fields, instead of starting over and searching for the `@` character once again and possibly having to move payload from `Nest` to `Heap`.

#### Checklist

ALLOW_MANY_COMMITS=true


icu4c/source/common/locid.cpp
```diff
@@ -1994,21 +1994,20 @@ Locale& Locale::init(StringPiece localeID, UBool canonicalize)
             U_ASSERT(fullNameBuffer.isEmpty());
             nest.init(language, script, region, variantBegin);
         } else {
-            std::unique_ptr<Heap>& heap = payload.emplace<std::unique_ptr<Heap>>(
-                std::make_unique<Heap>(language, script, region, variantBegin));
```
markusicu: About the first commit: It looks fine, but I don't understand `Bugfix: Don't access old std::variant<> data after emplace().` -- where was the access to old std::variant data?

roubert: The code that's being replaced does this:

```
⋯
Nest& nest = payload.emplace<Nest>();
⋯
char* fullName = nest.baseName;
⋯
std::unique_ptr<Heap>& heap = payload.emplace<std::unique_ptr<Heap>>(
    std::make_unique<Heap>(language, script, region, variantBegin));
⋯
heap->fullName = {fullName, static_cast<std::string_view::size_type>(length)};
⋯
```

At the end there, it reads from a pointer into a `Nest` object that has been replaced with a `Heap` object in the `std::variant<> payload` and this just happens to work because the data hasn't actually been overwritten yet.

The correct thing to do is to instead read the data before calling `emplace()`.

markusicu: Thanks, I see. You have to read a lot of code before the changes to see where fullName comes from...

PR #3599, opened 2025-08-20
ICU-23184 fix NFRuleSet constructor to return error to avoid null deref later

FrankYFTang:
#### Checklist
- [X] Required: Issue filed: ICU-23184

The fuzzer generated test data is very large (4300 lenght unicode string) . Not sure how to minimize it. Depends on the fuzzer to  test.

### DIFF
--- a/icu4c/source/test/intltest/itrbnf.cpp
+++ b/icu4c/source/test/intltest/itrbnf.cpp
@@ -85,6 +85,7 @@ void IntlTestRBNF::runIndexedTest(int32_t index, UBool exec, const char* &name, 
         TESTCASE(33, TestInfiniteRecursion);
         TESTCASE(34, testOmissionReplacementWithPluralRules);
         TESTCASE(35, TestNullDereferenceWRITE23149);
+        TESTCASE(36, TestNullDereferenceREAD23184);
 #else
         TESTCASE(0, TestRBNFDisabled);
 #endif
@@ -2761,6 +2762,23 @@ IntlTestRBNF::TestNullDereferenceWRITE23149() {
    UErrorCode status = U_ZERO_ERROR;
    // The following call should not crash
    icu::RuleBasedNumberFormat rbfmt(test, Locale("en"), perror, status);
+}
+
+void
+IntlTestRBNF::TestNullDereferenceREAD23184() {
+    icu::Formattable result;
+    UParseError perror;
+    UErrorCode status = U_ZERO_ERROR;
+    icu::RuleBasedNumberFormat rbfmt(u"x00:>%>>;%:;<0<<", Locale::getUS(), perror, status);
+    if (U_SUCCESS(status)) {
+       errln("Construct \"x00:>%%>>;%%:;<0<<\" should get error");
+    }
+
+    status = U_ZERO_ERROR;
+    icu::RuleBasedNumberFormat rbfmt2(u"x00:>%>>;%;<0<<", Locale::getUS(), perror, status);
+    if (U_SUCCESS(status)) {
+       errln("Construct \"x00:>%%>>;%%;<0<<\" should get error");
+    }
 }
 
 /* U_HAVE_RBNF */
--- a/icu4c/source/i18n/nfrs.cpp
+++ b/icu4c/source/i18n/nfrs.cpp
@@ -164,9 +164,11 @@ NFRuleSet::NFRuleSet(RuleBasedNumberFormat *_owner, UnicodeString* descriptions,
     // and delete it from the description
     if (description.charAt(0) == gPercent) {
         int32_t pos = description.indexOf(gColon);
-        if (pos == -1) {
+        // if there are no name or the name is "%".
+        if (pos < 2) {
             // throw new IllegalArgumentException("Rule set name doesn't end in colon");
             status = U_PARSE_ERROR;
+            return;
         } else {
             name.setTo(description, 0, pos);
             while (pos < description.length() && PatternProps::isWhiteSpace(description.charAt(++pos))) {
@@ -180,6 +182,7 @@ NFRuleSet::NFRuleSet(RuleBasedNumberFormat *_owner, UnicodeString* descriptions,
     if (description.isEmpty()) {
         // throw new IllegalArgumentException("Empty rule set description");
         status = U_PARSE_ERROR;
+        return;
     }
 
     fIsPublic = name.indexOf(gPercentPercent, 2, 0) != 0;
@@ -220,6 +223,9 @@ NFRuleSet::parseRules(UnicodeString& description, UErrorCode& status)
         }
         currentDescription.setTo(description, oldP, p - oldP);
         NFRule::makeRules(currentDescription, this, rules.last(), owner, rules, status);
+        if (U_FAILURE(status)) {
+            return;
+        }
         oldP = p + 1;
     }
 
@@ -244,6 +250,9 @@ NFRuleSet::parseRules(UnicodeString& description, UErrorCode& status)
             // same as the preceding rule's base value in fraction
             // rule sets)
             rule->setBaseValue(defaultBaseValue, status);
+            if (U_FAILURE(status)) {
+                return;
+            }
         }
         else {
             // if it's a regular rule that already knows its base value,
--- a/icu4c/source/i18n/rbnf.cpp
+++ b/icu4c/source/i18n/rbnf.cpp
@@ -1592,6 +1592,8 @@ RuleBasedNumberFormat::init(const UnicodeString& rules, LocalizationInfo* locali
             fRuleSets[curRuleSet] = new NFRuleSet(this, ruleSetDescriptions, curRuleSet, status);
             if (fRuleSets[curRuleSet] == nullptr) {
                 status = U_MEMORY_ALLOCATION_ERROR;
+            }
+            if (U_FAILURE(status)) {
                 return;
             }
             ++curRuleSet;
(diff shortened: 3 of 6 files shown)


icu4c/source/i18n/nfsubs.cpp
```diff
@@ -1306,6 +1306,9 @@ NumeratorSubstitution::doParse(const UnicodeString& text,
 
         while (workText.length() > 0 && workPos.getIndex() != 0) {
             workPos.setIndex(0);
+            if (getRuleSet() == nullptr) {
```
aheninger: Is this fix simply papering-over some other underlying problem that should have been detected and handled earlier? Is bailing out at this point safe, or are things left in some inconsistent state that could cause further problems? I think some deeper analysis of what actually happened with this fuzz test case is needed.

grhoten: How is this possible? A test would help.

The way that this is written is harder to read. Perhaps the following would be clearer.

```
            if (getRuleSet() != nullptr) {
                getRuleSet()->parse(workText, workPos, 1, nonNumericalExecutedRuleMask, recursionCount, temp); // parse zero or nothing at all
            }
            // else consider it a parse failure.
```

grhoten: Also the Java code needs to be kept in sync. This nullability change should apply there too.

grhoten: If you have rules and a phrase that trigger this issue, it would help to see it.

FrankYFTang:
ok, I reduced the 9600 char test case found by fuzzer and write a simple test case to cause the same issue
```
diff --git a/icu4c/source/test/intltest/itrbnf.cpp b/icu4c/source/test/intltest/itrbnf.cpp
index 5cd837ef17c..db8d63eae0e 100644
--- a/icu4c/source/test/intltest/itrbnf.cpp
+++ b/icu4c/source/test/intltest/itrbnf.cpp
@@ -85,6 +85,7 @@ void IntlTestRBNF::runIndexedTest(int32_t index, UBool exec, const char* &name,
         TESTCASE(33, TestInfiniteRecursion);
         TESTCASE(34, testOmissionReplacementWithPluralRules);
         TESTCASE(35, TestNullDereferenceWRITE23149);
+        TESTCASE(36, TestNullDereferenceREAD23184);
 #else
         TESTCASE(0, TestRBNFDisabled);
 #endif
@@ -2763,6 +2764,18 @@ IntlTestRBNF::TestNullDereferenceWRITE23149() {
    icu::RuleBasedNumberFormat rbfmt(test, Locale("en"), perror, status);
 }
 
+void
+IntlTestRBNF::TestNullDereferenceREAD23184() {
+    UnicodeString str(u"x00:>%>>;%:;<0<<");
+    icu::Formattable result;
+    UParseError perror;
+    icu::RuleBasedNumberFormat rbfmt(str, Locale::getUS(), perror, status);
+    if (U_SUCCESS(status)) {
+       rbfmt.parse(str, result, status);
+    }
+}
+
 /* U_HAVE_RBNF */
 #else
 
diff --git a/icu4c/source/test/intltest/itrbnf.h b/icu4c/source/test/intltest/itrbnf.h
index bd36336b5ec..5a01ab045d7 100644
--- a/icu4c/source/test/intltest/itrbnf.h
+++ b/icu4c/source/test/intltest/itrbnf.h
@@ -166,6 +166,7 @@ class IntlTestRBNF : public IntlTest {
     void TestInfiniteRecursion();
     void testOmissionReplacementWithPluralRules();
     void TestNullDereferenceWRITE23149();
+    void TestNullDereferenceREAD23184();
 
 protected:
     virtual void doTest(RuleBasedNumberFormat* formatter, const char* const testData[][2], UBool testParsing);
```

icu4c/source/i18n/nfsubs.cpp
FrankYFTang: done. Add java and C++ unit tests. Fix both java and c++

FrankYFTang:
PTAL

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -2763,6 +2764,18 @@ IntlTestRBNF::TestNullDereferenceWRITE23149() {
    icu::RuleBasedNumberFormat rbfmt(test, Locale("en"), perror, status);
 }
 
+void
+IntlTestRBNF::TestNullDereferenceREAD23184() {
+    UnicodeString str(u"x00:>%>>;%;<0<<");
```
grhoten: What is this rule monstrosity!?

It's worth pointing out that ICU4J returns "Rule set name doesn't end in colon". So we shouldn't even be getting to doParse. The fix should be earlier when parsing the rules. The ICU4C and ICU4J implementations should have parity and the same behavior.

The nameless rule starts with a non-number x00. It's an awkward rule. Normally, it could be x.x, 0.x, x.0, or any variation where the period is replaced by the region specific decimal separator. I guess it could be 0, if 0 is your decimal separator for a x.0 rule, which seems logically awkward.

`>%>>` is an incomplete rule because % is an incomplete rule name, but perhaps it's instead a DecimalFormat. If we get to here, I suspect that this also should have been detected as a rule parse error. When it has %, it should have been referencing a rule name, but also I think it's trying to reference part of a previous rule that doesn't exist.

Value 0 is literally a `%`. That seems valid but confusing.

Value 1 is `<0<<`. I suspect that this rule could be causing a problem. I think it's saying to use some of the previous rule from value 1, and use the decimal format.

grhoten:
Thanks for the test, it really helps to understand the issue at hand.

icu4c/source/test/intltest/itrbnf.cpp
FrankYFTang: I agree with you we should find a better fix. The problem is how and where the fix should be. I am not sure where should we place the check to determine this rule is incorrect.

FrankYFTang:
> The current proposed changes to doParse seem incomplete. If getRuleSet() is null, it should be returning U_INTERNAL_PROGRAM_ERROR or something like a NullPointerException.

The problem is none of the following functions on the stack take UErrorCode as parameter
```
    #0 0x56cf6ed25227 in icu_78::NFRuleSet::parse(icu_78::UnicodeString const&, icu_78::ParsePosition&, double, unsigned int, int, icu_78::Formattable&) const icu/icu4c/source/i18n/nfrs.cpp:724:13
    #1 0x56cf6ed3c864 in icu_78::NumeratorSubstitution::doParse(icu_78::UnicodeString const&, icu_78::ParsePosition&, double, double, signed char, unsigned int, int, icu_78::Formattable&) const icu/icu4c/source/i18n/nfsubs.cpp:1309:27
    #2 0x56cf6ed31657 in icu_78::NFRule::matchToDelimiter(icu_78::UnicodeString const&, int, double, icu_78::UnicodeString const&, icu_78::ParsePosition&, icu_78::NFSubstitution const*, unsigned int, int, double) const icu/icu4c/source/i18n/nfrule.cpp:1247:38
    #3 0x56cf6ed30477 in icu_78::NFRule::doParse(icu_78::UnicodeString const&, icu_78::ParsePosition&, signed char, double, unsigned int, int, icu_78::Formattable&) const icu/icu4c/source/i18n/nfrule.cpp:1056:32
    #4 0x56cf6ed2577b in icu_78::NFRuleSet::parse(icu_78::UnicodeString const&, icu_78::ParsePosition&, double, unsigned int, int, icu_78::Formattable&) const icu/icu4c/source/i18n/nfrs.cpp:768:39
    #5 0x56cf6ebe7499 in icu_78::RuleBasedNumberFormat::parse(icu_78::UnicodeString const&, icu_78::Formattable&, icu_78::ParsePosition&) const icu/icu4c/source/i18n/rbnf.cpp:1365:17
```

FrankYFTang:
So. I believe what @grhoten  said is the right fix is the status after the following code should set to error 
icu::RuleBasedNumberFormat rbfmt(u"x00:>%>>;%;<0<<", Locale::getUS(), perror, status);

so the code should not even call into parse since the RuleBasedNumberFormat is in a bogus state.

@grhoten , to help me to figure out the right fix, could you point out when the constructor parse the rule, in which point it should return error?

FrankYFTang:
could you suggest some check in the constructor that we should perform in the constructor to return an error code for this string?

grhoten:
I'd probably debug this new test, and set a breakpoint in NFRule::parseRuleDescriptor (nfrule.cpp). There may be untested conditions there, since the rules normally have a colon.

grhoten:
That function does return a U_PARSE_ERROR in certain states.

FrankYFTang:
> NFRule::parseRuleDescriptor

NFRule::parseRuleDescriptor got called twice, first time with description
"Inf: ∞" and second time "NaN: NaN"
I think one came from
RuleBasedNumberFormat::initializeDefaultNaNRule()
and the other came from 
RuleBasedNumberFormat::initializeDefaultInfinityRule()

grhoten:
It should also be called 2 more times for these 2 rules.
```
%;<0<<
```

FrankYFTang:
I fond a fix for c++ still try to figure out how to solve Java

FrankYFTang:
but I have not update the PR to include my last C++ fix yet. Just hold the review till I repush

FrankYFTang:
@grhoten @aheninger Please review the code. Now it return error in constructor

FrankYFTang:
@grhoten @aheninger @richgillam ping

grhoten (approved):
The new error handling seems reasonable to me.

PR #3600, opened 2025-08-20
ICU-23056 CLDR 48 alpha0 integration

DraganBesevic:
#### Checklist

Alpha0 integration for CLDR 48

Known issues:
~~[ICU-23182](https://unicode-org.atlassian.net/browse/ICU-23182), (Formatting for Japanese calendars)~~ Fixed in commit 4
[ICU-23183](https://unicode-org.atlassian.net/browse/ICU-23183)  (Time Zone round trip test)
[ICU-23185](https://unicode-org.atlassian.net/browse/ICU-23185)  (hz and hv patterns)
~~[ICU-23186](https://unicode-org.atlassian.net/browse/ICU-23186) (Japanese calendars fails round trip unit tests)~~ Fixed in commit 4
[ICU-23187](https://unicode-org.atlassian.net/browse/ICU-23187) (TestPersonNames test fail for kk_Arab and shn)
    - note, kk_Arab and shn are new locales not yet in ICU (until config file is updated)
[ICU-23188](https://unicode-org.atlassian.net/browse/ICU-23188) (Compact decimal format adds extra RLM mark for ar-EG)

Note that commit 4 reworks the ICU4C code for loading eraNames in table format which was added in PR [#3588](https://github.com/unicode-org/icu/pull/3588). The following ticket has been filed to rework the ICU4J code in a similar way: [ICU-23197](https://unicode-org.atlassian.net/browse/ICU-23197).

[ICU-23182]: https://unicode-org.atlassian.net/browse/ICU-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23183]: https://unicode-org.atlassian.net/browse/ICU-23183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23185]: https://unicode-org.atlassian.net/browse/ICU-23185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23186]: https://unicode-org.atlassian.net/browse/ICU-23186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23187]: https://unicode-org.atlassian.net/browse/ICU-23187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23188]: https://unicode-org.atlassian.net/browse/ICU-23188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/test/intltest/dtifmtts.cpp
+++ b/icu4c/source/test/intltest/dtifmtts.cpp
@@ -1246,7 +1246,7 @@ void DateIntervalFormatTest::expect(const char** data, int32_t data_length) {
 
         const UnicodeString& oneSkeleton(ctou(data[i++]));
         
-        DateIntervalFormat* dtitvfmt = DateIntervalFormat::createInstance(oneSkeleton, loc, ec);
+        LocalPointer<DateIntervalFormat> dtitvfmt(DateIntervalFormat::createInstance(oneSkeleton, loc, ec));
         if (!assertSuccess("createInstance(skeleton) in expect", ec)) return;
         FieldPosition pos(FieldPosition::DONT_CARE);
         dtitvfmt->format(&dtitv, str.remove(), pos, ec);
@@ -1270,7 +1270,6 @@ void DateIntervalFormatTest::expect(const char** data, int32_t data_length) {
         logln("interval date:" + str + "\"" + locName + "\", "
                  + "\"" + datestr + "\", "
               + "\"" + datestr_2 + "\", " + oneSkeleton);
-        delete dtitvfmt;
     }
 }
 
--- a/icu4c/source/test/intltest/numfmtst.cpp
+++ b/icu4c/source/test/intltest/numfmtst.cpp
@@ -977,7 +977,8 @@ NumberFormatTest::TestCurrency()
         }
         if (strcmp(localeID, "gl_ES@currency=ESP") == 0 &&
                 logKnownIssue("CLDR-18901", "Problem with ❰NBTSP❱")) {
-                  continue;
+                    delete currencyFmt;
+                    continue;
         }
         currencyFmt->format(1150.50, s);
         if(s!=expected){


pedberg-icu:
@DraganBesevic getting a crash e.g. in the clang-cpp17 tests:
```
      IntlCalendarTest {
         ... 
         TestForceGannenNumbering {
intltest: smpdtfmt.cpp:1290: void icu_78::_appendSymbol(icu_78::UnicodeString &, int32_t, const icu_78::UnicodeString *, int32_t): Assertion `0 <= value && value < symbolsCount' failed.
```

Depending on what is going on here: Maybe skip this test with the logKnownIssue already covering Gannen formatting?

DraganBesevic:
> @DraganBesevic getting a crash e.g. in the clang-cpp17 tests:
> 
> ```
>       IntlCalendarTest {
>          ... 
>          TestForceGannenNumbering {
> intltest: smpdtfmt.cpp:1290: void icu_78::_appendSymbol(icu_78::UnicodeString &, int32_t, const icu_78::UnicodeString *, int32_t): Assertion `0 <= value && value < symbolsCount' failed.
> ```
> 
> Depending on what is going on here: Maybe skip this test with the logKnownIssue already covering Gannen formatting?

That's weird. I wonder why adding a known issue would cause to crash in CPP 17. I will skip it as it is already a known issue

mihnita:
I was able to reproduce it on an Ubuntu system at home.

I am trying to figure out what the cause is. But it is very puzzling, I see no reason for this to fail with the CLDR update and not fail without it.

Nothing in this PR can explain what is going on (just by looking at it).
I'm pretty sure that it will all be "obvious" once I figure out the cause :-)

mihnita:
The failure happens in [icu4c/source/i18n/smpdtfmt.cpp, line 1536](https://github.com/unicode-org/icu/blob/b86b7a59d9c4039bc2b8456ec48f646662d75d55/icu4c/source/i18n/smpdtfmt.cpp#L1536), in this call:
```
_appendSymbol(appendTo, value, fSymbols->fNarrowEras, fSymbols->fNarrowErasCount);
```

`_appendSymbol` fails the `U_ASSERT(0 <= value && value < symbolsCount);` ([line 1290](https://github.com/unicode-org/icu/blob/b86b7a59d9c4039bc2b8456ec48f646662d75d55/icu4c/source/i18n/smpdtfmt.cpp#L1290))
because `value` is 235, and `symbolsCount` is 232.

The locale failing is Japanese with alternate calendar, 235 is Heisei
(see `icu4c/source/data/locales/root.txt` in pull request)

I am not sure how to fix this, since it is not just a few entries added, the structure seems changed.
Inspect the `icu4c/source/data/locales/root.txt` (or any other locale file).
The eras changed from an array:
```
eras{
  abbreviated{                                                   
    "Taika (645–650)",
    "Hakuchi (650–671)",
    "Hakuhō (672–686)",
    "Shuchō (686–701)",
    ...
    "Heisei",
    ...
```

to a map:
```
eras{
  abbreviated{
    0{"Taika (645–650)"}
    1{"Hakuchi (650–671)"}
    10{"Tenpyō (729–749)"}
    ...
    235{"Heisei"}
    ...
```

mihnita:
Related?

3a66f8c5fec 2025-08-11 22:24:01 +0000 "ICU-23172 part 1: Convert era names from array to map (#3588):"

mihnita:
Why it does not fail on non-Linux?
I suppose it is about how U_ASSERT is implemented.

On Linux the test results in a core-dump.

For macOS that should be enabled, it does not happen by default:
https://nasa.github.io/trick/howto_guides/How-to-dump-core-file-on-MacOS.html

Also, it looks like the test suite (on Linux) recovers after a core dump and continues.
I don't know if it is intentional.
But that makes it kind of hard to find where the failure happens.
You get one code dump for one test, and everything continues.

mihnita:
For the ja.txt file the eras/abbreviated has 237 entries, but the eras/narrow has only 232 entries.
Which "rhymes" with the `symbolsCount` being 232 in `U_ASSERT`.

---

This is very likely related to the new bugs filed by Dragan for the Japanese calendar that only happen with the new data.

AEApple:
> For the ja.txt file the eras/abbreviated has 237 entries, but the eras/narrow has only 232 entries. Which "rhymes" with the `symbolsCount` being 232 in `U_ASSERT`.
> 
> This is very likely related to the new bugs filed by Dragan for the Japanese calendar that only happen with the new data.

Per @pedberg-icu - Only the 5 most recent eras are the ones that differ from the abbreviated. It seems likely that those are not being copied for some reason, and if there were true that would explain why the formatting is not using the era code for those. It's not clear where in the integration this is happening.

pedberg-icu:
> > For the ja.txt file the eras/abbreviated has 237 entries, but the eras/narrow has only 232 entries. Which "rhymes" with the `symbolsCount` being 232 in `U_ASSERT`.
> > This is very likely related to the new bugs filed by Dragan for the Japanese calendar that only happen with the new data.
> 
> Per @pedberg-icu - Only the 5 most recent eras are the ones that differ from the abbreviated. It seems likely that those are not being copied for some reason, and if there were true that would explain why the formatting is not using the era code for those. It's not clear where in the integration this is happening.

Yeah, it seems like inheritance is not working for the narrow Japanese eras using the new table format for the era names data. But I did run the ICU4C and ICU4J tests with data generated using the new table format, including with early eras deleted, and did not have these problems. I will try to take a look this weekend.

markusicu:
The CLDR structure for eras is changing, allowing sparse eras, so that early (inaccurate) Japanese eras can be removed (and other changes endorsed by the design group). Peter's recent PR prepared the eras-loading code so that it can handle either the old or the new data structure. He generated the new structure to test with, but didn't add the new data in the same PR (because the generation pulled in a lot of unrelated stuff), and I guess he didn't see these errors with his generated data.

I reviewed the new loading code, and it seemed fine. Maybe we overlooked something. I actually suspect that the formatting code goes directly into the array when it should use a getter -- there is a new minEras field, and the array omits those first entries. There are also conditions when future eras at the end are dropped.

And, I don't know how loading vs. using abbreviated vs. wide vs. narrow symbols works. Maybe something expects the arrays to has the same lengths, but if there is no abbreviated symbol then the formatting code should pick the narrow one, or vice versa.

pedberg-icu:
@DraganBesevic The ICU4C code for loading the era names in the new table format is not handling inheritance from root when there are fewer eras defined in the `ja` locale than there are in `root` (which is the case for narrow Japanese-calendar eras in `ja`). We need to figure out how to really fix this. In the meantime, we can temporarily fix this by copying the last 5 narrow japanese eras from `root` to `ja`; in cldr and cldr-staging these are
```
						<era type="232">M</era>
						<era type="233">T</era>
						<era type="234">S</era>
						<era type="235">H</era>
						<era type="236">R</era>
```
and in ICU4C data these are
```
                    232{"M"}
                    233{"T"}
                    234{"S"}
                    235{"H"}
                    236{"R"}
```

pedberg-icu:
@DraganBesevic Not sure what is going on with the GHA ICU4C / clang errors:
```
     "ulocaletst.c:259 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "ca") should not return error and should return empty string."
     "ulocaletst.c:265 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "co") should not return error and should return empty string."
```
Seems like for now that should be a logKnownIssue test skip. And with the japanese calendar additions we should be able to **_remove_** the logKnownIssue test skips related to [ICU-23182](https://unicode-org.atlassian.net/browse/ICU-23182) and [ICU-23186](https://unicode-org.atlassian.net/browse/ICU-23186) and make sure those tests are working.

[ICU-23182]: https://unicode-org.atlassian.net/browse/ICU-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23186]: https://unicode-org.atlassian.net/browse/ICU-23186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

DraganBesevic:
> @DraganBesevic Not sure what is going on with the GHA ICU4C / clang errors:
> 
> ```
>      "ulocaletst.c:259 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "ca") should not return error and should return empty string."
>      "ulocaletst.c:265 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "co") should not return error and should return empty string."
> ```
> 
> Seems like for now that should be a logKnownIssue test skip. And with the japanese calendar additions we should be able to **_remove_** the logKnownIssue test skips related to [ICU-23182](https://unicode-org.atlassian.net/browse/ICU-23182) and [ICU-23186](https://unicode-org.atlassian.net/browse/ICU-23186) and make sure those tests are working.

@pedberg-icu , which unit test is this? I can only see those strings in `./cintltst/ulocaletst.c` , but that part of  code is already marked as a known issue ICU-22459

The failures I see now all seems to point to memory leaks in dtifmtts.cpp, but looking at my changes I don't think I introduce those as I don't think I added any returns with leftover objects

[ICU-23182]: https://unicode-org.atlassian.net/browse/ICU-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23186]: https://unicode-org.atlassian.net/browse/ICU-23186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pedberg-icu:
> > @DraganBesevic Not sure what is going on with the GHA ICU4C / clang errors:
> > ```
> >      "ulocaletst.c:259 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "ca") should not return error and should return empty string."
> >      "ulocaletst.c:265 ulocale_getUnicodeKeywordValue("fa-u-nu-thai", "co") should not return error and should return empty string."
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Seems like for now that should be a logKnownIssue test skip. And with the japanese calendar additions we should be able to **_remove_** the logKnownIssue test skips related to [ICU-23182](https://unicode-org.atlassian.net/browse/ICU-23182) and [ICU-23186](https://unicode-org.atlassian.net/browse/ICU-23186) and make sure those tests are working.
> 
> @pedberg-icu , which unit test is this?

Sorry, my mistake, I was looking at the wrong thing, that ulocaletst.c thing was just a log of a known issue. The real problem seems to be a memory leak in DateFormat or DateFormatSymbols, probably related to handling the new map format. I probably need to take a look at that. I have a copy of your branch, I can debug that.


[ICU-23182]: https://unicode-org.atlassian.net/browse/ICU-23182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ICU-23186]: https://unicode-org.atlassian.net/browse/ICU-23186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pedberg-icu:
The reported leaks - at least many of them - seem to be in the DateFormatSymbols copy constructor, where it copies the era names data (lines 407-409 in i18n/dtfmtsym.cpp). But when I try on my system to build & run with asan enabled, I do not get any leak reports. I am looking at better ways of loading the era names data; however it would be nice if others could look at this in parallel, I am time-limited with family visiting etc., thanks.

pedberg-icu:
@DraganBesevic I reworked the ICU4C code that loads the new table-style eraNames so that it can load missing values from parent locales; this way we do not need to manually add the missing Japanese narrow eras. This also removes most of the problems that led to the logKnownIssue tests skips for ICU-23182 and ICU-23186, so I removed those skips (I also had to update the tests for the fixed Meiji era start date). Then I squashed commits 4-7 into a single part 4 commit since they were all about the Japanese calendar stuff.

Unfortunately we still have a problem with leaks, though these changes reduced the leaks by around 50%. Continuing to investigate that, hoping I can get some help with that from somebody. I cannot repro the leaks on my system (macOS with clang, asan enabled).

pedberg-icu:
> Unfortunately we still have a problem with leaks, though these changes reduced the leaks by around 50%. Continuing to investigate that, hoping I can get some help with that from somebody. I cannot repro the leaks on my system (macOS with clang, asan enabled).

Ah, the remaining leaks actually seem to be due to logKnownIssue test skips, not related to the eraNames loading code. Fixed.

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1950,6 +1909,51 @@ initField(UnicodeString **field, int32_t& length, CalendarDataSink &sink, CharSt
...
+                        }
+                    }
+                }
+            }
+            return;
+        }
+        status = U_MISSING_RESOURCE_ERROR;
+    }
+}
+
```
pedberg-icu, line 1956: The following ticket has been filed to rework the ICU4J code in a similar way: [ICU-23197](https://unicode-org.atlassian.net/browse/ICU-23197)

FrankYFTang:
please see breakage at 
https://github.com/unicode-org/icu/actions/runs/17279025326/job/49043094938

pedberg-icu:
> please see breakage at https://github.com/unicode-org/icu/actions/runs/17279025326/job/49043094938

@DraganBesevic It seems like there is an ICU4J exhaustive test failure in collation, may need a logKnownIssue test skip ( have not checked but there may be a problem with ti-ER index exemplars):
```
Error:  Failures: 
Error:    AlphabeticIndexTest.TestEmpty:326->AbstractTestLog.errln:50 Exception when creating AlphabeticIndex for:	ti-ER
```

FrankYFTang:
> > please see breakage at https://github.com/unicode-org/icu/actions/runs/17279025326/job/49043094938
> 
> @DraganBesevic It seems like there is an ICU4J exhaustive test failure in collation, may need a logKnownIssue test skip ( have not checked but there may be a problem with ti-ER index exemplars):
> 
> ```
> Error:  Failures: 
> Error:    AlphabeticIndexTest.TestEmpty:326->AbstractTestLog.errln:50 Exception when creating AlphabeticIndex for:	ti-ER
> ```

the issue is from ti_ER locale, Also have breakage on the following
```
[ERROR] Failures: 
[ERROR]   ICUResourceBundleTest.TestLocaleDisplayNames:736->AbstractTestLog.errln:50 Could not get native localized country for ti_ER
[ERROR]   ULocaleTest.TestDisplayNames:1123->checkName:1377->checkInclusion:1396->AbstractTestLog.errln:50 loc: ti_ER name 'af' does not contain language 'ኣፍሪካንስ'
[ERROR] Errors: 
[ERROR]   LocaleDataTest.TestExemplarSet:220 » IllegalState pool.res has a different checksum than this bundle
[ERROR]   LocaleDataTest.TestExemplarSet2:277 » IllegalState pool.res has a different checksum than this bundle
[ERROR]   LocaleDataTest.TestLocaleDisplayPattern:470 » IllegalState pool.res has a different checksum than this bundle
[INFO] 
[ERROR] Tests run: 3221, Failures: 2, Errors: 3, Skipped: 5

```

PR #3604, opened 2025-08-21
ICU-22851 A somewhat more structured UnicodeSet parser.

eggrobin:
DRAFT: Not quite the same sequence of calls to lookup/lookupMatcher. I think that is fine, but if there is an easy way to get exactly the old behaviour I might try that…

#### Checklist
- [x] Required: Issue filed: ICU-22851


eggrobin:
@markusicu Do we need a separate ticket for this restructuring, or should I do it as part of ICU-23179?

richgillam:
Wow, there's a lot here.  I ran out of time to read through the whole thing ad was starting to lose my mental acuity well before that, so I don't think I have anything useful to say.  I can try again next week if you don't get useful review feedback from other people.

I read all the way through the spec, though, and I think I managed to keep my brain going through that.  I think it looks really good-- is the plan to bring all the various implementations into complete conformance with that spec, and is this one of the first steps toward doing so?  That seems like a good idea.  I'm assuming you've thought through the backward-compatibility issues in doing so and everybody's comfortable with them?

I hope you'll permit me one stupid question about the spec: I'm assuming that the spec allows for sets that contain both individual code points and strings, right?  I did see a lot of verbiage in there about bracketed elements and how they can represent either single code points or strings and how you want to do special things with them when they're single code points.  I don't remember seeing much verbiage that talks about the behavior of unicode sets that contain both strings and single code points.  At least to me, the semantics of this kind of thing are far from obvious, especially if a set contains both types.  I remember a lot of this being discussed when various people were proposing changes to `UnicodeSet` earlier, but it seems like all of that needs to be covered in the spec, and I didn't really see it there.  Is that planned?  Am I wrong to be concerned about this?

macchiati:
I wouldn't think that performance would be affected significantly, but we
probably should check at least with a simple spot test.

eggrobin:
@markusicu I filed a new ticket for this. Can you take a look?

eggrobin:
> At least to me, the semantics of this kind of thing are far from obvious, especially if a set contains both types.

Good point (Mark’s earlier comment has examples of this non-obviousness). I will hammer at the draft some more…

markusicu:
Looks like @poulsbo is not yet in the icu-team and so I can't add him as a reviewer, although in some ways he would be the obvious choice ;-)

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+        const UErrorCode& errorCode() const{
+          return errorCode_;
+        }
 
-    // Recognized special forms for chars, sets: c-c s-s s&s
+        UnicodeString debugString() const {
+            UnicodeString result;
+            if (U_FAILURE(errorCode_)) {
+                result.append(u"Ill-formed token (");
+                result.append(UnicodeString::fromUTF8(u_errorName(errorCode_)));
```
markusicu:
```suggestion
                result.append(UnicodeString(u_errorName(errorCode_), -1, US_INV));
```
We haven't yet actively removed EBCDIC support, so `char *` strings cannot generally be assumed to be in ASCII.

(Assuming UTF-8 is ok in ifdef-ed debug code.)

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+                result.append(u"), possibly ");
+            }
+            result.append(category_names_[category_]);
+            result.append(u" '");
+            result.append(sourceText_);
+            result.append(u"'");
+            return result;
```
markusicu: optional: you could chain some of these calls

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
 // Implementation: Pattern parsing
 //----------------------------------------------------------------
 
-namespace {
+class UnicodeSet::Lexer {
```
markusicu, line 228: Are you planning on a follow-up PR to keep Java code parallel?

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+          uint32_t unicodeSetOptions,
+          const SymbolTable *const symbols)
+        : pattern_(pattern), parsePosition_(parsePosition), chars_(chars),
+          charsOptions_(RuleCharacterIterator::PARSE_VARIABLES | RuleCharacterIterator::PARSE_ESCAPES |
+                        ((unicodeSetOptions & USET_IGNORE_SPACE) != 0
+                             ? RuleCharacterIterator::SKIP_WHITESPACE
+                             : 0)),
+          symbols_(symbols) {}
+
+    class LexicalElement {
```
markusicu, line 242: optional: why make this a nested class instead of a sibling?

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+        // '[', named-element, and property-query cannot be disabled by stand-in.
+        if (first == u'[' || first == u'\\') {
+            const RuleCharacterIterator::Pos afterFirst = getPos();
+            // This could be a property-query or named-element.
+            const UChar32 second = chars_.next(charsOptions_ & ~(RuleCharacterIterator::PARSE_ESCAPES |
+                                                                 RuleCharacterIterator::SKIP_WHITESPACE),
+                                               unusedEscaped, errorCode);
+            if ((first == u'[' && second == u':') ||
+                (first == u'\\' && (second == u'p' || second == u'P' || second == u'N'))) {
+                if (second == u'N') {
```
markusicu, line 453: Could you please split this function into somewhat smaller chunks?
For example, pull out something like `nextPropertyQueryToken(first, second, ...)` from here to just before `// Not a property-query.` ?

markusicu (commented):
I finally started to look... I won't try to understand everything.

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+                                unusedEscaped, errorCode);
+
+                while (!chars_.atEnd() && U_SUCCESS(errorCode)) {
+                    const RuleCharacterIterator::Pos beforeLast = getPos();
+                    UChar32 lastUnescaped =
+                        chars_.next(charsOptions_ & ~(RuleCharacterIterator::PARSE_ESCAPES |
+                                                      RuleCharacterIterator::SKIP_WHITESPACE),
+                                    unusedEscaped, errorCode);
+                    if (penultimateUnescaped == u'\\') {
+                        if (lastUnescaped == 'N') {
```
markusicu: optional: how about a `nextNamedElementToken(...)` for both this if-body and the earlier `\N{...}` parser? (avoid duplication)

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
-                lastItem = 0;
-                op = 0;
-            }
-
-            if (op == u'-' || op == u'&') {
-                patLocal.append(op);
+        if (first == u'\\') {
+            // Now try to parse the escape.
+            chars_.setPos(before);
+            UChar32 codePoint = chars_.next(charsOptions_, unusedEscaped, errorCode);
```
markusicu, line 480: question: does this do the actual unescaping?

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+        // UnicodeSet ::= property-query | named-element
+        // NOTE(egg): For now, we ignore the work that the lexer did to find out where the
+        // property-query or named-element ended in order to retain the existing buggy behaviour of
+        // variables containing property queries.
+        lexer.getCharacterIterator().skipIgnored(lexer.charsOptions());
+        UnicodeSet propertyQuery;
+        propertyQuery.applyPropertyPattern(lexer.getCharacterIterator(), prettyPrintedPattern, ec);
+        U_UNICODESET_RETURN_IF_ERROR(ec);
+        // But now, we go back to our lexing and advance through the property-query or named-element as
+        // lexed.  If there was no error, the old and the new code should agree on the extent.
```
markusicu, line 743: After this PR lands, the reference to new vs. old code is no longer meaningful, is it?

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
-            lastChar = c;
-            lastItem = 1;
-            break;
+            prettyPrintedPattern.append(u']');
+        } else if (const UnicodeSet *const standIn = lexer.lookahead().standIn();
+                   standIn != nullptr) {
+                *this = *standIn;
+            this->_toPattern(rebuiltPat, /*escapeUnprintable=*/false);
+            lexer.advance();
+                return;
```
markusicu:
```suggestion
            *this = *standIn;
            this->_toPattern(rebuiltPat, /*escapeUnprintable=*/false);
            lexer.advance();
            return;
```

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+    for (;;) {
+        if (lexer.acceptSetOperator(u'&')) {
+            // Intersection ::= Restriction & UnicodeSet
+            rebuiltPat.append(u'&');
+            UnicodeSet rightHandSide;
+            rightHandSide.parseUnicodeSet(lexer, rebuiltPat, options, caseClosure, depth + 1, ec);
+            U_UNICODESET_RETURN_IF_ERROR(ec);
+            retainAll(rightHandSide);
+        } else if (lexer.lookahead().isSetOperator(u'-')) {
+            // Here the grammar requires two tokens of lookahead to figure out whether the - the operator
```
markusicu:
```suggestion
            // Here the grammar requires two tokens of lookahead to figure out whether the - is the operator
```
(I think...)

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+            // of a Difference or an UnescapedHyphenMinus in the enclosing Union.
+            if (lexer.lookahead2().isSetOperator(u']')) {
+                // The operator is actually an UnescapedHyphenMinus; terminate the Restriction before it.
```
markusicu: question: don't we need to add the minus as a set character?

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,796 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+    }
+    lexer.advance();
+    _appendToPat(rebuiltPat, first, /*escapeUnprintable=*/false);
+    if (!lexer.lookahead().isSetOperator(u'-')) {
+        // No operator,
+        // Elements ::= Element
+        add(first);
+        return;
     }
+    // Here the grammar requires two tokens of lookahead to figure out whether the - the operator
```
markusicu:
```suggestion
    // Here the grammar requires two tokens of lookahead to figure out whether the - is the operator
```

markusicu:
Do we still need some of the existing functions in uniset_props.cpp like isPerlOpen() and resemblesPropertyPattern()?

markusicu:
Do we still need `applyPropertyPattern()`? If so, can we rewrite it using this new parser?

icu4c/source/common/uniset_props.cpp
eggrobin: Not here, because we haven’t consumed it yet (it was the first lookahead character).
With this return statement, we pop to parseTerm which then pops to parseUnion, which will `acceptSetOperator(u'-')` and `add(u'-')`. I can note that in the comment.

icu4c/source/common/uniset_props.cpp
eggrobin: Yes.

icu4c/source/common/uniset_props.cpp
eggrobin: Because we don’t name it outside of Lexer.

In Java it might want to be a sibling, because there nested class have different semantics besides the name scoping.

icu4c/source/common/uniset_props.cpp
eggrobin: The `debugString` is only used from ifdef-ed out debug paths. ifdefed out its definition instead.

icu4c/source/common/uniset_props.cpp
eggrobin: No (ICU does not need it, so that would be an unnecessary copy; here we get away with a string_view to a chunk of the original pattern).

icu4c/source/common/uniset_props.cpp
eggrobin: Well, the whole comment from NOTE(egg) to here is hopefully transitional; if the TC agrees, I will make a subsequent PR to drop make the behaviour of variables less buggy, and then we will no longer match the old code. Until then I think this comment is necessary to describe why this is written in the bonkers way it is.

eggrobin:
> Do we still need `applyPropertyPattern()`?

For now we still call it, with a comment explaining why we do this silly thing:

https://github.com/unicode-org/icu/blob/381616cb0c3e3e357450e4e402528811e82415ac/icu4c/source/common/uniset_props.cpp#L752-L761

> If so, can we rewrite it using this new parser?

Yes, but not if we want to retain the current insane behaviour of variables:
https://github.com/unicode-org/icu/blob/7d3c247e90232381eb84cd494be298d947294c6a/icu4c/source/test/intltest/usettest.cpp#L1842-L1856

So I will get rid of it in a subsequent PR (assuming the TC accepts changing the handling of variables so that they fit in the grammar).

eggrobin:
> Do we still need some of the existing functions in uniset_props.cpp like isPerlOpen() and resemblesPropertyPattern()?

`resemblesPropertyPattern` is used by `resemblesPattern`, which is `@stable ICU 2.4`.


`isPerlOpen` is used by both `resemblesPropertyPattern` and `applyPropertyPattern`.

icu4c/source/common/uniset_props.cpp
markusicu: Ok. FYI, nesting in Java is fine as well; use `static` if the inner class does not need to have an `outer` pointer.

markusicu: Also FYI -- I was just thinking about readability / compact class definitions, but this is ok.

icu4c/source/common/uniset_props.cpp
eggrobin: Oops, I forgot about this one.

I will do that now.

icu4c/source/common/uniset_props.cpp
```diff
@@ -220,438 +223,799 @@ UBool UnicodeSet::resemblesPattern(const UnicodeString& pattern, int32_t pos) {
...
+            rightHandSide.parseUnicodeSet(lexer, rebuiltPat, options, caseClosure, depth + 1, ec);
+            U_UNICODESET_RETURN_IF_ERROR(ec);
+            retainAll(rightHandSide);
+        } else if (lexer.lookahead().isSetOperator(u'-')) {
+            // Here the grammar requires two tokens of lookahead to figure out whether the - is the operator
+            // of a Difference or an UnescapedHyphenMinus in the enclosing Union.
+            if (lexer.lookahead2().isSetOperator(u']')) {
+                // The operator is actually an UnescapedHyphenMinus; terminate the Restriction
+                // before it.  We return to parseTerm, which immediately returns to parseUnion,
+                // which will accept the - and it to *this.
```
markusicu:
```suggestion
                // which will accept the - and add it to *this.
```

markusicu:
aside from another typo and the optional refactoring suggestions, this lgtm

PR #2284, opened 2023-01-14
ICU-21374 Add a CFI build bot for ICU4C

FrankYFTang:
##### Checklist

- [X] Required: Issue filed: https://unicode-org.atlassian.net/browse/ICU-21374
- [X] Required: The PR title must be prefixed with a JIRA Issue number. 
- [X] Required: The PR description must include the link to the Jira Issue, for example by completing the URL in the first checklist item
- [X] Required: Each commit message must be prefixed with a JIRA Issue number. 
- [X] Issue accepted (done by Technical Committee after discussion)
- [x] Tests included, if applicable
- [ ] API docs and/or User Guide docs changed or added, if applicable

Add the github action bot to build with cfi
Also fix all the known issues which require to change the C style cast:
1. Change to C++ style static_cast<> for code inside i18n and common while we know the class of the object, OR
2. Change to C++ style dynamic_cast<> for code inside i18n and common while we are not sure about the class of the object, OR
3. Change to C++ style dynamic_cast<> for code inside test and tool code, 
Also, change to use const_cast for casting between const / non-const

PR #3606, opened 2025-08-22
ICU-22775 fix StringSearch pattern CE mem mgmt

markusicu:
- the main bug in Java initializePatternPCETable() was that the pcetablesize never got updated
- in C++, if the pattern CE loops encountered a failure, they didn't free the array (unlikely but possible near OOM)
- I think I made the memory management less convoluted
- for good measure, I used the "append" functions to add the null terminator, rather than rely on these functions having over-allocated by one for the real collation elements

#### Checklist
- [x] Required: Issue filed: ICU-22775

### DIFF
--- a/icu4c/source/i18n/usearch.cpp
+++ b/icu4c/source/i18n/usearch.cpp
@@ -177,8 +177,9 @@ inline void * allocateMemory(uint32_t size, UErrorCode *status)
 * before calling this method. destination not to be nullptr and has at least
 * size destinationlength.
 * @param destination target array
-* @param offset destination offset to add value
 * @param destinationlength target array size, return value for the new size
+* @param destOnHeap whether the destination array is heap-allocated
+* @param offset destination offset to add value
 * @param value to be added
 * @param increments incremental size expected
 * @param status output error if any, caller to check status before calling
@@ -187,21 +188,24 @@ inline void * allocateMemory(uint32_t size, UErrorCode *status)
 */
 static
 inline int32_t * addTouint32_tArray(int32_t    *destination,
-                                    uint32_t    offset,
                                     uint32_t   *destinationlength,
+                                    bool        destOnHeap,
+                                    uint32_t    offset,
                                     uint32_t    value,
                                     uint32_t    increments,
                                     UErrorCode *status)
 {
-    uint32_t newlength = *destinationlength;
-    if (offset + 1 == newlength) {
-        newlength += increments;
+    if (offset >= *destinationlength) {
+        uint32_t newlength = offset + increments;
         int32_t* temp = static_cast<int32_t*>(allocateMemory(
                                          sizeof(int32_t) * newlength, status));
         if (U_FAILURE(*status)) {
-            return nullptr;
+            return destination;
         }
         uprv_memcpy(temp, destination, sizeof(int32_t) * (size_t)offset);
+        if (destOnHeap) {
+            uprv_free(destination);
+        }
         *destinationlength = newlength;
         destination        = temp;
     }
@@ -217,8 +221,9 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
 * before calling this method. destination not to be nullptr and has at least
 * size destinationlength.
 * @param destination target array
-* @param offset destination offset to add value
 * @param destinationlength target array size, return value for the new size
+* @param destOnHeap whether the destination array is heap-allocated
+* @param offset destination offset to add value
 * @param value to be added
 * @param increments incremental size expected
 * @param status output error if any, caller to check status before calling
@@ -227,29 +232,28 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
 */
 static
 inline int64_t * addTouint64_tArray(int64_t    *destination,
-                                    uint32_t    offset,
                                     uint32_t   *destinationlength,
+                                    bool        destOnHeap,
+                                    uint32_t    offset,
                                     uint64_t    value,
                                     uint32_t    increments,
                                     UErrorCode *status)
 {
-    uint32_t newlength = *destinationlength;
-    if (offset + 1 == newlength) {
-        newlength += increments;
+    if (offset >= *destinationlength) {
+        uint32_t newlength = offset + increments;
         int64_t* temp = static_cast<int64_t*>(allocateMemory(
                                          sizeof(int64_t) * newlength, status));
-
         if (U_FAILURE(*status)) {
-            return nullptr;
+            return destination;
         }
-
         uprv_memcpy(temp, destination, sizeof(int64_t) * (size_t)offset);
+        if (destOnHeap) {
+            uprv_free(destination);
+        }
         *destinationlength = newlength;
         destination        = temp;
     }
-
     destination[offset] = value;
-
     return destination;
 }
 
@@ -299,22 +303,22 @@ inline void initializePatternCETable(UStringSearch *strsrch, UErrorCode *status)
            U_SUCCESS(*status)) {
         uint32_t newce = getCE(strsrch, ce);
         if (newce) {
-            int32_t *temp = addTouint32_tArray(cetable, offset, &cetablesize,
-                                  newce,
-                                  patternlength - ucol_getOffset(coleiter) + 1,
-                                  status);
-            if (U_FAILURE(*status)) {
-                return;
-            }
+            cetable = addTouint32_tArray(
+                cetable, &cetablesize, /* destOnHeap= */ cetable != pattern->cesBuffer,
+                offset, newce, patternlength - ucol_getOffset(coleiter) + 1, status);
             offset ++;
-            if (cetable != temp && cetable != pattern->cesBuffer) {
-                uprv_free(cetable);
-            }
-            cetable = temp;
         }
     }
 
-    cetable[offset]   = 0;
+    cetable = addTouint32_tArray(
+        cetable, &cetablesize, /* destOnHeap= */ cetable != pattern->cesBuffer,
+        offset, 0, 1, status);
+    if (U_FAILURE(*status)) {
+        if (cetable != pattern->cesBuffer) {
+            uprv_free(cetable);
+        }
+        return;
+    }
     pattern->ces       = cetable;
     pattern->cesLength = offset;
 }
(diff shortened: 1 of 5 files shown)


icu4c/source/i18n/usearch.cpp
```diff
@@ -187,21 +188,24 @@ inline void * allocateMemory(uint32_t size, UErrorCode *status)
...
+                                    uint32_t    offset,
                                     uint32_t    value,
                                     uint32_t    increments,
                                     UErrorCode *status)
 {
-    uint32_t newlength = *destinationlength;
-    if (offset + 1 == newlength) {
-        newlength += increments;
+    if (offset >= *destinationlength) {
+        uint32_t newlength = offset + increments;
```
FrankYFTang, line 181: The *status could be either success or failure when calling into addTouint32_tArray, the code does not check the status and call allocateMemory unconditionally. Then allocateMemory call uprv_malloc unconditionally. So this could cause a possible memory leak here in the if (U_FAILURE(*status)) { return }.  if the addTouint32_tArray() function is called with a failed status code in the input. This issue is a new by the change here but a preexisting issue but preiously the loop condition ensure status code is always success when calling into addTouint32_tArray. 
I think we should add 
if (U_FAILURE(*status)) { return destination; } 
in the beginning of this function.
or, if not desirable, add
U_ASSSERT(U_SUCCESS(*status));

icu4c/source/i18n/usearch.cpp
```diff
@@ -227,29 +232,28 @@ inline int32_t * addTouint32_tArray(int32_t    *destination,
...
+                                    bool        destOnHeap,
+                                    uint32_t    offset,
                                     uint64_t    value,
                                     uint32_t    increments,
                                     UErrorCode *status)
 {
-    uint32_t newlength = *destinationlength;
-    if (offset + 1 == newlength) {
-        newlength += increments;
+    if (offset >= *destinationlength) {
```
FrankYFTang: same as above

icu4c/source/i18n/usearch.cpp
```diff
@@ -299,22 +303,22 @@ inline void initializePatternCETable(UStringSearch *strsrch, UErrorCode *status)
...
-                uprv_free(cetable);
-            }
-            cetable = temp;
         }
     }
 
-    cetable[offset]   = 0;
+    cetable = addTouint32_tArray(
+        cetable, &cetablesize, /* destOnHeap= */ cetable != pattern->cesBuffer,
+        offset, 0, 1, status);
```
FrankYFTang, line 297: This is another possible point to cause memory leak, if status is already failure, it will still call addTouint32_tArray and addTouint32_tArray may still call allocateMemory and call allocateMemory

icu4c/source/i18n/usearch.cpp
```diff
@@ -368,25 +372,21 @@ inline void initializePatternPCETable(UStringSearch *strsrch,
...
-        }
-
+        pcetable = addTouint64_tArray(
+            pcetable, &pcetablesize, /* destOnHeap= */ pcetable != pattern->pcesBuffer,
+            offset, pce, patternlength - ucol_getOffset(coleiter) + 1, status);
         offset += 1;
+    }
 
-        if (pcetable != temp && pcetable != pattern->pcesBuffer) {
+    pcetable = addTouint64_tArray(
```
FrankYFTang, line 363: same as above

icu4c/source/i18n/usearch.cpp
FrankYFTang: please look at the implementation of allocateMemory
it call uprv_malloc() without checking status. I think that is the major problem.

markusicu: Good catch. You are right, I didn't guard my new calls with status checks.
I changed these functions, inlining the allocation and always checking first for a failure code, as usual, so that no one gets confused.

icu4c/source/test/intltest/srchtest.cpp
```diff
@@ -2452,6 +2453,24 @@ void StringSearchTest::TestCoverage(){
...
+        u"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        u"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa響");
+    LocalPointer<RuleBasedCollator> collator(
+        dynamic_cast<RuleBasedCollator *>(
+            Collator::createInstance(Locale::getUS(), errorCode)),
+        errorCode);
+    collator->setAttribute(UCOL_STRENGTH, UCOL_PRIMARY, errorCode);
+    StringCharacterIterator text(u" ");
+    StringSearch stringSearch(pattern, text, collator.getAlias(), nullptr, errorCode);
+    stringSearch.next(errorCode);
```
FrankYFTang, line 2473: I think we should verify the value of the errorCode in each step in the test

markusicu:
feedback addressed, CI failure should be fixed -- PTAL

@FrankYFTang @eggrobin why does [ControlFlowIntegrity](https://clang.llvm.org/docs/ControlFlowIntegrity.html#bad-cast-checking) require dynamic_cast even when the subclass is known? In non-CFI configs, a static_cast works just fine. (Not a blocker, we have had this since https://github.com/unicode-org/icu/pull/2284, I just still don't understand the complaint. See my third commit here.)

icu4c/source/test/intltest/srchtest.cpp
markusicu: Why? That would be very unusual.
We have function chaining with UErrorCode, and the IcuTestErrorCode checks when it goes out of scope.

FrankYFTang:
> feedback addressed, CI failure should be fixed -- PTAL
> 
> @FrankYFTang @eggrobin why does [ControlFlowIntegrity](https://clang.llvm.org/docs/ControlFlowIntegrity.html#bad-cast-checking) require dynamic_cast even when the subclass is known? In non-CFI configs, a static_cast works just fine. (Not a blocker, we have had this since #2284, I just still don't understand the complaint. See my third commit here.)

see my email subject "CFI Needs Runtime Type Checks"

icu4c/source/test/intltest/srchtest.cpp
FrankYFTang: The problem is then the test will not fail if internally some of the step reset a failed error code to U_ZERO_ERROR. Let's say collator->setAttribute() set the errorCode to a fail code but somehow StringSearch  constructor ignore that error but reset  errorCode to U_ZERO_ERROR, we will not capture that error, right?

markusicu: If the code is that broken, then all bets are off. I really don't want to start a pattern of checking the UErrorCode at intermediate points. That is an ICU anti-pattern.

richgillam: For what little it's worth, I'm with Markus on this one...

FrankYFTang: > If the code is that broken, then all bets are off. I really don't want to start a pattern of checking the UErrorCode at intermediate points. That is an ICU anti-pattern.

I am not asking you to checkint  the UErrorCode at intermediate points inside run time code, this is testing code. Testing code should not assume each step is doing the right thing and only verify the last step. Testing code should assume any given step has the possibility to be broken and two bugs may produce an illusion of correct result. 

And this is not a new pattern
For example, looking for 
status.assertSuccess(); inside
https://github.com/unicode-org/icu/blob/500123b80313e4dcbe8b7aca8a026291c95a38d4/icu4c/source/test/intltest/collationtest.cpp#L372
These were introduced by you 12 years ago in 
https://github.com/unicode-org/icu/commit/e977c057a9172e123d8e93836e234ad7cdf371a2
I am simply asking you to write the same way as how you wrote it 12 years ago in that test

markusicu: > Testing code should not assume each step is doing the right thing and only verify the last step. Testing code should assume any given step has the possibility to be broken and two bugs may produce an illusion of correct result.

It's still a chain of functions that each take a UErrorCode. We _sometimes_ check for errors at more points than necessary, if something is likely to fail and to then make debugging easier.

> And this is not a new pattern For example, looking for status.assertSuccess(); inside
> 
> https://github.com/unicode-org/icu/blob/500123b80313e4dcbe8b7aca8a026291c95a38d4/icu4c/source/test/intltest/collationtest.cpp#L372

That checks for data loading problems before calling a bunch of UnicodeSet functions that don't take a UErrorCode. And it doesn't check for "set.isBogus()" each time.

icu4c/source/test/intltest/srchtest.cpp
```diff
@@ -2452,6 +2453,24 @@ void StringSearchTest::TestCoverage(){
...
+    UnicodeString pattern(
+        u"Xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        u"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        u"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+        u"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa響");
+    LocalPointer<RuleBasedCollator> collator(
+        dynamic_cast<RuleBasedCollator *>(
+            Collator::createInstance(Locale::getUS(), errorCode)),
+        errorCode);
+    collator->setAttribute(UCOL_STRENGTH, UCOL_PRIMARY, errorCode);
```
FrankYFTang, line 2470: Collator::createInstance could return nullptr and set errorCode to error code, calling collator->setAttribute without checking errorCode is anti pattern in ICU. If some necessary resources of US collator are filtering out, Collator::createInstance(Locale::getUS(), errorCode) may fail

markusicu: done

echeran (approved):
LGTM

PR #3608, opened 2025-08-25
ICU-23192 Replace use of CharString with FixedString for data members

roubert:
When `CharString` is used as a data member that never is modified (apart from initialization), it can be replaced with `FixedString` in order to save a little memory (which, in case anyone creates a lot of any of these objects, could add up).

#### Checklist
- [x] Required: Issue filed: ICU-23192

ALLOW_MANY_COMMITS=true

### DIFF
--- a/icu4c/source/common/fixedstring.h
+++ b/icu4c/source/common/fixedstring.h
@@ -32,7 +32,7 @@ public:
 
     FixedString(std::string_view init) {
         size_t size = init.size();
-        if (reserve(size + 1)) {
+        if (size > 0 && reserve(size + 1)) {
             uprv_memcpy(ptr, init.data(), size);
             ptr[size] = '\0';
         }
--- a/icu4c/source/common/fixedstring.h
+++ b/icu4c/source/common/fixedstring.h
@@ -13,6 +13,8 @@
 
 U_NAMESPACE_BEGIN
 
+class UnicodeString;
+
 /**
  * ICU-internal fixed-length char* string class.
  * This is a complement to CharString to store fixed-length strings efficiently
@@ -94,6 +96,8 @@ public:
 private:
     char* ptr = nullptr;
 };
+
+U_COMMON_API void copyInvariantChars(const UnicodeString& src, FixedString& dst, UErrorCode& status);
 
 U_NAMESPACE_END
 
--- a/icu4c/source/common/fixedstring.h
+++ b/icu4c/source/common/fixedstring.h
@@ -5,6 +5,7 @@
 #define FIXEDSTRING_H
 
 #include <string_view>
+#include <utility>
 
 #include "unicode/uobject.h"
 #include "unicode/utypes.h"
@@ -28,7 +29,7 @@ public:
     FixedString() = default;
     ~FixedString() { operator delete[](ptr); }
 
-    FixedString(const FixedString&) = delete;
+    FixedString(const FixedString& other) : FixedString(other.data()) {}
 
     FixedString(std::string_view init) {
         size_t size = init.size();
@@ -57,7 +58,7 @@ public:
         return *this;
     }
 
-    FixedString(FixedString&&) noexcept = delete;
+    FixedString(FixedString&& other) noexcept : ptr(std::exchange(other.ptr, nullptr)) {}
 
     FixedString& operator=(FixedString&& other) noexcept {
         operator delete[](ptr);
--- a/icu4c/source/common/unicode/localebuilder.h
+++ b/icu4c/source/common/unicode/localebuilder.h
@@ -18,7 +18,7 @@
  */
 
 U_NAMESPACE_BEGIN
-class CharString;
+class FixedString;
 
 /**
  * <code>LocaleBuilder</code> is used to build instances of <code>Locale</code>
@@ -297,7 +297,7 @@ private:
     char language_[9];
     char script_[5];
     char region_[4];
-    CharString *variant_;  // Pointer not object so we need not #include internal charstr.h.
+    FixedString *variant_;  // Pointer not object so we need not #include internal fixedstring.h.
     icu::Locale *extensions_;  // Pointer not object. Storage for all other fields.
 
 };
--- a/icu4c/source/i18n/measunit_impl.h
+++ b/icu4c/source/i18n/measunit_impl.h
@@ -11,6 +11,7 @@
 #include "unicode/measunit.h"
 #include "cmemory.h"
 #include "charstr.h"
+#include "fixedstring.h"
 
 U_NAMESPACE_BEGIN
 
@@ -249,11 +250,14 @@ class U_I18N_API_CLASS MeasureUnitImpl : public UMemory {
     /**
      * Used for currency units.
      */
-    static inline MeasureUnitImpl forCurrencyCode(StringPiece currencyCode) {
+    static inline MeasureUnitImpl forCurrencyCode(StringPiece currencyCode, UErrorCode& status) {
         MeasureUnitImpl result;
-        UErrorCode localStatus = U_ZERO_ERROR;
-        result.identifier.append(currencyCode, localStatus);
-        // localStatus is not expected to fail since currencyCode should be 3 chars long
+        if (U_SUCCESS(status)) {
+            result.identifier = currencyCode;
+            if (result.identifier.isEmpty() != currencyCode.empty()) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+            }
+        }
         return result;
     }
 
@@ -317,7 +321,7 @@ class U_I18N_API_CLASS MeasureUnitImpl : public UMemory {
     /**
      * The full unit identifier.  Owned by the MeasureUnitImpl.  Empty if not computed.
      */
-    CharString identifier;
+    FixedString identifier;
 
     /**
      * Represents the unit constant denominator.
--- a/icu4c/source/i18n/unicode/fmtable.h
+++ b/icu4c/source/i18n/unicode/fmtable.h
@@ -34,7 +34,7 @@
 
 U_NAMESPACE_BEGIN
 
-class CharString;
+class FixedString;
 
 namespace number::impl {
 class DecimalQuantity;
@@ -665,12 +665,12 @@ public:
     void adoptDecimalQuantity(number::impl::DecimalQuantity *dq);
 
     /**
-     * Internal function to return the CharString pointer.
+     * Internal function to return the FixedString pointer.
      * @param status error code
-     * @return pointer to the CharString - may become invalid if the object is modified
+     * @return pointer to the FixedString - may become invalid if the object is modified
      * @internal
      */
-    CharString *internalGetCharString(UErrorCode &status);
+    FixedString *internalGetFixedString(UErrorCode &status);
 
 #endif  /* U_HIDE_INTERNAL_API */
 
@@ -700,7 +700,7 @@ private:
         }               fArrayAndCount;
     } fValue;
 
-    CharString           *fDecimalStr;
+    FixedString* fDecimalStr;
 
     number::impl::DecimalQuantity *fDecimalQuantity;
 
--- a/icu4c/source/i18n/units_converter.cpp
+++ b/icu4c/source/i18n/units_converter.cpp
@@ -7,6 +7,7 @@
 
 #include "charstr.h"
 #include "cmemory.h"
+#include "cstring.h"
 #include "double-conversion-string-to-double.h"
 #include "measunit_impl.h"
 #include "putilimp.h"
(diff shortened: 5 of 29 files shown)


icu4c/source/i18n/units_data.h
```diff
@@ -31,18 +32,21 @@ class ConversionRateInfo : public UMemory {
...
+            status = U_MEMORY_ALLOCATION_ERROR;
+        }
     }
-    CharString sourceUnit;
-    CharString baseUnit;
-    CharString factor;
-    CharString offset;
-    CharString specialMappingName; // the name of a special mapping used instead of factor + optional offset.
-    CharString systems;
+    FixedString sourceUnit;
```
markusicu, line 44: good: fewer bytes used explicitly
bad: now we always heap-allocate each string

each memory allocation has some overhead in the allocator. when we take that into account, i am not sure that we save a lot when strings used to be internal to CharString.

i am not sure how to measure this, but it seems like if we could make FixedString a little bigger, say 16 bytes, and store very short strings internally, we would probably come out ahead.

there is also a performance issue. memory allocation takes some time. if the object that has a FixedString is long-lived, it's fine, but if we create & destroy these things a lot, like many server-side users do with formatters, this sort of change could hurt.

icu4c/source/i18n/units_data.cpp
```diff
@@ -109,15 +109,56 @@ class ConversionRateDataSink : public ResourceSink {
...
+                    status = U_MEMORY_ALLOCATION_ERROR;
+                    return;
+                }
+                {
+                    int32_t length = baseUnit.length();
+                    if (!cr->baseUnit.reserve(length + 1)) {
+                        status = U_MEMORY_ALLOCATION_ERROR;
+                        return;
+                    }
+                    baseUnit.extract(0, length, cr->baseUnit.getAlias(), length + 1, US_INV);
```
markusicu: It looks like we would benefit from adding a function like
```
    copyFrom(std::u16string_view s16, UErrorCode &errorCode)
```

icu4c/source/i18n/units_data.cpp
```diff
@@ -470,9 +522,10 @@ MaybeStackVector<UnitPreference> UnitPreferences::getPreferencesFor(StringPiece
             for (int32_t j = 0; unitsMatchSystem && j < measureUnit.singleUnits.length(); j++) {
                 const SingleUnitImpl* singleUnit = measureUnit.singleUnits[j];
                 const ConversionRateInfo* rateInfo = rates.extractConversionInfo(singleUnit->getSimpleUnitID(), status);
-                CharString systems(rateInfo->systems, status);
-                if (!systems.contains("metric_adjacent")) { // "metric-adjacent" is considered to match all the locale systems
-                    if (!systems.contains(localeSystem.data())) {
+                const char* systems = rateInfo->systems.data();
+                // "metric-adjacent" is considered to match all the locale systems
+                if (uprv_strstr(systems, "metric_adjacent") == nullptr) {
+                    if (uprv_strstr(systems, localeSystem.data()) == nullptr) {
```
markusicu, line 490: not having contains() makes this kind of code harder to read

icu4c/source/i18n/units_data.cpp
```diff
@@ -388,7 +434,9 @@ void U_I18N_API getAllConversionRates(MaybeStackVector<ConversionRateInfo> &resu
 const ConversionRateInfo *ConversionRates::extractConversionInfo(StringPiece source,
                                                                  UErrorCode &status) const {
     for (size_t i = 0, n = conversionInfo_.length(); i < n; ++i) {
-        if (conversionInfo_[i]->sourceUnit == source) return conversionInfo_[i];
+        if (uprv_strncmp(conversionInfo_[i]->sourceUnit.data(), source.data(), source.size()) == 0) {
+            return conversionInfo_[i];
+        }
```
markusicu, line 401: not having operator==(string_view) makes this kind of code harder to read

icu4c/source/i18n/units_converter.cpp
```diff
@@ -212,8 +213,8 @@ Factor loadSingleFactor(StringPiece source, const ConversionRates &ratesInfo, UE
         return {};
     }
 
-    Factor result = extractFactorConversions(conversionUnit->factor.toStringPiece(), status);
-    result.offset = strHasDivideSignToDouble(conversionUnit->offset.toStringPiece(), status);
+    Factor result = extractFactorConversions(conversionUnit->factor.data(), status);
+    result.offset = strHasDivideSignToDouble(conversionUnit->offset.data(), status);
```
markusicu, line 217: it looks like we are going back to passing strings with known lengths as NUL-terminated pointers and calling strlen() again to make the implicit string_view...

icu4c/source/i18n/unicode/fmtable.h
```diff
@@ -665,12 +665,12 @@ class U_I18N_API Formattable : public UObject {
...
     /**
-     * Internal function to return the CharString pointer.
+     * Internal function to return the FixedString pointer.
      * @param status error code
-     * @return pointer to the CharString - may become invalid if the object is modified
+     * @return pointer to the FixedString - may become invalid if the object is modified
      * @internal
      */
-    CharString *internalGetCharString(UErrorCode &status);
+    FixedString *internalGetCharString(UErrorCode &status);
```
markusicu: this function is now misnamed

icu4c/source/i18n/measunit_impl.h
```diff
@@ -251,9 +252,7 @@ class U_I18N_API_CLASS MeasureUnitImpl : public UMemory {
      */
     static inline MeasureUnitImpl forCurrencyCode(StringPiece currencyCode) {
         MeasureUnitImpl result;
-        UErrorCode localStatus = U_ZERO_ERROR;
-        result.identifier.append(currencyCode, localStatus);
-        // localStatus is not expected to fail since currencyCode should be 3 chars long
+        result.identifier = currencyCode;
```
markusicu: the old code couldn't fail with a valid currencyCode because the identifier would always store it internally; with a FixedString that always heap-allocates, we get a new OOM failure mode

icu4c/source/common/unicode/localebuilder.h
```diff
@@ -297,7 +297,7 @@ class U_COMMON_API LocaleBuilder : public UObject {
     char language_[9];
     char script_[5];
     char region_[4];
-    CharString *variant_;  // Pointer not object so we need not #include internal charstr.h.
+    FixedString *variant_;  // Pointer not object so we need not #include internal charstr.h.
```
markusicu: comment out of sync

icu4c/source/i18n/units_data.h
markusicu: After looking over the whole PR, I really think we should make FixedString a little bigger so that
- we can store short strings internally --> reduce heap allocations, and in some cases eliminate them for all valid inputs (e.g., currencyCode)
- we can store the string length --> avoid adding strlen() calls, bring back toStringPiece()

We don't need to store a buffer capacity because a FixedString doesn't grow.

Something like a union of
- internal
  - 1 byte as a discriminator, and for the length
  - 15 bytes contents and NUL
- heap
  - 1 byte as discriminator (same bits as in the other variant)
  - 3 bytes padding
  - 4 bytes length
  - 8 bytes pointer (if 64-bit CPU)

I think this is more or less what a modern std::string does. It would be nice if we could just use that...

icu4c/source/i18n/units_data.cpp
roubert: Done.

icu4c/source/common/unicode/localebuilder.h
roubert: Done.

icu4c/source/i18n/unicode/fmtable.h
roubert: Done.

icu4c/source/i18n/units_converter.cpp
roubert: Kind of, but these functions take `StringPiece` input parameters which here are initialized from a data type that guarantees `NUL` termination so the old risk of passing strings of indeterminate length remains gone.

icu4c/source/i18n/measunit_impl.h
roubert: There was always a new object being created here so it was always possible to run out of memory, the only thing that changes now is where that happens.

icu4c/source/i18n/units_data.cpp
roubert: Considering how many calls to `uprv_strncmp()` this code base already contains, I don't think that adding a few more will make any noticeable difference.

icu4c/source/i18n/units_data.cpp
roubert: Considering how many calls to `uprv_strstr()` this code base already contains, I don't think that adding a few more will make any noticeable difference.

icu4c/source/i18n/units_data.h
roubert: You're right that we don't have any good way to measure the performance of this, but as a kind of smoke test, to check that these changes don't inadvertently make performance worse, I ran `intltest` through `/usr/bin/time` and averaged over 100 runs this PR in its current state reduced _Elapsed (wall clock) time_ by 0.36% and reduced _Maximum resident set size_ by 0.0024% and while such tiny reductions aren't any argument for increased performance (that's only to be expected for users that allocate a lot of these objects), they're a pretty good sign that these changes don't make performance worse.

Then I think there's some real value in `FixedString` being a trivially simple class, to replace code that previously used `uprv_malloc()` / `uprv_free()` by just adding guarantees for deallocation and `NUL` termination without adding any overhead at all.

Doing `strlen()` on a short string is a very cheap operation while storing the length of an arbitrarily long string requires an additional 8 bytes for a `size_t`, so it'd require some serious benchmarking to find a good balance between the two, as long as the string is short and `NUL` termination is guaranteed, `strlen()` can easily be the most performant option.

That said, it certainly seems likely that it would be possible to create a data structure that balances the size wasted for short or empty strings against the size wasted for unnecessary heap allocations, but getting that balance right would require first developing some benchmarks that resonably accurately reflect real-world usage and that'd be quite a lot more work that what was proposed for this ticket.

icu4c/source/i18n/measunit_impl.h
markusicu: No, the result and result.identifier were stack-allocated, and for valid currency codes the result.identifier stack buffer was plenty large enough. With the new FixedString always heap-allocating, this is adding a new OOM failure mode.

icu4c/source/i18n/units_data.h
markusicu: i am not going to block this pr on this now, but we should come back and do the simple and obvious change in FixedString later.

we know that memory allocations are more expensive in a multi-threaded environment, such as on a server. there is a real cost, even if we can't measure it well with running intltest.

icu4c/source/common/fixedstring.h
```diff
@@ -94,6 +96,8 @@ class FixedString : public UMemory {
     char* ptr = nullptr;
 };
 
+void copyInvariantChars(const UnicodeString& src, FixedString& dst, UErrorCode& status);
```
markusicu: optional/later: I would have taken a u16string_view and copied the code from UnicodeString.extract(), to reduce dependencies.

icu4c/source/i18n/units_data.cpp
markusicu: i am not going to fight this now, but making code gradually better and more readable seems like a good thing. you are surprising me by pushing back...

icu4c/source/i18n/units_data.cpp
markusicu: i am not going to fight this now, but making code gradually better and more readable seems like a good thing. you are surprising me by pushing back...

icu4c/source/i18n/measunit_impl.h
roubert: Also stack memory is limited and can be exhausted, so when this function here allocated a new `MeasureUnitImpl` on the stack it was always possible for that to fail if the system was running low on memory.

markusicu: Sure, but that failure would have been outside of the constructor, and the constructor would not have been called. The problem is that the constructor does not have a UErrorCode param, so when it fails, it can't report it.

roubert: Oh, this function has one single call site, `MeasureUnit::initCurrency()`, which already does heap allocation and handles heap allocation errors, see the change done there in this PR to also verify that the allocation of `identifier` succeded.

markusicu: That is totally obscure. Please either add a comment to that effect here, or, better, change this function to take a UErrorCode.

roubert: Done.

PR #3610, opened 2025-08-27
ICU-23167 Change Chinese and Dangi calendar epoch to 1AD

FrankYFTang:
Remove unnecessary structure to support other epoch in Chinese and Dangi calendar. Rename and keep CYCLE_EPOCH for the calculation of 60 years cycle.
This should only change the meaning of values for set/get extended year and woy year.
Also need to address some int32_t overflow issue after the value of gyear shift

The extended year value of "chinese" calendar reduced 2637 (newValue = oldValue -2637), and the value of "dangi" calendar reduced 2333 (newValue = oldValue -2333). 

#### Checklist
- [X] Required: Issue filed: ICU-23167


FrankYFTang:
Notice during the change, I also discover an unrelated minor issue ICU-23198. That issue is not addressed in this PR but will handle separately

FrankYFTang:
@sffc ping

icu4c/source/i18n/chnsecal.h
```diff
@@ -253,7 +247,6 @@ class U_I18N_API ChineseCalendar : public Calendar {
   virtual int32_t getActualMaximum(UCalendarDateFields field, UErrorCode& status) const override;
 
   struct Setting {
-      int32_t epochYear;
```
sffc, line 256: Praise: Clever to use the mechanism already in place to make Chinese and Dangi have different extended years in order to change the extended year for both calendars.

icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
```diff
@@ -829,9 +828,9 @@ protected void handleComputeFields(int julianDay) {
         MonthInfo info = computeMonthInfo(days, gyear);
 
         // Extended year and cycle year is based on the epoch year
-        int extended_year = gyear - epochYear;
-        int cycle_year = gyear - CHINESE_EPOCH_YEAR;
-        if (info.month < 10 ||
+        int extended_year = gyear - CHINESE_EPOCH_YEAR;
+        int cycle_year = gyear - CYCLE_EPOCH;
+        if (info.month < 10 ||  // TODO(ICU-23198) < 10 or < 11 ????
```
sffc, line 833: I think 10 is correct; it just needs to be a month that is always after July and before December. So I think anything between 8 and 11 should work. The condition should return `false` if the Chinese date is before the Gregorian new year (most of the year) and `true` if it is after the Gregorian new year (the last 3-8 weeks).

sffc (approved):
Nice work!

What is `calendar.res` and why did it change in this PR?

icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
FrankYFTang: read https://unicode-org.atlassian.net/browse/ICU-23198

PR #3611, opened 2025-08-27
ICU-23177 Add width-aware day period support and fix data loading

sffc:
#### Checklist
- [x] Required: Issue filed: ICU-23177


sffc:
API proposal has been sent. This is the C++ version. I will try to add the Java version tomorrow.

icu4c/source/i18n/unicode/dtfmtsym.h
```diff
@@ -769,6 +789,13 @@ class U_I18N_API_CLASS DateFormatSymbols final : public UObject  {
     UnicodeString*  fAmPms;
     int32_t         fAmPmsCount;
 
+    /**
+     * Wide Ampm strings. For example: "ante meridiem" and "post meridiem".
+     * These strings are uncommon but exist in a handful of locales.
+     */
+    UnicodeString*  fWideAmPms;
+    int32_t         fWideAmPmsCount;
```
sffc, line 804: This does increase the stack size of DateFormatSymbols, but it is already so big that the delta is only about 2-3%.

pedberg-icu:
> API proposal has been sent. This is the C++ version. I will try to add the Java version tomorrow.

@sffc But this is not yet in the ICU 78 API Proposals document

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -2396,23 +2450,23 @@ DateFormatSymbols::initializeData(const Locale& locale, const char *type, UError
         assignArray(fStandaloneNarrowMonths, fStandaloneNarrowMonthsCount, fShortMonths, fShortMonthsCount);
     }
 
-    // Load AM/PM markers; if wide or narrow not available, use short
+    // Load AM/PM markers.
+    // Narrow falls back to Abbr which falls back to Wide.
```
pedberg-icu: That is not the way the dayPeriod fallback works in CLDR. There, within a given style (format or stand-alone), wide and narrow both fall back to abbreviated; and stand-alone abbreviated falls back to format abbreviated. The root locale only has the format abbreviated forms.

sffc:
Windows failure:

```
         TestAmPmLengths23114 {
         FAIL: DateFormatSymbols wide; got "\u0E01\u0E48\u0E2D\u0E19\u0E40\u0E17\u0E35\u0E48\u0E22\u0E07"; expected "\u00E0\u00B8\u0081\u00E0\u00B9\u02C6\u00E0\u00B8\u00AD\u00E0\u00B8\u2122\u00E0\u00B9\u20AC\u00E0\u00B8\u2014\u00E0\u00B8\u00B5\u00E0\u00B9\u02C6\u00E0\u00B8\u00A2\u00E0\u00B8\u2021"
      
         } ERRORS (1) in TestAmPmLengths23114
```

There is also an ICU4J failure that seems unrelated:

```
Error:  Tests run: 26, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.958 s <<< FAILURE! -- in com.ibm.icu.dev.test.collator.AlphabeticIndexTest
```

pedberg-icu:
I am traveling but should have a chance to take a closer look at this sometime in the next 3 days or so.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -2020,7 +2070,8 @@ protected void initializeData(ULocale desiredLocale, ICUResourceBundle b, String
         standaloneNarrowWeekdays[0] = "";  // 1-based
         System.arraycopy(snWeekdays, 0, standaloneNarrowWeekdays, 1, snWeekdays.length);
 
-        ampms = arrays.get("AmPmMarkers");
+        ampms = arrays.get("AmPmMarkersAbbr");
+        ampmsWide = arrays.get("AmPmMarkers");
```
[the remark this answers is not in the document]
pedberg-icu: Actually I think both the C and J implementations should load abbreviated first, then wide and narrow with each of the latter falling back to abbreviated if they are missing. That is the way the inheritance is specified in CLDR, and root _only_ provides the abbreviated forms.

pedberg-icu:
@sffc This all looks good to me, with my one comment being that we should change the loading order & fallback for both C and J so that we load abbreviated first, and then have both wide and narrow fall back to abbreviated if they are missing. That way C and J would be consistent with each other and with the CLDR fallback order (in which wide/narrow fallback to abbreviated, and root only provides abbreviated). I will go ahead and approve in case you need to get this in and make those changes later, but I would like those changes at some point.

pedberg-icu (approved):
This all looks good to me, with my one comment being that we should change the loading order & fallback for both C and J so that we load abbreviated first, and then have both wide and narrow fall back to abbreviated if they are missing. That way C and J would be consistent with each other and with the CLDR fallback order (in which wide/narrow fallback to abbreviated, and root only provides abbreviated). I will go ahead and approve in case you need to get this in and make those changes later, but I would like those changes at some point.

macchiati:
Hmmm. I wonder if there is a way with conformance tests to verify the order and targets of aliases. 
Just a thought: Perhaps we can have one or more artificial locales in CLDR with constant data having "holes" in it to force fallback. It wouldn't be accessible in production releases, but could be used in conformance tests to verify that the order and targets of inheritance matches order and targets of CLDR aliases. @sffc @echeran

sffc:
I don't believe that the manual fallback code as written in C++ actually does anything. The data loading sink should see the alias in root and do the fallback itself. I consider the C++ code to be merely a robust coding style, consistent with the rest of the symbols loading. In Java, most of the symbols also do not have explicit fallbacks written in code.

sffc:
> Just a thought: Perhaps we can have one or more artificial locales in CLDR with constant data having "holes" in it to force fallback. It wouldn't be accessible in production releases, but could be used in conformance tests to verify that the order and targets of inheritance matches order and targets of CLDR aliases. @sffc @echeran

Having a test locale in CLDR to exercise these sorts of things would be valuable. We could leverage en-XA perhaps.

icu4c/source/i18n/unicode/dtfmtsym.h
```diff
@@ -436,20 +436,27 @@ class U_I18N_API_CLASS DateFormatSymbols final : public UObject  {
     /**
      * Gets AM/PM strings with the specified width. For example: "A" and "P".
      * @param count    Filled in with length of the array.
+     * @param context   The usage context: FORMAT, STANDALONE. Currently ignored.
```
pedberg-icu: Should note that the values actually returned are the format styles.

icu4c/source/i18n/unicode/dtfmtsym.h
```diff
@@ -436,20 +436,27 @@ class U_I18N_API_CLASS DateFormatSymbols final : public UObject  {
...
-    U_I18N_API const UnicodeString* getAmPmStrings(int32_t& count, DtWidthType width) const;
+    U_I18N_API const UnicodeString* getAmPmStrings(int32_t& count,
+                                                   DtContextType context,
+                                                   DtWidthType width) const;
 
     /**
      * Sets AM/PM strings with the specified width. For example: "A" and "P".
      * @param ampms   The new AM/PM strings. (not adopted; caller retains ownership)
      * @param count   The number of strings in the array.
+     * @param context The usage context: FORMAT, STANDALONE. Currently ignored.
```
pedberg-icu: Note that the values returned are the format styles.

PR #3612, opened 2025-08-27
ICU-23179 Test more edge cases when mapping syntax characters to sets

eggrobin:
As for `^`, `-` is treated normally at the beginning of a UnicodeSet and as a set elsewhere.

Again, this is madness and we should probably get rid of this possibility, but for now let us make sure we know exactly which way madness lies.

#### Checklist
- [x] Required: Issue filed: ICU-23179

### DIFF
--- a/icu4c/source/test/intltest/usettest.cpp
+++ b/icu4c/source/test/intltest/usettest.cpp
@@ -2044,6 +2044,9 @@ void UnicodeSetTest::TestLookupSymbolTable() {
     symbols.add(u'{', UnicodeSet(u"[{leftCurlyBracket}]", errorCode));
     symbols.add(u'}', UnicodeSet(u"[{rightCurlyBracket}]", errorCode));
     symbols.add(u'$', UnicodeSet(u"[{dollarSign}]", errorCode));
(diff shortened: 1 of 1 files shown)


eggrobin:
@markusicu, friendly ping (#3604, the UnicodeSet parser rewrite, is waiting behind this one).

macchiati:
> As for ^, - is treated normally at the beginning of a UnicodeSet and as a set elsewhere.

I didn't understand "as a set elsewhere". Did you mean is treated as syntax to form a range elsewhere? Eg [a-b]?
BTW I think we only had the "is normal at the start" to match regexes.

eggrobin:
> I didn't understand "as a set elsewhere". Did you mean is treated as syntax to form a range elsewhere? Eg [a-b]?

The context here is that the symbol table can map characters to sets, e.g., it can say “`0` means `[a-z]`”. And this can be done to syntax characters, e.g., “`-` means `[1-9]`”. The effect of mapping syntax characters varies, and this is what this pull request tests. Obviously this is madness and we should disallow the mapping of syntax characters in a future version of ICU, but for now I am testing exactly what madness happens so I can preserve it in my rewrite of the parser.

macchiati:
Got it, thanks!

PR #3615, opened 2025-08-28
ICU-23189 Test on modern GCC and modern Clang for modern C++

eggrobin:
Note that the default GCC build is using GCC 11 already, as it is the default on ubuntu 22.04, so we don’t need a separate GCC 11 build. Bump that to 14, which is the most recent we can get on the latest GitHub runner (24.04), and make it build for C++20 and C++23.

Also make clang build for C++23. Try both Clang 18 and 19, as their libc++s have different sets of features (before this change, utfiterator did not compile with Clang 18 in C++23 mode, but compiled with Clang 19).

Fix UB in utfiteratortest.cpp which resulted in a segfault with GCC.

#### Checklist
- [x] Required: Issue filed: ICU-23189

ALLOW_MANY_COMMITS=meow

### DIFF
--- a/icu4c/source/common/unicode/unistr.h
+++ b/icu4c/source/common/unicode/unistr.h
@@ -1924,30 +1924,39 @@ public:
   inline UBool isBogus() const;
 
 #ifndef U_HIDE_DRAFT_API
+private:
+  // These type aliases are private; there is no guarantee that they will remain
+  // aliases to the same types in subsequent versions of ICU.
+  // Note that whether `std::u16string_view::const_iterator` is a pointer or a
+  // class that models contiguous_iterator is platform-dependent.
+  using unspecified_iterator = std::u16string_view::const_iterator;
+  using unspecified_reverse_iterator = std::u16string_view::const_reverse_iterator;
+
+public:
   /**
    * @return an iterator to the first code unit in this string.
    *     The iterator may be a pointer or a contiguous-iterator object.
    * @draft ICU 78
    */
-  auto begin() const { return std::u16string_view(*this).begin(); }
+  unspecified_iterator begin() const { return std::u16string_view(*this).begin(); }
   /**
    * @return an iterator to just past the last code unit in this string.
    *     The iterator may be a pointer or a contiguous-iterator object.
    * @draft ICU 78
    */
-  auto end() const { return std::u16string_view(*this).end(); }
+  unspecified_iterator end() const { return std::u16string_view(*this).end(); }
   /**
    * @return a reverse iterator to the last code unit in this string.
    *     The iterator may be a pointer or a contiguous-iterator object.
    * @draft ICU 78
    */
-  auto rbegin() const { return std::u16string_view(*this).rbegin(); }
+  unspecified_reverse_iterator rbegin() const { return std::u16string_view(*this).rbegin(); }
   /**
    * @return a reverse iterator to just before the first code unit in this string.
    *     The iterator may be a pointer or a contiguous-iterator object.
    * @draft ICU 78
    */
-  auto rend() const { return std::u16string_view(*this).rend(); }
+  unspecified_reverse_iterator rend() const { return std::u16string_view(*this).rend(); }
 #endif  // U_HIDE_DRAFT_API
 
   //========================================
--- a/.github/workflows/icu4c.yml
+++ b/.github/workflows/icu4c.yml
@@ -86,21 +86,28 @@ jobs:
           make install;
           PATH=$PREFIX/bin:$PATH make -C test/hdrtst check
 
-  #gcc 11 with c++ 20
-  gcc11-cpp20:
-    runs-on: ubuntu-22.04  # Updated in BRS
+  #gcc 14 with c++ 20 and later
+  gcc14-cpp20-23:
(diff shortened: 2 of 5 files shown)


eggrobin:
Well that is exciting:

```
         testUncommonForwardRange {
make[2]: *** [Makefile:120: check-local] Segmentation fault (core dumped)
```

icu4c/source/common/unicode/unistr.h
```diff
@@ -1924,30 +1924,39 @@ class U_COMMON_API UnicodeString : public Replaceable
   inline UBool isBogus() const;
 
 #ifndef U_HIDE_DRAFT_API
+private:
+  // These type aliases are private; there is no guarantee that they will remain
+  // aliases to the same types in subsequent versions of ICU.
+  // Note that whether `std::u16string_view::const_iterator` is a pointer or a
+  // class that models contiguous_iterator is platform-dependent.
+  using iterator = std::u16string_view::const_iterator;
```
eggrobin: @markusicu This type alias is very much not part of the API, but let me know if I should give it a scarier name (internal_iterator? unspecified_iterator? unstable_iterator? you_should_feel_bad_if_you_depend_on_the_type_of_this_iterator?)

roubert (commented):
I see two things here that need to be anchored in the TC before proceeding:

1. This adds 5 new jobs to the CI:

> gcc14-cpp20-23 (-std=c++23)
> succeeded 3 days ago in 9m 44s
> 
> clang-18-19-cpp-20-23-warning-as-errors (18, -std=c++23 -stdlib=libc++)
> succeeded 3 days ago in 11m 8s
> 
> clang-18-19-cpp-20-23-warning-as-errors (19, -std=c++20)
> succeeded 3 days ago in 10m 47s
> 
> clang-18-19-cpp-20-23-warning-as-errors (19, -std=c++20 -stdlib=libc++)
> succeeded 3 days ago in 12m 30s
> 
> clang-18-19-cpp-20-23-warning-as-errors (19, -std=c++23 -stdlib=libc++)
> succeeded 3 days ago in 10m 17s

If it could be done for free, it would of course be better to test every possible configuration, but each and every one of these jobs consume a noticable chunk of our limited CI resources so we need to make some concious decisions about what configurations are worth spending resources on. At some point, additional jobs consuming resources end up making the CI detrimentally slower and/or more flaky.

2. This makes supporting C++23 a de-facto promise. That's something that we'll eventually have to do some day, but is it really a promise that we want to start making already today?

Maybe the answer to both these questions is yes, this is worth it and something we want to start promising today, but it's something that needs to be discussed and agreed upon in the TC.

That said, I also wonder why "modern" Clang here is version 18 / 19 and not 20 / 21 and why GCC 14 instead of 15?

eggrobin:
> This makes supporting C++23 a de-facto promise. That's something that we'll eventually have to do some day, but is it really a promise that we want to start making already today?

Bear in mind that we wrote APIs targeting C++23 as part of #3499; so as a TC we are certainly aware that C++23 exists.
I would be OK with making these post-merge and allowing them to fail; but what I do not want is for us to unknowingly break C++23.

> That said, I also wonder why "modern" Clang here is version 18 / 19 and not 20 / 21

There is value in testing with 18 (since it did not work before this change). It would probably be a good idea to test with 20 or 21, perhaps instead of 19.

> and why GCC 14 instead of 15?

I don’t know how to easily get my hands on 15 (the latest GCC I can get from the latest runner GitHub gives me is 14).

icu4c/source/common/unicode/unistr.h
markusicu: unspecified_iterator sgtm

markusicu:
Whether we promise to support C++23 or not, our users expect that we do, or get into trouble as their projects move to C++23. They are starting to file bug reports, just like they have with earlier versions of C++, and we always make an effort to make stuff work. Getting ahead of it via CI seems good. If it hasn't been explicitly discussed in the TC, then an email to the TC mailing list should be fine to make it explicit.

Can we reduce the additional CI cost? Do we need so many combinations? Moving some to post-merge might be ok.

I suspect that we can't just switch existing CI checks from older versions to 23 because these C++ versions are not really backward compatible.

markusicu (commented):
good stuff

eggrobin:
> Can we reduce the additional CI cost? Do we need so many combinations?

I dropped clang 18 in C++20 mode, with a comment (it seems unlikely that this would have different errors than more modern clang in C++20 mode).

It looks like we have actionable warnings with clang 21, so I will try downgrading that until the warning disappears, and make a separate PR to bump it back up and fix the warnings.

markusicu:
> I dropped clang 18 in C++20 mode, with a comment

sgtm but i don't see that in your latest commit

eggrobin:
> sgtm but i don't see that in your latest commit

Ah yes, I forgot the `git commit -am ` part of the command…

eggrobin:
> why "modern" Clang here is version 18 / 19 and not 20 / 21

Turns out there is a better reason: we don’t build with 20 or 21 with warnings as error (UB flagged by [-Wnontrivial-memcall](https://clang.llvm.org/docs/DiagnosticsReference.html#wnontrivial-memcall), we should probably do something about that).

PR #3617, opened 2025-08-28
ICU-23199 Using 'C' in skeleton throws for some locales

mihnita:
#### Checklist
- [x] Required: Issue filed: ICU-23199

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
@@ -423,7 +423,11 @@ public class DateTimePatternGenerator implements Freezable<DateTimePatternGenera
 
         if (list != null) {
             defaultHourFormatChar = defaultCharFromLocale != null ? defaultCharFromLocale : list[0].charAt(0);
-            allowedHourFormats = Arrays.copyOfRange(list, 1, list.length - 1);
+            if (list.length > 1) {
+                allowedHourFormats = Arrays.copyOfRange(list, 1, list.length);
+            } else {
+                allowedHourFormats = LAST_RESORT_ALLOWED_HOUR_FORMAT;
+            }
         } else {
             allowedHourFormats = LAST_RESORT_ALLOWED_HOUR_FORMAT;
             defaultHourFormatChar = (defaultCharFromLocale != null) ? defaultCharFromLocale : allowedHourFormats[0].charAt(0);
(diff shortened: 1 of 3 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/text/DateTimePatternGenerator.java
```diff
@@ -423,7 +423,11 @@ private void getAllowedHourFormats(ULocale uLocale) {
 
         if (list != null) {
             defaultHourFormatChar = defaultCharFromLocale != null ? defaultCharFromLocale : list[0].charAt(0);
-            allowedHourFormats = Arrays.copyOfRange(list, 1, list.length - 1);
+            if (list.length > 1) {
```
mihnita, line 426: The `copyOfRange` was "off-by-one", too short.
> `to` - the final index of the range to be copied, exclusive. (This index may lie outside the array.)
> https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#copyOfRange-T:A-int-int-

When we have something like this in `supplementalData.xml`
```xml
<hours preferred="H" allowed="hB hb H h" regions="KE MM TZ UG"/>
```
the `preferred` & `allowed` keys are merged in the `list` (in `DayPeriodAllowedHoursSink`)

Here the code tries to do the opposite.
The first entry (`list[0]`) goes into `defaultHourFormatChar`.
The rest of the list, from index 1, goes into `allowedHourFormats`

So in the example:
- the `list` is `[ "H", "hB" "hb" "H" "h" ]`
- `allowedHourFormats` result should be `[ "hB" "hb" "H" "h" ]`
- but it is instead  `[ "hB" "hb" "H" ]`

One too short, because in `copyOfRange` `to` ... is exclusive.

---

The crash was caused by
```xml
<hours preferred="H" allowed="H" regions="AX BQ CP CZ DK FI ID IS ML NE RU SE SJ SK"/>
```

The `list` was `[ "H", "H" ]`, then `allowedHourFormats` was `[]` (empty array).

And later one this failed:
```java
} else { // patChr == 'C'
      String bestAllowed = allowedHourFormats[0];
```

---

The `if (list.length > 1)` is just for extra safety.
There is no entry in `<timeData>` that will get us there.

But it might happen if at some point we have something like this:
`<hours preferred="H" regions="AA BB"/>` or
`<hours preferred="H" allowed="" regions="AA BB"/>`

mihnita:
Thank you!
M

PR #3626, opened 2025-09-02
ICU-20392 Replace std::variant<> with a custom union to save 8 bytes

roubert:
There are no size guarantees for `std::variant<>` (the implementation in LLVM's _libc++_ will currently add at least 4 bytes for the discriminator, for example) so replacing it with a customized data structure instead gives the control required to make Locale objects 8 bytes smaller.

The storage required for different kinds of `Locale` objects then will look like this, each of them now 8 bytes smaller than they currently are:

|| legacy | current | new | _ratio_ |
| :--- | ---: | ---: | ---: | ---: |
| simple | 224 | 48 | 40 | 18% |
| complex | 227 | 117 | 109 | 48% |
| pathological | 245 | 237 | 229 | 93% |
| posix | 224 | 48 | 40 | 18% |
| fubar | 408 | 280 | 272 | 67% |

|||
| --- | --- |
| simple | `hi` |
| complex | `hi@t=en-h0-hybrid` |
| pathological | `sco__FONIPA_SCOTLAND@calendar=buddhist;collation=phonebook;t=tlh-fonxsamp-simple-h0-hybrid-m0-ungegn-1977;x=hello-world` |
| posix | `en_US_POSIX` |
| fubar | `vo__FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR_FUBAR` |

#### Checklist

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -35,11 +35,11 @@
 #include <string_view>
 #include <type_traits>
 #include <utility>
-#include <variant>
 
 #include "unicode/bytestream.h"
 #include "unicode/locid.h"
 #include "unicode/localebuilder.h"
+#include "unicode/localpointer.h"
 #include "unicode/strenum.h"
 #include "unicode/stringpiece.h"
 #include "unicode/uloc.h"
@@ -326,11 +326,111 @@ struct Locale::Heap : public UMemory {
         }
     }
 
-    // Move should be done on the std::unique_ptr object that owns this.
+    // Move should be done on the owner of the pointer to this object.
     Heap(Heap&&) noexcept = delete;
 
     ~Heap() = default;
 };
+
+Locale::Payload::HeapPtr& Locale::Payload::HeapPtr::operator=(const Heap& other) {
+    U_ASSERT(type == eBOGUS);
+    UErrorCode status = U_ZERO_ERROR;
+    ptr = new Heap(other, status);
+    if (ptr == nullptr || U_FAILURE(status)) {
+        delete ptr;
+    } else {
+        type = eHEAP;
+    }
+    return *this;
+}
+
+Locale::Payload::HeapPtr& Locale::Payload::HeapPtr::operator=(HeapPtr&& other) noexcept {
+    U_ASSERT(type == eBOGUS);
+    ptr = other.ptr;
+    type = eHEAP;
+    other.type = eBOGUS;
+    return *this;
+}
+
+template <typename BogusFn, typename NestFn, typename HeapFn, typename... Args>
+auto Locale::Payload::visit(BogusFn bogusFn, NestFn nestFn, HeapFn heapFn, Args... args) const {
+    switch (stat.type) {
+        case eBOGUS:
+            return bogusFn(args...);
+        case eNEST:
+            return nestFn(nest, args...);
+        case eHEAP:
+            return heapFn(*heap.ptr, args...);
+        default:
+            UPRV_UNREACHABLE_EXIT;
+    };
+}
+
+void Locale::Payload::copy(const Payload& other) {
+    other.visit([](Payload*) {},
+                [](const Nest& nest, Payload* dst) { dst->nest = nest; },
+                [](const Heap& heap, Payload* dst) { dst->heap = heap; },
+                this);
+}
+
+void Locale::Payload::move(Payload&& other) noexcept {
+    other.visit(
+        [](Payload*, Payload*) {},
+        [](const Nest& nest, Payload* dst, Payload*) { dst->nest = nest; },
+        [](const Heap&, Payload* dst, Payload* src) { dst->heap = std::move(src->heap); },
+        this,
+        &other);
+}
+
+Locale::Payload::~Payload() {
+    if (stat.type == eHEAP) { delete heap.ptr; }
+}
+
+Locale::Payload::Payload(const Payload& other) : stat{eBOGUS} { copy(other); }
+Locale::Payload::Payload(Payload&& other) noexcept : stat{eBOGUS} { move(std::move(other)); }
+
+Locale::Payload& Locale::Payload::operator=(const Payload& other) {
+    if (this != &other) {
+        setToBogus();
+        copy(other);
+    }
+    return *this;
+}
+
+Locale::Payload& Locale::Payload::operator=(Payload&& other) noexcept {
+    if (this != &other) {
+        setToBogus();
+        move(std::move(other));
+    }
+    return *this;
+}
+
+void Locale::Payload::setToBogus() {
+    this->~Payload();
+    stat.type = eBOGUS;
+}
+
+LocalPointer<Locale::Heap> Locale::Payload::release() {
+    U_ASSERT(stat.type == eHEAP);
+    stat.type = eBOGUS;
+    return LocalPointer<Heap>{heap.ptr};
+}
+
+template <> Locale::Nest& Locale::Payload::emplace() {
+    this->~Payload();
+    ::new(&nest) Nest();
+    return nest;
+}
+
+template <typename T, typename... Args> T& Locale::Payload::emplace(Args&&... args) {
+    U_ASSERT(stat.type != eHEAP);
+    heap.ptr = new Heap(std::forward<Args>(args)...);
+    stat.type = heap.ptr == nullptr ? eBOGUS : eHEAP;
+    return heap.ptr;
+}
+
+template <> Locale::Nest* Locale::Payload::get() { return stat.type == eNEST ? &nest : nullptr; }
+template <> Locale::Heap* Locale::Payload::get() { return stat.type == eHEAP ? heap.ptr : nullptr; }
 
 UOBJECT_DEFINE_RTTI_IMPLEMENTATION(Locale)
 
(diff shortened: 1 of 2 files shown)


roubert:
(After this, it'd be possible to make objects with heap allocation a little bit smaller still by refactoring `Payload` / `Heap` to move the `language`, `script`, `region` fields into the payload.)

icu4c/source/common/unicode/locid.h
```diff
@@ -1217,12 +1224,59 @@ class U_COMMON_API_CLASS Locale : public UObject {
...
+
+        struct HeapPtr {
+            ELocaleType type;
+            Heap* ptr;
+
+            HeapPtr& operator=(const Heap& other);
+            HeapPtr& operator=(HeapPtr&& other) noexcept;
+        } heap;
+
+        struct { ELocaleType type; } stat;
```
markusicu, line 1249: optional: what does "stat" mean?

icu4c/source/common/unicode/locid.h
```diff
@@ -1217,12 +1224,59 @@ class U_COMMON_API_CLASS Locale : public UObject {
...
+
+        Payload& operator=(const Payload& other);
+        Payload& operator=(Payload&& other) noexcept;
+
+        void setToBogus();
+        bool isBogus() const { return stat.type == eBOGUS; }
+
+        LocalPointer<Heap> release();
+
+        template <typename T, typename... Args> T& emplace(Args&&... args);
```
markusicu, line 1273: why add a space between `template` and the `<...>` ?

icu4c/source/common/locid.cpp
```diff
@@ -326,12 +326,109 @@ struct Locale::Heap : public UMemory {
...
+Locale::Payload::HeapPtr& Locale::Payload::HeapPtr::operator=(HeapPtr&& other) noexcept {
+    U_ASSERT(type == eBOGUS);
+    ptr = other.ptr;
+    type = eHEAP;
+    other.type = eBOGUS;
+    return *this;
+}
+
+template <typename Fn0, typename Fn1, typename Fn2, typename... Args>
+auto Locale::Payload::visit(Fn0 fn0, Fn1 fn1, Fn2 fn2, Args... args) const {
```
markusicu: Please rename to clarify, e.g.,
- `Fn0 fn0` --> `BogusFn bogusFn`
- `Fn1 fn1` --> `NestFn nestFn`
- etc. or similar

icu4c/source/common/unicode/locid.h
```diff
@@ -1217,12 +1224,59 @@ class U_COMMON_API_CLASS Locale : public UObject {
...
+      private:
+        Nest nest;
+
+        struct HeapPtr {
+            ELocaleType type;
+            Heap* ptr;
+
+            HeapPtr& operator=(const Heap& other);
+            HeapPtr& operator=(HeapPtr&& other) noexcept;
+        } heap;
```
markusicu, line 1247: I find it confusing to read the impl code where `heap` is not of type `Heap`. Please rename this to heapPtr.

icu4c/source/common/locid.cpp
markusicu: Also, I find this construct confusing. Not a blocker, but I think using a simple `switch` in the call sites rather than calling `visit` would actually be better.

icu4c/source/common/locid.cpp
```diff
@@ -326,12 +326,109 @@ struct Locale::Heap : public UMemory {
...
+            return fn2(*heap.ptr, args...);
+        default:
+            UPRV_UNREACHABLE_EXIT;
+    };
+}
+
+void Locale::Payload::copy(const Payload& other) {
+    other.visit([](Payload*) {},
+                [](const Nest& nest, Payload* dst) { dst->nest = nest; },
+                [](const Heap& heap, Payload* dst) { dst->heap = heap; }, this);
```
markusicu: At a minimum, let's not hide the non-fn parameters.
```suggestion
    other.visit([](Payload*) {},
                [](const Nest& nest, Payload* dst) { dst->nest = nest; },
                [](const Heap& heap, Payload* dst) { dst->heap = heap; },
                this);
```

icu4c/source/common/locid.cpp
```diff
@@ -326,12 +326,109 @@ struct Locale::Heap : public UMemory {
...
+}
+
+void Locale::Payload::move(Payload&& other) noexcept {
+    other.visit(
+        [](Payload*, Payload*) {},
+        [](const Nest& nest, Payload* dst, Payload*) { dst->nest = nest; },
+        [](const Heap&, Payload* dst, Payload* src) { dst->heap = std::move(src->heap); }, this, &other);
```
markusicu:
```suggestion
    other.visit(
        [](Payload*, Payload*) {},
        [](const Nest& nest, Payload* dst, Payload*) { dst->nest = nest; },
        [](const Heap&, Payload* dst, Payload* src) { dst->heap = std::move(src->heap); },
        this,
        &other);
```

icu4c/source/common/locid.cpp
```diff
@@ -326,12 +326,109 @@ struct Locale::Heap : public UMemory {
...
+    return LocalPointer<Heap>{heap.ptr};
+}
+
+template <> Locale::Nest& Locale::Payload::emplace() {
+    this->~Payload();
+    ::new(&nest) Nest();
+    return nest;
+}
+
+template <typename T, typename... Args> T& Locale::Payload::emplace(Args&&... args) {
```
markusicu, line 425: If the return type is always `Heap*` can we make that explicit and remove tparam T?

How about calling this function emplaceHeap() for clarity?

icu4c/source/common/locid.cpp
```diff
@@ -326,12 +326,109 @@ struct Locale::Heap : public UMemory {
...
+    stat.type = eBOGUS;
+}
+
+LocalPointer<Locale::Heap> Locale::Payload::release() {
+    U_ASSERT(stat.type == eHEAP);
+    stat.type = eBOGUS;
+    return LocalPointer<Heap>{heap.ptr};
+}
+
+template <> Locale::Nest& Locale::Payload::emplace() {
```
markusicu, line 419: How about calling this function emplaceNest() for clarity?

markusicu (commented):
Thank you!

We could recover the one byte taken from the Nest baseName by merging the type and variantBegin into bit fields of one single byte, but maybe that's too messy.

icu4c/source/common/locid.cpp
roubert: To me, it's an advantage to keep exactly the same interface for this as `std::variant<>` has, the call sites are easy to read and anyone familiar with the standard library will immedately know how it works.

icu4c/source/common/locid.cpp
roubert: To me, it's an advantage to keep exactly the same interface for this as `std::variant<>` has, the call sites are easy to read and anyone familiar with the standard library will immedately know how it works.

icu4c/source/common/unicode/locid.h
roubert: I agree that this naming is suboptimal but I'd like to keep it this way just to reduce churn, refactoring `struct Heap` to move its array fields into the payload will result in the names fitting again afterwards.

icu4c/source/common/unicode/locid.h
roubert: It's a Unix thing, `struct stat` is where metadata is stored. This is only used internally in `union Payload` so it can be named whatever and `stat` was just the first name that came into my mind.

icu4c/source/common/unicode/locid.h
roubert: I think you can guess the answer to this one: This code was formatted by `clang-format`.

icu4c/source/common/locid.cpp
roubert: A `switch` in the call sites would require access to `Payload` internals, which this construct doesn't require. It's a somewhat common construct in other libraries (including `std::visit<>`) and I particularly like how `Locale::getField()` now turns out, it becomes short, simple and wholly isolated from any `Payload` implementation details.

roubert:
> We could recover the one byte taken from the Nest baseName by merging the type and variantBegin into bit fields of one single byte,

While that's true, I can't think of any commonly used locale that would actually fill that additional byte, so it wouldn't be of much value to have it available.

icu4c/source/common/unicode/locid.h
markusicu: I don't remember seeing any other code where that space is inserted but whatever...

markusicu (approved):

roubert:
> (After this, it'd be possible to make objects with heap allocation a little bit smaller still by refactoring `Payload` / `Heap` to move the `language`, `script`, `region` fields into the payload.)

PR #3627

icu4c/source/common/unicode/locid.h
markusicu: removing "stat": https://github.com/unicode-org/icu/pull/3635

PR #3627, opened 2025-09-03
ICU-20392 Refactor struct Heap to store less data on the heap

roubert:
The `language`, `script`, `region` fields all fit within the payload, using storage that currently is filled only with padding, so by moving them there the amount of data that needs to be allocated on the heap is reduced to the arbitrary length fields.

#### Checklist


icu4c/source/common/locid.cpp
```diff
@@ -375,15 +399,14 @@ void Locale::Payload::copy(const Payload& other) {
...
 void Locale::Payload::move(Payload&& other) noexcept {
     other.visit(
-        [](Payload*, Payload*) {},
-        [](const Nest& nest, Payload* dst, Payload*) { dst->nest = nest; },
-        [](const Heap&, Payload* dst, Payload* src) { dst->heap = std::move(src->heap); },
-        this,
-        &other);
+        [](Payload*) {},
+        [](const Nest& nest, Payload* dst) { dst->nest = nest; },
+        [](const Heap& heap, Payload* dst) { dst->heap = std::move(heap); },
```
markusicu: remind me why we can move from a const reference?

icu4c/source/common/locid.cpp
```diff
@@ -2770,15 +2796,15 @@ Locale::setKeywordValue(StringPiece keywordName,
             std::string_view script = heap->script;
             std::string_view region = heap->region;
             if (Nest::fits(localeID.length(), language, script, region)) {
-                LocalPointer<Heap> save = payload.release();
-                U_ASSERT(save.isValid());
-                nest = &payload.emplace<Nest>();
+                int32_t variantBegin = heap->ptr->variantBegin;
+                U_ASSERT(variantBegin >= 0);
+                U_ASSERT(static_cast<size_t>(variantBegin) < sizeof Nest::baseName);
+                nest = &payload.emplace<Nest>(script, region, static_cast<uint8_t>(variantBegin));
```
markusicu: I see what you are doing here, and in this special constructor. It's very clever, but it seems misleading to disguise as a constructor what is really a "move Heap fields to Nest fields" function. It would be clearer / much easier to understand to name it appropriately and go back to just `&payload.emplace<Nest>()` and then calling that function.

icu4c/source/common/locid.cpp
```diff
@@ -256,6 +256,19 @@ U_NAMESPACE_BEGIN
...
 
+Locale::Nest::Nest(std::string_view script, std::string_view region, uint8_t variantBegin)
+    : variantBegin(variantBegin), baseName{'\0'} {
+    // When moving from Heap to Nest the language field can be left untouched
+    // (as it has the same offset in both) and only the script and region fields
+    // need to be copied to their new locations, which is safe to do because the
+    // new locations come before the old locations in memory and don't overlap.
+    static_assert(sizeof Nest::language + sizeof Nest::script <= ULOC_LANG_CAPACITY);
+    static_assert(sizeof Nest::language + sizeof Nest::script + sizeof Nest::region <=
+                  ULOC_LANG_CAPACITY + ULOC_SCRIPT_CAPACITY);
```
markusicu: This would be clearer as
```
static_assert(offsetof(Nest, region) <= offsetof(Heap, script));
static_assert(offsetof(Nest, variantBegin) <= offsetof(Heap, region));
```
because one would see references to Heap without having to inspect how its fields are sized, and because it fits more closely to what the comment says: "the new locations come before the old locations in memory"

icu4c/source/common/locid.cpp
roubert: Right, it needs a `const_cast`. Done.

icu4c/source/common/locid.cpp
roubert: For that to work it'd be necessary to have a special constructor that doesn't initialize the object (the default constructor zero initializes all these fields) and I don't think that having one special constructor plus one special helper function that need to be called one after the other is any clearer than just having one special constructor.

roubert: But maybe a different signature might help make it a little bit more clear that the special constructor isn't an ordinary constructor.

icu4c/source/common/unicode/locid.h
```diff
@@ -1231,15 +1270,7 @@ class U_COMMON_API_CLASS Locale : public UObject {
...
-        struct HeapPtr {
-            ELocaleType type;
-            Heap* ptr;
-
-            HeapPtr& operator=(const Heap& other);
-            HeapPtr& operator=(HeapPtr&& other) noexcept;
-        } heap;
-
+        Heap heap;
         struct { ELocaleType type; } stat;
```
markusicu, line 1274: question: do we actually need the struct stat? or would the type as a union variant suffice?

roubert: Do you mean just not wrapping it in a struct and using `ELocaleType type;` here directly instead? Sure, that would work fine, I just liked the symmetry of all 3 members being structs, it seemed conceptually simpler in my mind. But please feel free to change that if you think otherwise.

markusicu: > Do you mean just not wrapping it in a struct and using `ELocaleType type;` here directly instead?

Yes. I was wondering if there was a C++ reason for the struct-of-one-field.

> Sure, that would work fine, I just liked the symmetry of all 3 members being structs, it seemed conceptually simpler in my mind. But please feel free to change that if you think otherwise.

It's not really important, but it would avoid the confusing "stat" name...

markusicu: - https://github.com/unicode-org/icu/pull/3635

PR #3628, opened 2025-09-03
ICU-20392 Let the compiler do its job by simplifying Locale::Nest

roubert:
There's no need to define constructors, destructors and operators that just do the same thing as the compiler generated default ones do.

#### Checklist

### DIFF
--- a/icu4c/source/common/locid.cpp
+++ b/icu4c/source/common/locid.cpp
@@ -253,20 +253,6 @@ void copyToArray(std::string_view sv, T* that) {
 
 U_NAMESPACE_BEGIN
 
-Locale::Nest::Nest() : language{'\0'}, script{'\0'}, region{'\0'}, variantBegin{0}, baseName{'\0'} {}
-Locale::Nest::~Nest() = default;
-
-Locale::Nest::Nest(const Nest& other) {
-    uprv_memcpy(this, &other, sizeof *this);
-}
-
-Locale::Nest& Locale::Nest::operator=(const Nest& other) {
-    if (this != &other) {
-        uprv_memcpy(this, &other, sizeof *this);
-    }
-    return *this;
-}
-
 void Locale::Nest::init(std::string_view language,
                         std::string_view script,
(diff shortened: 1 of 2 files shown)


eggrobin:
> This is a plain struct that's perfectly fine to `memcpy()`

In practice, perhaps (though it really hides something nastier, see below). But this is UB, so a sufficiently clever compiler may eventually come to bite us with some aliasing analysis in the future.

But much more importantly, the reason why it is not trivially copyable is precisely that we have that copy constructor and that copy assignment operator with their memcpy. In this day and age, this looks like a bad case of trying to be smarter than the compiler. If the compiler can inline that memcpy, it gets turned into the same register moves, so at least it is not pessimizing (although we get a branch from that comparison with `this`). However since we implemented that in the `.cpp`, that will not happen without whole program optimization, and so we will get real calls to the operators and constructors (there will still not be any calls to memcpy happening of course). See https://godbolt.org/z/cj1fxahMT.

So I think it would be an improvement in terms of performance (aside from an elimination of UB) to make the destructor, copy constructor, and copy assignment `= default` in the .h.

And if you try to do that, you will notice some much nastier UB: there is a member `const ELocaleType type`, but the assignment mutates it.  Here I would expect compilers to actually eliminate reads of `type` after an assignment.

roubert:
Done.

icu4c/source/common/locid.cpp
```diff
@@ -253,20 +253,6 @@ void copyToArray(std::string_view sv, T* that) {
 
 U_NAMESPACE_BEGIN
 
-Locale::Nest::Nest() : language{'\0'}, script{'\0'}, region{'\0'}, variantBegin{0}, baseName{'\0'} {}
```
eggrobin, line 256: We might still need the default constructor, or alternatively (probably preferably), initializers in the .h , otherwise a default-constructed Nest might have uninitialized memory rather than NULs.

roubert: Done.

PR #3631, opened 2025-09-04
ICU-23054 Use Clang 21 rather than 19 as our most modern Clang in CI

eggrobin:
Following up on #3615. The -Wnontrivial-memcall was fixed by #3628 earlier today.

#### Checklist

ALLOW_MANY_COMMITS=meow

### DIFF
--- a/icu4c/source/common/unicode/utfiterator.h
+++ b/icu4c/source/common/unicode/utfiterator.h
@@ -754,7 +754,7 @@ public:
             UnitIter &p0, UnitIter &p, const LimitIter &limit) {
         constexpr bool isMultiPass = prv::forward_iterator<UnitIter>;
         // Very similar to U16_NEXT_OR_FFFD().
-        CP32 c = *p;
+        CP32 c = static_cast<CP32>(*p);
         ++p;
         if (!U16_IS_SURROGATE(c)) {
             if constexpr (isMultiPass) {
@@ -785,7 +785,7 @@ public:
     U_FORCE_INLINE static CodeUnits<CP32, UnitIter> decAndRead(UnitIter start, UnitIter &p) {
         // Very similar to U16_PREV_OR_FFFD().
         UnitIter p0 = p;
-        CP32 c = *--p;
+        CP32 c = static_cast<CP32>(*--p);
         if (!U16_IS_SURROGATE(c)) {
             return {c, 1, true, p, p0};
         } else {
@@ -984,7 +984,7 @@ public:
     U_FORCE_INLINE static UnsafeCodeUnits<CP32, UnitIter> readAndInc(UnitIter &p0, UnitIter &p) {
         constexpr bool isMultiPass = prv::forward_iterator<UnitIter>;
         // Very similar to U16_NEXT_UNSAFE().
-        CP32 c = *p;
+        CP32 c = static_cast<CP32>(*p);
         ++p;
         if (!U16_IS_LEAD(c)) {
             if constexpr (isMultiPass) {
@@ -1007,7 +1007,7 @@ public:
     U_FORCE_INLINE static UnsafeCodeUnits<CP32, UnitIter> decAndRead(UnitIter &p) {
         // Very similar to U16_PREV_UNSAFE().
         UnitIter p0 = p;
-        CP32 c = *--p;
+        CP32 c = static_cast<CP32>(*--p);
         if (!U16_IS_TRAIL(c)) {
             return {c, 1, p, p0};
         } else {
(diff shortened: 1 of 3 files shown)


roubert:
I think it's an excellent idea to use the latest stable release as the "modern" version of the compiler, but, well, the code first needs to be fixed so that it all builds and tests pass before that can be done. 😉

markusicu:
> Following up on #3619.

Isn't this a follow-up on PR #3615?
Are you intentionally using the generic ticket here rather than the gcc 15 fix ticket ICU-23189 on 3615?

.github/workflows/icu4c.yml
```diff
@@ -333,24 +333,22 @@ jobs:
...
+            - '21'
           flags:
             - '-std=c++20'
             - '-std=c++20 -stdlib=libc++'
             - '-std=c++23 -stdlib=libc++'
           exclude:
             # Most of C++20 was available in Clang 18, see https://clang.llvm.org/cxx_status.html#cxx20,
             # so testing C++20 with Clang 18 is unlikely to find errors that are found by the
-            # Clang 19 C++20 build.
+            # Clang 21 C++20 build.
```
markusicu: how about reducing the number of places one should remember to update?
```suggestion
            # C++20 build with the newer Clang version.
```

eggrobin:
> Isn't this a follow-up on PR https://github.com/unicode-org/icu/pull/3615?

Yes, it is indeed a follow up on the minitel PR.

> Are you intentionally using the generic ticket here rather than the gcc 15 fix ticket ICU-23189 on 3615?

Yes; the other one is closed, and this is just more version bumping and warning fixing.

> well, the code first needs to be fixed so that it all builds and tests pass before that can be done. 😉

Yes, that seems like a good idea… I somehow thought -Wnontrivial-memcall was the only one, evidently that’s not the case.

markusicu:
> implicit conversion from 'char16_t' to 'char32_t' may change the meaning of the represented code unit [-Werror,-Wcharacter-conversion]

This is pretty stupid. The only potential change in meaning is from a surrogate code _unit_ to a surrogate code _point_. Which is harmless when you look at (assertEquals()) code points. It's only a problem if you somehow forgot to actually decode UTF-16 surrogate pairs.

Can we suppress this warning, at least in the test suites?

markusicu:
> conditional expression between different Unicode character types 'char16_t' and 'char32_t' [-Werror,-Wcharacter-conversion]

in `u'b' : sub<CP32, behavior>(parts[1])`

--> different form of the same silly warning; if we can suppress -Wcharacter-conversion then it goes away; otherwise there are only a few of these, so we could change `u'b'` to `U'b'`

markusicu:
it would be really nice if clang didn't complain about these in the first place!

eggrobin:
> we could change u'b' to U'b'

I think that is a slight improvement since we are really talking about code points here, so I might as well do it.

roubert:
> it would be really nice if clang didn't complain about these in the first place!

If you make a compelling enough argument they might change that: https://github.com/llvm/llvm-project/issues/

markusicu:
> > it would be really nice if clang didn't complain about these in the first place!
> 
> If you make a compelling enough argument they might change that: https://github.com/llvm/llvm-project/issues/

Maybe I could ask @eggrobin with his SG16 connections to do so?

markusicu:
> > we could change u'b' to U'b'
> 
> I think that is a slight improvement since we are really talking about code points here, so I might as well do it.

FYI: I have always used u'b' literals for BMP because I find it looks/reads better. I find it annoying that we have to change code like this just because someone thought they should add a warning about zero-extending Unicode code point values. It is part and parcel of the Unicode design that you can zero-extend from ASCII to Latin-1 to the BMP to full Unicode.

eggrobin:
> someone thought they should add a warning about zero-extending Unicode code point values

The warning has false positives in that case, but it really is about avoiding this:
> if you somehow forgot to actually decode UTF-16 surrogate pairs.

We might be smart enough to avoid that, but the broader community of Clang users may not be. I have certainly encountered (and probably written) bugs of that kind.

markusicu:
Clang 21 release notes: https://releases.llvm.org/21.1.0/tools/clang/docs/ReleaseNotes.html
point to
- https://github.com/llvm/llvm-project/issues/138526

which points to a PR from Corentin who I think is in SG16 :-(

but Marc from Qt complained about this today, which yielded
- https://github.com/eisenwave/cpp-proposals/issues/58

markusicu (approved):
I really dislike the new warning, but if this is what it takes...

markusicu (approved):
:-/

PR #3634, opened 2025-09-05
ICU-23197 rework J DateFormatSymbols eraNames loading as in C, update tests

pedberg-icu:
- Rework ICU4J DateFormatSymbols eraNames loading code to match ICU4C rework in [PR #3600 commit 4](https://github.com/unicode-org/icu/pull/3600/commits/a95018f03db53308c0416f01e8d857688cecfee1)
    - Now that eraNames data are in map form and EraRules has been updated, rework the eraNames loading code originally added in PR [#3588](https://github.com/unicode-org/icu/pull/3588) so the CalendarSink treats the eraNames data as a table (more natural, and needed in C to remove leaks), then a new initEras function handles the conversion to array and also fills in missing values from parent(s). The latter solves the problem with not loading the narrow modern era names in Japanese, for example.
- Then update tests:
    - Remove the logKnownIssue test skips for ICU-23182/ICU-23186
    - Update for fixed Meiji start date

#### Checklist
- [x] Required: Issue filed: ICU-23197

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
@@ -26,6 +26,7 @@ import java.util.TreeMap;
 
 import com.ibm.icu.impl.CacheBase;
 import com.ibm.icu.impl.CalendarUtil;
+import com.ibm.icu.impl.EraRules;
 import com.ibm.icu.impl.ICUData;
 import com.ibm.icu.impl.ICUResourceBundle;
 import com.ibm.icu.impl.SoftCache;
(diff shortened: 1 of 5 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1898,6 +1870,42 @@ private DateFormatSymbols(ULocale desiredLocale, ICUResourceBundle b, String cal
...
+                String eraKey = Integer.toString(eraCode);
+                String eraName = eraNamesTable.get(eraKey);
+                if (eraName != null) {
+                    eraArray[eraCode] = eraName;
+                } else {
+                    eraArray[eraCode] = "";
+                    if (calErasWidthBundle != null) {
+                        ICUResourceBundle calErasWidthKeyBundle = calErasWidthBundle.findWithFallback(eraKey);
+                        if (calErasWidthKeyBundle != null) {
+                            eraName = calErasWidthKeyBundle.getString();
```
markusicu, line 1901: Why do we need to do a whole separate lookup here?
I don't know how the `maps` are collected, but it seems like having the sink read the data should assemble all of the values from the parent bundles into the same eraNamesTable. If not, then why is that not happening?
Is this just filling in era names of a different width?
If so, then a comment would help.

pedberg-icu: @markusicu For a map, the sink does not seem to fill in parent entries for keys that do not exist in the current bundle (I guess it has no way to know what those keys should be be), that is why we need to explicitly fill these in. Also true in ICU4C. I added a comment.

markusicu: When the sink is called at any one level, it does not have access to the data at other levels.
For a table, the sink should collect the data into a Map (or similar), and add data at each level (as it descends to the parents) if the key is still absent.
Once the sink loading is done, the Map should have a full set of entries from all bundles.

markusicu: Thanks for the comment, but it sounds like it's working around a bug in the sink code. It would be better to fix that instead.

pedberg-icu: @markusicu Probably (though I do not see an obvious issue there), but maybe we can do that separately in a later PR so this PR can unblock the changes Shane wants to make (and which will need to be integrated). I can leave the ticket open for a later PR. (I am currently traveling on vacation and have very limited time to investigate this further, but am fine if somebody else wants to)

PR #3646, opened 2025-09-11
ICU-23056 Update urename.h generator to write clang-tidy exception

richgillam:
Update the urename.h generator script to write the modified the #ifdef conditions surrounding the body, with the new clang-tidy exception.

#### Checklist

### DIFF
--- a/icu4c/source/tools/genren/genren.pl
+++ b/icu4c/source/tools/genren/genren.pl
@@ -106,8 +106,9 @@ print HEADER <<"EndOfHeaderComment";
 
 #if !U_DISABLE_RENAMING
 
-// Disable Renaming for Visual Studio's IntelliSense feature, so that 'Go-to-Definition' (F12) will work.
-#if !(defined(_MSC_VER) && defined(__INTELLISENSE__))
+// Disable Renaming for Visual Studio's IntelliSense feature and for LLVM's Clang-Tidy tool, so that
+// 'Go-to-Definition' (F12) and 'include-cleaner' respectively will work.
+#if !(defined(_MSC_VER) && defined(__INTELLISENSE__)) && !defined(__clang_analyzer__)
 
 /* We need the U_ICU_ENTRY_POINT_RENAME definition. There's a default one in unicode/uvernum.h we can use, but we will give
    the platform a chance to define it first.
@@ -250,7 +251,7 @@ foreach(sort keys(%CFuncs)) {
 
 print HEADER <<"EndOfHeaderFooter";
 
-#endif /* !(defined(_MSC_VER) && defined(__INTELLISENSE__)) */
+#endif /* !(defined(_MSC_VER) && defined(__INTELLISENSE__)) && !defined(__clang_analyzer__) */
 #endif /* U_DISABLE_RENAMING */
 #endif /* URENAME_H */
 


markusicu:
> somebody changed the #ifdef conditions surrounding the body

That was from @roubert ICU-23158
- https://github.com/unicode-org/icu/pull/3551

Please don't revert that. Fredrik and I apparently didn't think to update the script that generates this file.

richgillam:
> Please don't revert that. Fredrik and I apparently didn't think to update the script that generates this file.

Happy to oblige.  I thought it was the other way around-- that somebody updated the script without updating the file.  Will you update the script, or do you want me to do that?

markusicu:
> Will you update the script, or do you want me to do that?

Could you please do so, now that you know where that is and how to run it? :-)

richgillam:
Done!  PTAL.

—Rich

mihnita:
While doing the [Regenerate configure](https://unicode-org.github.io/icu/processes/release/tasks/build.html) step in BRS I didn't know where to stop and continued on "Update urename.h" (in the same document).
That is covered by the next BRS step, but it is the first time I do this, so...

What I noticed:
- the doc links to [icu4c/source/tools/genren/README](https://github.com/unicode-org/icu/blob/main/icu4c/source/tools/genren/README), which is severely out of date (talks about `svn export` and deleting `$(LEX)` (which does not exits) from a `Makefile`
- trying to call `make install-header` fails (missing `LETypes.h`)
- the best document on how to solve that seems to be ["Building the Paragraph Layout library with HarfBuzz"](https://unicode-org.github.io/icu/userguide/layoutengine/paragraph.html#building-the-paragraph-layout-library-with-harfbuzz)
- But on the http://harfbuzz.org site the "Build and install HarfBuzz" seems incomplete. There are instructions on how to build, but not how to install
- "Build and install the [icu-le-hb](https://github.com/harfbuzz/icu-le-hb) library" has no instructions on how to do that on the official site.

TLDR: it might be a good thing to update the documentation for "Update urename.h".
For someone doing this the first time without any help it can be a challenge.

I managed to make this work (build and install HarfBuzz, build and install icu-le-hb, manually copying icu-le-hb headers in icu4c/sources/layout.
But I am not sure what I did is correct, not enough to document it.

---

The second thing I noticed is that `icu4c/source/common/unicode/urename.h` is generated by `icu4c/source/tools/genren/genren.pl`
The header file was updated "by hand" in commit bb72a7ec6b.

So I updated the Perl script, and created https://github.com/unicode-org/icu/pull/3649

And I discovered that is conflicted with this one :-)

---

But the "please someone update the documentation" part is would still be nice...

richgillam:
> * the doc links to [icu4c/source/tools/genren/README](https://github.com/unicode-org/icu/blob/main/icu4c/source/tools/genren/README), which is severely out of date (talks about `svn export` and deleting `$(LEX)` (which does not exits) from a `Makefile`

Here's what I've actually been doing (which I believe _is_ covered in the README):

1. Delete "$(LEX)" from the makefile.
2. Run `make install-header`.  It should pass now.
3. Hand-edit the resulting `source/common/unicode/urename.h` file to remove `declspec` and to restore all the stuff that starts with `pl_` (the Harfbuzz stuff).  If the only changes in the new version of that file are the `declspec` and Harfbuzz stuff, then I don't bother with a PR and just mark `urename.h` as having not changed in the BRS spreadsheet; otherwise, I put up a PR (but I take care not to include the removal of `$(LEX)` from the makefile in the PR-- I usually just back out that change after I run the tool).

> But the "please someone update the documentation" part is would still be nice...

Yeah, I should do that.  Maybe the next time I go in to regenerate `urename.h`...

markusicu:
> * trying to call `make install-header` fails (missing `LETypes.h`)

We should not have any make target that tries to install a header file that no longer exists in the repo.

> Delete "$(LEX)" from the makefile.

Can we not delete it from the repo copy of the makefile?

richgillam:
> > Delete "$(LEX)" from the makefile.
> 
> Can we not delete it from the repo copy of the makefile?

I could leave it in there commented out, I guess...  The reason why I haven't made that change permanently is that then all the layout-related APIs get left out of `urename.h`.  I'll try to fix the documentation on the next go-round...

roubert:
> Fredrik and I apparently didn't think to update the script that generates this file.

Oops, it didn't even cross my mind that `genren.pl` might be overwriting those lines, I had simply assumed that it only touched the long list of `#define` statements for all the symbols that should be renamed.

roubert:
> But I am not sure what I did is correct, not enough to document it.

Just like everyone else, I have no idea how this is really supposed to be done, but what I do myself is that I use the following script to first build ICU4C without icu-le-hb and then use that as bootstrapping to build icu-le-hb:

```sh
#!/bin/sh
set -e
PARALLEL='-j -l30'

export CC='clang-19'
export CXX='clang++-19'

cd $HOME/src/unicode

(
  cd icu/icu4c/source
  ./runConfigureICU Linux --prefix="$TMP/icu"
  make $PARALLEL install
  git clean -dX
)

export PKG_CONFIG_PATH="$TMP/icu/lib/pkgconfig"

(
  cd icu-le-hb
  ./autogen.sh
  ./configure --prefix="$TMP/icu-le-hb"
  make $PARALLEL install
  git clean -d
)
```

Then I use the following script to run `runConfigureICU` using the newly built icu-le-hb:

```sh
#!/bin/sh
set -e
PREFIX="$TMP/icu"

export CC='clang-19'
export CXX='clang++-19'
export PKG_CONFIG_PATH="$TMP/icu-le-hb/lib/pkgconfig"

exec ./runConfigureICU Linux --prefix="$PREFIX" --enable-layoutex "$@"
```

With that all done, I then use the following script to run `genren.pl` and regenerate `urename.h`:

```sh
#!/bin/sh
set -e
PARALLEL='-j -l30'

export CC='clang-19'
export CXX='clang++-19'
export PKG_CONFIG_PATH="$TMP/icu-le-hb/lib/pkgconfig"

exec make $PARALLEL -C tools/genren install-header
```

That's my own best guess for how this all is supposed to work. What I like about doing it this way is that it doesn't require any manual copying or editing of any files, in the end _make install-header_ just generates the correct file in the correct place.

PR #3647, opened 2025-09-11
ICU-23214 generate brkitr/root.txt lstm from data

srl295:
Rather than hand editing brkitr/root.txt _every single time_, this changes the cldr-to-icu tooling to generate the data automatically.

Note the items are now in lexical order.

Also note TODOs in ICU-23215 and see ticket for background.

- [X] Required: Issue filed: ICU-23214

### DIFF
--- a/docs/processes/cldr-icu.md
+++ b/docs/processes/cldr-icu.md
@@ -297,17 +297,8 @@ ant copy-cldr-testdata
 5d. NOP
 (This step has been subsumed into 5c above)
 
-5e. For now, manually re-add the `lstm` entries in `data/brkitr/root.txt`
-```sh
-open $ICU4C_DIR/source/data/brkitr/root.txt
-```
-Paste the following block after the dictionaries block and before the final closing '}':
-```
-    lstm{
-        Thai{"Thai_graphclust_model4_heavy.res"}
-        Mymr{"Burmese_graphclust_model5_heavy.res"}
-    }
-```
+5e. NOP
+(This step is no longer necessary, see [ICU-23215](https://unicode-org.atlassian.net/browse/ICU-23215) for details.)
 
 5f. Update hard-coded lists in ICU
 
@@ -499,7 +490,7 @@ rebuilding of other kinds of data and/or code. For example:
 
 If you see a failure such as
 ```
-MeasureUnitTest	testCLDRUnitAvailability	Failure	(MeasureUnitTest.java:3410) : Unit present in CLDR but not available via constant in MeasureUnit: speed-beaufort 
+MeasureUnitTest	testCLDRUnitAvailability	Failure	(MeasureUnitTest.java:3410) : Unit present in CLDR but not available via constant in MeasureUnit: speed-beaufort
 ```
 then you will need to update the C and J library and test code for new measurement
 units, see the procedure at
--- a/icu4c/source/data/brkitr/root.txt
+++ b/icu4c/source/data/brkitr/root.txt
@@ -23,7 +23,7 @@ root{
         Thai:process(dependency){"thaidict.dict"}
     }
     lstm{
-        Thai{"Thai_graphclust_model4_heavy.res"}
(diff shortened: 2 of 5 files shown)


pedberg-icu:
@srl295 @FrankYFTang @markusicu As I recall there was a reason we specifically did _**not**_ want to do this (generate the include of `lstm` data) from source. It may have had to do with Google IP or something, or perhaps something about making it easy to remove the `lstm` data from certain configurations. When I am back I will try to dig up the history on this. But again, as I recall, that is why we have been manually adding it during ICU integration.

markusicu:
I don't remember why the tool doesn't write these lines. It seems like if we didn't want them, we wouldn't have them in the file that's checked into the repo. I am happy to defer others with relevant memories.

srl295:
> @srl295 @FrankYFTang @markusicu As I recall there was a reason we specifically did _**not**_ want to do this (generate the include of `lstm` data) from source. It may have had to do with Google IP or something, or perhaps something about making it easy to remove the `lstm` data from certain configurations. When I am back I will try to dig up the history on this. But again, as I recall, that is why we have been manually adding it during ICU integration.

The files are already in the repo in 

```
./icu4j/main/core/src/test/resources/com/ibm/icu/dev/data/testdata/Burmese_graphclust_model5_heavy.res
./icu4j/main/core/src/test/resources/com/ibm/icu/dev/data/testdata/Thai_graphclust_model4_heavy.res
./icu4j/main/core/src/test/resources/com/ibm/icu/dev/data/testdata/Thai_codepoints_exclusive_model5_heavy.res
```

but all THIS change does is generate the source res text that we expect to have so we can integrate.

pedberg-icu:
@srl295 @FrankYFTang @markusicu Here is what I saw from Frank in a 2021-08-18 entry in an e-mail thread “Re: Have you ever converted the lstm data to ICU4J” (thread involving Frank, me, Markus, Mark):
> The idea is that in the current state of the LSTM, we should NOT include them and just let people try to use them in 70.1 and find out all potential issues, and if we are all happy with it, we could then include them in the future release by default. In other words, it is "available in source but not turned on by default" case.  Therefore, BY DEFAULT we should not put them into the CLDR-ICU integration process since there is nothing we should integrate them for. I do not know how you get into that mode that you will see the error (I suspect you didn't clean in the icu4c/source directory). That two entries in the root.txt (without ":process(dependency)" predicate) should be just two strings in the resources and won't cause any dependency issue in the resource chain.

But presumably we are by now beyond that temporary inclusion for testing and so it seems that we should just update the source to generate the correct inclusion.

srl295:
> @srl295 @FrankYFTang @markusicu Here is what I saw from Frank in a 2021-08-18 entry in an e-mail thread “Re: Have you ever converted the lstm data to ICU4J” (thread involving Frank, me, Markus, Mark):
> 
> > The idea is that in the current state of the LSTM, we should NOT include them and just let people try to use them in 70.1 and find out all potential issues, and if we are all happy with it, we could then include them in the future release by default. In other words, it is "available in source but not turned on by default" case.  Therefore, BY DEFAULT we should not put them into the CLDR-ICU integration process since there is nothing we should integrate them for. I do not know how you get into that mode that you will see the error (I suspect you didn't clean in the icu4c/source directory). That two entries in the root.txt (without ":process(dependency)" predicate) should be just two strings in the resources and won't cause any dependency issue in the resource chain.
> 
> 
> 
> But presumably we are by now beyond that temporary inclusion for testing and so it seems that we should just update the source to generate the correct inclusion.

Since this change writes what's already in repo I don't see how it changes what's included. It just removes the manual step.

pedberg-icu (approved):

pedberg-icu (changes requested):
Need to update the CLDR-ICU integration instructions accordingly

srl295:
> Need to update the CLDR-ICU integration instructions accordingly

done

srl295:
rebased, should pass now

srl295:
@DraganBesevic FYI- no more manually re-adding lstm

PR #3657, opened 2025-09-14
ICU-23157 Do not use GMT Zero Format

yumaoka:
#### Checklist
- [x] Required: Issue filed: ICU-23157


icu4c/source/i18n/tzfmt.cpp
```diff
@@ -1559,7 +1559,8 @@ TimeZoneFormat::formatOffsetLocalizedGMT(int32_t offset, UBool isShort, UnicodeS
         return result;
     }
 
-    if (offset == 0) {
+    // Not using GMT zero format when long (ICU-23157)
+    if (offset == 0 && isShort) {
```
robertbastian: I think we also changed the spec for short, usage of this format should be completely gone.

yumaoka: When I attended the meeting, I thought the conclusion was not to use gmtZeroFormat for long localized GMT format, and use it only short localized GMT format, which is designed to display it in shortest form. Were there any updates since then?

yumaoka: I'm no longer attending CLDR-TC. If the spec was updated, please clarify this in the ICU ticket please.
I checked LDML changes. it looks the implementation note part was dropped, but the spec for gmtZeroFormat is still left in LDML.

yumaoka: Anyway, I now understand what's happening. I can update the code again.

yumaoka: Updated this PR. No longer use GMT zero format.
For now, we still keep APIs related to GMT zero format. We need API change request because these API are already public @stable. These API should be deprecated, but it's too late for ICU 78.

markusicu: In the ticket, I noted this as coming from @sffc:
> GMT short format already omits trailing “:00”; for offset 0 it shows “GMT”

I don't have enough context for whether this means that the short format should continue to use the "GMT zero format". @macchiati @pedberg-icu

markusicu: In the current state of the PR, I see that we used to output "GMT", and now we get short="GMT+0" and long="GMT+00:00". ok for short?

sffc:
The only source of truth for the decision is Rob's comment here: https://unicode-org.atlassian.net/browse/CLDR-18280?focusedCommentId=181756

> Design WG recommendation:
>
> Don’t use gmtZeroFormat anywhere, so we’ll have
>
> O => GMT+0, GMT+5, GMT+5:30
>
> OOOO => GMT+00:00, GMT+05:00, GMT+05:30

If I or anyone else said anything contradictory to that, the Design WG recommendation trumps it.

markusicu:
> The only source of truth for the decision is Rob's comment here [...]
> If I or anyone else said anything contradictory to that, the Design WG recommendation trumps it.

thanks for checking!

icu4c/source/i18n/tzfmt.cpp
yumaoka: This might be a bit confusing, but Europe/London winter time in English may still emit "GMT" with pattern "z". This is because CLDR en.xml has short standard name for metazone GMT.

```
			<metazone type="GMT">
				<long>
					<standard>Greenwich Mean Time</standard>
				</long>
				<short>
					<standard>GMT</standard>
				</short>
			</metazone>
```
For other locales, pattern "z" may emit "GMT+0", but this is a fallback localized GMT offset format when short standard name for metazone GMT is missing in a locale.

robertbastian: I'm aware. We could remove the short name, so that `z` produces "GMT+0" and `v` produces "London Time"

markusicu (approved):

PR #3660, opened 2025-09-15
ICU-23056 fix two doc tags

markusicu:
Found by the tool that Craig ran for finding draft ICU APIs.
These were listed with version N/A.

Unfortunately, there are a lot of unrelated changes here, because my Eclipse
- removed trailing spaces
- added missing `@Override` annotations

which other people's IDEs evidently don't do :-/

#### Checklist

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
@@ -58,10 +58,10 @@ import com.ibm.icu.util.ULocale.Category;
  * obtain a formatter for this calendar.
  *
  * <p>References:<ul>
- * 
+ *
  * <li>Dershowitz and Reingold, <i>Calendrical Calculations</i>,
  * Cambridge University Press, 1997</li>
- * 
+ *
  * <li>The <a href="http://www.tondering.dk/claus/calendar.html">
  * Calendar FAQ</a></li>
  *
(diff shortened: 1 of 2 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/util/IslamicCalendar.java
```diff
@@ -1179,9 +1207,11 @@ private static int firstIslamicStartYearFromGrego(int year) {
         return year - 579 + shift;
     }
 
-    /*
+    /**
+     * @internal
```
markusicu, line 1211: This is the "real" change here. This override method was added in
- https://github.com/unicode-org/icu/pull/3470

icu4j/main/core/src/main/java/com/ibm/icu/util/ChineseCalendar.java
```diff
@@ -1155,6 +1171,12 @@ protected int internalGetMonth(int defaultValue)
         return internalGetMonth();
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @stable ICU 2.8
+     */
+    @Override
```
markusicu, line 1179: This is the "real" change here. This override method was added in
- https://github.com/unicode-org/icu/pull/3348

The parent method Calendar.getActualMaximum() is stable ICU 2.0, but the whole class ChineseCalendar is from 2.8, so I think that makes sense as a version number here.

richgillam (approved):
LOKTM

echeran (approved):
LGTM.  The ICU version tag for the public API looks good. The internal API doesn't need a version number. Everything else is adding `@override` or removing trailing whitespace.

PR #3674, opened 2025-09-23
ICU-23202 MF2 to LDML 48: update tests and :number, :integer, :offset, :percent functions

mihnita:
I separated this in 2 commits.

One updates the test files (now in sync with the spec) and does minimal updates to the number formatter (mostly renaming `math` to `offset`, adding `percent`)
The second one doing some bigger refactoring. Mostly extract in a different method some number formatter code that is shared between format and select, implement `percent` (which is new), make `:offset` work differently than the ICU offset (https://github.com/unicode-org/message-format-wg/issues/1099)

---

Still to update the `:date`, `:time`, and `:datetime`, which now don't accept "bag of options" anymore, but a subset of semantic skeletons.

I don't expect much of a challenge there.
The changes would be isolated to the date formatter, and not big refactoring.
Just taking different public options (`dateLength`, `timePrecision` and so on) and mapping them to existing ICU `DateFormat` skeletons.

So I am optimistic that it will be this week, likely in a day or two.

#### Checklist
- [x] Required: Issue filed: ICU-23202


icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
```diff
@@ -130,10 +142,10 @@ public FormattedPlaceholder format(Object toFormat, Map<String, Object> variable
                 offset = 0;
             }
 
-            Double mathOperand = null;
-            if ("math".equals(kind)) {
-                ResolvedMathOptions resolvedMathOptions = ResolvedMathOptions.of(fixedOptions);
-                mathOperand = resolvedMathOptions.operand;
+            int offsetOperand = 0;
+            if ("offset".equals(kind)) {
```
markusicu: FYI: There is probably an ErrorProne about Yoda conditions...

icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
```diff
@@ -146,44 +158,38 @@ public FormattedPlaceholder format(Object toFormat, Map<String, Object> variable
...
                 // This is also what MessageFormat does.
                 throw new NullPointerException("Argument to format can't be null");
             } else if (toFormat instanceof Double) {
                 if (isInt) toFormat = Math.floor((double) toFormat);
-                double toFormatAdjusted =(double) toFormat - offset;
-                if (mathOperand != null) {
-                    toFormatAdjusted += mathOperand;
-                }
+                double toFormatAdjusted =(double) toFormat - offset + offsetOperand;
+                if (isPercent) toFormatAdjusted *= 100;
```
markusicu: Every "if" body should have `{ curly braces }` ... but we should be reformatting after 78 ...

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/CoreTest.java
```diff
@@ -39,8 +40,9 @@ public class CoreTest extends CoreTestFmwk {
             "spec/functions/date.json",
             "spec/functions/datetime.json",
             "spec/functions/integer.json",
-            "spec/functions/math.json", // FAILS 2 / 16, chaining to select
+            "spec/functions/offset.json", // FAILS 2 / 16, chaining to select
```
markusicu: still fails?

icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
echeran: Ditto here and below. Yeah, maybe the formatter will magically do the work when we use it soon enough.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/CoreTest.java
```diff
@@ -14,6 +14,7 @@
 @SuppressWarnings({"static-method", "javadoc"})
 @RunWith(JUnit4.class)
 public class CoreTest extends CoreTestFmwk {
+    private static final boolean DEBUG = false;
```
echeran, line 17: Should we remove debugging code from tests? I'm not sure what ICU preferred / prefers, or if it doesn't specify.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
```diff
@@ -363,12 +364,13 @@ public void testPluralWithOffset() {
     @Test
     public void testPluralWithOffsetAndLocalVar() {
```
echeran: observation: at this point, this test is practically identical to the one above (both tests use `.local` and no `icu:offset`)

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/CoreTest.java
mihnita: Without such code it is impossible to debug.

Because you get an error, but you don't know in what file it is, or in what test inside that file.
That's why I argued that each test should have a unique ID, but the idea was rejected. Because it is too much work to give and maintain IDs :-(

There might be a way to refactor this, for example pass the file name through the call chain, and show it in `assertEquals`.
But there is still no way to do it if you get an exception, for example.

So I got tired to add it and remove it again and again :-)

It looks like we have this as an established pattern.
See for example my cleanup here: https://github.com/unicode-org/icu/pull/3639

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/MessageFormat2Test.java
mihnita: Thanks, good catch.

Removed.

icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
mihnita: Done (in all MF2 files).

icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
mihnita: Done (in all MF2 files).

Although personally I don't think it's a problem.

The options are
```java
if (kind != null && kind.equals("*")) ...
```
and
```java
if (Objects.equals(kind, "*")) ...
```
None of them looks objectively better.

mihnita:
> rslgtm except for formatting nits not trying to follow the logic or match with the spec

ACK, thanks.

Would be nice if GitHub had a way to "align" pieces of code.
"The big change" is in fact code that moved from under `format` to a separate method, with minimal changes otherwise.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/CoreTest.java
mihnita: No.
I removed the comment, thanks.

mihnita (commented):
Everything implemented, thank you!

mihnita:
Everything implemented, thank you!

icu4j/main/core/src/main/java/com/ibm/icu/message2/NumberFunctionFactory.java
```diff
@@ -247,7 +247,7 @@ private boolean matches(Object value, String key, Map<String, Object> variableOp
                 return false;
             }
 
-            if ("integer".equals(kind)) {
+            if (Objects.equals(kind, "integer")) {
```
markusicu, line 250: just FYI -- if kind cannot be null, you can of course use `kind.equals("...")`

markusicu (approved):

PR #3675, opened 2025-09-23
ICU-23222 Enhance unit alias handling

younies:
# Summary

In this PR, we **introduced unit aliases** into the **unit parser Trie**.  When the parser **encounters a unit alias** (e.g., **"permillion" → "part-per-1e6"**),  it **ignores the alias**, **replaces it with the canonical unit identifier**,  and then **continues parsing** as if the alias had never appeared.  

This guarantees that **all units with aliases are normalized** to their replacements,  ensuring **consistent parsing results** across both aliases and canonical forms.

### Example:

```
Input: "meter-permillion-second"
          │
          ▼
[1] Parser reads "meter"
    ✔ Parsed: [meter]
    Remaining: "permillion-second"

[2] Parser reads "permillion"
    🔎 Alias detected → replace with "part-per-1e6"
    ✔ Parsed so far: [meter]
    Remaining becomes: "part-per-1e6-second"

[3] Continue parsing...
    - "part"   → Parsed: [meter, part]
    - "per"    → Parsed: [meter, part, per]
    - "1e6"    → Parsed: [meter, part, per, 1e6]
    - "second" → Parsed: [meter, part, per, 1e6, second]

[✔] Final Result:
    "meter-part-per-1e6-second"
```
#### Before / After (at a glance)

Input : meter-permillion-second
Output: meter-part-per-1e6-second

#### Mini Parse Tree 

```
UnitsExpression
├─ Unit("meter")
├─ Unit("part")
├─ Token("per")
├─ Const("1e6")
└─ Unit("second")
```

-----------------
# Problem Statement

The problem is that we need to scan a CLDR unit ID (string) for any known aliases and replace each alias with its canonical replacement.

### 1) Brute force

Scan for each alias across the entire string; when an alias is found, replace it in-place (or build a new string with the replacement).

Pros
- Simple to implement.

Cons
- Performance: O(N × M), where N = length of the ID and M = number of aliases (each alias checked against the whole string).

### 2) Dedicated alias Trie

Build a separate Trie containing all aliases. At the start of beside the parsing Trie, run a pass over the CLDR ID to find/replace aliases using this Trie.

Pros
- Completes alias work up-front; parsing code doesn’t need to think about aliases afterward.

Cons
- Extra pass on every parse to check for aliases (often there are none).
- The mutated ID must represent the entire CLDR ID (see Option 3 which does need only the remainig part (unparsed part)).
- Requires maintaining a second Trie alongside the Parsing Trie.

### 3) Integrate aliases into the Parsing Trie (current PR)

Add alias entries directly to the existing Parsing Trie. When the parser encounters an alias token, it ignores that token, splices in the alias’s replacement, and continues parsing the remaining (unparsed) tail of the CLDR ID.

Pros
- No pre-scan for aliases; replacements occur naturally during parsing.
- No need to rewrite/add the entire string—only the unparsed remainder after the alias.
- No separate alias Trie to maintain.

Cons
- Less obvious than Option 2 and requires careful handling of token boundaries and continuation logic.

-----------------
# Implementations

## ICU4J

- Refactor NumberFormatterApiTest to improve test case structure for unit aliases.
- Introduce UnitAliases class to manage unit aliases and their replacements.
- Update UnitsData to load aliases and integrate them into the unit parsing logic.
- Modify LongNameHandler to prepare for future refactoring related to unit aliases.
- Add alias handling in MeasureUnitImpl to support alias replacements during parsing.

## ICU4C

- Introduced a new UnitAliasesSink class to collect unit aliases and their replacements from resource tables.
- Added support for alias handling in the Parser class, allowing for dynamic replacement of unit identifiers.
- Updated the initUnitExtras function to populate global vectors for unit aliases and replacements.
- Enhanced NumberFormatterApiTest to include tests for new unit aliases and ensure compatibility with existing formats.
- Added comments for future refactoring of LongNameHandler to utilize the new aliasing system.

---------------
#### Checklist
- [x] Required: Issue filed: ICU-23222

### DIFF
--- a/icu4c/source/i18n/measunit_extra.cpp
+++ b/icu4c/source/i18n/measunit_extra.cpp
@@ -100,6 +100,9 @@ enum PowerPart {
 // "fluid-ounce-imperial".
 constexpr int32_t kSimpleUnitOffset = 512;
 
+// Trie value offset for aliases, e.g. "portion" replaced by "part"
+constexpr int32_t kAliasOffset = 51200; // This will give a very big space for the units ids.
+
 const struct UnitPrefixStrings {
     const char* const string;
     UMeasurePrefix value;
@@ -255,6 +258,58 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
     int32_t outIndex;
 };
 
+class UnitAliasesSink : public icu::ResourceSink {
+  public:
+    /**
+     * Constructor.
+     * @param unitAliases The output vector of unit alias identifiers (CharString).
+     * @param unitReplacements The output vector of replacements for the unit aliases (CharString).
+     */
+    explicit UnitAliasesSink(MaybeStackVector<CharString> &unitAliases,
+                             MaybeStackVector<CharString> &unitReplacements)
+        : unitAliases(unitAliases), unitReplacements(unitReplacements) {}
+
+    /**
+     * Adds the unit alias key and its replacement to the unitAliases and unitReplacements vectors.
+     * @param key The unit alias identifier (e.g., "meter-and-liter").
+     * @param value Should be a ResourceTable value containing the replacement,
+     *     when ures_getAllChildrenWithFallback() is called correctly for this sink.
+     * @param noFallback Ignored.
+     * @param status The standard ICU error code output parameter.
+     */
+    void put(const char *key, ResourceValue &value, UBool /*noFallback*/,
+             UErrorCode &status) override {
+        if (U_FAILURE(status)) return;
+
+        // Add the unit alias key to the unitAliases vector
+        int32_t keyLen = static_cast<int32_t>(uprv_strlen(key));
+        unitAliases.emplaceBackAndCheckErrorCode(status)->append(key, keyLen, status);
+        if (U_FAILURE(status)) {
+            return;
+        }
+
+        // Find the replacement for this unit alias from the alias table resource.
+        ResourceTable aliasTable = value.getTable(status);
+        if (U_FAILURE(status)) {
+            return;
+        }
+
+        if (!aliasTable.findValue("replacement", value)) {
+            status = U_MISSING_RESOURCE_ERROR;
+            return;
+        }
+
+        int32_t len;
+        const char16_t *uReplacement = value.getString(len, status);
+        unitReplacements.emplaceBackAndCheckErrorCode(status)->appendInvariantChars(uReplacement,
+                                                                                    len, status);
+    }
+
+  private:
+    MaybeStackVector<CharString> &unitAliases;
+    MaybeStackVector<CharString> &unitReplacements;
+};
+
 /**
  * A ResourceSink that collects information from `unitQuantities` in the `units`
  * resource to provide key->value lookups from base unit to category, as well as
@@ -320,6 +375,11 @@ class CategoriesSink : public icu::ResourceSink {
 };
 
 icu::UInitOnce gUnitExtrasInitOnce {};
+
+// Array of unit aliases.
+MaybeStackVector<icu::CharString> gUnitAliases;
+// Array of replacements for the unit aliases.
+MaybeStackVector<icu::CharString> gUnitReplacements;
 
 // Array of simple unit IDs.
 //
@@ -453,6 +513,24 @@ void U_CALLCONV initUnitExtras(UErrorCode& status) {
                                              simpleUnitsCount, b, kSimpleUnitOffset);
     ures_getAllItemsWithFallback(unitsBundle.getAlias(), "convertUnits", identifierSink, status);
 
+    // Populate gUnitAliases and gUnitReplacements.
+    LocalUResourceBundlePointer aliasBundle(ures_open(U_ICUDATA_ALIAS, "metadata", &status));
+    if (U_FAILURE(status)) {
+        return;
+    }
+    UnitAliasesSink aliasSink(gUnitAliases, gUnitReplacements);
+    ures_getAllChildrenWithFallback(aliasBundle.getAlias(), "alias/unit", aliasSink, status);
+    if (U_FAILURE(status)) {
+        return;
+    }
+
+    for (int32_t i = 0; i < gUnitAliases.length(); i++) {
+        b.add(gUnitAliases[i]->data(), i + kAliasOffset, status);
+        if (U_FAILURE(status)) {
+            return;
+        }
+    }
+
     // Build the CharsTrie
     // TODO: Use SLOW or FAST here?
     StringPiece result = b.buildStringPiece(USTRINGTRIE_BUILD_FAST, status);
@@ -479,8 +557,10 @@ public:
           this->fType = TYPE_INITIAL_COMPOUND_PART;
       } else if (fMatch < kSimpleUnitOffset) {
           this->fType = TYPE_POWER_PART;
-      } else {
+      } else if (fMatch < kAliasOffset) {
           this->fType = TYPE_SIMPLE_UNIT;
+      } else {
+          this->fType = TYPE_ALIAS;
       }
   }
 
@@ -505,6 +585,7 @@ public:
       TYPE_POWER_PART,
       TYPE_SIMPLE_UNIT,
       TYPE_CONSTANT_DENOMINATOR,
+      TYPE_ALIAS,
   };
 
   // Calling getType() is invalid, resulting in an assertion failure, if Token
@@ -549,6 +630,11 @@ public:
     int32_t getSimpleUnitIndex() const {
         U_ASSERT(getType() == TYPE_SIMPLE_UNIT);
         return fMatch - kSimpleUnitOffset;
+    }
+
+    int32_t getAliasIndex() const {
+        U_ASSERT(getType() == TYPE_ALIAS);
+        return static_cast<int32_t>(fMatch - kAliasOffset);
     }
 
     // TODO: Consider moving this to a separate utility class.
@@ -673,6 +759,10 @@ public:
             }
 
             if (singleUnitOrConstant.isConstantDenominator()) {
+                if (result.constantDenominator > 0) {
+                    status = kUnitIdentifierSyntaxError;
+                    return result;
+                }
                 result.constantDenominator = singleUnitOrConstant.getConstantDenominator();
                 result.complexity = UMEASURE_UNIT_COMPOUND;
                 continue;
@@ -727,6 +817,9 @@ private:
     // references to that string.
     StringPiece fSource;
     BytesTrie fTrie;
+
+    // Storage for modified source string when aliases are expanded
+    CharString fModifiedSource;
 
     // Set to true when we've seen a "-per-" or a "per-", after which all units
     // are in the denominator. Until we find an "-and-", at which point the
@@ -830,6 +923,19 @@ private:
             return {};
         }
 
+        // Handles the case where the alias replacement begins with "per-".
+        // For example:
+        //    if the alias is "permeter" and the replacement is "per-meter".
+        // NOTE: This case does not currently exist in CLDR, but this code anticipates possible future
+        // additions.
+        if (token.getType() == Token::TYPE_ALIAS) {
+            processAlias(token, status);
+            token = nextToken(status);
+            if (U_FAILURE(status)) {
+                return {};
+            }
+        }
+
         fJustSawPer = false;
 
         if (atStart) {
@@ -923,6 +1029,10 @@ private:
                 singleUnitResult.index = token.getSimpleUnitIndex();
                 break;
 
+            case Token::TYPE_ALIAS:
+                processAlias(token, status);
+                break;
+
             default:
                 status = kUnitIdentifierSyntaxError;
                 return {};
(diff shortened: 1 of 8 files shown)


icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -6012,29 +6012,65 @@ class TestCase {
...
-                new TestCase(MeasureUnit.LITER_PER_100KILOMETERS, "2 liters per 100 kilometers"),
-                new TestCase(MeasureUnit.PART_PER_MILLION, "2 parts per million"),
-                new TestCase(MeasureUnit.MILLIMETER_OF_MERCURY, "2 millimeters of mercury"),
-
-                // Replacements
-                new TestCase("millimeter-ofhg", "2 millimeters of mercury"),
-                new TestCase("liter-per-100-kilometer", "2 liters per 100 kilometers"),
+                // permillion
                 new TestCase("permillion", "2 parts per million"),
-                new TestCase("part-per-million", "2 parts per million"),
```
richgillam, line 6026: Why take this out?  Shouldn't it still work?

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
```diff
@@ -6012,29 +6012,65 @@ class TestCase {
         }
 
         TestCase[] testCases = {
-                // Aliases
-                new TestCase(MeasureUnit.MILLIGRAM_OFGLUCOSE_PER_DECILITER, "2 milligrams per deciliter"),
-                new TestCase(MeasureUnit.MILLIGRAM_PER_DECILITER, "2 milligrams per deciliter"),
-                new TestCase(MeasureUnit.LITER_PER_100KILOMETERS, "2 liters per 100 kilometers"),
-                new TestCase(MeasureUnit.PART_PER_MILLION, "2 parts per million"),
-                new TestCase(MeasureUnit.MILLIMETER_OF_MERCURY, "2 millimeters of mercury"),
```
richgillam, line 6020: Why take these out?  Are they covered somewhere else in here?

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
richgillam: For that matter, shouldn't `portion-per-million` or `portion-per-1e6` still work?

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -575,6 +579,13 @@ private UnitsParser(String identifier) {
 
             }
 
+            // Add aliases
+            List<UnitAliases.Alias> aliases = UnitsData.getAliases();
+            for (int i = 0; i < aliases.size(); i++) {
+                UnitAliases.Alias alias = aliases.get(i);
+                trieBuilder.add(alias.alias, i + UnitsData.Constants.kAliasOffset);
```
richgillam, line 586: What's `kAliasOffset` and why do you need it?

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/UnitAliases.java
```diff
@@ -0,0 +1,97 @@
...
+     * 
+     * @return an unmodifiable list of aliases
+     */
+    public List<Alias> getAliases() {
+        List<Alias> aliasList = new ArrayList<>(mapAliasToReplacement.size());
+        for (Map.Entry<String, String> entry : mapAliasToReplacement.entrySet()) {
+            aliasList.add(new Alias(entry.getKey(), entry.getValue()));
+        }
+        return Collections.unmodifiableList(aliasList);
+    }
```
richgillam, line 78: Who's calling this?

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: It is not taken out—you can see in the code below that all of them are listed. I just wanted to organize them with the other aliases in groups.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: I have added all of them, but they won’t work until CLDR adds both ‘million’ and ‘portion’ to the aliases. See logKnownIssue bellow.

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/MeasureUnitImpl.java
younies: In the parser Trie, we use offsets for each parsing category—for example, offsets for powers, SI prefixes, and unit IDs. These offsets are added to each type index in order to shift their ranges in the output of the trie.

Example: if unit IDs are:
 0 - meter
 1 - foot
 2 - inch

Then, they will be stored in the parsing Trie as:
 meter --> 0 + kSimpleUnitOffset 
 foot  --> 1 + kSimpleUnitOffset
 inch  --> 2 + kSimpleUnitOffset
 ... etc.

Now, since we want to add the aliases, they need to be placed in a different range. Therefore, I have introduced kAliasOffset.

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/UnitAliases.java
younies: This will be needed in the `UnitsData` class in order to retrieve all the aliases to be introduced to the parsing Trie builder.

icu4j/main/core/src/main/java/com/ibm/icu/impl/units/MeasureUnitImpl.java
```diff
@@ -750,6 +775,7 @@ private SingleUnitOrConstant nextSingleUnit() {
 
             // Read tokens until we have a complete SingleUnit or we reach the end.
             while (true) {
+                Token.Type type = token.getType();
```
younies: delete

younies: done

younies:
> How were you handling aliases before, and can you provide a summary explaining what you've changed?
> 
> What's the plan for C++? I'm assuming this isn't a Java-only problem.

I have added a summary for this PR, and I also merged the C++ changes into it instead of creating separate PRs.

richgillam:
> > How were you handling aliases before, and can you provide a summary explaining what you've changed?
> > What's the plan for C++? I'm assuming this isn't a Java-only problem.
> 
> I have added a summary for this PR, and I also merged the C++ changes into it instead of creating separate PRs.

I read your new summary, and it looks good, but it doesn't answer my first question above: How were you handling these aliases before?  Is the answer that we _weren't_ handling aliases before now?

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -255,6 +258,76 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
...
+     * @param status The standard ICU error code output parameter.
+     */
+    void put(const char * /*key*/, ResourceValue &value, UBool /*noFallback*/,
+             UErrorCode &status) override {
+        ResourceTable table = value.getTable(status);
+        if (U_FAILURE(status)) return;
+
+        // Collect the unit aliases from the table resource.
+        const char *unitAlias;
+        for (int32_t i = 0; table.getKeyAndValue(i, unitAlias, value); ++i) {
```
richgillam: You might want to consider using `getAllChildrenWithFallback()` instead of `getAllItemsWithFallback()`.  It'd let you avoid this loop and might help if instances of this resource ever use sideways inheritance.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -255,6 +258,76 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
...
+            int32_t unitAliasLen = static_cast<int32_t>(uprv_strlen(unitAlias));
+            CharString *unitAliasCharStr = unitAliases.emplaceBackAndCheckErrorCode(status);
+            if (U_FAILURE(status)) {
```
richgillam: If you functions you're calling do the right thing, you shouldn't have to call `U_FAILURE()` after every call.  If we've done our job right, each function should check the error code right at the top and dump out without doing anything if it's already showing an error value.  That will usually let you perform a bunch of operations at once and only check the error code at the end, making code like this a lot easier to read.

In this particular case, that'd also allow you to combine these things into one line:

```
unitAliases.emplaceBackAndCheckErrorCode(status)->append(unitAlias, unitAliasLen, status);
```

I personally don't think the `emplaceBack()` thing makes a whole lot of sense (or reads well) when it's split up into two lines like this.  If you keep it as two lines, it seems like you'd be better off just creating a `CharString` on the stack, setting it to the correct value, and then appending it to the `unitAliases` vector.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -923,6 +1034,36 @@ class Parser {
...
+            case Token::TYPE_ALIAS: {
+                auto aliasIndex = token.getAliasIndex();
+                if (aliasIndex < 0 || aliasIndex >= gUnitAliases.length()) {
+                    status = kUnitIdentifierSyntaxError;
+                    return {};
+                }
+                auto replacement = gUnitReplacements[aliasIndex];
+
+                // Create new source string: replacement + remaining unparsed portion
+                fModifiedSource.clear();
```
richgillam: If I understand correctly, you've incorporated the alias mechanism directly into your parser, but then, if you encounter a token with an alias, you just create a new version of the unit string and copy the alias's replacement text into the string you're parsing and continue parsing the _new_ string instead.  I wonder if it'd make things simpler to handle alias replacement in a separate preprocessing step instead of incorporating it directly into your parser?

(I'm not completely sold on this suggestion-- it may actually be simpler the way you're doing it now.)

icu4c/source/test/intltest/numbertest_api.cpp
```diff
@@ -6063,30 +6063,95 @@ void NumberFormatterApiTest::formatUnitsAliases() {
...
+            : measureUnit(nullptr), measureUnitString(muStr), expectedFormat(expected) {}
     } testCases[]{
-        // Aliases
-        {std::make_unique<MeasureUnit>(MeasureUnit::getMilligramOfglucosePerDeciliter()), nullptr, u"2 milligrams per deciliter"},
-        {std::make_unique<MeasureUnit>(MeasureUnit::getMilligramPerDeciliter()), nullptr, u"2 milligrams per deciliter"},
-        {std::make_unique<MeasureUnit>(MeasureUnit::getLiterPer100Kilometers()), nullptr,u"2 liters per 100 kilometers"},
-        {std::make_unique<MeasureUnit>(MeasureUnit::getPartPerMillion()), nullptr, u"2 parts per million"},
-        {std::make_unique<MeasureUnit>(MeasureUnit::getMillimeterOfMercury()), nullptr, u"2 millimeters of mercury"},
-
-        // Some replacements
```
richgillam, line 6074: You no longer have a comment over the new entries below saying what they are.  I'm guessing it'd be "part-per-million"?

richgillam (approved):
I think this looks pretty good.  I'd urge you to consider my suggestions, but I can live with it if this goes in as-is.  Thanks for addressing my previous comments.

icu4c/source/test/intltest/numbertest_api.cpp
younies: good catch, it is "permillion"

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
younies: now, I have removed the logKnownIssue as the new CLDR data has been merged

icu4c/source/i18n/measunit_extra.cpp
younies: very good advice, thanks :) 

done

icu4c/source/i18n/measunit_extra.cpp
younies: The problem is that we need to scan a CLDR unit ID (string) for any known aliases and replace each alias with its canonical replacement.

# Options
-------

## 1) Brute force
---------------
Scan for each alias across the entire string; when an alias is found, replace it in-place (or build a new string with the replacement).

Pros
- Simple to implement.

Cons
- Performance: O(N × M), where N = length of the ID and M = number of aliases (each alias checked against the whole string).


## 2) Dedicated alias Trie
-----------------------
Build a separate Trie containing all aliases. At the start of beside the parsing Trie, run a pass over the CLDR ID to find/replace aliases using this Trie.

Pros
- Completes alias work up-front; parsing code doesn’t need to think about aliases afterward.

Cons
- Extra pass on every parse to check for aliases (often there are none).
- The mutated ID must represent the entire CLDR ID (see Option 3 which does need only the remainig part (unparsed part)).
- Requires maintaining a second Trie alongside the Parsing Trie.


## 3) Integrate aliases into the Parsing Trie (current PR)
-------------------------------------------------------
Add alias entries directly to the existing Parsing Trie. When the parser encounters an alias token, it ignores that token, splices in the alias’s replacement, and continues parsing the remaining (unparsed) tail of the CLDR ID.

Pros
- No pre-scan for aliases; replacements occur naturally during parsing.
- No need to rewrite/add the entire string—only the unparsed remainder after the alias.
- No separate alias Trie to maintain.

Cons
- Less obvious than Option 2 and requires careful handling of token boundaries and continuation logic.

icu4c/source/i18n/measunit_extra.cpp
younies: Great suggestion, done :)

younies:
> > > How were you handling aliases before, and can you provide a summary explaining what you've changed?
> > > What's the plan for C++? I'm assuming this isn't a Java-only problem.
> > 
> > 
> > I have added a summary for this PR, and I also merged the C++ changes into it instead of creating separate PRs.
> 
> I read your new summary, and it looks good, but it doesn't answer my first question above: How were you handling these aliases before? Is the answer that we _weren't_ handling aliases before now?

We were not handling the aliases at the parsing level. We are handling them at the formatting level. I have added a todo to remove/modify this handling after this PR.

icu4c/source/i18n/measunit_extra.cpp
richgillam: Yeah, I think you're right.  Thanks for enumerating all this.

icu4c/source/i18n/measunit_extra.cpp
```diff
@@ -255,6 +258,58 @@ class SimpleUnitIdentifiersSink : public icu::ResourceSink {
...
+        int32_t len;
+        const char16_t *uReplacement = value.getString(len, status);
+        unitReplacements.emplaceBackAndCheckErrorCode(status)->appendInvariantChars(uReplacement,
+                                                                                    len, status);
+    }
+
+  private:
+    MaybeStackVector<CharString> &unitAliases;
+    MaybeStackVector<CharString> &unitReplacements;
+};
```
richgillam, line 311: This looks a lot better!  Thanks for making the change.

richgillam (approved):
Looks good!  Thanks for the changes and explanations.

PR #3677, opened 2025-09-24
ICU-22684 Removed @internal from all definitions that are already dec…

richgillam:
…lared private.

#### Checklist
- [x] Required: Issue filed: ICU-22684

### DIFF
--- a/icu4c/source/common/unicode/brkiter.h
+++ b/icu4c/source/common/unicode/brkiter.h
@@ -647,7 +647,6 @@ protected:
 
 private:
 
-    /** @internal (private) */
     Locale actualLocale;
     Locale validLocale;
     Locale requestLocale;
--- a/icu4c/source/common/unicode/bytestriebuilder.h
+++ b/icu4c/source/common/unicode/bytestriebuilder.h
@@ -150,9 +150,6 @@ private:
     virtual int32_t getMinLinearMatch() const override { return BytesTrie::kMinLinearMatch; }
     virtual int32_t getMaxLinearMatchLength() const override { return BytesTrie::kMaxLinearMatchLength; }
 
-    /**
-     * @internal (private)
-     */
     class BTLinearMatchNode : public LinearMatchNode {
     public:
         BTLinearMatchNode(const char *units, int32_t len, Node *nextNode);
--- a/icu4c/source/common/unicode/caniter.h
+++ b/icu4c/source/common/unicode/caniter.h
@@ -156,13 +156,11 @@ private:
 
     /**
      * Copy constructor. Private for now.
-     * @internal (private)
      */
     CanonicalIterator(const CanonicalIterator& other) = delete;
 
     /**
      * Assignment operator. Private for now.
-     * @internal (private)
      */
     CanonicalIterator& operator=(const CanonicalIterator& other) = delete;
 
--- a/icu4c/source/common/unicode/locid.h
+++ b/icu4c/source/common/unicode/locid.h
@@ -1151,7 +1151,6 @@ private:
      * @param canonicalize whether to call uloc_canonicalize on cLocaleID
      */
     Locale& init(const char* localeID, UBool canonicalize);
-    /** @internal */
     Locale& init(StringPiece localeID, UBool canonicalize);
 
     /*
@@ -1177,7 +1176,6 @@ private:
 
     /**
      * Locale data that can be nested directly within the union Payload object.
-     * @internal
      */
     struct Nest {
         static constexpr size_t SIZE = 32;
@@ -1229,7 +1227,6 @@ private:
 
     /**
      * Locale data that needs to be heap allocated in the union Payload object.
-     * @internal
      */
     struct Heap {
         struct Alloc;
@@ -1265,7 +1262,6 @@ private:
     /**
      * This is kind of std::variant but customized to not waste any space on the
      * discriminator or on any padding, and to copy any heap allocated object.
-     * @internal
      */
     union Payload {
       private:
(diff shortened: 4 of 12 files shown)


richgillam:
I have a PR up now, but the more I think about this, the more I wonder if this is actually worth doing.

I filed ICU-22684 in March of 2024.  At the time, I was trying to figure out a good way to deal with the `U_HIDE_INTERNAL_API` macro.  There was no easy way to tell whether a particular item marked with `@internal` needed to be surrounded with `U_HIDE_INTERNAL_API` or not.  Lots of `@internal` symbols do not have those guards around them, usually because the code won't actually compile without those symbols.  But it's often very hard to tell which symbols these are, unless the author has included extra comments that say.

One subclass of these symbols were class members that were already declared `private`, since it it's private, it's obviously also internal.  But taking all of these out (as I've done here) doesn't even come close to solving the larger problem, and some of those tags might have actual documentary value.

We've since basically abandoned `U_HIDE_INTERNAL_API`, and I'm wondering if we should just leave things that way.  We're using `@internal` to flag a lot of different things, each with different semantics, and I fear that getting this "right" is just going to be more trouble than it's worth at this point and that this PR isn't really adding any value.  It might make more sense to just mark the ticket as "Will Not Fix" and get on with our day.

I'd love to hear other people's opinions on this...

markusicu:
I see that you also have thoughts about this in ICU-20403 “Enhance tests for U_HIDE_ conditionals to catch issues like ICU-20306”.

We have tried to give our users a tool to prevent accidental use of unstable APIs, so that their code does not break when unstable APIs change or disappear. This might be most important for C functions, because users depend on stable C functions exported from the DLLs.

However, I don't really want anyone having to go through a super manual, cumbersome check-and-fix process.

I know that when I wrote a script to check this for draft & deprecated, I tried to extend it to cover internal, but despite futzing with U_FORCE_HIDE_INTERNAL_API I got totally tangled up. If you are waiving the white flag too, then maybe it's too difficult.

Maybe we should try what you suggested in ICU-20403 -- keep "internal" where it's easy to guard, and add a new "internalButWeCannotHelpYouDetectIt" elsewhere.

markusicu:
Having said that, your changes here lgtm.
Except, [doxygen is unhappy](https://github.com/unicode-org/icu/actions/runs/17962894117/job/51089763608?pr=3677).
I don't know why it can't see that these things are private, but I guess that is why these things (a subset of what you removed) do have the "internal" docs.

richgillam:
> Having said that, your changes here lgtm. Except, [doxygen is unhappy](https://github.com/unicode-org/icu/actions/runs/17962894117/job/51089763608?pr=3677). I don't know why it can't see that these things are private, but I guess that is why these things (a subset of what you removed) do have the "internal" docs.

I didn't fix that because I was questioning the whole premise of this fix, especially if I end up adding back a substantial subset of the things I took out.  If we can live until tomorrow, maybe we talk about this in ICU TC and decide then if it's worth it?

richgillam:
> We have tried to give our users a tool to prevent accidental use of unstable APIs, so that their code does not break when unstable APIs change or disappear. This might be most important for C functions, because users depend on stable C functions exported from the DLLs.  ... I know that when I wrote a script to check this for draft & deprecated, I tried to extend it to cover internal, but despite futzing with U_FORCE_HIDE_INTERNAL_API I got totally tangled up. If you are waiving the white flag too, then maybe it's too difficult.

I think it probably is.  We might be able to get somewhere with two (or more) different `@internal` tags, like you were proposing, but I'm worried that'll be complicated too, or that we'll end up with `@internalButWeCantCheckIt` in a majority of spots, making `U_HIDE_INTERNAL_API` of questionable use.

I think much of the problem here stems from the fact that C++ requires us to put many implementation details into public headers.  It might be possible to rearrange stuff so that most of the implementation details are inside implementation files (or internal headers), but I'm guessing there'll be a lot of situations where that's impossible.  Given that, it might be that the best we can do is to document which things are internal and hope users heed that and don't shoot themselves in the foot.  Which is essentially where we are.

That'd either mean that any future uses of `U_HIDE_INTERNAL_API` are on the honor system (we don't check anymore), or that the BRS task is just to check that turning on `U_HIDE_INTERNAL_API` still results in compilable code.  If it doesn't, either we'd have to file a new ticket to get the author of the offending code to fix it, or the BRS person would just have to take out any `U_HIDE_INTERNAL_API`s that cause things to fail to compile and leave it at that.  This wouldn't detect that new code has `U_HIDE_INTERNAL_API` in all the spots where it makes sense, though-- people would still be on the honor system there.

richgillam:
> Given that, it might be that the best we can do is to document which things are internal and hope users heed that and don't shoot themselves in the foot. Which is essentially where we are.

I think this is why I'm balking at landing this PR.  If we can't mechanize this test any better than we have, maybe it's better to leave all these `@internal`s in there-- they do have some documentary value.

markusicu:
> I didn't fix that because I was questioning the whole premise of this fix, especially if I end up adding back a substantial subset of the things I took out. If we can live until tomorrow, maybe we talk about this in ICU TC and decide then if it's worth it?

Of course, let's talk about it.
I was going to just approve, because I like the cleanup, but then looked at the test failure.

markusicu:
ICU-TC discussion today:
- Should try to keep using HIDE_INTERNAL where it’s easy
- Should add a CI check to make sure ICU headers build with HIDE_INTERNAL
- Don’t try to check that `@internal` is matched with HIDE_INTERNAL – no longer an active BRS task
- Rich might try to still remove some of the private internal tags

richgillam:
Okay, I fixed the Doxygen problem, and there seemed to be consensus in ICU TC that this PR was worth doing just for aesthetic reasons.  Would somebody mind reviewing and approving? @markusicu ?

markusicu (approved):

PR #3681, opened 2025-09-24
ICU-21821 un-hardcode scx set in icuexportdata

markusicu:
Add a new internal function that enumerates the data structure for the Script and Script_Extensions properties and uses the C "interface" USetAdder to report the code points that have Script_Extensions data. Test that it matches reality. Use it in icuexportdata, replacing a hardcoded list of these code points with a call to the new function.

USetAdder is a C "interface" for modifying a UnicodeSet, avoiding a code dependency from low-level properties/conversion/... on UnicodeSet. Callers have to "instantiate" it with function pointers.

It's a C "interface" because originally much of the properties and conversion code was in .c files.

#### Checklist
- [x] Required: Issue filed: ICU-21821


markusicu:
@Manishearth FYI

icu4c/source/common/uchar.cpp
```diff
@@ -616,6 +616,33 @@ uscript_getScriptExtensions(UChar32 c,
...
+    }
+    (void) value;
+    return true;
+}
+
+}
+
+// for icuexportdata
+U_CAPI void U_EXPORT2
+uprv_addScriptExtensionCodePoints(const USetAdder *sa, UErrorCode *pErrorCode) {
```
echeran: Named consistently after the property name, which is Script_Extensions

```suggestion
uprv_addScriptExtensionsCodePoints(const USetAdder *sa, UErrorCode *pErrorCode) {
```

markusicu: done

echeran (approved):
LGTM

richgillam:
Belated LGTM.  :-)

PR #3685, opened 2025-09-25
ICU-23159 Changed parseUCARules() in the genrb tool to use a CharStri…

richgillam:
…ng object for its filename manipulation.

Use of CharString should eliminate the possibility of buffer overflow problems.

I tested this by creating a directory with really long filenames (I used the filename in the ticket) and tried building ICU.  The `genrb` tool worked fine, but `genbrk` errored out (but didn't crash).  I also tried it with a normal-length path to confirm that ICU still built properly (that the changed code did the same thing as the original code).

#### Checklist
- [x] Required: Issue filed: ICU-23159

### DIFF
--- a/icu4c/source/tools/genrb/parse.cpp
+++ b/icu4c/source/tools/genrb/parse.cpp
@@ -320,9 +320,8 @@ parseUCARules(ParseState* state, char *tag, uint32_t startline, const struct USt
     struct SResource *result = nullptr;
     struct UString   *tokenValue;
     FileStream       *file          = nullptr;
-    char              filename[256] = { '\0' };
-    char              cs[128]       = { '\0' };
-    uint32_t          line;
+    CharString       filename;
+    uint32_t         line;
     UBool quoted = false;
     UCHARBUF *ucbuf=nullptr;
     UChar32   c     = 0;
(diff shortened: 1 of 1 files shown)


icu4c/source/tools/genrb/parse.cpp
```diff
@@ -345,32 +344,31 @@ parseUCARules(ParseState* state, char *tag, uint32_t startline, const struct USt
     /* make the filename including the directory */
     if (state->inputdir != nullptr)
     {
-        uprv_strcat(filename, state->inputdir);
+        filename.append(state->inputdir, -1, *status);
 
         if (state->inputdir[state->inputdirLength - 1] != U_FILE_SEP_CHAR)
         {
-            uprv_strcat(filename, U_FILE_SEP_STRING);
+            filename.append(U_FILE_SEP_STRING, -1, *status);
```
markusicu:
```suggestion
            filename.append(U_FILE_SEP_CHAR, *status);
```

richgillam:
> one nit
> 
> does this cover both “Buffer Overflow vulnerabilities” reported in the ticket?

Yes.  Both were in the same function, and I changed both locations in that function mentioned in the ticket.

icu4c/source/tools/genrb/parse.cpp
richgillam: I debated whether I could do that.  But since you're suggesting it, I guess I can.  :-)

PR #3695, opened 2025-09-26
ICU-23057 BRS78 Add ICU4J API signature file

mihnita:
#### Checklist
- [x] Required: Issue filed: ICU-23057


icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
```diff
@@ -2,8 +2,38 @@
 // License & terms of use: http://www.unicode.org/copyright.html
 package com.ibm.icu.text;
 
+/**
+ * This provides access for testing to the non-public {@code BreakTransliterator}.
+ *
+ * @internal
+ * @deprecated This API is ICU internal only.
+ */
+@Deprecated
```
markusicu, line 11: Why do we need these docs tags in perf-tests? Why are API tools looking at perf-tests?

mihnita:
**Warning:** in this PR I am also updating `APIChangeReport.html`

That is normally done in BRS line 20 by different people.
But the instructions for the step I'm doing (line 64), [_"Add ICU4J API signature file"_](https://unicode-org.github.io/icu/processes/release/tasks/docs.html#check-in-api-signature-data-file-icu4j) are to _"Resolve any warnings before proceeding."_

By doing that (and touching `BreakTransliteratorAccess` and `DateFormatSymbols`) I am affecting  `APIChangeReport.html`.

And `APIChangeReport.html` is generated by the same script I am running for the signature file (`releases_tools/api_reports.sh`), so there is no extra work.

Maybe we should consider unifying these two BRS tasks?
Or at least change the order and make sure they are done by the same people?

markusicu:
> Maybe we should consider unifying these two steps?

It should be easy to generate the API change report without updating the signature files.

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
mihnita: > Why are API tools looking at perf-tests?

I don't know why, but they do.
And I don't think this is the right time to tinker with that.

My best guess is that the filtering is done on classpath / namespace.

It was not a problem before because all the files in `perf-tests` 
are in `com.ibm.icu.dev.test.perf`.

This is a new file and it is in `com.ibm.icu.text`.
Can't be moved, because it is intentionally in the same package with `BreakTransliterator`, which is not public, it is package-accessible.

mihnita:
> It should be easy to generate the API change report without updating the signature files.

It is a shell script.
And both the signature generation and the API report require the same preliminary steps:
- build the api reporting tools (script)
- run `gatherapi` to collect the signature file (script)
- check bad tagging (@deprecated and @internals) (script)
- fix possible problems (manual) and loop

We can split the script it two, and repeat some of the steps.

But the annotations validations is done on the signature file, and fixing annotations affect the API report.
So the API report should happen after.

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
mihnita: I think I know why this is happening.

The API tools don't have any filtering of their own.

From `GatherAPIData.java`:
`* todo: provide command-line control of filters of which subclasses/packages to process`

The API info is gathered using the Javadoc extensions (taglets / doclets)
So we generate javadoc with our own taglet processors.

The maven javadoc plugin is configured like this:

```
<excludePackageNames>com.ibm.icu.impl,com.ibm.icu.impl.*,com.ibm.icu.dev.*,com.ibm.icu.samples,com.ibm.icu.samples.*</excludePackageNames>
```

That excludes `com.ibm.icu.dev.*`
But (or course) does not exclude `com.ibm.icu.text`.

It is not easy to run it on our release-only sub-projects, because we don't have a `pom.xml` under main.
So we do that in `icu4j/`, which includes everything (samples, demos, perf)

We can run it on `main/core`, `main/collation`, `main/charset` etc.
But not all at once, so we end up with separate reports.

My proposed maven refactoring would help, because by adding a pom.xml in main we can run maven workflows (like javadoc) on `main/` only.

That would be an easy way to do filtering on folder.
(it is unclear if the taglets even have access to folder info, or only to Java stuff like package / class names)

markusicu: I guess worst case we could invent a doc tag for a class that tells the API code to ignore it even if it's public and in what looks like an API package. It should be possible to tag the class and not have to tag every public member of it.

markusicu (approved):

mihnita:
Thanks.
I think this is something we can look into after we release.

icu4j/perf-tests/src/main/java/com/ibm/icu/text/BreakTransliteratorAccess.java
mihnita: Will see what the right fix should be.

This is new. We didn't have classes in the `com.ibm.icu.text` package before (outside main/)
So I guess this is something nobody imagined.

PR #3698, opened 2025-09-29
ICU-23228 Fix bugs from static analyzer

poliudian-iv:
When analyzing chromium code 
During the analysis chromium code with clang-sa, cppcheck and clang-tidy, we find a number of bugs.

Fix bugs from static analyzer (first part)
icu4c/source/common/cmemory.h - "Arguments must not be overlapping buffers"
icu4c/source/i18n/dtptngen.cpp - "Arguments must not be overlapping buffers"
icu4c/source/common/normalizer2impl.cpp - "Dereference of null pointer"(icu4c/source/source/i18n/rulebasedcollator.cpp:1376)
icu4c/source/i18n/uspoof.cpp - "Dereference of null pointer"
icu4c/source/common/uresbund.cpp  - "Access of the field 'fResBuf' at negative byte offset -1" when fResPathLen == 0
icu4c/source/i18n/calendar.cpp - "Division by zero" (getActualMaximum can return 0)

#### Checklist
- [x] Required: Issue filed: ICU-23228
- [x] Required: The PR title must be prefixed with a JIRA Issue number. Example: "ICU-23228 Fix Clang Static Analyzer findings"
- [x] Required: Each commit message must be prefixed with a JIRA Issue number. Example: "ICU-23228 Fix Clang Static Analyzer findings"

### DIFF
--- a/icu4c/source/i18n/rulebasedcollator.cpp
+++ b/icu4c/source/i18n/rulebasedcollator.cpp
@@ -665,8 +665,9 @@ RuleBasedCollator::setReorderCodes(const int32_t *reorderCodes, int32_t length,
     if(length == 1 && reorderCodes[0] == UCOL_REORDER_CODE_NONE) {
         length = 0;
     }
-    if(length == settings->reorderCodesLength &&
-            uprv_memcmp(reorderCodes, settings->reorderCodes, length * 4) == 0) {
+    if (length == settings->reorderCodesLength &&
+            (length == 0 || 
+                uprv_memcmp(reorderCodes, settings->reorderCodes, length * 4) == 0)) {
         return;
     }
     const CollationSettings &defaultSettings = getDefaultSettings();
--- a/icu4c/source/i18n/uspoof.cpp
+++ b/icu4c/source/i18n/uspoof.cpp
@@ -169,7 +169,7 @@ uspoof_clone(const USpoofChecker *sc, UErrorCode *status) {
     }
     if (U_FAILURE(*status)) {
         delete result;
-        result = nullptr;
+        return nullptr;
     }
     return result->asUSpoofChecker();
 }
--- a/icu4c/source/common/normalizer2impl.cpp
+++ b/icu4c/source/common/normalizer2impl.cpp
@@ -585,6 +585,10 @@ const char16_t *
 Normalizer2Impl::decompose(const char16_t *src, const char16_t *limit,
                            ReorderingBuffer *buffer,
                            UErrorCode &errorCode) const {
+    if(src==nullptr) {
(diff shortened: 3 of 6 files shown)


poliudian-iv:
@eggrobin, @markusicu 
Can you look this?
We have another patch in draft, with fixing static analyser bugs and wanna publish it after this

poliudian-iv:
@roubert, @richgillam
Is this PR by any chance lost? Can you look this?

icu4c/source/common/cmemory.h
```diff
@@ -486,6 +486,9 @@ icu::MaybeStackArray<T, stackCapacity>::MaybeStackArray(
 template<typename T, int32_t stackCapacity>
 inline MaybeStackArray <T, stackCapacity>&
 MaybeStackArray<T, stackCapacity>::operator=(MaybeStackArray <T, stackCapacity>&& src) noexcept {
+    if (this == &src) {
```
eggrobin: This is a guard for a self-move, on which see https://github.com/unicode-org/icu/pull/3723#discussion_r2440421009; let’s not go there.

poliudian-iv: Sure, i remove this change

icu4c/source/i18n/rulebasedcollator.cpp
```diff
@@ -665,7 +665,7 @@ RuleBasedCollator::setReorderCodes(const int32_t *reorderCodes, int32_t length,
     if(length == 1 && reorderCodes[0] == UCOL_REORDER_CODE_NONE) {
         length = 0;
     }
-    if(length == settings->reorderCodesLength &&
+    if(reorderCodes && length == settings->reorderCodesLength &&
```
roubert: Does this really do the same thing as before? Before, if both lengths were zero then I'd expect `uprv_memcmp()` to return zero, resulting in a return from the function here. Now that won't happen anymore if `reorderCodes == nullptr`, which seems like an unintentional change in behaviour.

icu4c/source/i18n/uspoof.cpp
```diff
@@ -170,6 +170,7 @@ uspoof_clone(const USpoofChecker *sc, UErrorCode *status) {
     if (U_FAILURE(*status)) {
         delete result;
         result = nullptr;
```
roubert: You should delete this line now when it no longer is used.

icu4c/source/i18n/rulebasedcollator.cpp
poliudian-iv: Hmm, sure, you're right
I replace it to checking length, if length the same and equal to zero, we must return

roubert (approved):
LGTM

PR #3699, opened 2025-09-29
ICU-23054 Remove obsolete TODO

mihnita:
#### Checklist


mihnita:
> I think these sorts of comments are good to keep. It is just not a TODO. I would suggest replacing the TODO with something more like

Sorry, I've only "seen the green" and submitted :-)
Good point, I'll add it back, without TODO.

PR #3702, opened 2025-09-30
ICU-23056 Update known issue log for volume units in Arabic locale

younies:
#### Checklist

### DIFF
--- a/icu4c/source/test/intltest/numbertest_skeletons.cpp
+++ b/icu4c/source/test/intltest/numbertest_skeletons.cpp
@@ -448,7 +448,8 @@ void NumberSkeletonTest::perUnitInArabic() {
 
             status.setScope(skeleton);
             if (u_strcmp(cas1.type, u"volume")==0 || u_strcmp(cas2.type, u"volume")==0) {
-                logKnownIssue("ICU-23104", "Strange handling of part-per-1e9 & volumes in skeletons");
+                logKnownIssue("ICU-23229", " Handling of volume units in Arabic locale are causing "
+                                           "status failure in NumberFormatter");
                 continue;
             }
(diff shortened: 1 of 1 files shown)


icu4c/source/test/intltest/numbertest_skeletons.cpp
```diff
@@ -448,7 +448,8 @@ void NumberSkeletonTest::perUnitInArabic() {
 
             status.setScope(skeleton);
             if (u_strcmp(cas1.type, u"volume")==0 || u_strcmp(cas2.type, u"volume")==0) {
-                logKnownIssue("ICU-23104", "Strange handling of part-per-1e9 & volumes in skeletons");
+                logKnownIssue("ICU-23229", " Handling of volume units in Arabic locale are causing "
```
markusicu, line 451: You can't use the same ticket in the PR description / commit message as what you are pointing to from the logKnownIssue() -- you would create a catch-22.

Assuming you intend to keep the ticket open beyond 78, it's probably best to change the PR+commit to use ICU-23056.

markusicu (changes requested):

icu4c/source/test/intltest/numbertest_skeletons.cpp
younies: Thanks, done

younies: I think that logging an issue with the same issue number should be legitimate. It is logging the issue in the code.

markusicu: It's useless when we have to close the ticket before the problem is fixed.

markusicu (approved):

markusicu:
Merged on your behalf because this is trivial, and we want to move towards rc.

PR #3708, opened 2025-10-02
ICU-23056 fix C API change report untagged

markusicu:
The Markdown file https://github.com/unicode-org/icu/blob/main/icu4c/APIChangeReport.md has a broken table of added APIs. We should try to fix the generator, but this PR should make it readable for the release candidate.

When new APIs are "untagged", the table row has extra line breaks which break the table layout.

When removed APIs are "untagged", the syntax seems off. This does not break the table but does not look good.

Most of the "untagged" functions are *deleted* boilerplate functions, so not actually API at all. Maybe we don't care about these? Except if they are newly deleted, then they used to explicitly exist and be callable. Hm.

The other new "untagged" function is `bool icu::message2::data_model::Binding::isLocal() const` which is part of the entirely `@internal` MF2 C++ API. I am adding internal/tech preview.

#### Checklist


richgillam (approved):
LOKTM

PR #3815, opened 2025-12-18
ICU-23269 Add PAC/BTI tags to generated aarch64 assembler data

rossburton:
Background: PAC and BTI are branch and pointer authentication technologies for aarch64:

https://developer.arm.com/community/arm-community-blogs/b/architectures-and-processors-blog/posts/enabling-pac-and-bti-on-aarch64

For PAC/BTI to be used at runtime, all of the ELF segments loaded need to confirm that they're compatible.  If given the right options both GCC and Clang automatically generate code that is compatible and add the annotations.

However, if data files are built into binary objects with tools/gen* using assembler, then these tags are not present and all of ICU cannot use branch protection or pointer authentication.

As the generated assembler is literally only data, we can just add the tags. This is similar to how Intel CET is enabled in the same file.

#### Checklist
- [x] Required: Issue filed: ICU-23269


rossburton:
Gentle ping. The CRA check is wrong, my employer (Arm) has signed a corporate CLA.

PR #3878, opened 2026-02-27
ICU-11443 WIP: Link detection according to UTS#58.

arnt:
This is a UTS58 link detector in Java. I haven't done the C++ side yet.

The new code is in .../LinkDetector.java and supporting files.

I also haven't done the PR submission chores or looked for a JIRA issue. But it's Friday, 17:25, I need to be somewhere at 18:30, and I want to go out feeling the joy of closure.

My rough plan is to look for a JIRA ticket and do other chores on Tuesday, update the Java implementation according to comments, and when the Java implementation looks good to merge I'll write a corresponding ICU4C implementation. At that point I'll also extend the Ruby implementation on which this is based.

For now this is just to give you a look at the code.


icu4j/main/core/src/main/java/com/ibm/icu/impl/LinkEmailProps.java
```diff
@@ -0,0 +1,71 @@
+// © 2025 and later: Unicode, Inc. and others.
```
macchiati, line 1: Most Unicode properties are supported more directly in ICU, so that additional files and parsing code are not necessary. Need to check with @markusicu as to whether the UTS58 properties are or will be.

icu4j/main/core/src/main/java/com/ibm/icu/text/IanaTlds.java
```diff
@@ -0,0 +1,210 @@
+// © 2025 and later: Unicode, Inc. and others.
```
macchiati, line 1: This might be better architected as a plain text file matching the table on https://www.iana.org/domains/root/db, and then changing this to just construct a HashSet statically from that file. (and maybe no .py file). That way it is a simple drop-in.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/util/LinkDetectorTest.java
```diff
@@ -0,0 +1,444 @@
+// © 2016 and later: Unicode, Inc. and others.
```
macchiati, line 1: Tests like these are better structure by having a plain text file that can be deployed across different implementations including different programming languages. 

The file can have a simple structure, a series of inputs and expected outputs. Either as a semicolon delimited file, or something with a simple format like JSON

macchiati:
BTW, just taking a quick look; will need to dig into the guts of the code later.

icu4j/main/core/src/main/java/com/ibm/icu/impl/LinkEmailProps.java
arnt: Right. If they are, I assume most or all of this can be dropped.

icu4j/main/core/src/main/java/com/ibm/icu/text/IanaTlds.java
arnt: Yes. I optimised for runtime performance. There's much to say for a simple drop-in.

icu4j/main/core/src/main/java/com/ibm/icu/impl/LinkEmailProps.java
arnt: FWIW, the UTS58 properties weren't visible to me when I worked on the code. I see the sense in including them in the core ICU property support and dropping the parsing/tree code here. However, I don't feel that this is my decision.

@markusicu please take it up with others who have a say in this and decide. I'll be happy to rewrite this to use the newly available core ICU support, if that's your decision.

icu4j/main/core/src/main/java/com/ibm/icu/text/IanaTlds.java
arnt: This is your decision, I think.
- As I did it, the runtime will be fast, there won't be ~1k IDN calls and Set.add calls on the first invocation.
- With a simple text file, the .py file at compile time can be much simpler.

I'm a simple contributor, I shouldn't try to make architectural decisions regarding performance at first call vs. build-time complexity.

I think the compile-time script has to stay, though, even if simpler than at present.

The build has to download the IANA file occasionally, without stopping the build if IANA isn't reachable. IIRC I chose to try a download if the local file is older than a week and ignore download failures. A week, a month, I don't think that matters. On one hand I've seen even big projects use a decade-old IANA file because there was no automation, on the other hand downloading every time is both slow and brittle. Anything between those extremes is good enough.

I've seen projects that run an occasional crontab for pom.xml version updates etc. That's another option for pulling iana updates.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/util/LinkDetectorTest.java
arnt: I interpret this as saying that tests are better structured if they have a more distant relationship to the code under test. Yes, more structure, but also more distance, and that's a price.

A few months have passed, but IIRC there are a few tests here that would not apply to my Ruby implementation of UTS58. Would you like me to remove those tests and restructure the rest, and restrict the Java tests to that which a Ruby implementation ought to do?

I can do it if you want, but if I do it I won't feel confident that every then and every else has a unit test.

arnt: I think maybe a good choice is to convert SOME of the unit tests into a table, and leave the ones that aren't easily included as they are. Agree?

arnt:
Hi,

I updated this to match my ruby and javascript implementations.

The most important change now: Somehow I'd forgotten to add the conformance tests to LinkDetectorTest.java. Duh. Now it's there, and the code passes all except two of them (it would be bad to linkify `https://whatever@example.com` as `mailto://whatever@example.com` IMO).

arnt:
I suppose we'd better close this.

PR #3914, opened 2026-03-20
ICU-22511 Fixed an infinite loop problem caused by backward collation…

yumaoka:
… code point iteration problem

#### Checklist
- [x] Required: Issue filed: ICU-22511

### DIFF
--- a/icu4c/source/i18n/utf16collationiterator.cpp
+++ b/icu4c/source/i18n/utf16collationiterator.cpp
@@ -295,9 +295,21 @@ FCDUTF16CollationIterator::previousCodePoint(UErrorCode &errorCode) {
                 return U_SENTINEL;
             }
             c = *--pos;
-            if(CollationFCD::hasLccc(c)) {
-                if(CollationFCD::maybeTibetanCompositeVowel(c) ||
-                        (pos != start && CollationFCD::hasTccc(*(pos - 1)))) {
+            if (CollationFCD::hasLccc(c)) {
+                UBool normalizePrevSegment = CollationFCD::maybeTibetanCompositeVowel(c);
+                if (!normalizePrevSegment && pos != start) {
+                    if (U16_IS_TRAIL(*(pos - 1))) {
+                        // CollationFCD::hasTccc(<low-surrogate>) always returns false.
+                        // To test possible trailing ccc, we need to check high surrogate
+                        // (or previous character for broken surrogate pair).
+                        if (pos - 1 != start) {
+                            normalizePrevSegment = CollationFCD::hasTccc(*(pos - 2));
+                        }
+                    } else {
+                        normalizePrevSegment = CollationFCD::hasTccc(*(pos - 1));
+                    }
+                }
+                if (normalizePrevSegment) {
                     ++pos;
                     if(!previousSegment(errorCode)) {
                         return U_SENTINEL;
--- a/icu4c/source/i18n/utf16collationiterator.cpp
+++ b/icu4c/source/i18n/utf16collationiterator.cpp
@@ -303,7 +303,7 @@ FCDUTF16CollationIterator::previousCodePoint(UErrorCode &errorCode) {
                         // To test possible trailing ccc, we need to check high surrogate
                         // (or previous character for broken surrogate pair).
                         if (pos - 1 != start) {
-                            normalizePrevSegment = CollationFCD::hasTccc(*(pos - 2));
+                            normalizePrevSegment = U16_IS_LEAD(*(pos - 2)) && CollationFCD::hasTccc(*(pos - 2));
                         }
                     } else {
                         normalizePrevSegment = CollationFCD::hasTccc(*(pos - 1));
--- a/icu4c/source/i18n/utf16collationiterator.cpp
+++ b/icu4c/source/i18n/utf16collationiterator.cpp
@@ -295,9 +295,21 @@ FCDUTF16CollationIterator::previousCodePoint(UErrorCode &errorCode) {
                 return U_SENTINEL;
             }
             c = *--pos;
-            if(CollationFCD::hasLccc(c)) {
-                if(CollationFCD::maybeTibetanCompositeVowel(c) ||
-                        (pos != start && CollationFCD::hasTccc(*(pos - 1)))) {
+            if (CollationFCD::hasLccc(c)) {
+                UBool normalizePrevSegment = CollationFCD::maybeTibetanCompositeVowel(c);
+                if (!normalizePrevSegment && pos != start) {
+                    if (U16_IS_TRAIL(*(pos - 1))) {
+                        // CollationFCD::hasTccc(<low-surrogate>) always returns false.
+                        // To test possible trailing ccc, we need to check high surrogate
+                        // (or previous character for broken surrogate pair).
+                        if (pos - 1 != start) {
+                            normalizePrevSegment = U16_IS_LEAD(*(pos - 2)) && CollationFCD::hasTccc(*(pos - 2));
+                        }
+                    } else {
+                        normalizePrevSegment = CollationFCD::hasTccc(*(pos - 1));
+                    }
+                }
+                if (normalizePrevSegment) {
                     ++pos;
                     if(!previousSegment(errorCode)) {
                         return U_SENTINEL;
--- a/icu4c/source/test/intltest/collationtest.cpp
+++ b/icu4c/source/test/intltest/collationtest.cpp
@@ -90,6 +90,7 @@ public:
     void TestCollatorPredicateTypes();
     void TestUCollatorPredicateTypes();
     void TestCollatorMap();
+    void TestColItrInfiniteLoop22511();
 
 private:
     void checkFCD(const char *name, CollationIterator &ci, CodePointIterator &cpi);
@@ -168,6 +169,7 @@ void CollationTest::runIndexedTest(int32_t index, UBool exec, const char *&name,
     TESTCASE_AUTO(TestCollatorPredicateTypes);
     TESTCASE_AUTO(TestUCollatorPredicateTypes);
     TESTCASE_AUTO(TestCollatorMap);
+    TESTCASE_AUTO(TestColItrInfiniteLoop22511);
     TESTCASE_AUTO_END;
 }
 
@@ -1912,6 +1914,7 @@ void CollationTest::TestHang22414() {
         errorCode.reset();
     }
 }
+
 void CollationTest::TestBuilderContextsOverflow() {
     IcuTestErrorCode errorCode(*this, "TestBuilderContextsOverflow");
     // ICU-20715: Bad memory access in what looks like a bogus CharsTrie after
@@ -2122,5 +2125,31 @@ void CollationTest::TestCollatorMap() {
     assertEquals("u16m.size()", 2, um.size());
     assertEquals(R"(u16m["a"])", 2, um[u"a"]);
 }
+
+void CollationTest::TestColItrInfiniteLoop22511() {
+    IcuTestErrorCode errorCode(*this, "TestColItrInfiniteLoop22511");
+    char16_t str1[] = {
+        0x0100, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
+    };
+    char16_t str2[] = {
+        0x0041, 0x0304, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
+    };
+    int32_t num_locales = 0;
+    const icu::Locale* locales = icu::Locale::getAvailableLocales(num_locales);
+    for (int32_t i = 0; i < num_locales; i++) {
+        errorCode.reset();
+        icu::Locale l = locales[i];
+        LocalPointer<Collator> coll(Collator::createInstance(l, errorCode));
+        errorCode.assertSuccess();
+        coll->setStrength(icu::Collator::IDENTICAL);
+        UCollationResult result = coll->compare(
+            str1, sizeof(str1)/sizeof(char16_t),
+            str2, sizeof(str2)/sizeof(char16_t),
+            errorCode);
+        errorCode.assertSuccess();
+        assertEquals(UnicodeString("Locale ") + l.getName(), UCOL_EQUAL, result);
+    }
+}
+
 
 #endif  // !UCONFIG_NO_COLLATION
--- a/icu4c/source/test/intltest/collationtest.cpp
+++ b/icu4c/source/test/intltest/collationtest.cpp
@@ -2128,12 +2128,25 @@ void CollationTest::TestCollatorMap() {
 
 void CollationTest::TestColItrInfiniteLoop22511() {
     IcuTestErrorCode errorCode(*this, "TestColItrInfiniteLoop22511");
-    char16_t str1[] = {
-        0x0100, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
+    const char16_t* testCases[][2] = {
+        {
+            u"\u0100\u032a\u01e0\U00011100\u031c",
+            u"A\u0304\u032a\u01e0\U00011100\u031c"  // Equivalent to above, but U+0100 is decomposed to U+0041 U+0304.
+        },
+        {
+            u"\u0100\u032a\u01e0\xdd00\u031c",      // High surrogate 0xd804 is dropped
+            u"A\u0304\u032a\u01e0\xdd00\u031c"      // Equivalent to above, but U+0100 is decomposed
+        },
+        {
+            u"\u0100\u032a\u01e0\xd804\u031c",      // Low surrogate 0xdd00 is dropped
+            u"A\u0304\u032a\u01e0\xd804\u031c"      // Equivalent to above, but U+0100 is decomposed
+        },
+        {nullptr, nullptr}
     };
-    char16_t str2[] = {
-        0x0041, 0x0304, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
-    };
+
+    StringPiece sp1 = u8"\u0100\u032a\u01e0\U00011100\u031c";   // UTF-8 equivalent to str1a
+    StringPiece sp2 = u8"A\u0304\u032a\u01e0\U00011100\u031c";  // UTF-8 equivalent to str1b
+
     int32_t num_locales = 0;
     const icu::Locale* locales = icu::Locale::getAvailableLocales(num_locales);
     for (int32_t i = 0; i < num_locales; i++) {
@@ -2142,12 +2155,16 @@ void CollationTest::TestColItrInfiniteLoop22511() {
         LocalPointer<Collator> coll(Collator::createInstance(l, errorCode));
         errorCode.assertSuccess();
         coll->setStrength(icu::Collator::IDENTICAL);
-        UCollationResult result = coll->compare(
-            str1, sizeof(str1)/sizeof(char16_t),
-            str2, sizeof(str2)/sizeof(char16_t),
-            errorCode);
+        for (int j = 0; testCases[j][0] != nullptr; j++) {
+            UCollationResult result = coll->compare(testCases[j][0], -1, testCases[j][1], -1, errorCode);
+            errorCode.assertSuccess();
+            assertEquals(UnicodeString("Locale ") + l.getName() + "UTF16 case:" + j, UCOL_EQUAL, result);
+        }
+
+        // Also test the UTF-8 versions
+        UCollationResult result = coll->compareUTF8(sp1, sp2, errorCode);
         errorCode.assertSuccess();
-        assertEquals(UnicodeString("Locale ") + l.getName(), UCOL_EQUAL, result);
+        assertEquals(UnicodeString("Locale ") + l.getName() + " UTF-8", UCOL_EQUAL, result);
     }
 }
 
--- a/icu4c/source/test/intltest/collationtest.cpp
+++ b/icu4c/source/test/intltest/collationtest.cpp
@@ -90,6 +90,7 @@ public:
     void TestCollatorPredicateTypes();
     void TestUCollatorPredicateTypes();
     void TestCollatorMap();
+    void TestColItrInfiniteLoop22511();
 
 private:
     void checkFCD(const char *name, CollationIterator &ci, CodePointIterator &cpi);
@@ -168,6 +169,7 @@ void CollationTest::runIndexedTest(int32_t index, UBool exec, const char *&name,
     TESTCASE_AUTO(TestCollatorPredicateTypes);
     TESTCASE_AUTO(TestUCollatorPredicateTypes);
     TESTCASE_AUTO(TestCollatorMap);
+    TESTCASE_AUTO(TestColItrInfiniteLoop22511);
     TESTCASE_AUTO_END;
 }
 
@@ -1912,6 +1914,7 @@ void CollationTest::TestHang22414() {
         errorCode.reset();
     }
 }
+
 void CollationTest::TestBuilderContextsOverflow() {
     IcuTestErrorCode errorCode(*this, "TestBuilderContextsOverflow");
     // ICU-20715: Bad memory access in what looks like a bogus CharsTrie after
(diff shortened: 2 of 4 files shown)


icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
```diff
@@ -137,8 +137,20 @@ public int previousCodePoint() {
...
                 c = seq.charAt(--pos);
                 if (CollationFCD.hasLccc(c)) {
-                    if (CollationFCD.maybeTibetanCompositeVowel(c)
-                            || (pos != start && CollationFCD.hasTccc(seq.charAt(pos - 1)))) {
+                    boolean normalizePrevSegment = CollationFCD.maybeTibetanCompositeVowel(c);
+                    if (!normalizePrevSegment && pos != start) {
+                        if (Character.isLowSurrogate(seq.charAt(pos - 1))) {
+                            // CollationFCD.hasTccc(<low-surrogate>) always returns false.
+                            // To test possible trailing ccc, we need to check high surrogate
+                            // (or previous character for broken surrogate pair).
```
markusicu, line 145: This feels wrong. If we have a trailing surrogate without a leading one, then we should treat the unpaired surrogate like an unassigned code point -- consistent with what we do elsewhere.

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
```diff
@@ -137,8 +137,20 @@ public int previousCodePoint() {
                 }
                 c = seq.charAt(--pos);
                 if (CollationFCD.hasLccc(c)) {
-                    if (CollationFCD.maybeTibetanCompositeVowel(c)
-                            || (pos != start && CollationFCD.hasTccc(seq.charAt(pos - 1)))) {
+                    boolean normalizePrevSegment = CollationFCD.maybeTibetanCompositeVowel(c);
+                    if (!normalizePrevSegment && pos != start) {
+                        if (Character.isLowSurrogate(seq.charAt(pos - 1))) {
+                            // CollationFCD.hasTccc(<low-surrogate>) always returns false.
+                            // To test possible trailing ccc, we need to check high surrogate
```
markusicu, line 144: good catch!!

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
```diff
@@ -137,8 +137,20 @@ public int previousCodePoint() {
...
+                            // (or previous character for broken surrogate pair).
+                            if (pos - 1 != start) {
+                                normalizePrevSegment = CollationFCD.hasTccc(seq.charAt(pos - 2));
+                            }
```
markusicu, line 150:
```suggestion
                            // CollationFCD.hasTccc(<low-surrogate>) always returns false.
                            // To test possible trailing ccc, we need to check the high (leading) surrogate.
                            if (pos - 1 != start) {
                                char prev = seq.charAt(pos - 2);
                                normalizePrevSegment = Character.isHighSurrogate(prev) && CollationFCD.hasTccc(prev);
                            }
```

icu4c/source/test/intltest/collationtest.cpp
```diff
@@ -2123,4 +2126,30 @@ void CollationTest::TestCollatorMap() {
...
 }
 
+void CollationTest::TestColItrInfiniteLoop22511() {
+    IcuTestErrorCode errorCode(*this, "TestColItrInfiniteLoop22511");
+    char16_t str1[] = {
+        0x0100, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
+    };
+    char16_t str2[] = {
+        0x0041, 0x0304, 0x032a, 0x01e0, 0xd804, 0xdd00, 0x031c
+    };
```
markusicu: Please use `u"string literals"`, using `\U000xxxxx` for the supplementary code point.

icu4c/source/test/intltest/collationtest.cpp
```diff
@@ -2123,4 +2126,30 @@ void CollationTest::TestCollatorMap() {
...
+    for (int32_t i = 0; i < num_locales; i++) {
+        errorCode.reset();
+        icu::Locale l = locales[i];
+        LocalPointer<Collator> coll(Collator::createInstance(l, errorCode));
+        errorCode.assertSuccess();
+        coll->setStrength(icu::Collator::IDENTICAL);
+        UCollationResult result = coll->compare(
+            str1, sizeof(str1)/sizeof(char16_t),
+            str2, sizeof(str2)/sizeof(char16_t),
+            errorCode);
```
markusicu:
```suggestion
        UCollationResult result = coll->compare(str1, str2, errorCode);
```

icu4c/source/test/intltest/collationtest.cpp
```diff
@@ -2123,4 +2126,30 @@ void CollationTest::TestCollatorMap() {
...
+        errorCode.assertSuccess();
+        assertEquals(UnicodeString("Locale ") + l.getName(), UCOL_EQUAL, result);
+    }
```
markusicu, line 2168: Please also call compareUTF8() with the UTF-8 versions of the same strings, to rule out a similar bug in the UTF-8 version of the iterator.

You could use something like
```
auto str1_8 = str1.toUTF8String<std::string>();
```
before the loop,
or write equivalent `u8"string literals"`.
(Note that in C++20, those change their base type to char8_t, but I think we made StringPiece accept that.)

icu4c/source/i18n/utf16collationiterator.cpp
```diff
@@ -295,9 +295,21 @@ FCDUTF16CollationIterator::previousCodePoint(UErrorCode &errorCode) {
...
+            if (CollationFCD::hasLccc(c)) {
+                UBool normalizePrevSegment = CollationFCD::maybeTibetanCompositeVowel(c);
+                if (!normalizePrevSegment && pos != start) {
+                    if (U16_IS_TRAIL(*(pos - 1))) {
+                        // CollationFCD::hasTccc(<low-surrogate>) always returns false.
+                        // To test possible trailing ccc, we need to check high surrogate
+                        // (or previous character for broken surrogate pair).
+                        if (pos - 1 != start) {
+                            normalizePrevSegment = CollationFCD::hasTccc(*(pos - 2));
+                        }
```
markusicu, line 307: see my comments on the Java code

markusicu (commented):
as expected, I got myself tangled in my own cleverness and complexity...

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
yumaoka: OK, I will make sure the code also take care unpaired surrogate case.

icu4c/source/test/intltest/collationtest.cpp
yumaoka: OK. Do you prefer to use \uXXXX notation for characters our of US-ASCII range? This test case checks equality - so it's difficult to see the difference between "Ā" (composed) and "Ā" (decomposed) without using \u notation.

1. u"Ā̪Ǡ̜𑄀";
2. u"Ā̪Ǡ\U00011100̜";
3. u"\u0100\u032a\u01e0\U00011100\u031c" & u"A\u0304\u032a\u01e0\U00011100\u031c"

If we use u"string literal", I personally feel it's easier to use \u or \U notation for all non-ASCII code points to illustrate the issue, just be cause this test string is artificially created test data (multiple combining characters)

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
yumaoka: When nextCodePoint() encounters a standalone low surrogate - 

```
                if (CollationFCD.hasTccc(c)) {
                    if (CollationFCD.maybeTibetanCompositeVowel(c)
                            || (pos != limit && CollationFCD.hasLccc(seq.charAt(pos)))) {
                        --pos;
                        nextSegment();
                        c = seq.charAt(pos++);
                    }
                }
```
CollationFCD.hasTccc(c) returns true. Then, whether normalizing the substring solely depends on the result of CollationFCD.hasLccc(seq.charAt(pos)))  except next position is at the limit nor Tibetan case.

When sub sequence is <1-non-high-surrogate> <2-low-surrogate> <3-combining-mark>

- When pos is at <1-non-high-surrogate>. No matter hasTccc(<1>) returns true or false, the normalization operation is not triggered becasuse hasLccc(<2>) returns false
- When pos is at <2-low-surrogate>, hasTccc(<2>) always returns true. The code unit at next position, in this case <3-combining-mark>, hasLccc(<3>) returns true. So it will trigger the normalization operation starting from <2-low-surrogate>. Is this expected behavior?

icu4c/source/test/intltest/collationtest.cpp
yumaoka: OK - I will add a case of UTF-8 string input.

icu4c/source/test/intltest/collationtest.cpp
markusicu: Writing a string literal with character escapes is fine. It's still much better than writing an array of code units.

Whether to use normal characters or escapes depends on whether you think the intent is clear.
I tend to write literal `ä` for the composite and `a\u0308` for the decomposed form, but `\u00E4` for the composite can be ok as well if that makes the string as a whole clearer.

C++ lets us write `\U...` for supplementary code points -- much better than writing a pair of surrogates. Sadly, Java doesn't support that.

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
markusicu: I am confused. Your code comment says that `hasTccc(<low-surrogate>) always returns false` which sounds right, but I haven't checked how I built that structure.

If that's the case, then `hasTccc(c)` being true would imply that c is either a BMP code point or a leading surrogate. (FYI: I find the "high" and "low" terminology confusing...)

Tibetan vowels are on the BMP, so easy.

I guess for that nextCodePoint() snippet to work properly, we would have to check if c and seq.charAt(pos) are a surrogate pair, and if so, check `hasLccc(seq.charAt(pos))` instead (with another pos vs. limit test).

And if c is a lead surrogate which is not followed by a trail surrogate, then we should treat c has not-hasTccc().

This gets convoluted here in the call site. I guess we should add small CollationFCD functions that check the boundary condition, like `boolean hasTcccFollowedByLccc(seq, c, pos, limit)` and `boolean hasLcccPrecededByTccc(seq, c, start, pos)` where c would be UTF-16 code units.

yumaoka: My understanding above was wrong. When iterating forward, sub region to be normalized will not include an orphan low-surrgate. Then when iterator comes back from the normalization buffer, it will start at the orphan low-surrogate, and it is consumed as an unassinged code point, which won't peak following codes.

The backward iteration should match the behavior. I will add some test codes and figure out what codition should be used for the unpaired surrogate case to match the forwarding behavior.

icu4c/source/test/intltest/collationtest.cpp
yumaoka: I tried to add a test case with unpaired surrogates, but the compiler does not allow this - e.g. `error: \udd00 is not a valid universal character`. So if we want to include the test case with unpaired surrogates, it looks I cannot use u"string-literal".

yumaoka: I understand we just need to use `\xd804` instead of `\udd00` to represent the malformed (unpaired surrogate). I will switch to u"string-literal".

icu4c/source/i18n/utf16collationiterator.cpp
yumaoka: Added the check for high surrogate

icu4c/source/test/intltest/collationtest.cpp
yumaoka: Updated the test case to use using u"string-literal". Also added cases with orphan high/low surrogate. Added compareUTF8() test case.

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
yumaoka: Confirmed your suggestion is sufficient for matching backward behavior with forward.

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
yumaoka: Done

icu4j/main/collate/src/main/java/com/ibm/icu/impl/coll/FCDUTF16CollationIterator.java
markusicu: > My understanding above was wrong. When iterating forward, sub region to be normalized will not include an orphan low-surrgate. Then when iterator comes back from the normalization buffer, it will start at the orphan low-surrogate, and it is consumed as an unassinged code point, which won't peak following codes.

Are you saying that the segments start/end in the middle of surrogate pairs? That would be trouble.

yumaoka: >Are you saying that the segments start/end in the middle of surrogate pairs? That would be trouble.

No. I confirmed this is not the case with tracing the test case.

yumaoka: What I worried about was only for orphan high/low surrogate case. I included the test case, and traced the code. It seems to work as designed. The normalization is never starting at the index between high and low surrogate after my initial fix.

yumaoka:
@markusicu I think I answered all of your questions. I currently don't have anything to update to this PR. When you have time, please check it again. Thanks.

FrankYFTang:
(with correction)
I re-verified the exact UTF-16 code units of the failing test sequence and confirmed that @yumaoka is entirely correct: the trigger string is not a pure BMP sequence.

Specifically, the vi_VN pathological sequence "à̰〭Ǡ̜𑄂" contains the Supplementary (Non-BMP) Code Point 𑄂 (U+11102), which is encoded in UTF-16 precisely as the surrogate pair \uD804\uDD02.

Why the Fix Directs the Root Cause: When FCDUTF16CollationIterator::previousCodePoint() processes this sequence in reverse:

It reads the low surrogate \uDD02.
Prior to this PR, evaluating the preceding high surrogate \uD804 using isolated CollationFCD::hasTccc() checks incorrectly returned false.
This caused the backward iterator to bypass necessary FCD previousSegment() normalization, desynchronizing the iteration boundary with the forward-pass, and triggering the infinite contraction/normalization loop in CollationIterator.
Because the infinite loop is strictly mediated by the UTF-16 surrogate boundary of U+11102, successfully reconstructing and validating the full supplementary code point via U16_IS_TRAIL/U16_IS_LEAD (C++) and isLowSurrogate/isHighSurrogate (Java) directly resolves the FCD desynchronization.

Upon applying this PR's previousCodePoint() patches to an un-capped ICU build, the infinite loop is completely eliminated. The getCollationKey and compare() operations for "à̰〭Ǡ̜𑄂" execute successfully and terminate naturally in < 1 millisecond.

Below are the updated, technically accurate regression unit tests for C++ and Java, correctly documenting the Supplementary U+11102 surrogate pair:

C++ (intltest)
```cpp

void CollationRegressionTest::TestSupplementarySurrogateDiscontiguousLoop() {
    IcuTestErrorCode errorCode(*this, "TestSupplementarySurrogateDiscontiguousLoop");
    // This test string intentionally includes the Supplementary (Non-BMP) Code Point U+11102 (𑄂),
    // encoded in UTF-16 as the surrogate pair \uD804\uDD02. 
    // It validates that FCDUTF16CollationIterator::previousCodePoint() correctly traverses 
    // the surrogate boundary in reverse without triggering FCD normalization desynchronization 
    // or an infinite loop in the CollationIterator.
    icu::UnicodeString s1 = icu::UnicodeString::fromUTF8("̀à̰〭Ǡ̜𑄂\u0010Āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāā́āāā");
    icu::UnicodeString s2 = icu::UnicodeString::fromUTF8("āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāà̰〭Ǡ̜𑄂");
    icu::LocalPointer<icu::Collator> coll(icu::Collator::createInstance(icu::Locale("vi_VN"), status), status);
    if (errorCode.errIfFailureAndReset("createInstance(vi_VN)")) {
        return;
    }
    coll->setStrength(icu::Collator::TERTIARY);
    
    // With PR #3914's FCD surrogate-boundary fix applied, this operation halts 
    // and completes successfully in < 1 ms with U_ZERO_ERROR.
    coll->compare(s1, s2);
}
```
Java (CollationRegressionTest.java)
```java


    @Test
    public void TestSupplementarySurrogateDiscontiguousLoop() {
        // This test string intentionally includes the Supplementary (Non-BMP) Code Point U+11102 (𑄂),
        // encoded in UTF-16 as the surrogate pair \uD804\uDD02. 
        // It validates that FCDUTF16CollationIterator.previousCodePoint() correctly traverses 
        // the surrogate boundary in reverse without triggering FCD normalization desynchronization 
        // or an infinite loop in the CollationIterator.
        String s1 = "̀à̰〭Ǡ̜𑄂\u0010Āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāā́āāā";
        String s2 = "āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāà̰〭Ǡ̜𑄂";
        try {
            Collator coll = Collator.getInstance(new ULocale("vi_VN"));
            coll.setStrength(Collator.TERTIARY);
            
            // With PR #3914's FCD surrogate-boundary fix applied, this operation halts 
            // and completes successfully in < 1 ms without throwing an exception.
            coll.compare(s1, s2);
        } catch (Exception e) {
            errln("Unexpected exception during comparison: " + e.getMessage());
        }
    }
```

yumaoka:
> I checked this approach, but noticed that because the changes in `previousCodePoint()` only apply when stepping across trail surrogates (`if (U16_IS_TRAIL(*(pos - 1)))` in C++ and `Character.isLowSurrogate(seq.charAt(pos - 1))` in Java),

Yes. The original failing test case was caused by this bug.

> it does not address infinite loops or timeouts when `CollationIterator::compare()` encounters pathological sequences composed entirely of Basic Multilingual Plane (BMP) characters (`<= 0xFFFF`).

I didn't know about the failing case with only BMP characters.

> For example, when `CollationIterator::nextCE32FromDiscontiguousContraction()` processes a BMP non-starter sequence, `backwardNumSkipped()` steps `pos` out of the normalized buffer, and on the next `handleNextCE32` call `switchToForward()` re-normalizes the segment again. On a string composed entirely of BMP combining marks and non-starters (`with zero surrogates`), `previousCodePoint()` executes identical logic to pre-patch `main`, and the iterator still gets trapped in an endless cycle.

OK, I can look into this case.

> Here is a 100% BMP regression unit test across C++ and Java where `compare()` still hangs / times out (`> 60 seconds`) with this PR applied:

So there should be yet another logic error if the interator trapped in an endless cycle.

yumaoka:
> Here is a 100% BMP regression unit test across C++ and Java where `compare()` still hangs / times out (`> 60 seconds`) with this PR applied:
> 
> ### C++ (`intltest`)
> ```c++
> void CollationRegressionTest::TestBMPDiscontiguousLoop() {
>     IcuTestErrorCode errorCode(*this, "TestBMPDiscontiguousLoop");
>     // All characters in this test case are on the Basic Multilingual Plane (BMP, <= 0xFFFF) with zero surrogates.
>     icu::UnicodeString s1 = icu::UnicodeString::fromUTF8("̀à̰〭Ǡ̜𑄂\u0010Āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāā́āāā");
>     icu::UnicodeString s2 = icu::UnicodeString::fromUTF8("āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāà̰〭Ǡ̜𑄂");
>     UErrorCode status = U_ZERO_ERROR;
>     icu::LocalPointer<icu::Collator> coll(icu::Collator::createInstance(icu::Locale("vi_VN"), status), status);
>     if (errorCode.errIfFailureAndReset("createInstance(vi_VN)")) {
>         return;
>     }
>     coll->setStrength(icu::Collator::TERTIARY);
>     coll->compare(s1, s2);
> }
> ```
> 
> ### Java (`CollationRegressionTest.java`)
> ```java
>     @Test
>     public void TestBMPDiscontiguousLoop() {
>         // All characters in this test case are on the Basic Multilingual Plane (BMP, <= 0xFFFF) with zero surrogates.
>         String s1 = "̀à̰〭Ǡ̜𑄂\u0010Āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāā́āāā";
>         String s2 = "āāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāāà̰〭Ǡ̜𑄂";
>         try {
>             Collator coll = Collator.getInstance(new ULocale("vi_VN"));
>             coll.setStrength(Collator.TERTIARY);
>             coll.compare(s1, s2);
>         } catch (Exception expected) {
>             // ...
>         }
>     }
> ```

I added these test codes to my working branch, but **I could not reproduce the problem** you mentioned. For just in case, I rebased my local workbranch to the latest ICU HEAD. But I still cannot reproduce the problem.
On Java, the test completed in less than 100ms without the expected exception thrown.

Can you tell me followings?
- Your platform (Mine is Ubuntu 24.04.4 LTS x64)
- Compiler version (Ubuntu clang version 18.1.3)
- Java version (openjdk 21.0.11 2026-04-21)
- Your ICU source revision

yumaoka:
I confirmed the test case provided by Frank is trapped in a loop on ICU4J main/HEAD. Probably same for ICU4C. But the test case had no problems with my fix. I will investigate why the test did not cause the problem with my fix.

yumaoka:
@FrankYFTang I carefully checked your test case. A substring of your test case "à̰〭Ǡ̜𑄂" actually triggered the infinite loop in ICU4J main branch. This problem can be reproduced with getCollationKey (compare method eventually generates collation key incrementally).
```
coll.getCollationKey("à̰〭Ǡ̜𑄂");
```
This test case string is actually below in UTF-16

- 00E0
- 302D
- 0330
- 01E0
- D804 - DD02
- 031C

So you said the test case only contains BMP characters (U+11102 = D804-DD02), but it's actually not. And with the fix in this PR, the problem is resolved.

Could you please apply the fix in this PR and try it again?

FrankYFTang:
> @FrankYFTang I carefully checked your test case. A substring of your test case "à̰〭Ǡ̜𑄂" actually triggered the infinite loop in ICU4J main branch. This problem can be reproduced with getCollationKey (compare method eventually generates collation key incrementally).
> 
> ```
> coll.getCollationKey("à̰〭Ǡ̜𑄂");
> ```
> 
> This test case string is actually below in UTF-16
> 
> * 00E0
> * 302D
> * 0330
> * 01E0
> * D804 - DD02
> * 031C
> 
> So you said the test case only contains BMP characters (U+11102 = D804-DD02), but it's actually not. And with the fix in this PR, the problem is resolved.
> 
> Could you please apply the fix in this PR and try it again?

I verify it fix

FrankYFTang (approved):
LGTM

FrankYFTang:
@markusicu could you also look at this again? thanks

markusicu (approved):
Sorry for the extreme delay :-(
I read the comment thread again, and the updated code.
Looks good and makes sense!

PR #3925, opened 2026-04-04
ICU-23247 Fix Errorprone DefaultCharset error

mihnita:
#### Checklist

### DIFF
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatDataDrivenTest.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatDataDrivenTest.java
@@ -20,7 +20,10 @@ import java.math.BigDecimal;
 import java.math.RoundingMode;
 import java.text.ParsePosition;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 
+@RunWith(JUnit4.class)
 public class NumberFormatDataDrivenTest {
 
     private static ULocale EN = new ULocale("en");
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 
 /**
  * A reader for text resource data in the current package or the package of a given class object.
@@ -95,7 +96,7 @@ public class ResourceReader implements Closeable {
         try {
             InputStreamReader isr =
                     (encoding == null)
-                            ? new InputStreamReader(is)
+                            ? new InputStreamReader(is, StandardCharsets.UTF_8)
                             : new InputStreamReader(is, encoding);
 
             this.reader = new BufferedReader(isr);
@@ -233,7 +234,7 @@ public class ResourceReader implements Closeable {
 
         InputStreamReader isr =
                 (encoding == null)
-                        ? new InputStreamReader(is)
+                        ? new InputStreamReader(is, StandardCharsets.UTF_8)
                         : new InputStreamReader(is, encoding);
         reader = new BufferedReader(isr);
         lineNo = 0;
--- a/icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/layout/LigatureTree.java
+++ b/icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/layout/LigatureTree.java
@@ -17,6 +17,7 @@ import com.ibm.icu.text.UTF16;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Writer;
+import java.nio.charset.Charset;
 
 public class LigatureTree {
     static class Lignode {
@@ -221,8 +222,7 @@ public class LigatureTree {
     }
 
     public void dump() {
-        PrintWriter pw = new PrintWriter(System.out);
-
+        PrintWriter pw = new PrintWriter(System.out, true, Charset.defaultCharset());
         root.dump(pw, 0);
         pw.flush();
     }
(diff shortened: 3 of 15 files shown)


icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatDataDrivenTest.java
```diff
@@ -20,7 +20,10 @@
 import java.math.RoundingMode;
 import java.text.ParsePosition;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
 
+@RunWith(JUnit4.class)
 public class NumberFormatDataDrivenTest {
```
markusicu, line 27: Did this test just not run before??

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/layout/LigatureTree.java
```diff
@@ -17,6 +17,7 @@
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.io.Writer;
+import java.nio.charset.Charset;
 
 public class LigatureTree {
```
markusicu, line 22: FYI -- I suspect that this class builds data for the long-deleted ICU layout engine.
Might be true for this whole ".../tool/layout" package.

icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
```diff
@@ -95,7 +96,7 @@ public ResourceReader(InputStream is, String resourceName, String encoding) {
         try {
             InputStreamReader isr =
                     (encoding == null)
-                            ? new InputStreamReader(is)
+                            ? new InputStreamReader(is, StandardCharsets.UTF_8)
```
markusicu, line 99: I am not sure about the changes in this file.
Where is the ResourceReader used?
Is changing from the default charset to UTF-8 a library-user-visible behavior change?
It might be safer to use Charset.defaultCharset() here.

icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/format/NumberFormatDataDrivenTest.java
mihnita: Yes.

I want to propose and make a cleanup of all tests.

Add `@RunWith` where it is missing, and rename some classes that don't start / end with Test, but contain `@Test` or `@RunWith` or they look like contain tests.

Because of unconventional naming we need to configure `maven-surefire-plugin`:
```xml
<plugin>
  <artifactId>maven-surefire-plugin</artifactId>
  <version>${maven-surefire-plugin.version}</version>
  <configuration>
    <includes>
      <!-- Some test files start with "IntlTest...". Many don't fit
        the Surefire test file name default patterns. To be safe, use
        a catchall pattern instead of the default patterns.

        FYI, the default patterns are documented on the website of the
        testing plugin:
        https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html
      -->
      <include>**/*.java</include>
    </includes>
  </configuration>
</plugin>
```

Convention over configuration.

For some reason this is the one that was reported as error.

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/layout/LigatureTree.java
mihnita: Probably.
I wanted to ask if it is OK to remove the whole folder.
But for now I fixed them.

icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
mihnita: `public ResourceReader(InputStream is, String resourceName, String encoding)` is never used in ICU proper.

It is public, so it is possible that someone uses it "outside".
But it is also `impl`.

What if I update the comment saying that the if an encoding is not specified then UTF-8 is used?

`_reset()` is used from `public ResourceReader(Class<?> rootClass, String resourceName, String encoding)`, `public ResourceReader(Class<?> rootClass, String resourceName)`, and `public void reset()`. And indirectly from all constructors (except the one above).

---

There is only constructor if this class used once, in `com.ibm.icu.dev.test.format.NumberFormatTest::TestCases()`:
`ResourceReader reader = new ResourceReader(is, caseFileName, "utf-8");`

The class is used (but not constructed) in `com.ibm.icu.impl.data.TokenIterator`, which is only used by `NumberFormatTest::TestCases()`

So overall the only class using `TokenIterator` and `ResourceReader` is `NumberFormatTest`.

We have tests for two classes that the ICU code does not use.

---

It feels wrong to try and read resources as non-UTF-8.
The resource files we produce are utf-8 when they contain non-ascii characters.
Reading them with anything else is a bug.

So this would only work correctly when someone use their own resources encoded as non-UTF-8 reading them with an `impl` class.

icu4j/tools/misc/src/main/java/com/ibm/icu/dev/tool/layout/LigatureTree.java
mihnita: These classes are also unfriendly to use.

There are 6 classes with `main` methods in that folder (`BuildMirroringTables`, `CanonGSUBBuilder`, `OpenTypeTagBuilder`, `ScriptNameBuilder`, `ShapingTypeBuilder`, `ThaiStateTableBuilder`)

They are not mentioned in `<icu_root>/docs` or anywhere else (readme, etc).

They don't have a help if you run them, they just fail.
They try to dump the result files in `src/com/ibm/icu/dev/tool/layout` (which does not exist), but they don't create that folder, and they don't tell you why they fail.

I suppose that the `src/com/ibm/icu/dev/tool/layout` folder existed before the maven migration (it is now `src/main/java/com/ibm/icu/dev/tool/layout`)
So nobody tried to use them from 2023-08-24, when I did the maven move.

They all still run, and produce output files (once you figure out how to invoke them).
And the outputs are the same before / after my change (except for `ScriptNameBuilder`, which failed before and after).

icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
markusicu: If this code is only used from a test, then this change is fine.

It would be great if you could make a note (file a ticket?) for moving ResourceReader and TokenIterator out into the test package, next to NumberFormatTest. No one benefits from test-only code in the library.

markusicu (approved):

icu4j/main/core/src/main/java/com/ibm/icu/impl/data/ResourceReader.java
mihnita: Created https://unicode-org.atlassian.net/browse/ICU-23362 and asked my email to take it.

mihnita:
Thank you!
M

PR #3926, opened 2026-04-04
ICU-23358 Make RBNF lenient parsing of ignorables more consistent between locales

grhoten:
Right now, there is inconsistent usage of the `%%lenient-parse` rule between locales in RBNF. The vast majority of them define the same set of characters as ignorable, except German, which adds a little extra for vowels with the umlaut. We should change the lenient parsing to not use the collator by default, unless it defines the `%%lenient-parse` rule. The letter folding, and the ignorable characters should be handled separately from a collator.

This improves the parsing time, and it usually reduces the overall heap usage when parsing leniently. All existing tests continue to pass with these changes.

#### Checklist
- [x] Required: Issue filed: ICU-23358


icu4c/source/data/rbnf/de.txt
```diff
@@ -5,7 +5,7 @@ de{
     RBNFRules{
         SpelloutRules{
             "%%lenient-parse:"
-            "&ue=\u00FC&ae=\u00E4&oe=\u00F6&[last primary ignorable ] << ' ' << ',' << '-' << '\u00AD';"
+            "&ue=\u00FC&ae=\u00E4&oe=\u00F6;"
```
grhoten, line 8: This is the only `%%lenient-parse` left. This one is hard to get rid of.

icu4c/source/data/rbnf/it.txt
```diff
@@ -14,14 +14,13 @@ it{
...
         SpelloutRules{
-            "%%lenient-parse:"
-            "&[last primary ignorable ] << ' ' << ',' << '-' << '\u00AD';"
             "%spellout-numbering-year:"
             "x.x: =0.0=;"
             "0: =%spellout-numbering=;"
             "%spellout-numbering:"
             "-x: meno >>;"
             "x.x: << virgola >>;"
+            "Inf: infinito;"
```
grhoten, line 23: This change got pulled in from the latest version of CLDR. Only the latest RBNF data is being brought in from CLDR.

icu4c/source/i18n/nfrule.cpp
```diff
@@ -1368,13 +1367,69 @@ NFRule::prefixLength(const UnicodeString& str, const UnicodeString& prefix, UErr
...
+            }
+            ++strIdx;
+            ++prefixIdx;
         }
+        if (matched) {
+            return strIdx;
+        }
+
+        // go through all this grief if we're in lenient-parse mode
+#if !UCONFIG_NO_COLLATION
```
grhoten, line 1465: The UCONFIG_NO_COLLATION option has been narrowed down to this smaller segment in both ICU4C and ICU4J. If collation is disabled, only German would be affected at this time. All of the mixing of ignorable characters and case folding is happening above.

icu4c/source/i18n/nfrule.cpp
```diff
@@ -1472,65 +1527,9 @@ NFRule::prefixLength(const UnicodeString& str, const UnicodeString& prefix, UErr
         fprintf(stderr, "prefix length: %d\n", result);
 #endif
         return result;
-#if 0
```
grhoten, line 1475: Remove permanently dead code collation code. It can be resurrected from the history if needed.

icu4c/source/test/intltest/itrbnf.cpp
```diff
@@ -1391,8 +1386,22 @@ IntlTestRBNF::TestSpanishSpellout()
...
+        formatter->setLenient(true);
+        static const char* lpTestData[][2] = {
+                // NFD form of "dieciséis": s + e + combining acute + i + s
+                { "diecise\\u0301is", "16" },
+                // Mixed case with NFD accent
+                { "DIECISE\\u0301IS", "16" },
+                // NFD form of "veintiséis"
+                { "veintise\\u0301is", "26" },
+                // Standard NFC form should still work
+                { "diecis\\u00e9is", "16" },
```
richgillam, line 1403: Dumb question: Should lenient parsing also match "dieciseis" without the accent?  Is that handled somewhere that I missed?

richgillam (dismissed):
Looks good!

icu4c/source/test/intltest/itrbnf.cpp
grhoten: That's a good question. I've added a test to verify that "dieciseis" matches 16 also. From my experience, when leniently parsing is turned on, it's fine to match when the accents are missing. The real world experience shows that people want to type fast, and they may not take the time to write the accent when autocorrect is off.

We try to be strict on the formatting. The default parsing wouldn't allow this, but you get what you asked for when lenient parsing is turned on.

richgillam: So it already works?  That's great!  I couldn't tell that from the code.  And yes, I totally agree with your reasoning here.

richgillam (approved):
Still LGTM.

icu4c/source/test/intltest/itrbnf.cpp
grhoten: Yes, the new test that I added after you asked this question confirms that this works. I didn't check if the old behavior worked this way.

PR #3928, opened 2026-04-06
ICU-23348 Fix thread safety issues with parsing and formatting in RuleBasedNumberFormat

grhoten:
I recommend moving the collator construction from `RuleBasedNumberFormat::getCollator` to `RuleBasedNumberFormat::setLenient(true)`. The getCollator casts away const, and this initialization doesn’t seem thread safe.

Also the formatting code can use `adjustForCapitalizationContext`, which uses the same break iterator between threads, which is not thread safe. It should clone the break iterator to keep the state separate between threads.

#### Checklist
- [x] Required: Issue filed: ICU-23348


richgillam (approved):
LGTM

PR #3932, opened 2026-04-08
ICU-3736 UAX44-LM2 loose matching for character names

eggrobin:
#### Checklist
- [x] Required: Issue filed: ICU-3736

### DIFF
--- a/icu4c/source/common/unames.cpp
+++ b/icu4c/source/common/unames.cpp
@@ -85,6 +85,195 @@ typedef struct {
     uint32_t tokenStringOffset, groupsOffset, groupStringOffset, algNamesOffset;
 } UCharNames;
 
+namespace {
+// For a character name `name`, a query string `query` matches `name` under UAX44-LM2 if and only if
+// the following returns true:
+//     auto matcher = CharacterNameQuery(query).matcher();
+//     for (const char c : name) {
+//         if (!matcher.consistentWith(c)) {
+//             return false;
+//         }
+//     }
+//     return matcher.matches();
+// `name` must be an exact character name, in particular, all uppercase, with no underscores, no
+// double hyphens, no isolated hyphens.  There are no constraints on `query`; for instance, this
+// class will match name="ZANABAZAR SQUARE LETTER -A" with query="Zanabazar-square_letter_-A".
+// `Matcher::consistentWith` is not retryable: once it returns false, the `Matcher` should not be
+// used again.
+class CharacterNameQuery {
+  public:
+    class Matcher {
+      public:
+        // `query` must outlive the constructed object.
+        Matcher(const CharacterNameQuery& query)
+            : query(query), skeletonIterator(query.skeleton.begin()) {}
+
+        bool consistentWith(char c) {
+            // Instead of constructing a skeleton from the name as in the constructor of
+            // CharacterNameQuery, we check character-by-character if the skeleton we would
+            // construct from the name would be consistent with `query->skeleton`.
+            // We require that this be called with characters from an actual character name, so we
+            // need not worry about case or underscores here.
+
+            if (skeletonIterator == query.skeleton.end()) {
+                return false;
+            }
+            if (c == ' ') {
+                // The last hyphen was word-final; check that there is a corresponding significant
+                // hyphen in the skeleton.  A hyphen in a character name cannot be both word-final
+                // and word-initial by rule R3 in
+                // https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-4/#G135165, so we do
+                // not have to worry about checking the skeleton for the same significant hyphen
+                // twice.
+                if (lastChar == '-') {
+                    if (*skeletonIterator++ != lastChar) {
+                        return false;
+                    }
+                }
+                lastChar = c;
+                return true;
+            } else if (c == '-') {
+                if (lastChar == ' ' ||
+                    (query.is1180 && skeletonIterator == query.skeleton.end() - 2)) {
+                    // If lastChar == ' ', this is a word-initial hyphen, so we know it is
+                    // significant.  Check that we are expecting it.
+                    // If we are looking for U+1180 and we matched everything but the trailing -E,
+                    // this could be that hyphen; move past it.  This could turn out to be a
+                    // different character name if there is something other than E afterwards, in
+                    // which case we should in principle have ignored the hyphen; but then since the
+                    // suffixes will differ, we will return false anyway.
+                    // For example, when searching for HANGUL JUNGSEONG O-E and checking consistency
+                    // with HANGUL JUNGSEONG O-U, if we computed both skeleta and compared them,
+                    // the skeleta would be HANGULJUNGSEONGO-E and HANGULJUNGSEONGOU, and the
+                    // comparison would fail on - vs. U.
+                    // Here, while feeding HANGUL JUNGSEONG O-U character-by-character, we assume
+                    // its skeleton would be HANGULJUNGSEONGO-, and fail on E vs. U.
+                    if (*skeletonIterator++ != c) {
+                        return false;
+                    }
+                }
+                // If lastChar is not ' ', we do not know whether this hyphen is word-final, so we
+                // cannot check against the skeleton.
+                lastChar = c;
+                return true;
+            } else {
+                if (*skeletonIterator++ != c) {
+                    return false;
+                }
+                lastChar = c;
+                return true;
+            }
+        }
+
+        bool consistentWith(const std::string_view substring) {
+            for (const char c : substring) {
+                if (!consistentWith(c)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        bool matches() {
+            // If a character name could end with a HYPHEN-MINUS, that HYPHEN-MINUS would be
+            // significant: we would need to check that if lastChar == '-',
+            // *skeletonIterator == '-', and then advance skeletonIterator.
+            // However, this is disallowed by rule R3 in
+            // https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-4/#G135165.
+            U_ASSERT(lastChar != '-');
+            return skeletonIterator == query.skeleton.end();
+        }
+
+        std::string_view remainingSignificantCharacters() const {
+            return query.skeleton.substr(skeletonIterator - query.skeleton.begin());
+        }
+
+      private:
+        const CharacterNameQuery& query;
+        std::string_view::const_iterator skeletonIterator;
+        // Initialized to ' ' so that a leading hyphen is treated like a hyphen that follows a
+        // space (non-medial and thus not ignorable).  There are in fact no leading hyphens in
+        // character names by rule R3 in
+        // https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-4/#G135165, and technically
+        // we do not read the value of this variable before writing to it because the first call to
+        // consistentWith never has c=' ' either by R4, but it seems cleanest to initialize lastChar
+        // nonetheless.
+        char lastChar = ' ';
+    };
+
+    explicit CharacterNameQuery(std::string_view query) {
+        // Construct a skeleton obtained by
+        // 1. removing medial hyphens (except the one in the name of U+1180);
+        // 2. removing spaces and underscores;
+        // 3. uppercasing,
+        // as described in https://www.unicode.org/reports/tr44/#UAX44-LM2.
+        // We do all three in a single pass.
+        char *skeletonLimit = skeletonData;
+        for (std::size_t i = 0; i < query.length(); ++i) {
+            U_ASSERT(skeletonLimit < skeletonData + sizeof(skeletonData));
+            if (skeletonLimit >= skeletonData + sizeof(skeletonData)) {
+                // The caller should limit the query length appropriately; if they do not, assert,
+                // and if assertions are disabled, query for the empty string (which will quickly
+                // find nothing).
+                skeletonLimit = skeletonData;
+                return;
+            }
+            if (query[i] == ' ' || query[i] == '_') {
+                continue;
+            }
+            if (query[i] == '-') {
+                bool isMedial;
+                bool is1180MedialHyphen = false;
+                if (i == 0 || i == query.length()) {
+                    isMedial = false;
+                } else {
+                    isMedial = isASCIILetterOrDigit(query[i - 1]) && isASCIILetterOrDigit(query[i + 1]);
+                }
+                // A medial hyphen is the hyphen in the name of U+1180 HANGUL JUNGSEONG O-E if what
+                // comes before skeletonizes to HANGULJUNGSEONGO and what comes after skeletonizes
+                // to E.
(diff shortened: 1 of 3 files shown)


icu4c/source/common/unames.cpp
```diff
@@ -85,6 +85,131 @@ typedef struct {
...
+                if (lastChar == '-') {
+                    if (*skeletonIterator++ != lastChar) {
+                        return false;
+                    }
+                }
+                lastChar = c;
+                return true;
+            } else if (c == '-') {
+                if (lastChar == ' ' ||
+                    (query->is1180 && skeletonIterator == query->skeleton.end() - 2)) {
```
richgillam: Some internal documentation in this function would be helpful.  Especially here, where I have no idea what `query->is1180` is all about.

icu4c/source/common/unames.cpp
```diff
@@ -85,6 +85,131 @@ typedef struct {
...
+                bool isMedial;
+                bool is1180MedialHyphen = false;
+                if (i == 0 || i == query.length()) {
+                    isMedial = false;
+                } else {
+                    isMedial = isASCIILetterOrDigit(query[i - 1]) && isASCIILetterOrDigit(query[i + 1]);
+                }
+                if (isMedial && i <= query.length() - 2 && uprv_toupper(query[i + 1]) == 'E' &&
+                    std::string_view(skeletonData, skeletonLimit - skeletonData) == "HANGULJUNGSEONGO") {
+                    is1180MedialHyphen = true;
```
richgillam, line 237: Again, I think a comment explaining this part would be helpful...

icu4c/source/common/unames.cpp
```diff
@@ -293,14 +418,6 @@ expandName(UCharNames *names,
...
                     WRITE_CHAR(buffer, bufferLength, bufferPos, c);
                 } else {
-                    /* stop, but skip the semicolon if we are seeking
-                       extended names and there was no 2.0 name but there
-                       is a 1.0 name. */
-                    if(!bufferPos && nameChoice == U_EXTENDED_CHAR_NAME) {
-                        if (static_cast<uint8_t>(';') >= tokenCount || tokens[static_cast<uint8_t>(';')] == static_cast<uint16_t>(-1)) {
-                            continue;
-                        }
-                    }
```
richgillam, line 303: Why are you taking this out?

icu4c/source/common/unames.cpp
```diff
@@ -387,26 +504,18 @@ compareName(UCharNames *names,
...
                     }
                 } else {
-                    /* stop, but skip the semicolon if we are seeking
-                       extended names and there was no 2.0 name but there
-                       is a 1.0 name. */
-                    if(otherName == origOtherName && nameChoice == U_EXTENDED_CHAR_NAME) {
-                        if (static_cast<uint8_t>(';') >= tokenCount || tokens[static_cast<uint8_t>(';')] == static_cast<uint16_t>(-1)) {
-                            continue;
-                        }
-                    }
```
richgillam, line 401: Again, why is this coming out?  Is this code no longer relevant, or did it move somewhere else?

icu4c/source/test/intltest/usettest.cpp
```diff
@@ -4715,10 +4715,24 @@ void UnicodeSetTest::TestToPatternOutput() {
...
             {uR"([\N{PRESENTATION FORM FOR VERTICAL RIGHT WHITE LENTICULAR BRAKCET}])", u"[︘]"},
             {uR"([\N{bell}])", u"[🔔]"},
-            // Ill-formed in ICU 78 and earlier:
+            // Ill-formed in ICU 78 and earlier, made well-formed by ICU-23350:
             {uR"([\N{PRESENTATION FORM FOR VERTICAL RIGHT WHITE LENTICULAR BRACKET}])", u"[︘]"},
+            // Loose matching: These were ill-formed in ICU 78 and earlier, and were
+            // made well-formed by ICU-3736.
+            {uR"([\N{Latin small ligature o-e}])", u"[œ]"},
             {uR"([\N{Hangul jungseong O-E}])", u"[ᆀ]"},
+            {uR"([\N{Hangul jungseong O -E}])", u"[ᆀ]"},
```
richgillam, line 4724: Can there be whitespace on either side of the hyphen (or both), or only before it?

richgillam (commented):
Maybe it's just because I'm not awake yet (or maybe I'm not the best choice of reviewer), but I had trouble finding my way through this.  A lot of it made sense, but it wasn't clear to me what the actual rules for matching are, and there seemed to be several spots in the code that assumed more knowledge on the part of the reader than I possess.  Can you help me understand what's going on here?

icu4c/source/common/unames.cpp
eggrobin: Unicode 1 names were removed from ICU in ICU 49.

icu4c/source/common/unames.cpp
eggrobin: I will add more comments; in the meantime, see https://www.unicode.org/reports/tr44/#UAX44-LM2.

richgillam: Wow, that made me dizzy...

Thanks for pointing me to that document.  The rules there aren't at all intuitive unless you know the motivation, but I feel more confident that they are indeed what you implemented.

richgillam (dismissed):
LOKTM

icu4c/source/test/intltest/usettest.cpp
eggrobin: Whitespace on either side makes it non-medial.

eggrobin:
I added some comments; they are pretty long because while writing them I noticed some subtle edge cases (which, as best I can tell, were handled correctly, but are noteworthy)…

icu4c/source/common/unames.cpp
```diff
@@ -85,6 +85,177 @@ typedef struct {
...
+                    // there is another letter beyond the E.  The insignificant characters are thus
+                    // only spaces and underscores.
+                    for (std::size_t j = i + 2; j < query.length(); ++j) {
+                        if (query[j] != ' ' && query[j] != '_') {
+                            is1180MedialHyphen = false;
+                        }
+                    }
+                }
+                if (!isMedial || is1180MedialHyphen) {
+                    *skeletonLimit++ = query[i];
```
markusicu, line 250: I am nervous about overflowing the skeletonData array. I guess the only public entry point is u_charFromName(), but it seems easy for the "120" here and there to get out of sync. I would feel better if you checked before writing; probably best with a tiny helper function that also sets a UErrorCode when we overflow. Or maybe the error handling here could just consist of setting the skeleton to a one-garbage-byte string?

icu4c/source/common/unames.cpp
```diff
@@ -85,6 +85,177 @@ typedef struct {
...
+// `name` must be an exact character name, in particular, all uppercase, with no underscores, no
+// double hyphens, no isolated hyphens.  There are no constraints on `query`; for instance, this
+// class will match name="ZANABAZAR SQUARE LETTER -A" with query="Zanabazar-square_letter_-A".
+// `Matcher::consistentWith` is not retryable: once it returns false, the `Matcher` should not be
+// used again.
+class CharacterNameQuery {
+  public:
+    class Matcher {
+      public:
+        Matcher(const CharacterNameQuery *const query)
```
markusicu: why not take the query by reference? it must not be nullptr.

icu4c/source/common/unames.cpp
```diff
@@ -85,6 +85,177 @@ typedef struct {
...
+            // significant: we would need to check that if lastChar == '-',
+            // *skeletonIterator == '-', and then advance skeletonIterator.
+            // However, this is disallowed by rule R3 in
+            // https://unicode.org/versions/Unicode17.0.0/core-spec/chapter-4/#G135165.
+            U_ASSERT(lastChar != '-');
+            return skeletonIterator == query->skeleton.end();
+        }
+
+        std::string_view remainingSignificantCharacters() const {
+            return query->skeleton.substr(skeletonIterator - query->skeleton.begin());
```
markusicu: optional: In Google code, we have an effort of constructing string_view's from strings and from pairs of iterators, and maybe then calling substr(), rather than constructing straight from pointer+length. ("safe buffers" project)

Here you could return `std::string_view(skeletonIterator, query->skeleton.end())`.

icu4c/source/common/unames.cpp
```diff
@@ -1088,20 +1244,16 @@ findAlgName(AlgorithmicRange *range, UCharNameChoice nameChoice, const char *oth
...
 
-        uint16_t i, count;
-
         /* compare prefix */
         while((c=*s++)!=0) {
-            if (c != *otherName++) {
+            if (!matcher.consistentWith(c)) {
                 return 0xffff;
             }
         }
```
markusicu: optional: we have two of these loops; maybe move them into a matcher.consistentWith(string_view prefix)?

markusicu (commented):
As usual, nice and elegant code where I feel like I learn good and modern techniques. I do have a few comments though.

icu4c/source/common/unames.cpp
eggrobin: > constructing string_view's […] and maybe then calling substr()

But that is what we have here? `query->skeleton` is a string_view, no pointers are harmed on this line.

markusicu: sorry, brain fart -- i must have seen the remaining length and nothing else...

icu4c/source/common/unames.cpp
eggrobin: Three even. Done.

eggrobin: `substring` rather than `prefix` though (it gets used for tokens in normal name matching).

icu4c/source/common/unames.cpp
eggrobin: Google style uses pointers when there is a lifetime requirement, as here, and I like that. Added a comment though, lifetime requirements should be documented…

markusicu: AFAICT Google code has been moving away from pointers to references when not nullable, including for outputs.

markusicu: Regardless of Google... I much prefer using a reference when the thing can't be null.

icu4c/source/common/unames.cpp
eggrobin: If it is checking an ICU-internal invariant (that the two 120s are equal), propagating a UErrorCode seems inappropriate. Added a U_ASSERT plus a fallback to an empty query if assertions are disabled.

icu4c/source/common/unames.cpp
eggrobin: > AFAICT Google code has been moving away from pointers to references when not nullable, including for outputs.

For outputs, yes; but not for lifetime dependencies if I recall correctly.

eggrobin: https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs with https://abseil.io/tips/116.

markusicu: I don't see how a pointer helps for that; it's certainly not something we do in ICU.
Please change query to a reference.

eggrobin: They do advise to make the _member_ a reference. Done.

markusicu: I guess I see the point:
> In some cases reference parameters can bind to temporaries, leading to lifetime bugs.

One of many C++ foot guns... I guess I can live with the pointer :-(

eggrobin: > I don't see how a pointer helps for that; 

Well, the ToTW has some pretty clear examples; it prevents passing temporaries.

But a cursory search for `outlive` finds at least one comment about a lifetime requirements on reference parameters in ICU, so I guess we do this here. Done.

icu4c/source/common/unames.cpp
```diff
@@ -559,11 +577,8 @@ compareName(UCharNames *names,
                 }
             } else {
                 /* write token word */
-                uint8_t *tokenString=tokenStrings+token;
-                while((c=*tokenString++)!=0) {
-                    if (!matcher.consistentWith(c)) {
-                        return false;
-                    }
+                if (!matcher.consistentWith(reinterpret_cast<const char *>(tokenStrings + token))) {
```
markusicu, line 580: Thanks -- but I just realized that we are now adding a strlen() call...
We could have the function take a `char *` pointer, or we could just live with it.

icu4c/source/common/unames.cpp
eggrobin: But we certainly do that in some places in ICU: see, e.g.,
https://github.com/unicode-org/icu/blob/c5946d703999261571f7ff6da4c55454beca17c1/icu4c/source/i18n/units_data.cpp#L51-L54

icu4c/source/common/unames.cpp
```diff
@@ -200,6 +210,14 @@ class CharacterNameQuery {
         // We do all three in a single pass.
         char *skeletonLimit = skeletonData;
         for (std::size_t i = 0; i < query.length(); ++i) {
+            U_ASSERT(skeletonLimit < skeletonData + sizeof(skeletonData));
+            if (skeletonLimit >= skeletonData + sizeof(skeletonData)) {
+              // The caller should limit the query length appropriately; if they do not, assert, and
+              // if assertions are disabled, query for the empty string (which will quickly find
+              // nothing).
+              skeletonLimit = skeletonData;
+              return;
```
markusicu: indent+2 please

icu4c/source/common/unames.cpp
eggrobin: Let’s live with it, if we care about the performance of lookups by name I think we have bigger fish to fry than strlen on small strings.

icu4c/source/common/unames.cpp
eggrobin: Oops. Done.

icu4c/source/common/unames.cpp
```diff
@@ -104,8 +104,9 @@ class CharacterNameQuery {
   public:
     class Matcher {
       public:
-        Matcher(const CharacterNameQuery *const query)
-            : query(query), skeletonIterator(query->skeleton.begin()) {}
+        // `query` must outlive the constructed object.
+        Matcher(const CharacterNameQuery& query)
+            : query(*query), skeletonIterator(query.skeleton.begin()) {}
```
markusicu: I think the `*` does not want to be there?

icu4c/source/common/unames.cpp
```diff
@@ -242,7 +260,7 @@ class CharacterNameQuery {
     }
 
     Matcher matcher() const {
-      return Matcher(this);
+      return Matcher(*this);
```
markusicu: indent+2

markusicu (approved):

eggrobin:
Hmm, looks like my post-review changes introduced a bug:
UnicodeSet(R"([{\N{Hangul syllable YA}\N{Hangul syllable ONG}}])").toPattern() expected [{야옹}], got [{쟥좢}]

etc.

richgillam (approved):
Still LOKTM.

PR #3938, opened 2026-04-13
ICU-23371 Reduce the memory footprint subdivisionData in ValidIdentifiers

gvictor:
HashSet internally uses HashMap requests extra memory footprints for each entry. Use Set.copyOf(Set) to avoid the memory overhead The heap dump shows ~351kB reduction.

Also, use Map.copyOf(map) instead of HashMap. It reduces the memory footprint by ~52kB.

#### Checklist
- [x] Required: Issue filed: ICU-23371

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
@@ -11,7 +11,6 @@ package com.ibm.icu.impl;
 import com.ibm.icu.impl.locale.AsciiUtil;
 import com.ibm.icu.util.UResourceBundle;
 import com.ibm.icu.util.UResourceBundleIterator;
-import java.util.Collections;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -74,18 +73,12 @@ public class ValidIdentifiers {
                 HashMap<String, Set<String>> _subdivisionData2 = new HashMap<String, Set<String>>();
                 // protect the sets
                 for (Entry<String, Set<String>> e : _subdivisionData.entrySet()) {
-                    Set<String> value = e.getValue();
-                    // optimize a bit by using singleton
-                    Set<String> set =
-                            value.size() == 1
-                                    ? Collections.singleton(value.iterator().next())
-                                    : Collections.unmodifiableSet(value);
-                    _subdivisionData2.put(e.getKey(), set);
+                    _subdivisionData2.put(e.getKey(), toImmutableSet(e.getValue()));
                 }
 
-                this.subdivisionData = Collections.unmodifiableMap(_subdivisionData2);
+                this.subdivisionData = toImmutableMap(_subdivisionData2);
             } else {
-                this.regularData = Collections.unmodifiableSet(plainData);
+                this.regularData = toImmutableSet(plainData);
                 this.subdivisionData = null;
             }
         }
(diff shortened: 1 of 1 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
```diff
@@ -49,9 +51,29 @@ public enum Datasubtype {
...
+        private static final Method SET_COPY_OF_METHOD;
+        private static final Method MAP_COPY_OF_METHOD;
+
         public final Set<String> regularData;
         public final Map<String, Set<String>> subdivisionData;
 
+        static {
+            Method setCopyOfMethod;
+            try {
+                setCopyOfMethod = Set.class.getMethod("copyOf", Collection.class);
```
mihnita: Both [`Set.copyOf`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Set.html#copyOf(java.util.Collection)) and [`Map.copyOf`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#copyOf(java.util.Map)) are public APIs in JDK 11.

And ICU dropped support for JDK 1.8 since last year (2025-03-31, [ICU-23072](https://unicode-org.atlassian.net/browse/ICU-23072), https://github.com/unicode-org/icu/commit/7eb090f910)

So there is no need for reflection.

markusicu:
@jabagawee FYI

icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
gvictor: Done.

I promise that I did try to Google it, and looked at the documentation here. But none of them showed that the min requirement is JDK 11.

https://unicode-org.github.io/icu/userguide/icu4j/faq.html#what-version-of-java-is-required-for-icu4j

icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
```diff
@@ -75,17 +74,13 @@ public ValiditySet(Set<String> plainData, boolean makeMap) {
                 // protect the sets
                 for (Entry<String, Set<String>> e : _subdivisionData.entrySet()) {
                     Set<String> value = e.getValue();
-                    // optimize a bit by using singleton
-                    Set<String> set =
-                            value.size() == 1
-                                    ? Collections.singleton(value.iterator().next())
-                                    : Collections.unmodifiableSet(value);
+                    Set<String> set = toImmutableSet(value);
                     _subdivisionData2.put(e.getKey(), set);
```
mihnita: Nitpick:

```java
Set<String> value = e.getValue();
Set<String> set = toImmutableSet(value);
_subdivisionData2.put(e.getKey(), set);
```
No need for all the intermediate variables:
```java
_subdivisionData2.put(e.getKey(), toImmutableSet(e.getValue()));
```

---

Nitpick means "only if you like that suggestion, if not, feel free to ignore without explaining why"
It will not affect my approval.

mihnita (dismissed):
Thank you!
M

icu4j/main/core/src/main/java/com/ibm/icu/impl/ValidIdentifiers.java
gvictor: Done.

mihnita (approved):
Thank you!
M

markusicu (approved):

pachecocordovamoiseseduardo-byte (approved):

PR #3939, opened 2026-04-13
ICU-23371 Reduce the memory footprints of TransliteratorRegistry

gvictor:
1. Intern CaseSensitiveString objects for source, target, variant in the registry
2. Use a default capacity of 1 for variant list. The source-target key  is mostly linked to 1 variant only.
3. Use ConcurrentHashMap instead of Collections.synchronizedMap(new HashMap())
4. Lazily initialize the script code cache in AnyTransliterator.

In total, it saves ~110 kB.

#### Checklist
- [x] Required: Issue filed: ICU-23371

### DIFF
--- a/icu4j/main/translit/src/main/java/com/ibm/icu/text/AnyTransliterator.java
+++ b/icu4j/main/translit/src/main/java/com/ibm/icu/text/AnyTransliterator.java
@@ -49,7 +49,7 @@ class AnyTransliterator extends Transliterator {
     static final String LATIN_PIVOT = "-Latin;Latin-";
 
     /** Cache mapping UScriptCode values to Transliterator*. */
-    private ConcurrentHashMap<Integer, Transliterator> cache;
+    private volatile ConcurrentHashMap<Integer, Transliterator> cache;
 
     /** The target or target/variant string. */
     private String target;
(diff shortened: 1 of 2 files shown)


mihnita (dismissed):
LGTM, thank you!
M

jabagawee:
does the lazy initialization reduce memory footprint? wouldn't it still take up the same amount of memory once it's initialized? also this lazy initialization introduces a new synchronization point when getting the cache instance. can we apply one of the concurrency patterns that would not cost us extra synchronization? i just made a change recently that ripped out a bunch of contention.

gvictor:
> does the lazy initialization reduce memory footprint? wouldn't it still take up the same amount of memory once it's initialized? 

It does reduce the memory footprint, because the cache in AnyTransliterator is always empty before it's being used. For Android specifically, when it's used in the post-forked app process, the page immediately becomes dirty even if the map is initialized in the pre-fork process. Initializing it before forking uses the extra memory.

> can we apply one of the concurrency patterns that would not cost us extra synchronization? i just made a change recently that ripped out a bunch of contention.

I will use a double check locking, and should avoid contention after the first read.

mihnita (approved):

markusicu (approved):

pachecocordovamoiseseduardo-byte (approved):

PR #3944, opened 2026-04-16
ICU-23365 move zero-init into declaration & add unit test

markusicu:
Improved version of
- https://github.com/unicode-org/icu/pull/3936
- from @OwenSanzas

#### Checklist
- [x] Required: Issue filed: ICU-23365

### DIFF
--- a/icu4c/source/test/intltest/trnserr.cpp
+++ b/icu4c/source/test/intltest/trnserr.cpp
@@ -27,6 +27,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include "cmemory.h"
 #include "unicode/rep.h"
 #include "unicode/locid.h"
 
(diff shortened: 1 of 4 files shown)


icu4c/source/test/intltest/trnserr.cpp
```diff
@@ -261,6 +265,23 @@ void TransliteratorErrorTest::TestRBTErrors() {
...
+    UParseError pe;
+    LocalPointer<Transliterator> trans(
+        Transliterator::createFromRules(u"id", rules, UTRANS_REVERSE, pe, errorCode));
+    // We expect this to fail but not crash.
+    if (errorCode.expectErrorAndReset(U_MEMORY_ALLOCATION_ERROR)) {
+        return;
+    }
+    // Just in case it does not fail.
+    UnicodeString text(u"hello");
+    trans->transliterate(text);
```
richgillam: What are these lines getting us?  Is the idea here that the `transliterate()` call might crash, or cause the destructor to crash?

richgillam (dismissed):
LOKTM

icu4c/source/test/intltest/trnserr.cpp
markusicu: I copied them from the previous PR's description, but of course we won't get there as long as createFromRules() fails for garbage input. So these are currently dead code. If for some reason the factory function creates some kind of Transliterator, then we would get here.

Do you want me to remove these lines?

richgillam: If the factory function creates some kind of Transliterator and we get to those two lines, are they what we want to do?  I'm still not sure these lines actually do.

I don't have strong feelings here, but if you also don't know what they do, maybe we're better off without them.

markusicu: Well, on the off-chance that we get a Transliterator object, we would be using it in the normal way. I could assert that the text is unchanged. However, that goes beyond reproducing the bug, so I removed these lines now. PTAL

richgillam: Yeah, that looks fine to me.

richgillam (approved):
Still LGTM.

pachecocordovamoiseseduardo-byte (approved):

PR #3948, opened 2026-04-17
ICU-23376 Minimize the number of states of the DFA

eggrobin:
This halves the size of the `line*.brk` (from 73 424 bytes to 37 224 for `line.brk`, similarly for the others); it does nothing for `char.brk` or `sent.brk`, and not much for `word.brk` (23 120 to 22 672).

The C++ implementation is atrocious because of our lack of decent data structures, this was 30 lines of Python and would probably not be very much more in modern C++…

CC @aheninger, @robertbastian.

#### Checklist
- [x] Required: Issue filed: ICU-23376

### DIFF
--- a/icu4c/source/common/rbbitblb.cpp
+++ b/icu4c/source/common/rbbitblb.cpp
@@ -561,7 +561,7 @@ void RBBITableBuilder::bofFixup() {
 //                        transition tables for these states, by the algorithm
 //                        of fig. 3.44 in Aho.
 //
-//                        Most of the comments are quotes of Aho's psuedo-code.
+//                        Most of the comments are quotes of Aho's pseudo-code.
 //
 //-----------------------------------------------------------------------------
 void RBBITableBuilder::buildStateTable() {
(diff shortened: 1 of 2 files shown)


markusicu:
Going down to 199 states should let us use an 8-bit-value trie and 8-bit-value state table, right?
Does the builder select those automatically if possible?

markusicu:
Java: future commit here or future PR?

eggrobin:
> Going down to 199 states should let us use an 8-bit-value trie and 8-bit-value state table, right?
Does the builder select those automatically if possible?

Looks like it:

https://github.com/unicode-org/icu/blob/c5946d703999261571f7ff6da4c55454beca17c1/icu4c/source/common/rbbitblb.cpp#L1362-L1364

markusicu:
> This halves the size of the `line*.brk` (from 73 424 bytes to 37 224 for `line.brk`

🤯

markusicu:
> > Going down to 199 states should let us use an 8-bit-value trie and 8-bit-value state table, right?
> > Does the builder select those automatically if possible?
> 
> Looks like it:

Yeah!

We might have a new problem then: Lack of code coverage for the 16-bit paths. Unless we have a unit test with strangely convoluted rules.

eggrobin:
> Java: future commit here or future PR?

I think either works, so depends on whether you review faster than I write the Java (you have an edge, I have half of Monday off). We probably can’t regenerate the checked-in .brks until I have done that because of round-trip tests.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+        uint32_t fAccepting;
+        uint32_t fLookAhead;
+        int32_t fTagsIdx;
+        bool operator==(StateType const &other) const {
+            return fAccepting == other.fAccepting &&
+                   fLookAhead == other.fLookAhead &&
+                   fTagsIdx == other.fTagsIdx;
+        }
+    };
+    // Group the states by types (but we have no maps so this is verbose).
```
markusicu, line 1036: FYI: We do have a very poor man's icu4c/source/common/uhash.h ...

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+    // of Algorithm 3.6.
+    struct TypeToStates {
+        TypeToStates(const StateType &type, UErrorCode &status)
+            : type(type), states(new UVector(status)) {
+            if (U_SUCCESS(status) && states == nullptr) {
+                status = U_MEMORY_ALLOCATION_ERROR;
+            }
+        }
+        StateType type;
+        std::unique_ptr<UVector> states;
```
markusicu: Why a unique_ptr of a UVector? Why not a UVector by value?

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+              return;
+            }
+            newEntry->states->addElement(i, *fStatus);
+            initialPartition.adoptElement(newEntry, *fStatus);
+        }
+    }
+    // The partition Π from Algorithm 3.6.
+    // (We could call it Π, but some member companies that integrate the ICU
+    // code base prohibit non-ASCII identifiers…).
+    std::unique_ptr<UVector> partition(new UVector(*fStatus));
```
markusicu: again, why not UVector by value?

icu4c/source/common/rbbitblb.cpp
markusicu: Also, inside ICU we usually use our own LocalPointer rather than unique_ptr.
I think this could save you a non-trivial amount of clutter below -- see one of my comments further down.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+    }
+    if (partition == nullptr) {
+      *fStatus = U_MEMORY_ALLOCATION_ERROR;
+      return;
+    }
+    partition->setDeleter(
+        [](void *p) { delete static_cast<UVector*>(p); });
+    
+    for (int32_t i = 0; i < initialPartition.size(); ++i) {
+        auto* const states = new UVector(*fStatus);
```
markusicu: I don't see where this pointer is used...?

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+    // transition from 𝑠 on the character class with index i) is in Π[σ(𝑠)[i]].
+    // We then have σ(𝑠) == σ(𝑡) if and only if “for all input symbols 𝑎, states
+    // 𝑠 and 𝑡 have transitions on 𝑎 to states in the same group of Π” (which is
+    // what defines the refinement of 𝐺 in Fig. 3.45).
+    auto partition_signature =
+        [&partition, this](int32_t stateIndex) -> std::unique_ptr<UVector> {
+            const RBBIStateDescriptor &state =
+                *static_cast<RBBIStateDescriptor*>(
+                    fDStates->elementAt(stateIndex));
+            std::unique_ptr<UVector> result(
```
markusicu: does UVector have move-copy? if so, we could create & return by value.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1001,275 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+                new UVector(state.fDtran->size(), *fStatus));
+            if (U_FAILURE(*fStatus)) {
+                return nullptr;
+            }
+            if (result == nullptr) {
+                *fStatus = U_MEMORY_ALLOCATION_ERROR;
+                return nullptr;
+            }
```
markusicu: These patterns should be spiffier with LocalPointer, something like
```
            LocalPointer<...> result(...create(..., *fStatus), *fStatus);
            if (U_FAILURE(*fStatus)) {
                return nullptr;
            }
```
Note the second UErrorCode, for the LocalPointer constructor. If there is no error but the pointer is nullptr, then it sets the mem-alloc-error. Less boilerplate code.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1317,15 +1586,9 @@ void RBBITableBuilder::removeSafeState(IntPair duplStates) {
  * RemoveDuplicateStates
  */
 int32_t RBBITableBuilder::removeDuplicateStates() {
-    IntPair dupls = {3, 0};
-    int32_t numStatesRemoved = 0;
-
-    while (findDuplicateState(&dupls)) {
```
markusicu, line 1323: Do we still need the findDuplicateState() and removeState() functions?

markusicu (commented):
I am not trying to understand the logic.

markusicu:
PS: Where I ask a question for why you choose something, and there is a good reason to do it that way, consider maybe adding a code comment.

icu4c/source/common/rbbitblb.cpp
eggrobin: We move them around, and you cannot move a UVector.

icu4c/source/common/rbbitblb.cpp
eggrobin: $\mathrm{Π} ≔ \mathrm{Π}_{\text{new}}$.

icu4c/source/common/rbbitblb.cpp
eggrobin: I tried with LocalPointer first, it was rather worse (and I thought we were moving towards unique_ptr).

icu4c/source/common/rbbitblb.cpp
eggrobin: It has neither.

icu4c/source/common/rbbitblb.cpp
eggrobin: Ah, I did not know about that. But on the other hand it is not implicitly constructible from nullptr_t so that made some other things very ugly.

I’d be tempted to fix this category of issue by having an ICU version of make_unique (with a status) rather than trying to keep LocalPointer up to date with the whole of unique_ptr.

icu4c/source/common/rbbitblb.cpp
eggrobin: Yeah, but linear search is fine, it is more the API of a map than the performance that I need here. I don’t think the boilerplate for a C hash table would help (and then I would need to be careful about how I hash, this needs to round-trip).

icu4c/source/common/rbbitblb.cpp
eggrobin: Probably not.

icu4c/source/common/rbbitblb.cpp
markusicu: Even if we at some point write a replacement for UVector, at some earlier point (but not here) we should fix simple things like this (and element iteration etc.) in UVector itself and improve such call sites.

markusicu: Why was LocalPointer worse?

I don't think we are moving away from it because it provides UErrorCode goodies like I mentioned.

icu4c/source/common/rbbitblb.cpp
markusicu: > Ah, I did not know about that. But on the other hand it is not implicitly constructible from nullptr_t so that made some other things very ugly.

- Please create a ticket for constructing a LocalPointer from a unique_ptr.
- Please consider changing all of these to LocalPointer so that we don't get into conversion trouble in the meantime.

> I’d be tempted to fix this category of issue by having an ICU version of make_unique (with a status) rather than trying to keep LocalPointer up to date with the whole of unique_ptr.

+1 (later)

icu4c/source/common/rbbitblb.cpp
eggrobin: > I don't think we are moving away from it because it provides UErrorCode goodies like I mentioned.

Right, now that I look more closely, there are many usages, but those are in tests.

icu4c/source/common/rbbitblb.cpp
eggrobin: I will keep the one unique_ptr where I use it to wrap a `uint16_t[]` though, since there is no double error handling here.

eggrobin: > Please create a ticket for constructing a LocalPointer from a unique_ptr.

ICU-23377.

> +1 (later)

Tickets are cheap: ICU-23378.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1001,6 +1003,252 @@ void  RBBITableBuilder::mergeRuleStatusVals() {
...
+        }
+        if (j == initialPartition.size()) {
+            const auto newEntry = new TypeToStates(type, *fStatus);
+            if (U_FAILURE(*fStatus)) {
+                return;
+            }
+            if (newEntry == nullptr) {
+              *fStatus = U_MEMORY_ALLOCATION_ERROR;
+              return;
+            }
```
markusicu: FYI: You could wrap the newEntry into a LocalPointer just for its mem-alloc-error handling, and immediately orphan() it into the adoptElement() ...

markusicu (dismissed):

icu4c/source/common/rbbitblb.cpp
eggrobin: Done, also below.

markusicu (dismissed):

markusicu:
Looking at your local time, I think you are practicing for the UTC meeting...

markusicu:
Sadly, CI checks are failing. I should bike home, and you should catch some sleep...

eggrobin:
> Looking at your local time, I think you are practicing for the UTC meeting...

Yeah, I am not sure what is going on here. This morning I woke up at seven in the morning. 24 h prior I was failing to sleep at seven in the morning.

eggrobin:
> Sadly, CI checks are failing

Yeah I think the unique_ptr is dragging in some silly symbols. Once we have a make_unique with a status we should not let that stop us, but for now I guess I will replace the remaining one with a LocalArray.

markusicu (approved):

eggrobin:
In any case, this needs TC approval, so it’s not getting merged tonight.

eggrobin:
It looks like the fuzzer failure is just telling us that this minimization algorithm can be a bit slow on pathological state machines.
On the rules `.+䶝.䶝.............;`, it starts out with 7398 states and does nothing, which means that it needs to refine the partition from two parts to 7398. On my machine this takes 10 s in debug mode and 2 s in release mode. It does not seem implausible that it would exceed 25 s with ubsan on the CI runners.

eggrobin:
@markusicu It looks like I successfully appeased the fuzzers. The windows-msys2-gcc-x86_64 failure seems unrelated, I see it on other PRs too.

icu4c/source/common/rbbitblb.cpp
```diff
@@ -1206,8 +1216,11 @@ void RBBITableBuilder::minimizeStates() {
     if (U_FAILURE(*fStatus)) {
         return;
     }
-    LocalArray<uint16_t> oldStateToPart(new uint16_t[fDStates->size()](), *fStatus);
-    if (U_FAILURE(*fStatus)) {
+    // We cannot use LocalArray nor new[] because, on uint16_t, they would call
+    // the global new[] and delete[].
```
markusicu, line 1220: Right, but we can use this internal class:
https://github.com/unicode-org/icu/blob/a011a81db68f099f4b2f9e403b350e3ffa166dfa/icu4c/source/common/cmemory.h#L179-L187

markusicu:
> The windows-msys2-gcc-x86_64 failure seems unrelated, I see it on other PRs too.

Yeah, it fails on main too: https://github.com/unicode-org/icu/actions/runs/25382815834
This is very recent :-(

icu4c/source/common/rbbitblb.cpp
eggrobin: We sure have a lot of those bootleg `unique_ptr`s…

markusicu:
> > The windows-msys2-gcc-x86_64 failure seems unrelated, I see it on other PRs too.
> 
> Yeah, it fails on main too: https://github.com/unicode-org/icu/actions/runs/25382815834 This is very recent :-(

Mihai has unblocked CI via PR #3970. Please rerun the failing test. If it still fails, try rebasing.

markusicu (approved):

pachecocordovamoiseseduardo-byte (approved):

PR #3950, opened 2026-04-21
ICU-23381 Fix null pointer dereference and memory leak in MessageFormat 2.0

nciric:
_Summary_
This PR addresses two potential safety vulnerabilities identified in the MessageFormat 2.0 evaluation pipeline (`messageformat2.cpp`), specifically a null pointer dereference and a memory leak triggered during out-of-memory (OOM) conditions.

_Details_

_1. Null Pointer Dereference in `formatExpression`_
In `MessageFormatter::formatExpression`, the method `formatOperand` is called to evaluate operands. If `formatOperand` fails due to an OOM condition (when allocating an `InternalValue`), it returns `nullptr` and sets the `UErrorCode` to a failure state.

However, the code was missing a check for this failure case immediately after the call. Furthermore, if the expression was a function call or fell through to the `else` block, the `status` was being reset to `U_ZERO_ERROR` without handling the previous error. This allowed execution to continue with a `nullptr` `randVal`, leading to undefined behavior (segmentation fault) when attempting to access it (e.g., in `takeArgument` or `evalFunctionCall`).

_Fix_: Added an early return if `U_FAILURE(status)` is true after calling `formatOperand`, preventing operations on a null pointer.

_2. Memory Leak in `checkDeclarations`_
In `MessageFormatter::checkDeclarations`, the code iterates over local variable bindings to build a linked environment chain. 
env = Environment::create(..., env, status);If `Environment::create` failed due to OOM, it returned `nullptr`. Because this `nullptr` was assigned directly to the `env` pointer, the reference to the previously allocated environment nodes in the chain was lost, resulting in a memory leak.

_Fix_: Result of `Environment::create` is now stored in a temporary pointer, and `env` is only updated if the operation was successful. This allows the caller's `LocalPointer` to correctly clean up the chain constructed so far.

_Testing_
- Built successfully.
- Ran the full ICU4C test suite (`make check`) and all tests passed, confirming no regressions in standard behavior.

#### Checklist
- [x] Required: Issue filed: ICU-23381

### DIFF
--- a/icu4c/source/i18n/messageformat2.cpp
+++ b/icu4c/source/i18n/messageformat2.cpp
@@ -277,6 +277,9 @@ FunctionOptions MessageFormatter::resolveOptions(const Environment& env, const O
     const Operand& rand = expr.getOperand();
     // Format the operand (formatOperand handles the case of a null operand)
     LocalPointer<InternalValue> randVal(formatOperand(fallback, globalEnv, rand, context, status));
+    if (U_FAILURE(status)) {
+        return {};
+    }
 
     FormattedPlaceholder maybeRand = randVal->takeArgument(status);
 
--- a/icu4c/source/i18n/messageformat2_evaluation.cpp
+++ b/icu4c/source/i18n/messageformat2_evaluation.cpp
@@ -170,11 +170,13 @@ PrioritizedVariant::~PrioritizedVariant() {}
     // ---------------- Environments and closures
 
     Environment* Environment::create(const VariableName& var, Closure&& c, Environment* parent, UErrorCode& errorCode) {
-        NULL_ON_ERROR(errorCode);
+        if (U_FAILURE(errorCode)) {
+            return parent;
+        }
         Environment* result = new NonEmptyEnvironment(var, std::move(c), parent);
         if (result == nullptr) {
             errorCode = U_MEMORY_ALLOCATION_ERROR;
-            return nullptr;
+            return parent;
         }
         return result;
     }


mihnita (dismissed):

catamorphism (dismissed):
Thanks!

icu4c/source/i18n/messageformat2.cpp
```diff
@@ -775,10 +778,13 @@ void MessageFormatter::checkDeclarations(MessageContext& context, Environment*&
...
 
         // Add the LHS to the environment for checking the next declaration
-        env = Environment::create(StandardFunctions::normalizeNFC(decl.getVariable()),
+        Environment* newEnv = Environment::create(StandardFunctions::normalizeNFC(decl.getVariable()),
                                   Closure(rhs, *env),
                                   env,
                                   status);
+        if (U_SUCCESS(status)) {
+            env = newEnv;
+        }
```
markusicu: This looks fragile. I would prefer changing Environment::create() so that if something goes wrong it returns the original `env` value.

nciric:
First, I am sorry for force-push on the PR - I asked AI agent to consider comment from Markus, and it made a fix and force-pushed instead of just adding another commit.

@markusicu is this what you had in mind?

markusicu:
> First, I am sorry for force-push on the PR - I asked AI agent to consider comment from Markus, and it made a fix and force-pushed instead of just adding another commit.

Live and learn... can you instruct the LLM to add commits rather than amending & force-pushing?

> @markusicu is this what you had in mind?

Yes, thanks. I do have a follow-up question for C++ gurus.

icu4c/source/i18n/messageformat2_evaluation.cpp
```diff
@@ -170,11 +170,13 @@ PrioritizedVariant::~PrioritizedVariant() {}
     // ---------------- Environments and closures
 
     Environment* Environment::create(const VariableName& var, Closure&& c, Environment* parent, UErrorCode& errorCode) {
-        NULL_ON_ERROR(errorCode);
+        if (U_FAILURE(errorCode)) {
+            return parent;
+        }
         Environment* result = new NonEmptyEnvironment(var, std::move(c), parent);
```
markusicu, line 176: @eggrobin @roubert @aheninger if the memory allocation fails, we don't call the constructor -- but will the `std::move(c)` expression have been evaluated already?

If so, then we should pass in `c` by copy.

eggrobin: I don’t think that’s a problem given that we take c by rvalue reference, which means the caller will have to move it in already. So even if this fails, the caller shouldn’t use the thing it passed as c.

markusicu (approved):

nciric:
> > First, I am sorry for force-push on the PR - I asked AI agent to consider comment from Markus, and it made a fix and force-pushed instead of just adding another commit.
> 
> Live and learn... can you instruct the LLM to add commits rather than amending & force-pushing?

I did, plus added that no code should be committed locally or remotely without my review.

I can't land this PR, so you'll have to do it.

pachecocordovamoiseseduardo-byte (approved):

PR #3951, opened 2026-04-21
ICU-23247 Fix raw types, other issues reported by -Xlint:unchecked

mihnita:
#### Checklist


icu4j/demos/src/main/java/com/ibm/icu/dev/demo/Launcher.java
```diff
@@ -142,9 +142,9 @@ public void actionPerformed(ActionEvent e) {
                     String demo = demoBase + '.' + demoShort;
                     showStatus(demoShort, "launching");
                     try {
-                        Class c = Class.forName(demo);
+                        Class<?> c = Class.forName(demo);
```
echeran, line 145: thought: I wonder if the `Class` type in Java is unique among the parameterized types in that the type parameter is largely not necessary. I imagine that you would use it for reflection purposes, and that the `Class` in particular is something whose type you won't know ahead of time until you parse & convert the stringified name into the object.

echeran (approved):
LGTM. I trust that when the `ArrayList`s are saved to a local variable of type `ArrayList` rather than `List`, those are during places where random access getting/setting is occurring. Which is reasonable.

icu4j/demos/src/main/java/com/ibm/icu/dev/demo/Launcher.java
mihnita: The type still helps. Probably less than for other classes, but still.

For example:
```java
Class<String> cs = String.class;
Class<?> ci = Integer.class;

String cos = cs.cast(o);
Object co = ci.cast(o);
Integer coi = ci.cast(o); // Error

Constructor<?> constro = ci.getConstructor();
Constructor<String> constrs = cs.getConstructor();

Object insto = constro.newInstance();
Integer insto = constro.newInstance(); // Error
String inst = constrs.newInstance();
```

What makes is less useful is that you often use reflection to get access to classes that are not accessible at compile time. So you can't say `Class<SomeUnknownClass>`.

If you use `Class` for something other than reflection, or reflection to access private fields / methods, then there is a bit of usefulness.

But yes, there are very few methods in `Class` where that type gets involved.

mihnita:
Thank you!

pachecocordovamoiseseduardo-byte (approved):

PR #3954, opened 2026-04-23
ICU-23384 Update to tzdb 2026b

robertbastian:
Files from https://github.com/unicode-org/icu-data/pull/52
including a manual data hack so that the Vancouver offsets continue to be -8+1 rather than -7,
to avoid confusing display names.

#### Checklist
- [x] Required: Issue filed: ICU-23384


icu4c/source/data/misc/zoneinfo64.txt
```diff
@@ -1045,12 +1045,9 @@ zoneinfo64:table(nofallback) {
...
+    typeOffsets:intvector { -29548, 0, -28800, 0, -28800, 3600, -25200, 0 }
```
markusicu, line 1049: With the hack, we technically don't need the last two numbers, right?
(They should be harmless / unused; add just 8 bytes.)

icu4c/source/data/misc/zoneinfo64.txt
```diff
@@ -1045,12 +1045,9 @@ zoneinfo64:table(nofallback) {
...
-    typeOffsets:intvector { -29548, 0, -28800, 0, -28800, 3600 }
-    typeMap:bin { "010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201020102010201" }
-    finalRule { "Canada" }
-    finalRaw:int { -28800 }
-    finalYear:int { 2008 }
```
markusicu, line 1053: I don't know what these lines do.
Should we retain one or both of them given the hack?
@yumaoka ?

markusicu (commented):
commenting here because of the visible diffs from before

icu4c/source/data/misc/zoneinfo64.txt
robertbastian: this configures the rule for after the last ~offset~ transition. So use the "Canada" rule, with a standard offset of -8 after the year 2008

as we want to be on permanent -8+1, we don't want to apply this rule anymore

icu4c/source/data/misc/zoneinfo64.txt
robertbastian: yes

richgillam (commented):
LOKTM if it LOKT everybody else.  :-)

markusicu (approved):

Yqwed:
I assume
```
        TimeZone timeZone = TimeZone.getTimeZone("America/Vancouver");
        long offsetSampleTimeMillis = 1816732800000L; // 28 Jul 2027
        int[] offsets = new int[2];
        timeZone.getOffset(offsetSampleTimeMillis, false /* local */, offsets);
 ```
setting `offsets` to `[-28800000, 3600000]` is intentional given the manual data hack?

I wonder if it will affect users or not.

robertbastian:
Yes, this is an intentional change in order to get the "Pacific Daylight Time" display name. If the offsets were `[-25200000, 0]`, this would be treated as a standard time, and produce "Pacific Standard Time", which is wrong.

It'll affect users in the sense that
* they get the right display name
* if they ask whether the zone is in DST, ICU will return yes
  * we believe this is the right behaviour for now, as long as we use "Pacific Time" display names, however is-dst checking is fairly philosophical anyway

pachecocordovamoiseseduardo-byte (approved):

PR #3955, opened 2026-04-23
ICU-23247 Fix all 'Finally' errorprone issues

mihnita:
#### Checklist

### DIFF
--- a/icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/docs/APIData.java
+++ b/icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/docs/APIData.java
@@ -64,46 +64,43 @@ public final class APIData {
         }
     }
 
+    private static APIData read(InputStream is, boolean internal) throws IOException {
+        try (InputStreamReader isr = new InputStreamReader(is, StandardCharsets.UTF_8);
+                BufferedReader br = new BufferedReader(isr)) {
+            return read(br, internal);
+        }
+    }
+
     public static APIData read(File file, boolean internal) {
         String fileName = file.getName();
-        ZipFile zf = null;
         try {
-            InputStream is;
             if (fileName.endsWith(".zip")) {
-                zf = new ZipFile(file);
-                Enumeration<? extends ZipEntry> entryEnum = zf.entries();
-                if (entryEnum.hasMoreElements()) {
-                    ZipEntry entry = entryEnum.nextElement();
-                    is = zf.getInputStream(entry);
-                    // we only handle one!!!
-                } else {
-                    throw new IOException("zip file is empty");
+                try (ZipFile zf = new ZipFile(file)) {
+                    Enumeration<? extends ZipEntry> entryEnum = zf.entries();
+                    if (entryEnum.hasMoreElements()) {
+                        ZipEntry entry = entryEnum.nextElement();
+                        // we only handle one!!!
+                        try (InputStream is = zf.getInputStream(entry)) {
+                            return read(is, internal);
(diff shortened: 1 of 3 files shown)


icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/docs/APIData.java
```diff
@@ -64,46 +64,42 @@ static APIData read(BufferedReader br, boolean internal) {
...
-                zf = new ZipFile(file);
+        if (fileName.endsWith(".zip")) {
+            try (ZipFile zf = new ZipFile(file)) {
                 Enumeration<? extends ZipEntry> entryEnum = zf.entries();
                 if (entryEnum.hasMoreElements()) {
                     ZipEntry entry = entryEnum.nextElement();
-                    is = zf.getInputStream(entry);
                     // we only handle one!!!
+                    try (InputStream is = zf.getInputStream(entry)) {
+                        return is;
```
roubert: How does this work!? Won't this call `InputStream.close()` when the `return` statement executes and the `try`-block finishes, thereby returning an already closed stream?

mihnita: You are right. I should have tested.
But now I fixed it, and I tested.
Thanks!

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/docs/APIData.java
```diff
@@ -64,46 +64,42 @@ static APIData read(BufferedReader br, boolean internal) {
...
+                        // we only handle one!!!
+                        try (InputStream is = zf.getInputStream(entry)) {
+                            return read(is, internal);
+                        }
+                    } else {
+                        throw new IOException("zip file is empty");
+                    }
+                }
+            } else if (fileName.endsWith(".gz")) {
+                try (InputStream is = new GZIPInputStream(new FileInputStream(file))) {
```
roubert: Does this really work as expected, ie. getting the `FileInputStream` that isn't assigned to any variable here closed in the case of an exception? I would have thought it necessary to split this into two semicolon separated statements just like you do in the `try`-block in your new `read()` above.

(Or, the other way around: If it works like this, there wouldn't be any need for the separate `isr` variable in the code above.)

roubert (commented):
Just one comment, the rest looks good to me now, but it seems you'll also have to run `mvn spotless:apply` before this becomes possible to merge.

icu4j/tools/build/src/main/java/com/ibm/icu/dev/tool/docs/APIData.java
mihnita: Thank you.

It works both ways.
The trouble is, there is no way to tell.

With `Foo f=new Foo(new Bar())` you have to trust that the `Foo` implementation of `close()` also closes the stream used in new.
All JDK classes do that. Not clearly documented, but they do.

But you can't tell for sure what a 3rd party class does. You can trust it, but...

When you do `try(Bar b=new Bar(); Foo f=new Foo(b))` the try will close both, for sure.
That is why in general I prefer the `;` separated style.

I checked the sources for `GZIPInputStream` and `close()` also closes the input stream.

But I split it anyway, for consistency, and for the reason above.

Thank you!
M.

mihnita:
Also formatted 
M

roubert (approved):
LGTM

mihnita:
Thank you!
M

PR #3967, opened 2026-05-01
ICU-23248 Errorprone doc: edit errorprone.cfg, not pom.xml

mihnita:
Editing `errorprone.cfg` is a lot easier than editing `pom.xml`.

And the tag should end in `errorprone.cfg` anyway.

So instead of (current doc):
1. `pom.xml`: find the right place in the big xml
1. `pom.xml`: add `-XepPatchLocation:IN_PLACE -XepPatchChecks:UnusedMethod`
1. run the patch step
1. `pom.xml`: revert the changes
1. `errorprone.cfg`: add `-Xep:UnusedMethod:ERROR`

After this change:
1. `errorprone.cfg`: add `-XepPatchLocation:IN_PLACE -XepPatchChecks:UnusedMethod`, anywhere
1. run the patch step
1. `errorprone.cfg`: remove `-XepPatchLocation:IN_PLACE`
1. `errorprone.cfg`: change `-XepPatchChecks:UnusedMethod` to `-Xep:UnusedMethod:ERROR`

#### Checklist
- [x] Required: Issue filed: ICU-23248

### DIFF
--- a/docs/devsetup/java/errorprone/index.md
+++ b/docs/devsetup/java/errorprone/index.md
@@ -74,13 +74,13 @@ Always check, visually and by running all tests.
 
 ### Preparation
 
-1. Edit the `pom.xml` in the root of the ICU repo.
-1. Search for the `errorprone-all` profile, and in the `compilerArgs` section of
-  that profile find the `<arg>` element that contains `-Xplugin:ErrorProne`.
-1. After `-Xplugin:ErrorProne` add `-XepPatchChecks:<listOfIssueTypes> -XepPatchLocation:IN_PLACE`.
-  For example `-XepPatchChecks:UnusedMethod`. \
-  You can also specify more than one kind of issues to fix, using comma as a separator,
-  for example `-XepPatchChecks:UnusedMethod,UnusedVariable`.
+1. Edit the `errorprone.cfg` file in the root of the ICU repo.
+1. Add two more flags: `-XepPatchLocation:IN_PLACE` and `-XepPatchChecks:<IssueType>`, on separate lines. \
+    For example:
+    ```
+    -XepPatchLocation:IN_PLACE
+    -XepPatchChecks:UnusedMethod
+    ```
 
 ### Apply the fixes
 
@@ -108,24 +108,24 @@ mvn spotless:apply
 mvn package
 ```
 
-**Visually inspect the changed files** (`git diff` or a GUI git client are handy here). Compare with the initial errorprone report.
+**Visually inspect the changed files** (`git diff` or a GUI git client are handy here).
+Compare with the initial errorprone report.
 
 ### Prepare to submit
 
-Update `errorprone.cfg` to declare the issue you just fix as an error. \
+1. Update `errorprone.cfg` to declare the issue you just fix as an error. \
 That way it will prevent it from coming back in the future.
-In our example add `-Xep:UnusedMethod:ERROR`.
-
-Revert the `pom.xml` changes (`git checkout -- pom.xml`).
-
-Run all the errorprone checks (see [“Running the checks”](#running-the-checks))
-to see that the all issues of that type are fixed.
-
-Run a regular build with testing:
-```sh
-mvn clean package
-```
-
-You might have to go back to [“Test, cleanup”](#test-cleanup) if something fails.
-
-Finally, create a PR, commit, push, send for review.
+    * Change `-XepPatchChecks:<IssueType>` to `-Xep:<IssueType>:ERROR`. \
+      In our example `-XepPatchChecks:UnusedMethod` to `-Xep:UnusedMethod:ERROR`.
+    * Move the `-Xep:<IssueType>:ERROR` together with the other `:ERROR` tags
+      if not there, keeping that block of tags sorted.
+    * Remove `-XepPatchLocation:IN_PLACE`.
+1. Check that everything is fine
+    * Run all the errorprone checks (see [“Running the checks”](#running-the-checks))
+to see that all the issues of that type are fixed.
+    * Run a regular build with testing:
+        ```sh
+        mvn clean package
+        ```
+1. You might have to go back to [“Test, cleanup”](#test-cleanup) if something fails.
+1. Finally, create a PR, commit, push, send for review.


docs/devsetup/java/errorprone/index.md
```diff
@@ -74,13 +74,15 @@ Always check, visually and by running all tests.
...
-  for example `-XepPatchChecks:UnusedMethod,UnusedVariable`.
+1. Edit the `errorprone.cfg` file in the root of the ICU repo.
+1. Add two more flags: `-XepPatchLocation:IN_PLACE` and `-XepPatchChecks:<IssueType>`, on separate lines. \
+   For example:
+   ```
+   -XepPatchLocation:IN_PLACE
+   -XepPatchChecks:UnusedMethod
+   ```
+
+A good place for the flags is at the end of the block of `:ERROR` tags (`-Xep:???:ERROR`).
```
echeran: You need 4 spaces (not 3) to keep the subsequent lines included as a part of the list item. The sentence that comes after the list is actually a detail that sounds more specific to step 2 than at the level of the after-the-list, so removed the line in between and changed the indentation so that it would be included, too.

```suggestion
    For example:
    ```
    -XepPatchLocation:IN_PLACE
    -XepPatchChecks:UnusedMethod
    ```
    A good place for the flags is at the end of the block of `:ERROR` tags (`-Xep:???:ERROR`).
```

docs/devsetup/java/errorprone/index.md
```diff
@@ -108,18 +110,22 @@ mvn spotless:apply
...
+**Visually inspect the changed files** (`git diff` or a GUI git client are handy here).
+Compare with the initial errorprone report.
 
 ### Prepare to submit
 
 Update `errorprone.cfg` to declare the issue you just fix as an error. \
-That way it will prevent it from coming back in the future.
-In our example add `-Xep:UnusedMethod:ERROR`.
+That way it will prevent it from coming back in the future. \
+In our example change `-XepPatchChecks:UnusedMethod` to `-Xep:UnusedMethod:ERROR`. \
```
echeran: Remove the trailing `\` at the end of line 118, too.
```suggestion
That way it will prevent it from coming back in the future.
In our example, change `-XepPatchChecks:UnusedMethod` to `-Xep:UnusedMethod:ERROR`.
```

docs/devsetup/java/errorprone/index.md
```diff
@@ -108,18 +110,22 @@ mvn spotless:apply
...
+not there, keeping that block of tags sorted.
 
-Revert the `pom.xml` changes (`git checkout -- pom.xml`).
+Remove `-XepPatchLocation:IN_PLACE`.
 
 Run all the errorprone checks (see [“Running the checks”](#running-the-checks))
-to see that the all issues of that type are fixed.
+to see that all the issues of that type are fixed.
 
 Run a regular build with testing:
```
echeran: Thoughts on turning this into a numbered list? If there is a sequential relationship across these paragraphs, then it's more organized, concise, and clear to make it a list, given the amount of text the reader has to read.

docs/devsetup/java/errorprone/index.md
mihnita: Applied.

The trouble is with the previews from different tools. In my case it looked fine :-(

docs/devsetup/java/errorprone/index.md
mihnita: Applied.

Although I am starting to notice that we have a difference of style.
I tend to add more line breaks, intentionally.

docs/devsetup/java/errorprone/index.md
mihnita: The order is not strict.
And the paragraphs are not on equal footing

1. Update `errorprone.cfg` to declare the issue you just fix as an error. \
    1.1 Change `-XepPatchChecks:<IssueType>` to `-Xep:<IssueType>:ERROR`
    1.2 Move the `-Xep:<IssueType>:ERROR` 
    1.3 Remove `-XepPatchLocation:IN_PLACE`.

The 1.# are sub-items to edit erroprone, but the order is not strict, you can swap them.
So they should probably be a bulleted sub-list.

2. Check that everything is fine:
   2.1 Run all the errorprone checks
   2.2 Run a regular build with testing:

2.1 and 2.2 would be sub-steps of something that does not exist "as such" now.
And they can be swapped, the order is not fixed.
So bulleted sub-list again.

3. You might have to go back

This breaks the "list" flow completely. It's a "goto" from inside a list item.

4. Finally, create a PR, commit, push, send for review.

Yes, that would be last.

---

So it feels to me like there is too much "use a bigger hammer" to make this a list.
With no much benefit, in my opinion.

Even if ordered, not everything requires a list.
I find it perfectly fine to have standalone paragraphs saying "do x" and "do y" without implying a list, even if they are supposed to be sequential.

I actually tried it in this case.
And I don't think it is better. But it is also a matter of opinion.

docs/devsetup/java/errorprone/index.md
echeran: Yes, and the conversation goes like this:

- me: Most people write paragraphs to group related ideas together. I do, too. I believe it helps with readability.
- you: I put in newlines to make sentences stand out, to give them emphasis

My response to that will be:

1. not writing prose in the way most people do causes friction for people reading
1. why is it distracting? -> because trying to add emphasis to several sentences in the manner invites the aphorism, "if everything is important, then nothing is important"
1. there are other formatting features (lists, code fences, note/warning callouts) to achieve special types of emphasis when needed
1. you can't force people to pay careful attention to every point, so attempting to use formatting as means won't work (point 2). the best we can do is call out the high value areas of attention, for which point 3 provides ways of doing that

echeran: I should not assume it to be implied, so I should say explicitly: Thanks for applying the suggestion, even if it's not the first style that you write in.

docs/devsetup/java/errorprone/index.md
mihnita: Here what this would look like as a list.

I don't like it more, and I find it harder to edit. But you might find it better. I don't know. I don't think there is right or wrong here.

---


### Prepare to submit

1. Update `errorprone.cfg` to declare the issue you just fix as an error. \
That way it will prevent it from coming back in the future.
    * Change `-XepPatchChecks:<IssueType>` to `-Xep:<IssueType>:ERROR`. \
      In our example `-XepPatchChecks:UnusedMethod` to `-Xep:UnusedMethod:ERROR`.
    * Move the `-Xep:<IssueType>:ERROR` together with the other `:ERROR` tags
      if not there, keeping that block of tags sorted.
    * Remove `-XepPatchLocation:IN_PLACE`.
1. Check that everything is fine
    * Run all the errorprone checks (see [“Running the checks”](#running-the-checks))
to see that all the issues of that type are fixed.
    * Run a regular build with testing:
        ```sh
        mvn clean package
        ```
1. You might have to go back to [“Test, cleanup”](#test-cleanup) if something fails.
1. Finally, create a PR, commit, push, send for review.

mihnita: Note: if you find the list version as above to be an improvement, I am not against changing and submitting it in list form.

echeran: I think the list version you created is a clear improvement. Thank you for that. I do prefer it strongly.

I happened to look at the formatted list before I saw the explanation in the comment preceding it. The visual effect that came from formatting of the list and the sublists (including your choices of ordered and unordered) told the whole story of the explanation without words.

mihnita: Between "clear improvement" and "prefer it strongly", I've done it :-)
Thank you!

mihnita:
Hmmm...

Look at the second commit (https://github.com/unicode-org/icu/pull/3967/changes/08920bf4125b6b91fe1ba30860544ac1cf88194f), _"Apply suggestions from code review"_.

There is something missing, at line 78 or so:

```diff
   For example:
-   ```
-   -XepPatchLocation:IN_PLACE
-   -XepPatchChecks:UnusedMethod
-   ```
-
-A good place for the flags is at the end of the block of `:ERROR` tags (`-Xep:???:ERROR`).
```

I don't remember applying that, not in that form.
And the end result looks incomplete (_"For example:"_ without an example)

So I think it was an accident, and I added it back in commit 3.

If that was intentional, and you think I should remove "For example" instead, let me know.
Although I think that the example helps.

echeran (approved):
LGTM. Thanks!

mihnita:
Thank you very much!
And have a nice week-end!
M.

PR #3973, opened 2026-05-07
ICU-23085 Refined Doclet CheckTags and fixed the existing API tagging problems

yumaoka:
- Dropped `@provisional` requirement
- Allowed `{@icu _usage_}` for Enum type
- Doclet CheckTags#run to return false when tag validation fails
- Fixed existing ICU4J API tagging issue in the main branch
- Updated maven checktags profile to fail on check tags validation failure

#### Checklist
- [x] Required: Issue filed: ICU-23085

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
@@ -15,9 +15,9 @@ import java.util.LinkedList;
 import java.util.List;
 
 /**
- * {@icu} BasicTimeZone extends <code>TimeZone</code> with additional methods to access time zone
- * transitions and rules. All ICU <code>TimeZone</code> concrete subclasses extend this class. APIs
- * added to <code>java.util.TimeZone</code> by <code>BasicTimeZone</code> are annotated with
+ * {@icu _usage_} BasicTimeZone extends <code>TimeZone</code> with additional methods to access time
+ * zone transitions and rules. All ICU <code>TimeZone</code> concrete subclasses extend this class.
+ * APIs added to <code>java.util.TimeZone</code> by <code>BasicTimeZone</code> are annotated with
  * <strong>'<span style="color:red">[icu]</span>'</strong>.
  *
  * @see com.ibm.icu.util.TimeZoneRule
@@ -679,9 +679,9 @@ public abstract class BasicTimeZone extends TimeZone {
     }
 
     /**
-     * {@icu} Options used by {@link #getOffsetFromLocal(long, LocalOption, LocalOption, int[])} to
-     * specify how to interpret an input time when it does not exist, or when it is ambiguous,
-     * around a time zone transition.
+     * {@icu _usage_} Options used by {@link #getOffsetFromLocal(long, LocalOption, LocalOption,
+     * int[])} to specify how to interpret an input time when it does not exist, or when it is
+     * ambiguous, around a time zone transition.
      *
      * @stable ICU 69
      */
--- a/icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
@@ -331,6 +331,11 @@ public abstract class UResourceBundle extends ResourceBundle {
         return getULocale().toLocale();
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @stable ICU 3.0
+     */
     @Override
     public String toString() {
         return "UResourceBundle{name=" + getBaseBundleName() + ", locale=" + getLocale() + '}';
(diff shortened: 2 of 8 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
```diff
@@ -15,9 +15,9 @@
 import java.util.List;
 
 /**
- * {@icu} BasicTimeZone extends <code>TimeZone</code> with additional methods to access time zone
- * transitions and rules. All ICU <code>TimeZone</code> concrete subclasses extend this class. APIs
- * added to <code>java.util.TimeZone</code> by <code>BasicTimeZone</code> are annotated with
+ * {@icu _usage_} BasicTimeZone extends <code>TimeZone</code> with additional methods to access time
```
markusicu, line 18: Is `{@icu _usage_}` actually the intended syntax?

icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
```diff
@@ -331,6 +331,11 @@ public Locale getLocale() {
         return getULocale().toLocale();
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * @stable ICU 79
```
markusicu: It can't be stable as of a future version...

markusicu (commented):
thanks for the cleanup!
basically rslgtm
Mihai ok?

icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
yumaoka: I will send a message to ICU TC list about this. toString() was implemented after 78. Before that, toString() used the Java's Object#toString(). I don't remember we discussed about policy for API status when we implement a method from JDK class. I looked at other cases and they are using `@stable` at the version when the implementation was provided. `@draft` looks awkward, becasuse the method implicitly existed.

icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
yumaoka: This works like a template and produces - 

```
Methods, fields, and other functionality specific to ICU are labeld '[icu]'.
```

CheckTags does not allow `{@icu}` in fields and methods in class description. I don't know when this was actually introduced. Whether we need `{@icu}` annotation in BasicTimeZone or not is acually questionable. But if methods in class already have `{@icu}`, we should have `{@icu _usage_}` in the class description. My understanding is `{@icu}` should be used only in class which has 1 to 1 JDK class mapping. BasicTimeZone is an ICU specific class extending TimeZone which has 1 to 1 JDK class corresponding. It makes sense to have `{@icu}` in TimeZone, but I feel it's not necessary for BasicTimeZone.

I will review classes with `{@icu}` and see if each of them really make sense.

icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
markusicu: Makes sense. I didn't realize that this was a new implementation.

I know that we have talked about tagging override methods, but I don't remember what we concluded :-(
If you observe a certain pattern, then let's follow it -- and we need to find a place where we document it. Maybe in the design chapter where we document API tags and stability.

icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
markusicu: Ok, makes sense, and I trust you.
If we just need `{@icu _usage_}` because for some unknown reason we can't just use `{@icu}`, then at some point we should tackle that...

> I will review classes with `{@icu}` and see if each of them really make sense.

+1

markusicu (dismissed):

icu4j/main/core/src/main/java/com/ibm/icu/util/BasicTimeZone.java
mihnita: I was also puzzled by `{@icu _usage_}` and did some digging into if (without seeing this comment).
And it is already used on several classes. But didn't know the reason.
So, thank you, I learned something.

icu4j/main/core/src/main/java/com/ibm/icu/util/UResourceBundle.java
mihnita: I think that technically is not a new API.
I was able to call `toString()` before, and I can call `toString()` now.
The behavior changed, but the API was there "since forever" (of course :-)

So I think stable is fine.

mihnita (dismissed):
Thank you!

markusicu (approved):

PR #3975, opened 2026-05-08
ICU-23401 Correctly implement rule LB3, thereby removing a few more states

eggrobin:
The line.txt rules have rule statuses distinguishing mandatory breaks (status UBRK_LINE_HARD=100) from break opportunities (status UBRK_LINE_SOFT=0).

However, they incorrectly return UBRK_LINE_SOFT at end of text unless the last character is a hard line break, contrary to https://www.unicode.org/reports/tr14/#LB3. This is unlikely to matter to callers, as the start and end of text would most likely be special-cased in any actual implementation, but it results in unnecessary splitting of states.

Correcting the rules to treat eot as a hard break removes four states from the state machine, as it no longer needs to make unnecessary distinctions between eot and LF etc.: https://github.com/eggrobin/unicodetools/compare/496e9a0b9924b6ea824ba6b675754bd9d41502f7...cb6c98a22c975391931d088dc3e4debf479fa3e6

#### Checklist
- [x] Required: Issue filed: ICU-23401


icu4c/source/common/rbbiscan.cpp
```diff
@@ -1258,7 +1258,7 @@ void RBBIRuleScanner::scanSet() {
     // Use tempSet to handle the case that the UnicodeSet contains
     // only string element, such as [{ab}] and treat it as empty set.
     tempSet.removeAllStrings();
-    if (tempSet.isEmpty()) {
+    if (tempSet.isEmpty() && !uset->contains(UnicodeString(u"eof"))) {
```
markusicu: you should not need the explicit UnicodeString constructor... unless the compiler literally wants to degrade the string literal to a pointer and we forbid the from-pointer constructor inside icu (don't remember). we could use a string_view literal...

markusicu:
@aheninger FYI

icu4c/source/common/rbbiscan.cpp
eggrobin: This is 
```
  template<typename S, typename = std::enable_if_t<ConvertibleToU16StringView<S>>>
  UNISTR_FROM_STRING_EXPLICIT UnicodeString(const S &text)
```
and `UNISTR_FROM_STRING_EXPLICIT` is defined as `explicit`.

Arguably I could be using UnicodeString::readOnlyAlias.

markusicu: Somehow I thought we had added something like UnicodeSet::contains(u16string_view) but we only really added special signatures on some UnicodeString functions...

markusicu (approved):

icu4c/source/common/rbbiscan.cpp
eggrobin: Maybe we should add that!

PR #3978, opened 2026-05-12
ICU-23403 Fix data race in CurrencyPrecision.withCurrency

Manishearth:
Clone the Precision object returned by constructFromCurrency before mutating its trailingZeroDisplay field. This prevents mutating shared static instances like FIXED_FRAC_0 or FIXED_FRAC_2 when multiple threads concurrently call withCurrency.

This was investigated and fixed by Gemini.

#### Checklist
- [x] Required: Issue filed: /ICU-23403

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/number/CurrencyPrecision.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/number/CurrencyPrecision.java
@@ -37,8 +37,9 @@ public abstract class CurrencyPrecision extends Precision {
     public Precision withCurrency(Currency currency) {
         if (currency != null) {
             Precision retval = constructFromCurrency(this, currency);
-            retval.trailingZeroDisplay = trailingZeroDisplay;
-            return retval;
+            // trailingZeroDisplay returns a clone if trailingZeroDisplay differs,
+            // preventing data races on shared static instances (e.g. FIXED_FRAC_0).
+            return retval.trailingZeroDisplay(trailingZeroDisplay);
         } else {
             throw new IllegalArgumentException("Currency must not be null");
         }
--- a/icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
@@ -346,6 +346,10 @@ public abstract class Precision {
      * @stable ICU 69
      */
     public Precision trailingZeroDisplay(TrailingZeroDisplay trailingZeroDisplay) {
+        // Question: Should we avoid cloning if one is null and the other is AUTO?
+        if (this.trailingZeroDisplay == trailingZeroDisplay) {
+            return this;
+        }
         Precision result = this.createCopy();
         result.trailingZeroDisplay = trailingZeroDisplay;
         return result;
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/number/NumberFormatterApiTest.java
@@ -6960,4 +6960,43 @@ public class NumberFormatterApiTest extends CoreTestFmwk {
         FormattedValueTest.checkFormattedValue(
                 message, result, result.toString(), expectedFieldPositions);
     }
+
+    @Test
+    public void testCurrencyPrecisionCacheMutation() {
+        Currency usd = Currency.getInstance("USD");
+
+        // Precision gets cached, we want to make sure the cache isn't overwritten.
+
+        // 1. Format using the cached Precision as is
+        String actual1 =
+                NumberFormatter.with()
+                        .precision(Precision.currency(CurrencyUsage.STANDARD))
+                        .locale(Locale.US)
+                        .unit(usd)
+                        .format(123.0)
+                        .toString();
+        assertEquals("Initial formatting with cached Precision", "$123.00", actual1);
+
+        // 2. Format modifying the Precision's trailingZeroDisplay
+        String actual2 =
+                NumberFormatter.with()
+                        .precision(
+                                Precision.currency(CurrencyUsage.STANDARD)
+                                        .trailingZeroDisplay(TrailingZeroDisplay.HIDE_IF_WHOLE))
+                        .locale(Locale.US)
+                        .unit(usd)
+                        .format(123.0)
+                        .toString();
+        assertEquals("Formatting with modified trailingZeroDisplay", "$123", actual2);
+
+        // 3. Format like the first time again, expect the same result
+        String actual3 =
+                NumberFormatter.with()
+                        .precision(Precision.currency(CurrencyUsage.STANDARD))
+                        .locale(Locale.US)
+                        .unit(usd)
+                        .format(123.0)
+                        .toString();
+        assertEquals("Subsequent formatting with cached Precision", "$123.00", actual3);
+    }
 }


icu4j/main/core/src/main/java/com/ibm/icu/number/CurrencyPrecision.java
```diff
@@ -36,7 +36,9 @@ public abstract class CurrencyPrecision extends Precision {
      */
     public Precision withCurrency(Currency currency) {
         if (currency != null) {
-            Precision retval = constructFromCurrency(this, currency);
+            // Clone the Precision object to prevent data races when mutating fields
+            // (like trailingZeroDisplay) on shared static instances (e.g., FIXED_FRAC_0).
+            Precision retval = constructFromCurrency(this, currency).createCopy();
             retval.trailingZeroDisplay = trailingZeroDisplay;
```
Manishearth: we could in theory add `.withTrailingDisplay` so we do not have to always copy, similar to withMode

https://github.com/unicode-org/icu/blob/main/icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java#L361-L368

Manishearth:
I'm not sure how to best test this.

Gemini generated the following complicated test


<details>


```java
    @Test
    public void testCurrencyPrecisionConcurrency() throws InterruptedException {
        final Currency usd = Currency.getInstance("USD"); // 2 decimals (maps to FIXED_FRAC_2)
        
        final CurrencyPrecision cp1 = Precision.currency(CurrencyUsage.STANDARD)
                .trailingZeroDisplay(TrailingZeroDisplay.HIDE_IF_WHOLE);
        final CurrencyPrecision cp2 = Precision.currency(CurrencyUsage.STANDARD)
                .trailingZeroDisplay(TrailingZeroDisplay.AUTO);
        int numThreads = 10;
        final int iterations = 5000;
        Thread[] threads = new Thread[numThreads];
        final java.util.concurrent.atomic.AtomicInteger failures = new java.util.concurrent.atomic.AtomicInteger(0);
        
        for (int i = 0; i < numThreads; i++) {
            final int threadId = i;
            threads[i] = new Thread(new Runnable() {
                @Override
                public void run() {
                    for (int j = 0; j < iterations; j++) {
                        if (threadId % 2 == 0) {
                            Precision p = cp1.withCurrency(usd);
                            String formatted = NumberFormatter.with()
                                    .precision(p)
                                    .locale(Locale.US)
                                    .unit(usd)
                                    .format(123.00)
                                    .toString();
                            // USD standard format for 123.00 is "$123.00".
                            // If HIDE_IF_WHOLE is working, it should be "$123".
                            if (!formatted.equals("$123")) {
                                failures.incrementAndGet();
                            }
                        } else {
                            Precision p = cp2.withCurrency(usd);
                            String formatted = NumberFormatter.with()
                                    .precision(p)
                                    .locale(Locale.US)
                                    .unit(usd)
                                    .format(123.00)
                                    .toString();
                            // If AUTO is working, it should be "$123.00".
                            if (!formatted.equals("$123.00")) {
                                failures.incrementAndGet();
                            }
                        }
                    }
                }
            });
        }
        for (Thread t : threads) {
            t.start();
        }
        for (Thread t : threads) {
            t.join();
        }
        
        assertEquals("Concurrently configured precisions should not corrupt each other's formatting behavior", 0, failures.get());
    }

```

</details>

markusicu:
Thanks! Discussion:
- the fix looks good for this call site, but...
- we do need a test; should be simple, something like
    - format something using the cached Precision as is
    - format something in a way that modifies the Precision
    - format like the first time again -- with the bug it gives the wrong result
- something like withTrailingDisplay() would help with performance
- the real problem is caching a mutable object
- we would like to solve that, for a robust solution
- @sffc we would like your help here
- we _assume_ that this is not a problem in C++ because we use `const` there which would force us to make a copy before modifying; someone should verify
- Shane: why is Precision mutable, at least in Java?
- if we don't want to radically change the design, consider making Precision Freezable, and freeze() the cached instances?

icu4j/main/core/src/main/java/com/ibm/icu/number/CurrencyPrecision.java
sffc: Yes, please avoid creating new Java objects when possible

sffc (commented):
This was added together with C++ in https://github.com/unicode-org/icu/commit/b79c299f90d4023ac237db3d0335d568bf21cd36

The C++ code appears to use the copy constructor; it is different than Java because it stores the settings locally in a struct instead of retaining static references to things.

sffc:
> the real problem is caching a mutable object

This is not intended to be a mutable object. The problem is code setting a private field at a distance.

sffc:
> if we don't want to radically change the design, consider making Precision Freezable, and freeze() the cached instances?

This helps nothing. Freezable is an ICU4J abstraction that relies on code implementing it correctly. It does not protect against code from somewhere else in the program setting private fields.

sffc:
You don't need to look far to see that the field is designed to be private and immutable. I don't remember the reason they are package-private instead of private, and why they are not `final`.

```java
public abstract class Precision {

    /* package-private final */ MathContext mathContext;
    /* package-private final */ TrailingZeroDisplay trailingZeroDisplay;
```

Manishearth:
Tried to add a simple test. We already had a `withTrailingDisplay`-equivalent method so I reused that.

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
```diff
@@ -346,6 +346,13 @@ public static CurrencyPrecision currency(CurrencyUsage currencyUsage) {
      * @stable ICU 69
      */
     public Precision trailingZeroDisplay(TrailingZeroDisplay trailingZeroDisplay) {
+        // In Precision, trailingZeroDisplay is null by default, which behaves identically
+        // to AUTO (see setResolvedMinFraction). Treat null and AUTO as equivalent to
```
Manishearth: the LLM noticed this null-auto correspondence, I verified it, btu I'm not 100% sure, this isn't my codebase

Manishearth:
The test won't test against the concurrency issue but it ought to test against the underlying bug. Good enough.

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
sffc: This "invariant" holds for basically all of the options. As in ECMA-402 and as in ICU4X, `null` and `AUTO` are subtly different. Please undo this change and if you like file a separate issue to change it consistently for all of the options, not just this one.

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
```diff
@@ -346,6 +346,9 @@ public static CurrencyPrecision currency(CurrencyUsage currencyUsage) {
      * @stable ICU 69
      */
     public Precision trailingZeroDisplay(TrailingZeroDisplay trailingZeroDisplay) {
+        if (this.trailingZeroDisplay == trailingZeroDisplay) {
```
sffc, line 350: sorry, I guess at least preserve the LLM's observation in a comment, something like

```
// Question: Should we avoid cloning if one is null and the other is AUTO?
```

I also observe that some of the other getters don't even have a `==` value check; they just unconditionally clone. That might be an oversight.

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
```diff
@@ -346,6 +346,9 @@ public static CurrencyPrecision currency(CurrencyUsage currencyUsage) {
...
+            return this;
+        }
         Precision result = this.createCopy();
         result.trailingZeroDisplay = trailingZeroDisplay;
```
sffc, line 354: Thought: Having to set the field after cloning the parent class is probably the reason why the field is not actually `final`. There's probably a way to make it actually be `final` but the code probably gets more complex.

sffc (approved):

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
```diff
@@ -346,6 +346,10 @@ public static CurrencyPrecision currency(CurrencyUsage currencyUsage) {
      * @stable ICU 69
      */
     public Precision trailingZeroDisplay(TrailingZeroDisplay trailingZeroDisplay) {
+        // Question: Should we avoid cloning if one is null and the other is AUTO?
```
markusicu, line 349: sounds like that would be a good optimization -- assuming null is possible

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
markusicu: We could add a constructor that takes a Precision source/template plus one or more overrides.

markusicu (approved):

markusicu:
> > if we don't want to radically change the design, consider making Precision Freezable, and freeze() the cached instances?
> 
> This helps nothing. Freezable is an ICU4J abstraction that relies on code implementing it correctly. It does not protect against code from somewhere else in the program setting private fields.

It's pretty effective in the normal case where external code can't just reach into the class internals.

markusicu:
> You don't need to look far to see that the field is designed to be private and immutable. I don't remember the reason they are package-private instead of private

truly private would help a lot...

Manishearth:
(n.b. I can't merge this myself)

PR #3979, opened 2026-05-12
ICU-23404 Dictionary break opportunities before a hard break should not be hard breaks

eggrobin:
People have been copying the existing rules to tweak them here and there; adding syntax to specify what status the dictionary breaks would have would be problematic for such forks, which would most likely never receive the new directive.  Use the rule status values instead.

#### Checklist
- [x] Required: Issue filed: ICU-23404

### DIFF
--- a/docs/userguide/boundaryanalysis/break-rules.md
+++ b/docs/userguide/boundaryanalysis/break-rules.md
@@ -270,6 +270,39 @@ See, for example, this snippet from the [line break rules](https://github.com/un
     $dictionary = [$SA];
 ```
 
+The status value of dictionary breaks is determined as follows:
+* the status value of the final break of the rule-based segment refined by the
+  dictionary breaks, if the largest status value defined in the rules is greater
+  than 100 (in that case, the rules are called *word-like*).
+* 0 otherwise.
+
+> **Note:** In practice, only word segmentation is word-like.
+> The need for a distinct behaviour was realized long after status values were
+> introduced: prior to ICU 79, all rules were considered word-like.  Using the
+> largest status value as a heuristic allows rules that are customized versions
+> of the word breaking rules to behave like word segmentation should, without
+> needing to introduce a new syntax to select the status of dictionary breaks.
+
+> **Example:**
+> With the rules
+> ```
+> $dictionary = [A-Z];
+> $ {100};
+> [A-Z] [A-Z];
+> ```
+> The string `ARMAVIRUMQUECANO` has a final break with status 100.
+> These rules are not word-like, so if dictionary breaking finds breaks between
+> `ARMA`, `VIRUMQUE`, and `CANO`,
+> These will have a status value of 0.
+> If however the following rule is added,
+> ```
+> . [?] {200};
+> ```
+> the rules become word-like, and breaks within `ARMAVIRUMQUECANO` all get a
+> status value of 100. Any dictionary breaks within
+> `QUOUSQUETANDEMABUTERECATILINAPATIENTIANOSTRA?` would get a status value of
+> 200.
+
 ## Rule Options
 
 | Option          | Description |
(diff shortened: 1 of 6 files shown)


eggrobin:
Note: In ICU-TC yesterday, Markus asked whether the language break engine should be setting those statuses. In an ideal world that would make sense, but the API of LanguageBreakEngine is status-unaware: it neither has a way to know which statuses are appropriate for the enclosing rules (which is the basic problem solved by looking at all the statuses here) nor to return the statuses for the breaks. This could be changed, but it would not be trivial. Further, `registerExternalBreakEngine` exists, so we would be breaking external break engines. We can do that since it is in tech preview, but this does not seem worth it for line break statuses, a feature whose existence no-one in ICU-TC remembered.

docs/userguide/boundaryanalysis/break-rules.md
markusicu, line 1: This lgtm for the new behavior, but I think it would be useful to document (briefly) that there is a change in ICU 79 compared with before.

docs/userguide/boundaryanalysis/break-rules.md
```diff
@@ -270,6 +270,39 @@ See, for example, this snippet from the [line break rules](https://github.com/un
...
+> With the rules
+> ```
+> $dictionary = [A-Z];
+> $ {100};
+> [A-Z] [A-Z];
+> ```
+> The string `ARMAVIRUMQUECANO` has a final break with status 100.
+> These rules are not word-like, so if dictionary breaking finds breaks between
+> `ARMA`, `VIRUMQUE`, and `CANO`,
+> These will have status value of 0.
```
markusicu:
```suggestion
> These will have a status value of 0.
```

docs/userguide/boundaryanalysis/break-rules.md
```diff
@@ -270,6 +270,39 @@ See, for example, this snippet from the [line break rules](https://github.com/un
     $dictionary = [$SA];
 ```
 
+The status value of dictionary breaks is determined as follows:
+* the status value of the final break of the rule-based segment refined by the
+  dictionary breaks, if the largest status value defined in the rules is greater
+  than 100 (in that case, the rules are called *word-like*).
```
markusicu, line 276: just curious: do we use status values > 100 only in the word rules?

docs/userguide/boundaryanalysis/break-rules.md
markusicu, line 1: optional: would it be useful to point out what you said in your PR note, that this is a heuristic which seems good enough, and would need API changes if we needed/wanted to be more explicit?

icu4j/main/core/src/main/java/com/ibm/icu/text/RuleBasedBreakIterator.java
```diff
@@ -1067,6 +1067,11 @@ void reset() {
             fFirstRuleStatusIndex = 0;
             fOtherRuleStatusIndex = 0;
             fBreaks.removeAllElements();
+            int maxStatus = 0;
+            for (int status : fRData.fStatusTable) {
+                maxStatus = Math.max(status, maxStatus);
+            }
+            isWordLike = maxStatus > 100;
         }
         ;
```
markusicu, line 1076: spurious semicolon?

eggrobin: Yes, pre-existing. This file is full of those, presumably because they were copied from C++ and then put on their own line when everything was spotlessed. Let’s clean these up separately.

docs/userguide/boundaryanalysis/break-rules.md
eggrobin: Yes. Line and sentence have 100 for hard line breaks, gcb has no statuses.

docs/userguide/boundaryanalysis/break-rules.md
eggrobin: I sort of do in the note:
> **Note:** In practice, only word segmentation is word-like.
> The need for a distinct behaviour was realized long after status values were
> introduced.  Using the largest status value allows rules that are customized
> versions of the word breaking rules to behave like word segmentation should,
> without needing to introduce a new syntax to select the status of dictionary
> breaks.

eggrobin: I can edit that note to stick in the word heuristic.

eggrobin: Added a mention of that in the note.

markusicu (approved):

PR #3994, opened 2026-05-20
ICU-23376 Minimize the number of states of the DFA: Java Edition

eggrobin:
Same as #3948 (but vastly more readable because we have such luxuries as maps).

#### Checklist
- [x] Required: Issue filed: ICU-23376

ALLOW_MANY_COMMITS=meow


eggrobin:
@markusicu FYI I intend to keep 608d8b805e3f6d7d00740dfe37baf38a4e2aedaf as its own commit once I squish things, because once we have decent data structures in the C++, we will probably not have LinkedHashMap equivalent, but we will definitely have a TreeMap equivalent, so we will want to revert it to match the C++.

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
```diff
@@ -800,6 +804,146 @@ void mergeRuleStatusVals() {
...
+        Map<StateType, List<Integer>> initialPartition = new LinkedHashMap<>();
+        for (int i = 0; i < fDStates.size(); ++i) {
+            final RBBIStateDescriptor state = fDStates.get(i);
+            var type = new StateType(state.fAccepting, state.fLookAhead, state.fTagsIdx);
+            initialPartition.putIfAbsent(type, new ArrayList<>());
+            initialPartition.get(type).add(i);
+        }
+        // The partition Π from Algorithm 3.6.
+        // (We could call it Π, but some member companies that integrate the ICU
+        // code base prohibit non-ASCII identifiers…).
```
eggrobin, line 858: Is that the case in Java too, or could I call it Π?

macchiati: It's ok in Java...

eggrobin: It’s been OK in C++ for 15 years too, but are there some Google Java style checks that prohibit it? (As far as I know Google C++ prohibits non-ASCII identifiers…)

markusicu: If we are sticking with Google style...
https://google.github.io/styleguide/javaguide.html#s5.1-identifier-names
> Identifiers use only ASCII letters and digits, and, in a small number of cases noted below, underscores. Thus each valid identifier name is matched by the regular expression `\w+` .

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
```diff
@@ -800,6 +804,146 @@ void mergeRuleStatusVals() {
...
+            partition.add(part);
+        }
+        // Given the index of a state 𝑠 in `fDStates`, returns a List of integers
+        // σ(𝑠) such that 𝑠.fDtran[i] (the index of the state reached by the
+        // transition from 𝑠 on the character class with index i) is in Π[σ(𝑠)[i]].
+        // We then have σ(𝑠) == σ(𝑡) if and only if “for all input symbols 𝑎, states
+        // 𝑠 and 𝑡 have transitions on 𝑎 to states in the same group of Π” (which is
+        // what defines the refinement of 𝐺 in Fig. 3.45).
+        Function<Integer, List<Integer>> partitionSignature =
+                (Integer stateIndex) -> {
```
markusicu, line 870: optional: Is this lambda better/clearer/... than a private final function with a name adjacent to this function?

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
eggrobin: Well, ICU certainly doesn’t abide by Google style, neither in Java nor in C++. The question is whether having a Π in third_party will cause trouble (e.g., if this is enforced by making the compiler emit an error on such identifiers), or whether nobody will care. In the latter case I strongly prefer mimicking the identifiers used in the textbooks/papers (doing **for**mula **tran**slation, if you will…).

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
```diff
@@ -800,6 +804,146 @@ void mergeRuleStatusVals() {
...
+            public int hashCode() {
+                return Objects.hash(fAccepting, fLookAhead, fTagsIdx);
+            }
+        }
+        Map<StateType, List<Integer>> initialPartition = new LinkedHashMap<>();
+        for (int i = 0; i < fDStates.size(); ++i) {
+            final RBBIStateDescriptor state = fDStates.get(i);
+            var type = new StateType(state.fAccepting, state.fLookAhead, state.fTagsIdx);
+            initialPartition.putIfAbsent(type, new ArrayList<>());
+            initialPartition.get(type).add(i);
```
markusicu: Double lookup. putIfAbsent() returns the previous value, but you can use computeIfAbsent() which returns the new value. (since Java 8)

Something like
```
... .computeIfAbsent(type, t -> new ArrayList<>()).add(i);
```

See https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#computeIfAbsent(K,java.util.function.Function)

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
```diff
@@ -800,6 +804,146 @@ void mergeRuleStatusVals() {
...
+            boolean refined = false;
+            // Group 𝐺 in Π (=`partition`).
+            for (List<Integer> group : partition) {
+                // Partition 𝐺 based on the signature, see above.
+                Map<List<Integer>, List<Integer>> groupRefinement = new LinkedHashMap<>();
+                // Index of a state in the group.
+                for (final int s : group) {
+                    final var signature = partitionSignature.apply(s);
+                    groupRefinement.putIfAbsent(signature, new ArrayList<>());
+                    groupRefinement.get(signature).add(s);
```
markusicu: computeIfAbsent()

markusicu (commented):
some Java comments; otherwise rubber stamp at the ready

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
eggrobin: Well, it needs to access some local variables, so it quickly means `partition` needs to go in its own class on which this would be a method. I could do that, not sure it is worth it.

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
eggrobin: Ah yes, I keep forgetting what that one is called.

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
markusicu: I have tried to find out in Google. There are recommendations for ASCII, and an ErrorProne check for UnicodeInCode motivated by confusables, but that can be turned off. I can't figure out how to search for symbols with non-ASCII characters. I can only guess that "physically" it probably works. It might still be asking for trouble somewhere because it's unusual and rarely exercised.

One of the arguments is that not everyone knows how to read, say, Greek letters.

I have a preference for being consistent between ICU4C and ICU4J.

@mihnita also prefers sticking with ASCII...

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
markusicu: good enough reason - tnx

icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/char.brk
markusicu, line 1: is this completely empty?

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java
eggrobin: > One of the arguments is that not everyone knows how to read, say, Greek letters.

Well, this comment by Andy applies to the new function as well:
https://github.com/unicode-org/icu/blob/8c187df31bf7e31a1724d98bcb60ba6b3fb6f5d2/icu4j/main/core/src/main/java/com/ibm/icu/text/RBBITableBuilder.java#L161-L163

And of course the dragon book has a Π, so a reader whose brain gets fried by that will have no luck making sense of that code anyway.

> It might still be asking for trouble somewhere because it's unusual and rarely exercised.

That might be true. (But of course, it is quite sad that we would give up on that in a Unicode project.)

icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/char.brk
eggrobin: The size _difference_ is 0 bytes, the size is not 0.
It changed at least because the states get reordered by the minimization algorithm, even though the minimization does not do anything here.

eggrobin:
@markusicu please note that you might need to perform your byte-flipping magic on those .brks before rubberstamping.

markusicu:
> The .brks (@markusicu to flip the bytes)

It might take me a couple of days to dig myself out of my current hole...
Could I ask you to regenerate these yourself?
[This is how I have done this](https://github.com/unicode-org/icu/blob/8c187df31bf7e31a1724d98bcb60ba6b3fb6f5d2/icu4c/source/data/unidata/changes.txt#L310-L344).

eggrobin:
> [This is how I have done this](https://github.com/unicode-org/icu/blob/8c187df31bf7e31a1724d98bcb60ba6b3fb6f5d2/icu4c/source/data/unidata/changes.txt#L310-L344).

Those instructions aren’t going to help me (short of setting up a new ICU development setup on a new machine, and that seems like an onerous first step).

I tried this https://github.com/unicode-org/icu/pull/3810#discussion_r2628972918, specifically,
```powershell
ls icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/*.brk | % { icu4c\source\tools\icupkg\x64\Debug\icupkg.exe -tb $_ }
```
That command modified 15 files here, but had no effect prior to this PR (whereas -tl had an effect).

Oddly running it with `-tl` at 12b069b22c41ad9a19d7a92c2a1156c9d388fe79 modified 7 files:
```
C:\Users\robin\Projects\Unicode\icu [𒁾𒃲𒁔𒆷 ≡ +0 ~7 -0 !]> git status
On branch 𒁾𒃲𒁔𒆷
Your branch is up to date with 'origin/𒁾𒃲𒁔𒆷'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line_cj.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line_loose.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line_loose_cj.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line_loose_phrase_cj.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/line_normal.brk
        modified:   icu4j/main/core/src/main/resources/com/ibm/icu/impl/data/icudata/brkitr/word_POSIX.brk
```

What kind of endian was I producing that is neither little nor big (nor big EBCDIC?)

eggrobin:
> What kind of endian was I producing that is neither little nor big (nor big EBCDIC?)

Looking at line.brk after running the tool with -tl, it looks like icupkg pads it to a multiple of 16 bytes with 8 AA bytes.

markusicu:
> > What kind of endian was I producing that is neither little nor big (nor big EBCDIC?)

I don't know. Traditionally, we have used big-endian (-tb) for Java.
ICU4J does not actually care any more, but it's nice to be consistent, and it gives us code coverage for BE.

> Looking at line.brk after running the tool with -tl, it looks like icupkg pads it to a multiple of 16 bytes with 8 AA bytes.

Yeah, because it treats everything as a piece in a .dat file. Each piece is 16-aligned, so that internal offsets are in alignment.

markusicu (approved):

PR #3995, opened 2026-05-22
Fix integer overflow in u_shapeArabic aggregate-tashkeel allocation

dxbjavid:
In u_shapeArabic the aggregate-tashkeel branch computes
2*sourceLength*U_SIZEOF_UCHAR in int arithmetic and hands the result to
uprv_malloc, so sourceLength above INT32_MAX/4 wraps. With a small
enough wrap the subsequent tempsource[j+=step] = source[i] writes
overrun the buffer. Mirror the numeric_limits guard handleGeneratedSpaces
already uses in this file.


dxbjavid:
Hi @markusicu 

Friendly follow-up on this PR. Is there anything else you'd like me to address, or is it currently awaiting review?

Thanks for your time.

PR #3997, opened 2026-05-27
ICU-23408 Fix null pointer dereference in DateTimePatternGenerator copy constructor

UnLucky252:
### Linked Jira issue
ICU-23408

### Summary
In `DateTimePatternGenerator`'s copy constructor, when one of the four `new`
calls (`fp`, `dtMatcher`, `distanceInfo`, `patternMap`) fails, `internalErrorCode`
is set to `U_MEMORY_ALLOCATION_ERROR`, but execution falls through to
`*this = other`, which dereferences those null pointers
(e.g. `*fp = *(other.fp)`, `dtMatcher->copyFrom(...)`).

### Cause
Missing `return;` after the error code assignment.

### Fix
Add `return;` so the constructor exits without touching the null pointers.

### Testing
Existing tests pass. No new test added — the OOM path requires allocator
injection that is not part of the standard test harness.

### Notes

#### Checklist
- [x] Required: Issue filed: ICU-23408

### DIFF
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -332,7 +332,7 @@ DateTimePatternGenerator::createEmptyInstance(UErrorCode& status) {
     return U_SUCCESS(status) ? result.orphan() : nullptr;
 }
 
-DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
+DateTimePatternGenerator::DateTimePatternGenerator() :
     UObject()
 {
     emptyString.getTerminatedBuffer();
@@ -341,7 +341,15 @@ DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
     distanceInfo = new DistanceInfo();
     patternMap = new PatternMap();
     if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
-        internalErrorCode = status = U_MEMORY_ALLOCATION_ERROR;
+        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
+    }
+}
+
+DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
+    DateTimePatternGenerator()
+{
+    if (U_FAILURE(internalErrorCode)) {
+        status = internalErrorCode;
     }
 }
 
@@ -352,24 +360,35 @@ DateTimePatternGenerator::DateTimePatternGenerator(const Locale& locale, UErrorC
 }
 
 DateTimePatternGenerator::DateTimePatternGenerator(const DateTimePatternGenerator& other) :
-    UObject()
+    DateTimePatternGenerator()
 {
-    emptyString.getTerminatedBuffer();
-    fp = new FormatParser();
-    dtMatcher = new DateTimeMatcher();
-    distanceInfo = new DistanceInfo();
-    patternMap = new PatternMap();
-    if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
-        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
-        return;
-    }
-    *this=other;
+    *this = other;
 }
 
 DateTimePatternGenerator&
 DateTimePatternGenerator::operator=(const DateTimePatternGenerator& other) {
     // reflexive case
     if (&other == this) {
+        return *this;
+    }
+    // Heal a previously-broken object: if any of the core helpers are missing
+    // (e.g. from a prior OOM during construction), try to allocate them now.
+    // If (re)allocation still fails, leave internalErrorCode set and bail out
+    // without dereferencing null pointers below.
+    if (fp == nullptr) {
+        fp = new FormatParser();
+    }
+    if (dtMatcher == nullptr) {
+        dtMatcher = new DateTimeMatcher();
+    }
+    if (distanceInfo == nullptr) {
+        distanceInfo = new DistanceInfo();
+    }
+    if (patternMap == nullptr) {
+        patternMap = new PatternMap();
+    }
+    if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
+        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
         return *this;
     }
     internalErrorCode = other.internalErrorCode;
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -361,6 +361,7 @@ DateTimePatternGenerator::DateTimePatternGenerator(const DateTimePatternGenerato
     patternMap = new PatternMap();
     if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
         internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
+        return;
     }
     *this=other;
 }
--- a/icu4c/source/i18n/dtptngen.cpp
+++ b/icu4c/source/i18n/dtptngen.cpp
@@ -348,7 +348,7 @@ DateTimePatternGenerator::DateTimePatternGenerator() :
 DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
     DateTimePatternGenerator()
 {
-    if (U_FAILURE(internalErrorCode)) {
+    if (U_FAILURE(internalErrorCode) && U_SUCCESS(status)) {
         status = internalErrorCode;
     }
 }
@@ -371,6 +371,14 @@ DateTimePatternGenerator::operator=(const DateTimePatternGenerator& other) {
     if (&other == this) {
         return *this;
     }
+    // Don't try to copy a broken source object: bail out cleanly instead of
+    // crashing on `*fp = *(other.fp);` etc., and avoid falsely "healing" by
+    // copying a failed internalErrorCode below.
+    if (U_FAILURE(other.internalErrorCode)) {
+        return *this;
+    }
+
+
     // Heal a previously-broken object: if any of the core helpers are missing
     // (e.g. from a prior OOM during construction), try to allocate them now.
     // If (re)allocation still fails, leave internalErrorCode set and bail out
--- a/icu4c/source/i18n/unicode/dtptngen.h
+++ b/icu4c/source/i18n/unicode/dtptngen.h
@@ -574,6 +574,12 @@ public:
 
 private:
     /**
+     * Default constructor. Allocates the core helper objects; sets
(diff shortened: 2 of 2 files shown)


markusicu:
The finding is correct, but the fix is incomplete.

The copy constructor is effectively the same as the default constructor plus the assignment operator. The first thing we should do is to actually make it so, invoking the default constructor before the assignment, so that we don't duplicate that code. (This class predates ICU requiring C++11 which added delegating constructors.)

The fix is incomplete because the same problem occurs when the default constructor and assignment are used explicitly, so the real fix is to deal with null pointers in the assignment operator.

There we need to decide a kind of policy question: Whether the assignment should bail out if U_FAILURE(internalErrorCode), or whether an assignment should try to "heal" a broken object by attempting to allocate the missing pieces. We do something like the latter in UnicodeString: Assignment and similar operations, where the target is a "bogus" string, replace the bogus state with the new state (e.g., copy of another string).

@aheninger @eggrobin @richgillam WDYT?

UnLucky252:
Thanks for the review!

Agreed — the immediate `return;` only covers the in-place dereference in the
copy constructor and doesn't cover the equivalent path through default-construct
+ explicit `operator=`. The real fix indeed belongs in `operator=`.

I'm happy to update the PR once you and @aheninger / @eggrobin / @richgillam
settle on the preferred policy for `operator=` on a target whose
`internalErrorCode` indicates a previous allocation failure
(bail out on `U_FAILURE(internalErrorCode)` vs. attempt to heal,
as `UnicodeString` does for the bogus state).

If a delegating-constructor refactor of the copy constructor
(`DateTimePatternGenerator(const DateTimePatternGenerator& other) : DateTimePatternGenerator() { *this = other; }`)
is welcome alongside the `operator=` fix, I'll include both in the same revision.

Will wait for direction before pushing changes.

Found via Svace (ISP RAS), happy to follow whatever approach the team prefers.

richgillam:
> There we need to decide a kind of policy question: Whether the assignment should bail out if U_FAILURE(internalErrorCode), or whether an assignment should try to "heal" a broken object by attempting to allocate the missing pieces.
> 
> @aheninger @eggrobin @richgillam WDYT?

If it's not too hard to do, I think I agree that "healing" would be preferable here.  In the copy-constructor case, the "healing" would fail again, so you don't gain anything there, but you don't lose anything either, and you only have to deal with the failed allocation in one place.

UnLucky252:
Pushed the refactor. Two commits on the branch — happy to squash into one
when you're satisfied with the approach (or you can squash-merge).

Changes:
- Added a private no-arg `DateTimePatternGenerator()` (header + impl) that
  performs the core allocations and sets `internalErrorCode` on failure.
- `DateTimePatternGenerator(UErrorCode&)` now delegates to it.
- The copy constructor now delegates and calls `operator=`, removing the
  duplicated allocation code.
- `operator=` now heals before the copy: if any of `fp` / `dtMatcher` /
  `distanceInfo` / `patternMap` are `nullptr`, attempt to (re)allocate. If
  that still fails, set `U_MEMORY_ALLOCATION_ERROR` and return without
  dereferencing. As @richgillam noted, healing in the copy-construction
  path will fail again on persistent OOM but doesn't lose anything.

Let me know if you'd prefer a different placement for the healing block
(e.g. inside a small helper) or if the no-arg constructor should be
documented differently.

markusicu:
We have discussed the "healing" of broken objects in the Technical Committee.
- We have done this (on purpose) so far for only a small set of classes.
- We cannot promise to do it for every class, so we will just document that.
- Otherwise it's ok to leave an object "broken", but it shouldn't crash when the brokenness is hard to discover, for example when there is no UErrorCode indicating that something is wrong.
- However, when it's relatively easy and harmless to "heal" an object on assignment, that's ok.
- See ICU-23419 where I plan to add documentation.

Having said that, this pull request
- does some nice refactoring
- "healing" looks not too onerous in this case
- still has some gaps; I will add comments
- fails on Windows, please take a look

eggrobin:
> fails on Windows, please take a look

The Windows build is independently broken.

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -352,17 +360,9 @@ DateTimePatternGenerator::DateTimePatternGenerator(const Locale& locale, UErrorC
 }
 
 DateTimePatternGenerator::DateTimePatternGenerator(const DateTimePatternGenerator& other) :
-    UObject()
+    DateTimePatternGenerator()
```
markusicu, line 363: FYI: nice example of how delegating to another constructor makes code simpler and less fragile

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -341,7 +341,15 @@ DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
...
     if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
-        internalErrorCode = status = U_MEMORY_ALLOCATION_ERROR;
+        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
+    }
+}
+
+DateTimePatternGenerator::DateTimePatternGenerator(UErrorCode &status) :
+    DateTimePatternGenerator()
+{
+    if (U_FAILURE(internalErrorCode)) {
```
markusicu: We don't want to clobber an existing failure code --
```suggestion
    if (U_FAILURE(internalErrorCode) && U_SUCCESS(status)) {
```

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -371,6 +371,26 @@ DateTimePatternGenerator::operator=(const DateTimePatternGenerator& other) {
     if (&other == this) {
         return *this;
     }
```
markusicu, line 373: Right after this line, I think we need
```
    if (U_FAILURE(other.internalErrorCode)) {
        return *this;
    }
```
so that
- we don't crash below (e.g., `*fp = *(other.fp);`) with the same problem in the source object that you are trying to fix in the target object
- when we copy the other.internalErrorCode we actually heal this object

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -371,6 +371,26 @@ DateTimePatternGenerator::operator=(const DateTimePatternGenerator& other) {
...
+        distanceInfo = new DistanceInfo();
+    }
+    if (patternMap == nullptr) {
+        patternMap = new PatternMap();
+    }
+    if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
+        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
+        return *this;
+    }
     internalErrorCode = other.internalErrorCode;
```
markusicu, line 402: Is the internalErrorCode ever set to something other than U_ZERO_ERROR or U_MEMORY_ALLOCATION_ERROR?
If there is another error condition, then we need to think about whether we can "heal" from that, and we may need to take a closer look at code changing the internalErrorCode.

icu4c/source/i18n/dtptngen.cpp
```diff
@@ -371,6 +371,26 @@ DateTimePatternGenerator::operator=(const DateTimePatternGenerator& other) {
...
+        dtMatcher = new DateTimeMatcher();
+    }
+    if (distanceInfo == nullptr) {
+        distanceInfo = new DistanceInfo();
+    }
+    if (patternMap == nullptr) {
+        patternMap = new PatternMap();
+    }
+    if (fp == nullptr || dtMatcher == nullptr || distanceInfo == nullptr || patternMap == nullptr) {
+        internalErrorCode = U_MEMORY_ALLOCATION_ERROR;
```
markusicu, line 399: FYI: If one of these was a nullptr and still is, then the internalErrorCode should already be set. If the internalErrorCode could hold some other error condition, then changing it could be problematic. See my comment below.

icu4c/source/i18n/dtptngen.cpp
UnLucky252: Done — added `&& U_SUCCESS(status)` so we don't overwrite a caller-set
failure code.

icu4c/source/i18n/dtptngen.cpp
UnLucky252: Thanks!

icu4c/source/i18n/dtptngen.cpp
UnLucky252: Added the early return on `U_FAILURE(other.internalErrorCode)` right
after the reflexive check, with a comment explaining both reasons
(avoid dereference and avoid false "heal").

icu4c/source/i18n/dtptngen.cpp
UnLucky252: Acknowledged — see the answer to your line 394 question below: yes,
internalErrorCode can hold non-OOM codes (from initData propagating
data-loading or locale errors). The heal-fail branch only runs when
one of the four `new`s in the immediate heal block returned nullptr,
so the U_MEMORY_ALLOCATION_ERROR write is correct for that branch.
If we want to be conservative against future code paths that might
nullptr one of these pointers under a non-OOM condition, I can guard
the heal-fail write with `if (U_SUCCESS(internalErrorCode))`. Happy
to do that if you prefer.

icu4c/source/i18n/dtptngen.cpp
UnLucky252: Yes — `internalErrorCode` is set via three paths in this file:

1. `U_MEMORY_ALLOCATION_ERROR` in the no-arg constructor (line 344)
   and in `operator=` when `skipMatcher` allocation fails (line ~396).
2. `other.internalErrorCode` in `operator=` (line ~374) — propagates
   the source's state, which can be any value.
3. `internalErrorCode = status;` in `initData` (line ~511) — `status`
   here can be any error returned by locale/data loading
   (`U_MISSING_RESOURCE_ERROR`, `U_INVALID_FORMAT_ERROR`, etc.), not
   only OOM.

So yes, non-OOM error codes can land in `internalErrorCode`. The heal
logic itself only re-allocates the four pointers (which only become
nullptr through OOM), so the heal-failure write to
`U_MEMORY_ALLOCATION_ERROR` is reached only on a genuine OOM in the
heal block.

The remaining question is the semantics of "successful heal" via
`operator=` over a non-OOM-failed target: today, after the heal block
proceeds, `internalErrorCode = other.internalErrorCode;` (line ~374)
replaces the target's state with the source's state — effectively
healing from any prior error, not just OOM. If you'd prefer the heal
to be OOM-only (i.e., refuse to heal if the target's
`internalErrorCode` is non-OOM), I can gate the heal block with a
check on the type of the existing error. Let me know which semantics
the TC prefers.

UnLucky252:
Pushed the two requested fixes (don't clobber status; bail out on broken
source) as a third commit on this branch. Replied inline to each of the
comments.

Re Windows: thanks @eggrobin for confirming that's independent of this PR.

PR #4004, opened 2026-05-27
reject RBNF plural rules whose divisor overflows to zero

rootvector2:
A slash divisor descriptor with a `$(cardinal,...)$` body sets `rulePatternFormat` but creates no modulus or integral-part substitution, so the divisor checks in those substitution constructors never run. When `radix^exponent` overflows 64 bits the power helper wraps to 0 (`util64_pow` in ICU4C, `NFRule.power` in ICU4J), and `doFormat()` then divides the formatted number by that zero divisor: integer divide by zero, SIGFPE on x86-64 in C, `ArithmeticException: / by zero` in Java.

`extractSubstitutions` now rejects the rule as soon as the plural format is built and `getDivisor() == 0`, mirroring the existing zero-divisor guards in the modulus and integral-part substitutions. ICU4C surfaces `U_PARSE_ERROR`; ICU4J throws `IllegalStateException`, matching how those substitutions already reject a zero divisor.

Confirmed the same crash exists in ICU4J (the Number Format Tester / `numbers.jsp` throws `/ by zero`), so the fix and the test are mirrored to both. The C++ `TestDividedByZero` and the new Java `TestDivisorOverflow` both fail without the fix and pass with it.

#### Checklist
- [ ] Required: The PR title must be prefixed with a JIRA Issue number. Example: "ICU-NNNNN Fix xyz"

### DIFF
--- a/icu4c/source/i18n/nfrule.cpp
+++ b/icu4c/source/i18n/nfrule.cpp
@@ -504,6 +504,13 @@ NFRule::extractSubstitutions(const NFRuleSet* owner,
         }
         rulePatternFormat = formatter->createPluralFormat(pluralType,
                 ruleText.tempSubString(endType + 1, pluralRuleEnd - endType - 1), status);
+        if (U_SUCCESS(status) && getDivisor() <= 0) {
+            // radix^exponent overflowed 64 bits, so doFormat() would divide the
+            // number by a zero (or wrapped-negative) divisor. Reject the rule, like
+            // the modulus and integral-part substitutions already do for a zero divisor.
+            status = U_PARSE_ERROR;
+            return;
+        }
     }
 }
 
--- a/icu4c/source/test/intltest/itrbnf.cpp
+++ b/icu4c/source/test/intltest/itrbnf.cpp
@@ -2834,6 +2834,15 @@ IntlTestRBNF::TestDividedByZero() {
     UErrorCode status = U_ZERO_ERROR;
     RuleBasedNumberFormat rbnf(u"7060920374060940374/4:[]", Locale::getUS(), perror, status);
     assertEquals("base is too large", U_NUMBER_ARG_OUTOFBOUNDS_ERROR, status);
+
+    // A $(cardinal,...)$ rule whose radix^exponent overflows 64 bits: the
+    // divisor is zero, so doFormat() would divide the number by it. The rule
+    // must be rejected at construction instead of dividing by zero.
+    status = U_ZERO_ERROR;
+    RuleBasedNumberFormat pluralDivisor(
+        u"0: zero;\n4611686018427387904/16: big $(cardinal,one{a}other{b})$;",
+        Locale::getEnglish(), perror, status);
+    assertEquals("zero divisor from overflow", U_PARSE_ERROR, status);
 }
 
 void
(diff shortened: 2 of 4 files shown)


markusicu:
- please restore and fill out the pull request template
- please create a Jira ticket with the problem description
- @grhoten @richgillam FYI

markusicu:
Please also check if this is a problem in Java -- ideally by porting the unit test.

grhoten (changes requested):
I confirmed that the [Number Format Tester](https://st.unicode.org/cldr-apps/numbers.jsp) also throws a "/ by zero" exception, which uses ICU4J. So the change needs to be mirrored to both.

These changes are just modifying how the failure is surfaced. I find it unlikely that this issue would have surfaced through typical usage of RBNF in any language. I think only some French variants use the divisor like this, and it's only for numbers 60-99. While it's a valid issue, it's unlikely to be encountered outside of synthetic rule making with large numbers.

Please follow the directions provided by Markus to get this change mergeable.

rootvector2:
Confirmed it's the same bug in ICU4J: `NFRule.power` wraps to 0 on that overflow and `doFormat` divides by it, which is the `/ by zero` from the Number Format Tester. Mirrored the guard in `NFRule.extractSubstitutions` (throws `IllegalStateException`, like the substitution zero-divisor checks already do) and ported the test as `RbnfTest.TestDivisorOverflow`. It fails without the fix and passes with it.

Agree this only changes how the failure surfaces, and you'd really only hit it with synthetic large-divisor rules rather than real RBNF data. Rejecting at construction just matches the existing zero-divisor guards instead of dividing by zero at format time.

Restored and filled out the PR template. On the Jira ticket: I'm not on the team so I can't file in the tracker directly. I can submit it through the bug page, or if it's easier for you to open one I'll prefix the title and commits with the number.

rootvector2:
Squashed to a single commit. Both bot checks stay red until the title has a ticket ID (the `single-commit` check doesn't count commits before that), so once there's an ICU number I'll prefix the title and the commit. Question above about filing the ticket still stands.

icu4c/source/i18n/nfrule.cpp
```diff
@@ -504,6 +504,13 @@ NFRule::extractSubstitutions(const NFRuleSet* owner,
         }
         rulePatternFormat = formatter->createPluralFormat(pluralType,
                 ruleText.tempSubString(endType + 1, pluralRuleEnd - endType - 1), status);
+        if (U_SUCCESS(status) && getDivisor() == 0) {
```
grhoten: This should probably be <= 0 in both implementations. A divisor should be a positive number. Bad things would probably happen if it were negative.

rootvector2: Makes sense, changed both to `<= 0`. Depending on radix the wrap can land on a negative value instead of 0, so rejecting anything non-positive is right. `TestDividedByZero` and `TestDivisorOverflow` still pass.

rootvector2:
any update?

grhoten (approved):
The latest changes look fine. Please follow the instructions when you created this pull request to get the remaining checks passing.

rootvector2:
Will do, filing the ticket myself. Once I have the ICU number I'll prefix the title and the commit so both checks go green.

PR #4005, opened 2026-05-27
ICU-23403 make number.Precision immutable

markusicu:
In the code review for
- https://github.com/unicode-org/icu/pull/3978

@sffc commented
- This is not intended to be a mutable object.
- You don't need to look far to see that the field is designed to be private and immutable. I don't remember the reason they are package-private instead of private, and why they are not final.

... so let's make Precision final.
- final fields
- override in copy constructors, not after construction
- making the fields final and adding getters also seemed good

#### Checklist
- [x] Required: Issue filed: ICU-23403

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
@@ -24,12 +24,50 @@ import java.math.MathContext;
  * @see NumberFormatter
  */
 public abstract class Precision {
+    /** For copy-constructing a Precision with overrides. */
+    enum OverrideTZD {
+        AUTO(TrailingZeroDisplay.AUTO),
+        HIDE_IF_WHOLE(TrailingZeroDisplay.HIDE_IF_WHOLE),
+        NULL(null);
 
-    /* package-private final */ MathContext mathContext;
-    /* package-private final */ TrailingZeroDisplay trailingZeroDisplay;
+        private final TrailingZeroDisplay tzd;
+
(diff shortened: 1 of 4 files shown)


markusicu:
@Manishearth FYI

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
```diff
@@ -24,12 +24,50 @@
  * @see NumberFormatter
  */
 public abstract class Precision {
+    /** For copy-constructing a Precision with overrides. */
+    enum OverrideTZD {
```
sffc, line 28: In #4002, the existing mathContext and trailingZeroDisplay are just always passed into the createCopy method, so it doesn't need OverrideTZD, which I assume you made to distinguish `null` as in "do not override" from `null` as in "a null TZD".

sffc (commented):
I made a similar PR, #4002. (a few hours older than this one, #4005.) It is similar but doesn't need the OverrideTZD thing, and also fixes ScientificNotation that has a similar issue with final-not-final fields.

icu4j/main/core/src/main/java/com/ibm/icu/number/Precision.java
markusicu: That's right. I looked at code cloning some Precision and then overriding one of its two core properties but not both.
I will take a look at your PR.

markusicu:
closing in favor of
- https://github.com/unicode-org/icu/pull/4002

PR #4006, opened 2026-05-28
ICU-23413 Fix null pointer dereference in RuleBasedNumberFormat::getRuleSetDisplayName

UnLucky252:
### Linked Jira issue
ICU-23413

### Summary
In `RuleBasedNumberFormat::getRuleSetDisplayName` (`icu4c/source/i18n/rbnf.cpp`),
the result of `localeName.getBuffer(len + 1)` was indexed via
`localeStr[len] = 0` in the following loop without a NULL check.

### Cause
`UnicodeString::getBuffer(int32_t)` returns NULL on allocation failure
(or when the string is bogus); the indexed write is then undefined
behavior.

### Fix
Return a bogus `UnicodeString` when `getBuffer` returns NULL.

### Testing

### Notes

#### Checklist
- [x] Required: Issue filed: ICU-23413

### DIFF
--- a/icu4c/source/i18n/rbnf.cpp
+++ b/icu4c/source/i18n/rbnf.cpp
@@ -943,6 +943,11 @@ RuleBasedNumberFormat::getRuleSetDisplayName(int32_t index, const Locale& locale
         UnicodeString localeName(localeParam.getBaseName(), -1, UnicodeString::kInvariant); 
         int32_t len = localeName.length();
         char16_t* localeStr = localeName.getBuffer(len + 1);
+        if (localeStr == nullptr) {
+            UnicodeString bogus;
+            bogus.setToBogus();
+            return bogus;
+        }
         while (len >= 0) {
             localeStr[len] = 0;
             int32_t ix = localizations->indexForLocale(localeStr);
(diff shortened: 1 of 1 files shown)


grhoten (commented):
Where's the test? Also wouldn't it be easier to reuse the bogus setting at the end of the function?

I'm not a fan of untestable code. So if there is a way to reuse an existing error condition, that would be better. There is  duplicate error state code in this function. Even if this condition can be triggered with only with OOM, reuse of error handling code is best.

grhoten:
The ruleset display names are not tested that well. It's unused functionality by the default rulesets. So even if the tests pass, I wouldn't assume that passing tests means that it's actually tested.

UnLucky252:
> Where's the test? Also wouldn't it be easier to reuse the bogus setting at the end of the function?
> 
> I'm not a fan of untestable code. So if there is a way to reuse an existing error condition, that would be better. There is duplicate error state code in this function. Even if this condition can be triggered with only with OOM, reuse of error handling code is best.

Refactored per your suggestion - on `getBuffer` failure we now fall
through to the existing bogus-return at the end of the function
instead of adding a second one. The bogus-return path is exercised
by existing tests via the `localizations == nullptr` / index-
out-of-range conditions, so no OOM-specific test is needed.

UnLucky252:
> The ruleset display names are not tested that well. It's unused functionality by the default rulesets. So even if the tests pass, I wouldn't assume that passing tests means that it's actually tested.

Pushed the refactor. Two commits on the branch - happy to squash
when you're satisfied (or you can squash-merge).

grhoten (approved):
This change is acceptable. Please address the issues in the other checks.

PR #4007, opened 2026-05-28
ICU-23414 Fix null pointer dereference in TransliterationRule copy constructor on allocation failure

UnLucky252:
### Linked Jira issue
ICU-23414

### Summary
In the `TransliterationRule` copy constructor (`icu4c/source/i18n/rbt_rule.cpp`),
the result of `uprv_malloc` for the `segments` array was passed to
`uprv_memcpy` without a NULL check.

### Cause
On OOM `uprv_malloc` returns NULL; the subsequent `uprv_memcpy` is then
undefined behavior.

### Fix
The constructor has no `UErrorCode`, so guard the `uprv_memcpy` with a
NULL check. On failure `segments` stays `nullptr`, which the rest of the
object already tolerates.

### Testing

### Notes

#### Checklist
- [x] Required: Issue filed: ICU-23414

### DIFF
--- a/icu4c/source/i18n/rbt_rule.cpp
+++ b/icu4c/source/i18n/rbt_rule.cpp
@@ -167,7 +167,9 @@ TransliterationRule::TransliterationRule(TransliterationRule& other) :
         data(other.data) {
     if (other.segmentsCount > 0) {
         segments = static_cast<UnicodeFunctor**>(uprv_malloc(other.segmentsCount * sizeof(UnicodeFunctor*)));
-        uprv_memcpy(segments, other.segments, (size_t)other.segmentsCount*sizeof(segments[0]));
+        if (segments != nullptr) {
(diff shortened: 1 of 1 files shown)


richgillam:
We looked at this in ICU TC.  If there's an OOM problem that causes allocation of `segments` to fail, this change will leave the object in a partially-initialized state, which seems likely to lead to other, harder-to-debug problems further down the road.  To deal with the allocation failure properly, we'd have to add a "bogus" flag and a bunch of extra checking.  Is any of that really worth it?

UnLucky252:
Understood - closing per the Won't Fix resolution on ICU-23414.
Agreed that a silent nullptr result in a copy constructor without
UErrorCode creates a partially-initialized object that callers
can't detect, and that a proper fix would require a broader
"bogus" flag refactor. Will mark the Svace finding as
Won't Fix / ICU design decision.

Thanks for the review!

PR #4008, opened 2026-05-28
ICU-23415 Fix null pointer dereference in RelativeDateFormat copy constructor on allocation failure

UnLucky252:
### Linked Jira issue
ICU-23415

### Summary
In the `RelativeDateFormat` copy constructor (`icu4c/source/i18n/reldtfmt.cpp`),
the result of `uprv_malloc` for `fDates` was passed to `uprv_memcpy`
without a NULL check.

### Cause
On OOM `uprv_malloc` returns NULL; the subsequent `uprv_memcpy` is then
undefined behavior.

### Fix
The constructor has no `UErrorCode`, so guard the `uprv_memcpy` with a
NULL check and reset `fDatesLen = 0` on failure so the object stays
internally consistent.

### Testing

### Notes

#### Checklist
- [x] Required: Issue filed: ICU-23415

### DIFF
--- a/icu4c/source/i18n/reldtfmt.cpp
+++ b/icu4c/source/i18n/reldtfmt.cpp
@@ -58,7 +58,11 @@ RelativeDateFormat::RelativeDateFormat(const RelativeDateFormat& other) :
     }
     if (fDatesLen > 0) {
         fDates = static_cast<URelativeString*>(uprv_malloc(sizeof(fDates[0]) * static_cast<size_t>(fDatesLen)));
-        uprv_memcpy(fDates, other.fDates, sizeof(fDates[0])*(size_t)fDatesLen);
+        if (fDates != nullptr) {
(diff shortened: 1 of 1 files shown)


richgillam:
Inspected in ICU TC.  Same issue as with ICU-21414: If the allocation fails and we just skip filling in the value, the object ends up in a partially-initialized state that'll just cause problems down the road.  To fix this right would require adding a "bogus" flag and all the machinery that comes with it.  Is that really worth it?  Will we see this error in practice, and why is something other than the program crashing the right answer?

UnLucky252:
Understood - closing per the Won't Fix resolution on ICU-23415.
Agreed that a silent nullptr result in a copy constructor without
UErrorCode creates a partially-initialized object that callers
can't detect, and that a proper fix would require a broader
"bogus" flag refactor. Will mark the Svace finding as
Won't Fix / ICU design decision.

Thanks for the review!

PR #4010, opened 2026-05-28
ICU-23364 Add support of the new dayOfMonths CLDR resources (and the "ddd" pattern symbol) to ICU

richgillam:
This PR adds support in ICU for the new `daysOfMonths` resources introduced to CLDR by https://unicode-org.atlassian.net/browse/CLDR-7408.  It includes:

- Temporary local ICU-data additions for the new resources added in CLDR (they're temporary until we get the "real" ones copied over in a future CLDR integration).
- An update to the CLDR-to-ICU tool to bring the new CLDR resources over into ICU.
- New API on DateFormatSymbols for the new resources (API proposal forthcoming).
- Code to properly handle the new `ddd` pattern/skeleton sequence in `SimpleDateFormat` (both formatting and parsing) and `DateTimePatternGenerator`.
- Unit tests for all of the above.

This PR is large; I split it into 10 separate commits to make it easier to review.

A few issues worth calling out:
- As I mentioned above, I manually brought over the changes to the ICU data that correspond to the new changes in CLDR.  The changes were generated by the CLDR-to-ICU tool with my change in place.  I'm expecting these changes to be stomped over by the next official CLDR integration (which hopefully won't actually change the `dayOfMonths` stuff), but this lets us test the implementation in advance of the next CLDR integration landing.
- The temporary ICU-data changes on the Java side required regenerating ALL of the Java `.res` files, and I think I may have accidentally picked up some stray changes we don't want.  I think we lost a few files we want to keep, and there are a couple stray unit-test changes unrelated to the work of this PR that I think happened because we picked up extraneous changes in the Java ICU data.  I'm assuming (hoping?) this will all get straightened out in the next CLDR integration.
- There is no testing included for the `<dayOfMonth type="x">` resources right now because nothing in the CLDR change uses them.  I _have_ tested these resources in Apple ICU on changes that aren't in CLDR-7408.  I made two CLDR changes in Apple CLDR that we don't have in open-source CLDR: 1) Support for ordinals (i.e., just "1er") in French, and 2) replacement of the "hanidays" numbering system with equivalent `dayOfMonths` resources.  I can file tickets to add these things to CLDR and add appropriate unit tests for them to ICU.  (As I said, I have all that in Apple ICU, so the relevant code here _has_ been tested.)

#### Checklist
- [x] Required: Issue filed: ICU-23364


richgillam:
**NOTE TO SELF:** In addition to the current failing tests in Java (which seem to be because I screwed up the Java data generation process and accidentally deleted some key files), I'm also getting reports that the Java code fails on skeletons that include `dd` (such as `yMMMMdd` and `yMMMdd`).  The C++ code apparently works.  Claude claimed I didn't need to make any `DateTimePatternGenerator` changes on the Java side, but this is apparently wrong and needs a closer look, plus we need to add the `dd` tests in both C++ and Java.

richgillam:
> **NOTE TO SELF:** In addition to the current failing tests in Java (which seem to be because I screwed up the Java data generation process and accidentally deleted some key files), I'm also getting reports that the Java code fails on skeletons that include `dd` (such as `yMMMMdd` and `yMMMdd`). The C++ code apparently works. Claude claimed I didn't need to make any `DateTimePatternGenerator` changes on the Java side, but this is apparently wrong and needs a closer look, plus we need to add the `dd` tests in both C++ and Java.

I think I've fixed this now...

macchiati (commented):
Java looks good; my C++ skills are too rusty for that part.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1253,6 +1266,41 @@ public void setZodiacNames(String[] zodiacNames, int context, int width) {
...
+     * com.ibm.icu.impl.StandardPlural#COUNT}. An empty entry means no data for that plural
+     * category. Currently only FORMAT context and ABBREVIATED width are supported; other
+     * combinations return null.
+     *
+     * @param context The formatting context. Currently only FORMAT is supported.
+     * @param width The width. Currently only ABBREVIATED is supported.
+     * @return the day-of-month ordinal name strings, or null if not available.
+     */
+    String[] getDayOfMonthOrdinalNames(int context, int width) {
+        if (context != FORMAT || width != ABBREVIATED) {
```
mihnita: Should we fallback to `dayOfMonthOrdinalNames` for contexts other than FORMAT and ABBREVIATED?
I tried checking what we do for other such values, and I think that we just ignore context / width.
See `getYearNames`, `getZodiacNames` (the only other methods that take context / width and ignore them)

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
```diff
@@ -1253,6 +1266,41 @@ public void setZodiacNames(String[] zodiacNames, int context, int width) {
...
+     * override for that day number. Currently only FORMAT context and ABBREVIATED width are
+     * supported; other combinations return null.
+     *
+     * @param context The formatting context. Currently only FORMAT is supported.
+     * @param width The width. Currently only ABBREVIATED is supported.
+     * @return the day-of-month cardinal strings, or null if not available.
+     */
+    String[] getDayOfMonthCardinalNames(int context, int width) {
+        if (context != FORMAT || width != ABBREVIATED) {
+            return null;
```
mihnita: Same comment as above.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -5172,7 +5201,124 @@ private void parsePattern() {
                 if (ch == 's') {
                     hasSecond = true;
                 }
+                if (ch == 'd') {
+                    dCount++;
+                    if (dCount >= 3 && ordinalPluralRules == null) {
```
mihnita, line 5206: Nitpick: do we want to do == 3, instead of >= ?
I know we don't have `ddd` right now in TR-35, but you never know.
If we do that later, changing an existing `dddd` behavior is harder.

If you think that it's better to make it exactly 3 then the loop should count the `d` in `dCount`, and the `dCount == 3` should be outside the loop.

---

I tried to see what we do for other "over long sequences" (think `EEEEEEEE` or `aaaaaaaaa`).
And in those cases we either fallback to some kind of best default (`EEEEEEEE` => "Thu", `aaaaaaaaa` => "PM"), or, for numeric values, we pad with zeros (`ddddd` => "00009").

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -5172,7 +5201,124 @@ private void parsePattern() {
...
+     * the caller should fall back to numeric formatting.
+     */
+    private String formatDayOfMonthName(
+            int value, NumberFormat currentNumberFormat, int count, int maxIntCount) {
+        // First check for a cardinal element (e.g., Chinese hanidays "\u521D\u4E00",
+        // "\u5341\u4E94").
+        String[] cardinalNames =
+                formatData.getDayOfMonthCardinalNames(
+                        DateFormatSymbols.FORMAT, DateFormatSymbols.ABBREVIATED);
+        if (cardinalNames != null
```
mihnita, line 5231: Nitpick: check for value <= 0.
I don't know if it is possible, since this is a private method.
Maybe it is handled already.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -5172,7 +5201,124 @@ private void parsePattern() {
...
+                        DateFormatSymbols.FORMAT, DateFormatSymbols.ABBREVIATED);
+        if (cardinalNames != null) {
+            int bestLen = 0;
+            int bestDay = -1;
+            for (int day = 1; day < cardinalNames.length; day++) {
+                String name = cardinalNames[day];
+                if (name == null || name.isEmpty()) {
+                    continue;
+                }
+                if (text.regionMatches(start, name, 0, name.length()) && name.length() > bestLen) {
```
mihnita: Nitpick (performance): do the `name.length() > bestLen` test first.
It's fast, only compares 2 integers, and if false we don't call the `regionMatches` at all (which is very likely slower).

I doubt that anyone will parse many-many such dates, but it is free performance.

mihnita: Should we call `regionMatches(boolean ignoreCase = true, ...`?
Maybe in English 1ST, 2ND ... is rare, but who knows in other languages.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
```diff
@@ -5172,7 +5201,124 @@ private void parsePattern() {
...
+            if (pat == null || pat.isEmpty()) {
+                continue;
+            }
+            int placeholderPos = pat.indexOf("{0}");
+            if (placeholderPos < 0) {
+                continue;
+            }
+            String prefix = pat.substring(0, placeholderPos);
+            String suffix = pat.substring(placeholderPos + 3);
+            if (!text.regionMatches(start, prefix, 0, prefix.length())) {
```
mihnita: Same question: do we want this case insensitive?

mihnita (commented):
Done with the Java review.

No complaints about anything Java, all I had were some nitpicks that (if you agree) will also affect the C++ code.
Because they are about behavior, not code.

Thank you,
Mihai

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
richgillam: That's an interesting point.  Right now, we only have `FORMAT` and `ABBREVIATED`, but we wanted the parameters in case we discovered language/calendar combinations that need other lengths or contexts (I'm pretty sure we'll eventually want two lengths for the day names in the Hindu calendars, for example).

I wasn't aware that there was a precedent for taking these parameters when they didn't do anything.  If `getYearNames` and `getZodiacNames` take these parameters and just ignore them rather than signaling an error, maybe we should do that here too and follow the precedent.  @macchiati , do you agree?

If no one objects in the next day or two, I'll go ahead and make this change.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
richgillam: As I mentioned above, I wanted to leave the door open for `dddd` and maybe even `ddddd`.  From our experience with the Hindu calendars here at Apple, I think there's a good chance we're going to want these at some point.  I'm hoping the current behavior of the code is to just have longer sequences of `d` behave the same as `ddd`, and it sounds like you're saying that's what we're doing, both here and in analogous situations.  (We want the behavior here to be kind of like `M`, where sequences of three or more `d`s always get treated as names rather than numbers.)

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
richgillam: Fair enough.  I'll make the change.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
richgillam: Good point.  I'll make the change.

richgillam: What do we do with month names?  Are they also case insensitive?  I think we should be consistent with analogous existing behavior.  (And no, I didn't think about this.)  I'll try to check this and make the change if needed.

icu4j/main/core/src/main/java/com/ibm/icu/text/DateFormatSymbols.java
macchiati: Yes. We ignore options that don't apply (yet, or never)

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
macchiati: Yes. That is the desired behavior, to fall back like the M do

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
macchiati: I can't remember about parsing, but suspect it is case insensitive

yumaoka (dismissed):
Reviewed ICU4C changes and looks good to me.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
mihnita: > I wanted to leave the door open for dddd and maybe even ddddd.

I'm all for that :-), in fact that's what I was trying to say, but I guess I was unclear.

This is the current behavior (with patterns, not skeleton, to avoid the whole skeleton-to-pattern magic):

```
MMMM d, y       => July 8, 2026
MMMM dd, y      => July 08, 2026
MMMM ddd, y     => July 008, 2026      !!! Going beyond what's defined
MMMM dddd, y    => July 0008, 2026     !!! Going beyond what's defined
MMMM ddddd, y   => July 00008, 2026    !!! Going beyond what's defined
```

What I would expect with this change:
```
MMMM d, y       => July 8, 2026
MMMM dd, y      => July 08, 2026
MMMM ddd, y     => July 8th, 2026
MMMM dddd, y    => July 0008, 2026     !!! Same as before this PR
MMMM ddddd, y   => July 00008, 2026    !!! Same as before this PR
```

I wanted to make sure that we don't end up with ordinal forms for d count >= 3.
So no:
```
MMMM dddd, y    => July 8th, 2026
or
MMMM dddd, y    => July 0008th, 2026
```

---

This is the behavior of other patterns when we go over the current limit
(they usually don't continue with the last):
```
E, MMMM d, y         => Wed, July 8, 2026
EE, MMMM d, y        => Wed, July 8, 2026
EEE, MMMM d, y       => Wed, July 8, 2026
EEEE, MMMM d, y      => Wednesday, July 8, 2026
EEEEE, MMMM d, y     => W, July 8, 2026 =>
EEEEEE, MMMM d, y    => We, July 8, 2026
EEEEEEE, MMMM d, y   => Wed, July 8, 2026  !!! Goes back to EEE, not EEEEEE

MMMM d, y, h:m a     => July 8, 2026, 9:10 AM
MMMM d, y, h:m aa    => July 8, 2026, 9:10 AM
MMMM d, y, h:m aaa   => July 8, 2026, 9:10 AM
MMMM d, y, h:m aaaa  => July 8, 2026, 9:10 AM
MMMM d, y, h:m aaaaa => July 8, 2026, 9:10 a
MMMM d, y, h:m aaaaaa => July 8, 2026, 9:10 AM  !!! Does not continue like aaaaa

MMMM d, y Q          => July 8, 2026 3
MMMM d, y QQ         => July 8, 2026 03
MMMM d, y QQQ        => July 8, 2026 Q3
MMMM d, y QQQQ       => July 8, 2026 3rd quarter
MMMM d, y QQQQQ      => July 8, 2026 3  !!! Goes back to Q, not QQQQ


M d, y               => 7 8, 2026
MM d, y              => 07 8, 2026
MMM d, y             => Jul 8, 2026
MMMM d, y            => July 8, 2026
MMMMM d, y           => J 8, 2026
MMMMMM d, y          => 000007 8, 2026  !!! Does not continue like MMMMM ("J")
```

So what I'm trying to say is that "dddd" (not currently defined) should give the same result after this PR as it did before (0008):

```java
Date testDate = new Date(2026 - 1900, 6, 8, 9, 10, 24);
DateFormat df = new SimpleDateFormat("MMMM dddd, y", ULocale.US);
assertEquals("July 0008, 2026", df.format(testDate));
```

mihnita: Note: maybe it does. I was not able to easily figure it out from the code, and I did not bother to patch the whole branch into my fork to try it.

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
mihnita: I tested the parsing, it is case insensitive (at least for month name, day of week, am/pm, timezone)

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
richgillam: >So what I'm trying to say is that "dddd" (not currently defined) should give the same result after this PR as it did before (0008)

I disagree.  I think it should give the same result as "ddd".  Having it do the same thing as it did before would be much hard to do and strikes me as nonsensical.  Why would we want that, especially if we're not doing that for "ddd"?

richgillam: To be clear, I'm expecting "MMMM dddd, yyyy" to give us "July 8th, 2026" just like "MMMM ddd, yyyy" does.

macchiati: Agreed. MMMM and MMMM are the closest analogs. A textual form needs to fall back to another textual form

We need to make this clear in cldr (and also IMO reserve the longer forms for the future)

mihnita: I only tried to see what other patterns do, to see if this is consistent.
I defer to Mark / you.

mihnita:
Note: in "my lingo" nitpick means _"if you think it's a good idea and you like it, do it, if not, feel free to ignore it without a need to explain why"_

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
richgillam: I reread this thread, and I have to say, if `MMMMMM` gives us "000007" and not "July", and `QQQQQ` gives us "3" instead of "3rd quarter", that feels very much like a bug in either ICU, or the spec, or both.  And I'm guessing the spec is silent on this point and it's a bug in ICU.  Would you mind filing a ticket?

richgillam:
@mihnita , @yumaoka , @macchiati Can at least one of you take another look and rubber-stamp this?

mihnita (approved):
Thank you,
M

icu4j/main/core/src/main/java/com/ibm/icu/text/SimpleDateFormat.java
mihnita: I've created [CLDR-19668](https://unicode-org.atlassian.net/browse/CLDR-19668) _"UTS `#`35 Date Format Patterns does not specify what happens for extra-long patterns"_

PR #4014, opened 2026-06-02
ICU-23421 code point iterators & ranges: add base(), front(), back()

markusicu:
This is a small addition to the [C++ code point iterator APIs](https://unicode-org.github.io/icu/userguide/strings/cpp-code-point-iterator.html). These are actually common functions in C++ standard iterators and ranges that we had overlooked.

From the ticket:
C++ “ranges” or “containers” usually have a front() function for the first element, and a back() function for the last one. This has turned out to be something that would be a useful addition to [Unsafe]UTFStringCodePoints. Simple example:
- https://en.cppreference.com/cpp/string/basic_string_view
- https://en.cppreference.com/cpp/string/basic_string_view/front
- https://en.cppreference.com/cpp/string/basic_string_view/back

A C++ iterator often wraps another, “more basic” iterator, and commonly provides a base() function to access that. [Unsafe]UTFIterator wraps a code unit iterator; base() would return a code unit iterator at the current logical position. It would be like fetching the current CodeUnits and calling their begin(), except that those cannot be fetched from an exclusive-end() iterator; base() would be much easier to use.

Real example: use std::advance() to move a code point iterator forward n code points, then call its base() to find the position in the input string/string_view, which will usually have moved by a larger number of code units.
- [std::ranges::advance - cppreference.com](https://en.cppreference.com/cpp/iterator/ranges/advance)

Similarly, a reverse_iterator usually has a base() function to return an original-direction iterator at the same logical position. The reverse_iterator wrappers over [Unsafe]UTFIterator should have that, too.
- https://en.cppreference.com/cpp/iterator/reverse_iterator/base

New API signatures in common/unicode/utfiterator.h. Note that the reverse_iterator classes are template specializations of C++ reverse_iterator and do not have the usual API docs.

#### Checklist
- [x] Required: Issue filed: ICU-23421

### DIFF
--- a/icu4c/source/common/unicode/utfiterator.h
+++ b/icu4c/source/common/unicode/utfiterator.h
@@ -1198,7 +1198,7 @@ public:
      * @draft ICU 78
      */
     U_FORCE_INLINE bool operator==(const UTFIterator &other) const {
-        return getLogicalPosition() == other.getLogicalPosition();
+        return base() == other.base();
     }
     /**
      * @param other Another iterator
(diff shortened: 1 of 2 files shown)


markusicu:
@echeran @richgillam @roubert FYI

eggrobin (approved):

PR #4016, opened 2026-06-02
ICU-22148 deprecate ICUUncheckedIOException, now is-an UncheckedIOException

markusicu:
... as agreed in the ICU-TC meeting on 2026-apr-02, with extra work to handle Java serialization.
- deprecate ICUUncheckedIOException
- change its base class from RuntimeException to UncheckedIOException
- this requires a non-null IOException as the cause, so adjust the constructors to make one if we don't already have one
- deserializing an object from ICU 78 or earlier:
    - Java inserts the new base class, but the cause can be null, or can be a Throwable that is not an IOException;
      problem: UncheckedIOException promises a non-null getCause(), and its getCause() casts the internal cause to an IOException;
      without intervention, a deserialized object could have its getCause() return null or throw a ClassCastException
    - add a new instance field for distinguishing the class format version
    - add a readResolve() method which deals with old objects,
      adding a cause if there is none, or
      replacing the deserialized object with a whole new one if there is an incompatible cause
      (because initCause() cannot be called if there is already a cause)
    - adjust the serialization test expectation: the result is functional, but not equal to a freshly constructed object
    - the tests do exercise objects with/without message, with/without cause

Don't change existing code that throws an ICUUncheckedIOException so that callers that catch that continue to work.

#### Checklist
- [x] Required: Issue filed: ICU-22148

### DIFF
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/serializable/CompatibilityTest.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/serializable/CompatibilityTest.java
@@ -185,7 +185,7 @@ public class CompatibilityTest extends CoreTestFmwk {
             jarFile = conn.getJarFile();
             Enumeration<JarEntry> entries = jarFile.entries();
             while (entries.hasMoreElements()) {
-                JarEntry entry = (JarEntry) entries.nextElement();
+                JarEntry entry = entries.nextElement();
                 if (!entry.isDirectory()) {
                     String entryName = entry.getName();
 
--- a/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/serializable/SerializableTestUtility.java
+++ b/icu4j/main/common_tests/src/test/java/com/ibm/icu/dev/test/serializable/SerializableTestUtility.java
@@ -82,7 +82,7 @@ public class SerializableTestUtility {
     }
 
     public static Handler getHandler(String className) {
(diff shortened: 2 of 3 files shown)


gvictor:
>  do I need to worry about serialization?

For compatibility with the older version of the object layout, if it doesn't cause a crash, but just a behavior change, it should be fine on Android. If it causes a runtime crash, it will be very helpful to let us know or put a small note in the ICU release note.

Looking at the change, I think that `ICUUncheckedIOException` should normally have a cause, and no runtime crash is expected. so it should be fine.

cc @vladcioricagoogle in case that he has other opinions

mihnita (approved):
Thank you!
M

markusicu:
Hi @gvictor

> > do I need to worry about serialization?
> 
> For compatibility with the older version of the object layout, if it doesn't cause a crash, but just a behavior change, it should be fine on Android. [...] `ICUUncheckedIOException` should normally have a cause, and no runtime crash is expected. so it should be fine.

You replied to my original question. Since then, I found a serialization test failure, debugged it, documented it, and fixed it (with a bespoke `readResolve()`). Please see the old narrative in the PR description, and the final writeup at the top. In short, deserializing an old object creates a functional new object, it's just not `equals` to a new object created with the same inputs.

markusicu:
I just finalized the PR description, deleting the debugging narrative and hoisting some of its details up. If you want to look at the narrative, please see the description's edit history.

PR #4019, opened 2026-06-05
ICU-23341 Support Gregorian era inheritance in Japanese calendar and …

yumaoka:
…DateFormat.

CLDR dropped Japanese calendar era data before Meiji. Also it introduced a new behavior that inheriting Gregorian calendar before Meiji (inheritEras: gregorian). This spec update requires following changes.

- DateFormatSmbols initialization code to load era data from calendar type specified by "inheritEras".
- Skip era index without definitions. (Japanese calendar inherit era 0 and 1 from Gregorian, but next era after 1 needs to be era 232 Meiji)

With these changes, this PR also removes `logKnownIssues` introduced by the previous CLDR data integration.

#### Checklist
- [x] Required: Issue filed: ICU-23341

### DIFF
--- a/icu4c/source/i18n/dtfmtsym.cpp
+++ b/icu4c/source/i18n/dtfmtsym.cpp
@@ -1896,6 +1896,31 @@ CalendarDataSink::~CalendarDataSink() {
 }
 
 //------------------------------------------------------
+static CharString
+&buildResourcePath(CharString &path, const char* segment1, UErrorCode &errorCode) {
+    return path.clear().append(segment1, -1, errorCode);
+}
+
+static CharString
+&buildResourcePath(CharString &path, const char* segment1, const char* segment2,
+                   UErrorCode &errorCode) {
+    return buildResourcePath(path, segment1, errorCode).append('/', errorCode)
+                                                       .append(segment2, -1, errorCode);
+}
+
+static CharString
+&buildResourcePath(CharString &path, const char* segment1, const char* segment2,
+                   const char* segment3, UErrorCode &errorCode) {
+    return buildResourcePath(path, segment1, segment2, errorCode).append('/', errorCode)
+                                                                 .append(segment3, -1, errorCode);
+}
+
+static CharString
+&buildResourcePath(CharString &path, const char* segment1, const char* segment2,
+                   const char* segment3, const char* segment4, UErrorCode &errorCode) {
+    return buildResourcePath(path, segment1, segment2, segment3, errorCode).append('/', errorCode)
+                                                                           .append(segment4, -1, errorCode);
+}
 
 static void
 initField(UnicodeString **field, int32_t& length, const char16_t *data, LastResortSize numStr, LastResortSize strLen, UErrorCode &status) {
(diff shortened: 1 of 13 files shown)


markusicu:
Java: please run spotless

C++: lots of failures, mostly on Windows; maybe missing U_I18N_API somewhere?
```
erarulestest.obj : error LNK2019: unresolved external symbol "public: int __cdecl icu_79::EraRules::getNextEraCode(int)const " (?getNextEraCode@EraRules@icu_79@@QEBAHH@Z) referenced in function "private: void __cdecl EraRulesTest::testJapanese(void)" (?testJapanese@EraRulesTest@@AEAAXXZ) [D:\a\icu\icu\icu4c\source\test\intltest\intltest.vcxproj]
```

yumaoka:
> Java: please run spotless

Reran spotless. I finished J version first and ran spotless. While working on C code, I want to make C and J code consistent as much as possible and modified J code.. and forgot to run spotless after that.

I think I need to install spotless plugin to IDE if available.

> C++: lots of failures, mostly on Windows; maybe missing U_I18N_API somewhere?

You're right. I missed U_I18N_API in a few places.

I updated the PR with the fixes. I also found integer 32 overflow warning in clang-ubsan build. This was caused by missing error status check before the reported location. I also fixed the issue. Anyway, I will check the updated CI build output.

richgillam (dismissed):
I'm not sure I'm the best one to review this, but it all looks plausible, and it does seem to be passing all the tests now...

robertbastian (commented):
tests lgtm

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
```diff
@@ -206,7 +241,47 @@ public int getMaxEraCode() {
...
+     * @return Next era code. If not available, maximum era code is returned.
+     */
+    public int getNextEraCode(int eraCode) {
+        int nextEra = eraCode + 1;
+        if (nextEra < minEra) {
+            if (inheritEraRules != null) {
+                if (nextEra > inheritEraRules.getMaxEraCode()) {
+                    nextEra = minEra;
+                } else {
+                    nextEra = inheritEraRules.getNextEraCode(eraCode);
```
markusicu, line 269: It looks like we could get stuck here:

If nextEra == inheritEraRules.getMaxEraCode(), with nextEra = eraCode + 1, then we go back to the one-less eraCode and then to inheritEraRules.getNextEraCode(eraCode) which ... returns inheritEraRules.getMaxEraCode()?

Am I missing something?
Is this covered by unit tests?

icu4c/source/i18n/erarules.cpp
```diff
@@ -103,22 +104,51 @@ static int32_t compareEncodedDateWithYMD(int encoded, int year, int month, int d
     }
 }
 
-EraRules::EraRules(LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn, int32_t minEraIn, int32_t numErasIn)
-    : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn) {
+EraRules::EraRules(const char* calTypeId, LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn,
+    int32_t minEraIn, int32_t numErasIn)
+    : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn), inheritEraRules(nullptr) {
+    uprv_strncpy(this->calTypeId, calTypeId, MAX_CAL_ID_LENGTH);
```
markusicu: We should try to avoid strncpy().
Are these strings all either hardcoded, that is, known, or length-checked upstream?
If so, then we can assume that the ID fits (`uprv_strcpy()`); maybe U_ASSERT() that they do (strlen).
If not, then we should check that the ID fits; if yes: memcpy()+NUL; if no: buffer-overflow-error.

icu4c/source/i18n/erarules.cpp
```diff
@@ -103,22 +104,51 @@ static int32_t compareEncodedDateWithYMD(int encoded, int year, int month, int d
     }
 }
 
-EraRules::EraRules(LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn, int32_t minEraIn, int32_t numErasIn)
-    : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn) {
+EraRules::EraRules(const char* calTypeId, LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn,
+    int32_t minEraIn, int32_t numErasIn)
+    : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn), inheritEraRules(nullptr) {
```
markusicu:
```suggestion
EraRules::EraRules(const char* calTypeId, LocalMemory<int32_t>& startDatesIn, int32_t startDatesLengthIn,
        int32_t minEraIn, int32_t numErasIn)
        : startDatesLength(startDatesLengthIn), minEra(minEraIn), numEras(numErasIn), inheritEraRules(nullptr) {
```

icu4c/source/i18n/erarules.cpp
```diff
@@ -103,22 +104,51 @@ static int32_t compareEncodedDateWithYMD(int encoded, int year, int month, int d
...
+    uprv_strncpy(this->calTypeId, calTypeId, MAX_CAL_ID_LENGTH);
+    this->calTypeId[MAX_CAL_ID_LENGTH] = 0;
     startDates = std::move(startDatesIn);
     initCurrentEra();
 }
 
 EraRules::~EraRules() {
+    if (inheritEraRules) {
```
markusicu: optional: unnecessary check: delete nullptr is fine

icu4c/source/i18n/erarules.cpp
```diff
@@ -257,19 +287,59 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
...
+        EraRules *tmpEraRules = createInstance(inheritCalType, includeTentativeEra, status);
+        if (tmpEraRules == nullptr) {
+            status = U_MEMORY_ALLOCATION_ERROR;
+            delete result;
+            return nullptr;
+        }
+        result->inheritEraRules = tmpEraRules;
+    }
+
     return result;
```
markusicu: if U_FAILURE then createInstance() should delete result and return nullptr.

icu4c/source/i18n/erarules.cpp
```diff
@@ -257,19 +287,59 @@ EraRules* EraRules::createInstance(const char *calType, UBool includeTentativeEr
     for (int32_t eraIdx = 0; eraIdx < eraStartDates.size(); eraIdx++) {
         startDates[eraIdx] = eraStartDates.elementAti(eraIdx);
     }
-    EraRules *result = new EraRules(startDates, eraStartDates.size(), minEra, numEras);
+    EraRules *result = new EraRules(calTypeId, startDates, eraStartDates.size(), minEra, numEras);
```
markusicu: would be safer to use a LocalPointer, and use its constructor that takes an extra UErrorCode which then auto-sets the OOM

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1957,48 +1982,77 @@ initField(UnicodeString **field, int32_t& length, CalendarDataSink &sink, CharSt
...
+
+    // Also load era names from resource corresponding to the inherited era rules if any.
+    const EraRules *inheritEraRules = eraRules->getInheritEraRules();
+    if (inheritEraRules != nullptr) {
+        loadEraNames(eras, length, eraNamesTable, calBaseBundle, eraWidth, inheritEraRules, status);
+    }
+}
+
+static void initEras(UnicodeString **field, int32_t& length, CalendarDataSink &sink, UResourceBundle *calBaseBundle,
+    const char *eraWidth, EraRules *eraRules, UErrorCode &status) {
```
markusicu: Continuation line should be indented so that it doesn't look like the start of the function body.
```suggestion
        const char *eraWidth, EraRules *eraRules, UErrorCode &status) {
```

icu4c/source/i18n/dtfmtsym.cpp
```diff
@@ -1957,48 +1982,77 @@ initField(UnicodeString **field, int32_t& length, CalendarDataSink &sink, CharSt
 }
 
 static void
-initEras(UnicodeString **field, int32_t& length, CalendarDataSink &sink, CharString &key, const UResourceBundle *ctebPtr, const char* eraWidth, int32_t maxEra, UErrorCode &status) {
-    if (U_SUCCESS(status)) {
-        length = 0;
-        UnicodeString keyUString(key.data(), -1, US_INV);
-        Hashtable *eraNamesTable = static_cast<Hashtable*>(sink.maps.get(keyUString));
+loadEraNames(UnicodeString *eras, int32_t length, Hashtable *eraNamesTable, UResourceBundle *calBaseBundle,
+    const char *eraWidth, const EraRules *eraRules, UErrorCode &status) {
```
markusicu: Continuation line should be indented so that it doesn't look like the start of the function body.
```suggestion
        const char *eraWidth, const EraRules *eraRules, UErrorCode &status) {
```

markusicu (commented):
- looks plausible
- small C++ issues
- I don't understand how the get-next-era code does not get into a loop

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
yumaoka: The test case is below -

https://github.com/unicode-org/icu/pull/4019/changes#diff-dba4fd54aad245317e580f0d7d79902ea8b7993272e77d874edcc28245ee56b9R92

icu4c/source/i18n/erarules.cpp
yumaoka: >We should try to avoid strncpy().
OK, I'll make the change.

>Are these strings all either hardcoded, that is, known, or length-checked upstream?
If so, then we can assume that the ID fits (uprv_strcpy()); maybe U_ASSERT() that they do (strlen).
If not, then we should check that the ID fits; if yes: memcpy()+NUL; if no: buffer-overflow-error.

calTypeId is coming from CLDR and used as a key in calendar rule data. Because it's only coming from data managed by us, we should not really need error checking, although I'll put an assert.

icu4c/source/i18n/erarules.cpp
yumaoka: OK, I will updated.

icu4j/main/core/src/main/java/com/ibm/icu/impl/EraRules.java
yumaoka: In case of Japanse calender

- EraRules has rule for Era 232 (Meiji - minimum era) to 236 (Reiwa - max era).
- EraRules has Gregorian EraRules as inheritEraRules
- Gregorian EraRules has two eras - 0 (BC - minmum era) and 1 (AD - max era).

getNextEraCode takes any arbitrary era code. The method add 1 to the input era code at the beginning.

- When eraCode is less than current EraRules minmum value (232 for Japanese calendar), checks with inheritEraRules (Gregorian).
  - If the value is greater than Gregorian's max era (1), then it returns current EraRule's minimum era. This is for jumping era Gregorian era 1 (AD) to Japanese era 232 (Meiji).
  - If the value is equal or less than Gregorian's max era (1), then it calls inheritEraRuls.getNextEra with the original input era code. Why does it do this? This design allows multiple EraRules inheritance (i.e. inheritEraRules may inherit another EraRules).

markusicu: Thanks... I think I got it now, and I think I got confused between the two layers.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/EraRulesTest.java
```diff
@@ -78,5 +79,34 @@ public void testJapanese() {
...
+        // Note: Current CLDR data uses 1868-10-23 as the start date of Meiji.
+        // This is not really accurate, because it does not count Lunar-solar
+        // calendar system used before Meiji 6. This might be changed in future.
+        assertEquals("1868-10-23", JapaneseCalendar.MEIJI, rules.getEraCode(1868, 10, 23));
+        assertEquals("1868-10-22", GregorianCalendar.AD, rules.getEraCode(1868, 10, 22));
+        assertEquals("0001-01-01", GregorianCalendar.AD, rules.getEraCode(1, 1, 1));
+        assertEquals("0000-12-31", GregorianCalendar.BC, rules.getEraCode(0, 12, 31));
+        assertEquals("-1000-01-01", GregorianCalendar.BC, rules.getEraCode(-1000, 1, 1));
+
+        final int[] expectedEraCodes = {0, 1, 232, 233, 234, 235, 236};
```
markusicu, line 92: How about also testing some weird inputs like -1, 2, 55, 231, 237, 238, 240, 555?

markusicu: Actually, you already have a middle value and maxEra, but some more edge cases might be nice.

icu4c/source/i18n/erarules.cpp
yumaoka: Done

icu4c/source/i18n/erarules.cpp
yumaoka: Done

icu4c/source/i18n/erarules.cpp
yumaoka: Dropped if.

icu4c/source/i18n/erarules.cpp
yumaoka: Done along with LocalPointer change

icu4c/source/i18n/erarules.cpp
yumaoka: Changed to LocalPointer.

icu4c/source/i18n/dtfmtsym.cpp
yumaoka: Done

icu4c/source/i18n/dtfmtsym.cpp
yumaoka: Done

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/calendar/EraRulesTest.java
yumaoka: Added a few more cases / minEra - 1 and maxEra + 1. These additional test cases should cover all conditions.

markusicu (approved):
thanks!

richgillam (approved):
Still LGTM.

PR #4025, opened 2026-06-14
ICU-23247 Fix error MSB8036: The Windows SDK version 10.0.0.0 was not found

mihnita:
All of this is to handle VS 2026, which is now the default Windows runner on GitHub.

---

VS 2026 (Platform Toolset = 'v145') dropped support for 32-bit ARM.

I changed the CI tests to check ARM64 instead, so ARM related problems will still be detected.
Users can build ARM 32 on their own machines, we just don't test it in CI anymore.

---

Although there seems to be a version missing (v144) in `Build.Windows.PlatformToolset.props`, that is not a mistake.

See https://developercommunity.microsoft.com/t/11019486

> _The error occurs because there is **no official v144 Platform Toolset** in VS. While the latest updates to VS 2022 used a compiler versioned 14.4x, the build system continued to use the label **v143** for compatibility. With your upgrade to **VS 2026 (v18.0)**, Microsoft has moved the primary toolset version to **v145**. Because your project file is explicitly looking for ‘**v144**’—a version that does not exist in the software’s registry—MSBuild is unable to locate the necessary build tools to complete the process._

---

#### Checklist

### DIFF
--- a/.github/workflows/icu4c.yml
+++ b/.github/workflows/icu4c.yml
@@ -400,7 +400,7 @@ jobs:
           - test_flags: 'x86 Debug'
             build_flags: '/p:Configuration=Debug /p:Platform=Win32'
           - test_flags: 'arm Release'
-            build_flags: '/p:Configuration=Release /p:Platform=ARM'
+            build_flags: '/p:Configuration=Release /p:Platform=ARM64'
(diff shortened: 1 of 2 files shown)


markusicu (approved):
thank you!!

PR #4033, opened 2026-06-17
ICU-23436 Custom properties in RBBI

eggrobin:
For use in the unicodetools to prepare publishable state tables; see https://github.com/unicode-org/properties/issues/567.

#### Checklist
- [x] Required: Issue filed: ICU-23436

### DIFF
--- a/icu4j/main/core/src/main/java/com/ibm/icu/text/RBBIRuleScanner.java
+++ b/icu4j/main/core/src/main/java/com/ibm/icu/text/RBBIRuleScanner.java
@@ -12,6 +12,7 @@ import com.ibm.icu.impl.Assert;
 import com.ibm.icu.impl.Utility;
 import com.ibm.icu.lang.UCharacter;
 import com.ibm.icu.lang.UProperty;
+import com.ibm.icu.text.UnicodeSet.XSymbolTable;
 import java.text.ParsePosition;
 import java.util.HashMap;
 
@@ -64,8 +65,11 @@ class RBBIRuleScanner {
 
     boolean fNoChainInRule; // True if the current rule starts with a '^'.
 
-    RBBISymbolTable fSymbolTable; // symbol table, holds definitions of
-    //   $variable symbols.
+    // Symbol table, holds definitions of $variable symbols.
+    RBBISymbolTable fSymbolTable;
+    // Either fSymbolTable, or an RBBIXSymbolTable with overriden properties,
+    // but whose variables are those of fSymbolTable.
+    SymbolTable fXSymbolTable;
 
     HashMap<String, RBBISetTableEl> fSetTable =
             new HashMap<>(); // UnicocodeSet hash table, holds indexes to
@@ -99,7 +103,7 @@ class RBBIRuleScanner {
     //  Constructor.
     //
     // ----------------------------------------------------------------------------------------
-    RBBIRuleScanner(RBBIRuleBuilder rb) {
+    RBBIRuleScanner(RBBIRuleBuilder rb, XSymbolTable customProperties) {
         fRB = rb;
         fLineNum = 1;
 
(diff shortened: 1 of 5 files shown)


icu4j/main/core/src/main/java/com/ibm/icu/text/RBBIRuleScanner.java
```diff
@@ -1118,4 +1127,43 @@ void scanSet() {
...
+        public UnicodeSet lookupSet(String s) {
+            return variableTable.lookupSet(s);
+        }
+
+        @Override
+        public String parseReference(String text, ParsePosition pos, int limit) {
+            return variableTable.parseReference(text, pos, limit);
+        }
+
+        // Note that we do not override XSymbolTable#scanVariable: we retain th
```
markusicu:
```suggestion
        // Note that we do not override XSymbolTable#scanVariable: we retain the
```

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBIRuleScanner.java
```diff
@@ -64,8 +65,11 @@ static class RBBIRuleChar {
 
     boolean fNoChainInRule; // True if the current rule starts with a '^'.
 
-    RBBISymbolTable fSymbolTable; // symbol table, holds definitions of
-    //   $variable symbols.
+    // Symbol table, holds definitions of $variable symbols.
+    RBBISymbolTable fSymbolTable;
```
markusicu, line 69: Does this still need to be a class field?
Should it ever be used directly, or should it only be used via the fXSymbolTable which is either an alias or a wrapper?

If we don't need a field for the RBBISymbolTable, then maybe we can retain the old field name fSymbolTable and make it type SymbolTable; store the alias or wrapper there.

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java
```diff
@@ -1342,4 +1344,50 @@ public void TestLookaheadPolychromy() {
...
+        s.clear();
+        shakespeareanSymbolTable.applyPropertyAlias("Name", "Hippolita", s);
+        final String hippolyta = Character.toString(s.charAt(0));
+        s.clear();
+        shakespeareanSymbolTable.applyPropertyAlias("Name", "Moth", s);
+        final String moth = Character.toString(s.charAt(0));
+        final var it =
+                new RuleBasedBreakIterator(
+                        "[\\N{Theſeus}] [\\N{Hippolita}]; .;", shakespeareanSymbolTable);
+        it.setText(moth + theseus + hippolyta + moth);
```
markusicu, line 1363: totally optional:
you could construct the string from the code points, rather than creating a single-character string for each of them first;
if they are all BMP code points, you could use a spiffier constructor

String​(char[] value)	
Allocates a new String so that it represents the sequence of characters currently contained in the character array argument.

String​(int[] codePoints, int offset, int count)	
Allocates a new String that contains characters from a subarray of the Unicode code point array argument.

icu4j/main/core/src/main/java/com/ibm/icu/text/RBBIRuleScanner.java
eggrobin: I don’t think that works from an OO standpoint: sometimes we use the RBBI-specific features of RBBISymbolTable, but fXSymbolTable is not necessarily an RBBISymbolTable (an RBBIXSymbolTable is not an RBBISymbolTable).

icu4j/main/core/src/test/java/com/ibm/icu/dev/test/rbbi/RBBITest.java
eggrobin: For the first one, I would need four ugly casts from int to char (because UnicodeSet.charAt returns an int). For the second one, I would need ugly indices. So in this one-off testonly case, I do not think this improves readability.

markusicu (dismissed):

robertbastian (approved):
.

PR #4034, opened 2026-06-18
ICU-23016 Bail out CollationFastLatin when it encounters a digit and …

yumaoka:
…digits are reordered.

When order of Latin letters and digits are changed by colReorder (kr) option, the comparison code in CollationFastLatin cannot handle it. The implementation was designed to bail out whenever it encounters the situation. However, it failed to detect this condition and returned an incorrect result.

To fix the issue, when digits are reordered, then CollationFastLatin::getOptions() sets the CollationSettings::NUMERIC bit, which in the fast-Latin code is only used to bail out when digits are encountered.

#### Checklist
- [x] Required: Issue filed: ICU-23016

### DIFF
--- a/icu4c/source/i18n/collationfastlatin.cpp
+++ b/icu4c/source/i18n/collationfastlatin.cpp
@@ -101,8 +101,13 @@ CollationFastLatin::getOptions(const CollationData *data, const CollationSetting
         for(UChar32 c = 0x30; c <= 0x39; ++c) { primaries[c] = 0; }
     }
 
+    // NUMERIC flag is also used for indicating digits are reordered
+    // in fast-Latin options.
+    int options = digitsAreReordered ?
+        settings.options | CollationSettings::NUMERIC : settings.options;
+
     // Shift the miniVarTop above other options.
-    return (static_cast<int32_t>(miniVarTop) << 16) | settings.options;
+    return (static_cast<int32_t>(miniVarTop) << 16) | options;
 }
 
 int32_t
--- a/icu4c/source/i18n/collationsettings.h
+++ b/icu4c/source/i18n/collationsettings.h
@@ -42,6 +42,10 @@ struct U_I18N_API CollationSettings : public SharedObject {
      *
      * Treat digit sequences as numbers with CE sequences in numeric order,
      * rather than returning a normal CE for each digit.
+     * 
+     * This bit is also used for indicating that digits are reordered in the fast-Latin code.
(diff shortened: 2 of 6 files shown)


icu4c/source/i18n/collationfastlatin.cpp
```diff
@@ -138,7 +142,9 @@ CollationFastLatin::compareUTF16(const uint16_t *table, const uint16_t *primarie
             if(c <= LATIN_MAX) {
                 leftPair = primaries[c];
                 if(leftPair != 0) { break; }
-                if(c <= 0x39 && c >= 0x30 && (options & CollationSettings::NUMERIC) != 0) {
+                if(c <= 0x39 && c >= 0x30
+                    && (((options & CollationSettings::NUMERIC) != 0) ||
+                         (options & CollationFastLatin::DIGITS_ARE_REORDERED) != 0)) {
```
markusicu: Looking at these changes, and looking at this file as a whole, we don't use NUMERIC for anything other than this bailout.
This suggests that we can simplify: Instead of using options bit 7, when digitsAreReordered we could simply set the NUMERIC bit. It would be kind of a "lie", but within the fast-Latin path it would just mean "bail out for digits".
WDYT?

markusicu (commented):
good fix, and as discussed, but looking at the actual changes, I think we can simplify...

icu4c/source/i18n/collationfastlatin.cpp
yumaoka: I actually considered this at first. I'm not confident that the NUMERIC bit is used for other purpose. In other words, something which is applicasble only when NUMRIC collation is used, but not collation reordering affecting ASCII digits. For fast latin path, I know it's only used for the purpose. But the regular path, I am not sure about it. Are you sure it's safe?

yumaoka: OK - never mind. The option bit map is only used for fast latin collation. I'll make the change.

yumaoka: @markusicu I've just updated the PR. Please review it again. Thanks

markusicu: Thanks! I updated the PR description and am looking at the changes again now.

markusicu (approved):
good and spiffy!

markusicu:
the commit message is a bit funky now...

PR #4037, opened 2026-06-20
ICU-23425 Fix broken links in documentation site

mihnita:
Most fixes only required a change in relative paths.

* A few web sites don't exist anymore. I replaced them with what I thought are good alternatives.
* In some cases I had to remove full paragraphs (for example mentions of samples in `icu4c\source\samples`).
* Some of the releasing steps depend on the ICU version.
And when the documentation was `"modify the [foo.bar](link) file"` mentioning a file that
was deleted in the meantime, I just removed the link, letting the file path be just text.
* There are still a few external links that are flaky (don't always answer), tbd what to do about them.

#### Checklist
- [x] Required: Issue filed: ICU-23425

### DIFF
--- a/docs/devsetup/source/index.md
+++ b/docs/devsetup/source/index.md
@@ -102,7 +102,7 @@ There are many resources available to help you work with git, here are a few:
 *   <https://try.github.io/> - Resources to learn Git
 
 Want to contribute back to ICU? See
-[How to contribute](../../userguide/processes/contribute.md).
+[How to contribute](../../userguide/dev/index.md).
 
 ## Repository Layout
 
@@ -138,7 +138,7 @@ There are other tags and branches which may be cleaned up/deleted at any time.
 *   long running shared feature branches (In general, feature work is done on
     personal forks of the repository.)
 
-See also the [Tips (for developers)](repository/tips/index.md) subpage.
+See also the [git and Github for ICU Developers](../../userguide/dev/gitdev.md) subpage.
 
 ## A Bit of History
 
--- a/docs/download/76.md
+++ b/docs/download/76.md
@@ -255,7 +255,7 @@ We accept patches for other platforms.
 For ICU 76, we have received a contribution to make ICU4C work again on z/OS,
 using a newer (clang-based) compiler. ([ICU-22714](https://unicode-org.atlassian.net/browse/ICU-22714) [icu/pull/3008](https://github.com/unicode-org/icu/pull/3008) + [ICU-22916](https://unicode-org.atlassian.net/browse/ICU-22916) [icu/pull/3208](https://github.com/unicode-org/icu/pull/3208))
 
-Windows: The minimum supported version is Windows 7. (See [How To Build And Install On Windows](../userguide/icu4c/build.html#how-to-build-and-install-on-windows) for more details.)
+Windows: The minimum supported version is Windows 7. (See [How To Build And Install On Windows](../userguide/icu4c/build#how-to-build-and-install-on-windows) for more details.)
 
 ## ICU4J Platform Support
 
--- a/docs/processes/release/tasks/integration.md
+++ b/docs/processes/release/tasks/integration.md
@@ -372,8 +372,7 @@ Instructions for verifying the XLIFF conversion tools.
 ## Test sample and demo programs
 
 Build and run all of the sample and demo apps that are included with ICU, on
-each of the reference platforms. A list of them is in the
-[readme](http://icu.sanjose.ibm.com/internal/checklist/icu-progs-list.html).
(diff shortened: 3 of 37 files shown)


mihnita:
Original report: https://mihai-nita.net/tmp/icu/report_fails.html

Everything is fixed, with 2 exceptions:
 * http://www.gnu.org/software/autoconf/
 * http://www.gnu.org/software/autoconf/manual/html_node/Specifying-Names.html#Specifying-Names

It looks like gnu.org is down:
* https://www.reddit.com/r/gnu/comments/1ub6giu/cant_access_gnuorg/
* https://uptime.com/gnu.org

I decided to wait for that to be fixed, instead of looking for a mirror.
I think that gnu.org is too important to just go away for good, and it is very likely to come back.

docs/devsetup/source/index.md
```diff
@@ -102,7 +102,7 @@ There are many resources available to help you work with git, here are a few:
 *   <https://try.github.io/> - Resources to learn Git
 
 Want to contribute back to ICU? See
-[How to contribute](../../userguide/processes/contribute.md).
+[How to contribute](../../userguide/dev/index.md).
```
markusicu, line 105: Should this go to https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md?

docs/download/76.md
```diff
@@ -255,7 +255,7 @@ We accept patches for other platforms.
 For ICU 76, we have received a contribution to make ICU4C work again on z/OS,
 using a newer (clang-based) compiler. ([ICU-22714](https://unicode-org.atlassian.net/browse/ICU-22714) [icu/pull/3008](https://github.com/unicode-org/icu/pull/3008) + [ICU-22916](https://unicode-org.atlassian.net/browse/ICU-22916) [icu/pull/3208](https://github.com/unicode-org/icu/pull/3208))
 
-Windows: The minimum supported version is Windows 7. (See [How To Build And Install On Windows](../userguide/icu4c/build.html#how-to-build-and-install-on-windows) for more details.)
+Windows: The minimum supported version is Windows 7. (See [How To Build And Install On Windows](../userguide/icu4c/build#how-to-build-and-install-on-windows) for more details.)
```
markusicu, line 258: Here and elsewhere: Should we have the .md suffix?
I believe that some versions of links end up displaying the right page but not connecting with the nav bar.

docs/processes/release/tasks/integration.md
```diff
@@ -372,8 +372,7 @@ Instructions for verifying the XLIFF conversion tools.
 ## Test sample and demo programs
 
 Build and run all of the sample and demo apps that are included with ICU, on
-each of the reference platforms. A list of them is in the
-[readme](http://icu.sanjose.ibm.com/internal/checklist/icu-progs-list.html).
```
markusicu, line 376: one of the best domain names ever ;-)

docs/userguide/datetime/index.md
```diff
@@ -81,7 +81,7 @@ details.*
 ### References
 
 The ICU system time zones are derived from the tz database (also known as the
-“Olson” database) at [ftp://elsie.nci.nih.gov/pub](ftp://elsie.nci.nih.gov/pub).
+“Olson” database) at [https://www.iana.org/time-zones](https://www.iana.org/time-zones).
```
markusicu: optional / possible simplification:
```suggestion
“Olson” database) at <https://www.iana.org/time-zones>.
```

docs/download/76.md
mihnita: I checked, and we are not consistent, we use a mixture, sometimes with sometimes without `.md`

I checked the generated site and both kinds of links work. So I decided to go without `.md` in all my fixes, somehow arbitrarily.

markusicu: Please double-check that both versions end up showing the nav bar with the current page open.

docs/devsetup/source/index.md
mihnita: I went with that because it takes you here:
https://unicode-org.github.io/icu/userguide/dev/

Which has entries for "Coding Guidelines", git for devs, etc.

The info overlaps quite a bit with https://github.com/unicode-org/icu/blob/main/CONTRIBUTING.md, but I favored staying on the same site.

I don't feel strongly about it. Let me know.

docs/download/76.md
mihnita: It works.
You can see / check it here:
https://mihnita.github.io/icu/download/76.html#icu4c-platform-support

> Windows: The minimum supported version is Windows 7. (See [How To Build And Install On Windows](https://mihnita.github.io/icu/userguide/icu4c/build#how-to-build-and-install-on-windows) for more details.)

markusicu (approved):

PR #4039, opened 2026-06-24
ICU-23425 More broken links, Jekyll bug

mihnita:
#### Checklist
- [x] Required: Issue filed: ICU-23425

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -55,7 +55,7 @@ color_scheme: nil
 # Aux links for the upper right navigation
 aux_links:
   "API Docs":
-    - "https://unicode-org.github.io/icu-docs"
+    - "https://unicode-org.github.io/icu-docs/"
   "ICU on GitHub":
     - "https://github.com/unicode-org/icu"
   "ICU Home Page":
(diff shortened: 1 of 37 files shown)


mihnita:
Testing the resulting online site the W3C link checker found more problems.

And some are in fact bug in Jekyll :-(

For example this works:
```md
See [ICU Data](../icu_data/index.md)
```

This does not:
```md
See [ICU
Data](../icu_data/index.md)
```

In the second case Jekyll does not change the `.md` to `.html` (leaves it `.md`) and results in a broken link.
Links with the visible text wrapped work fine and the proper `<a href>...</a>` is generated and works for external links.

Worse, I found Ruby + Jekyll virtually impossible to install and configure to the point where I can use it to generate the documentation offline.
I tried in on macOS, Windows, Ubuntu, Debian, and I wasted hours.

So the only way to fix things was to make changes, publish with an action on my own repo, then validate that site.

Add to that that many of the link checkers miss some kinds of links, or are limited to a certain number of links unless you subscribe, etc.

In the end https://validator.w3.org/checklink was very good, and free.
It is pretty slow, and the result is very verbose and hard to read, no a clean table.
But at least is not limited, and seems to detect most (all?) problems, including redirects.

---

Anyway, here is the last run of that action, corresponding to this PR:
https://github.com/mihnita/icu/actions/runs/28087248025

With the results here: https://mihnita.github.io/icu/

markusicu (approved):

mihnita:
Thank you!

PR #4042, opened 2026-06-25
Compute Arabic shaping scratch buffer sizes in size_t

ubeddulla:
The Arabic shaping helpers in `ushape.cpp` size their scratch buffers with `(sourceLength + 1) * U_SIZEOF_UCHAR` (and `(destSize + 1) * U_SIZEOF_UCHAR`) computed in 32-bit `int`, where the length ultimately comes from the `sourceLength` passed to the public `u_shapeArabic`; a large input wraps the multiply and the allocation that follows is too small for the writes done afterwards. `handleGeneratedSpaces` already has a `size_t` overflow guard, but it still does the multiply in `int`, so the value overflows before the guard can help on LP64, while `expandCompositCharAtBegin`, `expandCompositCharAtEnd` and `expandCompositChar` have no guard at all. This computes the sizes in `size_t` and adds the same guard to the three sibling helpers.

#### Checklist
- [ ] Required: The PR title must be prefixed with a JIRA Issue number. Example: "ICU-NNNNN Fix xyz"

### DIFF
--- a/icu4c/source/common/ushape.cpp
+++ b/icu4c/source/common/ushape.cpp
@@ -752,7 +752,7 @@ handleGeneratedSpaces(char16_t *dest, int32_t sourceLength,
         *pErrorCode = U_INDEX_OUTOFBOUNDS_ERROR;
         return 0;
     }
-    tempbuffer = static_cast<char16_t*>(uprv_malloc((sourceLength + 1) * U_SIZEOF_UCHAR));
+    tempbuffer = static_cast<char16_t*>(uprv_malloc((static_cast<size_t>(sourceLength) + 1) * U_SIZEOF_UCHAR));
     /* Test for nullptr */
     if(tempbuffer == nullptr) {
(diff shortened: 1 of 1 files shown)


ubeddulla:
I don't have Jira access to open an ICU ticket or get one accepted on my own, so I can't add the ICU-NNNNN prefix myself. If you can file one (or point me at an existing ticket), I'll update the PR title and reword the commit to match.

This overlaps with #4043 (same file, sibling helpers), so a shared ticket would be fine if that's simpler on your end.

markusicu:
> I don't have Jira access to open an ICU ticket or get one accepted on my own, so I can't add the ICU-NNNNN prefix myself.

Please create a Jira account and then a ticket: https://icu.unicode.org/bugs

> This overlaps with #4043 (same file, sibling helpers), so a shared ticket would be fine if that's simpler on your end.

A combined ticket for both sounds fine.

PR #4044, opened 2026-06-25
ICU-23429 MessageFormat: Implement :offset and :percent formatters and selectors

catamorphism:
Implement the :offset and :percent formatters and selectors
according to the spec for Unicode MessageFormat.

#### Checklist
- [x] Required: Issue filed: ICU-23429

### DIFF
--- a/icu4c/source/i18n/messageformat2_function_registry.cpp
+++ b/icu4c/source/i18n/messageformat2_function_registry.cpp
@@ -143,6 +143,7 @@ void MFFunctionRegistry::checkStandard() const {
     checkFunction(functions::CURRENCY);
     checkFunction(functions::NUMBER);
     checkFunction(functions::INTEGER);
+    checkFunction(functions::OFFSET);
     checkFunction(functions::STRING);
     checkFunction(functions::TEST_FUNCTION);
     checkFunction(functions::TEST_FORMAT);
@@ -387,7 +388,9 @@ bool isDigitSizeOption(std::u16string_view s) {
         || s == options::MINIMUM_FRACTION_DIGITS
         || s == options::MAXIMUM_FRACTION_DIGITS
         || s == options::MINIMUM_SIGNIFICANT_DIGITS
-        || s == options::MAXIMUM_SIGNIFICANT_DIGITS;
+        || s == options::MAXIMUM_SIGNIFICANT_DIGITS
+        || s == options::ADD
+        || s == options::SUBTRACT;
 }
 
 /* static */ void StandardFunctions::validateDigitSizeOptions(const FunctionOptions& opts,
@@ -536,6 +539,16 @@ StandardFunctions::Number::number(UErrorCode& success) {
 }
 
 /* static */ StandardFunctions::Number*
+StandardFunctions::Number::offset(UErrorCode& success) {
+    return create(NumberType::kOffset, success);
+}
+
+/* static */ StandardFunctions::Number*
+StandardFunctions::Number::percent(UErrorCode& success) {
+    return create(NumberType::kPercent, success);
+}
+
+/* static */ StandardFunctions::Number*
 StandardFunctions::Number::create(StandardFunctions::NumberType numberType, UErrorCode& success) {
     NULL_ON_ERROR(success);
 
@@ -603,7 +616,7 @@ static number::Precision significantDigitsWithRoundingPriority(int32_t min, int3
     return result;
 }
 
-number::Precision StandardFunctions::withRoundingIncrement(const FunctionOptions& opts, bool& hasIncrement, const DigitSizeOption& fractionDigits, const UChar* currency, UErrorCode& status) {
+number::Precision StandardFunctions::withRoundingIncrement(const FunctionOptions& opts, bool& hasIncrement, int32_t minFractionDigits, UErrorCode& status) {
     using namespace number;
 
     UErrorCode localStatus = U_ZERO_ERROR;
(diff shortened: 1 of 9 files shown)


catamorphism:
@mihnita Would you be able to take a look? I pinged @srl295 for a review, but he said he's traveling.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -618,28 +631,14 @@ number::Precision StandardFunctions::withRoundingIncrement(const FunctionOptions
...
-    }
     double incrementToUse = U_SUCCESS(localStatus) ? increment / 100.0 : 0.01;
-    return Precision::increment(incrementToUse).withMinFraction(minFractionToUse);
+    return Precision::increment(incrementToUse).withMinFraction(minFractionDigits);
 }
 
 /* static */ number::LocalizedNumberFormatter StandardFunctions::formatterForOptions(const Number& number,
                                                                                      const Locale& locale,
                                                                                      const FunctionOptions& opts,
+bool usePercent,
```
srl295: Indentation here

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -701,38 +700,67 @@ number::Precision StandardFunctions::withRoundingIncrement(const FunctionOptions
...
 
-            if (number.usePercent(opts)) {
+            if (number.numberType == NumberType::kPercent && usePercent) {
                 nf = nf.unit(NoUnit::percent()).scale(Scale::powerOfTen(2));
             }
         }
 
         int32_t maxSignificantDigits = number.maximumSignificantDigits(opts);
-        if (number.numberType != NumberType::kInteger) {
+        if (number.numberType != NumberType::kInteger && number.numberType != NumberType::kOffset) {
```
srl295: thinking of the select crasher.. might be better to have an explicit == here instead of !='s, to guard if a new number type is added.

icu4c/source/i18n/messageformat2_function_registry.cpp
```diff
@@ -701,38 +700,67 @@ number::Precision StandardFunctions::withRoundingIncrement(const FunctionOptions
...
+                    // Divide by 100 to get cents-based rounding
+                    // The default for currency formatting is the number of minor
+                    // unit digits provided by the ISO 4217 currency code list
+                    // (2 if the list doesn't provide that information)
+                    UErrorCode localStatus = U_ZERO_ERROR;
+                    if (fractionDigits.isAuto() || fractionDigits.isInvalid()) {
+                        minFractionDigits = ucurr_getDefaultFractionDigits(currency.getBuffer(), &localStatus);
+                        if (U_FAILURE(localStatus)) {
+                            localStatus = U_ZERO_ERROR;
+                            minFractionDigits = 2;
```
srl295: This ought to be a ucurr constant, default digits.

srl295 (commented):
No major blockers.
- the one indent would be good to fix. 
- at least one place could be a == opt-in list instead of a double != opt out (for future proofing) 
- future, i think the currency default of 2 digits should be shared w/ number formatting
- and in general, there's a lot of re-jiggering of the number formatter for currency- are there ways to use the currency formatting API more? Or are these API mismatches between ICU and MF2?  Or maybe add more APIs to ICU formatting to match MF2 surface?

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Fixed

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Fixed

icu4c/source/i18n/messageformat2_function_registry.cpp
catamorphism: Fixed

catamorphism:
> No major blockers.
> 
> * the one indent would be good to fix.
> * at least one place could be a == opt-in list instead of a double != opt out (for future proofing)
> * future, i think the currency default of 2 digits should be shared w/ number formatting

Fixed these.

> * and in general, there's a lot of re-jiggering of the number formatter for currency- are there ways to use the currency formatting API more? Or are these API mismatches between ICU and MF2?  Or maybe add more APIs to ICU formatting to match MF2 surface?

I see two places that change the number formatter for currency: the definition of `StandardFunctions::withRoundingIncrement()` and the code in `StandardFunctions::formatterForOptions()` that determines the minimum fraction digits. I added more comments in both of those places. I guess this code could be moved into the currency formatting API, but that would be a change to existing APIs, which would require a design proposal. Perhaps that could be done in the future? If there's other code besides those two places that you had in mind with this comment, let me now.

catamorphism:
@moconor Feel free to take a look as well.

srl295:
> > No major blockers.
> 
> > 
> 
> > * the one indent would be good to fix.
> 
> > * at least one place could be a == opt-in list instead of a double != opt out (for future proofing)
> 
> > * future, i think the currency default of 2 digits should be shared w/ number formatting
> 
> 
> 
> Fixed these.
> 

Taking a look

> 
> 
> > * and in general, there's a lot of re-jiggering of the number formatter for currency- are there ways to use the currency formatting API more? Or are these API mismatches between ICU and MF2?  Or maybe add more APIs to ICU formatting to match MF2 surface?
> 
> 
> 
> I see two places that change the number formatter for currency: the definition of `StandardFunctions::withRoundingIncrement()` and the code in `StandardFunctions::formatterForOptions()` that determines the minimum fraction digits. I added more comments in both of those places. I guess this code could be moved into the currency formatting API, but that would be a change to existing APIs, which would require a design proposal. Perhaps that could be done in the future? If there's other code besides those two places that you had in mind with this comment, let me now.
> 
> 

Definitely future.
I just think these concerns should live with the numebrformat code where possible

srl295 (approved):

catamorphism:
Created https://unicode-org.atlassian.net/browse/ICU-23515 to track the future refactoring.

